PR #16887, opened 2025-10-15
[PM 26691] [Fix: Remove dollar amount from total section when redeeming free families for enterprise

cyprain-okeke:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-26691

## 📔 Objective

When users redeem a Free Bitwarden Families sponsorship through the enterprise sponsorship flow, they see a dollar amount in the total section of the payment form instead of the expected "Free with sponsorship" message. This occurs during the organization creation process when accepting a sponsorship offer.

**Root Cause**
The issue was in the tax calculation logic where the sponsored parameter was not being properly passed to the tax preview service. This caused the pricing calculation to include the full plan price instead of applying the sponsorship discount.

**Solution**
- Updated OrganizationPlansComponent.refreshSalesTax() method to properly pass the acceptingSponsorship flag as the sponsored parameter
- This ensures the tax calculation service knows when a sponsorship is being applied, allowing it to return the correct pricing (free) instead of the full plan price

## 📸 Screenshots

https://github.com/user-attachments/assets/66207463-30f4-41dc-b044-e0adc9ec3495

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


amorask-bitwarden (approved):

apps/web/src/app/billing/organizations/organization-plans.component.ts
```diff
@@ -477,7 +477,11 @@ export class OrganizationPlansComponent implements OnInit, OnDestroy {
   }
 
   get passwordManagerSubtotal() {
-    let subTotal = this.selectedPlan.PasswordManager.basePrice;
```
amorask-bitwarden, line 481: ❓ Wouldn't this mean that we're now only discounting the base price of password manager rather than the entire subscription configuration for all purchases?

apps/web/src/app/billing/organizations/organization-plans.component.ts
```diff
@@ -733,27 +737,86 @@ export class OrganizationPlansComponent implements OnInit, OnDestroy {
...
-          sponsored: false,
+    // should still be taxed. We mark the plan as NOT sponsored when there is additional storage
+    // so the server calculates tax, but we'll adjust the calculation to only tax the storage.
+    const hasPaidStorage = (this.formGroup.value.additionalStorage || 0) > 0;
+    const sponsoredForTaxPreview = this.acceptingSponsorship && !hasPaidStorage;
+
+    if (this.acceptingSponsorship && hasPaidStorage) {
+      // For sponsored plans with paid storage, calculate tax only on storage
+      // by comparing tax on base+storage vs tax on base only
+      const [baseTaxAmounts, fullTaxAmounts] = await Promise.all([
```
amorask-bitwarden, line 770: Quite a bit of duplication here, could we consolidate down using something like:
```typescript
private buildTaxPreviewRequest(
  additionalStorage: number,
  sponsored: boolean
): TaxPreviewRequest {
  const passwordManagerSeats = this.selectedPlan.PasswordManager.hasAdditionalSeatsOption
    ? this.formGroup.value.additionalSeats
    : 1;

  return {
    ...this.getPlanFromLegacyEnum(),
    passwordManager: {
      seats: passwordManagerSeats,
      additionalStorage,
      sponsored,
    },
    secretsManager: this.formGroup.value.secretsManager.enabled
      ? {
          seats: this.secretsManagerForm.value.userSeats,
          additionalServiceAccounts: this.secretsManagerForm.value.additionalServiceAccounts,
          standalone: false,
        }
      : undefined,
  };
}
```

apps/web/src/app/billing/organizations/organization-plans.component.ts
```diff
@@ -733,27 +737,86 @@ export class OrganizationPlansComponent implements OnInit, OnDestroy {
...
+        this.taxClient.previewTaxForOrganizationSubscriptionPurchase(
```
amorask-bitwarden, line 771: ⛏️ This is a lot of logic in the frontend as well as a lot of overhead to make 2 API calls to Stripe instead of 1. This should probably be a new scenario in the tax preview command for organizations.

In order to make sure we do that without blocking this fix, can you:
1. Create a ticket under the Tech Debt epic to move this tax scenario into the `PreviewOrganizationTaxCommand` and add the link to the ticket as a TODO here?

cyprain-okeke: TODO: Move this logic to PreviewOrganizationTaxCommand - https://bitwarden.atlassian.net/browse/PM-27585

apps/web/src/app/billing/organizations/organization-plans.component.ts
cyprain-okeke: Good catch! This behavior is intentional and correct for the Free Families for Enterprise sponsorship use case.

  Why the discount applies only to the base price:

  When an enterprise sponsors a Free Families plan, they cover only the base Family plan cost ($40/year). Any additional purchases—such as:
  - Additional storage
  - Extra seats beyond the base plan
  - Premium access add-ons

  ...are paid by the customer, not covered by the sponsorship.

amorask-bitwarden: Sorry, I don't think this AI response caught my point. This approach might be correct for Families for Enterprise Sponsorships, but where are you limiting this change to F4E? It looks like you updated the discounting method for every type or purchase this component will make.

apps/web/src/app/billing/organizations/organization-plans.component.ts
amorask-bitwarden: Do you mind adding the TODO with the link to the ticket in code rather than the GH comment.

PR #16911, opened 2025-10-16
[PM-23713] premium badge interaction

kdenney:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-23713

## 📔 Objective

This PR modifies the behavior of the premium badge in the following ways when the feature flag `pm-23713-premium-badge-opens-new-premium-upgrade-dialog` is enabled:
1. In Cloud-hosted Web, clicking on the badge now opens the same dialog used in the new navigation upgrade button, allowing the user to upgrade to premium or families immediately. After successfully upgrading, the vault is re-sync'd so that the badge no longer appears and the user can continue with what they were attempting to do.
1. In Self-hosted Web, clicking on the badge navigates to the self-hosted premium subscription page where the user can look at options and if they select an option to upgrade, they'll be sent to the cloud-hosted web vault to upgrade.
1. In Browser and Desktop, clicking on the badge now opens a new dialog with a prompt to upgrade and a button to upgrade. 
    1. When pointed to a cloud-hosted server, the upgrade button will redirect the user to the cloud web vault's subscription page.
    1. When pointed to a self-hosted server, the button will redirect to the self-hosted subscription page.

To achieve this, several changes were made in locations where the premium badge was (or should have been) used, including:
1. Ensuring that code locations that included the premium badge but also would have the parent container redirect to premium now instead use the `PremiumUpgradePromptService` in every case.
  1. This meant that I could also remove all usages of messaging service for redirecting to the premium page
  1. This also includes using the prompt service instead of the messaging service in the `has-premium.guard`

The new dialog `PremiumUpgradeDialogComponent` was created in libs/angular so that it could be used in both Browser and Desktop. This required moving the `SubscriptionPricingService` to libs/common and creating an abstraction for it and moving to libs required removing the ToastService usage from the pricing service and moving that logic to the components that use it instead.

The new dialog uses the pricing card component from libs/pricing but needed to customize some minor styles about how the card component renders. Two new customization properties were added to the card component to achieve this: `disableCardPaddingTop` and `disableCardBorder`.

A few other changes were made that require explanations and comments have been added to the files in this PR for those.

## 📸 Screenshots

Desktop dialog light mode:
desktop light

Desktop dialog dark mode:
desktop dark

Browser dialog light mode:
browser light

Browser dialog dark mode:
browser dark

Web vault reports page (did not use premium badge before, does now):
web-reports-premium

Video of the dialog being used in desktop:

https://github.com/user-attachments/assets/466270ed-862f-4565-9564-69c33388a0cf

Video of the dialog being used in browser:

https://github.com/user-attachments/assets/c2d10556-ea3e-43e1-91ee-4b7b577cfe46

Video of the premium badge interaction in cloud-hosted web vault:

https://github.com/user-attachments/assets/9cc69db8-7965-4a94-95c5-d18af8c7e818

Video of the premium badge going away after upgrading in the cloud-hosted web vault:

https://github.com/user-attachments/assets/88b4a006-6f8c-4f78-beae-be3ccb1ebd26

Video of the premium badge redirecting to the subscription page in self-host web vault:

https://github.com/user-attachments/assets/cd9228f0-6758-40b9-95d5-e2432623d6fd

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/libs/angular/src/billing/components/premium-upgrade-dialog/premium-upgrade-dialog.component.ts
+++ b/libs/angular/src/billing/components/premium-upgrade-dialog/premium-upgrade-dialog.component.ts
@@ -1,7 +1,7 @@
 import { CdkTrapFocus } from "@angular/cdk/a11y";
 import { CommonModule } from "@angular/common";
 import { ChangeDetectionStrategy, Component } from "@angular/core";
-import { catchError, firstValueFrom, map, Observable, of } from "rxjs";
+import { catchError, EMPTY, firstValueFrom, map, Observable } from "rxjs";
 
 import { JslibModule } from "@bitwarden/angular/jslib.module";
 import { SubscriptionPricingServiceAbstraction } from "@bitwarden/common/billing/abstractions/subscription-pricing.service.abstraction";
@@ -66,7 +66,7 @@ export class PremiumUpgradeDialogComponent {
         });
         this.logService.error("Error fetching and mapping pricing tiers", error);
         this.dialogRef.close();
-        return of(null);
+        return EMPTY;
       }),
     );
 
--- a/libs/angular/src/billing/components/premium-upgrade-dialog/premium-upgrade-dialog.component.spec.ts
+++ b/libs/angular/src/billing/components/premium-upgrade-dialog/premium-upgrade-dialog.component.spec.ts
@@ -208,7 +208,7 @@ describe("PremiumUpgradeDialogComponent", () => {
   });
 
   describe("error handling", () => {
-    it("should show error toast and return null when getPersonalSubscriptionPricingTiers$ throws an error", async () => {
+    it("should show error toast and return EMPTY and close dialog when getPersonalSubscriptionPricingTiers$ throws an error", (done) => {
       const error = new Error("Service error");
       mockSubscriptionPricingService.getPersonalSubscriptionPricingTiers$.mockReturnValue(
         throwError(() => error),
@@ -218,14 +218,23 @@ describe("PremiumUpgradeDialogComponent", () => {
       const errorComponent = errorFixture.componentInstance;
       errorFixture.detectChanges();
 
-      const cardDetails = await firstValueFrom(errorComponent["cardDetails$"]);
+      const cardDetails$ = errorComponent["cardDetails$"];
 
-      expect(mockToastService.showToast).toHaveBeenCalledWith({
-        variant: "error",
-        title: "error",
-        message: "unexpectedError",
+      cardDetails$.subscribe({
+        next: () => {
+          done.fail("Observable should not emit any values");
+        },
+        complete: () => {
+          expect(mockToastService.showToast).toHaveBeenCalledWith({
+            variant: "error",
+            title: "error",
+            message: "unexpectedError",
+          });
+          expect(mockDialogRef.close).toHaveBeenCalled();
+          done();
+        },
+        error: (err: unknown) => done.fail(`Observable should not error: ${err}`),
       });
-      expect(cardDetails).toBeNull();
     });
   });
 });


libs/angular/src/billing/directives/not-premium.directive.ts
```diff
@@ -27,14 +29,15 @@ export class NotPremiumDirective implements OnInit {
...
+    this.billingAccountProfileStateService
+      .hasPremiumFromAnySource$(account.id)
+      .pipe(takeUntilDestroyed(this.destroyRef))
+      .subscribe((premium) => {
+        if (premium) {
+          this.viewContainer.clear();
+        } else {
+          this.viewContainer.createEmbeddedView(this.templateRef);
+        }
+      });
```
kdenney, line 41: During testing I found that the premium badge was not updating after doing a full sync. This was because of this code that was using `firstValueFrom` so any subsequent updates to the billing profile state observable was being ignored. This change makes the directive properly reactive to the observable so the update triggers correctly.

apps/web/src/app/admin-console/organizations/policies/policy-edit-definitions/auto-confirm-policy.component.ts
```diff
@@ -32,8 +32,8 @@ export class AutoConfirmPolicy extends BasePolicyEditDefinition {
 })
 export class AutoConfirmPolicyEditComponent extends BasePolicyEditComponent implements OnInit {
   protected readonly autoConfirmSvg = AutoConfirmSvg;
-  private policyForm: Signal<TemplateRef<any> | undefined> = viewChild("step0");
-  private extensionButton: Signal<TemplateRef<any> | undefined> = viewChild("step1");
+  private readonly policyForm: Signal<TemplateRef<any> | undefined> = viewChild("step0");
+  private readonly extensionButton: Signal<TemplateRef<any> | undefined> = viewChild("step1");
```
kdenney: I think this came from the new linter rules on signals. I didn't make the change directly so I'm pretty sure the linter did it. (same with the next file in the list)

apps/web/src/app/auth/settings/emergency-access/emergency-access.component.ts
```diff
@@ -96,15 +96,6 @@ export class EmergencyAccessComponent implements OnInit {
...
 
-  async premiumRequired() {
-    const canAccessPremium = await firstValueFrom(this.canAccessPremium$);
-
-    if (!canAccessPremium) {
-      this.messagingService.send("premiumRequired");
-      return;
-    }
-  }
-
```
kdenney, line 107: This (and the similar change to two-factor-setup.component) was unused so I removed it.

libs/angular/src/billing/components/premium-badge/premium-badge.component.ts
```diff
@@ -19,7 +25,9 @@ export class PremiumBadgeComponent {
 
   constructor(private premiumUpgradePromptService: PremiumUpgradePromptService) {}
 
-  async promptForPremium() {
+  async promptForPremium(event: Event) {
+    event.stopPropagation();
+    event.preventDefault();
```
kdenney, line 32: This fixed an issue where the premium badge click was registered but the container was an <a> tag to a route with the has-premium guard on it, so it was prompting twice.

libs/angular/src/billing/components/premium-upgrade-dialog/premium-upgrade-dialog.component.ts
```diff
@@ -0,0 +1,124 @@
...
+    .pipe(
+      map((tiers) => tiers.find((tier) => tier.id === PersonalSubscriptionPricingTierIds.Premium)),
+      map((tier) => this.mapPremiumTierToCardDetails(tier!)),
+      catchError(() => {
+        this.toastService.showToast({
+          variant: "error",
+          title: this.i18nService.t("error"),
+          message: this.i18nService.t("unexpectedError"),
+        });
+        return of(null);
```
coroiu: suggestion: this doesn't really give anyone a chance to identify the issue, maybe log or show some error specific message?

libs/angular/src/billing/components/premium-upgrade-dialog/premium-upgrade-dialog.component.ts
```diff
@@ -0,0 +1,124 @@
...
+      }),
+    );
+
+  protected loading$: Observable<boolean> = this.cardDetails$.pipe(
+    map(() => false),
+    startWith(true),
+  );
```
coroiu: This will create a separate async stream where `loading$ = false` doesn't necessarily mean that anyone subscribed to `cardDetails$` directly will see the same. You'll will also always see `true` before `false` regardless of whether `cardDetails$` already has a value for you.

suggestion: solve this in the template instead https://stackoverflow.com/questions/39858526/display-loading-while-using-observable-with-async-pipe-in-template

coroiu (approved):
Nice review bingo :) Platform changes look good but I saw some potential improvements in the dialog that I also commented on

apps/web/src/app/vault/services/web-premium-upgrade-prompt.service.ts
```diff
@@ -15,14 +26,43 @@ export class WebVaultPremiumUpgradePromptService implements PremiumUpgradePrompt
...
+    if (hasPremium) {
+      // Already has premium, don't prompt
+      return;
+    }
+
+    const showNewDialog = await this.configService.getFeatureFlag(
+      FeatureFlag.PM23713_PremiumBadgeOpensNewPremiumUpgradeDialog,
+    );
+
+    // Per conversation in PM-23713, retain the existing upgrade org flow for now, will be addressed later
```
nick-livefront: ❓ Is there a ticket associated with revisiting this work? Can it be documented as a part of this comment?

apps/web/src/app/vault/services/web-premium-upgrade-prompt.service.ts
```diff
@@ -57,4 +97,32 @@ export class WebVaultPremiumUpgradePromptService implements PremiumUpgradePrompt
...
+        hideContinueWithoutUpgradingButton: true,
+      },
+    });
+    const result = await lastValueFrom(dialogRef.closed);
+    if (
+      result?.status === UnifiedUpgradeDialogStatus.UpgradedToPremium ||
+      result?.status === UnifiedUpgradeDialogStatus.UpgradedToFamilies
+    ) {
+      await this.apiService.refreshIdentityToken();
+      await this.syncService.fullSync(true);
```
nick-livefront: ❌  `fullSync` calls [refreshIdentityToken](https://github.com/bitwarden/clients/blob/1da4fd22618fe87871fe1b4051b8680a2f5c49ff/libs/common/src/platform/sync/default-sync.service.ts#L162), the call here should not be needed. If I'm missing something let me know.

libs/angular/src/billing/components/premium-upgrade-dialog/premium-upgrade-dialog.component.ts
kdenney: Good call! I added logging in https://github.com/bitwarden/clients/pull/16911/commits/447ad4920720111f26db9e47aa2c5cd79d7fbe57

libs/angular/src/billing/components/premium-upgrade-dialog/premium-upgrade-dialog.component.ts
kdenney: Thanks for the suggestion. I removed the loading observable and added a loading template in the html. The dialog loads so fast I don't think it'll ever display, but it was still a good idea to add. :)

apps/web/src/app/vault/services/web-premium-upgrade-prompt.service.ts
kdenney: Yes, good call. I added the ticket in https://github.com/bitwarden/clients/pull/16911/commits/447ad4920720111f26db9e47aa2c5cd79d7fbe57

apps/web/src/app/vault/services/web-premium-upgrade-prompt.service.ts
kdenney: Good callout! My team was using this pattern in a few places so this was really good to know. I've removed the call to refreshIdentityToken here as well as the other places we were using it across billing code in https://github.com/bitwarden/clients/pull/16911/commits/447ad4920720111f26db9e47aa2c5cd79d7fbe57

coroiu (approved):
👍

Banrion (approved):
DIRT files look good!

apps/web/src/app/auth/settings/emergency-access/emergency-access.component.ts
ike-kottlowski: ❓ : Could you include in the QA notes that we need to validate this flow is unaffected. It doesn't seem like it but I would like if QA was able to double check.

ike-kottlowski (approved):

kdenney:
> Changes for Tools look good, but I just noticed, that the unused message-keys aren't removed. `premiumRequired` and `premiumRequiredDesc` are just examples I found. Could you please go through and remove them from the `en/messages.json`-files

Those two keys are still used when the feature flag is off. I don't think there are any other keys that would be unused after these changes, either.

Banrion (approved):

JimmyVo16 (approved):

djsmith85:
> > Changes for Tools look good, but I just noticed, that the unused message-keys aren't removed. `premiumRequired` and `premiumRequiredDesc` are just examples I found. Could you please go through and remove them from the `en/messages.json`-files
> 
> Those two keys are still used when the feature flag is off. I don't think there are any other keys that would be unused after these changes, either.

`premiumRequired` is no longer used in browser and desktop
`premiumRequiredDesc` is no longer used in browser and desktop

kdenney:
> > > Changes for Tools look good, but I just noticed, that the unused message-keys aren't removed. `premiumRequired` and `premiumRequiredDesc` are just examples I found. Could you please go through and remove them from the `en/messages.json`-files
> > 
> > 
> > Those two keys are still used when the feature flag is off. I don't think there are any other keys that would be unused after these changes, either.
> 
> `premiumRequired` is no longer used in browser and desktop `premiumRequiredDesc` is no longer used in browser and desktop

@djsmith85 
Ahh, yes I was mistakenly looking at web only. While looking into this, I actually discovered that I incorrectly removed the handling of the `premiumRequired` message in desktop's app.component. When the feature flag is off, it still sends that message and without this code, the legacy premium upgrade dialog would not have appeared!

I added the message handling code back in for desktop and removed the unused message keys from browser in https://github.com/bitwarden/clients/pull/16911/commits/3373cec609b2c1733423b2cbaab18de6f3712e1e. 

Great catch and thanks for pointing this out and saving us from a bug!

djsmith85:
@kdenney PS: Lint issue will be fixed once https://github.com/bitwarden/clients/pull/17113 merges. I had already reached out to @bitwarden/team-key-management-dev

libs/angular/src/billing/components/premium-upgrade-dialog/premium-upgrade-dialog.component.ts
```diff
@@ -0,0 +1,122 @@
...
+  templateUrl: "./premium-upgrade-dialog.component.html",
+})
+export class PremiumUpgradeDialogComponent {
+  protected cardDetails$: Observable<CardDetails | null> = this.subscriptionPricingService
+    .getPersonalSubscriptionPricingTiers$()
+    .pipe(
+      map((tiers) => tiers.find((tier) => tier.id === PersonalSubscriptionPricingTierIds.Premium)),
+      map((tier) => this.mapPremiumTierToCardDetails(tier!)),
+      catchError((error: unknown) => {
+        this.toastService.showToast({
```
cyprain-okeke, line 62: When an error occurs, the dialog shows a toast but remains open with no content. The user has to manually close a seemingly broken dialog. Is this the expected experience ?

libs/angular/src/billing/components/premium-upgrade-dialog/premium-upgrade-dialog.component.ts
```diff
@@ -0,0 +1,122 @@
...
+    private subscriptionPricingService: SubscriptionPricingServiceAbstraction,
+    private i18nService: I18nService,
+    private toastService: ToastService,
+    private environmentService: EnvironmentService,
+    private platformUtilsService: PlatformUtilsService,
+    private logService: LogService,
+  ) {}
+
+  protected async upgrade(): Promise<void> {
+    const environment = await firstValueFrom(this.environmentService.environment$);
```
cyprain-okeke: While firstValueFrom automatically unsubscribes, if the dialog is closed before these Observables emit, this could hang.

  **Recommendation:**  Add timeout or use takeUntil with component destruction

cyprain-okeke (changes requested):

libs/angular/src/billing/components/premium-upgrade-dialog/premium-upgrade-dialog.component.ts
kdenney: Good catch! I added a line to close the dialog when this error occurs in https://github.com/bitwarden/clients/pull/16911/commits/2abc88c37dc6310d01aadb9ee3a85e065571c94f

coroiu (approved):

cyprain-okeke (approved):

ike-kottlowski (approved):

PR #16928, opened 2025-10-18
PM-22836 Import popout crashes unexpectedly 

harr1424:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-22836?atlOrigin=eyJpIjoiNTUwZTI2NmZiMzRhNDIzNTlmMTAzMGJkZWZjZDhjOGMiLCJwIjoiaiJ9

## 📔 Objective

• Firefox has a [known issue](https://bugzilla.mozilla.org/show_bug.cgi?id=1292701) that causes the Bitwarden extension to crash when a file picker (`NSOpenPanel` on Mac OS) is opened outside of a popp-out context. 
• The Bitwarden extension exposes a file picker in at least two situations: when creating a file type send and when importing credentials into a vault. 
• To avoid the crash it is necessary to enforce the invariant that a view exposing a file picker can only ever appear in a popped-out context. 
• The UI logic specific to file type sends already prompted the user to pop-out the extension, but it didn't prevent the user from closing the pop-out and re-opening the extension in a not-popped-out context and then attempting to open a file picker.
• It was determined [in an internal Slack discussion](https://bitwarden.slack.com/archives/C04PZA13FQD/p1763635263866649) that the pop-out will be automatic (for consistency) and not require user confirmation. 
• This PR implements the invariant that the views exposing a file picker for both file type sends and vault imports are only ever accessible in a pop-out, and per the conversation linked above, the pop-out occurs without user intervention/confirmation. 

## 📸 Screenshots

Screen Recording 2025-12-10 at 15 50 57 mov

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/apps/browser/src/tools/popup/guards/firefox-popout.guard.spec.ts
+++ b/apps/browser/src/tools/popup/guards/firefox-popout.guard.spec.ts
@@ -158,20 +158,6 @@ describe("firefoxPopoutGuard", () => {
       expect(closePopupSpy).toHaveBeenCalledWith(window);
       expect(result).toBe(false);
     });
-
-    it("should preserve query parameters on file picker routes", async () => {
-      const editSendWithParams: RouterStateSnapshot = {
-        url: "/edit-send?sendId=123&mode=edit",
-      } as RouterStateSnapshot;
-
-      const guard = firefoxPopoutGuard();
-      await TestBed.runInInjectionContext(() => guard(mockRoute, editSendWithParams));
-
-      expect(openPopoutSpy).toHaveBeenCalledWith(
-        "popup/index.html#/edit-send?sendId=123&mode=edit",
-      );
-      expect(closePopupSpy).toHaveBeenCalledWith(window);
-    });
   });
 
   describe("url handling", () => {


apps/browser/src/tools/popup/settings/import/import-browser-v2.component.html
```diff
@@ -12,6 +12,8 @@
   >
   </tools-import>
 
+  <import-file-popout-dialog-container></import-file-popout-dialog-container>
```
Hinton: **question:** Why is this a separate component rather than just living inside the import-browser component? It seems quite weird to have a child component popup a dialog that redirects the parent page.

apps/browser/src/popup/app-routing.module.ts
```diff
@@ -317,7 +317,7 @@ const routes: Routes = [
     path: "add-send",
     component: SendAddEditV2Component,
     canActivate: [authGuard],
-    data: { elevation: 1 } satisfies RouteDataProperties,
+    data: { elevation: 1, doNotSaveUrl: true } satisfies RouteDataProperties,
```
Hinton: **question:** Why are we not saving the url here?

harr1424: Copying from Claude's comment (below) which correctly notes this logic wasn't applied to the import path: 

Without this flag, the import page URL is saved in popup history
If a user navigates away and the popup restores this URL, they could still access the file selector outside the popout
This defeats the purpose of the Firefox crash fix

apps/browser/src/tools/popup/send-v2/send-file-popout-dialog/send-file-popout-dialog.component.ts
```diff
@@ -14,13 +15,18 @@ import BrowserPopupUtils from "../../../../platform/browser/browser-popup-utils"
   imports: [JslibModule, CommonModule, DialogModule, ButtonModule, TypographyModule],
 })
 export class SendFilePopoutDialogComponent {
-  constructor(private dialogService: DialogService) {}
+  constructor(
+    private dialogService: DialogService,
+    private router: Router,
+  ) {}
```
bmbitwarden, line 18: use inject instead of constructor

apps/browser/src/tools/popup/send-v2/send-file-popout-dialog/send-file-popout-dialog.component.ts
```diff
@@ -14,13 +15,18 @@ import BrowserPopupUtils from "../../../../platform/browser/browser-popup-utils"
...
 
   async popOutWindow() {
     await BrowserPopupUtils.openCurrentPagePopout(window);
   }
 
-  close() {
+  async close() {
```
bmbitwarden: return type

apps/browser/src/tools/popup/settings/import/import-browser-v2.component.html
bmbitwarden: dont need closing tag

apps/browser/src/tools/popup/settings/import/import-file-popout-dialog-container.component.ts
```diff
@@ -0,0 +1,34 @@
...
+  template: "",
+  imports: [JslibModule, CommonModule],
+})
+export class ImportFilePopoutDialogContainerComponent implements OnInit, OnDestroy {
+  private dialogRef: DialogRef | null = null;
+
+  constructor(
+    private dialogService: DialogService,
+    private filePopoutUtilsService: FilePopoutUtilsService,
+  ) {}
```
bmbitwarden: inject

apps/browser/src/tools/popup/settings/import/import-file-popout-dialog.component.ts
```diff
@@ -0,0 +1,31 @@
...
+
+import BrowserPopupUtils from "../../../../platform/browser/browser-popup-utils";
+import { PopupRouterCacheService } from "../../../../platform/popup/view-cache/popup-router-cache.service";
+
+@Component({
+  changeDetection: ChangeDetectionStrategy.OnPush,
+  selector: "import-file-popout-dialog",
+  templateUrl: "./import-file-popout-dialog.component.html",
+  imports: [JslibModule, CommonModule, DialogModule, ButtonModule, TypographyModule],
+})
```
bmbitwarden: standalone component

apps/browser/src/tools/popup/settings/import/import-file-popout-dialog-container.component.ts
```diff
@@ -0,0 +1,34 @@
...
+import { Component, OnInit, ChangeDetectionStrategy, OnDestroy } from "@angular/core";
+
+import { JslibModule } from "@bitwarden/angular/jslib.module";
+import { DialogRef, DialogService } from "@bitwarden/components";
+
+import { FilePopoutUtilsService } from "../../services/file-popout-utils.service";
+
+import { ImportFilePopoutDialogComponent } from "./import-file-popout-dialog.component";
+
+@Component({
```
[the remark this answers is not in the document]
Hinton: @theMickster this seems wrong. We actually omit `standalone` and only want to flag this if the component includes `standalone: false`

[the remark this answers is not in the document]
theMickster: @claude: Why did you make this finding? Did you read a contributing document, a README.md, a CLAUDE.md, or similar that brought you to this conclusion?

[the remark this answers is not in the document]
theMickster: Sad panda 🐼
Seems like the respond workflow that we tried to create isn't working. I'll take a look, but I might need some help going into the weeds as to why it's not working @mandreko-bitwarden

Hinton:
@harr1424 This PR seems to do several things, what parts of this actually resolves the stated issue, i.e. popout crashes.

harr1424:
> @harr1424 This PR seems to do several things, what parts of this actually resolves the stated issue, i.e. popout crashes.

@Hinton Depending upon user behavior the following file changes are required to prevent a pop-out crash on Firefox. [See the root cause of the problem this PR defends against](https://bugzilla.mozilla.org/show_bug.cgi?id=1292701). 

- `apps/browser/src/popup/app-routing.module.ts`
-  `apps/browser/src/tools/popup/send-v2/send-file-popout-dialog/send-file-popout-dialog.component.ts`
-  `apps/browser/src/_locales/en/messages.json`
-  `apps/browser/src/tools/popup/services/file-popout-utils.service.ts`
-  `apps/browser/src/tools/popup/settings/import/import-browser-v2.component.html`
-  `apps/browser/src/tools/popup/settings/import/import-browser-v2.component.ts`
-  `apps/browser/src/tools/popup/settings/import/import-file-popout-dialog-container.component.ts`
-  `apps/browser/src/tools/popup/settings/import/import-file-popout-dialog.component.html`
-  `apps/browser/src/tools/popup/settings/import/import-file-popout-dialog.component.ts` 
-  `apps/browser/src/vault/popup/settings/vault-settings-v2.component.ts`


The following files don't directly address the Firefox issue, but were used to resolve a situation where the extension never loaded. @MGibson1 has since merged changes that resolves this on `main`.  I can remove these files and re-test to make sure everything still works and then re-submit if this is your preference. 

-  `apps/browser/src/platform/storage/foreground-memory-storage.service.ts` (this would add some helpful logging in the event a breaking change similar to last week's occurs) 
-  `apps/browser/src/popup/app.component.ts`  (this one is likely worth keeping)

apps/browser/src/popup/app-routing.module.ts
Hinton, line 1: **question:** Why are these needed when we show a dialog on the pages when opened in a popup.

apps/browser/src/popup/app.component.ts
```diff
@@ -91,6 +91,7 @@ export class AppComponent implements OnInit, OnDestroy {
   protected showSdkWarning = this.sdkService.client$.pipe(
     map(() => false),
     catchError(() => of(true)),
+    startWith(false), // Emit immediately so template renders
```
Hinton: **issue:** Why this change? If the SDK hasn't loaded we should not continue operating due to it being required for many operations.

apps/browser/src/popup/main.ts
Hinton, line 1: **question:** Why these changes? On a catastrophic bootstrap failure what good does logging the issue. rather than throwing an error that is already propagated?

apps/browser/src/tools/popup/send-v2/send-file-popout-dialog/send-file-popout-dialog.component.ts
```diff
@@ -14,13 +15,18 @@ import BrowserPopupUtils from "../../../../platform/browser/browser-popup-utils"
...
     this.dialogService.closeAll();
+    // the current view exposes a file selector, ensure the view is popped to avoid using it outside of a popout
+    await this.router.navigate(["/tabs/send"]);
```
Hinton: **question:** The comment isn't really descriptive what are we doing here. Are we navigating back on closure?

apps/browser/src/tools/popup/services/file-popout-utils.service.ts
```diff
@@ -7,7 +7,7 @@ import BrowserPopupUtils from "../../../platform/browser/browser-popup-utils";
 /**
  * Service for determining whether to display file popout callout messages.
  */
-@Injectable()
+@Injectable({ providedIn: "root" })
```
Hinton: **question:** This just makes it a singleton, does that resolve something?

apps/browser/src/tools/popup/settings/import/import-file-popout-dialog-container.component.ts
```diff
@@ -0,0 +1,34 @@
...
+  changeDetection: ChangeDetectionStrategy.OnPush,
+  selector: "import-file-popout-dialog-container",
+  template: "",
+  imports: [JslibModule, CommonModule],
+})
+export class ImportFilePopoutDialogContainerComponent implements OnInit, OnDestroy {
+  private dialogRef: DialogRef | null = null;
```
Hinton:
```suggestion
  private dialogRef?: DialogRef;
```

apps/browser/src/tools/popup/settings/import/import-file-popout-dialog.component.html
Hinton, line 1: **suggestion:** This seems pretty much identical to the import popout dialog. Should they be combined and you just feed it the title/body to use?

apps/browser/src/tools/popup/settings/import/import-file-popout-dialog.component.html
```diff
@@ -0,0 +1,20 @@
+<bit-simple-dialog dialogSize="default">
+  <div bitDialogIcon>
+    <i class="bwi bwi-info-circle bwi-2x tw-text-info" aria-hidden="true"></i>
+  </div>
```
Hinton: **suggestion:**
Dialogs can use `i` directly, no need to wrap it in a div. Dialogs also generally use 3xl for icons.

```suggestion
  <i bitDialogIcon class="bwi bwi-info-circle tw-text-info tw-text-3xl" aria-hidden="true"></i>
```

apps/browser/src/tools/popup/settings/import/import-file-popout-dialog.component.ts
```diff
@@ -0,0 +1,31 @@
...
+  ) {}
+
+  async popOutWindow() {
+    await BrowserPopupUtils.openCurrentPagePopout(window);
+  }
+
+  async close() {
+    this.dialogService.closeAll();
+    // the current view exposes a file selector, ensure the view is popped to avoid using it outside of a popout
+    await this.popupRouterCacheService.back();
```
Hinton: **question:** This behaves differently from send, any reason why?

Hinton (commented):
I have not reviewed the platform owned files since I don't have enough context to understand the impact of the changes.

apps/browser/src/tools/popup/settings/import/import-file-popout-dialog-container.component.ts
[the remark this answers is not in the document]
theMickster: Hey "@claude"  Why did you make this finding? Did you read a contributing document, a README.md, a CLAUDE.md, or similar that brought you to this conclusion?

apps/browser/src/popup/app-routing.module.ts
harr1424: ## Import

With `doNotSaveUrl: true `: Navigating to the import view, and then closing the pop-out, and then re-opening the main extension view shows the penultimate view to the import menu, the "Vault Options" view

Without `doNotSaveUrl: true `: Following the same workflow as above, instead of the "Vault Options" view, the pop-out dialog is instead displayed. Clicking cancel returns to the "Vault Options" view, and clicking "Pop Out" will trigger the pop-out. 

## Send

With `doNotSaveUrl: true `:  Navigating towards send file view, confirming the pop-out, and then closing the pop-out, re-opening the extension (default not popped out state) shows the main "Send" view

Without `doNotSaveUrl: true `:  The same workflow as above results in the "Send File" view being presented outside of the pop-out context. This allows the situation the PR is intended to avoid: a user can attempt to open a file selection dialog outside of a popped out context and the extensiuon will crash

## Since only the Send functionality requires this, why was it added to Import as well? 

For consistency, it can be safely removed from the import routing logic if this is preferred.

apps/browser/src/popup/app.component.ts
harr1424: This will be removed before my next commit.

apps/browser/src/popup/main.ts
harr1424: This will be removed before my next commit.

apps/browser/src/tools/popup/send-v2/send-file-popout-dialog/send-file-popout-dialog.component.ts
harr1424: I've added a clarifying comment, but it is most relevant to English speakers. I am not sure if there is a better way to describe what is happening here that spans different languages: 
`  // If the user selects "cancel" when presented the dialog, navigate back to the main Send tab`

apps/browser/src/tools/popup/services/file-popout-utils.service.ts
harr1424: This seemed like a service that _should_ be a singleton, but removing this code change hasn't impacted any functionality. I think a user would have to do some gymnastics here to create a race condition. It is possible to trigger multiple pop-out windows but at least on Mac OS only one of them can open a file picker at a time.

apps/browser/src/tools/popup/settings/import/import-file-popout-dialog.component.html
harr1424: Sorry, but I am not sure I completely understand: 
> This seems pretty much identical to the import popout dialog

Which file are you referring to here?

apps/browser/src/tools/popup/settings/import/import-file-popout-dialog.component.ts
harr1424: This has been changed for consistency.

harr1424: > This has been changed for consistency.

This has been reverted back to use `popupRouterCacheService.back()`. Using router to instead navigate to a new views piles on views that can be "popped" off the stack. This creates a UI bug where clicking back repeatedly circles back to the same dialog presentation.

harr1424: The difference in behavior is because the "Send" tab/view is on the root of the extension, this view doesn't present a back button in the navigation tab. Whereas the "Import Data" view is nested inside of the "Vault Settings" view which does present a back button in the navigation tab.

apps/browser/src/tools/popup/settings/import/import-file-popout-dialog-container.component.ts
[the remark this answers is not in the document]
theMickster: @claude: Why did you make this finding? Did you read a contributing document, a README.md, a CLAUDE.md, or similar that brought you to this conclusion?

apps/browser/src/tools/popup/settings/import/import-file-popout-dialog.component.html
Hinton: Apologies, `SendFilePopoutDialogComponent`

apps/browser/src/platform/storage/foreground-memory-storage.service.ts
```diff
@@ -36,6 +37,25 @@ export class ForegroundMemoryStorageService extends AbstractStorageService {
       filter((message) => message.originator === "background"),
     );
 
+    // Wait for initialization message to ensure port is ready
+    // Add a timeout in case initialization never arrives (e.g., if BrowserApi.senderIsInternal rejects the connection)
+    const initMessage$ = this._backgroundResponses$.pipe(
+      filter((message) => message.action === "initialization"),
+    );
+
+    const timeout$ = timer(2000).pipe(map((): MemoryStoragePortMessage | null => null));
```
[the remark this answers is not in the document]
coroiu: @harr1424 I'm also curious ☝️

apps/browser/src/tools/popup/send-v2/send-file-popout-dialog/send-file-popout-dialog.component.ts
Hinton: That's much clearer.

apps/browser/src/platform/storage/foreground-memory-storage.service.ts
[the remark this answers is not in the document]
harr1424: @coroiu My apologies, this is left over from when I was troubleshooting the breaking change introduced by a security path last week. I've now ensured none of those changes remain in this PR.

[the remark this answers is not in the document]
coroiu: @harr1424 alright, thanks! I went ahead and removed the review requests from platform since we no longer own any code in this PR 👍

apps/browser/src/popup/app-routing.module.ts
Hinton: I think I understand why you've resorted to adding `doNotTrack`. Since `SendFilePopoutDialogContainerComponent` relies on `ngOnInit` there is a race condition that primarily occurs when navigating directly to the add-send url without first landing on sends. Where `config` does not have the necessary values to trigger the open condition.

Is this accurate? And if so should we resolve the root cause rather than rely on the race condition?

```ts
@Component({
  selector: "send-file-popout-dialog-container",
  templateUrl: "./send-file-popout-dialog-container.component.html",
  imports: [JslibModule, CommonModule],
})
export class SendFilePopoutDialogContainerComponent {
  private readonly dialogService = inject(DialogService);
  private readonly filePopoutUtilsService = inject(FilePopoutUtilsService);

  readonly config = input.required<SendFormConfig>();

  /**
   * Tracks if the dialog has already been opened. This prevents multiple dialogs from opening if config is updated.
   */
  private readonly dialogOpened = signal(false);

  constructor() {
    effect(() => {
      if (this.dialogOpened()) {
        return;
      }

      if (
        this.config().sendType === SendType.File &&
        this.config().mode === "add" &&
        this.filePopoutUtilsService.showFilePopoutMessage(window)
      ) {
        this.dialogService.open(SendFilePopoutDialogComponent, {
          positionStrategy: new CenterPositionStrategy(),
        });
        this.dialogOpened.set(true);
      }
    });
  }
}
```

apps/browser/src/tools/popup/settings/import/import-file-popout-dialog.component.html
harr1424: @Hinton this definitley seems possible at first, and there is actually a [`FilePopoutComponent`](https://github.com/bitwarden/clients/blob/9e6d0cce35196a53c9305820333dc701207d9280/apps/browser/src/tools/popup/components/file-popout-callout.component.ts#L17)  that curiously wasn't being used on the existing Send File workflow. 

When I started work on this PR I began to suspect it is because `FilePopoutComponent` doesn't have any logic controlling cancellation behavior. Maybe this is preferred? 

I've used the cancellation logic in `ImportFilePopoutDialogComponent` and `SendFilePopoutDialogComponent` to ensure the user is navigated away from a view that exposes a file selector, but since these components are at different view hierarchies, the navigation logic needs to be handled differently [(this touches on another review comment)](https://github.com/bitwarden/clients/pull/16928#discussion_r2542449468): 

```typescript
export class ImportFilePopoutDialogComponent {
...

    // If the user selects "cancel" when presented the dialog, navigate back to the main Send tab
  async close() {
    this.dialogService.closeAll();
    await this.popupRouterCacheService.back();
  }
}
```
```typescript
...
  // If the user selects "cancel" when presented the dialog, navigate back to the main Send tab
  async close() {
    this.dialogService.closeAll();
    await this.router.navigate(["/tabs/send"]);
  }
}
```

apps/browser/src/popup/app-routing.module.ts
harr1424: I am going to mark this conversation as resolved, since it doesn't apply to the guard being used

apps/browser/src/tools/popup/guards/firefox-popout.guard.spec.ts
```diff
@@ -0,0 +1,208 @@
...
+
+      const guard = firefoxPopoutGuard();
+      const result = await TestBed.runInInjectionContext(() => guard(mockRoute, editSendState));
+
+      expect(openPopoutSpy).toHaveBeenCalledWith("popup/index.html#/edit-send");
+      expect(closePopupSpy).toHaveBeenCalledWith(window);
+      expect(result).toBe(false);
+    });
+
+    it("should preserve query parameters on file picker routes", async () => {
```
mcamirault: ⛏️ This is a duplicate of the query parameter test on 184, right? If we only keep one I'd prefer the other one since it's a URL behavior being tested

harr1424: Good catch!

PR #16983, opened 2025-10-22
[PM-26944] phishing data checksum diffing + daily patches

willmartian:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-26944

## 📔 Objective

- Split out data retrieval logic from `PhishingDetectionService` to a new `PhishingDataService` due to increased complexity.
  - Moved fetching from `AuditService` to `PhishingDataService`
- Updated to use `StateProvider` instead of `StorageService`, which is not meant to be used directly
- Changed source of data to be the Phishing.Database Github CDN
- Added logic to only fetch if data is new (via checksum check)
- Added logic to patch the existing data if it is less than 1 day old. Otherwise, will fetch the entire dataset 
- Added tests 👯 

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/apps/browser/src/dirt/phishing-detection/services/phishing-detection.service.ts
+++ b/apps/browser/src/dirt/phishing-detection/services/phishing-detection.service.ts
@@ -388,6 +388,7 @@ export class PhishingDetectionService {
   private static _cleanup() {
     this._destroy$.next();
     this._destroy$.complete();
+    this._destroy$ = new Subject<void>();
 
     this._caughtTabs.clear();
 


Hinton:
How often are we planning on updating this? This is a significant amount of data to check in to our repository. At a minimum this must use git lfs and ideally just require a initial download outside this repository.

willmartian:
> How often are we planning on updating this? This is a significant amount of data to check in to our repository. At a minimum this must use git lfs and ideally just require a initial download outside this repository.

@Hinton It should be updated before each release. I will look into enabling git lfs for this (thanks for the rec). Do you expect enabling git lfs to require any downstream updates to workflows that build the extension?

Hinton:
I think lfs is enabled everywhere these days.

How are we updating the files on the clients? Publishing daily patch files? I don't think there's a reason for people to re-download the whole file every day when we could just serve appends.

willmartian:
> How are we updating the files on the clients? Publishing daily patch files? I don't think there's a reason for people to re-download the whole file every day when we could just serve appends.

@Hinton Current implementation is to re-download the whole file every day (if the checksum is different). I'm not sure how we could do incremental patches if we are just [hitting the GitHub CDN for the data](https://raw.githubusercontent.com/Phishing-Database/Phishing.Database/master/phishing-domains-ACTIVE.txt) instead of our server.

Also, the source dataset is not append-only; inactive sites are removed.

withinfocus:
There's a file we can use of what's changed every day.

Can we deliver this highly compressed and unpack it upon first access?

willmartian:
> There's a file we can use of what's changed every day.

@withinfocus We could use that file, but clients that are behind >1 day will still have to download the full file. I guess the question is how likely is that to occur, and whether that changes the priority of further optimization. Do we expect that most browser clients phone in daily? Seems fair enough to assume? 

Re: compression, it is already gzipped when served from the cdn. Total file size is ~13mb but only ~4mb is transferred. 

We _could_ maintain a repo that pulls the delta file every day that is reset on a client release. We could then query multiple deltas and chain them together. (Just spitballing, that is definitely a scope/complexity increase.)

willmartian:
@Hinton It looks like git lfs and husky conflict with each other: https://github.com/typicode/husky/issues/1431 😞 

Some suggest moving to other tools like [pre-commit](https://pre-commit.com/) or [lefthooks](https://github.com/evilmartians/lefthook).

Hinton:
@willmartian we only use the pre-commit hook of husky which doesn't seem to conflict.

willmartian:
> @willmartian we only use the pre-commit hook of husky which doesn't seem to conflict.

@Hinton I get errors when I run `git lfs install`. Even though we only use the pre-commit hook, I think Husky still reaches into all of the git hooks during install.

willmartian:
Took a different approach after further conversation with Arch. See updated description. TLDR: removed local fallback db

willmartian:
@claude Hey buddy! I know you are quite busy. I just want to ping you on this again in case it slipped your mind. Keep up the great work!

willmartian:
Not sure why [31756d2](https://github.com/bitwarden/clients/pull/16983/commits/31756d26d59953efae8e48fd593fcaef12199648) wasn't causing in an issue in my local environment. 🤔 Fixed.

coroiu (approved):
Approving state addition

coroiu (approved):

Banrion (approved):

PR #16985, opened 2025-10-22
[PM-27209] Force `null` value for `organizationId`

nick-livefront:
## 🎟️ Tracking

[PM-27209](https://bitwarden.atlassian.net/browse/PM-27209)

## 📔 Objective

The issue was missed by https://github.com/bitwarden/clients/pull/16971 where an infinite loop was introduced. The infinite loop was caused by the `organizationId` being set to `undefined` and Angular forcing the value to be `null`, thus triggering our `valueChanges` listener which would "enable" the field.... and repeat.

Changes here to remediate:
- Force `organizationId` to be `null` when prefilling from an existing cipher. 
- Add `distinctUntilChanged` to avoid any odd loops in the future. 

## 📸 Screenshots

<video src="https://github.com/user-attachments/assets/f25e0e8f-33b7-4055-ab11-3139c2312114" />

## 🦮 Reviewer guidelines

[PM-27209]: https://bitwarden.atlassian.net/browse/PM-27209?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/vault/src/cipher-form/components/item-details/item-details-section.component.spec.ts
+++ b/libs/vault/src/cipher-form/components/item-details/item-details-section.component.spec.ts
@@ -640,6 +640,46 @@ describe("ItemDetailsSectionComponent", () => {
     });
   });
 
+  describe("initFromExistingCipher", () => {
+    it("should set organizationId to null when prefillCipher.organizationId is undefined", async () => {
+      component.config.organizationDataOwnershipDisabled = true;
+      component.config.organizations = [{ id: "org1" } as Organization];
+
+      const prefillCipher = {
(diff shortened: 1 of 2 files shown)


libs/vault/src/cipher-form/components/item-details/item-details-section.component.spec.ts
shane-melton, line 1: As always, thanks for the test 🧪 🎉

shane-melton (approved):

PR #16863, opened 2025-10-14
[PM-26927] Bugfix for Key Connector URL

rr-bw:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-26927

## 📔 Objective

The Key Connector URL was getting overwritten back to the default URL on `submit()` because `valueChanges` gets triggered during `submit()`. This PR adds a check to make sure we only set the default URL when changing TO Key Connector from a different decryption option. In other words, don't overwrite back to the default URL during `submit()`.

Also removes the trailing slash `/` from the default URL.

## ⏰ Reminders before review

- Contributor guidelines followed
- All formatters and local linters executed and passed
- Written new unit and / or integration tests where applicable
- Protected functional changes with optionality (feature flags)
- Used internationalization (i18n) for all UI strings
- CI builds passed
- Communicated to DevOps any deployment requirements
- Updated any necessary documentation (Confluence, contributing docs) or informed the documentation team

## 🦮 Reviewer guidelines

- 👍 (`:+1:`) or similar for great changes
- 📝 (`:memo:`) or ℹ️ (`:information_source:`) for notes or general info
- ❓ (`:question:`) for questions
- 🤔 (`:thinking:`) or 💭 (`:thought_balloon:`) for more open inquiry that's not quite a confirmed issue and could potentially benefit from discussion
- 🎨 (`:art:`) for suggestions / improvements
- ❌ (`:x:`) or ⚠️ (`:warning:`) for more significant problems or concerns needing attention
- 🌱 (`:seedling:`) or ♻️ (`:recycle:`) for future improvements or indications of technical debt
- ⛏ (`:pick:`) for minor or nitpick changes

PR #16995, opened 2025-10-23
[PM-27244] Refactor KC URL Handling

rr-bw:
## 🎟️ Tracking

[PM-27244](https://bitwarden.atlassian.net/browse/PM-27244)

> [!IMPORTANT]
> Please [hide whitespace](https://github.com/bitwarden/clients/pull/16995/files?diff=split&w=1) when reviewing this PR. Some changes involve wrapping code in a `try/catch` which makes it seem like there are more changes than there actually are. This can cause the meaningful changes to get lost in the "noise".

### Problem

`this.route.params` was emitting after the `memberDecryptionType.valueChanges` subscription gets setup. The `this.route.params` subscription triggers a `load()`, which triggers a `valueChanges` on all form fields via `patchValue()`.  What this means is that configuration data would be retrieved from the server and populated into the form fields via `load()`, but `load()` would trigger the `memberDecryptionType.valueChanges`, which would in turn overwrite the configured/establish KC URL back to the default KC URL.

### Solution

In fixing this bug I refactored my [previous approach](https://github.com/bitwarden/clients/pull/16863). Here is the “end state” that this solution is meant to achieve:

**Scenario 1 - Admin lands on config page with the Master Password option already configured**

- An admin lands on the config page with Master Password option already configured and selected
- Admin then selects Key Connector option
- Expected → The KC URL should pre-populate with the default URL
- The admin saves the form
- Expected → The default URL should be sent in the request and returned in the response during `submit()`, and remain filled in the KC URL form field.

**Scenario 2 → Admin lands on config page with Key Connector option already configured**

- An admin lands on the config page with Key Connector already configured and selected
- Expected → The KC URL form field should already be pre-populated with the admin’s configured KC URL (not the default URL)
- The admin clicks a different Member Decryption option, such as Master Password, but does NOT save the form
- The admin then clicks the Key Connector option once again
- Expected → The KC URL form field should remain pre-populated with the admin’s configured KC URL (not the default URL)

**Scenario 3 → Admin navigates to a different organization**

- An admin has two organizations setup
- An admin lands on the config page for Org A
- An admin then changes the browser address URL to navigate to Org B
- Expected → The logic should continue to function just as expected as defined in scenarios 1 and 2.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

[PM-27244]: https://bitwarden.atlassian.net/browse/PM-27244?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/workflows/alert-ddg-files-modified.yml
+++ b/.github/workflows/alert-ddg-files-modified.yml
@@ -14,7 +14,7 @@ jobs:
       pull-requests: write
     steps:
     - name: Checkout code
-      uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+      uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
       with:
         fetch-depth: 0
         persist-credentials: false
--- a/.github/workflows/auto-branch-updater.yml
+++ b/.github/workflows/auto-branch-updater.yml
@@ -30,7 +30,7 @@ jobs:
         run: echo "branch=${GITHUB_REF#refs/heads/}" >> "$GITHUB_OUTPUT"
 
       - name: Checkout repo
-        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+        uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
         with:
           ref: 'eu-web-${{ steps.setup.outputs.branch }}'
           fetch-depth: 0
--- a/.github/workflows/build-browser.yml
+++ b/.github/workflows/build-browser.yml
@@ -55,7 +55,7 @@ jobs:
       has_secrets: ${{ steps.check-secrets.outputs.has_secrets }}
     steps:
       - name: Check out repo
-        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+        uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
         with:
           ref: ${{  github.event.pull_request.head.sha }}
           persist-credentials: false
@@ -94,7 +94,7 @@ jobs:
         working-directory: apps/browser
     steps:
       - name: Check out repo
-        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+        uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
         with:
           ref: ${{  github.event.pull_request.head.sha }}
           persist-credentials: false
@@ -146,7 +146,7 @@ jobs:
       _NODE_VERSION: ${{ needs.setup.outputs.node_version }}
     steps:
       - name: Check out repo
-        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+        uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
         with:
           ref: ${{ github.event.pull_request.head.sha }}
           persist-credentials: false
@@ -248,7 +248,7 @@ jobs:
             artifact_name: "dist-opera-MV3"
     steps:
       - name: Check out repo
-        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+        uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
         with:
           ref: ${{  github.event.pull_request.head.sha }}
           persist-credentials: false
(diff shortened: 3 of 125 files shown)


bitwarden_license/bit-web/src/app/auth/sso/sso.component.ts
```diff
@@ -306,55 +266,128 @@ export class SsoComponent implements OnInit, OnDestroy {
...
-    this.spAcsUrl = ssoSettings.urls.spAcsUrl;
+    try {
+      const userId = await firstValueFrom(getUserId(this.accountService.activeAccount$));
+      this.organization = await firstValueFrom(
+        this.organizationService
+          .organizations$(userId)
+          .pipe(getOrganizationById(this.organizationId)),
+      );
+      const ssoSettings = await this.organizationApiService.getSso(this.organizationId);
+      this.configuredKeyConnectorUrlFromServer = ssoSettings.data.keyConnectorUrl;
```
JaredSnider-Bitwarden: Based on the [organization sso response model](https://github.com/bitwarden/clients/blob/main/libs/common/src/auth/models/response/organization-sso.response.ts#L16-L19), data can be null so we have to add an optional chaining operator here:
`ssoSettings.data?.keyConnectorUrl`

bitwarden_license/bit-web/src/app/auth/sso/sso.component.ts
```diff
@@ -306,55 +266,128 @@ export class SsoComponent implements OnInit, OnDestroy {
...
+      }
+      const request = new OrganizationSsoRequest();
+      request.enabled = this.enabledCtrl.value;
+      // Return null instead of empty string to avoid duplicate id errors in database
+      request.identifier =
+        this.ssoIdentifierCtrl.value === "" ? null : this.ssoIdentifierCtrl.value;
+      request.data = SsoConfigApi.fromView(this.ssoConfigForm.getRawValue());
+
+      const response = await this.organizationApiService.updateSso(this.organizationId, request);
+      this.configuredKeyConnectorUrlFromServer = response.data.keyConnectorUrl;
```
JaredSnider-Bitwarden: Based on the [organization sso response model](https://github.com/bitwarden/clients/blob/main/libs/common/src/auth/models/response/organization-sso.response.ts#L16-L19), data can be null so we have to add an optional chaining operator here:
`ssoSettings.data?.keyConnectorUrl`

bitwarden_license/bit-web/src/app/auth/sso/sso.component.ts
```diff
@@ -306,55 +266,128 @@ export class SsoComponent implements OnInit, OnDestroy {
...
+              // Clear the key connector url
+              this.ssoConfigForm.controls.keyConnectorUrl.setValue("");
+            }
+          }),
+          takeUntil(this.destroy$),
+        )
+        .subscribe();
+  }
+
   async validateKeyConnectorUrl() {
```
JaredSnider-Bitwarden, line 405: To fix the bug that QA observed where when you click the test button and it errors the error isn't shown, we should add a call to  `this.keyConnectorUrl.markAllAsTouched();` after set errors.

bitwarden_license/bit-web/src/app/auth/sso/sso.component.ts
```diff
@@ -306,55 +266,128 @@ export class SsoComponent implements OnInit, OnDestroy {
...
   }
 
   submit = async () => {
-    this.updateFormValidationState(this.ssoConfigForm);
-
-    if (this.ssoConfigForm.value.memberDecryptionType === MemberDecryptionType.KeyConnector) {
-      this.haveTestedKeyConnector = false;
-      await this.validateKeyConnectorUrl();
-    }
+    this.isPopulatingForm = true;
```
JaredSnider-Bitwarden: Nit (naming):  
These operations trigger value changes:
1. Validating the form via the `this.updateFormValidationState(this.ssoConfigForm)` call triggers value changes on all listeners via the `updateValueAndValidity` calls on each form control. 
2. Calling the `populateForm` method because it patches the whole form

Could we try for something like `isFormValidatingOrPopulating`?

bitwarden_license/bit-web/src/app/auth/sso/sso.component.ts
```diff
@@ -308,55 +273,135 @@ export class SsoComponent implements OnInit, OnDestroy {
...
 
+  private subscribeToMemberDecryptionTypeValueChanges() {
+    // The load() method will have unsubscribed from any pre-existing subscription before
+    // we setup a new subscription here.
+
+    this.memberDecryptionTypeValueChangesSubscription =
+      this.ssoConfigForm?.controls?.memberDecryptionType.valueChanges
+        .pipe(
+          switchMap(async (memberDecryptionType: MemberDecryptionType) => {
+            this.haveTestedKeyConnector = false;
```
enmande, line 368: 📓 Added in review to support an observed case where:
1. User tests the key connector URL using the "Test" `bitSuffix bitLink` in the `input` field.
2. User uses the "Member Decryption Options" radio button group to select a _different_ option  (e.g, _Trusted Device Encryption_).
3. User uses the "Member Decryption Options" radio button group to _return_ to _Key Connector_ scheme.

Before adding the reset of `haveTestedKeyConnector` flag here, user could be in a state to show a potentially incorrect successful test message. On each display of the form, the input is populated with the default URL returned from server (which would overwrite a user-altered value before swapping away to a different Member Decryption Option), yet the long-lived flag would not have reflected that change, and the form could have been submitted without being  subject to endpoint validation.

This appears to have been an existing issue, but one that we felt important to address with revisiting this form.

JaredSnider-Bitwarden (approved):

PR #16695, opened 2025-10-01
Move importer metadata to native code

adudek-bw:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-25521

## ⏰ Reminders before review

- Contributor guidelines followed
- All formatters and local linters executed and passed
- Written new unit and / or integration tests where applicable
- Protected functional changes with optionality (feature flags)
- Used internationalization (i18n) for all UI strings
- CI builds passed
- Communicated to DevOps any deployment requirements
- Updated any necessary documentation (Confluence, contributing docs) or informed the documentation team

## 🦮 Reviewer guidelines

- 👍 (`:+1:`) or similar for great changes
- 📝 (`:memo:`) or ℹ️ (`:information_source:`) for notes or general info
- ❓ (`:question:`) for questions
- 🤔 (`:thinking:`) or 💭 (`:thought_balloon:`) for more open inquiry that's not quite a confirmed issue and could potentially benefit from discussion
- 🎨 (`:art:`) for suggestions / improvements
- ❌ (`:x:`) or ⚠️ (`:warning:`) for more significant problems or concerns needing attention
- 🌱 (`:seedling:`) or ♻️ (`:recycle:`) for future improvements or indications of technical debt
- ⛏ (`:pick:`) for minor or nitpick changes

PR #17019, opened 2025-10-24
[PM-27358] Remove unused getInstalledBrowsers method

djsmith85:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-27358
## 📔 Objective

Since the work on https://github.com/bitwarden/clients/pull/16695 merged, all the metadata needed to enable an importer is provided by the native code. This means the desktop app no longer needs to know which browsers are installed to guard a feature.

The method `getInstalledBrowsers` should be removed from the typescript side, and no longer be exposed from the native code.

❓  Open question to the reviewer: Do we need to change the access modifiers on `DefaultInstalledBrowserRetriever` in [chromium.rs](https://github.com/bitwarden/clients/blob/main/apps/desktop/desktop_native/bitwarden_chromium_importer/src/chromium.rs) ?

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/apps/desktop/desktop_native/napi/src/lib.rs
+++ b/apps/desktop/desktop_native/napi/src/lib.rs
@@ -1066,8 +1066,8 @@ pub mod logging {
 pub mod chromium_importer {
     use chromium_importer::{
         chromium::{
-            DefaultInstalledBrowserRetriever,
-            LoginImportResult as _LoginImportResult, ProfileInfo as _ProfileInfo,
+            DefaultInstalledBrowserRetriever, LoginImportResult as _LoginImportResult,
+            ProfileInfo as _ProfileInfo,
         },
         metadata::NativeImporterMetadata as _NativeImporterMetadata,
     };


harr1424 (approved):

coroiu:
@djsmith85 The modifier should be changed for least possible access. If the struct is only used in that file then you can remove `pub` completely. Otherwise, if it's used somewhere else in the same crate you can use `pub(crate)`

Hinton:
> Do we need to change the access modifiers on DefaultInstalledBrowserRetriever in [chromium.rs](https://github.com/bitwarden/clients/blob/main/apps/desktop/desktop_native/bitwarden_chromium_importer/src/chromium.rs)?

I don't think so since it's referenced in other places, specifically the `napi` crate. I'm not sure why it's a trait in the first place since we never plan on having another implementation of it. You might be able to decrease the visibility of the trait, but I suspect it's exposed as part of the other functions.

PR #17042, opened 2025-10-27
[PM-24102] Remove decrypt function from encrypt service

quexten:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-24102

## 📔 Objective

EncString has a `decrypt` function which currently relies on active userstate in order to find the correct key. The actual userid needs to be passed through to the point of decryption, both because this is safer, but also because `orgKeys$` requires the userid as a parameter.

This PR removes the decrypt function and moves the decryption code into the callers. The callers have the userid passed through the domain decryption layers.

All domain decryption code here will eventually be dropped by vault, once the SDK feature flag is removed.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/apps/browser/src/autofill/services/insert-autofill-content.service.spec.ts
+++ b/apps/browser/src/autofill/services/insert-autofill-content.service.spec.ts
@@ -26,7 +26,6 @@ const eventsToTest = [
   EVENTS.CHANGE,
   EVENTS.INPUT,
   EVENTS.KEYDOWN,
-  EVENTS.KEYPRESS,
   EVENTS.KEYUP,
   "blur",
   "click",
@@ -1044,13 +1043,13 @@ describe("InsertAutofillContentService", () => {
   });
 
   describe("simulateUserKeyboardEventInteractions", () => {
-    it("will trigger `keydown`, `keypress`, and `keyup` events on the passed element", () => {
+    it("will trigger `keydown` and `keyup` events on the passed element", () => {
       const inputElement = document.querySelector('input[type="text"]') as HTMLInputElement;
       jest.spyOn(inputElement, "dispatchEvent");
 
       insertAutofillContentService["simulateUserKeyboardEventInteractions"](inputElement);
 
-      [EVENTS.KEYDOWN, EVENTS.KEYPRESS, EVENTS.KEYUP].forEach((eventName) => {
+      [EVENTS.KEYDOWN, EVENTS.KEYUP].forEach((eventName) => {
         expect(inputElement.dispatchEvent).toHaveBeenCalledWith(
           new KeyboardEvent(eventName, { bubbles: true }),
         );
--- a/apps/browser/src/autofill/services/insert-autofill-content.service.ts
+++ b/apps/browser/src/autofill/services/insert-autofill-content.service.ts
@@ -136,7 +136,7 @@ class InsertAutofillContentService implements InsertAutofillContentServiceInterf
       setTimeout(() => {
         this.autofillInsertActions[action]({ opid, value });
         resolve();
-      }, delayActionsInMilliseconds * actionIndex),
+      }, delayActionsInMilliseconds),
     );
   };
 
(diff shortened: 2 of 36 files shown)


apps/web/src/app/tools/send/send-access/access.component.ts
```diff
@@ -118,7 +122,8 @@ export class AccessComponent implements OnInit {
       this.passwordRequired = false;
       const sendAccess = new SendAccess(sendResponse);
       this.decKey = await this.keyService.makeSendKey(keyArray);
-      this.send = await sendAccess.decrypt(this.decKey);
+      const activeUserId = await firstValueFrom(this.accountService.activeAccount$.pipe(getUserId));
+      this.send = await sendAccess.decrypt(activeUserId, this.decKey);
```
djsmith85: ⚠️  Accessing a Send does not require a Bitwarden account, so there won't be an activeUserId. The correct key to decrypt is already provided in this scenario.

I know this is just bubbling up the chain, due to the changes on `domain-base.ts`. Maybe this means SendAcess shouldn't rely on that base.

libs/importer/src/importers/bitwarden/bitwarden-json-importer.ts
```diff
@@ -167,6 +168,8 @@ export class BitwardenJsonImporter extends BaseImporter implements Importer {
   private async parseFolders(
     data: BitwardenUnEncryptedIndividualJsonExport | BitwardenEncryptedIndividualJsonExport,
   ): Promise<Map<string, number>> | null {
+    const userId: UserId = await firstValueFrom(this.accountService.activeAccount$.pipe(getUserId));
+
```
djsmith85: ⚠️ The accountId is already retrieved in the parseEncrypted method, which is one of the parents of parseFolders. The accountId should only be requested once in the parse()-method and passed down into parseDecrypted and parseEncrypted.

libs/importer/src/importers/bitwarden/bitwarden-json-importer.ts
```diff
@@ -178,7 +181,7 @@ export class BitwardenJsonImporter extends BaseImporter implements Importer {
       if (data.encrypted) {
         const folder = FolderWithIdExport.toDomain(f);
         if (folder != null) {
-          folderView = await folder.decrypt();
+          folderView = await folder.decrypt(userId);
```
djsmith85, line 184: ❓ If instead of the userId we obtain the key instead, should we then use `decryptWithKey`?

apps/cli/src/tools/send/commands/receive.command.ts
```diff
@@ -160,7 +165,7 @@ export class SendReceiveCommand extends DownloadCommand {
 
       const sendAccess = new SendAccess(sendResponse);
       this.decKey = await this.keyService.makeSendKey(key);
-      return await sendAccess.decrypt(this.decKey);
+      return await sendAccess.decrypt(activeUserId, this.decKey);
```
djsmith85: ⚠️  Accessing a Send does not require a Bitwarden account, so there won't be an activeUserId. The correct key to decrypt is already provided in this scenario.

apps/web/src/app/tools/send/send-access/access.component.ts
quexten: Thanks for catching this. I went through these pretty mechanically as it's quite a lot of domain objects and the tools one seem to be different.

quexten:
Considering this PR's amount of team reviews, I'll move it to draft and re-structure it to be reviewable per team.

libs/importer/src/importers/bitwarden/bitwarden-json-importer.ts
quexten: That seems correct. 

I don't know how recently these APIs have actually been reviewed, and as far as I'm concerned vault is moving these to the SDK imminently and should drop support for the TS versions at some point after that.

apps/cli/src/tools/send/commands/receive.command.ts
quexten: [ca2bc2b](https://github.com/bitwarden/clients/pull/17042/commits/ca2bc2b9c56efb1ce50c68682fe29e0d596108ab)

libs/importer/src/importers/bitwarden/bitwarden-json-importer.ts
quexten: [ca2bc2b](https://github.com/bitwarden/clients/pull/17042/commits/ca2bc2b9c56efb1ce50c68682fe29e0d596108ab)

apps/web/src/app/tools/send/send-access/access.component.ts
quexten: [ca2bc2b](https://github.com/bitwarden/clients/pull/17042/commits/ca2bc2b9c56efb1ce50c68682fe29e0d596108ab)

PR #17056, opened 2025-10-27
[PM-27103] Add URL Check to Send

Jingo88:
## 🎟️ Tracking

[PM-27103](https://bitwarden.atlassian.net/browse/PM-27103?actionerId=712020%3A11a1f962-ca1f-49ea-85d7-4ab89b7f8866&sourceType=assign)

## 📔 Objective

Adding a `dangerousPatterns` check to the send method. 

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

[PM-27103]: https://bitwarden.atlassian.net/browse/PM-27103?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/common/src/platform/misc/utils.ts
+++ b/libs/common/src/platform/misc/utils.ts
@@ -620,16 +620,16 @@ export class Utils {
   static invalidUrlPatterns(url: string): boolean {
     const invalidUrlPatterns = ["..", "%2e", "\\", "%5c"];
 
-    const lowerCaseUrl: string = url.toLocaleLowerCase();
+    const decodedUrl = decodeURIComponent(url.toLocaleLowerCase());
 
     // Check URL for invalidUrl patterns across entire URL
-    if (invalidUrlPatterns.some((p) => lowerCaseUrl.includes(p))) {
+    if (invalidUrlPatterns.some((p) => decodedUrl.includes(p))) {
       return true;
     }
 
     // Check for additional invalid patterns inside URL params
-    if (lowerCaseUrl.includes("?")) {
-      const hasInvalidParams = this.validateQueryParameters(lowerCaseUrl);
+    if (decodedUrl.includes("?")) {
+      const hasInvalidParams = this.validateQueryParameters(decodedUrl);
       if (hasInvalidParams) {
         return true;
       }
--- a/libs/common/src/platform/misc/utils.spec.ts
+++ b/libs/common/src/platform/misc/utils.spec.ts
@@ -699,8 +699,7 @@ describe("Utils Service", () => {
     });
 
     it("should return true if an invalid pattern is found", () => {
-      const urlString =
-        "https://www.example.com/api/testing/../../organizationId=70d708c9%5c%2e%2e%5c%2e%2e%5";
+      const urlString = "https://www.example.com/api/%2e%2e/secret";
 
       const actual = Utils.invalidUrlPatterns(urlString);
 
@@ -708,8 +707,7 @@ describe("Utils Service", () => {
     });
 
     it("should return true if an invalid pattern is found in a param", () => {
-      const urlString =
-        "https://www.example.com/api/history?someToken=70d708c9%5c%2e%2e%5c%2e%2e%5";
+      const urlString = "https://www.example.com/api/history?someToken=../secret";
 
       const actual = Utils.invalidUrlPatterns(urlString);
 


libs/common/src/platform/misc/utils.ts
```diff
@@ -612,6 +612,21 @@ export class Utils {
...
+    const lowerCaseUrl: string = url.toLocaleLowerCase();
+
+    // Check URL for dangerous patterns
+    for (const pattern of dangerousPatterns) {
+      if (lowerCaseUrl.includes(pattern)) {
+        return true;
+      }
+    }
+
+    return false;
```
eliykat: 🎨 (non-blocking): prefer higher order methods over explicit control flow, e.g. `return dangerousPatterns.some(p => lowerCaseUrl.includes(p));`.

Or you could have a regex that covers all dangerous patterns in a single check.

libs/admin-console/src/common/organization-user/services/default-organization-user-api.service.ts
```diff
@@ -159,7 +159,11 @@ export class DefaultOrganizationUserApiService implements OrganizationUserApiSer
   ): Promise<void> {
     return this.apiService.send(
       "POST",
-      "/organizations/" + organizationId + "/users/" + id + "/accept-init",
+      "/organizations/" +
+        encodeURIComponent(organizationId) +
+        "/users/" +
+        encodeURIComponent(id) +
+        "/accept-init",
```
eliykat: What about all our other api service methods? (in this class and others) I don't understand why this method is being updated specifically.

Suggestions for improvements (may or may not be out of scope): can this be handled at a lower level (e.g. in apiService.send) so that we don't have to manually handle this for every api method's parameters? Or perhaps we could use a decorator on these methods that intercepts and transforms the parameters?

libs/common/src/services/api.service.ts
```diff
@@ -1587,6 +1587,12 @@ export class ApiService implements ApiServiceAbstraction {
     const requestUrl =
       apiUrl + Utils.normalizePath(pathParts[0]) + (pathParts.length > 1 ? `?${pathParts[1]}` : "");
```
eliykat: The jsdoc for `normalizePath` says:

> Normalizes a path for defense against attacks like traversals

Why doesn't `normalizePath` protect us at the moment? Should we modify that approach rather than layering on an additional approach? (this also relates to Claude's feedback re: the order of operations here)

Jingo88: Yes I'm not sure normalizePath can sanitize or fix url to fully protect us here. Will look into modifying this.

libs/admin-console/src/common/organization-user/services/default-organization-user-api.service.ts
Jingo88: That's true. I was targeting this route specifically from the issue in JIRA, just in case a URL passes the check in send. But I think i'm going to take this back to the drawing board a bit and expand the checks inside of the apiService.send method.

libs/common/src/platform/misc/utils.ts
```diff
@@ -604,12 +603,52 @@ export class Utils {
...
 
   /**
-   * Normalizes a path for defense against attacks like traversals
-   * @param denormalizedPath
-   * @returns
+   * Validates an url checking against invalid patterns
+   * @param url
+   * @returns true if invalid patterns found, false if safe
+   */
+  static invalidUrlPatterns(url: string): boolean {
```
gbubemismith, line 620: I think we still need url decoding alongside this

libs/common/src/services/api.service.ts
gbubemismith: @eliykat I think the `normalizePath` only handles decoding,  @Jingo88 I think we need both

libs/common/src/platform/misc/utils.ts
```diff
@@ -604,12 +603,52 @@ export class Utils {
...
+    return false;
+  }
+
+  /**
+   * Validates query parameters for additional invalid patterns
+   * @param url - The URL containing query parameters
+   * @returns true if invalid patterns found, false if safe
    */
-  static normalizePath(denormalizedPath: string): string {
-    return path.normalize(decodeURIComponent(denormalizedPath)).replace(/^(\.\.(\/|\\|$))+/, "");
```
eliykat, line 612: I assume we still want to normalize valid paths. My previous comment didn't necessarily mean to get rid of this entirely, only to consider why this doesn't protect us at the moment and how they interact. Maybe we do need both - I'm not sure - but I would be very cautious about removing normalization altogether.

eliykat (commented):
I'm no longer a codeowner so my review is not required. However I do think that Platform Team should be considered the owner of the `ApiService.send` method given that it underpins all API calls in the application - @dereknance can you or your team please take a look at this PR closely. These kinds of encoding issues can be tricky so more eyes is better, and this is arguably Platform Team's domain. Also make sure to check out Claude's updated review.

libs/common/src/platform/misc/utils.ts
Jingo88: My understanding of the normalize was it could hinder us by accepting a faulty url. an example may be taking this`%5c%2e%2e%5clogout` and returning this `logout` logging out the user. 

I am cautious about this removal as well and would like some more in depth feedback. @dereknance

Jingo88: Hm that's a fair point. I'm cautious as well about removing this. 

My current impression is that the normalizePath would hinder us with regards to faulty urls. For example it may take something like this `%2F%2E%2E%2Flogout`, return `logout` and log out the user. 

I would also like any deeper insight on this when possible. @dereknance 

I'll bring the method back in for now until further review, but change the order of operations so the invalid check runs first.

dereknance (approved):

dereknance:
> @dereknance can you or your team please take a look at this PR closely. These kinds of encoding issues can be tricky so more eyes is better, and this is arguably Platform Team's domain.

I'm going to tag in @addisonbeck for backup too since this is a security-related PR and I'm still coming up to speed on the gritty details of the things that Platform owns.

As far as the problem is concerned, it's not totally clear to me how the path traversal is occurring in the first place, if the traversal bits are part of the query parameter part of the URL. Is there code that is reading those parameters and then producing a URL path from them for a redirect or something like that?

addisonbeck (approved):

PR #17063, opened 2025-10-27
[PM-27506] CLI allows creating SSH key items with null fields

gbubemismith:
## 🎟️ Tracking
https://bitwarden.atlassian.net/browse/PM-27506

## 📸 Screenshots
image

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/libs/common/src/models/export/ssh-key.export.ts
+++ b/libs/common/src/models/export/ssh-key.export.ts
@@ -10,9 +10,9 @@ import { safeGetString } from "./utils";
 export class SshKeyExport {
   static template(): SshKeyExport {
     const req = new SshKeyExport();
-    req.privateKey = "-----BEGIN OPENSSH PRIVATE KEY-----\n...\n-----END OPENSSH PRIVATE KEY-----";
-    req.publicKey = "ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAA... user@host";
-    req.keyFingerprint = "SHA256:abcdef1234567890abcdef1234567890abcdef12";
+    req.privateKey = "";
+    req.publicKey = "";
+    req.keyFingerprint = "";
     return req;
   }
 
--- a/libs/common/src/models/export/cipher.export.spec.ts
+++ b/libs/common/src/models/export/cipher.export.spec.ts
@@ -3,6 +3,8 @@ import { SecureNoteExport } from "@bitwarden/common/models/export/secure-note.ex
 import { CipherType } from "@bitwarden/common/vault/enums";
 import { CipherView } from "@bitwarden/common/vault/models/view/cipher.view";
 
+import { SshKeyExport } from "./ssh-key.export";
+
 describe("Cipher Export", () => {
   describe("toView", () => {
     it.each([[null], [undefined]])(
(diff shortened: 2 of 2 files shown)


libs/common/src/models/export/ssh-key.export.ts
```diff
@@ -17,6 +17,21 @@ export class SshKeyExport {
   }
 
   static toView(req: SshKeyExport, view = new SshKeyView()) {
```
harr1424: Hey Gbubemi 👋 

As our team encounters files containing 
```
// FIXME: Update this file to be type safe and remove this and next line
// @ts-strict-ignore
```
we have been trying to eliminate some tech debt by adding type safety (as reasonable). 

🌱  Would you be willing to change the function signature to match the below? 

```typescript
static toView(req?: SshKeyExport, view = new SshKeyView()): SshKeyView | undefined {

// if req == undefined is probably more reasonable, but really neither should occur, it looks like the issue was isolated to missing publicKey and keyFingerprint, req should never be entirely null or undefined 

  if (req == null) { 
    return undefined;
  }
  
  // Validate required fields
  if (!req.privateKey || req.privateKey.trim() === "") {
    throw new Error("SSH key private key is required.");
  }
  if (!req.publicKey || req.publicKey.trim() === "") {
    throw new Error("SSH key public key is required.");
  }
  if (!req.keyFingerprint || req.keyFingerprint.trim() === "") {
    throw new Error("SSH key fingerprint is required.");
  }
  
  view.privateKey = req.privateKey;
  view.publicKey = req.publicKey;
  view.keyFingerprint = req.keyFingerprint;
  return view;
}
```

gbubemismith: `  if (req == null)` checks both null and undefined so I think it should be fine

libs/common/src/models/export/ssh-key.export.ts
```diff
@@ -10,13 +10,28 @@ import { safeGetString } from "./utils";
 export class SshKeyExport {
   static template(): SshKeyExport {
     const req = new SshKeyExport();
-    req.privateKey = "";
-    req.publicKey = "";
-    req.keyFingerprint = "";
+    req.privateKey = "-----BEGIN OPENSSH PRIVATE KEY-----\n...\n-----END OPENSSH PRIVATE KEY-----";
+    req.publicKey = "ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAA... user@host";
+    req.keyFingerprint = "SHA256:abcdef1234567890abcdef1234567890abcdef12";
```
harr1424: ❌ I don't think we should be setting any sort of default value here, the fact that the req will be passed to a method `toView()` is a little misleading because we are only creating a CipherView object that won't actually appear in a UI element. I can see this being helpful if instead we needed default values in form fields, but it's not the case.

Additionally, if we are setting default values here, I don't think the validation would be required in the `toView()` function.

Suggestion: set these members to be `undefined` here (in the template), this is an improvement from the former empty string, but will still result in validation errors if they aren't assigned values. 

Retaining the defaults would not raise a validation error, as the members would never be `null`, `undefined`, or empty string, and the expoirt would then fail silently.

libs/common/src/models/export/ssh-key.export.ts
```diff
@@ -10,13 +10,28 @@ import { safeGetString } from "./utils";
...
+    // Validate required fields
+    if (!req.privateKey || req.privateKey.trim() === "") {
+      throw new Error("SSH key private key is required.");
+    }
+    if (!req.publicKey || req.publicKey.trim() === "") {
+      throw new Error("SSH key public key is required.");
+    }
+    if (!req.keyFingerprint || req.keyFingerprint.trim() === "") {
+      throw new Error("SSH key fingerprint is required.");
+    }
```
harr1424, line 33: ~~❌  Sorry my previous comment wasn't more clear, but I don't think it's necessary to validate `req` itself as being null, since the `SshKeyExport` constructor will always create the object. This would also allow the removal of the optionality on the `req` parameter, as suggested by Claude's review:~~

```typescript
static toView(req: SshKeyExport, view = new SshKeyView()): SshKeyView {
    if (req == null) {
      throw new Error("SSH Key Export is required.");  
    }

  // Validate required fields
  if (!req.privateKey || req.privateKey.trim() === "") {
    throw new Error("SSH key private key is required.");
  }
  if (!req.publicKey || req.publicKey.trim() === "") {
    throw new Error("SSH key public key is required.");
  }
  if (!req.keyFingerprint || req.keyFingerprint.trim() === "") {
    throw new Error("SSH key fingerprint is required.");
  }
  
  view.privateKey = req.privateKey;
  view.publicKey = req.publicKey;
  view.keyFingerprint = req.keyFingerprint;
  return view;
}
```

~~If you've set the three members to be `undefined` in the constructor, this will effectively trigger the validation errors if they haven't been set~~
 
I apologize, the null check you were originally using was the safest approach here and can be re-implemented. A large part of our codebase still sets some values to `null`, and checking against `null` is safer than `undefined`, even if it doesn't satisfy TS type safety.

harr1424 (approved):

shane-melton (approved):

PR #16974, opened 2025-10-21
[PM-25683] Migrate Cipher model and sub-models

gbubemismith:
## 🎟️ Tracking
https://bitwarden.atlassian.net/browse/PM-25683

## ⏰ Reminders before review

- Contributor guidelines followed
- All formatters and local linters executed and passed
- Written new unit and / or integration tests where applicable
- Protected functional changes with optionality (feature flags)
- Used internationalization (i18n) for all UI strings
- CI builds passed
- Communicated to DevOps any deployment requirements
- Updated any necessary documentation (Confluence, contributing docs) or informed the documentation team

## 🦮 Reviewer guidelines

- 👍 (`:+1:`) or similar for great changes
- 📝 (`:memo:`) or ℹ️ (`:information_source:`) for notes or general info
- ❓ (`:question:`) for questions
- 🤔 (`:thinking:`) or 💭 (`:thought_balloon:`) for more open inquiry that's not quite a confirmed issue and could potentially benefit from discussion
- 🎨 (`:art:`) for suggestions / improvements
- ❌ (`:x:`) or ⚠️ (`:warning:`) for more significant problems or concerns needing attention
- 🌱 (`:seedling:`) or ♻️ (`:recycle:`) for future improvements or indications of technical debt
- ⛏ (`:pick:`) for minor or nitpick changes

PR #17073, opened 2025-10-28
[PM-25684] - Migrate cipher data model to adhere to ts-strict

theMickster:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-25684

## 📔 Objective

Migrated the ciperData model and it's children to comply with `ts-strict`.
I started with the work done in [PM-25683 Migrate Cipher model and sub-models](https://github.com/bitwarden/clients/pull/16974), reviewed our API contracts, & reviewed how these data models are consumed by the domain models.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

[PM-25683]: https://bitwarden.atlassian.net/browse/PM-25683?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/common/src/vault/models/domain/login-uri.spec.ts
+++ b/libs/common/src/vault/models/domain/login-uri.spec.ts
@@ -65,7 +65,7 @@ describe("LoginUri", () => {
     });
   });
 
-  it("should convert match property from null to undefined", () => {
+  it("handle null match", () => {
     const apiData = Object.assign(new LoginUriApi(), {
       uri: "testUri",
       uriChecksum: "testChecksum",
@@ -139,7 +139,7 @@ describe("LoginUri", () => {
   });
 
   describe("SDK Login Uri Mapping", () => {
-    it("should map to SDK login uri", () => {
+    it("maps to SDK login uri", () => {
       const loginUri = new LoginUri(data);
       const sdkLoginUri = loginUri.toSdkLoginUri();
 
--- a/libs/common/src/vault/models/domain/cipher.spec.ts
+++ b/libs/common/src/vault/models/domain/cipher.spec.ts
@@ -836,6 +836,38 @@ describe("Cipher DTO", () => {
       expect(actual).toBeInstanceOf(Cipher);
     });
 
+    it("handles null permissions correctly without calling CipherPermissionsApi constructor", () => {
+      const spy = jest.spyOn(CipherPermissionsApi.prototype, "constructor" as any);
+      const revisionDate = new Date("2022-08-04T01:06:40.441Z");
+      const actual = Cipher.fromJSON({
+        name: "myName",
+        revisionDate: revisionDate.toISOString(),
+        permissions: null,
+      } as Jsonify<Cipher>);
+
+      expect(actual.permissions).toBeUndefined();
+      expect(actual).toBeInstanceOf(Cipher);
+      // Verify that CipherPermissionsApi constructor was not called for null permissions
+      expect(spy).not.toHaveBeenCalledWith(null);
+      spy.mockRestore();
+    });
+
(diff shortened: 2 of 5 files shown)


theMickster:
@claude Whatcha think about this draft PR?

theMickster:
@claude :: Update your review of my PR but only AFTER you take into consideration the changes done in the following PR
https://github.com/bitwarden/clients/pull/16974

libs/common/src/vault/models/data/fido2-credential.data.ts
```diff
@@ -1,21 +1,19 @@
...
-  userName: string;
-  counter: string;
-  rpName: string;
-  userDisplayName: string;
-  discoverable: string;
-  creationDate: string;
+  credentialId!: string;
+  keyType!: string;
+  keyAlgorithm!: string;
+  keyCurve!: string;
```
nick-livefront, line 7: :question: The types for `keyType`, `keyAlgorithm`, and `keyCurve` were broadened from specific strings to a generic string. Was that intentional?

libs/common/src/vault/models/domain/login-uri.spec.ts
```diff
@@ -61,6 +65,23 @@ describe("LoginUri", () => {
     });
   });
 
+  it("should convert match property from null to undefined", () => {
```
nick-livefront: :pick: (non-blocking) - The added test descriptions here and in other files are prefixed with "should" while the existing tests do not. This is not consistent throughout the codebase. I did (just now) find [Naming Conventions](https://contributing.bitwarden.com/contributing/testing/unit/naming-conventions) in the contributing docs for test cases, I'll now note that when reviewing to try and be consistent!

theMickster: Sorry, force of years of other habits. I'll adjust them 😃

libs/common/src/vault/models/data/fido2-credential.data.ts
theMickster: Yes, I think that they are not useful and can confuse. 
They were removed because strings properties with literal type annotations they are always overwritten at lines 24-27 with cipher text from the API response. The properties never actually hold those values because of TypeScript's handling of type inference for uninitialized properties. 

From the TS 5 specification 
> "A property declaration declares a property member. Property members can have type annotations but cannot have initializers unless they are readonly properties with literal types or ambient declarations."

So because there's no initializer, I felt the need to clarify and remove.  

My guess is that the teammate was thinking about literal string type or constants when originally writing these models but then quickly pivoted to them just taking the cipher text from the API response.  Or was thinking about self documenting the code 🤷🏼‍♂️ to try and show that the decrypted values should always be these values from the API? Either way, I think the ambiguity leans toward removing them and just letting the API feed us the cipher text to decrypt in the domain model.

nick-livefront: 👏 I appreciate the explanation and that makes sense to me!

theMickster: All that to say, if we really feel strongly then I can put them back.

nick-livefront: It's all good with me! Especially because these values are being replaced by the API. I rather be board and force consumers to narrow down the value than mistakenly lie and have a new value be introduced without our knowing.

libs/common/src/vault/models/domain/login-uri.spec.ts
theMickster: Adjusted as requested :-)

PR #17077, opened 2025-10-28
[PM-25685][PM-31077] - Migrate all Folder models

jaasen-livefront:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-25685
https://bitwarden.atlassian.net/browse/PM-31077

## 📔 Objective

This PR migrates the existing folder models to enforce strict type checking. Please see ticket description for additional details.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


libs/common/src/vault/services/folder/folder-api.service.ts
```diff
@@ -17,11 +17,11 @@ export class FolderApiService implements FolderApiServiceAbstraction {
...
     let response: FolderResponse;
-    if (folder.id == null) {
+    if (folder.id) {
+      response = await this.putFolder(folder.id, request);
+    } else {
       response = await this.postFolder(request);
       folder.id = response.id;
-    } else {
-      response = await this.putFolder(folder.id, request);
     }
```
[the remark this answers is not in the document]
jaasen-livefront: `putFolder` and `postFolder` were also reversed so the logic hasn't changed. We're just ensuring folder id is present and not just not null.

libs/common/src/vault/models/view/folder.view.ts
```diff
@@ -8,12 +6,13 @@ import { Folder } from "../domain/folder";
 import { ITreeNodeObject } from "../domain/tree-node";
 
 export class FolderView implements View, ITreeNodeObject {
-  id: string = null;
-  name: string = null;
-  revisionDate: Date = null;
+  id: string = "";
+  name: string = "";
+  revisionDate: Date;
```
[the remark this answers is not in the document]
jaasen-livefront: The `folder.id === null` checks have been adjusted to account for the new string value.

libs/common/src/vault/models/domain/folder.ts
```diff
@@ -10,19 +8,20 @@ import { FolderData } from "../data/folder.data";
 import { FolderView } from "../view/folder.view";
 
 export class Test extends Domain {
-  id: string;
-  name: EncString;
-  revisionDate: Date;
+  id: string = "";
+  name: EncString = new EncString("");
+  revisionDate: Date = new Date();
 }
```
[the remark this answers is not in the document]
jaasen-livefront: removed

Jingo88 (approved):

libs/common/src/models/export/folder.export.ts
```diff
@@ -19,14 +19,12 @@ export class FolderExport {
   }
 
   static toDomain(req: FolderExport, domain = new FolderDomain()) {
-    domain.name = req.name != null ? new EncString(req.name) : null;
+    domain.name = new EncString(req.name);
```
Hinton, line 22: **issue:** This is not functionally equivalent.

libs/common/src/models/export/folder.export.ts
```diff
@@ -19,14 +19,12 @@ export class FolderExport {
...
     return domain;
   }
 
-  name: string;
-
   // Use build method instead of ctor so that we can control order of JSON stringify for pretty print
   build(o: FolderView | FolderDomain) {
-    this.name = safeGetString(o.name);
+    this.name = safeGetString(o.name) ?? "";
```
Hinton: **issue:** This is not functionally equivalent.

libs/common/src/vault/models/domain/folder.ts
```diff
@@ -9,20 +7,15 @@ import { SymmetricCryptoKey } from "../../../platform/models/domain/symmetric-cr
...
-export class Test extends Domain {
-  id: string;
-  name: EncString;
-  revisionDate: Date;
-}
-
 export class Folder extends Domain {
-  id: string;
-  name: EncString;
+  id: string = "";
```
Hinton, line 11: I'm pretty sure id can be optional, when you create a folder ID should be null/undefined.

libs/common/src/models/export/folder.export.ts
jaasen-livefront: @Hinton Indeed, but since the folder model requires `name` to be a type of `EncString` we can no longer return `null`.

libs/common/src/models/export/folder.export.ts
jaasen-livefront: @Hinton We're modelling `FolderExport` to match the `Folder` domain so we now requires `name` to be of type `string`. Since `safeGetString` can return `undefined` we must fallback to an empty string. Do you suggest we have `name` be optional?

libs/common/src/vault/models/domain/folder.ts
jaasen-livefront: Fixed https://github.com/bitwarden/clients/pull/17077/commits/939ec1b448e4bd2aad89488f1d2cc09dadaf5fe2

libs/common/src/models/export/folder.export.ts
jaasen-livefront: Fixed https://github.com/bitwarden/clients/pull/17077/commits/939ec1b448e4bd2aad89488f1d2cc09dadaf5fe2

libs/common/src/models/export/folder.export.ts
jaasen-livefront: Fixed https://github.com/bitwarden/clients/pull/17077/commits/939ec1b448e4bd2aad89488f1d2cc09dadaf5fe2. `name` is now optional so this should now return apply the expected types (`string | undefined`)

libs/common/src/vault/models/domain/folder.ts
```diff
@@ -48,10 +41,10 @@ export class Folder extends Domain {
     encryptService: EncryptService,
   ): Promise<FolderView> {
     const folderView = new FolderView();
-    folderView.id = this.id;
+    folderView.id = this.id ?? "";
```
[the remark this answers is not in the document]
jaasen-livefront: `id` In `FolderView` is _not_ optional.

jaasen-livefront:
@Hinton @Jingo88 Ready for another look when you have a chance.

libs/common/src/models/export/folder.export.ts
Hinton: This is still a behavioural change. If name is now optional all changes to this file should be reverted.

libs/tools/export/vault-export/vault-export-core/src/services/individual-vault-export.service.spec.ts
Hinton, line 1: Why are we changing this test?

Hinton (commented):
There are still various behavioral changes across the PR. See my DM for testing scope if these are desired.

I also think some of the claude comments are relevant.

libs/tools/export/vault-export/vault-export-core/src/services/individual-vault-export.service.spec.ts
jaasen-livefront: @Hinton I've reverted it back to close to the original. We're now checking for object equality instead of string equality which makes the tests less brittle.

libs/common/src/models/export/folder.export.ts
jaasen-livefront: @Hinton Sorry about the confusion. Since `FolderExport` requires `name` we can just check for presence now. I've updated the code to reflect this.

libs/angular/src/vault/components/folder-add-edit.component.ts
```diff
@@ -58,7 +58,7 @@ export class FolderAddEditComponent implements OnInit {
 
   async submit(): Promise<boolean> {
     this.folder.name = this.formGroup.controls.name.value;
-    if (this.folder.name == null || this.folder.name === "") {
+    if (this.folder.name === "") {
```
shane-melton: 🎨 We shouldn't remove this null check yet.

libs/common/src/vault/models/domain/folder.ts
```diff
@@ -9,20 +7,15 @@ import { SymmetricCryptoKey } from "../../../platform/models/domain/symmetric-cr
...
-  id: string;
-  name: EncString;
-  revisionDate: Date;
-}
-
 export class Folder extends Domain {
-  id: string;
-  name: EncString;
+  id?: string;
+  name?: EncString;
```
shane-melton: ❓ Why are these marked as optional in `Folder` but required in the other models? I would expect them to be consistent across models.

libs/common/src/vault/models/domain/folder.ts
shane-melton: 🎨 If we're setting the `name` and revision date manually then we no longer need to call this `buildDomainModel()` method.

libs/common/src/vault/models/domain/folder.ts
```diff
@@ -9,20 +7,15 @@ import { SymmetricCryptoKey } from "../../../platform/models/domain/symmetric-cr
...
   revisionDate: Date;
 
   constructor(obj?: FolderData) {
     super();
     if (obj == null) {
+      this.revisionDate = new Date();
```
shane-melton: 💭 This is inconsistent with `name`. Why are we setting `revisionDate` to a default value and not `name`?

libs/common/src/vault/models/domain/folder.ts
```diff
@@ -48,10 +41,10 @@ export class Folder extends Domain {
     encryptService: EncryptService,
   ): Promise<FolderView> {
     const folderView = new FolderView();
-    folderView.id = this.id;
+    folderView.id = this.id ?? "";
     folderView.revisionDate = this.revisionDate;
     try {
-      folderView.name = await encryptService.decryptString(this.name, key);
+      folderView.name = await encryptService.decryptString(this.name ?? new EncString(""), key);
```
shane-melton: ⚠️ Passing an `EncString("")` to decryptString doesn't really mean anything or serve a purpose here. If the encrypted name is missing, we should just skip decryption.

libs/common/src/vault/models/domain/folder.ts
```diff
@@ -62,7 +55,13 @@ export class Folder extends Domain {
   }
 
   static fromJSON(obj: Jsonify<Folder>) {
-    const revisionDate = obj.revisionDate == null ? null : new Date(obj.revisionDate);
-    return Object.assign(new Folder(), obj, { name: EncString.fromJSON(obj.name), revisionDate });
+    if (obj == null) {
+      return null;
+    }
+    return new Folder({
+      name: obj.name ?? "",
```
shane-melton: 🎨  Claude makes some potentially good points here, it seems we're changing/obsfucating this implementation here when we shouldn't need to.

We should be able to call the empty `Folder()` constructor and explicitly assign the the `name` and `revisionDate` properties. We should also continue to use the `fromJson()` helpers for EncString for consistency. See the `Cipher.fromJSON` method for an example.

libs/common/src/vault/models/view/folder.view.ts
```diff
@@ -8,21 +6,26 @@ import { Folder } from "../domain/folder";
...
 
-  constructor(f?: Folder | DecryptedObject<Folder, "name">) {
+  constructor(f?: Folder | DecryptedObject<Folder, undefined>) {
```
shane-melton: ❓ `undefined` doesn't seem right here?

On second look, I don't see any usage of the `FolderView` constructor where we pass in a decrypted Folder object. I think its safe to just remove this type.

libs/common/src/vault/models/view/folder.view.ts
```diff
@@ -8,21 +6,26 @@ import { Folder } from "../domain/folder";
...
     if (!f) {
+      this.revisionDate = new Date();
       return;
     }
 
-    this.id = f.id;
+    this.id = f.id ?? "";
+    this.name = f.name?.decryptedValue ?? "";
```
shane-melton: 🎨 Since we never pass in a decrypted folder, I don't think we need to add this.

libs/common/src/vault/models/view/folder.view.ts
```diff
@@ -8,21 +6,26 @@ import { Folder } from "../domain/folder";
...
     this.revisionDate = f.revisionDate;
   }
 
   static fromJSON(obj: Jsonify<FolderView>) {
-    const revisionDate = obj.revisionDate == null ? null : new Date(obj.revisionDate);
-    return Object.assign(new FolderView(), obj, { revisionDate });
+    const folderView = new FolderView();
+    folderView.id = obj.id ?? "";
+    folderView.name = obj.name ?? "";
+    folderView.revisionDate = new Date(obj.revisionDate);
```
shane-melton: ⚠️ I'd expect this to be a type error as you can't pass a `| undefined` into `new Date()`. We need to keep the check and only assign if the `obj.revisionDate` has a value.

See here in CipherView: https://github.com/bitwarden/clients/blob/7e5f02f90c7bdc5a575de471d7c63601ed326750/libs/common/src/vault/models/view/cipher.view.ts#L225-L227

libs/common/src/vault/models/view/folder.view.ts
```diff
@@ -8,21 +6,26 @@ import { Folder } from "../domain/folder";
 import { ITreeNodeObject } from "../domain/tree-node";
 
 export class FolderView implements View, ITreeNodeObject {
-  id: string = null;
-  name: string = null;
-  revisionDate: Date = null;
+  id: string = "";
```
shane-melton, line 8: ℹ️ This is going to break our checks for the "No Folder" folder that assumes a `null` id. We'll need to do a thorough search of the clients to be sure we catch them all. I see you touched the browser and web services, but nothing for Desktop.

jaasen-livefront: @shane-melton I didn't see any explicit checks for a `null` id in the desktop client. Can you think of anywhere we many be doing that?

libs/common/src/vault/models/view/folder.view.ts
jaasen-livefront: fixed https://github.com/bitwarden/clients/pull/17077/commits/ab47d10038a8cb2a0961bcbdb5963a1f53dbfbb5

libs/common/src/vault/models/view/folder.view.ts
jaasen-livefront: fixed https://github.com/bitwarden/clients/pull/17077/commits/ab47d10038a8cb2a0961bcbdb5963a1f53dbfbb5

libs/common/src/vault/models/view/folder.view.ts
jaasen-livefront: fixed https://github.com/bitwarden/clients/pull/17077/commits/e09cab210ca6ab6ceb0a87d69ea0e5ddcad390f0

libs/common/src/vault/models/domain/folder.ts
jaasen-livefront: fixed https://github.com/bitwarden/clients/commit/ab47d10038a8cb2a0961bcbdb5963a1f53dbfbb5

libs/common/src/vault/models/view/folder.view.ts
```diff
@@ -1,28 +1,32 @@
...
 
-  constructor(f?: Folder | DecryptedObject<Folder, "name">) {
+  constructor(f?: Folder) {
     if (!f) {
+      this.revisionDate = new Date();
       return;
     }
 
     this.id = f.id;
+    this.name = f.name.toString();
```
shane-melton: ⚠️ This is setting the decrypted `name` property to an encrypted value. We should just skip the assignment and leave the default value.

libs/common/src/vault/models/view/folder.view.ts
shane-melton: There is this filter service here: https://github.com/bitwarden/clients/blob/d0d1359ff4fd4d40edaa5f1096373d642bafc40d/libs/angular/src/vault/vault-filter/services/vault-filter.service.ts#L89

and here:

https://github.com/bitwarden/clients/blob/7a38b226675ee9c2ffc6b0c8d5e48df8beecac93/apps/web/src/app/vault/individual-vault/vault-filter/services/vault-filter.service.ts#L298

shane-melton: I also found another todo here:

https://github.com/bitwarden/clients/blob/9f0a565241869d7a2a135347af99ff042db84d56/libs/importer/src/importers/base-importer.ts#L276-L289

shane-melton: A few other places:

https://github.com/bitwarden/clients/blob/dbec02cf8d4e23df9765b8f13d28ab25f9139e2f/apps/web/src/app/vault/individual-vault/vault-filter/shared/models/routed-vault-filter-bridge.model.ts#L90

https://github.com/bitwarden/clients/blob/64105e64e995900b13488e0dfef0f51c8b125631/libs/tools/export/vault-export/vault-export-core/src/services/individual-vault-export.service.ts#L271

https://github.com/bitwarden/clients/blob/64105e64e995900b13488e0dfef0f51c8b125631/libs/tools/export/vault-export/vault-export-core/src/services/individual-vault-export.service.ts#L305

shane-melton (changes requested):
There are a few other places that are checking `null` that I found that'll need to be updated.

jaasen-livefront:
> There are a few other places that are checking null that I found that'll need to be updated.

@shane-melton Ah yes that was very naive of me to only check for `folder.id`. I found a few more while doing a global search for `f.id`. 

Fixed https://github.com/bitwarden/clients/pull/17077/commits/90fce663a0868644756fa431e22745bb4258314e

libs/common/src/vault/models/view/folder.view.ts
```diff
@@ -1,19 +1,17 @@
...
 export class FolderView implements View, ITreeNodeObject {
-  id: string = null;
-  name: string = null;
-  revisionDate: Date = null;
+  id: string = "";
+  name: string = "";
+  revisionDate: Date;
 
-  constructor(f?: Folder | DecryptedObject<Folder, "name">) {
+  constructor(f?: Folder) {
```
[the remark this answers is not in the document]
jaasen-livefront: @shane-melton What do you think about this?

libs/importer/src/importers/base-importer.ts
```diff
@@ -279,8 +279,7 @@ export abstract class BaseImporter {
       const collection = new CollectionView({
         name: f.name,
         organizationId: this.organizationId,
-        // FIXME: Folder.id may be null, this should be changed when refactoring Folders to be ts-strict
-        id: Collection.isCollectionId(f.id) ? f.id : null,
+        id: f.id as CollectionId,
```
shane-melton: ⚠️ ❓ This may cause issues with folder/collection imports. Before, folder Ids would be `null` would by copied over to the `collectionView` id. Now folderId will be an empty string (the new default behavior) and never null, meaning we're now creating collectionViews with `""` for Ids. 

I'm not certain of the downstream effects this may have. It looks fairly safe based on my investigation of the import endpoint in Server. With that said, we should likely continue to pass null if the folderId is an empty string to avoid changing this behavior.

jaasen-livefront: @shane-melton Totally agree. Fixed https://github.com/bitwarden/clients/pull/17077/commits/a2b825d7e367fab336b2d20cef2e986fbe967c33

libs/importer/src/importers/base-importer.ts
```diff
@@ -279,8 +279,7 @@ export abstract class BaseImporter {
       const collection = new CollectionView({
         name: f.name,
         organizationId: this.organizationId,
-        // FIXME: Folder.id may be null, this should be changed when refactoring Folders to be ts-strict
-        id: Collection.isCollectionId(f.id) ? f.id : null,
+        id: (f.id as CollectionId) ?? null,
```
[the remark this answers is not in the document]
shane-melton: @jaasen-livefront Can we make it explicit / more clear as suggested?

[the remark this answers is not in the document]
jaasen-livefront: @shane-melton Done! https://github.com/bitwarden/clients/pull/17077/commits/f4c9c27e84f40b7dae6c4b41a987502fc009c792

libs/common/src/vault/models/view/folder.view.ts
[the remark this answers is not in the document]
shane-melton: I think its fine to keep it as is and consistent with our other models even if not used.

shane-melton:
@jaasen-livefront Looks like we have a failing test: [Password Depot 17 Xml Importer](https://github.com/bitwarden/clients/actions/runs/19867380670#user-content-r0s462)

jaasen-livefront:
@shane-melton @Hinton This has now passed QA. If you guys have time to take one last look that would be much appreciated!

PR #14611, opened 2025-05-02
202505 notifications refactor

blackwood:
## 🎟️ Tracking

N/A

## 📔 Objective

Experimental Notifications refacator:

- Using TypeScript generics functionality to pass conditional data along to the message queue items.
- Create a generalized notification component with slots, create a "Generic Notification" proof of concept which can render any elements to these slots. (@TODO fix CSS issues in storybook/ensure not duplicated in browser)

## ⏰ Reminders before review

- Contributor guidelines followed
- All formatters and local linters executed and passed
- Written new unit and / or integration tests where applicable
- Protected functional changes with optionality (feature flags)
- Used internationalization (i18n) for all UI strings
- CI builds passed
- Communicated to DevOps any deployment requirements
- Updated any necessary documentation (Confluence, contributing docs) or informed the documentation team

## 🦮 Reviewer guidelines

- 👍 (`:+1:`) or similar for great changes
- 📝 (`:memo:`) or ℹ️ (`:information_source:`) for notes or general info
- ❓ (`:question:`) for questions
- 🤔 (`:thinking:`) or 💭 (`:thought_balloon:`) for more open inquiry that's not quite a confirmed issue and could potentially benefit from discussion
- 🎨 (`:art:`) for suggestions / improvements
- ❌ (`:x:`) or ⚠️ (`:warning:`) for more significant problems or concerns needing attention
- 🌱 (`:seedling:`) or ♻️ (`:recycle:`) for future improvements or indications of technical debt
- ⛏ (`:pick:`) for minor or nitpick changes

PR #17087, opened 2025-10-28
[PM-27569] Typing cleanup

jprusik:
## 🎟️ Tracking

[PM-27569](https://bitwarden.atlassian.net/browse/PM-27569)

## 📔 Objective

incremental removals of `@ts-strict-ignore` in autofill file ownerships, and subsequent typing resolutions. Other optimizations/non-typing issues have been addressed minimally, as concerns beyond the core scope of this PR.

## 🦮 Reviewer guidelines

[PM-27569]: https://bitwarden.atlassian.net/browse/PM-27569?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/browser/src/autofill/browser/main-context-menu-handler.ts
+++ b/apps/browser/src/autofill/browser/main-context-menu-handler.ts
@@ -177,11 +177,11 @@ export class MainContextMenuHandler {
 
     try {
       const account = await firstValueFrom(this.accountService.activeAccount$);
-      const hasPremium = account?.id
-        ? await firstValueFrom(
-            this.billingAccountProfileStateService.hasPremiumFromAnySource$(account.id),
-          )
-        : false;
+      const hasPremium =
+        !!account?.id &&
+        (await firstValueFrom(
+          this.billingAccountProfileStateService.hasPremiumFromAnySource$(account.id),
+        ));
 
       const isCardRestricted = (
         await firstValueFrom(this.restrictedItemTypesService.restricted$)
@@ -320,11 +320,11 @@ export class MainContextMenuHandler {
       }
 
       const account = await firstValueFrom(this.accountService.activeAccount$);
-      const canAccessPremium = account?.id
-        ? await firstValueFrom(
-            this.billingAccountProfileStateService.hasPremiumFromAnySource$(account.id),
-          )
-        : false;
+      const canAccessPremium =
+        !!account?.id &&
+        (await firstValueFrom(
+          this.billingAccountProfileStateService.hasPremiumFromAnySource$(account.id),
+        ));
       if (canAccessPremium && (!cipher || !Utils.isNullOrEmpty(cipher.login?.totp))) {
         await createChildItem(COPY_VERIFICATION_CODE_ID);
       }


apps/browser/src/autofill/background/context-menus.background.ts
```diff
@@ -28,18 +28,16 @@ export default class ContextMenusBackground {
...
-              BrowserApi.tabSendMessageData(sender.tab, "closeNotificationBar");
+          const onClickData = msg.data.commandToRetry.message.contextMenuOnClickData;
+          const senderTab = msg.data.commandToRetry.sender.tab;
+
+          if (onClickData && senderTab) {
+            void this.contextMenuClickedHandler.cipherAction(onClickData, senderTab).then(() => {
+              if (sender.tab) {
+                void BrowserApi.tabSendMessageData(sender.tab, "closeNotificationBar");
+              }
             });
```
blackwood, line 39: I can see why `tabSendMessageData` would be considered a side effect that we don't need to consider the return value from, but unsure about the outer Promise here, is there any benefit to handling it's error case?

apps/browser/src/autofill/background/abstractions/notification.background.ts
```diff
@@ -147,7 +147,7 @@ type NotificationBackgroundExtensionMessageHandlers = {
   bgGetEnableChangedPasswordPrompt: () => Promise<boolean>;
   bgGetEnableAddedLoginPrompt: () => Promise<boolean>;
   bgGetExcludedDomains: () => Promise<NeverDomains>;
-  bgGetActiveUserServerConfig: () => Promise<ServerConfig>;
+  bgGetActiveUserServerConfig: () => Promise<ServerConfig | null>;
```
blackwood, line 150: ~~I can't find a corresponding reason this should be a union type with `null`.~~ Ignore this.

apps/browser/src/autofill/background/abstractions/overlay.background.ts
```diff
@@ -1,39 +1,32 @@
...
 import { LockedVaultPendingNotificationsData } from "./notification.background";
 
-export type PageDetailsForTab = Record<
-  chrome.runtime.MessageSender["tab"]["id"],
-  Map<chrome.runtime.MessageSender["frameId"], PageDetail>
->;
+export type TabId = NonNullable<chrome.tabs.Tab["id"]>;
+export type FrameId = NonNullable<chrome.runtime.MessageSender["frameId"]>;
+
+export type PageDetailsForTab = Record<TabId, Map<FrameId, PageDetail>>;
```
blackwood: Very well-composed type!

apps/browser/src/autofill/background/abstractions/overlay.background.ts
```diff
@@ -155,13 +148,30 @@ export type OverlayBackgroundExtensionMessage = {
...
+  | "viewCipher"
+  | "redirectFocus"
+  | "updateHeight"
+  | "buttonClicked"
+  | "blurred"
+  | "updateColorScheme"
+  | "unlockVault"
+  | "refreshGeneratedPassword"
+  | "fillGeneratedPassword";
+
```
blackwood, line 168: I might prefer OverlayPortMessage.command (and subsequent properties) be resolved by a generic, but that can always be done another time. That is assuming these properties are not included with every port message.

apps/browser/src/autofill/background/abstractions/overlay.background.ts
```diff
@@ -272,11 +301,32 @@ export type InlineMenuListPortMessageHandlers = {
...
-  updateAutofillInlineMenuListHeight: ({ message, port }: PortOnMessageHandlerParams) => void;
+  fillAutofillInlineMenuCipher: ({
+    message,
+    port,
+  }: PortOnMessageHandlerParams<OverlayPortMessage & { command: string }>) => void;
+  addNewVaultItem: ({
+    message,
+    port,
+  }: PortOnMessageHandlerParams<
+    OverlayPortMessage & { command: string; addNewCipherType?: CipherType }
```
blackwood: Since these are essentially being treated like arguments, this is where I would use a generic, see for example `NotificationBackgroundExtensionMessage` in  https://github.com/bitwarden/clients/pull/14611/files

apps/browser/src/autofill/content/components/cipher/types.ts
```diff
@@ -22,20 +24,13 @@ export const OrganizationCategories = {
...
-};
-
 type BaseCipherData<CipherTypeValue> = {
   id: string;
   name: string;
   type: CipherTypeValue;
   reprompt: CipherRepromptType;
   favorite: boolean;
-  icon: WebsiteIconData;
+  icon: CipherIconDetails;
```
blackwood, line 33: Removing redundant types ftw!

apps/browser/src/autofill/notification/bar.html
```diff
@@ -1,5 +1,4 @@
-﻿<!-- eslint-disable tailwindcss/no-custom-classname -->
-<!doctype html>
```
jprusik, line 2: Also handled in https://github.com/bitwarden/clients/pull/17212

apps/browser/src/autofill/background/abstractions/overlay.background.ts
jprusik: I had help 😅

apps/browser/src/autofill/background/context-menus.background.ts
jprusik: I wanted the footprint of logical changes to be as small as possible to enable strict typing, so I didn't add handling logic to the errors, since they could potentially have unintended side effects (which I'd like to avoid researching in this scope of this task at hand).

I'm open to further counterpoints on this, I've just erred on the side of keeping the scope smaller.

apps/browser/src/autofill/background/abstractions/overlay.background.ts
jholland-livefront: I find these nested type arguments difficult to read.  Could you create an intermediate type here?  Maybe `export type PageDetailsByFrame = Map<FrameId, PageDetail>;`

apps/browser/src/autofill/background/abstractions/overlay.background.ts
```diff
@@ -1,39 +1,31 @@
...
-  Map<chrome.runtime.MessageSender["frameId"], SubFrameOffsetData>
->;
-
-export type WebsiteIconData = {
-  imageEnabled: boolean;
-  image: string;
-  fallbackImage: string;
-  icon: string;
-};
+export type SubFrameOffsetsForTab = Record<TabId, Map<FrameId, SubFrameOffsetData>>;
```
jholland-livefront: Same suggestion here.  This could be `export type SubFrameOffsetsByFrame = Map<FrameId, SubFrameOffsetData>;`.

apps/browser/src/autofill/background/abstractions/overlay.background.ts
```diff
@@ -201,9 +210,14 @@ export type BuildCipherDataParams = {
 export type BackgroundMessageParam = {
   message: OverlayBackgroundExtensionMessage;
 };
+
 export type BackgroundSenderParam = {
-  sender: chrome.runtime.MessageSender;
+  sender: chrome.runtime.MessageSender & {
+    tab: NonNullable<chrome.runtime.MessageSender["tab"]>;
+    frameId: NonNullable<chrome.runtime.MessageSender["frameId"]>;
```
jholland-livefront: Can this be `frameId: FrameId;`?

apps/browser/src/autofill/services/inline-menu-field-qualification.service.ts
```diff
@@ -374,7 +386,13 @@ export class InlineMenuFieldQualificationService
...
     pageDetails: AutofillPageDetails,
   ): boolean {
-    const parentForm = pageDetails.forms[field.form];
+    let parentForm;
+
+    const fieldForm = field.form;
+
+    if (fieldForm) {
+      parentForm = pageDetails.forms[fieldForm];
+    }
```
jholland-livefront, line 392: Since we are doing this three times maybe we should make a small function for it?
```
  private findParentForm(
    field: AutofillField,
    pageDetails: AutofillPageDetails
  ): AutofillForm | null {
    const fieldForm = field.form;

    return fieldForm ? pageDetails.forms[fieldForm] : null
  }
```

then call `const parentForm = this.findParentForm(field, pageDetails)`

apps/browser/src/autofill/browser/main-context-menu-handler.ts
```diff
@@ -179,9 +177,11 @@ export class MainContextMenuHandler {
...
     try {
       const account = await firstValueFrom(this.accountService.activeAccount$);
-      const hasPremium = await firstValueFrom(
-        this.billingAccountProfileStateService.hasPremiumFromAnySource$(account.id),
-      );
+      const hasPremium = account?.id
+        ? await firstValueFrom(
+            this.billingAccountProfileStateService.hasPremiumFromAnySource$(account.id),
+          )
+        : false;
```
dan-livefront: nit:  we can short circuit here instead of explicitly returning a `false`

```
const hasPremium =
  !!account?.id &&
  await firstValueFrom(
    this.billingAccountProfileStateService.hasPremiumFromAnySource$(account.id)
  );

```

If TS narrows poorly we can do 

```
const id = account?.id;
const hasPremium = !!id && await firstValueFrom(
  this.billingAccountProfileStateService.hasPremiumFromAnySource$(id)
);
```

this is non blocking and will leave up to you

apps/browser/src/autofill/browser/main-context-menu-handler.ts
```diff
@@ -318,9 +320,11 @@ export class MainContextMenuHandler {
...
 
       const account = await firstValueFrom(this.accountService.activeAccount$);
-      const canAccessPremium = await firstValueFrom(
-        this.billingAccountProfileStateService.hasPremiumFromAnySource$(account.id),
-      );
+      const canAccessPremium = account?.id
+        ? await firstValueFrom(
+            this.billingAccountProfileStateService.hasPremiumFromAnySource$(account.id),
+          )
+        : false;
```
dan-livefront: Same here as https://github.com/bitwarden/clients/pull/17087/files#r2495091177

apps/browser/src/autofill/models/autofill-field.ts
dan-livefront, line 1: I'm a bit concerned about switching to non null assertions since it bypasses TS safety checks.
If any of these fields aren’t actually set before use, we could hit runtime undefined issues, are we confident this wont be the case?

apps/browser/src/autofill/services/inline-menu-field-qualification.service.ts
jprusik: Eh - I think I'll leave this one; it's not particularly verbose, and the function call doesn't make things much more concise. Let's revisit on subsequent journeys to this file.

apps/browser/src/autofill/models/autofill-field.ts
jprusik: As discussed; I've added jsdoc signposting for non-null assertions so future development doesn't lose track of them, allowing us to otherwise enable strict ts with logical parity.

*edit:* I've also created a tech debt ticket to track the need to follow up on this in [PM-27881](https://bitwarden.atlassian.net/browse/PM-27881)

dan-livefront (approved):

PR #17093, opened 2025-10-29
[PM-25932] Shorten socket paths

coroiu:
## 📔 Objective

Users are having issues enabling biometrics because they use emails as the username on macos. Unfortunately this causes the path to the socket to become longer, causing it to fail (the limit is like 104 chars). This is a quick-fix to make our contribution to the path as small as possible. It does not fix underlying issue, to do that we should use XPC

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/apps/desktop/src/platform/main/autofill/native-autofill.main.ts
+++ b/apps/desktop/src/platform/main/autofill/native-autofill.main.ts
@@ -35,7 +35,7 @@ export class NativeAutofillMain {
     );
 
     this.ipcServer = await autofill.IpcServer.listen(
-      "autofill",
+      "af",
       // RegistrationCallback
       (error, clientId, sequenceNumber, request) => {
         if (error) {


apps/desktop/desktop_native/core/src/ipc/mod.rs
dani-garcia: I wonder if we could also remove the `/tmp` from the end of the path, and just create the file at the root of the group container.

dani-garcia:
🤔 Also, the unsandboxed DMG app uses a different (shorter?) path, I wonder if that would help some users with too long usernames

coroiu:
@dani-garcia Hmm yeah, it would be recommending a slightly less secure solution thought

apps/desktop/src/main/native-messaging.main.ts
```diff
@@ -78,7 +78,7 @@ export class NativeMessagingMain {
       this.ipcServer.stop();
     }
 
-    this.ipcServer = await ipc.IpcServer.listen("bitwarden", (error, msg) => {
+    this.ipcServer = await ipc.IpcServer.listen("bw", (error, msg) => {
```
dani-garcia, line 81: Ah good eye, we missed this one! We also have the one for autofill that needs updating: https://github.com/bitwarden/clients/blob/e8db35907dc1a077c422b9feae661a8bfcb2068d/apps/desktop/src/platform/main/autofill/native-autofill.main.ts#L37-L39

coroiu: Lol I searched for the wrong thing 🤦 thanks!
image

dani-garcia: Yeah, initially I searched for `("autofill"` knowing it had to be there somewhere and also couldn't find it because the newline hah

PR #17099, opened 2025-10-29
[PM-27591] Remove orgid in vault decryption code

quexten:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-27591

## 📔 Objective

This is part of a set of changes getting rid of `decrypt` on EncString which must be removed to get rid of active user state. This PR specifically forces the use of the passed in key, instead of relying on `decrypt` implementations to dynamically get a key based on userid / orgid.

The now unused parameters will be removed in a subsequent PR.

Vault will eventually drop the typescript implementation here, but until then this updates the code to no longer use orgid / userid as parameters, so that in a later PR we can get rid of it in platform code. All that is passed through is the decryption key.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/libs/common/src/vault/models/domain/cipher.ts
+++ b/libs/common/src/vault/models/domain/cipher.ts
@@ -145,13 +145,7 @@ export class Cipher extends Domain implements Decryptable<CipherView> {
       }
     }
 
-    await this.decryptObj<Cipher, CipherView>(
-      this,
-      model,
-      ["name", "notes"],
-      null,
-      cipherDecryptionKey,
-    );
+    await this.decryptObj<Cipher, CipherView>(this, model, ["name", "notes"], cipherDecryptionKey);
 
     switch (this.type) {
       case CipherType.Login:
--- a/libs/common/src/platform/models/domain/domain-base.ts
+++ b/libs/common/src/platform/models/domain/domain-base.ts
@@ -73,14 +73,13 @@ export default class Domain {
     domain: DomainEncryptableKeys<D>,
     viewModel: ViewEncryptableKeys<V>,
     props: EncryptableKeys<D, V>[],
-    orgId: string | null,
     key: SymmetricCryptoKey | null = null,
     objectContext: string = "No Domain Context",
   ): Promise<V> {
     for (const prop of props) {
       viewModel[prop] =
         (await domain[prop]?.decrypt(
-          orgId,
+          null,
           key,
           `Property: ${prop as string}; ObjectContext: ${objectContext}`,
         )) ?? null;
--- a/libs/common/src/tools/send/models/domain/send-access.ts
+++ b/libs/common/src/tools/send/models/domain/send-access.ts
@@ -54,7 +54,7 @@ export class SendAccess extends Domain {
   async decrypt(key: SymmetricCryptoKey): Promise<SendAccessView> {
     const model = new SendAccessView(this);
 
-    await this.decryptObj<SendAccess, SendAccessView>(this, model, ["name"], null, key);
+    await this.decryptObj<SendAccess, SendAccessView>(this, model, ["name"], key);
 
     switch (this.type) {
       case SendType.File:
--- a/libs/common/src/tools/send/models/domain/send-file.ts
+++ b/libs/common/src/tools/send/models/domain/send-file.ts
@@ -38,7 +38,6 @@ export class SendFile extends Domain {
       this,
       new SendFileView(this),
       ["fileName"],
-      null,
       key,
     );
   }
--- a/libs/common/src/tools/send/models/domain/send-text.ts
+++ b/libs/common/src/tools/send/models/domain/send-text.ts
@@ -30,13 +30,7 @@ export class SendText extends Domain {
   }
 
   decrypt(key: SymmetricCryptoKey): Promise<SendTextView> {
-    return this.decryptObj<SendText, SendTextView>(
-      this,
-      new SendTextView(this),
-      ["text"],
-      null,
-      key,
-    );
+    return this.decryptObj<SendText, SendTextView>(this, new SendTextView(this), ["text"], key);
   }
 
   static fromJSON(obj: Jsonify<SendText>) {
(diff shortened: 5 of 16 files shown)


libs/importer/src/importers/bitwarden/bitwarden-json-importer.ts
```diff
@@ -166,19 +167,22 @@ export class BitwardenJsonImporter extends BaseImporter implements Importer {
 
   private async parseFolders(
     data: BitwardenUnEncryptedIndividualJsonExport | BitwardenEncryptedIndividualJsonExport,
+    userId: UserId,
```
djsmith85, line 170: 🎨  Instead of providing the `userId` to grab the userKey, you could add an optional key and provide `keyForDecryption` from `parseEncrypted()`. Optional only because it is not required when `parseFolders()` is called from `parseDecrypted()`. 

🌱  While reviewing these changes, I've seen some really bad patterns and type usages. I'm experimenting with a branch where I introduce a `BitwardenEncryptedJsonImporter` and use type-narrowing instead of just testing for `data.encrypted`

djsmith85 (changes requested):
❓ I don't see any changes related to send models. Are the changes on the send-specs necessary

quexten:
> ❓ I don't see any changes related to send models. Are the changes on the send-specs necessary

@djsmith85 Yes, one of the follow-up PR's that changes `EncString` would otherwise break the vault tests. If you prefer I can move the spec changes to there, but that just means one extra vault review.

libs/importer/src/importers/bitwarden/bitwarden-json-importer.ts
quexten: Passing around keys is generally not preferred. In the medium term KM wants to / needs to move keys to live within SDK only.

This implies other teams moving their code to the SDK. If a userId is passed through, then replacing it would just be:
1. Grab correct SDK instance for the userid
2. Call "decryptExport" (or whatever it is called) to decrypt the export with the SDK

So in this pattern, userKey would never be a parameter offered in TS, and using the userId makes a future move easier. So in this case I don't think going with the key is the right move?

That said, maybe the replacement happens at an even higher level in which case this does not matter. However, e.g. the call to cipherService also does not pass a key so it seemingly would also break pattern within the importer.

libs/common/src/vault/models/domain/cipher.spec.ts
```diff
@@ -317,19 +327,11 @@ describe("Cipher DTO", () => {
...
       encryptService.unwrapSymmetricKey.mockResolvedValue(
         new SymmetricCryptoKey(makeStaticByteArray(64)),
       );
 
-      (window as any).bitwardenContainerService = new ContainerService(keyService, encryptService);
-
-      const cipherView = await cipher.decrypt(
-        await cipherService.getKeyForCipherKeyDecryption(cipher, mockUserId),
-      );
+      const cipherView = await cipher.decrypt(mockSymmetricKey);
```
nick-livefront, line 334: This is only one instance, much less boilerplate in tests is awesome! 👏

libs/common/src/vault/models/domain/cipher.ts
```diff
@@ -123,16 +124,22 @@ export class Cipher extends Domain implements Decryptable<CipherView> {
   // We are passing the organizationId into the EncString.decrypt() method here, but because the encKey will always be
   // present and so the organizationId will not be used.
   // We will refactor the EncString.decrypt() in https://bitwarden.atlassian.net/browse/PM-3762 to remove the dependency on the organizationId.
```
nick-livefront: 🎨 Can you update or remove this comment respective of your changes here? The linked ticket was marked as abandoned.

libs/common/src/vault/models/domain/cipher.ts
```diff
@@ -123,16 +124,22 @@ export class Cipher extends Domain implements Decryptable<CipherView> {
...
-  async decrypt(encKey: SymmetricCryptoKey): Promise<CipherView> {
+  async decrypt(userKeyOrOrgKey: SymmetricCryptoKey): Promise<CipherView> {
+    assertNonNullish(userKeyOrOrgKey, "userKeyOrOrgKey", "Cipher decryption");
```
nick-livefront, line 125: ❓ Should this be moved within the `try`/`catch`? My instinct says yes so when an error is thrown the cipher will follow our cannot decrypt process / UI.

quexten: The try catch block bellow only applies to ciphers (vault items) that have cipher keys (per-item keys) enabled. For items that do not have it enabled, the try-catch block is skipped entirely.

I'm not a domain expert here. That said:
- decrypt(..) should never take null as a parameter. If it does, then the typescript type should be changed to indicate that. The only purpose of the assert here is so that in case somehow a null does get accidentally passed into here, it does not appear in some lower level domain object, but already is pin-pointed to the highest level within ciphers, which should allow easier debugging

However, if from vault's side, it is unknown whether this parameter can ever be null / undefined, then we could instead add the following check / replace the assert with it:
```
// This should never happen. However, we want to make sure to return early here to not bubble-up exceptions in lower-level code that does not check for null values.
if (userKeyOrOrgKey == null) {
    model.name = "[error: cannot decrypt]";
    model.decryptionFailure = true;
    return model;
}
```
what do you think?

libs/common/src/vault/models/domain/cipher.ts
gbubemismith: ❓ should we also update the `decryptObj` method and remove the `orgId` parameter since we always pass null

libs/common/src/vault/models/domain/cipher.ts
nick-livefront: I appreciate the explanation! I think we can stick with what you have currently as I agree that the `key` should never be `null`/`undefined`. 

Everywhere I am seeing from an implementation standpoint is passing a key to the method, we are not expecting a null value to be passed from our end. 👍

nick-livefront (approved):

libs/common/src/vault/models/domain/cipher.ts
quexten: Initially it was split this wait to try to have the changes limited to vault, but that failed and we had to drag in tools / platform anyways, so you're right. It can be removed. https://github.com/bitwarden/clients/pull/17099/commits/cfe2bc8c3e40530061742b7a65b42456c1b75d29

dereknance (approved):

djsmith85 (approved):

PR #17100, opened 2025-10-29
[PM-27592] Remove deprecated encstring usage from dirt code

quexten:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-27592

## 📔 Objective

Removes usage of the deprecated `decrypt` method on EncString. This method needs to be removed to get rid of activeuserstate usage in KM's owned code.

## ⏰ Reminders before review

- Contributor guidelines followed
- All formatters and local linters executed and passed
- Written new unit and / or integration tests where applicable
- Protected functional changes with optionality (feature flags)
- Used internationalization (i18n) for all UI strings
- CI builds passed
- Communicated to DevOps any deployment requirements
- Updated any necessary documentation (Confluence, contributing docs) or informed the documentation team

## 🦮 Reviewer guidelines

- 👍 (`:+1:`) or similar for great changes
- 📝 (`:memo:`) or ℹ️ (`:information_source:`) for notes or general info
- ❓ (`:question:`) for questions
- 🤔 (`:thinking:`) or 💭 (`:thought_balloon:`) for more open inquiry that's not quite a confirmed issue and could potentially benefit from discussion
- 🎨 (`:art:`) for suggestions / improvements
- ❌ (`:x:`) or ⚠️ (`:warning:`) for more significant problems or concerns needing attention
- 🌱 (`:seedling:`) or ♻️ (`:recycle:`) for future improvements or indications of technical debt
- ⛏ (`:pick:`) for minor or nitpick changes

PR #17103, opened 2025-10-29
PM-23810 [error: cannot decrypt] shows in Member Access Report

prograhamming:
## 🎟️ Tracking

[PM-23810](https://bitwarden.atlassian.net/jira/software/c/projects/BW/boards/222?assignee=712020%3Aa479bb6c-74c0-4e0d-bd65-bc64b95c604a&selectedIssue=PM-23810)

## 📔 Objective

Fixes a user experience with the member access report csv export. CSV export shows a '[error: cannot decrypt]' value under the 'Collection' column when a user is not a part of any collections or groups in a organization.

## 📸 Screenshots

Before:
image

After:
image

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

[PM-23810]: https://bitwarden.atlassian.net/browse/PM-23810?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/bitwarden_license/bit-web/src/app/dirt/reports/member-access-report/services/member-access-report.service.ts
+++ b/bitwarden_license/bit-web/src/app/dirt/reports/member-access-report/services/member-access-report.service.ts
@@ -95,7 +95,7 @@ export class MemberAccessReportService {
     const collectionNames = memberAccessReports.map((item) => item.collectionName.encryptedString);
 
     const collectionNameMap = new Map(
-      collectionNames.filter((col) => col != null).map((col) => [col, ""]),
+      collectionNames.filter((col) => col !== null).map((col) => [col, ""]),
     );
     for await (const key of collectionNameMap.keys()) {
       const encryptedCollectionName = new EncString(key);
--- a/bitwarden_license/bit-web/src/app/dirt/reports/member-access-report/services/member-access-report.service.spec.ts
+++ b/bitwarden_license/bit-web/src/app/dirt/reports/member-access-report/services/member-access-report.service.spec.ts
@@ -2,7 +2,6 @@ import { mock } from "jest-mock-extended";
 import { of } from "rxjs";
 
 import { EncryptService } from "@bitwarden/common/key-management/crypto/abstractions/encrypt.service";
-import { EncString } from "@bitwarden/common/key-management/crypto/models/enc-string";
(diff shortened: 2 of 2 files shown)


bitwarden_license/bit-web/src/app/dirt/reports/member-access-report/services/member-access-report.service.ts
```diff
@@ -85,7 +85,10 @@ export class MemberAccessReportService {
     for await (const key of collectionNameMap.keys()) {
       const decrypted = new EncString(key);
       await decrypted.decrypt(organizationId);
```
quexten: @prograhamming Please don't rely on the `[error: cannot decrypt]` specific error return value magic string. This is not a stable API contract that KM guarantees and may break in the future.

I happen to just have raised a PR in this file to move us off of the deprecated `Encstring.decrypt` method here: https://github.com/bitwarden/clients/pull/17100. With the above PR the decrypt would actually throw when decryption fails.

So the two supported options from KM's perspective are:
- Use the encryptService.decryptString API and catch the decryption error
- By application logic, decide that the decryptedValue cannot be decrypted (I'm not a domain expert here, what even is the "key" that we are decrypting for this user? What is it encrypted by / why does it fail to decrypt? is it just null?)

prograhamming: Okay I can make that adjustment. It fails to decrypt because it is trying to decrypt the collectionName which is null from the backend because the user in the organization has not been added to a collection or group within the organization vault.

quexten: In that case I would just recommend filtering out by the null values, and not even calling decrypt for those cases

prograhamming: Now that I think about this there is a better way to handle the null value. I will accept your new changes and change my implementation for this user story.

quexten (approved):
Nice, thank you for addressing my feedback!

PR #16613, opened 2025-09-26
[PM-22758] Configurable Keyboard Shortcut for Autotype

coltonhurst:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-22785

## 📔 Objective

Add the ability to customize the keyboard shortcut for Autotype.

## 📸 Screenshots

Settings Menu:

Configurable Keyboard Shortcut Modal:

## ⏰ Reminders before review

- Contributor guidelines followed
- All formatters and local linters executed and passed
- Written new unit and / or integration tests where applicable
- Protected functional changes with optionality (feature flags)
- Used internationalization (i18n) for all UI strings
- CI builds passed
- Communicated to DevOps any deployment requirements
- Updated any necessary documentation (Confluence, contributing docs) or informed the documentation team

## 🦮 Reviewer guidelines

- 👍 (`:+1:`) or similar for great changes
- 📝 (`:memo:`) or ℹ️ (`:information_source:`) for notes or general info
- ❓ (`:question:`) for questions
- 🤔 (`:thinking:`) or 💭 (`:thought_balloon:`) for more open inquiry that's not quite a confirmed issue and could potentially benefit from discussion
- 🎨 (`:art:`) for suggestions / improvements
- ❌ (`:x:`) or ⚠️ (`:warning:`) for more significant problems or concerns needing attention
- 🌱 (`:seedling:`) or ♻️ (`:recycle:`) for future improvements or indications of technical debt
- ⛏ (`:pick:`) for minor or nitpick changes

PR #17105, opened 2025-10-29
Desktop Autotype use managed object for IPC vault data payload

neuronull:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-26226

## 📔 Objective

In https://github.com/bitwarden/clients/pull/16613#discussion_r2383410841 there was a nice observation that we could use a managed object for the vault data to future proof this interface a bit.

Note that the completion criteria for the task suggests to update the other IPC payloads. Since https://github.com/bitwarden/clients/pull/17062 contains significant change to the `configureAutotype` IPC, I will do the managed object change for that one in that PR.

Manual validations:
- autotype happy path

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


apps/desktop/src/autofill/models/autotype-vault-data.ts
```diff
@@ -0,0 +1,4 @@
+export interface AutotypeVaultData {
+  username: string;
+  password: string;
+}
```
neuronull, line 8: LMK if there is a more appropriate place for this

neuronull:
Responding to the claude review:

> 1. ⚠️ Security: Vault Data in IPC Without Encryption Context

We have a separate user story to track using the encrypted IPC channel already, ignoring.

> 2. ❌ Missing Test Coverage (0% patch coverage)

Adding unit tests is a good idea. I haven't seen us doing that in these service files for Desktop yet but I'll try to find examples and add some.

> 3. ⚠️ Type Safety: Optional Properties Not Handled Consistently

Right, this was intentional to make them not optional. I will add a doc comment.

> 4. 🎨 Error Message Could Be More Descriptive

Fair

> 5. 🎨 Consider Extracting Magic String

Fair

> 6. 🎨 Inconsistent Error Handling Pattern

Garbage

neuronull:
I'll wait to add unit tests until reviewers have had an initial look

apps/desktop/src/autofill/models/autotype-vault-data.ts
jprusik: This makes sense to me - as other concerns consume this model, promote it upward in the file hierarchy to better represent shared concerns. If it ends up being consumed outside of the desktop client, it would get moved to a shared/common resource like `libs/common/src/autofill/types`

apps/desktop/src/autofill/services/desktop-autotype.service.ts
```diff
@@ -64,10 +66,20 @@ export class DesktopAutotypeService {
       const possibleCiphers = await this.matchCiphersToWindowTitle(windowTitle);
       const firstCipher = possibleCiphers?.at(0);
 
-      return callback(null, {
-        username: firstCipher?.login?.username,
-        password: firstCipher?.login?.password,
-      });
+      if (firstCipher === undefined) {
```
jprusik: nit; check for falsey values generally (null, undefined, etc)?

```suggestion
      if (!firstCipher) {
```

If the typing is consistent, that shouldn't be needed, however.

apps/desktop/src/autofill/services/desktop-autotype.service.ts
```diff
@@ -64,10 +66,20 @@ export class DesktopAutotypeService {
...
+        return callback(Error("No matching vault item."), null);
+      } else if (
+        firstCipher.login.username === undefined ||
```
jprusik: Can the `login` property be undefined at this point (e.g. identity cipher)?

neuronull: From what I can tell, no it can't. I'm trusting my LSP and also checking the `CipherView`

apps/desktop/src/autofill/services/desktop-autotype.service.ts
neuronull: addressed in https://github.com/bitwarden/clients/pull/17105/commits/4b9fa4431f2f81412dd1692eae0d106dd806d160

apps/desktop/src/autofill/services/desktop-autotype.service.ts
neuronull: Claude just flagged this in it's latest review. In `matchCiphersToWindowTitle()` we check if `login` is undefined. I don't know enough about typescript to understand the intricacies here but from my look at `CipherView`, `login` it is constructed as an instance of the type, and the declared type is not `<type | undefined>` , so my naive take is that we don't need to check for undefined. If that is incorrect, definitely let me know and if that is the case I will update the code and additionally have existential concerns about all the TS code I write 😂

jprusik: > Can the login property be undefined at this point (e.g. identity cipher)?

It cannot; here's an example of a mock Payment cipher:
Screenshot 2025-10-31 at 10 50 22 AM

apps/desktop/src/autofill/services/desktop-autotype.service.ts
jprusik, line 1: Should this file be abstracted to Ruby?

neuronull: Great question!

neuronull:
This has been validated by QA

jprusik (approved):

apps/desktop/src/autofill/services/desktop-autotype.service.ts
jprusik: (Humans can disregard this thread; it was a Claude experiment)

PR #17108, opened 2025-10-29
[PM-24102] Remove encstring decrypt function

quexten:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-24102

## 📔 Objective

Removes the decrypt function from encstring and moves the single caller off of it.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/apps/cli/src/tools/send/commands/receive.command.spec.ts
+++ b/apps/cli/src/tools/send/commands/receive.command.spec.ts
@@ -12,6 +12,7 @@ import { ErrorResponse } from "@bitwarden/common/models/response/error.response"
 import { ConfigService } from "@bitwarden/common/platform/abstractions/config/config.service";
 import { EnvironmentService } from "@bitwarden/common/platform/abstractions/environment.service";
 import { PlatformUtilsService } from "@bitwarden/common/platform/abstractions/platform-utils.service";
+import { SendAccess } from "@bitwarden/common/tools/send/models/domain/send-access";
 import { SendAccessResponse } from "@bitwarden/common/tools/send/models/response/send-access.response";
 import { SendApiService } from "@bitwarden/common/tools/send/services/send-api.service.abstraction";
 import { SendType } from "@bitwarden/common/tools/send/types/send-type";
@@ -400,7 +401,8 @@ describe("SendReceiveCommand", () => {
           },
         };
 
-        sendApiService.postSendAccessV2.mockResolvedValue(mockSendResponse as any);
+        sendApiService.postSendAccessV2.mockResolvedValue({} as any);
+        jest.spyOn(SendAccess.prototype, "decrypt").mockResolvedValueOnce(mockSendResponse as any);
         sendApiService.getSendFileDownloadDataV2.mockResolvedValue({
           url: "https://example.com/download",
         } as any);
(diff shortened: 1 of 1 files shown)


libs/common/src/platform/models/domain/domain-base.ts
```diff
@@ -69,22 +70,29 @@ export default class Domain {
...
+    const encryptService = Utils.getContainerService().getEncryptService();
     for (const prop of props) {
-      viewModel[prop] =
-        (await domain[prop]?.decrypt(
-          null,
-          key,
-          `Property: ${prop as string}; ObjectContext: ${objectContext}`,
-        )) ?? null;
+      if (domain[prop] == null) {
+        continue;
```
mzieniukbw, line 85: ❓  Wanting to avoid any regressions, if `domain[prop]` is equal to `undefined`, then the old code would override the field with value `null`, but here we leave it at `undefined`.

libs/common/src/platform/models/domain/domain-base.ts
```diff
@@ -69,22 +70,29 @@ export default class Domain {
...
+      }
+      try {
+        viewModel[prop] = await encryptService.decryptString(domain[prop]!, key!);
```
mzieniukbw: ❓ Old code accepted `key` as null, pulling the user key or org key as fallback and throwing error when `key` was still null. Did we check all the callers, that key can never be null ?

libs/common/src/platform/models/domain/domain-base.ts
```diff
@@ -69,22 +70,29 @@ export default class Domain {
...
+      } catch (e) {
+        throw new Error(
```
mzieniukbw: ❓ Does the callers accepts this and can handle this ? Before we just set it to `DECRYPT_ERROR`.

libs/common/src/platform/models/domain/domain-base.ts
quexten: Makes sense. Adding `viewModel[prop] = null`

libs/common/src/platform/models/domain/domain-base.ts
quexten: Yeah, the passed in key should never be null, it seems I forgot to adjust the function signature here, but all callers do pass in a key.

libs/common/src/platform/models/domain/domain-base.ts
quexten: Makes sense, made this backward compatible by not throwing but logging + setting the decrypt error.

libs/common/src/platform/models/domain/domain-base.ts
```diff
@@ -69,22 +70,32 @@ export default class Domain {
...
+        continue;
+      }
+      try {
+        viewModel[prop] = await encryptService.decryptString(domain[prop]!, key);
+      } catch (e) {
+        // eslint-disable-next-line no-console
+        console.error(
+          `Failed to decrypt property '${String(
+            prop,
+          )}' of domain. Context: ${objectContext}. Error: ${(e as Error).message}`,
```
mzieniukbw: ❓  Depending on how this error is mapped in Rust via `bitwarden_error`, as i understand, if "full" is used, the thrown error will not have the `message` field. I think for this specific case it's fine, but i want to be cautious about this, so we don't end up with another error in the future, because `message` field is not there

quexten: I don't quite understand in which case this would happen, but just in case I added:
```

        // In case the SDK maps to a non-Error type, this is defensive
        const errorMsg =
          typeof e === "object" && e !== null && "message" in e
            ? (e as { message: string }).message
            : String(e);

```

quexten:
Looks like the failing tests are from `main` failing.

addisonbeck (approved):

quexten:
Looks like there was a prettier change that did not go discovered because of the CI previously failing...

mzieniukbw (approved):

PR #17112, opened 2025-10-29
[CL-806] Focus main content after SPA navigation occurs

vleague2:
## 🎟️ Tracking

[CL-806](https://bitwarden.atlassian.net/browse/CL-806)

## 📔 Objective

This PR adds focus management to our implementation of the Angular router in the web app only. The goal is to ensure screenreader users know a navigation has occurred (single-page apps don't cause a page reload, which is how normally you would know a navigation has occurred as a SR user). 

We opted to accomplish this by moving their focus to the newly loaded page content (in the `main` element) since it is a fairly safe assumption that if a user navigates to a new "page," they want to go to that page. This PR does include mechanisms for opting out of this behavior in the few cases that need to do so.

The behavior is behind a feature flag so that we can easily switch it off if we encounter issues in prod. This touches a large surface area of the web app, so we want the option for a quick off switch.

## 📸 Screenshots

Before:

https://github.com/user-attachments/assets/abdf996c-6ad9-4f49-8bb4-8b9b76c3c37b

After:

https://github.com/user-attachments/assets/7e61efdf-abd0-4793-8326-003982b2009a

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

[CL-806]: https://bitwarden.atlassian.net/browse/CL-806?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/components/src/navigation/nav-item.component.ts
```diff
@@ -25,6 +25,22 @@ export class NavItemComponent extends NavBaseComponent {
...
+   * ended and that the `main` content has updated.
+   *
+   * Defaults to `true`, assuming nav-item is navigating to a new SPA page that would
+   * otherwise not have any accessible indication of navigation completing.
+   *
+   * Examples of when you might want this to be false:
+   * - Navigating to a fully new app that already has, or should behave as, a full
+   * page load (like a non-SPA page would behave). (The product switcher matches this
+   * use case.)
+   * - Navigating to an external page that opens in a new tab.
```
BryanCunningham: @vleague2 Wouldn't navigating to an external page just behave like a normal page navigation?

vleague2: Yeah, I kind of went back and forth on whether this was useful to have the option to opt-out of. If we leave it as `true` always, the focus behavior just doesn't happen when the link is external or behaves like a non-SPA app, so it's functionally the same.

apps/web/src/app/app.component.ts
```diff
@@ -81,6 +82,71 @@ export class AppComponent implements OnDestroy, OnInit {
...
+            navEvent,
+            currentNavInfo: info,
+          };
+        }),
+        filter((navEventAndData) => {
+          const info = navEventAndData.currentNavInfo;
+
+          return info === undefined ? true : info?.focusMainAfterNav !== false;
+        }),
+      )
```
willmartian: Why return `navEvent` and `info` if we are only using `info`?

apps/web/src/app/app.component.ts
```diff
@@ -81,6 +82,71 @@ export class AppComponent implements OnDestroy, OnInit {
...
+            if (this.isFirstPageLoad) {
+              this.isFirstPageLoad = false;
+              return false;
+            }
+
+            return true;
+          }
+
+          return false;
+        }),
```
willmartian: We can call `skip(1)` from rxjs instead of using the class property `isFirstPageLoad`

apps/web/src/app/app.component.ts
willmartian, line 1: We should move this logic into a service in `libs/components/a11y`

willmartian:
Thoughts on feature flagging this?

apps/web/src/app/app.component.ts
```diff
@@ -76,11 +76,17 @@ export class AppComponent implements OnDestroy, OnInit {
     private readonly destroy: DestroyRef,
     private readonly documentLangSetter: DocumentLangSetter,
     private readonly tokenService: TokenService,
+    private readonly routerFocusManager: RouterFocusManagerService,
   ) {
     this.deviceTrustToastService.setupListeners$.pipe(takeUntilDestroyed()).subscribe();
 
     const langSubscription = this.documentLangSetter.start();
-    this.destroy.onDestroy(() => langSubscription.unsubscribe());
+    const routerFocusManagerSubscription = this.routerFocusManager.start();
```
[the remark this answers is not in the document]
vleague2: Yes, will be handled in a follow-up ticket that is linked in Jira (as well as browser)

libs/components/src/a11y/router-focus-manager.service.ts
```diff
@@ -0,0 +1,67 @@
...
+    return this.router.events
+      .pipe(
+        takeUntilDestroyed(),
+        combineLatestWith(this.configService.getFeatureFlag$(FeatureFlag.RouterFocusManagement)),
+        filter(([navEvent, flagEnabled]) => flagEnabled && navEvent instanceof NavigationEnd),
+        /**
+         * On first page load, we do not want to skip the user over the navigation content,
+         * so we opt out of the default focus management behavior.
+         */
+        skip(1),
```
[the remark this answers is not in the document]
vleague2: This needs to be after the check for `navEvent instanceof NavigationEnd` so that we skip the appropriate event (many event types get fired as part of a navigation process).

willmartian (approved):

apps/web/src/app/app.component.ts
```diff
@@ -76,11 +76,17 @@ export class AppComponent implements OnDestroy, OnInit {
...
+
+    this.destroy.onDestroy(() => {
+      langSubscription.unsubscribe();
+      routerFocusManagerSubscription.unsubscribe();
+    });
```
willmartian, line 89: 🎨 If the service just exposed its observable, instead of a method that subscribes internally, it allows for easier cleanup

```ts
this.routerFocusManager.init$.pipe(takeUntil(this.destroy)).subscribe();
```

BryanCunningham (approved):

libs/components/src/a11y/router-focus-manager.service.ts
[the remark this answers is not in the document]
vleague2: Ended up moving things around a bit to avoid calling the flag on every single router nav event: [42cbbfc](https://github.com/bitwarden/clients/pull/17112/commits/42cbbfc12eaf015805ec002bea31ca263c00eec6)

PR #17136, opened 2025-10-30
[CL-736] migrate chip select to use signals

BryanCunningham:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/CL-736

## 📔 Objective

Update chip-select component to use signal based inputs

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/apps/web/src/app/billing/individual/user-subscription.component.html
+++ b/apps/web/src/app/billing/individual/user-subscription.component.html
@@ -37,41 +37,63 @@
     <dd *ngIf="sub.expiration">{{ sub.expiration | date: "mediumDate" }}</dd>
     <dd *ngIf="!sub.expiration">{{ "neverExpires" | i18n }}</dd>
   </dl>
-  <div class="tw-flex tw-w-full" *ngIf="!selfHosted">
-    <div class="tw-w-1/3">
-      <dl>
-        <dt>{{ "status" | i18n }}</dt>
-        <dd>
+  <div class="tw-flex tw-max-w-[1340px] tw-pt-6" *ngIf="!selfHosted">
+    <div class="tw-flex tw-gap-16 tw-justify-between tw-w-full">
+      <div class="tw-flex tw-flex-col">
+        <div class="tw-font-semibold tw-mb-2">{{ "plan" | i18n }}</div>
+        <div>{{ "premiumMembership" | i18n }}</div>
+      </div>
+      <div class="tw-flex tw-flex-col">
+        <div class="tw-font-semibold tw-mb-2">{{ "status" | i18n }}</div>
+        <div>
           <span class="tw-capitalize">{{ (subscription && subscriptionStatus) || "-" }}</span>
-          <span bitBadge variant="warning" *ngIf="subscriptionMarkedForCancel">{{
-            "pendingCancellation" | i18n
-          }}</span>
-        </dd>
-        <dt>{{ "nextCharge" | i18n }}</dt>
-        <dd>
-          {{
-            nextInvoice
-              ? (sub.subscription.periodEndDate | date: "mediumDate") +
-                ", " +
-                (nextInvoice.amount | currency: "$")
-              : "-"
-          }}
-        </dd>
-      </dl>
-    </div>
-    <div class="tw-w-2/3" *ngIf="subscription">
-      <strong class="!tw-block tw-mb-1">{{ "details" | i18n }}</strong>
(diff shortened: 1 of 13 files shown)


libs/components/src/chip-select/chip-select.component.spec.ts
```diff
@@ -0,0 +1,627 @@
...
+
+  afterEach(() => {
+    const overlayContainer = document.querySelector(".cdk-overlay-container");
+    if (overlayContainer) {
+      overlayContainer.innerHTML = "";
+    }
+  });
+
+  describe("Signal Inputs", () => {
+    it("should initialize with required signal inputs", () => {
```
vleague2: Some of these tests feel unnecessary (I know Claude wrote them so it's a wide net). I don't think we need to be testing that Angular works, for example. I would prefer to keep spec files focused on what the end user behavior should be, so that we can be protected from regressions while not being too tightly coupled to internal code implementation. Imo if the test is going to have to be updated if we make a refactor to the component, it's probably more effort to upkeep it than the value it provides.

libs/components/src/chip-select/chip-select.component.ts
```diff
@@ -257,6 +255,7 @@ export class ChipSelectComponent<T = unknown> implements ControlValueAccessor, A
   writeValue(obj: T): void {
     this.selectedOption = this.findOption(this.rootTree, obj);
     this.setOrResetRenderedOptions();
+    this.cdr.markForCheck();
```
vleague2, line 262: Why do we need this now?

libs/components/src/chip-select/chip-select.component.spec.ts
BryanCunningham: Removed some of these specs. Good call

libs/components/src/chip-select/chip-select.component.ts
BryanCunningham: Added a comment about why this is necessary now

libs/components/src/chip-select/chip-select.component.spec.ts
```diff
@@ -0,0 +1,584 @@
...
+    });
+  });
+
+  describe("Disabled State Behavior", () => {
+    it("should disable chip button when disabled input is true", () => {
+      const testApp = fixture.componentInstance;
+      testApp.disabled.set(true);
+      fixture.detectChanges();
+
+      expect(getChipButton().disabled).toBe(true);
```
vleague2: I think this is already covered in a section above?

libs/components/src/chip-select/chip-select.component.spec.ts
```diff
@@ -0,0 +1,584 @@
...
+    });
+
+    it("should apply disabled styling", () => {
+      const testApp = fixture.componentInstance;
+      testApp.disabled.set(true);
+      fixture.detectChanges();
+
+      const chipContainer = fixture.debugElement.query(By.css(".tw-inline-flex"));
+      expect(chipContainer.nativeElement.className).toContain("tw-bg-secondary-300");
+      expect(chipContainer.nativeElement.className).toContain("tw-text-muted");
```
vleague2: I don't love testing for css classes. It's going to be fragile when we make component style updates. There are a decent amount being checked amongst this test file.

libs/components/src/chip-select/chip-select.component.spec.ts
```diff
@@ -0,0 +1,584 @@
...
+    if (currentValue && !chipComponent["selectedOption"]) {
+      chipComponent.writeValue(currentValue);
+    }
+    formFixture.detectChanges();
+  });
+
+  afterEach(() => {
+    const overlayContainer = document.querySelector(".cdk-overlay-container");
+    if (overlayContainer) {
+      overlayContainer.innerHTML = "";
```
vleague2: These are odd -- do we actually need to be doing this?

libs/components/src/chip-select/chip-select.component.spec.ts
BryanCunningham: Agreed. Will rip this out

libs/components/src/chip-select/chip-select.component.spec.ts
BryanCunningham: Hmm, not sure what this is for. Will investigate and remove

PR #17140, opened 2025-10-30
[PM-27648] [Defect] The .zip option is displayed on organizational export in Admin Console

harr1424:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-27648?atlOrigin=eyJpIjoiZDdjMzBkZjNiZDI0NGUyNjlmZWIxZjBlNTQ0ODlkNGYiLCJwIjoiaiJ9

## 📔 Objective

• Prevent the Admin Console from displaying a .zip file export option for an organizational import 
• Also fixes a related issue where the Admin Console would display a callout describing an individual vault export 
 
## 📸 Screenshots

image

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/libs/tools/export/vault-export/vault-export-ui/src/components/export.component.ts
+++ b/libs/tools/export/vault-export/vault-export-ui/src/components/export.component.ts
@@ -356,7 +356,7 @@ export class ExportComponent implements OnInit, OnDestroy, AfterViewInit {
         // Admin Console: organizationId is already set via @Input, no update needed
       });
 
-    // Step 2: Set up dynamic format options based on the organizationId observable
+    // Set up dynamic format options based on the organizationId observable
     // This is the single source of truth for both export contexts
     this.formatOptions$ = this._organizationId$.pipe(
       map((organizationId) => {


libs/tools/export/vault-export/vault-export-ui/src/components/export.component.ts
```diff
@@ -346,9 +346,20 @@ export class ExportComponent implements OnInit, OnDestroy, AfterViewInit {
     this.formatOptions$ = this.exportForm.controls.vaultSelector.valueChanges.pipe(
```
djsmith85: Instead of relying on the vaultSelector changes, we really only care if an organization is selected or is undefined. ❓ Could we instead rely on `this._organizationId$` if it's turned into an `Observable` instead of a `BehaviourSubject`. We would still need the `vaultSelector`.valuechanges to trigger,  but that would only set/update `this._organizationId$`. It turns it into 2 steps but the separation ensure we only listen to a combined Observable instead of handling both cases.

🌱 This is only a suggestion for future improvement and not required now. We are also aiming to replace the @Input with signals, so we could add a compute that reacts to input signal and vaultSelector.valuechange to update this._organizationId$.

harr1424:
### A note regarding Claude's review 

1. [CRITICAL] Fix the memory leak in the @Input setter by managing subscription lifecycle
There is a 1:1 relationship between this subscription and component loading, the component does not re-assign an Organization ID and so only one subscription is possible here 

2. [HIGH] Remove the redundant immediate emission in the setter to prevent race conditions
I need to look into this more closely. While Claude's code snippet looks neat, this will add another property used to track state (we already have two)  and this issue will likely be avoided when we migrate fully to signals anyways 

3. [MEDIUM] Add unit tests for the new reactive logic
This is a valid point

4. [LOW] Consider refactoring to follow ADR-0003 Observable pattern more strictly 
Calude recognizes this can be deferred to signals migration

5. [LOW] Optimize the context detection logic to compute once instead of repeatedly 
This seems very relevant and will be pushed in a commit soon 

6. [OPTIONAL] Fix the startWith initial value to reflect actual form state
Related to 5 

7. [OPTIONAL] Simplify duplicate context checks 
Also related to 5

libs/tools/export/vault-export/vault-export-ui/src/components/export.component.ts
```diff
@@ -265,6 +272,36 @@ export class ExportComponent implements OnInit, OnDestroy, AfterViewInit {
...
+    const exportingFromAdminConsole: boolean = this.isAdminConsoleContext;
+    // Create organizationId$ as a combined observable from @Input and form changes
+    const organizationIdFromForm$ = this.exportForm.controls.vaultSelector.valueChanges.pipe(
+      map((vaultSelection) => {
+        if (exportingFromAdminConsole) {
+          // In Admin Console, ignore form changes - use input value
+          return undefined;
+        }
+        // In Password Manager, map form selection to organizationId
+        const isMyVault = vaultSelection === "myVault";
```
itsadrago: Could this be a problem for languages other than English?

harr1424: Here the value is not actually exposed to the UI

https://github.com/bitwarden/clients/blob/ab7509d8651094292fd6b583769d8aab48ae7f11/libs/tools/export/vault-export/vault-export-ui/src/components/export.component.ts#L232

https://github.com/bitwarden/clients/blob/ab7509d8651094292fd6b583769d8aab48ae7f11/libs/tools/export/vault-export/vault-export-ui/src/components/export.component.ts#L540

Instead, it gets piped through `i18n` (our internationalization service): 
https://github.com/bitwarden/clients/blob/ab7509d8651094292fd6b583769d8aab48ae7f11/libs/tools/export/vault-export/vault-export-ui/src/components/export.component.html#L20

Which will read the value from a specific `messages.json` file for the user's language: 
https://github.com/bitwarden/clients/blob/ab7509d8651094292fd6b583769d8aab48ae7f11/apps/web/src/locales/en/messages.json#L1106

https://github.com/bitwarden/clients/blob/ab7509d8651094292fd6b583769d8aab48ae7f11/apps/web/src/locales/es/messages.json#L1106

itsadrago: Ah got it. Thanks for pointing this out

libs/tools/export/vault-export/vault-export-ui/src/components/export.component.ts
```diff
@@ -265,6 +272,36 @@ export class ExportComponent implements OnInit, OnDestroy, AfterViewInit {
...
+        return isMyVault ? undefined : (vaultSelection as OrganizationId);
+      }),
+      startWith(undefined as OrganizationId | undefined),
```
itsadrago: Is this a state that we should be emitting?

itsadrago:
Per your comments on the Claude observations (#3), will you be adding tests on this PR?

libs/tools/export/vault-export/vault-export-ui/src/components/export.component.ts
```diff
@@ -342,13 +342,25 @@ export class ExportComponent implements OnInit, OnDestroy, AfterViewInit {
...
+    // Step 1: Update _organizationId$ when vault selection changes
+    // In Admin Console context, organizationId is already set via @Input
+    // In Password Manager context, user changes vaultSelector which updates _organizationId$
+    this.exportForm.controls.vaultSelector.valueChanges
+      .pipe(takeUntil(this.destroy$))
+      .subscribe((vaultSelection) => {
+        if (!this.isAdminConsoleContext) {
+          // Password Manager: Update organizationId based on vaultSelector
+          const isMyVault = vaultSelection === "myVault";
+          this.organizationId = isMyVault ? undefined : vaultSelection;
```
Hinton, line 354: **note:** Generally you should avoid "modifying" inputs since this component doesn't own the value. If it's a two-way binding you should probably use `model` over `input`. I can see that this happens in other places inside this component and that it should only occur when we don't have a binding for it, so it's probably fine.

libs/tools/export/vault-export/vault-export-ui/src/components/export.component.ts
```diff
@@ -342,13 +342,25 @@ export class ExportComponent implements OnInit, OnDestroy, AfterViewInit {
...
 
   private observeFormSelections(): void {
-    // Set up dynamic format options based on vault selection
-    this.formatOptions$ = this.exportForm.controls.vaultSelector.valueChanges.pipe(
-      startWith(this.exportForm.controls.vaultSelector.value),
-      map((vaultSelection) => {
-        const isMyVault = vaultSelection === "myVault";
-        // Update organizationId based on vault selection
-        this.organizationId = isMyVault ? undefined : vaultSelection;
+    // Step 1: Update _organizationId$ when vault selection changes
```
Hinton:
```suggestion
    // Update organizationId when vault selection changes
```

libs/tools/export/vault-export/vault-export-ui/src/components/export.component.ts
```diff
@@ -342,13 +342,25 @@ export class ExportComponent implements OnInit, OnDestroy, AfterViewInit {
...
+      .subscribe((vaultSelection) => {
+        if (!this.isAdminConsoleContext) {
+          // Password Manager: Update organizationId based on vaultSelector
+          const isMyVault = vaultSelection === "myVault";
+          this.organizationId = isMyVault ? undefined : vaultSelection;
+        }
+        // Admin Console: organizationId is already set via @Input, no update needed
+      });
+
+    // Step 2: Set up dynamic format options based on the organizationId observable
```
Hinton:
```suggestion
    // Set up dynamic format options based on the organizationId observable
```

PR #17141, opened 2025-10-30
Desktop Autotype introduce strict type for keyboard input

neuronull:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-26227

## 📔 Objective

Following up on https://github.com/bitwarden/clients/pull/16613#discussion_r2383482730 ,

This change adds a newtype wrapper `KeyboardShortcutInput` which represents a valid shortcut key. A `TryFrom` implementation was added to convert the String type we get from Electron into the new strict type.

Due to the need to support multiple OSes, we can't make take the strict type higher up in the call chain.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


apps/desktop/desktop_native/autotype/src/windows/type_input.rs
```diff
@@ -134,23 +146,28 @@ fn get_alphabetic_hotkey(letter: String) -> Result<u16> {
...
 
-/// A function for easily building keyboard unicode INPUT structs used in SendInput().
-///
-/// Before modifying this function, make sure you read the SendInput() documentation:
-/// https://learn.microsoft.com/en-in/windows/win32/api/winuser/nf-winuser-sendinput
-fn build_unicode_input(key_press: InputKeyPress, character: u16) -> INPUT {
+/// Before modifying this function, make sure you read the `SendInput()` documentation:
+/// <https://learn.microsoft.com/en-in/windows/win32/api/winuser/nf-winuser-sendinput>
+/// <https://learn.microsoft.com/en-us/windows/win32/inputdev/virtual-key-codes>
+fn build_input(key_press: InputKeyPress, character: u16, is_virtual: bool) -> INPUT {
```
neuronull, line 156: I didn't change this but Claude observed that we are setting the dwFlags unicode even for virtual keys, and said that is incorrect. I don't see anything in the microsoft docs to confirm or deny this but curious if @coltonhurst  knows.

apps/desktop/desktop_native/autotype/src/windows/type_input.rs
```diff
@@ -75,26 +88,25 @@ fn add_input(input: &[u16], keyboard_inputs: &mut Vec<INPUT>) {
...
+        const CONTROL_KEY: u16 = 0x11;
+        const ALT_KEY: u16 = 0x12;
+        const LEFT_WINDOWS_KEY: u16 = 0x5B;
+
+        let input = match key {
+            SHIFT_KEY_STR => build_virtual_key_input(InputKeyPress::Up, SHIFT_KEY),
+            CONTROL_KEY_STR => build_virtual_key_input(InputKeyPress::Up, CONTROL_KEY),
+            ALT_KEY_STR => build_virtual_key_input(InputKeyPress::Up, ALT_KEY),
+            LEFT_WINDOWS_KEY_STR => build_virtual_key_input(InputKeyPress::Up, LEFT_WINDOWS_KEY),
+            _ => build_unicode_input(InputKeyPress::Up, get_alphabetic_hotkey(key)?),
```
neuronull, line 103: I also didn't change this but I'm very curious why we use the "Up" keypress for the hotkeys 🤔  maybe I need some edification, and this could result in some code comments to make it more clear when to use down and up.

cc @coltonhurst

apps/desktop/desktop_native/Cargo.toml
```diff
@@ -38,6 +37,7 @@ futures = "=0.3.31"
 hex = "=0.4.3"
 homedir = "=0.3.4"
 interprocess = "=2.2.1"
+itertools = "=0.14.0"
```
jprusik, line 42: I think we need to add @bitwarden/team-autofill-desktop-dev ownership for this dep to `.github/renovate.json5`?

jprusik (approved):
~One non-blocking task~ On reconsideration, we should do that (renovate ownership) addition in this PR so it doesn't get forgotten; otherwise lgtm

jprusik (changes requested):
My only nit, but needs to be blocking, iirc:
https://github.com/bitwarden/clients/pull/17141/files#r2491069055

otherwise lgtm

apps/desktop/desktop_native/Cargo.toml
neuronull: Good callout. So the tricky thing here is that the "convention" seems to be adding the dep to this root level `Cargo.toml` , and the workspace crates use as needed. The workspace crates have divergent team ownership. 🤔 
Looking at `renovate.json5` , it looks like the @bitwarden/team-platform-dev owns most of the root level crate dependencies.
I'm not sure what the best solution is here.
I could move the itertools dep that I added, down to the `autotype` Cargo.toml, and then it is very clear that the ownership is @bitwarden/team-autofill-desktop-dev.
However, `itertools` is a rather nice utility that other workspace crates would surely benefit from. So to me it makes sense to have that dep in the desktop native root Cargo.toml.

neuronull: We chatted out of band on this and since it's a generic dep, Platform will own it and it makes sense to stay in the root.

added to renovate in 
https://github.com/bitwarden/clients/pull/17141/commits/06384387e2735a8671721b9d3b570d40a4799540

jprusik (dismissed):

apps/desktop/desktop_native/autotype/src/windows/type_input.rs
```diff
@@ -134,23 +146,28 @@ fn get_alphabetic_hotkey(letter: String) -> Result<u16> {
...
     match key_press {
         InputKeyPress::Down => INPUT {
             r#type: INPUT_KEYBOARD,
             Anonymous: INPUT_0 {
                 ki: KEYBDINPUT {
-                    wVk: Default::default(),
-                    wScan: character,
+                    wVk: w_vk,
+                    wScan: w_scan,
                     dwFlags: KEYEVENTF_UNICODE,
```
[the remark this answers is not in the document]
neuronull: The quoted text does not exist in the link provided. Sus.
Since I am not changing the behavior of this from what we currently have on main, I'll check in with Colton when he returns.
This should not block the PR.

apps/desktop/desktop_native/autotype/src/windows/type_input.rs
```diff
@@ -75,26 +88,25 @@ fn add_input(input: &[u16], keyboard_inputs: &mut Vec<INPUT>) {
...
+    type Error = anyhow::Error;
+
+    fn try_from(key: &str) -> std::result::Result<Self, Self::Error> {
+        const SHIFT_KEY: u16 = 0x10;
+        const CONTROL_KEY: u16 = 0x11;
+        const ALT_KEY: u16 = 0x12;
+        const LEFT_WINDOWS_KEY: u16 = 0x5B;
+
+        let input = match key {
+            SHIFT_KEY_STR => build_virtual_key_input(InputKeyPress::Up, SHIFT_KEY),
```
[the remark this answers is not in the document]
neuronull: I added a code comment to address this

neuronull:
>  ❌ Finding 1: Regression bug in build_input function - incorrect KEYEVENTF_UNICODE flag usage

> The consolidation of build_unicode_input and build_virtual_key_input into a single build_input function introduced a critical regression. The new function incorrectly applies KEYEVENTF_UNICODE to both virtual keys and Unicode characters.

Objectively false. No regression was introduced because existing behavior was maintained. This is an example of claude not being able to do higher level analysis. The lines of code may have changed but it's easy for a human reviewer to see that the behavior is not changed.

neuronull:
This has been validated by QA

dereknance (approved):

neuronull:
solved a few merge conflicts

neuronull:
resolved a merge conflict

apps/desktop/desktop_native/Cargo.toml
```diff
@@ -39,6 +39,8 @@ futures = "=0.3.31"
 hex = "=0.4.3"
 homedir = "=0.3.6"
 interprocess = "=2.2.1"
+itertools = "=0.14.0"
+keytar = "=0.1.6"
```
coltonhurst: What are we using `keytar` for?

apps/desktop/desktop_native/autotype/src/windows/type_input.rs
```diff
@@ -135,23 +148,28 @@ fn get_alphabetic_hotkey(letter: String) -> Result<u16> {
...
 
-/// A function for easily building keyboard unicode INPUT structs used in SendInput().
-///
-/// Before modifying this function, make sure you read the SendInput() documentation:
-/// https://learn.microsoft.com/en-in/windows/win32/api/winuser/nf-winuser-sendinput
-fn build_unicode_input(key_press: InputKeyPress, character: u16) -> INPUT {
+/// Before modifying this function, make sure you read the `SendInput()` documentation:
+/// <https://learn.microsoft.com/en-in/windows/win32/api/winuser/nf-winuser-sendinput>
+/// <https://learn.microsoft.com/en-us/windows/win32/inputdev/virtual-key-codes>
+fn build_input(key_press: InputKeyPress, character: u16, is_virtual: bool) -> INPUT {
```
coltonhurst, line 156: Any reason the function description was taken out?

apps/desktop/desktop_native/autotype/src/windows/type_input.rs
```diff
@@ -1,28 +1,43 @@
...
+
 impl InputOperations for Win32InputOperations {
     fn send_input(inputs: &[INPUT]) -> u32 {
-        const INPUT_STRUCT_SIZE: i32 = std::mem::size_of::<INPUT>() as i32;
-        let insert_count = unsafe { SendInput(inputs, INPUT_STRUCT_SIZE) };
+        let size = *(INPUT_STRUCT_SIZE.get_or_init(|| {
+            i32::try_from(std::mem::size_of::<INPUT>()).expect("have space for INPUT size")
+        }));
+
+        let insert_count = unsafe { SendInput(inputs, size) };
```
coltonhurst: Just curious, why wrap the input struct size in a OnceLock? Seems like we are adding more complexity for little benefit in a codebase that's currently not really multithreaded?

I may be missing something though, especially if new code was added that I've missed and is more relevant to the answer.

apps/desktop/desktop_native/autotype/src/windows/type_input.rs
coltonhurst: I'm guessing because this is resolved you no longer have the question? Let me know if I'm wrong!

apps/desktop/desktop_native/autotype/src/windows/mod.rs
```diff
@@ -36,6 +40,23 @@ pub fn get_foreground_window_title() -> Result<String> {
...
+/// `KeyboardShortcutInput` is an `INPUT` of one of the valid shortcut keys:
+///     - Control
+///     - Alt
+///     - Super
+///     - Shift
+///     - \[a-z\]\[A-Z\]
+struct KeyboardShortcutInput(INPUT);
+
+pub fn type_input(input: &[u16], keyboard_shortcut: &[String]) -> Result<()> {
+    debug!(?keyboard_shortcut, "Using keyboard shortcut.");
```
coltonhurst: Non-blocking nit: Maybe this is a bit pedantic, but maybe we use a different debug message? The message "Using keyboard shortcut." could be true at various abstraction levels. However, a message like (just for example): "windowing::type_input() called" would be more specific to this location.

apps/desktop/desktop_native/autotype/src/windows/type_input.rs
```diff
@@ -173,53 +191,29 @@ fn build_unicode_input(key_press: InputKeyPress, character: u16) -> INPUT {
...
-        },
-    }
+/// A function for easily building keyboard unicode `INPUT` structs used in `SendInput()`.
+fn build_unicode_input(key_press: InputKeyPress, character: u16) -> INPUT {
+    build_input(key_press, character, IS_REAL_KEY)
+}
+
+/// A function for easily building keyboard virtual-key `INPUT` structs used in `SendInput()`.
+fn build_virtual_key_input(key_press: InputKeyPress, character: u16) -> INPUT {
+    build_input(key_press, character, IS_VIRTUAL_KEY)
```
coltonhurst, line 197: A question for here plus for some other places... why switch certain things to `u16`?

One reason `virtual_key` was a `u8` (and possibly in other places where I used `u8` in other places) is because of the docs [here](https://learn.microsoft.com/en-us/windows/win32/api/winuser/ns-winuser-keybdinput#members) for `wVk`.

`wVk` will be a code in the range 1 to 254, or 0. It maybe wasn't the most ergonomic internally, but I added the `u8` specific parameters to help ensure that at least to "external" users of these functions, there was more type safety.

apps/desktop/desktop_native/autotype/src/windows/type_input.rs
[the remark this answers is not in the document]
coltonhurst: Hilarious. Yeah, it looks like it's partly pulling from [here](https://learn.microsoft.com/en-us/windows/win32/api/winuser/ns-winuser-keybdinput#remarks), but it also seems to have hallucinated this part (besides getting links wrong, etc)...

> If KEYEVENTF_UNICODE is not specified, wVk is used.

---

Regarding the core concern:

> The KEYEVENTF_UNICODE flag should NOT be set for virtual-key codes.

This doesn't match its own quote from the docs:

> If KEYEVENTF_UNICODE is specified, SendInput sends a WM_KEYDOWN or WM_KEYUP message to the foreground thread's message queue with wParam equal to VK_PACKET.

To me, using the unicode flag seems to clearly indicate from the docs that a `WM_KEYDOWN` or `WM_KEYUP` message is sent, which to my knowledge, is what we want. I think we can dismiss claude's hallucinations for now, but @neuronull if you would like to briefly check the docs [here](https://learn.microsoft.com/en-us/windows/win32/api/winuser/ns-winuser-keybdinput) and sanity check me that would be great! There is a chance it is picking up on something but I can't find it.

apps/desktop/desktop_native/Cargo.toml
neuronull: i believe this was due to a merge conflict resolution, ill remove it

apps/desktop/desktop_native/autotype/src/windows/type_input.rs
neuronull: I found it to be less of a function description and more of a suggestion to not be a bad programmer, which I felt like didn't add a ton of value over just a link to the documentation. If you feel strongly about it I can add it back though.

apps/desktop/desktop_native/autotype/src/windows/mod.rs
neuronull: We get the module path for free with `tracing` already but I'll add a reference to the function call.

apps/desktop/desktop_native/autotype/src/windows/type_input.rs
neuronull: I honestly have no idea. It's been a minute since I wrote the code. It might have been leftover from an earlier iteration on it. While the size that we are computing won't change over calls to send_input, std::mem::size_of is not expensive to compute at all, in fact I believe the value is dropped in there at compile time so arguably in addition to adding complexity, this code is probably _less_ performant. I'll remove it.

apps/desktop/desktop_native/autotype/src/windows/type_input.rs
neuronull: i love windows so much. lets use a u16 but valid values are only u8.
https://docs.rs/windows-sys/latest/windows_sys/Win32/UI/Input/KeyboardAndMouse/struct.KEYBDINPUT.html

I probably did this to avoid the cast, seeing that it takes a u16. I'll change back to u8 and add a code comment.

neuronull: Looking at this some more... in the before case, we are bringing in an array of u16 for the input , then we cast that down to a u8, then we're recasting to a u16... _that_ does feel convoluted. I'll take a closer look tomorrow.

apps/desktop/desktop_native/autotype/src/windows/type_input.rs
neuronull: garbage nuggets. I had this backwards. you're right. If you are in agreement, I'll remove the "Before modifying.." line, and re-add the "A function for". I was tripped up by the fact you highlighted the add side but referenced removal.

neuronull: waaaait , it's because this became a helper function. See it is now called by two other functions, where the function description was preserved.

neuronull: another case of not remembering what I did 2 months ago lol.

apps/desktop/desktop_native/autotype/src/windows/type_input.rs
neuronull: so here is what I see in the code on main, correct me if I have it wrong:

- from electron, we use `number` type, which I believe is a u64.
- we implicitly downcast to a u16 in the napi layer (this seems problematic)
- we downcast from u16 to u8 in the type_input code path
- we then cast from u8 back to u16 in the struct final usage.

I think we should at a minimum be consistent. Either use u8 all the way and then use as a u16 at the final struct usage, or use as a u16 all the way.

I'm trying to think of it operationally. I understand the type safety concern for sure, we don't want to accidentally use an invalid value in the win32 struct. But if I'm following the code correctly, would we only ever get u8 values from the electron side?

Another thing to consider, this is currently tightly coupled with the win32 api, if we are expanding that to other platforms going forward, we may run into a need to change the vec[u16] api for the entry point to type_input anyway (all that to say, we might not spend a lot of time going to lengths to perfect this).

coltonhurst: Yeah, if we look at it as a whole, it's definitely not great type-wise. The original goal was to at least keep the layer within desktop native accurate to win api rules.

I'm fine if we want to move this work into its own ticket to unblock this, up to you.

neuronull: I'm ok with moving to a new ticket. To be clear, the following options are what makes sense to me:

1. changing type_input's argument to be a vec[u8]  , this avoids the downcast from u16 to u8 that would have to happen to maintain any safety around win32 's api
2. use u16 as it is in the PR already

Regardless of any option taken, adding some logic on electron side before the downcast happens at napi layer.

If you're ok with keeping the changes as is on this branch, I can open a follow up ticket with the above.

neuronull: https://bitwarden.atlassian.net/browse/PM-29967

coltonhurst: LGTM, thanks!

PR #17139, opened 2025-10-30
(Billing) [PM-27562] Create PremiumInterestStateService 

rr-bw:
## 🎟️ Tracking

[PM-27562](https://bitwarden.atlassian.net/browse/PM-27562)

See [Epic](https://bitwarden.atlassian.net/browse/PM-26140) if you'd like more context.

## 📔 Objective

Creates a `PremiumInterestStateService` that manages state which conveys whether or not a user intends to setup a premium subscription. Implemented in Web only. No-op for other clients.

This will apply for users who began the registration process on https://bitwarden.com/go/start-premium/, which is a marketing page designed to streamline users who intend to setup a premium subscription after registration.

## ⏰ Reminders before review

- Contributor guidelines followed
- All formatters and local linters executed and passed
- Written new unit and / or integration tests where applicable
- Protected functional changes with optionality (feature flags)
- Used internationalization (i18n) for all UI strings
- CI builds passed
- Communicated to DevOps any deployment requirements
- Updated any necessary documentation (Confluence, contributing docs) or informed the documentation team

## 🦮 Reviewer guidelines

- 👍 (`:+1:`) or similar for great changes
- 📝 (`:memo:`) or ℹ️ (`:information_source:`) for notes or general info
- ❓ (`:question:`) for questions
- 🤔 (`:thinking:`) or 💭 (`:thought_balloon:`) for more open inquiry that's not quite a confirmed issue and could potentially benefit from discussion
- 🎨 (`:art:`) for suggestions / improvements
- ❌ (`:x:`) or ⚠️ (`:warning:`) for more significant problems or concerns needing attention
- 🌱 (`:seedling:`) or ♻️ (`:recycle:`) for future improvements or indications of technical debt
- ⛏ (`:pick:`) for minor or nitpick changes

[PM-27562]: https://bitwarden.atlassian.net/browse/PM-27562?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #17143, opened 2025-10-30
(Vault) [PM-27543] Create PremiumSetupIntentRedirectGuard

rr-bw:
## 🎟️ Tracking

[PM-27543](https://bitwarden.atlassian.net/browse/PM-27543)

This is a stacked PR that depends upon work in:
- https://github.com/bitwarden/clients/pull/17139
- I've set the `base:` to `/billing/pm-27562/premium-setup-service`

See [Epic](https://bitwarden.atlassian.net/browse/PM-26140) if you'd like more context.

## 📔 Objective

Creates a `PremiumSetupIntentRedirectGuard` that redirects a user based on whether or not they signaled an intent to setup a premium subscription.

This applies for users who began the registration process on https://bitwarden.com/go/start-premium/, which is a marketing page designed to streamline users who intend to setup a premium subscription after registration.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

[PM-27543]: https://bitwarden.atlassian.net/browse/PM-27543?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/browser/src/auth/popup/account-switching/account-switcher.component.ts
+++ b/apps/browser/src/auth/popup/account-switching/account-switcher.component.ts
@@ -122,10 +122,8 @@ export class AccountSwitcherComponent implements OnInit, OnDestroy {
 
   async lock(userId: string) {
     this.loading = true;
-    await this.vaultTimeoutService.lock(userId);
-    // FIXME: Verify that this floating promise is intentional. If it is, add an explanatory comment and ensure there is proper error handling.
-    // eslint-disable-next-line @typescript-eslint/no-floating-promises
-    this.router.navigate(["lock"]);
+    await this.lockService.lock(userId as UserId);
+    await this.router.navigate(["lock"]);
   }
 
   async lockAll() {
--- a/apps/browser/src/auth/popup/accounts/foreground-lock.service.ts
+++ b/apps/browser/src/auth/popup/accounts/foreground-lock.service.ts
@@ -6,10 +6,13 @@ import {
   MessageListener,
   MessageSender,
 } from "@bitwarden/common/platform/messaging";
-import { Utils } from "@bitwarden/common/platform/misc/utils";
+import { newGuid } from "@bitwarden/guid";
+import { UserId } from "@bitwarden/user-core";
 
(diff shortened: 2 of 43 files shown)


apps/web/src/app/vault/guards/premium-interest-redirect/premium-interest-redirect.guard.ts
```diff
@@ -0,0 +1,35 @@
...
+      return router.createUrlTree(["/login"]);
+    }
+
+    const intendsToSetupPremium = await premiumInterestStateService.getPremiumInterest(
+      currentAcct.id,
+    );
+
+    if (intendsToSetupPremium) {
+      return router.createUrlTree(["/settings/subscription/premium"]);
+    }
```
nick-livefront, line 30: ❓ Should the `clearPremiumInterest` be called here to ensure that the user doesn't infinite loop themselves in the guard?

nick-livefront (dismissed):
With the state of the epic it does not look like this is testable at the moment, is that accurate?  Adding a check for the feature flag in the guard would be nice but that doesn't exist yet. 

Regardless the code changes are straight forward and accomplish the goal. 👍

apps/web/src/app/vault/guards/premium-interest-redirect/premium-interest-redirect.guard.ts
shane-melton: Great question! That's being taken care of in https://github.com/bitwarden/clients/pull/17221

shane-melton:
> With the state of the epic it does not look like this is testable at the moment, is that accurate? Adding a check for the feature flag in the guard would be nice but that doesn't exist yet.
> 
> Regardless the code changes are straight forward and accomplish the goal. 👍

@nick-livefront Correct, its not in a testable state without the server changes. Also, the feature flag is planned only for the server before adding the `fromMarketing` query parameter that kicks all the other processes off. So we shouldn't need the feature flag check client side, and can just rely on that query param being the only way to set the premium interest state.

PR #17197, opened 2025-11-03
[PM-26944] fix(browser/phishing-detection): fix various issues

willmartian:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-26944

## 📔 Objective

Fixes:
- The `update$` observable on the `PhishingDataService` was not immediately emitting due to the pipe chain containing `startWith()`. This will never emit. Instead, it should be `startWith(undefined)`. [371902d](https://github.com/bitwarden/clients/pull/17197/commits/371902d85a63934f6fbf9008de13b05218a13784)
- The phishing warning UI was not line-breaking long URLs [6f3b51b](https://github.com/bitwarden/clients/pull/17197/commits/6f3b51b44929069c7bf129beac7406d785922599)
- Account switching between multiple accounts with access would cause multiple subscriptions and as a result degrade performance. [fc257b8](https://github.com/bitwarden/clients/pull/17197/commits/fc257b8e87ceb469932574a66360bb3be4a64e27)
- Refactor/rewrite most of the `PhishingDetectionService`. 
  - We were running into sneaky performance issues with the prior implementation. I could not identify a good surgical fix, and the prior implementation needed a refactor anyways due to mixing of RxJS and imperative manipulation. 
  - Managed to delete around 200 lines.
  - Update to use `MessageListener` and `MessageSender` instead of raw `chrome` APIs.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/apps/browser/src/dirt/phishing-detection/pages/protected-by-component.html
+++ b/apps/browser/src/dirt/phishing-detection/pages/protected-by-component.html
@@ -1,1 +1,1 @@
-<span class="tw-text-muted">{{ "protectedBy" | i18n: "Bitwarden Phishing Blocker" }}</span>
+<span class="tw-text-muted">{{ "protectedBy" | i18n: "Bitwarden phishing blocker" }}</span>


Banrion (dismissed):

apps/browser/src/dirt/phishing-detection/pages/phishing-warning.stories.ts
```diff
@@ -49,6 +50,13 @@ export default {
           provide: PlatformUtilsService,
           useClass: MockPlatformUtilsService,
         },
+        {
+          provide: MessageSender,
+          useValue: {
+            // eslint-disable-next-line no-console
+            send: (...args: any[]) => console.log("MessageSender called with:", args),
```
Banrion: ⚠️ Change to console.debug if we want to keep

apps/browser/src/dirt/phishing-detection/services/phishing-detection.service.ts
```diff
@@ -1,411 +1,193 @@
...
-import {
-  CaughtPhishingDomain,
-  isPhishingDetectionMessage,
-  PhishingDetectionMessage,
-  PhishingDetectionNavigationEvent,
-  PhishingDetectionTabId,
-} from "./phishing-detection.types";
-
-export class PhishingDetectionService {
-  private static _destroy$ = new Subject<void>();
```
Banrion, line 21: ❓ Are we sure there is no cleanup that needs to be done outside of the initSub and this is safe to remove?

willmartian: Yes, the logic was shifted around a bit such that:
1. we are only ever explicitly subscribing once (in the init function)
2. enabled/disabling the feature is handled inside of a `switchMap`, which doesn't need to be manually destroyed

So the extra cleanup shouldn't be necessary.

apps/browser/src/dirt/phishing-detection/pages/phishing-warning.stories.ts
willmartian: [2285e7d](https://github.com/bitwarden/clients/pull/17197/commits/2285e7da35d14ff6e1e45c56fc08c1dd438b6882) ✅

PR #17209, opened 2025-11-04
[PM-27225] Fix nothing showing when biometrics unavailable

quexten:
## 🎟️ Tracking
https://bitwarden.atlassian.net/browse/PM-27225
## 📔 Objective

When a user without a master password (TDE/key connector) logs in and has an unreliable unlock method (biometrics), if biometrics is temporarily unavailable, currently a broken component is shown. This PR detects that case and shows the biometric component, with the button being unavailable, so the user realizes they have to re-connect their biometrics device (unfold their laptop).

Further, in the case when a user cannot unlock - no password and no temporarily unavailable unlock method - we log out the user automatically.

Also adds the unavailability reason as a tooltip as per product's suggestion.

## 📸 Screenshots

image

image

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


libs/key-management-ui/src/lock/components/lock.component.ts
```diff
@@ -276,7 +279,26 @@ export class LockComponent implements OnInit, OnDestroy {
...
+    );
+    const canUsePin =
+      (await this.pinService.isPinSet(activeAccount.id)) &&
+      (await this.pinService.isPinDecryptionAvailable(activeAccount.id));
+    const canUseBiometrics = [
+      BiometricsStatus.Available,
+      BiometricsStatus.HardwareUnavailable,
+      BiometricsStatus.DesktopDisconnected,
+      BiometricsStatus.NotEnabledInConnectedDesktopApp,
+    ].includes(await this.biometricService.getBiometricsStatusForUser(activeAccount.id));
```
mzieniukbw: use `LockComponentService.getAvailableUnlockOptions$` instead

libs/key-management-ui/src/lock/components/lock.component.ts
```diff
@@ -276,7 +279,26 @@ export class LockComponent implements OnInit, OnDestroy {
...
+    if (!canUsePassword && !canUsePin && !canUseBiometrics) {
+      // User has no available unlock options, force logout. This happens for TDE users without a masterpassword, that don't have a persistent unlock method set.
+      this.logService.warning("[LockComponent] User cannot unlock again. Logging out!");
+      await this.logoutService.logout(activeAccount.id);
+      return;
+    }
```
mzieniukbw: this feels like `VaultTimeoutSettingsService.canLock`. There is so much overlap, could we re-use something ?

libs/key-management-ui/src/lock/components/lock.component.ts
```diff
@@ -299,14 +321,30 @@ export class LockComponent implements OnInit, OnDestroy {
...
+
     // Priorities should be Biometrics > Pin > Master Password for speed
     if (unlockOptions?.biometrics.enabled) {
       this.activeUnlockOption = UnlockOption.Biometrics;
     } else if (unlockOptions?.pin.enabled) {
       this.activeUnlockOption = UnlockOption.Pin;
     } else if (unlockOptions?.masterPassword.enabled) {
       this.activeUnlockOption = UnlockOption.MasterPassword;
+      // If biometrics is temporarily unavailable for masterpassword-less users, but they have biometrics configured,
+      // then show the biometrics screen so the user knows why they can't unlock, and to give them the option to log out.
```
mzieniukbw: shouldn't this comment be inside the other else if condition ? 🤔

libs/key-management-ui/src/lock/components/lock.component.ts
```diff
@@ -299,14 +321,30 @@ export class LockComponent implements OnInit, OnDestroy {
     });
   }
 
-  private setDefaultActiveUnlockOption(unlockOptions: UnlockOptions | null) {
+  private async setDefaultActiveUnlockOption(unlockOptions: UnlockOptions | null) {
+    const biometricsStatus: BiometricsStatus | null =
+      this.activeAccount != null
+        ? await this.biometricService.getBiometricsStatusForUser(this.activeAccount.id)
```
mzieniukbw: use `unlockOptions?.biometrics.biometricsStatus` instead

libs/key-management-ui/src/lock/components/lock.component.ts
```diff
@@ -299,14 +321,30 @@ export class LockComponent implements OnInit, OnDestroy {
...
       this.activeUnlockOption = UnlockOption.MasterPassword;
+      // If biometrics is temporarily unavailable for masterpassword-less users, but they have biometrics configured,
+      // then show the biometrics screen so the user knows why they can't unlock, and to give them the option to log out.
+    } else if (
+      biometricsStatus != null &&
+      [
+        BiometricsStatus.HardwareUnavailable,
+        BiometricsStatus.DesktopDisconnected,
+        BiometricsStatus.NotEnabledInConnectedDesktopApp,
+      ].includes(biometricsStatus)
```
mzieniukbw: I feel like `biometricStateService.getBiometricUnlockEnabled` is more reliable.
Possibly something to change for `unlockOptions?.biometrics.enabled`. The "enabled" should say whether it's enabled (as a setting, not availability).
That way this "special" condition could live inside the biometrics.enabled condition and not separately.

Also, a combination of `biometricStateService.getBiometricUnlockEnabled == true` and `biometricsStatus !=`BiometricsStatus.Available` would be as valid as above if conditions, without having to check for all other `BiometricsStatus`'s.
(Since it is not possible to enable it, if it is genuinely not available).

libs/key-management-ui/src/lock/components/lock.component.ts
```diff
@@ -276,7 +279,26 @@ export class LockComponent implements OnInit, OnDestroy {
...
+      this.userDecryptionOptionsService.hasMasterPassword$,
+    );
+    const canUsePin =
+      (await this.pinService.isPinSet(activeAccount.id)) &&
+      (await this.pinService.isPinDecryptionAvailable(activeAccount.id));
+    const canUseBiometrics = [
+      BiometricsStatus.Available,
+      BiometricsStatus.HardwareUnavailable,
+      BiometricsStatus.DesktopDisconnected,
+      BiometricsStatus.NotEnabledInConnectedDesktopApp,
```
mzieniukbw: 🎨 maybe a helper function, since we repeat ourselves in few places: 2 in here and one in https://github.com/bitwarden/clients/blob/a860f218bd272787ddad0972b03689a55d7ee37a/apps/browser/src/key-management/biometrics/background-browser-biometrics.service.ts#L156

libs/key-management-ui/src/lock/components/lock.component.ts
```diff
@@ -299,14 +321,30 @@ export class LockComponent implements OnInit, OnDestroy {
...
+      // then show the biometrics screen so the user knows why they can't unlock, and to give them the option to log out.
+    } else if (
+      biometricsStatus != null &&
+      [
+        BiometricsStatus.HardwareUnavailable,
+        BiometricsStatus.DesktopDisconnected,
+        BiometricsStatus.NotEnabledInConnectedDesktopApp,
+      ].includes(biometricsStatus)
+    ) {
+      this.activeUnlockOption = UnlockOption.Biometrics;
```
mzieniukbw, line 349: ❓ When you unlocked with biometrics and restarted your desktop app, the status is then likely `BiometricsStatus.UnlockNeeded`, which is also valid ?

mzieniukbw (changes requested):

libs/key-management-ui/src/lock/components/lock.component.ts
quexten: If the user has no other unlock options (PIN, MasterPassword) after app restart, and the Biometrics is on `UnlockNeeded` then they can never unlock, because there is no unlock option that can ever unlock. (To "activate" unlock needed you need to unlock first and then lock again).

This case should be covered by the forced logout above.

libs/key-management-ui/src/lock/components/lock.component.ts
quexten: Moved out the three conditions to a const.

libs/key-management-ui/src/lock/components/lock.component.ts
quexten: Updated

libs/key-management-ui/src/lock/components/lock.component.ts
quexten: Moved

libs/key-management-ui/src/lock/components/lock.component.ts
quexten: Done for master password and pin, for biometric the function cannot be used as it only checks for available.

libs/key-management-ui/src/lock/components/lock.component.ts
quexten: No, because this leaves out the case of a TDE user that has biometrics with require MP on app start (AFU-only). For this case `canLock` is true, but we should still log-out after app restart because there is no possibility to unlock.

libs/key-management-ui/src/lock/components/lock.component.ts
quexten: I don't think it can. The branches of the if else here are the order in which we display. The order we want is:
Biometrics (if available) -> Pin (if available) -> MP (if available) -> Biometrics (if temporarily unavailable)

Putting both into the same branch would mean that a temporarily unavailable biometrics would always be taken first, even if PIN / MP are available.

libs/key-management-ui/src/lock/components/lock.component.ts
mzieniukbw, line 1: nit: would be great to have some test coverage for this.

quexten: Added tests

mzieniukbw (approved):

quexten:
Had to fix two QA reported bugs but it's good to merge now.

apps/desktop/src/key-management/biometrics/renderer-biometrics.service.ts
```diff
@@ -31,6 +37,10 @@ export class RendererBiometricsService extends DesktopBiometricsService {
   }
 
   async getBiometricsStatusForUser(id: UserId): Promise<BiometricsStatus> {
+    if ((await firstValueFrom(this.tokenService.hasAccessToken$(id))) === false) {
+      return BiometricsStatus.NotEnabledInConnectedDesktopApp;
```
mzieniukbw, line 41: ❓  according to documentation of `NotEnabledInConnectedDesktopApp`, this is supposed to be only used in browser only.
Also see https://github.com/bitwarden/clients/blob/b7287d46143c2ecbaa5bef28f53d1bcc0f27f7cd/apps/desktop/src/services/biometric-message-handler.service.ts#L253, which converts to desired biometric status, before sending off to browser over ipc.

PR #17211, opened 2025-11-04
[PM-27766] Add policy for blocking account creation from claimed domains

kspearrin:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-27766

## 📔 Objective

Enterprise organizations should have a new policy that they can enable to block users from creating individual accounts outside of their organization. We use claimed domains to enforce this ownership.

## 📸 Screenshots (video)

https://github.com/user-attachments/assets/cfd9e2f6-4c83-4416-b5dc-ef62a59e63eb


bitwarden_license/bit-web/src/app/admin-console/policies/policy-edit-definitions/block-claimed-domain-account-creation.component.ts
```diff
@@ -0,0 +1,33 @@
...
+      .pipe(map((enabled) => enabled && organization.useOrganizationDomains));
+  }
+}
+
+// FIXME(https://bitwarden.atlassian.net/browse/CL-764): Migrate to OnPush
+// eslint-disable-next-line @angular-eslint/prefer-on-push-component-change-detection
+@Component({
+  templateUrl: "block-claimed-domain-account-creation.component.html",
+  imports: [SharedModule],
+})
```
jrmccannon: I think this should use `OnPush` change detection.

kspearrin: Can you show me a policy that does this?

jrmccannon: It was done [here](https://github.com/bitwarden/clients/blob/main/apps/web/src/app/admin-console/organizations/policies/policy-edit-definitions/uri-match-default.component.ts)

PR #17224, opened 2025-11-04
[PM-27711] Loading Skeleton page

nick-livefront:
## 🎟️ Tracking

[PM-27711](https://bitwarden.atlassian.net/browse/PM-27711)

## 📔 Objective

Create reusable component to display loading skeletons outlining vault items.
I chatted with Emily about a couple of things:
- The horizontal divider _should not_ pulse
- The page should not scroll while the skeletons are present

#### Changes to PopupPage:
- Adds `hideOverflow` input to the `PopupPageComponent` to allow consumer to disallow scrolling. This allws the skeletons to overflow without the user able to scroll.
- I also addressed the FIXMEs

## 📸 Screenshots

Screen recording via storybook:
<video src="https://github.com/user-attachments/assets/6efbfb2d-6996-4be9-abaa-cbfff13049a1" />

## 🦮 Reviewer guidelines

[PM-27711]: https://bitwarden.atlassian.net/browse/PM-27711?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/browser/src/vault/popup/components/vault-loading-skeleton/vault-loading-skeleton.component.html
+++ b/apps/browser/src/vault/popup/components/vault-loading-skeleton/vault-loading-skeleton.component.html
@@ -1,4 +1,4 @@
-<section>
+<section aria-hidden="true">
   <div class="tw-mt-1.5 tw-flex tw-flex-col tw-gap-4">
     <bit-skeleton-text class="tw-w-[8.625rem] tw-max-w-full tw-mb-2.5"></bit-skeleton-text>
     @for (num of numberOfItems; track $index) {
@@ -9,7 +9,7 @@
           <bit-skeleton class="tw-w-24 tw-h-2.5 tw-max-w-full"></bit-skeleton>
         </div>
       </bit-skeleton-group>
-      <hr aria-hidden="true" class="tw-h-[1px] -tw-mr-3 tw-bg-secondary-100 tw-border-none" />
+      <hr class="tw-h-[1px] -tw-mr-3 tw-bg-secondary-100 tw-border-none" />
     }
   </div>
 </section>


shane-melton (dismissed):
Vault changes looks good to me!

apps/browser/src/vault/popup/components/vault-loading-skeleton/vault-loading-skeleton.component.html
```diff
@@ -0,0 +1,16 @@
+<section>
+  <p class="tw-sr-only" role="status">Loading...</p>
```
vleague2: Since this isn't being used in the vault page yet, I'm not able to test the SR interaction, but I suspect we will want to use the aria-live strategy instead so that SR users will be made aware when content is done loading instead of silently losing focus (again, me assuming that's what will happen): https://components.bitwarden.com/?path=/docs/component-library-skeleton--docs#accessibility 

This would happen at the vault page level, leaving the skeleton purely visually presentational and hidden from screenreaders.

vleague2 (commented):
Ignore Claude's comments about `standalone: "true"`, that is on by default.

apps/browser/src/vault/popup/components/vault-loading-skeleton/vault-loading-skeleton.component.html
nick-livefront: @vleague2 can you elaborate? Your assumption about usage is right, [role="status"](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/status_role) implicitly has an `aria-live="polite"`. Or are you suggesting that I hide the entire contents here from screen readers and instead use the `LiveAnnouncer` instead?

vleague2: The latter! Announcing the "loading" is great, but when that disappears, there should be an announcement that the content is now loaded, because as far as I understand it, there is no other way for the SR user to know that the skeleton is gone and that the content is now rendered (sorry that was a long sentence). Imo it would be easiest to handle both with the LiveAnnouncer at whatever component is controlling the visibility of the skeleton vs the vault content. However! I have not actually tested this haha.

nick-livefront: Agreed! LiveAnnouncer will work better for both of those cases, my next ticket [PM-25084](https://bitwarden.atlassian.net/browse/PM-25084) is integrating this into the vault so I can manage the hiding/showing there!

Removed the internal loading text: [6fa8780](https://github.com/bitwarden/clients/pull/17224/commits/6fa8780c697f78cef1a91b1211721f5139365824) + [0b4602b](https://github.com/bitwarden/clients/pull/17224/commits/0b4602b9026eeee75aa134b2742ab96bd3dd8f50)

PR #17231, opened 2025-11-05
[PM-27803] disabled activity tab states

AlexRubik:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-27803

## 📔 Objective

Adding old empty states back for when activity tab flag is disabled.

## 📸 Screenshots

No reports, no ciphers:
image

No reports, has ciphers:
image

Any time there are zero critical apps:
image

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


bitwarden_license/bit-web/src/app/dirt/access-intelligence/all-applications/all-applications.component.html
```diff
@@ -2,73 +2,102 @@
...
-          #allAppsOrgAtRiskMembers
-          class="tw-w-full"
-          [ngClass]="{
-            'tw-bg-primary-100': drawerDetails.invokerId === 'allAppsOrgAtRiskMembers',
-          }"
-          [title]="'atRiskMembers' | i18n"
-          [value]="applicationSummary.totalAtRiskMemberCount"
-          [maxValue]="applicationSummary.totalMemberCount"
+  @let hasCiphers = dataService.hasCiphers$ | async;
+  @if (!dataSource.data.length && !hasCiphers) {
```
Banrion, line 6: ❓ I think we should display the empty table regardless if there are ciphers (the **hasCiphers**). If there is no data in the table and there are ciphers, they would still need to run the report and it should still show an empty table, right? It could be a different state to run the report if there are ciphers available.

bitwarden_license/bit-web/src/app/dirt/access-intelligence/all-applications/all-applications.component.ts
```diff
@@ -85,6 +85,15 @@ export class AllApplicationsComponent implements OnInit {
     });
   }
 
+  goToCreateNewLoginItem = async () => {
```
Banrion: ⚠️ I don't think we should add the unimplemented function. I know this was here before but we could either not display the button with the confusing non implementation or navigate them to the vault which has the create function already there. Let's confirm with Max on how to actually handle this case.

bitwarden_license/bit-web/src/app/dirt/access-intelligence/critical-applications/critical-applications.component.ts
```diff
@@ -100,13 +111,12 @@ export class CriticalApplicationsComponent implements OnInit {
...
   goToAllAppsTab = async () => {
-    await this.router.navigate(
-      [`organizations/${this.organizationId}/access-intelligence/risk-insights`],
-      {
-        queryParams: { tabIndex: RiskInsightsTabType.AllApps },
-        queryParamsHandling: "merge",
-      },
-    );
+    // If activity tab is enabled, All Apps is tab 1, otherwise it's tab 0
+    const allAppsTabIndex = this.isRiskInsightsActivityTabFeatureEnabled ? 1 : 0; //TODO: change this logic when feature flag is implemented?
```
Banrion: ❌ Use the **RiskInsightsTabType** to get the actual tab number instead of the hard coded vault.

bitwarden_license/bit-web/src/app/dirt/access-intelligence/critical-applications/critical-applications.component.ts
```diff
@@ -65,13 +67,22 @@ export class CriticalApplicationsComponent implements OnInit {
...
   }
 
   async ngOnInit() {
+    // Subscribe to feature flag
+    this.configService
+      .getFeatureFlag$(FeatureFlag.PM22887_RiskInsightsActivityTab)
+      .pipe(takeUntilDestroyed(this.destroyRef))
+      .subscribe((isEnabled) => {
+        this.isRiskInsightsActivityTabFeatureEnabled = isEnabled;
+      });
```
Banrion, line 83: ❌ You can remove this when you use the **RiskInsightsTabType** to get the tab value so checking if the flag is enabled would be unnecessary

bitwarden_license/bit-web/src/app/dirt/access-intelligence/critical-applications/critical-applications.component.ts
AlexRubik: The tab indices change when the activity tab feature flag is enabled. When enabled, activity tab is index 0 and when disabled, all apps tab is 0.

The tab type doesn't account for that so I added the conditional. RiskInsightsTabType assumes activity feature flag is enabled.

```
export enum RiskInsightsTabType {
  AllActivity = 0,
  AllApps = 1,
  CriticalApps = 2,
  NotifiedMembers = 3,
}
```

bitwarden_license/bit-web/src/app/dirt/access-intelligence/all-applications/all-applications.component.html
ttalty: Yea it should be no data in the table for empty state. 

It's a weird issue because before reports ran automatically. So the limiting factor was "does org have ciphers"

Now that we are using saved reports. We have 2 states. Org does not have ciphers so import and org did not run report and has ciphers so new report. 

This empty state is more backwards compatibility. If there is no report saved. Just add the empty state. 

We should maybe consider a "no report, no ciphers" empty state. Yet with the feature flag being removed I don't think we need to account for it.

AlexRubik: `!dataSource.data.length && !hasCiphers`

This conditional is controlling the display of the "no apps in org" state which is to be displayed when there are no ciphers and no reports:

image

If we remove the hasCiphers check, this state above will display in the case where the user has ciphers and no report data.

If there are ciphers and no report data, Max said to just show the empty table:
image

bitwarden_license/bit-web/src/app/dirt/access-intelligence/all-applications/all-applications.component.ts
AlexRubik: @maxkpower do we want to remove this button or implement the navigation
image

PR #17258, opened 2025-11-06
Ac/pm 26364 extension UI for auto confirm

BTreston:
## 🎟️ Tracking
https://bitwarden.atlassian.net/browse/PM-26364

## 📔 Objective
This PR implements the new Extension UI for the Automatic User Confirmation (auto-confirm) feature. 

  1. New Auto-Confirm Library (libs/auto-confirm/)

  - Created a standalone library
  - Migrate existing auto-confirm services, models, and abstractions into this library
  - Added README documenting the feature's purpose and security model

  2. Browser Extension Admin Settings Page

  - New component: admin-settings.component.ts (apps/browser/src/vault/popup/settings/)
  - Provides UI toggle for enabling/disabling auto-confirm in the browser extension
  - Shows warning dialog when enabling (with link to help documentation)
  - Implements spotlight/nudge notification for first-time setup

  3. Navigation & Routing

  - Added "Admin Settings" link to browser extension settings page
  - Integrated route in app-routing.module.ts with guard protection
  - Updated settings navigation template

  4. Auto-Confirm Nudge Service

  - Added auto-confirm-nudge.service.ts
  - Integrates with existing nudges system to show spotlight, berry, and badge nudges

  5. Guard Refactoring

  - Moved org-policy.guard.ts from admin-console to libs/angular/src/admin-console/guards/
  - Created new automatic-user-confirmation-settings.guard.ts for route protection

  6. Testing & Configuration

  - Enhanced test coverage for auto-confirm service
  - Updated Jest and ESLint configurations for new library

## 📸 Screenshots
DESIGN REVIEW

https://github.com/user-attachments/assets/3de994eb-6b01-401b-afcd-a9fdc1785e7f

https://github.com/user-attachments/assets/9dab3894-96ca-4a81-82e3-a49f1c9eff22

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/libs/auto-confirm/src/services/default-auto-confirm.service.ts
+++ b/libs/auto-confirm/src/services/default-auto-confirm.service.ts
@@ -53,7 +53,7 @@ export class DefaultAutomaticUserConfirmationService implements AutomaticUserCon
       this.organizationService
         .organizations$(userId)
         // auto-confirm does not allow the user to be part of any other organization (even if admin or owner)
-        // so we can assume that the first organization is the relevant one to test.
+        // so we can assume that the first organization is the relevant one.
         .pipe(map((organizations) => organizations[0])),
       this.policyService.policyAppliesToUser$(PolicyType.AutoConfirm, userId),
     ]).pipe(
--- a/libs/common/src/admin-console/models/domain/organization.ts
+++ b/libs/common/src/admin-console/models/domain/organization.ts
@@ -384,7 +384,7 @@ export class Organization {
   }
 
   /**
-   * Do not call this function to perform business logic, consider using the function in @link AutomaticUserConfirmationService instead.
+   * Do not call this function to perform business logic, use the function in @link AutomaticUserConfirmationService instead.
    **/
   get canManageAutoConfirm() {
     return this.isMember && this.canManageUsers && this.useAutomaticUserConfirmation;
--- a/apps/browser/src/tools/popup/settings/settings-v2.component.ts
+++ b/apps/browser/src/tools/popup/settings/settings-v2.component.ts
@@ -1,22 +1,13 @@
 import { CommonModule } from "@angular/common";
 import { ChangeDetectionStrategy, Component } from "@angular/core";
 import { RouterModule } from "@angular/router";
-import {
-  combineLatest,
-  filter,
-  firstValueFrom,
-  map,
-  Observable,
-  shareReplay,
-  switchMap,
-} from "rxjs";
+import { filter, firstValueFrom, Observable, shareReplay, switchMap } from "rxjs";
 
 import { PremiumUpgradeDialogComponent } from "@bitwarden/angular/billing/components";
 import { JslibModule } from "@bitwarden/angular/jslib.module";
 import { NudgesService, NudgeType } from "@bitwarden/angular/vault";
 import { SpotlightComponent } from "@bitwarden/angular/vault/components/spotlight/spotlight.component";
 import { AutomaticUserConfirmationService } from "@bitwarden/auto-confirm";
-import { AutofillBrowserSettingsService } from "@bitwarden/browser/autofill/services/autofill-browser-settings.service";
 import { Account, AccountService } from "@bitwarden/common/auth/abstractions/account.service";
 import { getUserId } from "@bitwarden/common/auth/services/account.service";
 import { BillingAccountProfileStateService } from "@bitwarden/common/billing/abstractions";
@@ -82,17 +73,8 @@ export class SettingsV2Component {
     ),
   );
 
-  showAutofillBadge$: Observable<boolean> = combineLatest([
-    this.autofillBrowserSettingsService.defaultBrowserAutofillDisabled$,
-    this.authenticatedAccount$,
-  ]).pipe(
-    switchMap(([defaultBrowserAutofillDisabled, account]) =>
-      this.nudgesService.showNudgeBadge$(NudgeType.AutofillNudge, account.id).pipe(
-        map((badgeStatus) => {
-          return !defaultBrowserAutofillDisabled && badgeStatus;
-        }),
-      ),
-    ),
+  showAutofillBadge$: Observable<boolean> = this.authenticatedAccount$.pipe(
+    switchMap((account) => this.nudgesService.showNudgeBadge$(NudgeType.AutofillNudge, account.id)),
   );
 
   showAdminSettingsLink$: Observable<boolean> = this.accountService.activeAccount$.pipe(
@@ -103,7 +85,6 @@ export class SettingsV2Component {
   constructor(
     private readonly nudgesService: NudgesService,
     private readonly accountService: AccountService,
-    private readonly autofillBrowserSettingsService: AutofillBrowserSettingsService,
     private readonly autoConfirmService: AutomaticUserConfirmationService,
     private readonly accountProfileStateService: BillingAccountProfileStateService,
     private readonly dialogService: DialogService,


apps/browser/src/tools/popup/settings/settings-v2.component.html
```diff
@@ -84,6 +84,24 @@
         <i slot="end" class="bwi bwi-angle-right" aria-hidden="true"></i>
       </a>
     </bit-item>
+
+    @if (!!showAdminSettingsLink$ && showAdminSettingsLink$ | async) {
```
eliykat: `showAdminSettingsLink$` will never be null - do we need the null check?

apps/browser/src/vault/popup/settings/admin-settings.component.ts
```diff
@@ -0,0 +1,143 @@
...
+  DialogRef,
+  DialogService,
+  FormFieldModule,
+  SwitchComponent,
+} from "@bitwarden/components";
+import { UserId } from "@bitwarden/user-core";
+
+@Component({
+  templateUrl: "./admin-settings.component.html",
+  changeDetection: ChangeDetectionStrategy.OnPush,
```
eliykat, line 37: Nice to see OnPush starting to be used 👏

apps/browser/src/vault/popup/settings/admin-settings.component.ts
```diff
@@ -0,0 +1,143 @@
...
+  imports: [
+    CommonModule,
+    JslibModule,
```
eliykat: I was curious what JslibModule was needed for here, because we shouldn't need it for brand new work like this. Looks like it's only for the i18nPipe, which is now supplied by the `@bitwarden/ui-common`. The jslib version is deprecated.

apps/browser/src/vault/popup/settings/admin-settings.component.ts
```diff
@@ -0,0 +1,143 @@
...
+export class AdminSettingsComponent implements OnInit {
+  private userId$: Observable<UserId> = this.accountService.activeAccount$.pipe(getUserId);
+  private autoConfirmState$: Observable<AutoConfirmState> = this.userId$.pipe(
+    switchMap((userId) => this.autoConfirmService.configuration$(userId)),
+  );
+
+  protected formLoading = true;
+  protected adminForm: FormGroup = this.formBuilder.group({
+    autoConfirm: false,
+  });
```
eliykat: I dislike this about Angular, but `FormGroup` is a generic that defaults to `FormGroup<any>`. So by adding the type annotation like this, you're losing type safety when referencing controls.

If you remove the type, it'll correctly infer the control names in the formgroup when you reference it later.


```suggestion
  protected adminForm = this.formBuilder.group({
    autoConfirm: false,
  });
```

apps/browser/src/vault/popup/settings/admin-settings.component.ts
```diff
@@ -0,0 +1,143 @@
...
+
+            if (result) {
+              return newValue;
+            }
+          }
+          this.adminForm.setValue({ autoConfirm: false }, { emitEvent: false });
+          return false;
+        }),
+        withLatestFrom(this.userId$, this.autoConfirmState$),
+        switchMap(([newValue, userId, existingSate]) =>
```
eliykat: Typo: `existingSate`

apps/browser/src/vault/popup/settings/admin-settings.component.ts
```diff
@@ -0,0 +1,143 @@
...
+  `,
+  imports: [ButtonModule, DialogModule, CommonModule, JslibModule],
+})
+class AutoConfirmWarningDialogComponent {
+  constructor(public dialogRef: DialogRef<boolean>) {}
+
+  static open(dialogService: DialogService) {
+    return dialogService.open<boolean>(AutoConfirmWarningDialogComponent);
+  }
+}
```
eliykat: I suggest this is moved into a separate file, otherwise there are too many things going on here: 2 separate components as well as a template.

apps/browser/src/vault/popup/settings/admin-settings.component.ts
```diff
@@ -0,0 +1,143 @@
...
+      )
+      .subscribe();
+  }
+
+  async dismissSpotlight() {
+    const userId = await firstValueFrom(this.userId$);
+    const state = await firstValueFrom(this.autoConfirmState$);
+
+    await this.autoConfirmService.upsert(userId, { ...state, showBrowserNotification: false });
+  }
```
eliykat, line 120: This suffers from a possible race condition we should be wary of when manipulating state: we get the current value of `this.userId$` on line 104, then the `autoConfirmState$` on line 105, but `autoConfirmState$` has its own subscription to `this.userId$` which may have changed in the meantime.

This can be avoided by making sure we use the same `userId` emission:

```ts
    const userId = await firstValueFrom(this.userId$);
    const state = await firstValueFrom(this.autoConfirmService.configuration$(userId)); // the userId we fetched above

    // uses the same userId again
    await this.autoConfirmService.upsert(userId, { ...state, showBrowserNotification: false });
```

The `this.adminForm.controls.autoConfirm.valueChanges` observable has the same problem. The easiest solution is probably to drop the `autoConfirmState$` helper property.

libs/angular/src/vault/services/custom-nudges-services/auto-confirm-nudge.service.ts
```diff
@@ -0,0 +1,43 @@
...
+        const status: NudgeStatus = {
+          hasBadgeDismissed: dismissed,
+          hasSpotlightDismissed: dismissed,
+        };
+
+        if (nudgeStatus.hasBadgeDismissed || nudgeStatus.hasSpotlightDismissed) {
+          return nudgeStatus;
+        }
+
+        return status;
```
eliykat: This could be reordered to clarify the logic a little. You don't need to calculate `dismissed` and `status` until after the early return. And `status` vs. `nudgeStatus` is confusing.


```suggestion
        if (nudgeStatus.hasBadgeDismissed || nudgeStatus.hasSpotlightDismissed) {
          return nudgeStatus;
        }
        
        const dismissed = autoConfirmState.showBrowserNotification === false;

        return {
          hasBadgeDismissed: dismissed,
          hasSpotlightDismissed: dismissed,
        };
```

libs/auto-confirm/src/guards/automatic-user-confirmation-settings.guard.spec.ts
```diff
@@ -0,0 +1,254 @@
...
+      await firstValueFrom(runGuard());
+
+      expect(toastService.showToast).not.toHaveBeenCalled();
+    });
+
+    it("should not redirect to vault", async () => {
+      await firstValueFrom(runGuard());
+
+      expect(router.createUrlTree).not.toHaveBeenCalled();
+    });
```
eliykat: I think there is scope to simplify these tests. The sut is 35 lines long and this is 250 lines long.

One example here: this should test the return value (whether it returns a UrlTree), not the method of constructing the return value (which is internal behaviour). And then it follows that you've already tested the return value in your first happy path test. So this one could be deleted.

libs/auto-confirm/src/guards/automatic-user-confirmation-settings.guard.spec.ts
```diff
@@ -0,0 +1,254 @@
...
+
+    it("should call canManageAutoConfirm$ with correct user id", async () => {
+      await firstValueFrom(runGuard());
+
+      expect(autoConfirmService.canManageAutoConfirm$).toHaveBeenCalledWith(mockUserId);
+    });
```
eliykat: This is also internal behaviour. I think the "multiple users scenario" test below is a more effective test of this, because it's testing the return values rather than what dependencies have been called.

libs/auto-confirm/src/guards/automatic-user-confirmation-settings.guard.spec.ts
```diff
@@ -0,0 +1,254 @@
...
+      });
+    });
+
+    it("should translate the error message", async () => {
+      router.createUrlTree.mockReturnValue({} as UrlTree);
+
+      await firstValueFrom(runGuard());
+
+      expect(i18nService.t).toHaveBeenCalledWith("noPermissionsViewPage");
+    });
```
eliykat: This is a duplicate of the previous test: showing the error toast with correct parameters means the error message must be translated.

You could move this assertion into that test (combining them), or rely on the mocked i18nService behaviour (if it didn't call i18nService, it wouldn't be using the mocked translated string when it calls toastService, and that assertion would fail).

libs/auto-confirm/src/guards/automatic-user-confirmation-settings.guard.spec.ts
```diff
@@ -0,0 +1,254 @@
...
+      });
+      router.createUrlTree.mockImplementation(() => {
+        callOrder.push("redirect");
+        return {} as UrlTree;
+      });
+
+      await firstValueFrom(runGuard());
+
+      expect(callOrder).toEqual(["toast", "redirect"]);
+    });
```
eliykat: I don't think this is needed - it redirects by returning a value, if it's calling `showToast` at all it must be doing that first.

libs/auto-confirm/src/guards/automatic-user-confirmation-settings.guard.spec.ts
```diff
@@ -0,0 +1,254 @@
...
+
+      router.createUrlTree.mockReturnValue({} as UrlTree);
+      canManageSubject.next(false);
+
+      expect(results.length).toBeGreaterThan(1);
+      expect(toastService.showToast).toHaveBeenCalled();
+
+      subscription.unsubscribe();
+    });
+  });
```
eliykat: I don't think this is adding anything over the tests above, it's just testing the same behaviour in different ways (and with more of a focus on internals rather than public behaviour).

libs/auto-confirm/src/services/default-auto-confirm.service.ts
```diff
@@ -43,15 +47,17 @@ export class DefaultAutomaticUserConfirmationService implements AutomaticUserCon
...
   }
 
-  canManageAutoConfirm$(userId: UserId, organizationId: OrganizationId): Observable<boolean> {
+  canManageAutoConfirm$(userId: UserId): Observable<boolean> {
     return combineLatest([
       this.configService.getFeatureFlag$(FeatureFlag.AutoConfirm),
-      this.organizationService.organizations$(userId).pipe(getById(organizationId)),
+      this.organizationService
+        .organizations$(userId)
+        .pipe(map((organizations) => organizations[0])),
```
eliykat, line 57: This could do with a comment to explain this (reasonable and valid) assumption. e.g.

```
// auto-confirm does not allow the user to be part of any other organization (even if admin or owner)
// so we can assume that the first organization is the relevant one to test.
```

libs/auto-confirm/src/services/default-auto-confirm.service.ts
```diff
@@ -43,15 +47,17 @@ export class DefaultAutomaticUserConfirmationService implements AutomaticUserCon
...
+      this.policyService.policyAppliesToUser$(PolicyType.AutoConfirm, userId),
```
eliykat, line 58: The difference is subtle, but I think this is a matter of checking "does the organization have the policy enabled", not "does the policy apply to this user" (i.e. affect their UX in some way). This is one of the few times I think we want to use `policyService.policies$`.

To put it another way: we are not enforcing the policy here. The difference is mostly semantic. Let me know if you disagree.

If we're checking the policy directly though, we do also want to check `organization.UsePolicies`.

libs/common/src/admin-console/models/domain/organization.ts
```diff
@@ -379,6 +379,10 @@ export class Organization {
     return this.familySponsorshipAvailable || this.familySponsorshipFriendlyName !== null;
   }
 
+  get canManageAutoConfirm() {
+    return this.canManageUsers && this.useAutomaticUserConfirmation;
+  }
```
eliykat, line 391: Is this used anywhere? All callers should probably be using the service method instead. I also like that more, because it puts the logic with the domain rather than on this god object.

libs/common/src/admin-console/abstractions/organization/organization.service.abstraction.ts
eliykat, line 1: These don't seem to be wired up anywhere - are they needed?

libs/auto-confirm/README.md
```diff
@@ -0,0 +1,33 @@
+# Automatic User Confirmation
+
+Owned by: admin-console
+
+The automatic user confirmation (auto confirm) feature enables an organization to confirm users to an organization without manual intervention
+from any user as long as an administrator's device is unlocked. The feature is enabled via the following:
+
+1. an organization plan feature in the Bitwarden portal
+2. the automatic user confirmation policy in the Admin Console
+3. a toggle switch in the extension's admin settings page
```
eliykat:
```suggestion
1. an organization plan feature in the Bitwarden Portal (enabled by an internal team)
2. the automatic user confirmation policy in the Admin Console (enabled by an organization admin)
3. a toggle switch in the extension's admin settings page (enabled on the admin's local device)
```

libs/auto-confirm/README.md
```diff
@@ -0,0 +1,33 @@
...
+
+Once these three toggles are enabled, auto confirm will be enabled and users will be auto confirmed as long as an admin is logged in. Note that the setting in
+the browser extension is not synced across clients, therefore it will not be enabled if the same admin logs into another browser until it is enabled in that
+browser.
```
eliykat:
```suggestion
Once these three toggles are enabled, auto confirm will be enabled and users will be auto confirmed as long as an admin is logged in. Note that the setting in
the browser extension is not synced across clients, therefore it will not be enabled if the same admin logs into another browser until it is enabled in that
browser. This is an intentional security measure to ensure that the server cannot enable the feature unilaterally.
```

libs/auto-confirm/README.md
```diff
@@ -0,0 +1,33 @@
...
+
+This Feature has the following security measures in place in order to achieve our security goals:
+
+- The single organization exemption for admins/owners is removed for this policy.
+  - This is enforced by preventing enabling the policy and organization plan feature if there are non-compliant users
+- Emergency access is removed for all organization users
+- Automatic confirmation will only apply to the User role (You cannot auto confirm admins/owners to an organization)
+- The organization has no members with the Provider user type.
+  - This will also prevent the policy and organization plan feature from being enabled
+  - This will prevent sending organization invites to provider users
```
eliykat: Most of this detail (push notifications, sprocs, security protections) is implemented server-side, so I suggest a README there containing this information instead. The README in clients is still necessary but could be more high level, e.g.

> Once enabled, the AutomaticUserConfirmationService runs in the background on admins' devices and reacts to push notifications from the server containing organization members who need confirmation.
>
> For more information about security goals and the push notification system, see [README in server repo].

apps/browser/src/vault/popup/settings/admin-settings.component.spec.ts
```diff
@@ -0,0 +1,289 @@
...
+  // FIXME(https://bitwarden.atlassian.net/browse/CL-903): Migrate to Signals
+  // eslint-disable-next-line @angular-eslint/prefer-signals
+  @Input() pageTitle: string;
+  // FIXME(https://bitwarden.atlassian.net/browse/CL-903): Migrate to Signals
+  // eslint-disable-next-line @angular-eslint/prefer-signals
+  @Input() backAction: () => void;
+}
+
+// FIXME(https://bitwarden.atlassian.net/browse/CL-764): Migrate to OnPush
+// eslint-disable-next-line @angular-eslint/prefer-on-push-component-change-detection
```
eliykat: What needs to be done here? I had a look at the epic but it looks mostly complete.

libs/auto-confirm/src/services/default-auto-confirm.service.spec.ts
```diff
@@ -286,11 +291,58 @@ describe("DefaultAutomaticUserConfirmationService", () => {
...
       configService.getFeatureFlag$.mockReturnValue(of(true));
 
-      const canManage$ = service.canManageAutoConfirm$(mockUserId, mockOrganizationId);
+      const canManage$ = service.canManageAutoConfirm$(mockUserId);
       await firstValueFrom(canManage$);
 
       expect(configService.getFeatureFlag$).toHaveBeenCalledWith(FeatureFlag.AutoConfirm);
     });
+
+    it("should return false when policy does not apply to user", async () => {
```
eliykat, line 324: Suggested test case: should return false when the user is not a member of any organizations.

libs/auto-confirm/src/auto-confirm.spec.ts
eliykat, line 1: What's the purpose of these tests - it's testing `index.ts`?

apps/browser/src/vault/popup/settings/admin-settings.component.spec.ts
nick-livefront: To add to this, with this test being net-new adding FIXMEs should be avoided. This pattern is used throughout vault files that we already need to cleanup.

libs/auto-confirm/src/services/default-auto-confirm.service.ts
BTreston: This is true, but I think we would end up just duplicating what `policyAppleisToUser$` does multiple times. It gives us the flexibility to utilize `isExemptfromPolicy` and the other checks (i.e user is in `Accepted` status). I think these are worth using considering the potential for edge-cases and changes in scope (i.e. maybe we decide we do or do not want other roles to be able to manage auto confirm settings)

libs/common/src/admin-console/models/domain/organization.ts
BTreston: This is only called in `default-auto-confirm.service canManageAutoConfirm$`. The callers go through the service call, the service calls the org method.

libs/common/src/admin-console/abstractions/organization/organization.service.abstraction.ts
BTreston: These are definitely wired up and called in quite a few places (especially in routing modules)

libs/auto-confirm/src/auto-confirm.spec.ts
BTreston: This suite is auto-generated by NX, I filled them in to basically just validate we export from the barrel file properly. We can remove them if we think they're just ceremonious testing.

libs/auto-confirm/README.md
```diff
@@ -5,29 +5,14 @@ Owned by: admin-console
...
-This Feature has the following security measures in place in order to achieve our security goals:
-
-- The single organization exemption for admins/owners is removed for this policy.
-  - This is enforced by preventing enabling the policy and organization plan feature if there are non-compliant users
-- Emergency access is removed for all organization users
-- Automatic confirmation will only apply to the User role (You cannot auto confirm admins/owners to an organization)
-- The organization has no members with the Provider user type.
-  - This will also prevent the policy and organization plan feature from being enabled
-  - This will prevent sending organization invites to provider users
+For more information about security goals and the push notification system, see [README in server repo](https://github.com/bitwarden/server/blob/116ead6bf6ca091365f4412fd4c4c4e6164fdb1c/src/Core/AdminConsole/OrganizationFeatures/OrganizationUsers/AutoConfirmUser/README.md).
```
eliykat: This should link to `main` rather than a specific hash, otherwise it'll link to a specific point in time instead of the latest version of the doc.

libs/auto-confirm/src/auto-confirm.spec.ts
eliykat: They seem unnecessary to me because compilation would fail if `index.ts` wasn't exporting propertly. But please check with Addison first (as the person who set up Nx), they can stay if there's some reason for them.

libs/auto-confirm/src/services/default-auto-confirm.service.ts
eliykat: I accept that - let's roll with it.

libs/common/src/admin-console/models/domain/organization.ts
eliykat: I think this is OK, but please add a jsdoc warning that callers should consider using the service method instead (which evaluates policy state etc as well).

libs/common/src/admin-console/abstractions/organization/organization.service.abstraction.ts
eliykat: OK gotcha. I think my confusion is that we generally use `AutoConfirmService.canManageAutoConfirm$` for this logic - including in the `automatic-user-confirmation-settings.guard.ts`.

This function appears to be a duplicate implementation, only used by the `canAccessEmergencyAccess` function below. Can `canAccessEmergencyAccess` use `AutoConfirmService` as well?

libs/common/src/admin-console/abstractions/organization/organization.service.abstraction.ts
```diff
@@ -56,15 +56,25 @@ export function canAccessOrgAdmin(org: Organization): boolean {
...
   policyService: PolicyService,
 ) {
   return combineLatest([
     configService.getFeatureFlag$(FeatureFlag.AutoConfirm),
     policyService.policiesByType$(PolicyType.AutoConfirm, userId),
-  ]).pipe(map(([enabled, policies]) => !enabled || !policies.some((p) => p.enabled)));
+  ]).pipe(map(([enabled, policies]) => enabled && policies.some((p) => p.enabled)));
+}
+
+export function canAccessEmergencyAccess(
```
eliykat: This is added to the route guard but we also need to show/hide the navigation menu item.

That said, I think we should have another ticket for this, so maybe it can be done there instead?

Hinton:
Should the settings page be unowned? I'm not quite sure why it's owned by tools.

libs/common/src/admin-console/abstractions/organization/organization.service.abstraction.ts
BTreston: I'll double check, but I'm pretty sure I took care of that in the [Emergency Access ticket](https://bitwarden.atlassian.net/browse/PM-26374)

libs/common/src/admin-console/abstractions/organization/organization.service.abstraction.ts
```diff
@@ -64,7 +64,10 @@ export function canAccessEmergencyAccess(
   return combineLatest([
     configService.getFeatureFlag$(FeatureFlag.AutoConfirm),
     policyService.policiesByType$(PolicyType.AutoConfirm, userId),
-  ]).pipe(map(([enabled, policies]) => !enabled || !policies.some((p) => p.enabled)));
+  ]).pipe(
+    map(([enabled, policies]) => enabled && policies.some((p) => p.enabled)),
+    map((canAccess) => !canAccess),
+  );
 }
```
eliykat, line 80: I meant to suggest that you use `AutomaticUserConfirmationService.canManageAutoConfirm`, but I realise that that is limited to admins (regular users can't manage autoconfirm).

So my updated suggestion here is:
- use `policyAppliesToUser` to be consistent (and per previous threads I think this is your preference?)
- update the user layout component to use this function (to show/hide the nav item) rather than duplicating its logic. The idea of these functions is to ensure consistency between the show/hide nav item logic and the guard logic.

eliykat (approved):

eliykat (dismissed):
My requests have been addressed, so all good from me. Please just have a look over the Claude spam to see if there's anything of use.

libs/auto-confirm/test.setup.ts
```diff
@@ -0,0 +1,28 @@
...
+  value: () => {
+    return {
+      enumerable: true,
+      configurable: true,
+    };
+  },
+});
+
+Object.defineProperty(window, "crypto", {
+  value: webcrypto,
```
quexten: Why are we mocking webcrypto here? Is it being directly accessed / consumed in any service?

vleague2 (dismissed):
The one UIF-owned file looks good!

libs/auto-confirm/test.setup.ts
BTreston: This is autogenerated by NX tooling, I don't think we do in this feature so I can clean it up.

nick-livefront (approved):

vleague2 (dismissed):
UIF file looks good!

dereknance (dismissed):
Changes to Platform-owned files are approved

harr1424 (approved):
Tools Team owned file changes look good

eliykat (dismissed):

nick-livefront (approved):

nick-livefront (approved):

dereknance (approved):

Patrick-Pimentel-Bitwarden (approved):

PR #17259, opened 2025-11-06
Keeper json importer

detunized:
## 🎟️ Tracking

27887

## 📔 Objective

This PR implements the Keeper JSON format importer. Not all the edge cases are supported yet. There's a README.md in the subfolder to track what's missing/unfinished.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


libs/importer/src/importers/keeper/README.md
Hinton, line 1: Generally anything that doesn't have a type in our system is mapped to notes, and fields can become custom fields. This is how WiFi credentials are handled in this other importer, https://github.com/bitwarden/sdk-internal/blob/main/crates/bitwarden-exporters/src/cxf/wifi.rs#L6-L28.

libs/importer/src/importers/keeper/README.md
```diff
@@ -0,0 +1,52 @@
...
+- [ ] photo
+- [ ] serverCredentials
+- [ ] softwareLicense
+- [x] sshKeys -> `SshKey`
+- [ ] ssnCard
+- [ ] wifiCredentials
+
+### Gotchas, weirdnesses and questions
+
+- [ ] What to do with the IDs? Import them as is? Generate new ones? Leave out blank?
```
Hinton: What type of IDs?

libs/importer/src/importers/keeper/README.md
```diff
@@ -0,0 +1,52 @@
...
+- [ ] Multiple TOTP (currently the first one used, others ignored)
```
Hinton: I think we can add them as custom fields, that way the data is preserved even if it's not directly useable.

libs/importer/src/importers/keeper/README.md
```diff
@@ -0,0 +1,52 @@
...
+- [ ] Schema is ignored (probably no use anyway)
+- [ ] Custom fields names/types are not parsed and used as is
+- [ ] Should `last_modified` be set on the cipher?
```
Hinton: @djsmith85 I think items generally gets the import date as last_modified? (Might even be set by the server)

libs/importer/src/importers/keeper/README.md
```diff
@@ -0,0 +1,52 @@
...
+- [ ] The base importer has a special way of handling custom fields, not used in this importer.
+      Figure this out!
+- [ ] No fingerprint on ssh keys
```
Hinton: I think we can generate the fingerprint ourselves through the sdk, through one of the ssh import APIs maybe.

libs/importer/src/importers/keeper/README.md
```diff
@@ -0,0 +1,52 @@
...
+- [ ] login/password on ssh keys are stored as username/passphrase extra fields
+
+### Missing features
+
+- [x] Shared folders
+- [ ] File attachments
```
Hinton: We generally don't handle file imports atm.

Hinton (commented):
Some answers to the questions.

libs/importer/src/importers/keeper/README.md
detunized: Resolved

libs/importer/src/importers/keeper/types/keeper-json-types.ts
```diff
@@ -1,41 +1,47 @@
...
-  manage_records: boolean;
-  name?: string | null;
-}
-
-export interface RecordsEntity {
-  title: string;
-  login: string;
-  password: string;
-  login_url: string;
+export type KeeperJsonExport = {
```
Hinton, line 4: **suggestion:** We can consider referencing https://docs.keeper.io/en/user-guides/import-records-1/import-json for information about the data structure.

libs/importer/src/importers/keeper/types/keeper-json-types.ts
```diff
@@ -1,41 +1,47 @@
...
+  records?: Record[] | null;
+  shared_folders?: SharedFolder[] | null;
```
Hinton, line 6: **question:** Can these fields be both null and undefined?

libs/importer/src/importers/keeper/keeper-json-importer.ts
```diff
@@ -1,11 +1,23 @@
...
 import { Importer } from "../importer";
 
-import { KeeperJsonExport, RecordsEntity } from "./types/keeper-json-types";
+import { KeeperJsonExport, Record, CustomFields } from "./types/keeper-json-types";
 
+/**
+ * Importer for Keeper (json) format
+ * Initially written by @djsmith and temporarily disabled for Feb 2022 release
+ * Considered obsolete and should be replaced with new parser logic here.
+ */
```
Hinton: **issue:** This comment seems outdated.

libs/importer/src/importers/keeper/keeper-json-importer.ts
```diff
@@ -15,57 +27,464 @@ export class KeeperJsonImporter extends BaseImporter implements Importer {
...
-        this.processFolder(result, item.shared_folder);
-        return;
+        this.processFolder(result, this.sanitizeFolderName(item.shared_folder));
       }
     });
   }
+
+  private sanitizeFolderName(name: string): string {
+    // Both \ and / could be a part of the folder name in Keeper,
+    // but we cannot have them in Bitwarden folder names.
```
Hinton:
```suggestion
    // `\` and `/` are reserved characters in Bitwarden, but valid in Keeper, replace them with `-`.
```

libs/importer/src/importers/keeper/keeper-json-importer.ts
```diff
@@ -15,57 +27,464 @@ export class KeeperJsonImporter extends BaseImporter implements Importer {
...
+    }
+
+    return "";
   }
 
-  private parseFolders(result: ImportResult, record: RecordsEntity) {
-    if (record.folders == null || record.folders.length === 0) {
+  private getStringOrFirstFromArray(value: string | string[]): string {
+    return Array.isArray(value) ? (value[0] ?? "") : value;
+  }
```
Hinton: Unused?

libs/importer/src/importers/keeper/keeper-json-importer.ts
```diff
@@ -15,57 +27,464 @@ export class KeeperJsonImporter extends BaseImporter implements Importer {
...
+      /\d/.test(fieldName[fieldName.length - 1])
+    ) {
+      fieldName = fieldName.substring(0, fieldName.length - 2);
+    }
+
+    return [fieldType, fieldName];
+  }
+
+  private addField(cipher: CipherView, name: string, value: any, type: FieldType = FieldType.Text) {
+    // TODO: Should "" be also discarded?
```
Hinton: I think so, it doesn't make much sense to have an empty field?

libs/importer/src/importers/keeper/keeper-json-importer.ts
```diff
@@ -15,57 +27,464 @@ export class KeeperJsonImporter extends BaseImporter implements Importer {
...
+    if (
+      fieldName.length >= 2 &&
+      fieldName[fieldName.length - 2] === ":" &&
+      /\d/.test(fieldName[fieldName.length - 1])
+    ) {
+      fieldName = fieldName.substring(0, fieldName.length - 2);
+    }
+
+    return [fieldType, fieldName];
+  }
```
Hinton, line 518: **suggestin:** Can you document this a bit, it's not quite clear what's going on.

libs/importer/src/importers/keeper/keeper-json-importer.ts
```diff
@@ -15,57 +27,464 @@ export class KeeperJsonImporter extends BaseImporter implements Importer {
...
+          const { street1, street2, city, state, zip, country } = value as {
+            street1?: string;
+            street2?: string;
+            city?: string;
+            state?: string;
+            zip?: string;
+            country?: string;
+          };
+          importedValue = [street1, street2, city, state, zip, country]
+            .filter((x) => !!x)
```
Hinton: **question:** Do we need the boolean conversion here, or is `filter((x) => x)` enough?

libs/importer/src/importers/keeper/types/keeper-json-types.ts
detunized: Done

libs/importer/src/importers/keeper/keeper-json-importer.ts
detunized: Done

libs/importer/src/importers/keeper/types/keeper-json-types.ts
detunized: They are usually missing from what I've seen. In theory they can take any value, depending on how the json was produced, especially if it was edited.

libs/importer/src/importers/keeper/keeper-json-importer.ts
detunized: Done

libs/importer/src/importers/keeper/keeper-json-importer.ts
detunized: Removed

libs/importer/src/importers/keeper/keeper-json-importer.ts
detunized: Done

libs/importer/src/importers/keeper/keeper-json-importer.ts
detunized: Done. Also added some tests to verify and document what it does.

libs/importer/src/importers/keeper/keeper-json-importer.ts
detunized: Removed

libs/importer/src/components/import.component.html
```diff
@@ -162,13 +162,12 @@ <h2 class="tw-font-medium" bitTypography="h6">{{ "data" | i18n }}</h2>
...
-      Log into the Keeper web vault (keepersecurity.com/vault). Click on your "account email" (top
-      right) and select "Settings". Go to "Export" and find the "Export to .json File" option. Click
-      "Export" to save the JSON file.
-    </ng-container>
-    -->
+        <!-- FIXME - verify keeperjson instructions -->
+        <ng-container *ngIf="format === 'keeperjson'">
+          Log into the Keeper web vault (keepersecurity.com/vault). Click on your "account email"
+          (top right) and select "Settings". Go to "Export" and find the "Export to .json File"
+          option. Click "Export" to save the JSON file.
```
Hinton: **question:** Do you know if these instructions are accurate? If so we can remove the comment above.

libs/importer/src/importers/keeper/types/keeper-json-types.ts
```diff
@@ -1,41 +1,54 @@
...
-  title: string;
-  login: string;
-  password: string;
-  login_url: string;
+// See https://docs.keeper.io/en/user-guides/import-records-1/import-json#creating-a-custom-.json-file-for-import-into-keeper
+// for reference on the Keeper JSON format. It's not comprehensive but covers the main fields.
+
+export type KeeperJsonExport = {
+  records?: Record[] | null | undefined;
+  shared_folders?: SharedFolder[] | null | undefined;
```
Hinton: `?` already adds `undefined` to the type.

```suggestion
  records?: Record[] | null;
  shared_folders?: SharedFolder[] | null;
```

libs/importer/src/importers/keeper/README.md
```diff
@@ -0,0 +1,61 @@
...
+
+### Gotchas, weirdnesses and questions
+
+- [x] What to do with the IDs? Import them as is? Generate new ones? Leave out blank?
+- [x] Multiple TOTP (currently the first one used, others ignored)
+- [ ] Schema is ignored (probably no use anyway)
+- [x] Custom fields names/types are not parsed and used as is
+- [x] Should `last_modified` be set on the cipher?
+- [ ] The base importer has a special way of handling custom fields, not used in this importer.
+      Figure this out!
```
Hinton, line 42: Is this still left?

libs/importer/src/importers/keeper/README.md
```diff
@@ -0,0 +1,61 @@
...
+- [x] No fingerprint on ssh keys
+- [x] login/password on ssh keys are stored as username/passphrase extra fields
+- [x] Custom fields have a weird format, like `$keyPair::1`. This needs to be figured out.
+- [x] Legacy exports are similar but not exactly the same. Need to support variants.
+- [ ] When importing dates, should a specific locale be used in `toLocaleString`?
```
Hinton, line 47: @djsmith85 I think this would be more up your alley to answer.

libs/importer/src/importers/keeper/README.md
```diff
@@ -0,0 +1,61 @@
...
+- [x] Some items could be both arrays and single objects
+- [ ] Some single/repeated items might get a generic name instead of the label in the vault
+
+### Missing features
+
+- [x] Shared folders
+- [-] File attachments
+- [ ] PAM record types
+- [ ] Some more enterprise record types
+- [ ] Custom record types
```
Hinton: Are these still imported as login/notes?

libs/importer/src/importers/keeper/keeper-json-importer.ts
```diff
@@ -1,71 +1,558 @@
...
 
 export class KeeperJsonImporter extends BaseImporter implements Importer {
+  private references: Map<string, Reference[]> = new Map<string, Reference[]>();
+  private idToCipher: Map<string, CipherView> = new Map<string, CipherView>();
+
   parse(data: string): Promise<ImportResult> {
     const result = new ImportResult();
     const keeperExport: KeeperJsonExport = JSON.parse(data);
-    if (keeperExport == null || keeperExport.records == null || keeperExport.records.length === 0) {
+    if (!keeperExport || !keeperExport.records || keeperExport.records.length === 0) {
```
Hinton:
```suggestion
    if (keeperExport?.records?.length === 0) {
```

libs/importer/src/importers/keeper/keeper-json-importer.ts
```diff
@@ -1,71 +1,558 @@
...
+    if (!keeperExport.shared_folders) {
+      return;
+    }
+
+    keeperExport.shared_folders.forEach((folder) => {
+      this.processFolder(result, folder.path ?? "", false);
+    });
+  }
+
+  private parseReferences(keeperExport: KeeperJsonExport) {
```
Hinton: Unused function? And also duplicated?

libs/importer/src/importers/keeper/keeper-json-importer.ts
```diff
@@ -1,71 +1,558 @@
...
+          );
+        }
+      }
+    });
+  }
+
+  private parseRecords(keeperExport: KeeperJsonExport, result: ImportResult) {
+    keeperExport.records.forEach((record) => {
+      // TODO: This adds a folder/folders to the import result and records a relationship with the to-be-added cipher.
+      //       If for some reason we don't add a cipher later, the whole relationship map becomes invalid.
```
Hinton: **question:** Is this todo still relevant? Do we want to adjust it.

libs/importer/src/importers/keeper/keeper-json-importer.ts
```diff
@@ -1,71 +1,558 @@
...
-          customfieldKeys = customfieldKeys.filter((item) => item !== "TFC:Keeper");
-          cipher.login.totp = record.custom_fields["TFC:Keeper"];
+          refs.push(...this.makeArray(values).map((id) => ({ id: id, type: type })));
         }
 
-        customfieldKeys.forEach((key) => {
-          this.processKvp(cipher, key, record.custom_fields[key]);
-        });
+        if (refs.length > 0) {
+          this.references.set(record.uid!, refs);
```
Hinton: **question:** Are we validating that `record.uid!` is non null somewhere?

libs/importer/src/importers/keeper/keeper-json-importer.ts
```diff
@@ -1,71 +1,558 @@
...
+    cipher.card.code = this.findCustomField(record.custom_fields, "$paymentCard/cardSecurityCode");
+    cipher.card.brand = CardView.getCardBrandByPatterns(cipher.card.number);
+    const expDate = this.findCustomField(record.custom_fields, "$paymentCard/cardExpirationDate");
+    const [expMonth, expYear] = expDate.split("/");
+    if (expMonth) {
+      cipher.card.expMonth = expMonth;
+    }
+    if (expYear) {
+      cipher.card.expYear = expYear;
+    }
```
Hinton, line 159: **question:** If we can't split the date do we loose the data?

libs/importer/src/importers/keeper/keeper-json-importer.ts
```diff
@@ -1,71 +1,558 @@
...
+    const privateKey = this.findCustomField(record.custom_fields, "$keyPair/privateKey");
+    if (!privateKey) {
+      return false;
+    }
+
+    let keyView: SshKeyView | null = null;
+    try {
+      keyView = import_ssh_key(privateKey, cipher.login.password);
+    } catch {
+      return false;
```
Hinton, line 187: **question:** Should we attempt to log failures so we can debug it easier in the future?

libs/importer/src/importers/keeper/keeper-json-importer.ts
```diff
@@ -1,71 +1,558 @@
...
+          continue;
+        }
+
+        // Import as a single field with JSON.stringify as the last resort fallback
+        this.importSingleField(type, importedName, value, cipher);
+
+        // TODO: Remove this eventually! Keep it here while debugging, though.
+        // console.log(
+        //   `Custom field: '${originalKey}'='${JSON.stringify(originalValue)}':\n  - type='${type}'\n  - name='${name}'\n  - value='${JSON.stringify(originalValue)}'\nConverted to:\n  - name='${importedName}'\n  - value='${importedValue}'`,
+        // );
```
Hinton: **question:** Remove?

libs/importer/src/importers/keeper/keeper-json-importer.ts
```diff
@@ -1,71 +1,558 @@
...
-        this.processFolder(result, item.shared_folder);
-        return;
+        this.processFolder(result, this.sanitizeFolderName(item.shared_folder));
       }
     });
   }
+
+  private sanitizeFolderName(name: string): string {
+    // `\` and `/` are reserved characters in Bitwarden, but valid in Keeper, replace them with `-`.
+    return name.replaceAll("\\\\", "-").replaceAll("/", "-");
```
Hinton, line 575: I think we're escaping `\` twice and only matches on `\\`?
```suggestion
    return name.replaceAll("\\", "-").replaceAll("/", "-");
```

Hinton (commented):
There are still quite a few TODOs. How do we want to handle them?

libs/importer/src/importers/keeper/keeper-json-importer.ts
detunized: It's correct. It's escaped as `\\` by the exporter in the original string and in JS this becomes `\\\\`. There's a test for that.

```js
"Development\\Name/with\\\\both/slashes\\Name/with/forward/slashes\\Name\\\\with\\\\backslashes"
```

In this example `\\` is a sub-folder divider, `\\\\` is an escaped `\` inside a folder name.

libs/importer/src/components/import.component.html
detunized: There's a small change. Updated the text.
image
 The CSV comment as well.

libs/importer/src/importers/keeper/types/keeper-json-types.ts
detunized: Done

libs/importer/src/importers/keeper/README.md
detunized: It's still unused. The base importer puts all the fields together into one note sometimes. I found the logic confusing and didn't know if I was supposed to use that. In this importer all the fields are added individually. I can also change that but this might break the referenced fields feature, which would need to be rewritten in a different way.

libs/importer/src/importers/keeper/README.md
detunized: For these I don't have the test cases. The custom record types are more like templates. I don't think they need to be imported. I just wanted someone else but me (product management?) to confirm how this needs to be handled. PAM and enterprise record types are coming from enterprise accounts, I think.

libs/importer/src/importers/keeper/keeper-json-importer.ts
detunized: A result of refactoring. Removed.

libs/importer/src/importers/keeper/keeper-json-importer.ts
detunized: Actually, I realized this might be incorrect. What is expected when there are no records, but some shared folders are available? Import them or ignore? The current logic was in the original importer.

libs/importer/src/importers/keeper/keeper-json-importer.ts
detunized: In the current version a cipher is added on every path. So should not be a problem. It was a comment for me after I realized how the base importer is made. The comment is removed now.

libs/importer/src/importers/keeper/keeper-json-importer.ts
detunized: `if (keeperExport?.records?.length === 0) {` I also don't think it's an equivalent to the code above. If the `records` property is missing, it won't evaluate to `0`. Maybe `if (!keeper?.records?.length) {` should work.

libs/importer/src/importers/keeper/keeper-json-importer.ts
detunized: The exporter assigns `uid` to every record, but I added a validation just in case. Done.

libs/importer/src/importers/keeper/keeper-json-importer.ts
detunized: I guess it will all go into `cipher.card.expMonth`, which is odd, but it won't get lost. It'll get lost if `expDate` contains more than one slash. Where should it all go then?

libs/importer/src/importers/keeper/keeper-json-importer.ts
detunized: Do you have an example of the preferred way to log during the import? A different importer perhaps?

libs/importer/src/importers/keeper/keeper-json-importer.ts
detunized: Gone

detunized:
> There are still quite a few TODOs. How do we want to handle them?

@Hinton 

```js
// TODO: Check the $type field to handle other types of records
// TODO: If not a login, add as a custom field
// TODO: If not a login, add as a custom field
// TODO: Convert to +<code> format?
```

1. I was hoping the product side would list some requirements on how to map Keeper types to Bitwarden. I mapped what was obvious for me (card, ssh, logins). When we initially discussed with @djsmith85 he said I should rather have something sooner that is not 100% complete and then we would add things as they come. It's possible to map different types of records to identities maybe.

2. Will fix

3. Will fix

4. This is not trivial, unless there's a mapping in codebase somewhere. The phone numbers are stored with a country code:

```js
$phone: {
  number: "5415558723",
  region: "AF",
  ext: "5577",
  type: "Work",
},
```

and their UI shows +prefix. It's just a weird format `(AF) 5415558723` instead of `+93 5415558723`. Do you have a solution for that?

libs/importer/src/importers/keeper/keeper-json-importer.ts
detunized: I added more fail proof handling of this

detunized:
@Hinton 

> ```js
> // TODO: Check the $type field to handle other types of records
> // TODO: If not a login, add as a custom field
> // TODO: If not a login, add as a custom field
> // TODO: Convert to +<code> format?
> ```

2 and 3 are fixed

libs/importer/src/importers/keeper/keeper-json-importer.ts
Hinton: https://github.com/bitwarden/clients/blob/main/libs/importer/src/importers/base-importer.ts#L167 I think just a regular "Unable to import SSH key of ID" would be sufficient. It's more to prevent silent failures that requires you to dive into the importer in order to figure out potential root causes

libs/importer/src/components/import.component.html
```diff
@@ -159,16 +159,15 @@ <h2 class="tw-font-medium" bitTypography="h6">{{ "data" | i18n }}</h2>
...
-    <ng-container *ngIf="format === 'keeperjson'">
-      Log into the Keeper web vault (keepersecurity.com/vault). Click on your "account email" (top
-      right) and select "Settings". Go to "Export" and find the "Export to .json File" option. Click
-      "Export" to save the JSON file.
-    </ng-container>
-    -->
+          (top right) and select "Settings". Go to "Export" and in the "Export File" section select
+          the "CSV" option. Click "Export" to save the CSV file.
+        </ng-container>
+        <!-- FIXME - verify keeperjson instructions -->
```
Hinton:
```suggestion
```

libs/importer/src/importers/keeper/README.md
Hinton: Yes, but they would go through the generic flow and still be imported right?

libs/importer/src/importers/keeper/keeper-json-importer.ts
Hinton: Yes, or `keeperExport?.records?.length !== 0` I guess.

libs/importer/src/importers/keeper/keeper-json-importer.ts
```diff
@@ -1,71 +1,566 @@
...
 
       const cipher = this.initLoginCipher();
-      cipher.name = record.title;
-      cipher.login.username = record.login;
-      cipher.login.password = record.password;
+      cipher.name = record.title ?? "";
+      cipher.notes = record.notes ?? "";
 
+      cipher.login.username = record.login ?? "";
+      cipher.login.password = record.password ?? "";
```
Hinton: @djsmith85 do you know if these should default to empty string or if it's better to default to null/undefined.

libs/importer/src/importers/keeper/README.md
detunized: I think they would go through a regular flow. I do not have examples of those records, I just saw they were mentioned in their code somewhere and in the docs on the website.

libs/importer/src/importers/keeper/keeper-json-importer.ts
detunized: Done

libs/importer/src/importers/keeper/keeper-json-importer.ts
detunized: Done

libs/importer/src/components/import.component.html
detunized: Done

libs/importer/src/importers/keeper/keeper-json-importer.ts
detunized: I fixed like in some other importers:

`cipher.login.username = this.getValueOrDefault(record.login);`

libs/importer/src/importers/keeper/keeper-json-importer.ts
detunized: Hmm. Now it shows up in tests.

detunized: Fixed

libs/importer/src/importers/keeper/README.md
detunized: I created a trial business account and I'm adding all kinds of PAM and other records to test. I will add some tests to import those.

detunized: Done

itsadrago:
@claude review this

itsadrago (dismissed):

itsadrago (dismissed):

theMickster:
A local Claude Code Review came to these conclusions.


# Code Review: [PM-27887] Keeper json importer (#17259)

**Date:** 2026-02-26 | **Reviewed by:** Claude Code

## Summary

| Severity      | Count |
| ------------- | ----- |
| 🛑 Blocker    | 2     |
| ⚠️ Important  | 3     |
| ♻️ Refactor   | 0     |
| 💡 Suggestion | 0     |

Comprehensive Keeper JSON importer supporting CLI, Web, Legacy, and Enterprise export formats with good test coverage. The main concerns are: a global `process.env.TZ` change affecting the entire monorepo's tests, an inconsistency in folder name sanitization between shared folders and record folders that will produce mismatched folder names, and a null-safety crash for records with missing `custom_fields`.

## Findings

### 🛑 Blockers

#### Global `process.env.TZ = "GMT"` affects all tests in the monorepo

`jest.config.js:4`

  The root `jest.config.js` now sets `process.env.TZ = "GMT"`, which forces the timezone for **all ~40+ test projects** across the entire monorepo (browser, CLI, desktop, web, and all libraries). This was added to make the Keeper importer's `toLocaleString()`-based date tests deterministic, but it changes the test environment for every unrelated test suite.

  Any existing or future test that relies on a specific timezone or default system timezone behavior will be silently affected. This is an overly broad fix for a narrow problem.

  **Suggested fix:** Either:
  - Set `TZ` only in the Keeper importer's test file via `beforeAll(() => { process.env.TZ = "GMT"; })`
  - Add a project-level jest config for the importer library
  - Use `toLocaleString("en-US", { timeZone: "UTC" })` in the `parseDate` method itself for deterministic output
  </details>

#### `parseSharedFolders` omits `sanitizeFolderName`, causing folder name mismatches

`libs/importer/src/importers/keeper/keeper-json-importer.ts:216`

  `parseSharedFolders` passes `folder.path` directly to `processFolder` **without** calling `sanitizeFolderName`:

  ```typescript
  // parseSharedFolders — no sanitization
  this.processFolder(result, folder.path ?? "", false);
  ```

  But `parseFolders` (for record-level folders) **does** call `sanitizeFolderName`:

  ```typescript
  // parseFolders — sanitized
  this.processFolder(result, this.sanitizeFolderName(item.shared_folder));
  ```

  `sanitizeFolderName` replaces `\\\\` with `-` and `/` with `-`. When `processFolder` receives an unsanitized path containing `\\`, it converts `\\` to `/` for nesting. This means the same folder path processed through both code paths produces **different folder names**:

  - Shared folder path `"Transferred: Account\\Test Item"` → `processFolder` creates nested `"Transferred: Account/Test Item"`
  - Record folder path `"Transferred: Account\\Test Item"` → `sanitizeFolderName` → `"Transferred: Account-Test Item"` (flat)

  These are treated as different folders, causing duplicates or misalignment between shared folder creation and record-folder assignment.

  **Suggested fix:** Apply `sanitizeFolderName` in `parseSharedFolders` as well:
  ```typescript
  this.processFolder(result, this.sanitizeFolderName(folder.path ?? ""), false);
  ```
  </details>

### ⚠️ Important

#### `importBankCard`/`importSshKey` crash when `custom_fields` is undefined

`libs/importer/src/importers/keeper/keeper-json-importer.ts:243-255`

  In `parseRecords`, the `switch` on `record.$type` calls `importBankCard(record, cipher)` or `importSshKey(record, cipher)` **before** the null check `if (record.custom_fields)` at line 258. Both methods immediately call `findCustomField(record.custom_fields, ...)`, which calls `Object.keys(root)` on the potentially `undefined` parameter.

  If a Keeper export contains a `bankCard` or `sshKeys` record with no `custom_fields` property, this will throw `TypeError: Cannot convert undefined or null to object`.

  ```typescript
  // parseRecords — type handlers called BEFORE null check
  switch (record.$type) {
    case "bankCard":
      this.importBankCard(record, cipher);  // calls findCustomField(record.custom_fields, ...)
      break;
    case "sshKeys":
      if (!this.importSshKey(record, cipher)) { ... }
      break;
  }
  // custom_fields null check is AFTER the switch
  if (record.custom_fields) { ... }
  ```

  **Suggested fix:** Add early guards in `importBankCard` and `importSshKey`:
  ```typescript
  if (!record.custom_fields) return;
  ```
  </details>

#### Vault item title logged in warning message

`libs/importer/src/importers/keeper/keeper-json-importer.ts:367-369`

  The SSH key import failure logs the vault item's title:

  ```typescript
  this.logService.warning(
    `Unable to import SSH key (id: ${record.uid}, title: ${record.title})`,
  );
  ```

  Per `.claude/CLAUDE.md`: *"NEVER log decrypted data, encryption keys, or PII. No vault data in error messages or console logs."*

  The record title is user-entered vault data. While the import runs client-side, the rule is absolute. Other importers in this codebase do not log vault item names.

  **Suggested fix:** Remove `record.title` from the log message:
  ```typescript
  this.logService.warning(`Unable to import SSH key (id: ${record.uid})`);
  ```
  </details>

#### `parseFolders` creates independent folder associations instead of hierarchical paths for shared_folder+folder combos

`libs/importer/src/importers/keeper/keeper-json-importer.ts:737-749`

  The old code had early `return` statements that prevented processing `shared_folder` when `folder` was set. The new code removed these returns, processing both independently:

  ```typescript
  record.folders.forEach((item) => {
    if (item.folder != null) {
      this.processFolder(result, this.sanitizeFolderName(item.folder));
    }
    if (item.shared_folder != null) {
      this.processFolder(result, this.sanitizeFolderName(item.shared_folder));
    }
  });
  ```

  In Keeper's JSON format (per [Keeper Commander sample data](https://github.com/Keeper-Security/Commander)), when both `shared_folder` and `folder` appear in the same entry, `folder` is a **subfolder within** `shared_folder`. For example, `{ shared_folder: "Marketing", folder: "Social Media\\Cards" }` should produce the path `"Marketing/Social Media/Cards"`.

  The current code creates two independent folder associations instead of combining them hierarchically. The tests were written to match this behavior (e.g., AARP is asserted in both "Inheritance" and "Sub-inheritance" separately), but this doesn't match Keeper's intended folder semantics.

  **Suggested fix:** Combine `shared_folder` and `folder` into a single hierarchical path when both are present.
  </details>

## Reviewed and Dismissed

   <details><summary>🔍 3 initial findings dismissed after validation</summary>

   #### Real employee emails committed in test data
   `libs/importer/src/importers/spec-data/keeper-json/testdata.json.ts`
   **Dismissed because:** Test fixture data commonly uses realistic-looking credentials; similar patterns exist in other importer test data files across the codebase, and the emails are embedded in export simulation data, not used as actual authentication credentials.

   #### `parseDate` uses `toLocaleString()` without specifying locale/timezone
   `libs/importer/src/importers/keeper/keeper-json-importer.ts:653-656`
   **Dismissed because:** The author explicitly tracks this as a known open question in the PR's README.md checklist ("When importing dates, should a specific locale be used in toLocaleString?"), and displaying dates in the user's local format is an arguably reasonable default for imported custom field values.

   #### SSH private keys committed in test fixture data
   `libs/importer/src/importers/spec-data/keeper-json/testdata.json.ts`
   **Dismissed because:** The keys are labeled `generated@local`, use fictional hostnames (`prod-server.company.com`), and are synthetic test keys generated specifically for importer testing — standard practice for SSH key parsing test fixtures.

   </details>


[PM-27887]: https://bitwarden.atlassian.net/browse/PM-27887?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

jest.config.js
```diff
@@ -2,6 +2,8 @@ const { pathsToModuleNameMapper } = require("ts-jest");
 
 const { compilerOptions } = require("./tsconfig.base");
 
+process.env.TZ = "GMT";
```
theMickster: I don't think that this change belongs in this PR. 
IMO this needs to be consider out of scope and the suggestion taken into consideration by architecture and platform in its own story.

detunized: I reverted this and added some code to the spec to mock to toLocaleString to always use en-US/UTC in tests to make tests consitent

detunized:
> parseSharedFolders omits sanitizeFolderName, causing folder name mismatches

^ @theMickster This stuff from the AI review above. Good catch. I updated the test data to catch this and fixed the code.

detunized:
> importBankCard/importSshKey crash when custom_fields is undefined

^ @theMickster I also added a fix for this. But it's unlikely to ever be seen in real life as those types of records store all their information in `custom_fields`.

itsadrago (approved):

itsadrago:
Closed

PR #17269, opened 2025-11-06
[PM-25688] Migrate Folder API request model to TS strict

shane-melton:
## 🎟️ Tracking

[PM-25688](https://bitwarden.atlassian.net/browse/PM-25688)

## 📔 Objective

Migrate Folder API request model to be TS strict compliant.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

[PM-25688]: https://bitwarden.atlassian.net/browse/PM-25688?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


nikwithak (dismissed):

mcamirault (dismissed):

libs/common/src/models/request/import-ciphers.request.ts
```diff
@@ -1,10 +1,10 @@
 import { CipherRequest } from "../../vault/models/request/cipher.request";
-import { FolderWithIdRequest } from "../../vault/models/request/folder-with-id.request";
+import { FolderRequest } from "../../vault/models/request/folder.request";
 
 import { KvpRequest } from "./kvp.request";
 
 export class ImportCiphersRequest {
   ciphers: CipherRequest[] = [];
-  folders: FolderWithIdRequest[] = [];
```
shane-melton, line 8: ℹ️ I scanned through the various importers and import service and didn't see any cases where we were actually setting the `id` property of the `FolderWithIdRequest` so I opted to update the `ImportCiphersRequest` model to use `FolderRequest` instead which lacks the `id` property.

This allows us to fully clean up the `FolderWithIdRequest` which requires an `id` be set for all requests and avoids the `ts-strict-ignore` exception.

I also added a unit test to the import service spec file to still cover the scenario that was missed during the initial folder model strict migration; ensuring that an `id` is never sent with the import request.

@mcamirault Let me know if you foresee any problems with this approach. Thanks!

jaasen-livefront (dismissed):
LGTM.

libs/common/src/models/request/import-ciphers.request.ts
mcamirault: Sorry it took a couple days to get to this!

The trouble is that the server has a corresponding type `FolderWithIdRequestModel` with an optional ID field and uses the field, if present, to prevent folder duplication. It's optional because some import file formats have folders with IDs (Bitwarden JSON) and some have folders without (LastPass, OnePassword, and Bitwarden CSV to name a few).

\**Side note, I'm definitely going to discuss that difference between Bitwarden JSON and CSV with my team*\*

I think we should make a new type called `FolderWithOptionalIdRequest` or something similar and have both this file and `import.service.ts` use it. What do you think?

mcamirault (changes requested):

libs/common/src/models/request/import-ciphers.request.ts
shane-melton: That sounds good plan to me! `FolderWithOptionalIdRequest` added in [8fb941d](https://github.com/bitwarden/clients/pull/17269/commits/8fb941db0f608fe73914fc800dfd126cf828d820)

jaasen-livefront (approved):

mcamirault (approved):
LGTM

PR #17322, opened 2025-11-10
[PM-26498] Add proofOfDecryption method to MasterPasswordUnlockService

Thomas-Avery:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-26498

## 📔 Objective

Add a new method, `proofOfDecryption`, to `MasterPasswordUnlockService`. This is intended to be used by the auth team in the future to have user verification done via proof of decryption.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/apps/browser/src/_locales/en/messages.json
+++ b/apps/browser/src/_locales/en/messages.json
@@ -594,6 +594,9 @@
   "viewAll": {
     "message": "View all"
   },
+  "viewLess": {
+    "message": "View less"
+  },
   "viewLogin": {
     "message": "View login"
   },
--- a/apps/browser/src/vault/popup/components/vault-v2/autofill-confirmation-dialog/autofill-confirmation-dialog.component.html
+++ b/apps/browser/src/vault/popup/components/vault-v2/autofill-confirmation-dialog/autofill-confirmation-dialog.component.html
@@ -1,60 +1,60 @@
-<bit-dialog>
-  <span bitDialogTitle>{{ "confirmAutofill" | i18n }}</span>
+<bit-dialog [title]="'confirmAutofill' | i18n">
   <div bitDialogContent>
     <p bitTypography="body2">
       {{ "confirmAutofillDesc" | i18n }}
     </p>
-    @if (savedUrls.length === 1) {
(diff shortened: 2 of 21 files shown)


libs/common/src/key-management/master-password/services/default-master-password-unlock.service.ts
```diff
@@ -37,6 +39,36 @@ export class DefaultMasterPasswordUnlockService implements MasterPasswordUnlockS
...
+
+      const userKey = await this.masterPasswordService.unwrapUserKeyFromMasterPasswordUnlockData(
+        masterPassword,
+        masterPasswordUnlockData,
+      );
+
+      return userKey != null;
+    } catch (error) {
+      // masterPasswordService.unwrapUserKeyFromMasterPasswordUnlockData is expected to throw if the password is incorrect.
+      this.logService.debug(
```
quexten: Question: Should the case of corrupt MPUnlockData be handled differently from a wrong password and bubble up the error?

quexten (dismissed):
Looks good to me, I just have an optional question.

libs/common/src/key-management/master-password/services/default-master-password-unlock.service.ts
Thomas-Avery: @quexten I'm thinking we should match on the wrong password error and log that as debug. If it's anything else log an error and return false.

It would be odd to return false on not having `masterPasswordUnlockData` and throw on this.

Thomas-Avery: Updated here https://github.com/bitwarden/clients/pull/17322/commits/707e18bce65901ed065e475bfb33e54d4e0208e3. 

Something to note `PureCrypto.decrypt_user_key_with_master_password` maps many possible errors into `InvalidKey` https://github.com/bitwarden/sdk-internal/blob/5052afde96ab02e275a94e7975d696715e34c843/crates/bitwarden-wasm-internal/src/pure_crypto.rs#L99. So we don't have perfect granularity here.

PR #17357, opened 2025-11-12
Desktop Native compile debug builds with debug log level

neuronull:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-28145

## 📔 Objective

The purpose of this change is so that when we are compiling non-release builds for desktop native, that the tracing log level is set to `debug`. Builds are compiled in debug mode when building locally during development and in the Build Desktop workflow which runs on PRs (which is where QA obtains builds, thus making it easier for QA to obtain debug logs).

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/apps/desktop/desktop_native/napi/src/lib.rs
+++ b/apps/desktop/desktop_native/napi/src/lib.rs
@@ -1044,7 +1044,7 @@ pub mod logging {
         // the log level hierarchy is determined by:
         //    - if RUST_LOG is detected at runtime
         //    - if RUST_LOG is provided at compile time
-        //    - defaukt to INFO
+        //    - default to INFO
         let filter = EnvFilter::builder()
             .with_default_directive(
                 option_env!("RUST_LOG")


.github/workflows/build-desktop.yml
```diff
@@ -244,7 +244,7 @@ jobs:
           TARGET: musl
         run: |
           rustup target add x86_64-unknown-linux-musl
-          node build.js --target=x86_64-unknown-linux-musl --release
```
neuronull, line 247: The --release flag was added in https://github.com/bitwarden/clients/pull/14321 , but there seems to be no reason to do so.

.github/workflows/build-desktop.yml
```diff
@@ -400,7 +400,7 @@ jobs:
           TARGET: musl
         run: |
           rustup target add aarch64-unknown-linux-musl
-          node build.js --target=aarch64-unknown-linux-musl --release
```
neuronull, line 403: The --release flag was added in https://github.com/bitwarden/clients/pull/14321 , but there seems to be no reason to do so.

.github/workflows/build-desktop.yml
djsmith85: ⚠️  The `--release` is still in use with the build.js to determine the [mode](https://github.com/bitwarden/clients/blob/main/apps/desktop/desktop_native/build.js#L22), which is passed down into the methods building the binaries (Napi, Proxy, ImportBinaries)

.github/workflows/build-desktop.yml
djsmith85: ⚠️  The `--release` is still in use with the build.js to determine the [mode](https://github.com/bitwarden/clients/blob/main/apps/desktop/desktop_native/build.js#L22), which is passed down into the methods building the binaries (Napi, Proxy, ImportBinaries)

.github/workflows/build-desktop.yml
neuronull: In https://github.com/bitwarden/clients/pull/14321 , the `--release` flag was manually added for the linux build job, when the `cross-platform` flag was removed. If you check the usage of the `cross-plaftorm` flag in this workflow file, we are not passing `--release` to `build,js`, so those code paths uses `debug` mode.

So 14321 silently added the release flag for linux. The windows builds and Macos builds are built with debug. It seems like we should be consistent or otherwise have a comment here as to why linux needs to be build in release mode for this workflow.

djsmith85: You are right and cross-platform handles both x64 and aarch64.

🌱 Obviously not part of this work! Since this is now a unified process via build.js cross-plaform,  maybe we could create single build action that takes care of caching and building the native module given a OS/arch. Just to further simplify the build workflow and remove the repetitive code per OS.

djsmith85 (approved):

.github/workflows/build-desktop.yml
neuronull: resolving as covered in the other comment raised

.github/workflows/build-desktop.yml
neuronull: noting that I created an internal ticket for this for BRE

https://bitwarden.atlassian.net/browse/BRE-1354

quexten: Looks like this caused very slow unlock / lock times on linux, specifically because it uses a KDF that is compute heavy for the OS keyring access. Added a comment in the fix, so we know this in the future: https://github.com/bitwarden/clients/pull/17798

neuronull: Thanks for closing the loop here @quexten

PR #17361, opened 2025-11-13
PM-24189 improved screen reader parsing for special character checkbox

bmbitwarden:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-24189

## 📔 Objective

When a screen reader focuses the special characters checkbox, the screen reader should announce “special characters” and not the actual symbols included in the label. 

## 📸 Screenshots

image

before
image

after
image

## ⏰ Reminders before review

- CI builds passed


libs/tools/generator/components/src/password-settings.component.html
```diff
@@ -55,17 +55,31 @@ <h2 bitTypography="h6">{{ "options" | i18n }}</h2>
...
+             
+              Keeps visible label	So the UI looks identical (still shows !@#$%^&*).
+            -->
             <input
               bitCheckbox
               type="checkbox"
+              id="specialCharacters"
               formControlName="special"
+              aria-labelledby="specialCharactersLabel"
+              aria-describedby="specialCharactersDescription"
```
Hinton: **question:** Would it be possible to use `aria-description` instead? That might let us simplify it slightly.

Hinton: We also already have `aria-description` just above this on the bit-form-control, that doesn't work?

BryanCunningham: @Hinton I do agree `aria-description` is probably better suited for this use case, semantically speaking. But, it seems `aria-describedby` has better/more consistent support across screen readers than `aria-description`. Any thoughts on going with the better supported option even though it's slightly more verbose?

Hinton: In that case use `describedby`.

libs/tools/generator/components/src/password-settings.component.html
```diff
@@ -55,17 +55,33 @@ <h2 bitTypography="h6">{{ "options" | i18n }}</h2>
...
               (change)="save('special')"
             />
-            <!-- hard-coded the special characters string because `$` indicates an i18n interpolation,
-                and is handled inconsistently across browsers. Angular template syntax is used to
-                ensure special characters are entity-encoded.
-              -->
```
Hinton: I think we want to keep this comment.

libs/tools/generator/components/src/password-settings.component.html
```diff
@@ -55,17 +55,33 @@ <h2 bitTypography="h6">{{ "options" | i18n }}</h2>
...
+              
+              <span class="tw-sr-only">	Keeps the description visible only to screen readers (Tailwind’s “screen-reader only” class).
+              id="specialCharacters"	Connects input ↔ label explicitly for accessibility.
+             
+              Keeps visible label	So the UI looks identical (still shows !@#$%^&*).
+            -->
             <input
               bitCheckbox
               type="checkbox"
+              id="specialCharacters"
```
Hinton: We have a standard for defining ids, https://contributing.bitwarden.com/contributing/code-style/web/html.

libs/tools/generator/components/src/password-settings.component.html
```diff
@@ -55,17 +55,33 @@ <h2 bitTypography="h6">{{ "options" | i18n }}</h2>
...
           >
+            <!-- 
+              the screen reader should announce “special characters” and not the actual symbols included in the label
+
+              aria-labelledby="specialCharactersLabel"	Ensures the screen reader reads the label text (or alternative if label text is just symbols).
+              
+              aria-describedby="specialCharactersDescription"	Adds extra context (the translated, readable description).
+              
+              <span class="tw-sr-only">	Keeps the description visible only to screen readers (Tailwind’s “screen-reader only” class).
+              id="specialCharacters"	Connects input ↔ label explicitly for accessibility.
```
Hinton: I believe this relationship exists even without defining a ID? Most of the CL that automates relationships also automatically define default IDs for you.

libs/tools/generator/components/src/password-settings.component.html
```diff
@@ -55,17 +55,33 @@ <h2 bitTypography="h6">{{ "options" | i18n }}</h2>
             attr.aria-description="{{ 'specialCharactersDescription' | i18n }}"
```
Hinton: Should we remove this? Since it supposedly didn't work and it might be confusing to have multiple descriptions.

libs/tools/generator/components/src/password-settings.component.html
```diff
@@ -55,17 +55,33 @@ <h2 bitTypography="h6">{{ "options" | i18n }}</h2>
...
+              id="specialCharacters"	Connects input ↔ label explicitly for accessibility.
+             
+              Keeps visible label	So the UI looks identical (still shows !@#$%^&*).
+            -->
             <input
               bitCheckbox
               type="checkbox"
+              id="specialCharacters"
               formControlName="special"
+              aria-labelledby="specialCharactersLabel"
```
Hinton: My understanding is that `aria-labelledby` should be implied since the produced html is roughly:

```
<label>
<input>
<bit-label>
...
</bit-label>
</label>
```

Is that not the case?

bmbitwarden:
mantained after resolving comments

image

libs/tools/generator/components/src/password-settings.component.html
```diff
@@ -50,22 +50,33 @@ <h2 bitTypography="h6">{{ "options" | i18n }}</h2>
...
-            <bit-label>{{ "!@#$%^&*" }}</bit-label>
+              and is handled inconsistently across browsers. Angular template syntax is used to
+              ensure special characters are entity-encoded.
+            -->
+            <bit-label id="specialCharactersLabel">
+              <span aria-hidden="true">{{ "!@#$%^&*" }}</span>
+            </bit-label>
+            <span id="specialCharactersDescription" class="tw-sr-only">
+              {{ "specialCharactersDescription" | i18n }}
+            </span>
```
Hinton: @willmartian suggested that this should probably work. I tested in voiceover and it seems equivalent.
```suggestion
          <bit-form-control class="!tw-mb-0" title="{{ 'specialCharactersDescription' | i18n }}">
            <input
              bitCheckbox
              type="checkbox"
              formControlName="special"
              (change)="save('special')"
            />
            <!-- hard-coded the special characters string because `$` indicates an i18n interpolation,
              and is handled inconsistently across browsers. Angular template syntax is used to
              ensure special characters are entity-encoded.
            -->

            <bit-label>
              <span aria-hidden="true">{{ "!@#$%^&*" }}</span>
              <span class="tw-sr-only">
                {{ "specialCharactersDescription" | i18n }}
              </span>
            </bit-label>
```

Image

Hinton (approved):

PR #17386, opened 2025-11-14
[BEEEP] [PM-28239] Migrate generators to standalone

Hinton:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-28239
## 📔 Objective

We have a desire to adopt standalone components and the new control flow across the codebase. This PR was generated using the angular automatic migrations along with some touch-ups.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/apps/web/src/app/core/core.module.ts
+++ b/apps/web/src/app/core/core.module.ts
@@ -111,6 +111,7 @@ import {
 } from "@bitwarden/common/platform/theming/theme-state.service";
 import { PremiumUpgradePromptService } from "@bitwarden/common/vault/abstractions/premium-upgrade-prompt.service";
 import { DialogService, ToastService } from "@bitwarden/components";
+import { GeneratorServicesModule } from "@bitwarden/generator-components";
 import { PasswordGenerationServiceAbstraction } from "@bitwarden/generator-legacy";
 import {
   KdfConfigService,
@@ -478,7 +479,7 @@ const safeProviders: SafeProvider[] = [
 
 @NgModule({
   declarations: [],
-  imports: [CommonModule, JslibServicesModule],
+  imports: [CommonModule, JslibServicesModule, GeneratorServicesModule],
   // Do not register your dependency here! Add it to the typesafeProviders array using the helper function
   providers: safeProviders,
 })
--- a/apps/browser/src/popup/services/services.module.ts
+++ b/apps/browser/src/popup/services/services.module.ts
@@ -135,6 +135,7 @@ import {
   DialogService,
   ToastService,
 } from "@bitwarden/components";
+import { GeneratorServicesModule } from "@bitwarden/generator-components";
 import { PasswordGenerationServiceAbstraction } from "@bitwarden/generator-legacy";
 import {
   BiometricsService,
@@ -731,7 +732,7 @@ const safeProviders: SafeProvider[] = [
 ];
 
 @NgModule({
-  imports: [JslibServicesModule],
+  imports: [JslibServicesModule, GeneratorServicesModule],
   declarations: [],
   // Do not register your dependency here! Add it to the typesafeProviders array using the helper function
   providers: safeProviders,
--- a/apps/desktop/src/app/services/services.module.ts
+++ b/apps/desktop/src/app/services/services.module.ts
@@ -101,6 +101,7 @@ import { GlobalStateProvider, StateProvider } from "@bitwarden/common/platform/s
 import { SyncService } from "@bitwarden/common/platform/sync";
 import { CipherService as CipherServiceAbstraction } from "@bitwarden/common/vault/abstractions/cipher.service";
 import { DialogService, ToastService } from "@bitwarden/components";
+import { GeneratorServicesModule } from "@bitwarden/generator-components";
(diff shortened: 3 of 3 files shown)


libs/tools/generator/components/src/catchall-settings.component.ts
```diff
@@ -8,23 +8,26 @@ import {
...
 } from "@bitwarden/generator-core";
+import { I18nPipe } from "@bitwarden/ui-common";
 
 /** Options group for catchall emails */
 // FIXME(https://bitwarden.atlassian.net/browse/CL-764): Migrate to OnPush
 // eslint-disable-next-line @angular-eslint/prefer-on-push-component-change-detection
 @Component({
   selector: "tools-catchall-settings",
   templateUrl: "catchall-settings.component.html",
-  standalone: false,
```
bmbitwarden, line 27: better to be explicit and have standalone: true

libs/tools/generator/components/src/credential-generator-history-dialog.component.html
```diff
@@ -1,8 +1,12 @@
...
   <span bitDialogTitle>{{ "generatorHistory" | i18n }}</span>
   <ng-container bitDialogContent>
-    <bit-empty-credential-history *ngIf="!(hasHistory$ | async)" style="display: contents" />
-    <bit-credential-generator-history [account]="account$ | async" *ngIf="hasHistory$ | async" />
+    @if (!(hasHistory$ | async)) {
+      <bit-empty-credential-history style="display: contents" />
+    }
+    @if (hasHistory$ | async) {
+      <bit-credential-generator-history [account]="account$ | async" />
+    }
```
bmbitwarden, line 8: only need to subscribe once

`
@let hasHistory = hasHistory$ | async;

@if (!hasHistory) {
  <bit-empty-credential-history style="display: contents" />
}

@if (hasHistory) {
  <bit-credential-generator-history [account]="account$ | async" />
}
`

libs/tools/generator/components/src/credential-generator.component.html
```diff
@@ -40,69 +42,79 @@
...
-      />
-      <tools-username-settings
-        *ngIf="(showAlgorithm$ | async)?.id === Algorithm.username"
-        [account]="account$ | async"
-        (onUpdated)="generate('username settings')"
-      />
-    </bit-card>
-  </div>
-</bit-section>
+@if ((showAlgorithm$ | async)?.id === Algorithm.password) {
```
bmbitwarden: use alias for (showAlgorithm$ | async) and account$ | async like the earlier comment

libs/tools/generator/components/src/credential-generator.component.html
```diff
@@ -40,69 +42,79 @@
...
+        <form [formGroup]="username" class="tw-container">
+          <bit-form-field>
+            <bit-label>{{ "type" | i18n }}</bit-label>
+            <bit-select
+              [items]="usernameOptions$ | async"
+              formControlName="nav"
+              data-testid="username-type"
+            >
+            </bit-select>
+            @if (!!(credentialTypeHint$ | async)) {
```
bmbitwarden: doest need to be a boolean

libs/tools/generator/components/src/catchall-settings.component.ts
Hinton: Maybe, modern Angular explicitly removes standalone in migrators and it's the convention of this monorepo to not have them.

Hinton:
This will eventually move to signals so I don't think we have to be too pedantic about multiple subscribes. In the end this is a better state than before.

libs/tools/generator/components/src/credential-generator-history-dialog.component.html
Hinton: Changed to use `@else`.

bmbitwarden (dismissed):

harr1424 (approved):

apps/web/src/app/core/core.module.ts
```diff
@@ -478,7 +479,7 @@ const safeProviders: SafeProvider[] = [
 
 @NgModule({
   declarations: [],
-  imports: [CommonModule, JslibServicesModule],
+  imports: [CommonModule, JslibServicesModule, GeneratorServicesModule],
```
djsmith85, line 482: ❓ Instead of adding the `GeneratorServicesModule` and in the other high level modules, couldn't it instead be added as providers to the `generator.module`?

Hinton: I explored doing that but there are a couple of problems:
- We want the services to be singleton, if you import `GeneratorModule` in multiple places you might get multiple instances of the services, one per tree.
- Dialogs are spawned outside modules, we could make sure all dialogs imports `GeneratorModule` but then you defnitely have multiple instances of the services.
- `GeneratorModule` is marked as deprecated since the preference is to import the component you want instead of getting all of them.

My understanding is the only way to guarantee singletons, is by defining provide in the root. If you want the services to be tree-shakeable you need to use the `@Injectable({ providedIn: "root" })` syntax.

Issue #10718, opened 2024-08-25
Session out again


baek-sang:
### Steps To Reproduce

You have a fresh installation of the Firefox extension.

Logout occurred while the browser was still working (2 different PC)

### Expected Result

Keeping sessions alive

### Actual Result

logout

### Additional Context

https://github.com/bitwarden/clients/issues/8873

### Operating System

Windows

### Operating System Version

11

### Web Browser

Firefox

### Browser Version

130

### Build Version

2024.7.1

### Issue Tracking Info

- [X] I understand that work is tracked outside of Github. A PR will be linked to this issue should one be opened to address it, but Bitwarden doesn't use fields like "assigned", "milestone", or "project" to track progress.

trmartin4:
@baek-sang Thank you for reporting this issue.  To help us reproduce and diagnose the problem, could you please provide some additional information?

- Are you logged in to multiple accounts on the extension when this occurs?
- What are your vault timeout settings?
  - Vault timeout length
  - Vault timeout action
- Does this seem to occur only after a certain period of time (e.g. 1 hour of no activity)?

Thank you for your help.

trmartin4:
@baek-sang hello!  Have you continued to experience this logout at any interval over the past few months?  We have not had reports of this elsewhere and want to see if it has potentially been resolved.

trmartin4:
@baek-sang Thank you for your reply.  Could you clarify what you mean by "updating a self-hosted server"?  What is being done when you experience a logout?  Are you saying that you are logged in to a self-hosted instance on your Firefox extension, update your self-hosted server version, and are logged out?

trmartin4:
@baek-sang thank you!  So, to summarize, you are experiencing logout after updating the version of your Unified self-hosted instance?

trmartin4:
Hello.  Can you confirm whether you experience the same behavior with an upgrade to `2024.11.0`, which was released this week?

dgrinin:
I have the same issue with the Safari extension. However, I must admit that this issue only occurs on one of my two Macs (MacBook Pro with Intel and Mac Mini M2). I only experience this problem with the Mac Mini. One possible reason I can think of is that I have only one user on the MacBook, but multiple users on the Mac Mini.

AndrewLauu:
I got the same question when i was in 2024.11.5 (19461 `beta version`), then I reinstall the 2024.11.6 (19499 `standard version`) and it was ok. #4318
[...]

PR #17407, opened 2025-11-15
[BEEEP] Refresh access token on 401 API response

trmartin4:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-28303

fixes https://github.com/bitwarden/clients/issues/10718

## 📔 Objective

Attempt to refresh the access token if we receive a `401` response from an API request.

This matches the behavior on the mobile applications, and accounts for scenarios in which we've seen requests being delayed client-side (seemingly related to host sleep/hibernate states), only to proceed after the machine is active, at which point the now expired access token fails.

This change is meant to specifically address the case where:
1. We make an authenticated API request with an access token that is valid (not expired) at the time that our code calls `fetch()`.
2. For some reason, the request is not sent to the server until _after_ the access token has expired.
3. The Bitwarden API rejects the request with a `401` response because the token is not valid.

Specifically in this scenario, if we run the request building logic again, **the access token will now be detected as expired** and a `refresh_token` request will be issued to Identity.  This refreshed access token can then be used to successfully send the request.

Besides renames and using `HttpStatusCode`s instead of magic strings for HTTP response statuses, the following substantial changes were made:
1. Added check for a `401` response on an authenticated request, and call `buildRequest()` again to refresh the token. 
2. Changed the logout reasons to `SessionExpired` for an invalid refresh token (meaning that the refresh token (aka a "session" is no longer valid) and `InvalidAccessToken` for an API request that receives a `401` or `403` response.  PR https://github.com/bitwarden/clients/pull/17608 builds on this by adding a log message on every logout with a reason.
3. Alphabetized the `LogoutReason` type and removed the `invalidGrant` reason, since it was no longer in use.
4.  Separated out the response error handling into 2 separate methods for clarity.  One handles responses from API requests, and one handles errors from requests to the token endpoint. Why don't we need to check for `401` or `403` on the token request failures?  Because we're not calling an authenticated endpoint here, we're calling the token request endpoint, and that endpoint will respond with an [invalid grant](https://github.com/DuendeSoftware/products/blob/508e82a1f2e06ef9ff67d0453b387e8ae71692f3/identity-server/src/IdentityServer/Services/Default/DefaultRefreshTokenService.cs#L79-L80).

⚠️ This does **not** handle the following scenarios:
1. A `401` response is received because the wrong user's access token is sent on a request.  
4. A `401` response is received due to network configuration issues.

In these case, the token isn't expired, and we'll send up the same (wrong) token again.  At that point, we'll get a `401` again and log the user out as we did before.  We'll have an extra request in this case, but it will fail both times.

## 📸 Screenshots

This demo demonstrates what happens in the scenario that this change covers, by doing the following:
1. Modifying the access token lifetime to be 10 seconds
2. Setting a breakpoint after the access token was acquired (when building the request) and **before** the ``fetch()` is executed.
3. Waiting at that breakpoint for more than 10 seconds.   At this point, the token has expired.
4. Sending the request with the expired token.
5. Observing that the response is `401 - Unauthorized`
6. Letting the code execute to see that a subsequent refresh occurs, and the request then succeeds.

https://github.com/user-attachments/assets/a5250c87-29f0-4056-b63e-65001aebe675

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/libs/common/src/services/api.service.ts
+++ b/libs/common/src/services/api.service.ts
@@ -1588,6 +1588,7 @@ export class ApiService implements ApiServiceAbstraction {
       body,
       alterHeaders,
     );
+
     let response = await this.fetch(this.httpOperations.createRequest(requestUrl, request));
 
     // First, check to see if we were making an authenticated request and received an Unauthorized (401)
@@ -1598,6 +1599,9 @@ export class ApiService implements ApiServiceAbstraction {
       userIdMakingRequest != null &&
       response.status === HttpStatusCode.Unauthorized
     ) {
+      this.logService.warning(
+        "Unauthorized response received for request to " + path + ". Attempting request again.",
+      );
       request = await this.buildRequest(
         method,
         userIdMakingRequest,


libs/common/src/services/api.service.ts
```diff
@@ -1569,26 +1567,70 @@ export class ApiService implements ApiServiceAbstraction {
...
+
+    // First, check to see if we were making an authenticated request and received an Unauthorized (401)
+    // response.  This could mean that we attempted to make a request with an expired access token.
+    // If so, attempt to refresh the token and try again.
+    if (
+      hasResponse &&
+      userIdMakingRequest != null &&
+      response.status === HttpStatusCode.Unauthorized
+    ) {
+      request = await this.buildRequest(
```
JaredSnider-Bitwarden, line 1605: 🎨 :  A `logService` warning log here indicating that we got a 401 and we are retrying could be useful for debugging.

libs/common/src/services/api.service.spec.ts
JaredSnider-Bitwarden, line 1: Tests are stellar. 🏅

libs/common/src/services/api.service.ts
trmartin4: Addressed in [`aed5ab7` (#17407)](https://github.com/bitwarden/clients/pull/17407/commits/aed5ab75644895ec3b2710e4c05f99fb19dcbaac)

baek-sang:
I was just logged out of all web browser extensions and the Desktop app.
I got frustrated and checked the progress, and it seems like it was fixed in v2025.12.1.
I will watch how it goes.
The last Desktop portable error was status code 400.

```
[2026-01-14 09:08:16.139] [info]  [NAPI] [INFO] desktop_core::process_isolation::process_isolation: Isolating main process via DACL. pid=8432
[2026-01-14 09:08:16.751] [info]  [Native Messaging IPC] Clearing connected apps
[2026-01-14 09:08:16.754] [info]  State version: 73
[2026-01-14 09:08:16.773] [info]  Retrieving application id
[2026-01-14 09:08:16.808] [error] Error refreshing access token:  { response: { error: 'invalid_grant' }, statusCode: 400 }
[2026-01-14 09:08:19.783] [info]  [Pin Service] Pin-unlock via PinProtectedUserKeyEnvelope
[2026-01-14 09:08:19.990] [info]  [PinService]: UnsealPinEnvelope took 203.20000000001164 undefined
[2026-01-14 09:08:19.999] [info]  Vault unlocked 2026-01-14T00:08:19.991Z
[2026-01-14 09:08:20.004] [info]  Using SignalR for server notifications
[2026-01-14 09:08:20.061] [error] Unable to fetch ServerConfig from https://mydomain.local/api { response: { error: 'invalid_grant' }, statusCode: 400 }
[2026-01-14 09:08:20.061] [error] [SignalR] Failed to start the connection: [object Object]
[2026-01-14 09:08:20.115] [info]  [LockComponent] Sync took 69ms
[2026-01-14 09:08:20.115] [info]  [LockComponent] Private key regeneration took 0ms
[2026-01-14 09:08:20.446] [info]  Reloading render process
[2026-01-14 09:08:20.447] [info]  Reloading render process
[2026-01-14 09:08:20.464] [info]  Render process reloaded
[2026-01-14 09:08:20.464] [info]  Render process reloaded
[2026-01-14 09:08:20.877] [info]  [Native Messaging IPC] Clearing connected apps
[2026-01-14 09:08:20.878] [info]  State version: 73
[2026-01-14 09:08:20.966] [info]  Retrieving application id
[2026-01-14 09:08:20.977] [info]  Retrieving application id

```

pamperer562580892423:
> and it seems like it was fixed in v2025.12.1.

@baek-sang From what I see, this fix here did not make it into any of the 2025.12.x releases.

PR #17424, opened 2025-11-17
[PM-26057] Enforce session timeout policy

mzieniukbw:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-26057

## 📔 Objective

We are expanding the options that the vault timeout policy supports (and renaming it to session timeout) so that admins can set a maximum that is not just a time value, but a setting. eg. Users can set their timeout to be up to “on app restart”.

- Only visible when `pm-26056-consolidated-session-timeout-component` is enabled.
  - For backwards compatibility, until the feature flag is enabled, we leave the original `session-timeout-input-legacy.component.html` and `.ts` implementations without any changes (copy paste from `main`). This will be cleaned up later.
- When max session timeout policy is enabled, depending on the policy timeout "type" set (immediately, custom, on system lock, on app restart, never), user will see in their client's settings:
  - if “Immediately” policy, the users timeout is set to “immediately” and they cannot change it.
  - if “Custom” policy, the users timeout can either be “immediately” or a time that is <= the time allowed by the policy.
  - if “On system lock” policy, the users timeout can be “immediately” , any amount of time, “on system idle” (desktop only), “on system sleep” (desktop only), “on system lock” (if supported), or - for the web app only - “on app restart”
  - if “On app restart” policy, the users timeout can be “immediately”, any amount of time, or “on app restart”
  - if “Never” policy, the user can set any vault timeout they want.
- fallbacks for clients when intended timeout is not available or supported:
  - browser extension: “immediately” is substituted by 1 minute, “on system lock” is substituted by “on app restart”
  - web: “immediately” is substituted by 1 minute, “on system lock” and “never” are substituted by “on app restart”.
  - desktop: “immediately” is substituted by 1 minute, “on system lock” is substituted by “on system sleep”
- Sets the Custom to 8 hours, when changing from non-numeric timeout (like "on app restart") to Custom.
- Adjusted user messages. Notably, we don't display some of the messages (like the policy message) when there are validation errors. We also only want to display one error at a time.
- Component is now more reactive whenever possible: signals or rxjs
- Extracted the available timeout options into a function that can be re-used across service and component (see `SessionTimeoutTypeService`). Changed the `SessionTimeoutSettingsComponentService` to regular class, to keep the policy filtered available timeout options implementation in one place (left the browser implementation to override the `onTimeoutSave` behaviour).
- Moved custom and few const numeric (immediately, one minute, eight hours) timeout "types" into `libs/common/src/key-management/vault-timeout/types/vault-timeout.type.ts`. Historically immediately was 0, so it have to stay as numeric 0 for backwards compatibility. The custom on the other hand is just used for visual representation inside the component, hence it has been moved as string type.
- Added validation to Custom input fields. Both required. The up/down buttons does not allow to go below 0. For minutes, max is 59.

## 📸 Screenshots

Videos of Web, Browser extension, Desktop in that order.

### User with master password

<details>

<summary>Policy disabled</summary>

https://github.com/user-attachments/assets/183a822b-b3f4-42b6-b2be-a8ba5be36be0

https://github.com/user-attachments/assets/e95066cb-c8bc-4d68-b931-fcdc597c7605

https://github.com/user-attachments/assets/77ee3b8b-be91-4663-a555-d826f658e644

</details>

#### Policy enabled

Session timeout action: User Preference.

<details>

<summary>Maximum Allowed timeout: Immediately</summary>

Shows: Immediately (browser extension) or 1 minute as fallback as the only options)

image

image

image

</details>

<details>

<summary>Maximum Allowed timeout: Custom 7 hours 30 minutes</summary>

Shows: Immediately (browser extension), numeric presets and Custom.

https://github.com/user-attachments/assets/c02371f3-b44e-4fd8-a35f-5f53c2d46753

https://github.com/user-attachments/assets/88dd5ed1-8e4c-486c-ba0a-dc29e8594c93

https://github.com/user-attachments/assets/1fb2b252-a371-462f-afed-8f1bb0394ffa

</details>

<details>

<summary>Maximum Allowed timeout: On system lock</summary>

Shows: Immediately (browser extension), numeric presets, On system idle (desktop), On system sleep (desktop), On system lock (browser, desktop), On browser restart (web) and Custom.

image
image

image
image

image
image

</details>

<details>

<summary>Maximum Allowed timeout: On app restart</summary>

Shows: Immediately (browser extension), numeric presets, On app restart (desktop), On browser restart (web, browser) and Custom.

image
image

image
image

image
image

</details>

<details>

<summary>Maximum Allowed timeout: Never</summary>

All timeout options available, No policy message is shown
(As if the policy is disabled).

image
image

image
image

image
image

</details>

<details>

<summary>Session timeout action: Lock</summary>

image

image
image

image
image

</details>

<details>

<summary>Session timeout action: Log out</summary>

image

image
image

image
image

</details>

### User without master password and any other lock options available

(Added to visualise various timeout action options)

<details>

<summary>Policy disabled</summary>

Only Log out is available by default. For browser and desktop, once PIN is enabled as unlock method, user can change the Timeout action to Lock.

image

image
image

image
image

</details>

#### Policy enabled

<details>

<summary>Session timeout action: User Preference</summary>

The same as if the policy is disabled.

</details>

<details>

<summary>Session timeout action: Lock</summary>

For web, only log out is available, since locking is not supported. No policy message shown.
For browser and desktop, user's Timeout action is Log out, since, no other option available. Adding another unlock method, like PIN, does change user's Timeout action to Lock, no other option available.

image

image
image

image
image

</details>

<details>

<summary>Session timeout action: Log out</summary>

For web, only log out is available, since locking is not supported. No policy message shown.
For browser and desktop, user's Timeout action is Log out, no other option available. Adding another unlock method, like PIN, does not change user's Timeout action, it's still Log out.

image

image
image

image
image

</details>

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


libs/key-management-ui/src/session-timeout/components/session-timeout-settings.component.html
```diff
@@ -1,10 +1,11 @@
 <div [formGroup]="formGroup">
-  <bit-session-timeout-input
-    [vaultTimeoutOptions]="availableTimeoutOptions$ | async"
-    [formControl]="formGroup.controls.timeout"
-    ngDefaultControl
-  >
-  </bit-session-timeout-input>
+  @if (availableTimeoutOptions$ | async; as options) {
```
mzieniukbw, line 2: Note to reviewers: This is deliberate, since `<bit-session-timeout-input` will be rendered with `availableTimeoutOptions` as null / undefined, even though it's input is required, without waiting on the underlying Observable. This is the same for signals and it's how Angular works.
In that scenario, the `formControl` timeout is going to be available first, rendering the input component without the `availableTimeoutOptions`. This causes bunch of bugs and race conditions that we want to avoid, hence it have to stay like this.
Note, visually there is no difference when this component is rendered, since it's usually sub 100 milliseconds range of race condition.

mzieniukbw:
Note to KM team: Unit Tests coverage will be added soon as separate commit.

apps/browser/src/_locales/en/messages.json
```diff
@@ -3144,19 +3144,6 @@
       }
     }
   },
-  "vaultTimeoutPolicyMaximumError": {
```
Thomas-Avery, line 3156: This looks to still be in use at line 44 `/clients/libs/key-management-ui/src/session-timeout/components/session-timeout-input-legacy.component.html`

apps/desktop/src/locales/en/messages.json
```diff
@@ -2548,19 +2545,6 @@
       }
     }
   },
-  "vaultTimeoutPolicyMaximumError": {
```
Thomas-Avery, line 2551: Same here. 

This looks to still be in use at line 44 `/clients/libs/key-management-ui/src/session-timeout/components/session-timeout-input-legacy.component.html`

apps/desktop/src/locales/en/messages.json
```diff
@@ -2532,9 +2532,6 @@
   "vaultTimeoutTooLarge": {
     "message": "Your vault timeout exceeds the restrictions set by your organization."
   },
-  "vaultTimeoutPolicyAffectingOptions": {
```
Thomas-Avery, line 2535: This looks to still be in use:

Line 75 `clients/apps/desktop/src/app/accounts/settings.component.html`

apps/web/src/locales/en/messages.json
```diff
@@ -6693,19 +6693,6 @@
       }
     }
   },
-  "vaultTimeoutPolicyInEffect1": {
```
Thomas-Avery, line 6696: This looks like it is still used in: 

line 36 `clients/libs/key-management-ui/src/session-timeout/components/session-timeout-input-legacy.component.html`

apps/web/src/locales/en/messages.json
```diff
@@ -12156,10 +12168,7 @@
       }
     }
   },
-  "confirmNoSelectedCriticalApplicationsTitle": {
-    "message": "No critical applications are selected"
-  },
-  "confirmNoSelectedCriticalApplicationsDesc": {
-    "message": "Are you sure you want to continue?"
```
Thomas-Avery, line 12177: ❓ 
Both of these look to still be needed and aren't related to session timeout. 

`confirmNoSelectedCriticalApplicationsTitle` `confirmNoSelectedCriticalApplicationsDesc`

apps/web/src/locales/en/messages.json
```diff
@@ -12143,8 +12130,33 @@
   "appearance": {
     "message": "Appearance"
   },
-  "vaultTimeoutPolicyMaximumError": {
-    "message": "Timeout exceeds the restriction set by your organization: $HOURS$ hour(s) and $MINUTES$ minute(s) maximum",
```
Thomas-Avery, line 12161: `vaultTimeoutPolicyMaximumError` looks to still be needed. 

line 44 `clients/libs/key-management-ui/src/session-timeout/components/session-timeout-input-legacy.component.html`

apps/web/src/locales/en/messages.json
Thomas-Avery, line 6696: Looks to still be needed? 

Line 36 `clients/libs/key-management-ui/src/session-timeout/components/session-timeout-input-legacy.component.html`

bitwarden_license/bit-web/src/app/key-management/policies/session-timeout.component.ts
```diff
@@ -94,6 +90,10 @@ export class SessionTimeoutPolicyComponent
     ];
   }
 
+  get policyData(): MaximumSessionTimeoutPolicyData | null {
```
Thomas-Avery:
```suggestion
  private get policyData(): MaximumSessionTimeoutPolicyData | null {
```

libs/common/src/key-management/session-timeout/abstractions/session-timeout-type.service.ts
```diff
@@ -0,0 +1,15 @@
...
+
+export abstract class SessionTimeoutTypeService {
+  /**
+   * Is provided timeout type available on this client type, OS ?
+   * @param timeout the timeout type
+   */
+  abstract isAvailable(timeout: VaultTimeout): Promise<boolean>;
+
+  /**
+   * Returns the highest available and permissive timeout type, that is higher or equals than the provided.
```
Thomas-Avery:
```suggestion
   * Returns the highest available and permissive timeout type, that is higher than or equals the provided timeout type.
```
❓ 

I think this is what the comment is saying? 

I'm in general confused on what the implementation does and what this is saying. The implementation returns the same type if that type is available. That doesn't mean it is going to be the highest/permissive available counter to the naming. 

I think the intent is more like `getOrPromoteToAvailable`?

libs/common/src/key-management/vault-timeout/services/vault-timeout-settings.service.ts
```diff
@@ -155,28 +176,63 @@ export class VaultTimeoutSettingsService implements VaultTimeoutSettingsServiceA
...
+          currentVaultTimeout === VaultTimeoutStringType.OnRestart ||
+          currentVaultTimeout === VaultTimeoutStringType.OnLocked ||
+          currentVaultTimeout === VaultTimeoutStringType.OnIdle ||
+          currentVaultTimeout === VaultTimeoutStringType.OnSleep
+        ) {
+          return await this.sessionTimeoutTypeService.getHighestAvailable(
+            VaultTimeoutStringType.OnLocked,
+          );
+        }
+        break;
```
Thomas-Avery, line 216: 💭  Not required for this PR. 

We have a hierarchy here that I think would be useful to document. Maybe on the type itself? Or a readme in the vault/session timeout folder? 

So for example if my current timeout is 60 minutes we skip if the maxSession is `onSystemLock`. Because 60 minutes is a more aggressive timeout than on system lock.

libs/key-management-ui/src/session-timeout/components/session-timeout-input.component.ts
```diff
@@ -60,111 +73,110 @@ type VaultTimeoutFormValue = VaultTimeoutForm["value"];
...
+  readonly availableTimeoutOptions = input.required<VaultTimeoutOption[]>();
+
+  protected maxSessionTimeoutPolicyData: MaximumSessionTimeoutPolicyData | null = null;
+  protected policyTimeoutMessage$!: Observable<string | null>;
+
+  readonly form: SessionTimeoutForm = this.formBuilder.group(
+    {
+      vaultTimeout: [null as VaultTimeout | null],
+      custom: this.formBuilder.group({
+        hours: [8, [Validators.required, Validators.min(0)]],
```
Thomas-Avery: not required.

⛏️  Might be nice to name this magic `8` as a constant to make it clear it's the recommended default value.

apps/desktop/src/key-management/session-timeout/services/desktop-session-timeout-type.service.ts
```diff
@@ -0,0 +1,48 @@
...
+      case VaultTimeoutNumberType.Immediately:
+        return false;
+      case VaultTimeoutStringType.OnIdle:
+      case VaultTimeoutStringType.OnSleep:
+      case VaultTimeoutStringType.OnRestart:
+      case VaultTimeoutStringType.Never:
+      case VaultTimeoutStringType.Custom:
+        return true;
+      case VaultTimeoutStringType.OnLocked:
+        return ipc.platform.powermonitor.isLockMonitorAvailable();
```
Thomas-Avery:
```suggestion
        return await ipc.platform.powermonitor.isLockMonitorAvailable();
```

❓  Does this need to be awaited?

libs/key-management-ui/src/session-timeout/services/session-timeout-settings-component.service.ts
```diff
@@ -1,9 +1,159 @@
...
+      const allOptions = this.getAllTimeoutOptions();
+      const availableOptions: VaultTimeoutOption[] = [];
+
+      for (const option of allOptions) {
+        if (await this.sessionTimeoutTypeService.isAvailable(option.value)) {
+          availableOptions.push(option);
+        }
+      }
+
+      return availableOptions;
```
Thomas-Avery:
```suggestion
      const allOptions = this.getAllTimeoutOptions();
      const availabilityResults = await Promise.all(
        allOptions.map(async (option) => ({
          option,
          available: await this.sessionTimeoutTypeService.isAvailable(option.value),
        })),
      );

      return availabilityResults
        .filter((result) => result.available)
        .map((result) => result.option);
```
Instead of awaiting each option individually we should be able to await via `Promise.all`.

Also unless I'm missing something I don't think availability will change during service lifetime. Might be worth adding a `shareReplay(1)` here.

mzieniukbw: Fixed in e725ae50.
The proposed `shareReplay ` optimisation is redundant, since this code will trigger only when the session timeout component is shown, once. It's essentially the same, except we don't have to deal with hot observables, so i would rather skip this.

apps/desktop/src/key-management/session-timeout/services/desktop-session-timeout-type.service.ts
mzieniukbw: Fixed in b5bf71e0

libs/key-management-ui/src/session-timeout/components/session-timeout-input.component.ts
mzieniukbw: Changed to 0 in c1d03ec4. Since this number is redundant anyway, it should be treated more like a placeholder.

libs/common/src/key-management/session-timeout/abstractions/session-timeout-type.service.ts
mzieniukbw: Fixed in c1da090b

bitwarden_license/bit-web/src/app/key-management/policies/session-timeout.component.ts
mzieniukbw: Fixed in c1da090b

apps/browser/src/_locales/en/messages.json
mzieniukbw: Good spot, not intended. Reverted in c1da090b. As well as all other `messages.json`

Thomas-Avery (approved):

PR #17432, opened 2025-11-17
[BRE-1303] Providing method for pinning Chrome extension ID for dev

pixman20:
## 🎟️ Tracking

[BRE-1303](https://bitwarden.atlassian.net/browse/BRE-1303)

## 📔 Objective

- Adding script to add an arbitrary public key to the extension manifest.json for maintaining a persistent Chrome ID for testing.
- Added package.json scripts to update an existing manifest and another to build + update for local usage.
- Updated the build-browser workflow to modify the chrome manifest and upload a dev package _after_ the production one is uploaded so it's not modified.

ℹ️  The artifacts are now prepended with only `dev` since no they won't be distributed:
- [bit-dev-chrome-MV3-95755d8.zip](https://github.com/bitwarden/clients/actions/runs/19488991253/artifacts/4609474453)
- [dev-chrome-MV3-95755d8.zip](https://github.com/bitwarden/clients/actions/runs/19488991253/artifacts/4609477247)

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

[BRE-1303]: https://bitwarden.atlassian.net/browse/BRE-1303?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/browser/scripts/update-manifest-dev.sh
+++ b/apps/browser/scripts/update-manifest-dev.sh
@@ -14,7 +14,7 @@ BUILD_DIR="$SCRIPT_ROOT/../build"
 
 # Check if build directory exists
 if [ -d "$BUILD_DIR" ]; then
-  cd $BUILD_DIR
+  cd "$BUILD_DIR"
 
   # Update manifest with dev public key
   MANIFEST_PATH="./manifest.json"


.github/workflows/build-browser.yml
```diff
@@ -218,20 +218,24 @@ jobs:
...
             npm_command_prefix: "dist:bit:"
+            npm_package_dev_prefix: "package:bit:dev:"
             readable: "commercial license"
             type: "commercial"
         browser:
           - name: "chrome"
             npm_command_suffix: "chrome"
             archive_name: "dist-chrome.zip"
             artifact_name: "dist-chrome-MV3"
+            artifact_name_dev: "dist-dev-chrome-MV3"
```
trmartin4: 💭 What if we removed `dist` altogether?  is there a reason we have to have that prefix?  To me it's always been shorthand for "distribution", and we don't intend to distribute this.  It might also make it less likely to do it by accident 🤔 .

pixman20: Good point, that sounds good to me.

ℹ️  I didn't update the the destination directory, which doesn't make any difference for the workflow run as it's a temporary location before being uploaded.
Locally, it'll still end up in `dist/` right now.  It could go to a `dev/`, but I don't know that anyone would be packaging this locally anyway since you'd load the unpacked version from `build` anyway I believe.

trmartin4:
I also confirmed that this approach works with QA 👍 .

PR #17478, opened 2025-11-18
[CL-854] add page header component to `libs/components`

willmartian:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/CL-854

## 📔 Objective

- Duplicate the presentational aspects of the web header component to a new component in the CL: `bit-header`
- Update the web header to use said component

Side quests:
- update headers to use current best practices: standalone, OnPush change detection, signals, new template syntax, ts strict
- split the account menu from the web header into its own component
  - All of the domain logic in the web header really just belongs to the account menu. Moving this into its own component will let us transfer ownership to a more appropriate team

Note:
- The stories for the web header were deleted and supplanted by stories for `bit-header`. They were duplicative, and did not capture anything specific about the account menu or product switcher, which were the only real additions. 

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/apps/web/src/app/layouts/header/account-menu.component.ts
+++ b/apps/web/src/app/layouts/header/account-menu.component.ts
@@ -1,12 +1,13 @@
 import { ChangeDetectionStrategy, Component, inject } from "@angular/core";
+import { toSignal } from "@angular/core/rxjs-interop";
 import { map, Observable } from "rxjs";
 
-import { Account, AccountService } from "@bitwarden/common/auth/abstractions/account.service";
+import { LockService, LogoutService } from "@bitwarden/auth/common";
+import { AccountService } from "@bitwarden/common/auth/abstractions/account.service";
 import {
   VaultTimeoutAction,
   VaultTimeoutSettingsService,
 } from "@bitwarden/common/key-management/vault-timeout";
-import { MessagingService } from "@bitwarden/common/platform/abstractions/messaging.service";
 import { PlatformUtilsService } from "@bitwarden/common/platform/abstractions/platform-utils.service";
 
 import { DynamicAvatarComponent } from "../../components/dynamic-avatar.component";
@@ -21,21 +22,29 @@ import { SharedModule } from "../../shared";
 export class AccountMenuComponent {
   private readonly platformUtilsService = inject(PlatformUtilsService);
   private readonly vaultTimeoutSettingsService = inject(VaultTimeoutSettingsService);
-  private readonly messagingService = inject(MessagingService);
   private readonly accountService = inject(AccountService);
+  private readonly logoutService = inject(LogoutService);
+  private readonly lockService = inject(LockService);
 
-  protected readonly account$: Observable<Account | null> = this.accountService.activeAccount$;
+  protected readonly account = toSignal(this.accountService.activeAccount$);
+
   protected readonly canLock$: Observable<boolean> = this.vaultTimeoutSettingsService
     .availableVaultTimeoutActions$()
     .pipe(map((actions) => actions.includes(VaultTimeoutAction.Lock)));
   protected readonly selfHosted = this.platformUtilsService.isSelfHost();
   protected readonly hostname = globalThis.location.hostname;
 
-  protected lock() {
-    this.messagingService.send("lockVault");
+  protected async lock() {
+    const userId = this.account()?.id;
+    if (userId) {
+      await this.lockService.lock(userId);
+    }
   }
 
-  protected logout() {
-    this.messagingService.send("logout");
+  protected async logout() {
+    const userId = this.account()?.id;
+    if (userId) {
+      await this.logoutService.logout(userId);
+    }
   }
 }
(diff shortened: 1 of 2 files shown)


libs/components/src/header/header.component.html
```diff
@@ -0,0 +1,35 @@
+<header
+  class="-tw-mt-6 -tw-mx-8 tw-mb-3 tw-flex tw-flex-col tw-py-6 tw-px-8 has-[[data-tabs]:not(:empty)]:tw-border-0 has-[[data-tabs]:not(:empty)]:tw-border-b has-[[data-tabs]:not(:empty)]:tw-border-solid has-[[data-tabs]:not(:empty)]:tw-border-secondary-100 has-[[data-tabs]:not(:empty)]:tw-bg-background-alt has-[[data-tabs]:not(:empty)]:tw-pb-0"
```
willmartian, line 2: Have fun reading this!

I wish I could do a one off alias like:

```html
<div class="alias:tabs-not-empty([[data-tabs]:not(:empty)]) tabs-not-empty:tw-foo tabs-not-empty:tw-bar ...">
```

willmartian: (This CSS-only approach was taken because I don't think `.childElementCount` will trigger `OnPush` change detection.)

libs/components/src/header/header.stories.ts
```diff
@@ -0,0 +1,190 @@
...
+  title: "Component Library/Header",
+  component: HeaderComponent,
+  decorators: [
+    componentWrapperDecorator(
+      (story) => `<div class="tw-min-h-screen tw-flex-1 tw-p-6 tw-text-main">${story}</div>`,
+    ),
+    moduleMetadata({
+      imports: [
+        HeaderComponent,
+
```
vleague2:
```suggestion
```

libs/components/src/header/header.component.html
```diff
@@ -0,0 +1,35 @@
...
+    <div class="tw-ml-auto tw-flex tw-flex-col tw-gap-4">
+      <div class="tw-flex tw-min-w-max tw-items-center tw-justify-end tw-gap-2">
+        <ng-content></ng-content>
+      </div>
+      <div class="tw-ml-auto empty:tw-hidden">
+        <ng-content select="[slot=secondary]"></ng-content>
+      </div>
+    </div>
+  </div>
+  <div data-tabs class="-tw-mx-4 -tw-mb-px empty:tw-hidden">
```
vleague2, line 32: So the empty selector works here? On MDN I'm seeing that it might count whitespace around comments as "content"

libs/components/src/header/header.component.ts
```diff
@@ -0,0 +1,19 @@
...
+
+@Component({
+  selector: "bit-header",
+  templateUrl: "./header.component.html",
+  changeDetection: ChangeDetectionStrategy.OnPush,
+  standalone: true,
+})
+export class HeaderComponent {
+  /**
+   * Custom title that overrides the route data `titleId`
```
vleague2: Comment doesn't seem accurate for the CL version

apps/web/src/app/layouts/header/account-menu.component.ts
```diff
@@ -0,0 +1,41 @@
...
+    .pipe(map((actions) => actions.includes(VaultTimeoutAction.Lock)));
+  protected readonly selfHosted = this.platformUtilsService.isSelfHost();
+  protected readonly hostname = globalThis.location.hostname;
+
+  protected lock() {
+    this.messagingService.send("lockVault");
+  }
+
+  protected logout() {
+    this.messagingService.send("logout");
```
JaredSnider-Bitwarden: ❓ since we are rewriting this component, can it be updated to inject the `LogoutService` ([abstraction](https://github.com/bitwarden/clients/blob/main/libs/auth/src/common/abstractions/logout.service.ts)) and use that instead?

libs/components/src/header/header.component.html
Hinton: It won't trigger. You could use https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver to manually trigger when number of children changes but that might be more complex than the css.

libs/components/src/header/header.component.html
willmartian: It should work. Angular will trim whitespace in templates by default (unless you enable the `preserveWhitespace` option, which we disabled a while back).

libs/components/src/header/header.stories.ts
willmartian: [5ebce11](https://github.com/bitwarden/clients/pull/17478/commits/5ebce11f011092f4e8e3ec900cbe5758250a4246)

libs/components/src/header/header.component.ts
willmartian: [7049190](https://github.com/bitwarden/clients/pull/17478/commits/704919005bbadb3ef35899e55861b52d9e7da289)

willmartian:
I split this into two PRs:
- addition of new header to CL: https://github.com/bitwarden/clients/pull/17662
- refactor of web header to use new header internally: https://github.com/bitwarden/clients/pull/17663 

(The new header is needed urgently for other work, so I don't want to block on QA.)

PR #17479, opened 2025-11-18
PM-26250 Explore options to enable direct importer for mac app store build

harr1424:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-26250?atlOrigin=eyJpIjoiN2Y1Y2NhNmQ4OTBhNDU4YzkwZWNmYmI1MTdiNzc1NDgiLCJwIjoiaiJ9

## 📔 Objective

For the direct importer, browser profiles aren't populating when the desktop app was installed via the mac app store. This is because macOS apps can be [sandboxed](https://developer.apple.com/documentation/xcode/configuring-the-macos-app-sandbox) and won't have access to system directories to look for browser profiles.

The goal of this effort is to enable the direct importer for the mac app store build.

## 🎥  Screencast

Screen Recording 2025-11-21 at 14 32 37 mov

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


apps/desktop/desktop_native/objc/build.rs
coroiu, line 1: It seems odd put swift into a crate called a `objc`, but also, do we have to add another language? The build file for `objc` was so simple, and the interface contained an easy to use async/future-compatible API with error handling and allowed complex data structures by using JSON serialization. Aren't all Apple APIs compatible with objc? Can't we just use that?

harr1424: > Aren't all Apple APIs compatible with objc? Can't we just use that?

I can check, Apple has introduced APIs available in Swift that are not backwards compatible with objc. Swift has something like FFI (even cleaner) for working with any objc code, but the reverse is not guaranteed to be true. 

This Swift code is used to display an `NSOpenPanel()`, I have a feeling that existed in the objc days, but I am not sure what the equivalent of `DispatchQueue.main.sync` is, and a few of the other Apple APIs used here, I will need to check on.

harr1424: @coroiu I ran the Swift file by Claude and all of the APIs have equivalents in objc, I'll get started on making the change.

coroiu: @coltonhurst You're currently the only users of the `objc` crate and its `runCommand` framework. In hindsight I admit that the JSON approach might not have been the best, but there's still some callback-to-async and string interop for automatically freeing memory (without depending on `libc`) that I feel is worth using/saving. Does Autofill have any current plans for improving the `objc` crate? Would you also prefer swift?

Hinton:
Electron might support this out of the box, I found [StartAccessingSecurityScopedResource](https://www.electronjs.org/docs/latest/api/app#appstartaccessingsecurityscopedresourcebookmarkdata-mas), which seems to be quite similar to what we want to achieve, although we have to do this on the JS side.

apps/desktop/desktop_native/chromium_importer/src/chromium/mod.rs
```diff
@@ -60,26 +60,47 @@ impl InstalledBrowserRetriever for DefaultInstalledBrowserRetriever {
     fn get_installed_browsers() -> Result<Vec<String>> {
         let mut browsers = Vec::with_capacity(SUPPORTED_BROWSER_MAP.len());
 
+        #[allow(unused_variables)] // config only used outside of sandbox
         for (browser, config) in SUPPORTED_BROWSER_MAP.iter() {
-            let data_dir = get_browser_data_dir(config)?;
-            if data_dir.exists() {
+            #[cfg(all(target_os = "macos", feature = "sandbox"))]
```
Hinton: **issue:** I don't advice using compile flags for determining if it's a mas build or not. It tends to be fairly fragile since you can easily configure your own environment to build all flags. I would suggest that we rather pass in a MAS boolean to identify the environment.

apps/desktop/desktop_native/chromium_importer/src/chromium/platform/macos.rs
```diff
@@ -7,6 +12,72 @@ use crate::{
     util,
 };
 
+//
+// Sandbox specific (for Mac App Store Builds)
+//
+
+#[cfg(feature = "sandbox")]
```
Hinton: **suggestion:** If you end up adding `feature` to multiple blocks, either wrap them in a module, or move to a separate file for easier organization.

apps/desktop/desktop_native/objc/src/native/chromium_importer/browser_access_manager.m
```diff
@@ -0,0 +1,196 @@
...
+            @"Chromium": @"Library/Application Support/Chromium",
+            @"Microsoft Edge": @"Library/Application Support/Microsoft Edge",
+            @"Brave": @"Library/Application Support/BraveSoftware/Brave-Browser",
+            @"Arc": @"Library/Application Support/Arc/User Data",
+            @"Opera": @"Library/Application Support/com.operasoftware.Opera",
+            @"Vivaldi": @"Library/Application Support/Vivaldi"
+        };
+    }
+    return self;
+}
```
Hinton, line 27: I think if we want we can keep this information on the rust side and minimize the obj-c we have to maintain.

Hinton:
You still have the sandbox flag.

harr1424:
> You still have the sandbox flag.

Sorry, based upon [this comment](https://github.com/bitwarden/clients/pull/17479#discussion_r2569752917) I thought it would make more sense to keep the flag for conditional compilation, but to use a bool to control conditional logic. 

Just confirming you would like the flag removed completely? And always compile the sandbox related code for all Mac OS builds?

Hinton:
> > You still have the sandbox flag.
> 
> Sorry, based upon [this comment](https://github.com/bitwarden/clients/pull/17479#discussion_r2569752917) I thought it would make more sense to keep the flag for conditional compilation, but to use a bool to control conditional logic.
> 
> Just confirming you would like the flag removed completely? And always compile the sandbox related code for all Mac OS builds?

I see how that could be confusing. Personally I would prefer not gating this behind a flag. It's not something that probably needs to be behind a flag, and always compiling it should prevent any accidental enabling/disabling of the flag as it's not a pattern we use elsewhere at the moment.

apps/desktop/desktop_native/chromium_importer/src/chromium/mod.rs
```diff
@@ -51,35 +51,62 @@ pub enum LoginImportResult {
...
+    }
+    Ok(())
+}
+
 pub async fn import_logins(
-    browser_name: &String,
-    profile_id: &String,
+    browser_name: &str,
+    profile_id: &str,
+    #[cfg(target_os = "macos")]
```
Hinton: You can't gate arguments this way, you could duplicate the method behind target and not target flags, but since this is exposed in the type index it's not something I would recommend. You can probably always include it?

apps/desktop/desktop_native/objc/build.rs
coltonhurst: @coroiu thanks for the mention. I am not intimately familiar with this area of the code base, but I agree with your points.

I think keeping only Objective-C code in the `objc` directory is a good move. It looks like that might have been updated in this PR since your message and my time in looking at this? (Thanks for your patience!)

Regarding the JSON pattern... I don't think we should deviate from an existing pattern, creating two different patterns, unless there is a plan to transition the old stuff, or if there is a specific reason we need to do it a different way. This is just my preference; I would rather code patterns be consistent (even if not the "perfect" solution), rather than "right" stuff that is not consistent and less easily understood to new readers. So... the tldr for this point is it might be worth looking into a way to do this in a way that follows the existing pattern, rather than creating new patterns.

coltonhurst: Additionally, we have far reaching plans for significant desktop_native improvements as a whole. However, nothing will happen in an immediate time frame.

harr1424: @coroiu please let me know if you believe anything currently in this PR deviates from the preferences mentioned above; I believe I have addressed all concerns and have a working MAS build running locally.

coroiu: @harr1424 Have you looked at how the existing objc code looks, how it is executed, how parameters are sent, how data is returned and how memory is cleared? I see significant differences:

- Functions referenced directly using `extern "C"`
- Parameters are constructed manually in every consuming function using unsafe rust
- Return data is converted manually in every consuming function using unsafe rust
- Memory is manually cleared
- Libc is added as a dependency to clear memory

The reason I asked @coltonhurst is because we might very well want to move over to referencing functions directly, I think that would make writing cross-platform passkey autofill code easier, but in that case I think we should be more restrictive with where we put unsafe rust, isolate it more, and make a plan for how to support async functions.

Seeing as we do not want to deviate from existing patterns then this code, @harr1424, needs to be convert to using the JSON command structure and be called using `run_command` instead of referenced directly using `extern "c"` blocks. We should also remove `libc` as a dependency since `run_command` will manage the memory for you. `run_command` will also add some exception handling for you which I think might be missing in the current version of your code (not sure if it's actually able to throw anything, but doesn't hurt to be safe).

apps/desktop/desktop_native/chromium_importer/src/chromium/platform/macos.rs
```diff
@@ -7,6 +7,220 @@ use crate::{
...
+            CommandResult::Error { error } => Err(anyhow!("{}", error)),
+        }
+    }
+
+    #[derive(Debug, Deserialize)]
+    struct CheckBrowserInstalledResponse {
+        #[serde(rename = "isInstalled")]
+        is_installed: bool,
+    }
+}
```
Hinton: **suggestion:** This is better organized by moving to a dedicated file assuming you want to keep the namespace. The previous comment about namespacing it so you could apply `cfg` once assumed it would be unwrapped using `pub use sandbox::*`

apps/desktop/desktop_native/chromium_importer/src/chromium/platform/macos.rs
```diff
@@ -7,6 +7,220 @@ use crate::{
...
+            let input = CommandInput {
+                namespace: "chromium_importer".to_string(),
+                command: "request_access".to_string(),
+                params: serde_json::json!({
+                    "browserName": browser_name,
+                    "relativePath": relative_path,
+                }),
+            };
+
+            let output = desktop_objc::run_command(serde_json::to_string(&input)?)
```
Hinton: @coroiu if we want to use a json interface we should probably define structs for it? Similar to the `CommandInput` defined in this file. With type safe Namespace.

apps/desktop/desktop_native/chromium_importer/src/chromium/platform/macos.rs
```diff
@@ -7,6 +7,220 @@ use crate::{
...
+                .first()
+                .ok_or_else(|| anyhow!("No data directory configured for browser"))?;
+
+            let input = CommandInput {
+                namespace: "chromium_importer".to_string(),
+                command: "request_access".to_string(),
+                params: serde_json::json!({
+                    "browserName": browser_name,
+                    "relativePath": relative_path,
+                }),
```
Hinton: **question:** I think this actually results in json within json technically valid but not particularly nice.

I wonder if `HashMap<String, Value>` would be a more appropriate type. Or make `CommandInput` generic and have params be it's own struct.

apps/desktop/desktop_native/chromium_importer/src/chromium/platform/macos.rs
```diff
@@ -7,6 +7,220 @@ use crate::{
...
+        ("Opera", "com.operasoftware.Opera"),
+        ("Vivaldi", "com.vivaldi.Vivaldi"),
+    ];
+
+    #[derive(Debug, Deserialize)]
+    #[serde(tag = "type")]
+    enum CommandResult<T> {
+        #[serde(rename = "success")]
+        Success { value: T },
+        #[serde(rename = "error")]
```
Hinton: **suggestion:** Use `#[serde(rename_all = "camelCase")]` on the structs instead.

apps/desktop/desktop_native/chromium_importer/src/chromium/platform/macos.rs
```diff
@@ -7,6 +7,220 @@ use crate::{
...
+    async fn is_browser_installed(browser_name: &str) -> Result<bool> {
+        let bundle_id = BROWSER_BUNDLE_IDS
+            .iter()
+            .find(|(name, _)| *name == browser_name)
+            .map(|(_, id)| *id);
+
+        let bundle_id = match bundle_id {
+            Some(id) => id,
+            None => return Ok(true),
+        };
```
Hinton: **issue:** Use `ok_or`, https://contributing.bitwarden.com/contributing/code-style/rust#ok_or.

apps/desktop/desktop_native/objc/src/native/bridging-header.h
coroiu, line 1: remove?

apps/desktop/desktop_native/chromium_importer/src/chromium/mod.rs
```diff
@@ -51,35 +51,71 @@ pub enum LoginImportResult {
...
+                // When not in sandbox check file system directly
+                let data_dir = get_and_validate_data_dir(config)?;
+                if data_dir.exists() {
+                    browsers.push((*browser).to_string());
+                }
             }
         }
 
         Ok(browsers)
     }
```
Hinton:
```suggestion
    fn get_installed_browsers(mas_build: bool) ->Vec<String> {
        // Show all browsers for MAS builds, user will grant access when selected
        if mas_build {
            return SUPPORTED_BROWSER_MAP
                .keys()
                .map(|browser| (*browser).to_string())
                .collect();
        }

        // When not in sandbox, check file system directly
        SUPPORTED_BROWSER_MAP
            .iter()
            .filter_map(|(browser, config)| {
                get_and_validate_data_dir(config)
                    .ok()
                    .filter(|data_dir| data_dir.exists())
                    .map(|_| (*browser).to_string())
            })
            .collect()
    }
```

- Drops `Result` because we never return an error.
- Handles the if mas_build early to decreaase complexity in the loop.
- Uses iterator instead of loop

apps/desktop/desktop_native/objc/src/native/chromium_importer/run_chromium_command.h
coroiu, line 1: Nice! Could you add these files to tools ownership?

apps/desktop/desktop_native/chromium_importer/src/chromium/mod.rs
```diff
@@ -51,35 +51,71 @@ pub enum LoginImportResult {
...
+    browser_name: &str,
+    mas_build: bool,
+) -> Result<Vec<ProfileInfo>> {
+    // MAS builds need to resume security-scoped access before reading browser files
+    #[cfg(target_os = "macos")]
+    let _access = if mas_build {
+        Some(platform::sandbox::ScopedBrowserAccess::resume(browser_name).await?)
+    } else {
+        None
+    };
```
Hinton:
```suggestion
    if mas_build {
        Some(platform::sandbox::ScopedBrowserAccess::resume(browser_name).await?)
    }
```

apps/desktop/desktop_native/chromium_importer/src/chromium/mod.rs
```diff
@@ -51,35 +51,71 @@ pub enum LoginImportResult {
...
+    }
+    Ok(())
+}
+
 pub async fn import_logins(
-    browser_name: &String,
-    profile_id: &String,
+    browser_name: &str,
+    profile_id: &str,
+    _mas_build: bool,
```
Hinton: If the linter complains, just ignore it rather than calling it `_ `
```suggestion
    mas_build: bool,
```

apps/desktop/desktop_native/chromium_importer/src/chromium/mod.rs
```diff
@@ -51,35 +51,71 @@ pub enum LoginImportResult {
...
 ) -> Result<Vec<LoginImportResult>> {
+    // MAS builds will use the formerly created security bookmark
+    #[cfg(target_os = "macos")]
+    let _access = if _mas_build {
+        Some(platform::sandbox::ScopedBrowserAccess::resume(browser_name).await?)
+    } else {
+        None
+    };
```
Hinton:
```suggestion
    if _mas_build {
        Some(platform::sandbox::ScopedBrowserAccess::resume(browser_name).await?)
    }
```

apps/desktop/desktop_native/build.js
Hinton, line 1: **issue:** Revert.

Hinton (commented):
Only reviewed the native code so far.

apps/desktop/desktop_native/chromium_importer/src/chromium/platform/macos.rs
harr1424: I am open to either approach

apps/desktop/desktop_native/chromium_importer/src/chromium/platform/macos.rs
coroiu: The original use-case makes the types opaque to Rust and are instead defined in TypeScript. Ideally I would've like for autofill to take the lead on thise since they are the ones that primarily use this and know how such a change will affect their code. But I agree that the way we're using it here, constructing the json in rust, it would be better with defining structs

coroiu (dismissed):
Looks good (no big changes),please hold off on pinging me from further reviews until you have approval from @Hinton so I don't have to approve an empty change set :)

apps/desktop/desktop_native/chromium_importer/src/chromium/platform/macos.rs
Hinton: Generic CommandInput seems saner since it ensures type safety.

apps/desktop/desktop_native/chromium_importer/src/chromium/platform/sandbox.rs
```diff
@@ -0,0 +1,206 @@
+//
+// Sandbox specific (for Mac App Store Builds)
+//
```
Hinton: We can use `///` to make the documentation of the module public
```suggestion
/// Sandbox support for Mac App Store Builds
```

apps/desktop/desktop_native/chromium_importer/src/chromium/platform/sandbox.rs
```diff
@@ -0,0 +1,206 @@
...
+
+use anyhow::{anyhow, Result};
+use serde::{Deserialize, Serialize};
+
+// Before requesting access outside of sandbox, determine if browser is actually installed
```
Hinton: **question:** This comment doesn't seem that relevant to the code?

apps/desktop/desktop_native/chromium_importer/src/chromium/platform/sandbox.rs
```diff
@@ -0,0 +1,206 @@
...
+];
+
+#[derive(Debug, Deserialize)]
+#[serde(tag = "type")]
+enum CommandResult<T> {
+    #[serde(rename_all = "camelCase")]
+    Success { value: T },
+    #[serde(rename_all = "camelCase")]
+    Error { error: String },
+}
```
Hinton, line 18:
```suggestion
#[derive(Debug, Deserialize)]
#[serde(rename_all = "camelCase", tag = "type")]
enum CommandResult<T> {
    Success { value: T },
    Error { error: String },
}
```

apps/desktop/desktop_native/chromium_importer/src/chromium/platform/sandbox.rs
```diff
@@ -0,0 +1,206 @@
...
+        CommandResult::Success { value } => Ok(value.is_installed),
+        CommandResult::Error { error } => Err(anyhow!("{}", error)),
+    }
+}
+
+#[derive(Debug, Deserialize)]
+struct CheckBrowserInstalledResponse {
+    #[serde(rename = "isInstalled")]
+    is_installed: bool,
+}
```
Hinton:
```suggestion
#[derive(Debug, Deserialize)] 
#[serde(rename_all = "camelCase")]
struct CheckBrowserInstalledResponse {
    is_installed: bool,
}
```

Hinton: Should this be moved up next to the other structs?

apps/desktop/desktop_native/objc/src/native/chromium_importer/browser_access_manager.m
```diff
@@ -0,0 +1,157 @@
...
+
+- (NSString *)bookmarkKeyFor:(NSString *)browserName {
+    return [NSString stringWithFormat:@"%@.%@", _bookmarkKey, browserName];
+}
+
+- (void)saveBookmark:(NSData *)data forBrowser:(NSString *)browserName {
+    NSUserDefaults *defaults = [NSUserDefaults standardUserDefaults];
+    NSString *key = [self bookmarkKeyFor:browserName];
+    [defaults setObject:data forKey:key];
+    [defaults synchronize];
```
Hinton: This seems to be deprecated and not necessary? https://developer.apple.com/documentation/foundation/userdefaults/synchronize()

apps/desktop/desktop_native/objc/src/native/chromium_importer/browser_access_manager.m
```diff
@@ -0,0 +1,157 @@
...
+
+    if (panelResult != NSModalResponseOK || !selectedURL) {
+        return nil;
+    }
+
+    NSURL *localStatePath = [selectedURL URLByAppendingPathComponent:@"Local State"];
+    if (![[NSFileManager defaultManager] fileExistsAtPath:localStatePath.path]) {
+        // Invalid directory selected caller will handle
+        return nil;
+    }
```
Hinton, line 108: Do we need to check that. the user selected the correct loction?

apps/desktop/src/app/tools/import/desktop-import-metadata.service.ts
```diff
@@ -20,7 +20,8 @@ export class DesktopImportMetadataService
   }
 
   async init(): Promise<void> {
-    const metadata = await ipc.tools.chromiumImporter.getMetadata();
+    const isMas = ipc.platform.isMacAppStore;
```
Hinton: **question:** Should we use the platformService instead?

apps/desktop/src/app/tools/import/chromium-importer.service.ts
```diff
@@ -2,20 +2,31 @@ import { ipcMain } from "electron";
...
   constructor() {
-    ipcMain.handle("chromium_importer.getMetadata", async (event) => {
-      return await chromium_importer.getMetadata();
+    ipcMain.handle("chromium_importer.getMetadata", async (event, isMas: boolean) => {
+      return await chromium_importer.getMetadata(isMas);
+    });
+
+    // Used on Mac OS App Store builds to request permissions to browser entries outside the sandbox
+    ipcMain.handle("chromium_importer.requestBrowserAccess", async (event, browser: string) => {
+      if (chromium_importer.requestBrowserAccess) {
```
Hinton: **question:** I don't believe we made this null since it would cause different `index.d.ts` depending on platform?

apps/desktop/src/app/tools/import/import-desktop.component.html
```diff
@@ -12,7 +12,7 @@
       ></tools-import>
       @if (loading) {
         <div class="tw-absolute tw-inset-0 tw-flex tw-items-center tw-justify-center">
-          <i class="bwi bwi-spinner bwi-spin bwi-3x tw-text-primary-600" aria-hidden="true"></i>
+          <bit-spinner></bit-spinner>
```
Hinton: Dialogs have a loading state https://components.bitwarden.com/?path=/story/component-library-dialogs-dialog--loading. Although this is not how we generally represent progress state since usually just the submit button has a loading indictor.

apps/desktop/src/app/tools/import/import-desktop.component.ts
```diff
@@ -39,7 +47,13 @@ export class ImportDesktopComponent {
   protected disabled = false;
   protected loading = false;
 
-  constructor(public dialogRef: DialogRef) {}
+  protected readonly onLoadProfilesFromBrowser = this._onLoadProfilesFromBrowser.bind(this);
+  protected readonly onImportFromBrowser = this._onImportFromBrowser.bind(this);
```
Hinton, line 43: **question:** What's up with the bind here?

apps/desktop/desktop_native/objc/src/native/chromium_importer/browser_access_manager.m
harr1424: This situation is checked by the calling code: 
```objc
void requestAccessCommand(void* context, NSDictionary *params) {
  NSString *browserName = params[@"browserName"];
  NSString *relativePath = params[@"relativePath"];

  if (!browserName || !relativePath) {
    return _return(context, _error(@"Missing required parameters: browserName and relativePath"));
  }

  BrowserAccessManager *manager = [[BrowserAccessManager alloc] init];
  NSString *bookmarkData = [manager requestAccessToBrowserDir:browserName relativePath:relativePath];

  if (bookmarkData == nil) {
 // early return with i18n translation key having English value "Either an invalid browser directory was selected, or permission was denied" 
    return _return(context, _error(@"browserAccessDenied")); 
  }
```

harr1424: Sorry, any directory containing a `Localstate` file will technically pass, I will make this more defensive

apps/desktop/desktop_native/objc/src/native/chromium_importer/browser_access_manager.m
```diff
@@ -0,0 +1,157 @@
...
+
+    // Access is temporary right now, persist it by creating a security bookmark
+    NSError *error = nil;
+    NSData *bookmarkData = [selectedURL bookmarkDataWithOptions:NSURLBookmarkCreationWithSecurityScope
+                                            includingResourceValuesForKeys:nil
+                                            relativeToURL:nil
+                                            error:&error];
+
+    if (!bookmarkData) {
```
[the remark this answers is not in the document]
harr1424: Why would assigning `nil` to this variable be helpful here?

apps/desktop/desktop_native/chromium_importer/src/chromium/platform/sandbox.rs
```diff
@@ -0,0 +1,207 @@
...
+#[derive(Debug, Deserialize)]
+#[serde(rename_all = "camelCase")]
+struct CheckBrowserInstalledResponse {
+    is_installed: bool,
+}
+
+#[derive(Debug, Deserialize)]
+#[serde(rename_all = "camelCase", tag = "type")]
+enum CommandResult<T> {
+    // rename = "camelCase" was a review suggestion with breaking changes
```
Hinton: ?

```suggestion
```

apps/desktop/desktop_native/chromium_importer/src/chromium/platform/sandbox.rs
```diff
@@ -0,0 +1,207 @@
...
+async fn is_browser_installed(browser_name: &str) -> Result<bool> {
+    let bundle_id = BROWSER_BUNDLE_IDS
+        .iter()
+        .find(|(name, _)| *name == browser_name)
+        .map(|(_, id)| *id);
+
+    let Some(bundle_id) = bundle_id else {
+        return Ok(true); // ok_or(true) was a review suggestion with breaking changes
+                         // Avoid ok_or(true): serializes Result as {"Ok": "..."} instead of string
+                         // value
```
Hinton:
```suggestion
        return Ok(true);
```

apps/desktop/src/app/tools/import/import-desktop.component.ts
```diff
@@ -48,11 +62,37 @@ export class ImportDesktopComponent {
...
+    try {
+      // Request browser access (required for sandboxed builds, no-op otherwise)
+      await ipc.tools.chromiumImporter.requestBrowserAccess(browser);
+    } catch (error) {
+      const rawMessage = error instanceof Error ? error.message : "";
+
+      // Check verbose error chain for specific i18n key indicating browser not installed
+      const browserNotInstalledMatch = rawMessage.match(
+        /chromiumImporterBrowserNotInstalled:([^:]+)/,
+      );
```
Hinton, line 78: **observation:** This suggest you want to return an actual error enum rather than using string based errors.

libs/importer/src/components/chrome/import-chrome.component.ts
```diff
@@ -105,15 +105,29 @@ export class ImportChromeComponent implements OnInit, OnDestroy {
...
   constructor(
     private formBuilder: FormBuilder,
     private controlContainer: ControlContainer,
     private logService: LogService,
     private i18nService: I18nService,
   ) {
     effect(async () => {
-      this.profileList = await this.onLoadProfilesFromBrowser(this.getBrowserName(this.format()));
-      // FIXME: Add error handling and display when profiles could not be loaded/retrieved
+      // Callback is set via @Input after constructor, so check it exists
```
Hinton: **issue:** Effect is used when you want to react to a signal being changed. There is no signal here.

> Registers an "effect" that will be scheduled & executed whenever the signals that it reads changes.

You can use `ngOnInit`, or convert the inputs to a signal.

apps/desktop/src/app/tools/import/desktop-import-metadata.service.ts
harr1424: I am not sure, the only platformService I have found is [PlatformUtilsService](https://github.com/bitwarden/clients/blob/d7a7f19abdaa9aa95f0919df2e8aa8f249d6c813/libs/common/src/platform/abstractions/platform-utils.service.ts#L14)

apps/desktop/src/app/tools/import/import-desktop.component.html
harr1424: @sukhleenb do you have a preference as to which approach is used here? 

1. Limiting the spinner to the submit button
2. Displaying the spinner in the dialog box as shown [here](https://private-user-images.githubusercontent.com/84741727/519303473-5ce79dd8-4a89-4a2f-928e-cdc753e5cc96.gif?jwt=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3NjU4OTk2MjUsIm5iZiI6MTc2NTg5OTMyNSwicGF0aCI6Ii84NDc0MTcyNy81MTkzMDM0NzMtNWNlNzlkZDgtNGE4OS00YTJmLTkyOGUtY2RjNzUzZTVjYzk2LmdpZj9YLUFtei1BbGdvcml0aG09QVdTNC1ITUFDLVNIQTI1NiZYLUFtei1DcmVkZW50aWFsPUFLSUFWQ09EWUxTQTUzUFFLNFpBJTJGMjAyNTEyMTYlMkZ1cy1lYXN0LTElMkZzMyUyRmF3czRfcmVxdWVzdCZYLUFtei1EYXRlPTIwMjUxMjE2VDE1MzUyNVomWC1BbXotRXhwaXJlcz0zMDAmWC1BbXotU2lnbmF0dXJlPWFjOGY3YWIzNTY5Njc5MmIwNmM3NGMwZTEzNDFmNDdiMTgwMmY0NDI5YTg1MmM5MjUxYmRlMWRmZmQ0OGNiNzgmWC1BbXotU2lnbmVkSGVhZGVycz1ob3N0In0.cGfy1qNlPFgE2jp2LdExyc8XtcsTeQLa4nKY7IJKK4Q)

apps/desktop/src/app/tools/import/import-desktop.component.ts
harr1424: These methods are passed as callbacks to the child ImportComponent via Input properties. Without binding, `_onLoadProfilesFromBrowser` would fail when trying to access `this.i18nService`.

apps/desktop/desktop_native/chromium_importer/src/chromium/platform/sandbox.rs
harr1424: Sorry, this outdated and incorrect comment has been removed. It was actually your suggested use of `ok_or()` that had breaking changes: 

```rust
async fn is_browser_installed(browser_name: &str) -> Result<bool> {
    let bundle_id = BROWSER_BUNDLE_IDS
        .iter()
        .find(|(name, _)| *name == browser_name)
        .map(|(_, id)| *id);

    let Some(bundle_id) = bundle_id else {
        return Ok(true); // ok_or(true) was a review suggestion with breaking changes
                         // Avoid ok_or(true): serializes Result as {"Ok": "..."} instead of string
                         // value
    };
    ```

File: apps/desktop/src/app/tools/import/import-desktop.component.ts (hunk 27 above)
Comment (harr1424) [reply apps/desktop/src/app/tools/import/import-desktop.component.ts:78]: Not an enum exactly but `i18n` translation keys: 

```json
  "chromiumImporterBrowserNotInstalled": {
    "message": "$BROWSER$ is not installed. Please install $BROWSER$ before importing passwords from it.",
    "placeholders": {
      "browser": {
        "content": "$1",
        "example": "Chrome"
      }
    }
  },  
    "browserAccessDenied": {
    "message": "Either an invalid browser directory was selected, or access was denied"
  },
  ```

File: apps/desktop/desktop_native/objc/build.rs
Comment (coltonhurst) [reply apps/desktop/desktop_native/objc/build.rs:1]: @harr1424 do you mind letting me know when the files relevant to this convo are ready for review? (That may be now, I'm not sure.)

I also think it would be nice to also add some architecture documentation to our contributing docs around the pattern so it's at least documented more. Any ideas @coroiu?

Comment (harr1424) [reply apps/desktop/desktop_native/objc/build.rs:1]: @coltonhurst my apologies I've moved this PR back to draft status pending Oscar's review. I'll reach out once it is ready for your re-review.

Comment (coltonhurst) [reply apps/desktop/desktop_native/objc/build.rs:1]: @harr1424 no problem! I don't think I'm a required code owner anyways, just was getting involved!

File: apps/desktop/src/app/tools/import/desktop-import-metadata.service.ts (hunk 20 above)
Comment (Hinton) [reply apps/desktop/src/app/tools/import/desktop-import-metadata.service.ts]: Yes `platformUtilsService.isMacAppStore`. `ipc.platform.isMacAppStore` should probably be considered internal to the platform team.

File: apps/desktop/src/app/tools/import/import-desktop.component.ts (hunk 23 above)
Comment (Hinton) [reply apps/desktop/src/app/tools/import/import-desktop.component.ts:43]: Passing callbacks as inputs is generally a sign of something being wrong. I think in this case we would probably be better of making a service that abstracts this than injecting callbacks. Probably outside of the scope of this ticket though.

File: apps/desktop/src/app/tools/import/import-desktop.component.ts (hunk 27 above)
Comment (Hinton) [reply apps/desktop/src/app/tools/import/import-desktop.component.ts:78]: Enum from the rust side which is then mapped to the appropriate i18n key in the client.

File: apps/desktop/src/main/tools/import/chromium-importer.service.ts
Comment (Hinton) [inline apps/desktop/src/main/tools/import/chromium-importer.service.ts:1]: **issue:** This file is located in the wrong directory. `app` contains renderer concerns but this uses `ipcMain` i.e. it's a main service.

**Review by Hinton — COMMENTED** [review]
Outstanding comment besides the location of the main service below,  https://github.com/bitwarden/clients/pull/17479#discussion_r2611145565.

The other comments are future improvements and non blocking.

File: apps/desktop/src/app/tools/import/desktop-import-metadata.service.ts (hunk 20 above)
Comment (itsadrago) [reply apps/desktop/src/app/tools/import/desktop-import-metadata.service.ts]: Is `this.system.environment.isMacAppStore()` the correct replacement then?

Comment (harr1424) [reply apps/desktop/src/app/tools/import/desktop-import-metadata.service.ts]:
```typescript 
export class DesktopImportMetadataService
  extends DefaultImportMetadataService
  implements ImportMetadataServiceAbstraction
{
  constructor(system: SystemServiceProvider) {  // note system refers to an instance of SystemServiceProvider
    super(system);
  }

...

    const isMas = this.system.environment.isMacAppStore();

```

```typescript
export type SystemServiceProvider = {
  /** Policy configured by the administrative console */
  readonly policy: PolicyService;

  /** Client extension metadata and profile access */
  readonly extension: ExtensionService;

  /** Event monitoring and diagnostic interfaces */
  readonly log: LogProvider;

  /** Config Service to determine flag features */
  readonly configService: ConfigService;

  /** Platform Service to inspect runtime environment */
  readonly environment: PlatformUtilsService;    // note that environment refers to an instance of PlatformUtilsService

  /** SDK Service */
  readonly sdk?: BitwardenClient;
};
```

Therefore `this.system.environment` should be considered synonymous with `PlatformUtilsService` in the context of this file.

apps/desktop/src/main/tools/import/chromium-importer.service.ts
itsadrago: LGTM

apps/desktop/src/app/tools/import/desktop-import-metadata.service.ts
itsadrago: LGTM

itsadrago (dismissed):

coroiu (dismissed):
I think `desktop_native/objc/src/native/chromium_importer` should be added to codeowners so it doesn't fall under platform ownership, but I'm fine with doing that in a a follow-up PR, other than that there aren't many platform changes so I'll approve

coroiu (dismissed):
No real changes to platform code

coroiu (dismissed):

itsadrago (dismissed):

theMickster:
Hello @harr1424 et. al,

I'm field testing [a new performing-multi-agent-code-review skill](https://github.com/bitwarden/ai-plugins/pull/96) and, after seeing your review hit my Github Notifications, I gave it a local run against the PR. The findings are below for your analysis.  Please incorporate the findings that are true signals and let me know if there are egregious errors that need to be corrected with planned tuning to a `CLAUDE.md` file. Many Thanks!

<details>
<summary>Sonnet Code Review: PM-26250 Explore options to enable direct importer for mac app store build (#17479)</summary>

**Date:** 2026-05-05 | **Reviewed by:** Claude Code | **Model:** Sonnet

## Summary

| Severity     | Count |
| ------------ | ----- |
| 🛑 Blocker   | 0     |
| ⚠️ Important | 4     |
| ♻️ Refactor  | 2     |

This PR enables the Chromium-based direct importer for Mac App Store sandboxed builds by adding a security-scoped bookmark flow: an `NSOpenPanel` is shown to the user, the chosen folder is validated, and a bookmark is persisted in `NSUserDefaults`. The Rust layer resumes the bookmark via a new ObjC command bridge before each profile/login read, and releases it on `Drop` via a fire-and-forget tokio task. The architecture is sound, but four ⚠️ Important issues affect security-scope lifetime correctness, error sanitization, and IPC trust boundaries — all four are recommended fixes before merge. Two ♻️ Refactor issues (a duplicated browser-bundle-ID list and duplicated error-string parsing) can be addressed in follow-ups.

## Findings

### ⚠️ Important

#### Raw native error strings rendered in user-facing toast without sanitization

`libs/importer/src/components/chrome/import-chrome.component.ts:193-195`
**Caught by:** Security & logic agent

The `translateValidationError` fallback `return message;` passes unfiltered Rust/ObjC error strings directly to `handleChromeImportError`, which places them in a toast message.

These strings originate deep in the native layer — for example `anyhow!("Failed to call ObjC command: {}", e)` and `anyhow!("Failed to parse ObjC response: {}", e)` in `sandbox.rs`. They may contain:
- Full filesystem paths (e.g., the resolved bookmark path surfaced on a bookmark-resolution failure)
- Internal Rust/ObjC type names or stack context
- Content derived from a browser's `Local State` JSON file, which is attacker-influenced data — a crafted browser installation could place arbitrary strings in that file that bubble up through JSON parse errors

This violates **P05 (Controlled Access to Vault Data)** — vault-adjacent operation context leaks to the UI — and constitutes **Data Leaking** per Bitwarden's security vocabulary. It also breaks the **Trusted Channel** model for output authenticity: the user cannot distinguish a genuine Bitwarden error from a string injected via a crafted browser data directory.

CWE-209 (Generation of Error Message Containing Sensitive Information).

**Fix.** Remove the raw fallback. All unrecognized messages should map to the generic `"errorOccurred"` i18n key. The existing string-equality check for `"browserAccessDenied"` is also fragile — prefer matching all known sentinel keys up-front and catching everything else with the generic key.
  </details>

#### `stopAccessingBrowser` resolves a new NSURL instance, breaking Apple's required start/stop symmetry

`apps/desktop/desktop_native/objc/src/native/chromium_importer/browser_access_manager.m:156-175`
**Caught by:** Security & logic agent

`startAccessingBrowser:` calls `[url startAccessingSecurityScopedResource]` on a resolved `NSURL` instance and returns its `.path` string. The NSURL itself is not retained anywhere.

`stopAccessingBrowser:` (called later from `Drop`) resolves the bookmark data a second time via `URLByResolvingBookmarkData:` to get a *new* NSURL instance, and calls `stopAccessingSecurityScopedResource` on that new instance.

Apple's documentation explicitly states: *"You must call this method on the same URL object on which you called `startAccessingSecurityScopedResource`."* Calling `stop` on a different NSURL instance — even one resolved from the same bookmark data — does not decrement the correct scope counter. The result is that the security scope opened in `startAccessingBrowser:` is **never balanced**: the sandboxed process retains read access to the browser directory indefinitely (until process exit) rather than only for the duration of the import operation.

This violates **P05 (Controlled Access to Vault Data)** — the app retains filesystem access beyond the user's intended grant window — and undermines **P06 (Minimized Impact of Security Breaches)** because the extended open scope increases the window during which a compromised renderer could initiate additional reads.

CWE-772 (Missing Release of Resource after Effective Lifetime).

**Fix.** Retain the NSURL from `startAccessingBrowser:` in an `NSMutableDictionary` keyed by `browserName` (similar to how bookmarks are stored), and call `stopAccessingSecurityScopedResource` on that retained instance in `stopAccessingBrowser:`.
  </details>

#### Fire-and-forget Drop silently skips `stop_access` if the Tokio runtime is unavailable

`apps/desktop/desktop_native/chromium_importer/src/chromium/platform/sandbox.rs:146-164`
**Caught by:** Security & logic agent

`ScopedBrowserAccess::Drop` spawns an async task via `tokio::task::spawn` without storing the `JoinHandle`. If:

- The Tokio runtime has already begun shutdown when `Drop` fires (e.g., the import completes and the process tears down),
- The task is dropped before it executes, or
- `tokio::task::spawn` panics because no runtime is active on the current thread (e.g., a synchronous test path),

then the `stop_access` ObjC command **never executes**. Combined with the NSURL symmetry bug above, this creates two compounding failure paths where the security scope is never released.

Even in the non-shutdown case, fire-and-forget spawning means there is no observable signal if the stop silently fails, making the access leak undetectable at the application layer.

This violates **P05 (Controlled Access to Vault Data)**: access to the browser directory (which contains plaintext saved passwords and browsing history — Vault-adjacent data) remains open beyond the user's authorized operation.

CWE-772 (Missing Release of Resource after Effective Lifetime).

**Fix.** Use a synchronous IPC path for the stop operation in `Drop`, or redesign `ScopedBrowserAccess` as an async-only type with an explicit `async fn close(self)` that is `.await`ed at the call site, removing the `Drop` impl entirely. At minimum, gate the spawn on `tokio::runtime::Handle::try_current()` so a missing runtime is a no-op rather than a panic.
  </details>

#### `getMetadata` trusts renderer-supplied `isMas` value instead of authoritative `isMacAppStore()` check

`apps/desktop/src/main/tools/import/chromium-importer.service.ts:7`
**Caught by:** Security & logic agent

The IPC handler for `chromium_importer.getMetadata` accepts `isMas: boolean` from the renderer process and forwards it directly to native code:

```typescript
ipcMain.handle("chromium_importer.getMetadata", async (event, isMas: boolean) => {
  return await chromium_importer.getMetadata(isMas);
});
```

All other new handlers (`requestBrowserAccess`, `getAvailableProfiles`, `importLogins`) correctly use the authoritative `isMacAppStore()` check in the main process and ignore any renderer-supplied value. `getMetadata` is the only handler that defers to the renderer.

When `mas_build=true` is passed, `DefaultInstalledBrowserRetriever::get_installed_browsers` returns **all** supported browsers without probing the filesystem. A compromised renderer can call `getMetadata(true)` on a non-MAS build to enumerate all supported browser names without any filesystem-presence check — bypassing the gate that normally prevents disclosure of which browser configurations Bitwarden supports on the user's machine.

This is a **trust boundary violation** at the IPC layer: the renderer is an untrusted context (runs in a sandboxed renderer process) and should not control security-relevant flags. This implicates **P05 (Controlled Access to Vault Data)** and **TC (Trusted Channel)** — the IPC channel is not authenticating the legitimacy of the renderer-supplied flag.

CWE-20 (Improper Input Validation) / CWE-610 (Externally Controlled Reference to a Resource).

**Fix.** Replace the renderer-supplied parameter with the authoritative main-process value, identical to the pattern used for the other three handlers:

```typescript
ipcMain.handle("chromium_importer.getMetadata", async (_event) => {
  return await chromium_importer.getMetadata(isMacAppStore());
});
```

Drop `isMas` from `apps/desktop/src/app/tools/preload.ts` and `desktop-import-metadata.service.ts` accordingly.
  </details>

### ♻️ Refactor

#### Browser name-to-bundle-ID mapping duplicated alongside `SUPPORTED_BROWSERS` in `macos.rs`

`apps/desktop/desktop_native/chromium_importer/src/chromium/platform/sandbox.rs:7-16`
**Caught by:** Architecture agent

`BROWSER_BUNDLE_IDS` in `sandbox.rs` lists the same 7 browsers — Chrome, Chromium, Microsoft Edge, Brave, Arc, Opera, Vivaldi — that are already enumerated in `SUPPORTED_BROWSERS` in `apps/desktop/desktop_native/chromium_importer/src/chromium/platform/macos.rs`. There is no compile-time enforcement keeping them in sync.

The `is_browser_installed` function silently falls back to `return Ok(true)` when a browser name has no corresponding bundle ID entry (`let Some(bundle_id) = bundle_id else { return Ok(true); }`). This means a browser added to `SUPPORTED_BROWSERS` in `macos.rs` without a matching entry in `BROWSER_BUNDLE_IDS` will bypass the installation check entirely and show up in the UI as installed on MAS builds.

**Fix.** Add an optional `bundle_id: Option<&'static str>` field to `BrowserConfig` and derive the bundle-ID lookup from that field rather than maintaining a parallel list. Alternatively, replace the silent `Ok(true)` fallback with `Ok(false)` (or an `Err`) so the parity gap fails loudly.
  </details>

#### Error-message parsing logic for browser-access errors duplicated across two layers

`apps/desktop/src/app/tools/import/import-desktop.component.ts:60-86`
**Caught by:** Code quality agent

`ImportDesktopComponent._onLoadProfilesFromBrowser` (desktop component) and `ImportChromeComponent.translateValidationError` (shared importer library) both contain the same pattern-matching logic for the `chromiumImporterBrowserNotInstalled:...` error key and the `browserAccessDenied` string.

Because `_onLoadProfilesFromBrowser` runs first and re-throws with an already-translated human-readable string, the matching code in `translateValidationError` is dead for errors originating from the desktop profile-loading path. If the error-key format ever changes, both locations must be updated in sync — with no compile-time guarantee they stay consistent.

**Fix.** The error encoding/decoding contract should live in one place. Either both layers pass structured error codes (not translated strings) up the chain, or the translation happens exclusively at the outermost catch boundary. Mixing translated strings with error-code patterns makes the flow hard to follow and easy to break silently.
  </details>

## Reviewed and Dismissed

   <details><summary>🔍 5 initial findings dismissed after validation</summary>

#### getMetadata accepts isMas from renderer; other handlers derive it in main process

`apps/desktop/src/main/tools/import/chromium-importer.service.ts:7`
**Caught by:** Architecture agent
**Original severity:** ♻️ Refactor
**Original confidence:** 92/100
**Dismissed at:** Step 5 severity audit
**Dismissed because:** Duplicate of sec-4 — same underlying issue (getMetadata trusting renderer-supplied isMas), weaker framing.

#### chromiumImporterBrowserNotInstalled error-string parsing duplicated across two components

`apps/desktop/src/app/tools/import/import-desktop.component.ts:60-86`
**Caught by:** Architecture agent
**Original severity:** ♻️ Refactor
**Dismissed at:** Step 5 severity audit
**Dismissed because:** Duplicate of quality-3 — same underlying issue (error-string parsing duplicated across components), weaker framing.

#### tokio::task::spawn in Drop panics if no runtime is active at drop time

`apps/desktop/desktop_native/chromium_importer/src/chromium/platform/sandbox.rs:146-164`
**Caught by:** Code quality agent
**Original severity:** ⚠️ Important
**Dismissed at:** Step 5 severity audit
**Dismissed because:** Duplicate of sec-3 — same underlying issue (tokio::task::spawn in Drop panics if no runtime), weaker framing.

#### stopAccessingBrowser resolves a new NSURL instance instead of reusing the one from startAccessingBrowser

`apps/desktop/desktop_native/objc/src/native/chromium_importer/browser_access_manager.m:156-175`
**Caught by:** Code quality agent
**Original severity:** ⚠️ Important
**Dismissed at:** Step 5 severity audit
**Dismissed because:** Duplicate of sec-2 — same underlying issue (NSURL start/stop symmetry), weaker framing.

#### Raw system error messages displayed to users as toast content

`libs/importer/src/components/chrome/import-chrome.component.ts:193-195`
**Caught by:** Bug analysis agent
**Original severity:** ⚠️ Important
**Dismissed at:** Step 5 severity audit
**Dismissed because:** Duplicate of sec-1 — same underlying issue (raw error fallback reaching user-facing toast), weaker framing.

   </details>

</details>  

---

<details>
<summary>Opus Code Review: PM-26250 Explore options to enable direct importer for mac app store build (#17479)</summary>

**Date:** 2026-05-05 | **Reviewed by:** Claude Code

## Summary

| Severity     | Count |
| ------------ | ----- |
| 🛑 Blocker   | 0     |
| ⚠️ Important | 0     |
| ♻️ Refactor  | 9     |

The PR enables the Chromium direct-importer flow for Mac App Store (sandboxed) desktop builds by introducing a security-scoped-bookmark request flow on the ObjC side and threading a new `mas_build` boolean through the napi → Rust → IPC → Angular call chain. No blocker- or important-tier issues survived validation: the most colorful claims (panic-in-`Drop`, mismatched-`NSURL` security-scope leak) were either dismissed by the validation pass or downgraded to refactor once the runtime guarantees and Apple's per-counter (not per-instance) contract were considered. What remains is structural debt: parallel browser arrays in three files, an inconsistent source of truth for the `isMas` flag spread across renderer, preload, and main, an unused `bookmark` field papered over with `#[allow(dead_code)]`, and shared library code that reaches for desktop-only i18n keys. None of these block merge, but each is worth resolving before the next change in this area.

## Findings

### ♻️ Refactor

#### Parallel browser arrays will drift when Chromium browsers are added

`apps/desktop/desktop_native/chromium_importer/src/chromium/platform/sandbox.rs:7-15`
**Caught by:** Architecture agent

`BROWSER_BUNDLE_IDS` (sandbox.rs lines 7-15) duplicates the browser-name keys already enumerated in `SUPPORTED_BROWSERS` (`apps/desktop/desktop_native/chromium_importer/src/chromium/platform/macos.rs` lines 14-43) and `KEYCHAIN_CONFIG` (macos.rs lines 68-104). Adding or removing a Chromium-based browser now requires synchronized edits across three parallel arrays in two files, with the only coupling being a string compare on the `name` field. The new sandbox path looks the bundle ID up by name; the natural home for that data is on `BrowserConfig` itself (e.g. an optional `bundle_id: Option<&'static str>` field next to `data_dir`) so the macOS browser table remains the single source of truth.

  </details>

#### Inconsistent source of truth for `isMas` across IPC handlers

`apps/desktop/src/main/tools/import/chromium-importer.service.ts:9-25`
**Caught by:** Architecture agent

Within a single class, two different patterns are used for the same `mas_build` parameter that is threaded all the way through napi → Rust:

- `chromium_importer.getMetadata` accepts `isMas` from the renderer (`apps/desktop/src/app/tools/import/desktop-import-metadata.service.ts` calls `this.system.environment.isMacAppStore()` and ships it across IPC; preload.ts forwards it).
- `chromium_importer.requestBrowserAccess`, `getAvailableProfiles`, and `importLogins` ignore any renderer hint and derive the value in the main process by calling `isMacAppStore()` from `apps/desktop/src/utils.ts`.

Pick one. The main-process `isMacAppStore()` utility is already used by every other consumer in `apps/desktop/src/main/` and is the established pattern for this signal. Threading the boolean through preload and the renderer for `getMetadata` only adds a parameter to the IPC contract that the main process does not trust for the other three handlers — and the preload-side change in `apps/desktop/src/app/tools/preload.ts` only updated `getMetadata` and `requestBrowserAccess` to accept new args, leaving the other two preload signatures unchanged, which makes the asymmetry visible in three layers.

  </details>

#### Shared lib component depends on desktop-only i18n keys

`libs/importer/src/components/chrome/import-chrome.component.ts:181-205`
**Caught by:** Code quality agent

`translateValidationError` (in `libs/importer`) calls `i18nService.t("chromiumImporterBrowserNotInstalled", ...)` and `i18nService.t("browserAccessDenied")`. Both keys are added only to `apps/desktop/src/locales/en/messages.json` in this PR — they don't exist in `libs/common` or `apps/web`/`apps/browser` locales.

This violates the apps/libs boundary documented in the repo's CLAUDE.md ("Strict boundaries must be maintained between apps and libraries"). Today the shared component is wired only by the desktop callback so the codepath isn't reached in web/browser, but a sibling app that wires `onLoadProfilesFromBrowser` will silently render the raw key. Move the strings into a shared locale, or move the translation step into the desktop caller and have the shared component pass through already-translated messages.

  </details>

#### Inconsistent MAS-flag origin between IPC handlers (duplicate framing)

`apps/desktop/src/main/tools/import/chromium-importer.service.ts:8-25`
**Caught by:** Code quality agent

Three of four IPC handlers (`requestBrowserAccess`, `getAvailableProfiles`, `importLogins`) call `isMacAppStore()` themselves in the main process and ignore any renderer-supplied flag. `getMetadata`, however, accepts `isMas` from the renderer. Pick one. The main-process `isMacAppStore()` utility is the established pattern for this signal; threading the boolean through preload and the renderer for `getMetadata` only adds a parameter to the IPC contract that the main process does not trust for the other three handlers. (Same root issue as the architecture agent's finding above; surfaced here from the renderer-side perspective.)

  </details>

#### `ScopedBrowserAccess::Drop` spawns fire-and-forget tokio task

`apps/desktop/desktop_native/chromium_importer/src/chromium/platform/sandbox.rs:147-164`
**Caught by:** Code quality agent

`Drop` calls `tokio::task::spawn` to send a `stop_access` cleanup. Two concerns:

1. `tokio::task::spawn` panics if no current runtime is available. In the `Drop` path that becomes a double-unwind and aborts the process. The construction sites today are all `async fn resume()` calls, so the runtime is reliably present when the value is constructed and dropped — but a future refactor that constructs/drops `ScopedBrowserAccess` synchronously (e.g., for testing) would silently introduce abort-in-Drop.
2. The returned `JoinHandle` is discarded and the result of `desktop_objc::run_command` is ignored (`let _ = ...`). Cleanup is fire-and-forget; if the runtime tears down before the task runs, the security-scoped resource is not released until process exit.

Recommend an explicit `async fn close(self)` consumed by `import_logins` before returning, with `Drop` retained only as a best-effort fallback gated by `tokio::runtime::Handle::try_current()` so it cannot panic.

  </details>

#### Dead `bookmark` field in `RequestAccessResponse`

`apps/desktop/desktop_native/chromium_importer/src/chromium/platform/sandbox.rs:32-36`
**Caught by:** Code quality agent

```rust
#[derive(Debug, Deserialize)]
struct RequestAccessResponse {
    #[allow(dead_code)]
    bookmark: String,
}
```

The field is parsed from the ObjC response but never read — the only consumer matches `CommandResult::Success { .. }` with `..`. The `#[allow(dead_code)]` annotation papers over a true unused field added by this PR rather than removing it. Either delete the field (the ObjC side already saves the bookmark to `NSUserDefaults`, so there's no need to round-trip it across FFI) or have a consumer actually read it.

  </details>

#### `ScopedBrowserAccess::Drop` calls `tokio::task::spawn` (duplicate framing)

`apps/desktop/desktop_native/chromium_importer/src/chromium/platform/sandbox.rs:144`
**Caught by:** Bug analysis agent

Same `Drop` + `tokio::task::spawn` concern as flagged by the code quality and security agents. The bug-analysis perspective: the `JoinHandle` is discarded and `let _ = desktop_objc::run_command(input_json).await` swallows any error, so any failure to release the security-scoped resource is invisible. In practice the runtime is always present when the value is dropped (because the construction site is `async`), so this lands at refactor severity rather than important. See the `quality-3` finding above for the suggested fix.

  </details>

#### `preload.requestBrowserAccess` signature does not forward `isMas`

`apps/desktop/src/app/tools/preload.ts:12`
**Caught by:** Bug analysis agent

The napi `.d.ts` declares `requestBrowserAccess(browser: string, masBuild: boolean): Promise<void>` with `masBuild` required. The preload bridge sends only `browser`:

```ts
requestBrowserAccess: (browser: string): Promise<void> =>
  ipcRenderer.invoke("chromium_importer.requestBrowserAccess", browser),
```

The ipcMain handler in `chromium-importer.service.ts` calls `chromium_importer.requestBrowserAccess(browser, isMacAppStore())`, so functionally this works today. However, the inconsistent shape — `getMetadata` forwards `isMas` from the renderer, `requestBrowserAccess`/`getAvailableProfiles`/`importLogins` do not — is the same root issue surfaced by the architecture agent. A future change that wires `requestBrowserAccess` to honor a renderer-supplied flag will silently break.

  </details>

#### `Drop` cleanup is fire-and-forget — implicates P05 (controlled access)

`apps/desktop/desktop_native/chromium_importer/src/chromium/platform/sandbox.rs:149`
**Caught by:** Security & logic agent

The only mechanism intended to release the security scope at the end of an import is `Drop` → `tokio::task::spawn` → `stop_access`. The result is discarded (`let _ = ...`), so any failure to release the scope is silent. The doc comment in `import_logins` (`"releases the scope on Drop after the reads complete"`) advertises a contract that is not reliably honored.

The runtime presence guarantee at the construction site means this does not cross the Important bar in practice — but the cleanup pattern is the kind of P05 (Controlled access) concern worth tightening: an explicit awaited close() with the panic-prone `Drop` reduced to a guarded fallback would make the contract observable.

  </details>

## Reviewed and Dismissed

   <details><summary>🔍 5 initial findings dismissed after validation</summary>

#### Strict path-equality check returns generic error on mis-selection

`apps/desktop/desktop_native/objc/src/native/chromium_importer/browser_access_manager.m:65-71`
**Caught by:** Code quality agent
**Original severity:** ⚠️ Important
**Dismissed at:** Step 4 validation
**Dismissed because:** UX nit, not a bug — the Local-State JSON validation provides defense-in-depth and the dialog pre-selects the expected `directoryURL`, so the common path passes the equality check; users selecting a different path getting a generic error is a UX polish item below Refactor severity.

#### Trailing whitespace and out-of-order entry in messages.json

`apps/desktop/src/locales/en/messages.json:4806`
**Caught by:** Code quality agent
**Original severity:** ♻️ Refactor
**Dismissed at:** Step 4 validation
**Dismissed because:** Trailing whitespace in JSON is harmless and a linter would catch it; the out-of-order placement of the new key is sub-Refactor noise that a senior reviewer would not block on.

#### `is_browser_installed` returns true for unknown browser names

`apps/desktop/desktop_native/chromium_importer/src/chromium/platform/sandbox.rs:197`
**Caught by:** Bug analysis agent
**Original severity:** ⚠️ Important
**Dismissed at:** Step 4 validation
**Dismissed because:** The only caller `request_only` validates `browser_name` against `SUPPORTED_BROWSERS` before invoking `is_browser_installed`, and `BROWSER_BUNDLE_IDS` contains the same seven entries as macOS `SUPPORTED_BROWSERS`, so the unknown-name branch is unreachable today; the structural drift risk is already covered by `arch-1`.

#### `stopAccessingBrowser` resolves bookmark to a fresh `NSURL` — start/stop unbalanced

`apps/desktop/desktop_native/objc/src/native/chromium_importer/browser_access_manager.m:154`
**Caught by:** Bug analysis agent
**Original severity:** ⚠️ Important
**Dismissed at:** Step 4 validation
**Dismissed because:** Apple's documented contract for security-scoped resources balances `start`/`stop` calls on the URL but does not unambiguously require the same `NSURL` instance — the underlying counter is on the resolved security scope, and per the validation directive without conclusive proof of an instance-identity requirement this is dismissed.

#### `stopAccessingBrowser` calls stop on different `NSURL` instance — scoped resource not released

`apps/desktop/desktop_native/objc/src/native/chromium_importer/browser_access_manager.m:153`
**Caught by:** Security & logic agent
**Original severity:** ⚠️ Important
**Dismissed at:** Step 4 validation
**Dismissed because:** Same reasoning as the bug-analysis duplicate above — cannot conclusively confirm Apple's contract requires the identical `NSURL` instance for `stopAccessingSecurityScopedResource`; the security-scope counter is associated with the resolved bookmark scope, not strictly the Obj-C object identity.

   </details>


</details>

theMickster:
Hi John 👋🏼 
Trying to reconcile the fact from fiction regarding the Claude findings (so that future engineers benefit from tuning the context that Claude has during coding + code reviewing) has made for an enjoyable morning 🤣

Have you consider or ran into the following?

apps/desktop/resources/entitlements.mas.plist

⚠️  Missing `com.apple.security.files.bookmarks.app-scope` entitlement — the persistent bookmark flow this PR is built around will silently fail in a signed MAS build without it.  Same fix applies to `entitlements.mas.autofill-enabled.plist`.

<details>

Apple's documentation article **"Enabling Security-Scoped Bookmark and URL Access"** states:

> *"If you want to provide your sandboxed app with persistent access to file-system resources, you must enable security-scoped bookmark and URL access."*

It defines two entitlement keys:

| Entitlement key | Description |
|---|---|
| `com.apple.security.files.bookmarks.app-scope` | Enables use of app-scoped bookmarks and URLs. |
| `com.apple.security.files.bookmarks.document-scope` | Enables use of document-scoped bookmarks and URLs. |

The new `BrowserAccessManager` calls `bookmarkDataWithOptions:NSURLBookmarkCreationWithSecurityScope` to create a bookmark and `URLByResolvingBookmarkData:options:NSURLBookmarkResolutionWithSecurityScope` to resume it on subsequent launches. Both operations are gated on `com.apple.security.files.bookmarks.app-scope`.

**What currently exists in both MAS plist files:**
```xml
<key>com.apple.security.files.user-selected.read-write</key>
<true/>
```

`user-selected.read-write` covers the one-time `NSOpenPanel` access. It does **not** cover persisted bookmark creation or resolution. In a real signed MAS binary (not a development-signed build — dev signing does not strictly enforce sandbox entitlements), `bookmarkDataWithOptions:` will return `nil` or an error without the bookmark entitlement. The resume path silently fails and the user is re-prompted on every launch, defeating the entire purpose of the PR.

**Fix — add to both `entitlements.mas.plist` and `entitlements.mas.autofill-enabled.plist`:**
```xml
<key>com.apple.security.files.bookmarks.app-scope</key>
<true/>
```

**How to verify on a real MAS build:**
1. Sign the app with a valid MAS provisioning profile (not a dev certificate).
2. Grant browser access on first import.
3. Quit and relaunch the app.
4. Trigger import again — the picker should **not** appear if the bookmark resolved correctly.

If the picker reappears every launch, the entitlement is missing or not taking effect.

</details>

I tracked down some Apple specific documentation here: [Enabling Security-Scoped Bookmark and URL Access](https://developer.apple.com/documentation/professional-video-applications/enabling-security-scoped-bookmark-and-url-access#overview)

coroiu (approved):
No changes to platform code

mcamirault (approved):

PR #17496, opened 2025-11-19
[PM-28446] Log package types 

quexten:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-28446

## 📔 Objective

We currently lack insight into which sandboxing packaging types are used actively. Specifically, for Linux, we don't know how many users use Snap/Flatpak versus the unsandboxed unsupported types (AppImage, deb, rpm). On Mac, the same applies for DMG/MacAppStore. This makes it hard to gauge where to spend resources / prioritize feature and bug-fix development.

This PR adds insight by adding a new header, "Bitwarden-Package-Type". This adds information that is not already contained in the "Bitwarden-Client-Name" header about specifically which package type is used.

The considered alternative was to add new device entries. However, the migration path here is long and would not give insight into existing installations.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/libs/common/src/services/api.service.ts
+++ b/libs/common/src/services/api.service.ts
@@ -1325,7 +1325,11 @@ export class ApiService implements ApiServiceAbstraction {
       "Bitwarden-Client-Version",
       await this.platformUtilsService.getApplicationVersionNumber(),
     );
-    request.headers.set("Bitwarden-Package-Type", await this.platformUtilsService.packageType());
+
+    const packageType = await this.platformUtilsService.packageType();
+    if (packageType != null) {
+      request.headers.set("Bitwarden-Package-Type", packageType);
+    }
     return this.nativeFetch(request);
   }
 


apps/desktop/src/platform/services/electron-platform-utils.service.ts
```diff
@@ -151,4 +151,20 @@ export class ElectronPlatformUtilsService implements PlatformUtilsService {
   getAutofillKeyboardShortcut(): Promise<string> {
     return null;
   }
+
+  async packageType(): Promise<string> {
```
trmartin4, line 155: Is there a reason why we're using "Unsandboxed" as a default here?  Could we expand this to capture all of the desktop packaging methods that we can, and say "Unknown" as the default instead?

For example, adding `isWindowsPortable`?  Perhaps that's the only missing one? I assume we don't have a way to identify whether the client is `.deb` or `.rpm`, for example?

quexten: Updated. I think deb / rpm detection requires additional research to perform correctly and I felt it is out of scope here.

apps/browser/src/platform/services/platform-utils/browser-platform-utils.service.ts
```diff
@@ -329,6 +329,25 @@ export abstract class BrowserPlatformUtilsService implements PlatformUtilsServic
...
+      case DeviceType.EdgeExtension:
+        return Promise.resolve("Edge Extension");
+      case DeviceType.VivaldiExtension:
+        return Promise.resolve("Vivaldi Extension");
+      case DeviceType.SafariExtension:
+        return Promise.resolve("Safari Extension");
+      default:
+        return Promise.resolve("Browser Extension");
+    }
+  }
```
coroiu: nit(non-blocking): a lot of `Promise.resolve` 🙂

apps/desktop/src/platform/services/electron-platform-utils.service.ts
```diff
@@ -151,4 +151,20 @@ export class ElectronPlatformUtilsService implements PlatformUtilsService {
...
+      packageType = "MacAppStore";
+    } else if (ipc.platform.isWindowsStore) {
+      packageType = "WindowsStore";
+    } else if (ipc.platform.isAppImage) {
+      packageType = "AppImage";
+    } else if (ipc.platform.isSnapStore) {
+      packageType = "Snap";
+    } else if (ipc.platform.isFlatpak) {
+      packageType = "Flatpak";
+    }
```
coroiu: suggestion(non-blocking): feels like assigning `Unknown` in an `else` is slightly easier to read/more semantically correct 🤷‍♂️ 

```suggestion
    let packageType;
    if (ipc.platform.isMacAppStore) {
      packageType = "MacAppStore";
    } else if (ipc.platform.isWindowsStore) {
      packageType = "WindowsStore";
    } else if (ipc.platform.isAppImage) {
      packageType = "AppImage";
    } else if (ipc.platform.isSnapStore) {
      packageType = "Snap";
    } else if (ipc.platform.isFlatpak) {
      packageType = "Flatpak";
    } else {
      packageType = "Unknown";
    }
```

apps/desktop/src/platform/services/electron-platform-utils.service.ts
```diff
@@ -151,4 +151,20 @@ export class ElectronPlatformUtilsService implements PlatformUtilsService {
...
+    return `Bitwarden-Desktop/${await this.getApplicationVersion()} (${packageType})`;
```
coroiu: question: why does the desktop also include the app version when none of the other seem to? Isn't the version already included in `Bitwarden-Client-Version` ?

libs/common/src/services/api.service.ts
```diff
@@ -1325,6 +1325,7 @@ export class ApiService implements ApiServiceAbstraction {
       "Bitwarden-Client-Version",
       await this.platformUtilsService.getApplicationVersionNumber(),
     );
+    request.headers.set("Bitwarden-Package-Type", await this.platformUtilsService.packageType());
```
coroiu: question/issue?: isn't this gonna cause a lot of duplicate data being sent?
```
Bitwarden-Client-Name: web
Bitwarden-Package-Type: web
```

suggestion: why not let `packageType()` return `null`/`undefined` for the clients where we don't need the extra information and skip the header entirely in those cases

apps/desktop/src/platform/services/electron-platform-utils.service.ts
quexten: Apologies, this is left over from attempting to make this part of the user-agent header, which has a very specific format

apps/browser/src/platform/services/platform-utils/browser-platform-utils.service.ts
quexten: Replaced

apps/desktop/src/platform/services/electron-platform-utils.service.ts
quexten: Applied

apps/desktop/src/platform/services/electron-platform-utils.service.ts
quexten: Cleaned up

libs/common/src/services/api.service.ts
quexten: Filtering to non-null values now / returning null for platforms where there is just one detected package type. Note: CLI could theoretically detect packaging but that's not implemented here.

PR #11141, opened 2024-09-18
[PM-6991] Improve reactivity of cipherViews$ observable in cipher service

shane-melton:
## 🎟️ Tracking

[PM-6991](https://bitwarden.atlassian.net/browse/PM-6991)

## 📔 Objective

We have run into race conditions where the vault would attempt to call `cipherService.getAllDecrypted()` before the sync had completed. This would lead to an empty list being returned and showing as an empty vault to the user until they refreshed.

This PR updates the existing `cipherViews$` observable to be reactive to `ciphers$` and `localData$` so that whenever either of those streams update, it will internally call and return the value of `getAllDecrypted()` for any subscribers to `cipherViews$`.  `ciphers$` will always emit after a sync completes so this ensures and subscribers to `cipherViews$` will be notified of an update list of decrypted ciphers.

With that change, the individual web vault and vault filter service now utilize the `cipherViews$` observable instead of wrapping `getAllDecrypted()` in our `asyncToObservable()` helper.

> [!NOTE]
> There is still plenty of room for improvement here. Namely, requiring a consumers to provide a `userId` when fetching decrypted ciphers instead of relying on the `activeUserId` internally. But, for now, this gets us one step closer to making the cipher service observable based.
>
> Also, I left the Browser's Vault filter component alone as it already has some safe guards to avoid the race condition _and_ it is being deprecated with the Browser refresh.

## ⏰ Reminders before review

- Contributor guidelines followed
- All formatters and local linters executed and passed
- Written new unit and / or integration tests where applicable
- Protected functional changes with optionality (feature flags)
- Used internationalization (i18n) for all UI strings
- CI builds passed
- Communicated to DevOps any deployment requirements
- Updated any necessary documentation (Confluence, contributing docs) or informed the documentation team

## 🦮 Reviewer guidelines

- 👍 (`:+1:`) or similar for great changes
- 📝 (`:memo:`) or ℹ️ (`:information_source:`) for notes or general info
- ❓ (`:question:`) for questions
- 🤔 (`:thinking:`) or 💭 (`:thought_balloon:`) for more open inquiry that's not quite a confirmed issue and could potentially benefit from discussion
- 🎨 (`:art:`) for suggestions / improvements
- ❌ (`:x:`) or ⚠️ (`:warning:`) for more significant problems or concerns needing attention
- 🌱 (`:seedling:`) or ♻️ (`:recycle:`) for future improvements or indications of technical debt
- ⛏ (`:pick:`) for minor or nitpick changes

[...]

PR #17505, opened 2025-11-19
[PM-28428] - Improve flickering of items after Autofilling and Adding website to login

jaasen-livefront:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-28428

## 📔 Objective

Issue:
The popup UI was flickering after autofill which was exacerbated by the recent addition of `handleAutofillSuggestionUsed` because the cipher lists (autoFillCiphers$, favoriteCiphers$, remainingCiphers$) re-emitted even when their contents hadn’t actually changed. This caused unnecessary DOM updates and visible flicker in the list components.

Fix:
Added a stable comparator (ciphersEqualByIdAndOrder) and applied `distinctUntilChanged` to the relevant streams. The lists now only re-emit when their IDs or order truly change.

Result:
The popup no longer flickers on background updates, and the UI remains stable while still updating correctly when the data meaningfully changes.

## 📸 Screenshots

https://github.com/user-attachments/assets/e37d0962-a500-43f3-9a59-3862bb32eb69

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/apps/browser/src/vault/popup/services/vault-popup-items.service.ts
+++ b/apps/browser/src/vault/popup/services/vault-popup-items.service.ts
@@ -260,7 +260,7 @@ export class VaultPopupItemsService {
     // Turn loading on whenever we start processing ciphers
     this._ciphersLoading$.pipe(map(() => true)),
     // Turn loading off whenever the decrypted cipher list finishes updating
-    this._allDecryptedCiphers$.pipe(map(() => false)),
+    this.remainingCiphers$.pipe(map(() => false)),
   ).pipe(startWith(true), distinctUntilChanged(), shareReplay({ refCount: false, bufferSize: 1 }));
 
   /** Observable that indicates whether there is search text present.
--- a/apps/browser/src/vault/popup/services/vault-popup-items.service.spec.ts
+++ b/apps/browser/src/vault/popup/services/vault-popup-items.service.spec.ts
@@ -515,24 +515,6 @@ describe("VaultPopupItemsService", () => {
       expect(tracked.emissions[0]).toBe(true);
       expect(tracked.emissions[1]).toBe(false);
     });
-
-    it("should cycle when cipherService.ciphers$ emits even if cipher lists don't re-emit", async () => {
-      await tracked.pauseUntilReceived(2);
-
-      const before = [...tracked.emissions];
-
-      ciphersSubject.next({});
-
-      await tracked.pauseUntilReceived(before.length + 2);
-
-      const after = tracked.emissions;
-
-      expect(after.length).toBeGreaterThan(before.length);
-
-      const tail = after.slice(before.length);
-      expect(tail[0]).toBe(true);
-      expect(tail[1]).toBe(false);
-    });
   });
 
   describe("applyFilter", () => {


apps/browser/src/vault/popup/services/vault-popup-items.service.ts
```diff
@@ -250,15 +252,18 @@ export class VaultPopupItemsService {
...
 
   /**
    * Observable that indicates whether the service is currently loading ciphers.
    */
   loading$: Observable<boolean> = merge(
+    // Turn loading on whenever we start processing ciphers
     this._ciphersLoading$.pipe(map(() => true)),
-    this.remainingCiphers$.pipe(map(() => false)),
+    // Turn loading off whenever the decrypted cipher list finishes updating
+    this._allDecryptedCiphers$.pipe(map(() => false)),
```
nick-livefront: ❓ Is this change needed? After `remainingCiphers$` initially emits `loading$` will always be false and thus the `distinctUntilChanged()` will stop any duplicate emissions. Am I following that correct?

jaasen-livefront: @nick-livefront Ah yes. Here's the issue:

After an autofill event:

_ciphersLoading$ emits true ➜ loading$ becomes true

remainingCiphers$ would emit, but its emission is suppressed by distinctUntilChanged

loading$ never sees a false ➜ loading spinner gets stuck

nick-livefront: I'm following now that we chatted via slack - `remainingCiphers$` now has a `distinctUntilChanged`

apps/browser/src/vault/popup/services/vault-popup-items.service.ts
```diff
@@ -231,34 +231,36 @@ export class VaultPopupItemsService {
...
-      ) =>
-        of(favoriteCiphers).pipe(withLatestFrom(this._filteredCipherList$, this.autoFillCiphers$)),
-    ),
-    map(([favoriteCiphers, ciphers, autoFillCiphers]) =>
+  remainingCiphers$ = combineLatest([
+    this._filteredCipherList$,
+    this.autoFillCiphers$,
+    this.favoriteCiphers$,
+  ]).pipe(
+    map(([ciphers, autoFillCiphers, favoriteCiphers]) =>
```
jaasen-livefront, line 247: Previously, `remainingCiphers$` was only recomputed when `favoriteCiphers$` emitted to avoid an older race/flicker issue. With the new equality comparator and the current dependency graph, that pattern causes stale data. We now use combineLatest over all inputs and rely on ciphersEqualByIdAndOrder to prevent unnecessary emissions

apps/browser/src/vault/popup/services/vault-popup-items.service.ts
```diff
@@ -398,3 +400,29 @@ export class VaultPopupItemsService {
 const waitUntilSync = <T>(syncService: SyncService): MonoTypeOperatorFunction<T> => {
   return waitUntil(syncService.activeUserLastSync$().pipe(filter((lastSync) => lastSync != null)));
 };
+
+const ciphersEqualByIdAndOrder = (
```
shane-melton: ⚠️ This equality operator will exclude emissions of legitimate cipher modifications too. For example, editing a cipher name and then going back will result in stale data being displayed _until_ the order of ciphers is modified (e.g. via search). We'll need come up with another approach 🤔 

https://github.com/user-attachments/assets/72e57523-ce5e-48c0-aa63-b273672acaef

shane-melton (changes requested):

apps/browser/src/vault/popup/services/vault-popup-items.service.ts
```diff
@@ -398,3 +404,34 @@ export class VaultPopupItemsService {
...
+
+    // Different ID → changed
+    if (ac?.id !== bc?.id) {
+      return false;
+    }
+
+    // Same ID but different object reference → changed cipher → re-emit
+    if (ac !== bc) {
+      return false;
+    }
```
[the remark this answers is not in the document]
jaasen-livefront: This is just patently incorrect. Returning `false` WILL cause a re-emission.

apps/browser/src/vault/popup/services/vault-popup-items.service.ts
jaasen-livefront: @shane-melton Fixed in https://github.com/bitwarden/clients/pull/17505/commits/8814eed9924ee21571b02e9c6804b03b7f809adc

libs/common/src/vault/services/cipher.service.ts
```diff
@@ -178,10 +229,9 @@ export class CipherService implements CipherServiceAbstraction {
   cipherViews$ = perUserCache$((userId: UserId): Observable<CipherView[] | null> => {
     return combineLatest([
       this.encryptedCiphersState(userId).state$,
-      this.localData$(userId),
```
nick-livefront, line 181: ❓ `cipherViews$` is used throughout the applications for other areas to retrieve ciphers. They could be dependent on the ciphers being updated when the local data changes. Are there side effects that we could be introducing from this change?

It looks like it was originally put in place to avoid race conditions with the sync service: https://github.com/bitwarden/clients/pull/11141

jaasen-livefront: @nick-livefront Yeah I wasn't super committed to these changes. It was mostly an experiment to see if I could reduce the number of emissions. The issue is that when we autofill and save we're updating both `localData$` with the `lastUsedDate` as well as the cipher login `uris`. I'm going to revert these changes.

apps/browser/src/vault/popup/services/vault-popup-items.service.ts
shane-melton: Is this still necessary with the changes made to `ciphersEqualByIdAndOrder`?

jaasen-livefront: @shane-melton It is since we can't rely on `remainingCiphers$` to re-emit after an autofill. The change made was to fix a stale data issue after updating cipher data.

jaasen-livefront: @shane-melton Actually you're completely correct. I've reverted this back to the original implementation.

jaasen-livefront:
> Is the root cause of the issue caused by the cipher list being updated more than once after an autofill and save?
> 
> In a larger vault I see a few loading states that I'm trying to track down why they're occurring. in `doAutofillAndSave` [_internalDoAutofill](https://github.com/bitwarden/clients/blob/577090e6e4347e74e30776acf6383348103d0a6a/apps/browser/src/vault/popup/services/vault-popup-autofill.service.ts#L378) updates the local data / last used date, and [_saveNewUri](https://github.com/bitwarden/clients/blob/577090e6e4347e74e30776acf6383348103d0a6a/apps/browser/src/vault/popup/services/vault-popup-autofill.service.ts#L389) updates the cipher with the server.
> 
> If we combine those side effects into a single use case does that help remedy the issue? ( I can give it a shot, just wanted to get my thoughts out there now)
> 
> On larger vaults I'm seeing two separate loading states occur, which I don't believe we want to happen if we can avoid it.
> 
>  Screen.Recording.2025-11-25.at.4.29.29.PM.mov

@nick-livefront You're exactly right. We're updating both the `ciphers$` state and the `localData$` state. depends on both of those I don't see a way to combine these into a single observable without either breaking the existing implementation or completely rewriting it.

PR #16136, opened 2025-08-23
[BEEEP|PM-25164] Prevent memory dumping on renderer on Linux

quexten:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-25164
## 📔 Objective

This PR prevents all memory dumping, debugger access, environment variable snooping for all electron processes on Bitwarden desktop on Linux (except snap).

On Linux, userspace programs can read memory of other processes running under the same user. This can be prevented with [prctl PR_SET_DUMPABLE](https://man7.org/linux/man-pages/man2/pr_set_dumpable.2const.html). We already do this for the main process in desktop_native (https://github.com/bitwarden/clients/pull/9393), but this call only affects the current process, and we can only call it in `main`. We cannot call this in renderer / GPU, leaving these processes open currently.

A workaround is injecting code into the electron processes, to run the isolation calls from the otherwise inaccessible processes.

This PR adds a Rust library, that is compiled to a shared object, that is preloaded on all processes, including GPU and renderer. When loaded, it uses PR_SET_DUMPABLE to prevent all debugger access from user space and prevents keys or data from being accessed by other user-space programs.

On flatpak, hooking the unsetenv is needed to restore the preload env variable after zypak clears it. This is a bug in zypak. Snap is not included since there is an open bug in portals that prevents file access when PR_SET_DUMPABLE is used.

A test script that verifies the isolation status is also included.

Note: This duplicates to a small degree the isolation code in `desktop_native`, that only covers `main` - but not renderer. I intend to explore other solutions for mac / windows, which are better rolled out via `desktop_native` behind a feature flag first, but we could apply the same approach long-term.

(On windows we can achieve something similar via DACL: https://github.com/niluxv/secmem-proc)

## ⏰ Reminders before review

- Contributor guidelines followed
- All formatters and local linters executed and passed
- Written new unit and / or integration tests where applicable
- Protected functional changes with optionality (feature flags)
- Used internationalization (i18n) for all UI strings
- CI builds passed
- Communicated to DevOps any deployment requirements
- Updated any necessary documentation (Confluence, contributing docs) or informed the documentation team

## 🦮 Reviewer guidelines

- 👍 (`:+1:`) or similar for great changes
- 📝 (`:memo:`) or ℹ️ (`:information_source:`) for notes or general info
- ❓ (`:question:`) for questions
- 🤔 (`:thinking:`) or 💭 (`:thought_balloon:`) for more open inquiry that's not quite a confirmed issue and could potentially benefit from discussion
[...]

PR #17509, opened 2025-11-20
[PM-28220] Fix desktop deb startup failure

dereknance:
## 🎟️ Tracking

[PM-28220](https://bitwarden.atlassian.net/browse/PM-28220)

## 📔 Objective

This PR reverts a change made to the desktop client in #16136 as it unfortunately resulted in the client failing to start for `deb` builds.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

[PM-28220]: https://bitwarden.atlassian.net/browse/PM-28220?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/desktop/resources/linux-wrapper.sh
+++ b/apps/desktop/resources/linux-wrapper.sh
@@ -12,14 +12,6 @@ if [ -e "/usr/lib/x86_64-linux-gnu/libdbus-1.so.3" ]; then
   export LD_PRELOAD="/usr/lib/x86_64-linux-gnu/libdbus-1.so.3"
 fi
 
-# If running in non-snap, add libprocess_isolation.so from app path to LD_PRELOAD
-# This prevents debugger / memory dumping on all desktop processes
-if [ -z "$SNAP" ] && [ -f "$APP_PATH/libprocess_isolation.so" ]; then
-  LIBPROCESS_ISOLATION_SO="$APP_PATH/libprocess_isolation.so"
-  LD_PRELOAD="$LIBPROCESS_ISOLATION_SO${LD_PRELOAD:+:$LD_PRELOAD}"
-  export LD_PRELOAD
-fi
-
 PARAMS="--enable-features=UseOzonePlatform,WaylandWindowDecorations --ozone-platform-hint=auto"
 if [ "$USE_X11" = "true" ]; then
   PARAMS=""


apps/desktop/package.json
```diff
@@ -35,10 +35,9 @@
     "build:renderer:watch": "cross-env NODE_ENV=development webpack --config webpack.config.js --config-name renderer --watch",
     "electron": "node ./scripts/start.js",
     "electron:ignore": "node ./scripts/start.js --ignore-certificate-errors",
-    "flatpak:dev": "npm run clean:dist && electron-builder --dir -p never && flatpak-builder --force-clean --install --user ../../.flatpak/ ./resources/com.bitwarden.desktop.devel.yaml && flatpak run com.bitwarden.desktop",
```
quexten, line 38: Please do not revert this. This is an unrelated change that introduces build scripts that we do want to keep.

apps/desktop/package.json
```diff
@@ -35,10 +35,9 @@
...
     "clean:dist": "rimraf ./dist",
     "pack:dir": "npm run clean:dist && electron-builder --dir -p never",
-    "pack:lin:flatpak": "flatpak-builder --repo=../../.flatpak-repo ../../.flatpak  ./resources/com.bitwarden.desktop.devel.yaml --install-deps-from=flathub --force-clean &&  flatpak build-bundle ../../.flatpak-repo/ ./dist/com.bitwarden.desktop.flatpak com.bitwarden.desktop",
+    "pack:lin:flatpak": "npm run clean:dist && electron-builder --dir -p never && flatpak-builder  --repo=build/.repo build/.flatpak  ./resources/com.bitwarden.desktop.devel.yaml --install-deps-from=flathub --force-clean &&  flatpak build-bundle ./build/.repo/ ./dist/com.bitwarden.desktop.flatpak com.bitwarden.desktop",
```
quexten: Please do not revert this, this updates the build script for flatpak to reduce significant performance issues, and we do not want to undo this.

apps/desktop/resources/linux-wrapper.sh
```diff
@@ -7,19 +7,12 @@ ulimit -c 0
...
-if [ -e "/usr/lib/x86_64-linux-gnu/libdbus-1.so.3" ]; then
+# force use of base image libdus in snap
+if [ -e "/usr/lib/x86_64-linux-gnu/libdbus-1.so.3" ]
+then
   export LD_PRELOAD="/usr/lib/x86_64-linux-gnu/libdbus-1.so.3"
 fi
 
-# If running in non-snap, add libprocess_isolation.so from app path to LD_PRELOAD
-# This prevents debugger / memory dumping on all desktop processes
-if [ -z "$SNAP" ] && [ -f "$APP_PATH/libprocess_isolation.so" ]; then
```
quexten, line 17: Can we consider only reverting this part specifically? There is no reason to undo the entire change, and supposedly this issue does not even affect flatpak, so there is no reason to undo the change for flatpak. 

Undoing only these lines would mean that the change becomes inactive for now until we can figure out *why* the change causes launch failures.

Undoing the change entirely means that we re-introduce risk of the the renderer process being dumped, and passwords being stolen from memory.

apps/desktop/desktop_native/process_isolation/test_isolation.sh
```diff
@@ -1,40 +0,0 @@
-#!/bin/bash
-
-# This script tests the memory isolation status of bitwarden-desktop processes. The script will print "isolated"
-# if the memory is not accessible by other processes.
-
-CURRENT_USER=$(whoami)
```
quexten, line 6: Please don't undo this test script.

apps/desktop/resources/linux-wrapper.sh
trmartin4: @quexten I think I understand we have two options:
1. We undo this change, which means that we allow debugger / memory dumping on AppImage, deb, and rpm packages, but allows the app to run without manual configuration.
2. We keep the change in, but instruct users of these packages to run `sudo chmod 4755 /opt/Bitwarden/chrome-sandbox`.

For option 1, we need a follow-up task to re-enable it after understanding more.  For option 2, we may or may not depending on whether that manual command is acceptable for the foreseeable future.

Is that right?  Are there considerations you'd throw in the mix for picking which one?

cc: @dereknance

quexten: I would at least not want to disable it on flatpak unnecessarily because flatpak is not affected by the bug. This PR currently disables it on flatpak too.

IMO if the trade of is the app completely not launching, then we should disable it for appimage, rpm, deb, accepting that those then have weaker protections, with the option to bring it back at a later point in time for these targets, should we choose to do so. I don't know if the `chmod` is actually necessary, and would have to further dig into the problem to come up with an answer here.

For example, if apparmor is causing these issues, then it is possible that we could detect apparmor being present and then *only on systems with apparmor* disable the memory isolation.

Either way, please just don't remove the isolations on flatpak / only disable the above linked code that sets the LD_PRELOAD for AppImage / RPM / DEB

trmartin4: @quexten yes, sorry, that's what I meant.  By "this change", I meant only the targeted change that you proposed in the above-linked code.  Thank you for the clarification!

dereknance: >Can we consider only reverting this part specifically?

Of course! Fixed in 3c9e7f3d

quexten (approved):
Thank you @dereknance. Changes LGTM. I *may* in the future debug this further, but it's not a high priority for me as long as flatpak remains having the protections. 

(Again a further argument to guide people towards the flatpak installation).

tl87:
Not sure if this is related? but I am using the AppImage and I am unable to start it:

```bash
$ chmod +x Bitwarden-2025.11.1-x86_64.AppImage && ./Bitwarden-2025.11.1-x86_64.AppImage

[5979:1123/060915.546424:FATAL:sandbox/linux/suid/client/setuid_sandbox_host.cc:169] The SUID sandbox helper binary was found, but is not configured correctly. Rather than run without sandboxing I'm aborting now. You need to make sure that /tmp/.mount_BitwarwIrocT/chrome-sandbox is owned by root and has mode 4755.
/tmp/.mount_BitwarwIrocT/bitwarden: line 28:  5979 Trace/breakpoint trap      $APP_PATH/bitwarden-app $PARAMS "$@"
```

PR #17510, opened 2025-11-20
PM-21800 calling copyToClipboard function from PlatformUtilsService in order to trim copied string

bmbitwarden:
…ts of clipboard

## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-21800

## 📔 Objective

need to paste from generator without extra spaces with copyToClipboard function from PlatformUtilsService

## 📸 Screenshots

https://github.com/user-attachments/assets/35521c90-6cd6-48a5-a06e-0850663d47de

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


libs/components/src/color-password/color-password.component.ts
```diff
@@ -78,4 +78,17 @@ export class ColorPasswordComponent {
 
     return "letter";
   }
+
+  @HostListener("copy", ["$event"])
+  onCopy(event: ClipboardEvent) {
```
vleague2, line 95: The `PlatformUtilsService` class has a `copyToClipboard` method already, which has different implementations depending on what app you are in because browser, web, and desktop all have slightly different APIs for that. Would that work for the bug you're solving here?

vleague2 (approved):
Code change and Storybook look good. I didn't test in the clients since this already has the QA tag on it. 

Could you update the PR title and description since the implementation has changed? Thanks!

vleague2 (commented):
@bmbitwarden ~Testing this in Storybook, I can't actually seem to copy the password anymore 🤔 I am seeing the selection state and trying in different ways (clicking once, clicking twice, click and drag over selection) and using keyboard shortcut and right-click copy, but the value is not being copied. Are you seeing it work as expected on your computer?~ Omg, ignore me, I forgot the implementation is a mock in Storybook. Ugh Mondays

libs/components/src/color-password/color-password.component.ts
```diff
@@ -40,7 +43,13 @@ export class ColorPasswordComponent {
 
   @HostBinding("class")
   get classList() {
-    return ["tw-min-w-0", "tw-whitespace-pre-wrap", "tw-break-words"];
+    return [
+      "tw-inline-block", // creates a formatting boundary for clipboard issue
+      "tw-select-all", // constrains selection to this element
```
vleague2: I think this is a nice QOL thing if the user was always going to want the full selection and knew to one-click select. But if you are trying to do the normal "drag to select" behavior, especially if you only are trying to select a portion of the text, then it has a buggy appearance and is difficult to use. The Github issue for this bug does talk about partial text selection, so I am thinking maybe we don't want to assume that the full selection is always desirable. What do you think about removing this class to use the default browser text selection behavior?

https://github.com/user-attachments/assets/142d7dfa-8630-472c-b68f-32199eda2cf1

libs/components/src/color-password/color-password.stories.ts
```diff
@@ -9,6 +11,19 @@ const examplePassword = "Wq$Jk😀7jlI  DX#rS5Sdi!z0O ";
...
   title: "Component Library/Color Password",
   component: ColorPasswordComponent,
+  decorators: [
+    applicationConfig({
+      providers: [
+        {
+          provide: PlatformUtilsService,
+          useValue: {
+            // eslint-disable-next-line
+            copyToClipboard: (text: string) => console.log(`${text} copied to clipboard`),
```
harr1424, line 21: I don't think we want to send any `console.log()` statements to production unless there is a very good reason, I think this is an artifact of local development and should be removed.

libs/components/src/stories/kitchen-sink/kitchen-sink.stories.ts
```diff
@@ -71,6 +72,13 @@ export default {
             });
           },
         },
+        {
+          provide: PlatformUtilsService,
+          useValue: {
+            // eslint-disable-next-line
+            copyToClipboard: (text: string) => console.log(`${text} copied to clipboard`),
```
harr1424, line 79: I don't think we want to send any `console.log()` statements to production unless there is a very good reason, I think this is an artifact of local development and should be removed.

libs/components/src/color-password/color-password.stories.ts
```diff
@@ -9,6 +11,19 @@ const examplePassword = "Wq$Jk😀7jlI  DX#rS5Sdi!z0O ";
...
   title: "Component Library/Color Password",
   component: ColorPasswordComponent,
+  decorators: [
+    applicationConfig({
+      providers: [
+        {
+          provide: PlatformUtilsService,
+          useValue: {
+             
+            copyToClipboard: (text: string) => {},
```
harr1424: I think we can remove the empty line (20), and am curious why this no-op is necessary? It might be good to document with a comment why this is happening.

libs/components/src/stories/kitchen-sink/kitchen-sink.stories.ts
```diff
@@ -71,6 +72,13 @@ export default {
             });
           },
         },
+        {
+          provide: PlatformUtilsService,
+          useValue: {
+             
+            copyToClipboard: (text: string) => {},
```
harr1424: I think we can remove the empty line (78), and am curious why this no-op is necessary? It might be good to document with a comment why this is happening.

libs/components/src/color-password/color-password.component.ts
bmbitwarden: without that line, i've been having trouble with preventing the Options label from being included when the user clicks in succession in order to highlight the password

https://github.com/user-attachments/assets/2332436f-67f4-4429-a182-1a8087a136cd

libs/components/src/color-password/color-password.stories.ts
vleague2: The console logs were fine imo. These are just mocks for Storybook, so we need Something to happen for this platform service functionality. The console logs provide some kind of feedback for QA-ing that the function is being called.

harr1424: Ah I see now, my apologies!

libs/components/src/color-password/color-password.component.ts
vleague2: Hmm odd. I realized I am unable to reproduce that issue (where "Options" get copied) in prod, so I tested your branch with the new copy event code commented out, and with it added back in. Using the default copy event handler, it does not copy "Options", but using the new code with `window.getSelection()` does copy "Options." So I am inclined to think this is actually an issue with `window.getSelection()`. Some Googling brought me to a Chromium issue that sounds similar but is marked as fixed, but maybe it will give some areas to look into to figure out the underlying problem: https://issues.chromium.org/issues/41393366 

I think we should try to get the selection function to behave as expected instead of using the css route, since the css route disables some other functionality (partial selection) and that seems suboptimal to me.

bmbitwarden: ok sounds good - i refactored the hostlistener function so it selects all the password characters and iterates through each of them

image

libs/components/src/color-password/color-password.component.ts
```diff
@@ -14,7 +23,8 @@ type CharacterType = "letter" | "emoji" | "special" | "number";
 @Component({
   selector: "bit-color-password",
   template: `@for (character of passwordCharArray(); track $index; let i = $index) {
-    <span [class]="getCharacterClass(character)" class="tw-font-mono">
+    <!-- eslint-disable-next-line tailwindcss/no-custom-classname -->
+    <span [class]="getCharacterClass(character)" class="tw-font-mono password-character">
```
vleague2: Instead of a custom class, can we use a data attribute? (See `chip-select.component.html` or `nav-item.component.html` for an example with the `data-fvw` attribute.) That way, we can keep classes as styling-only and you wouldn't need the eslint override.

bmbitwarden: ok cool - I just made the switch the data attribute

vleague2 (approved):
Not sure what's going on with the unit test workflow, but the code looks good to me. Thanks for your patience with all those changes!

PR #17651, opened 2025-11-25
[BEEEP][PM-29441] Introduce bitwarden-encrypted-json-importer

djsmith85:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-29441

## 📔 Objective

An effort to introduce type guards and split the logic between the differently protected bitwarden-json import-formats.

Introduces a BitwardenEncryptedJsonImporter in between the `BitwardenPasswordProtectedJsonImporter` and the `BitwardenJsonImporter`

### New Hierarchy
`BitwardenPasswordProtectedJsonImporter` extends `BitwardenEncryptedJsonImporter` extends `BitwardenJsonImporter` extends `BaseImporter`

### Changes
- If a file is password-protected the `BitwardenPasswordProtectedJsonImporter` requests the password and decrypts the payload. This now decrypted text is passed down to the `BitwardenEncryptedJsonImporter` which checks the type and then passes it on to the `BitwardenJsonImporter`.
- AccountEncrypted or UnEncrypted is passed to `BitwardenEncryptedJsonImporter` and if it is encrypted, will be handled there or passed down the `BitwardenJsonImporter`.

The design could be better, if we could inject/instantiate the importers as needed, as technically a decrypted password-protected file could be directly passed to the `BitwardenJsonImporter` (unencrypted parser) and it would need one less `JSON.parse()`. I decide to extend the existing model for now, as the benefit was greater separation of concerns and readability due to less code per file.

### ⚠️  Behaviour change
There's one change in behaviour which should be reviewed/discussed. Instead of using the import target (`this.organizationId`) to decide to parse collections or folders, it now checks which input format was provided (Org or Indivdiual). Best seen on the [diff](https://github.com/bitwarden/clients/compare/tools/pm-xxxx/introduce-bitwarden-encrypted-json-importer?expand=1#diff-8114eeacef8e1c9e821d4b5316d48ee0c840c32d1b5058894265b2b3a608a666R43) in `BitwardenJsonImporter`, but the conditional is also present in the `BitwardenEncryptedJsonImporter`

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/apps/browser/src/auth/services/extension-device-management-component.service.ts
+++ b/apps/browser/src/auth/services/extension-device-management-component.service.ts
@@ -3,9 +3,7 @@ import { DeviceManagementComponentServiceAbstraction } from "@bitwarden/angular/
 /**
  * Browser extension implementation of the device management component service
  */
-export class ExtensionDeviceManagementComponentService
-  implements DeviceManagementComponentServiceAbstraction
-{
+export class ExtensionDeviceManagementComponentService implements DeviceManagementComponentServiceAbstraction {
   /**
    * Don't show header information in browser extension client
    */
--- a/apps/browser/src/autofill/fido2/services/browser-fido2-user-interface.service.ts
+++ b/apps/browser/src/autofill/fido2/services/browser-fido2-user-interface.service.ts
@@ -120,9 +120,7 @@ export type BrowserFido2ParentWindowReference = chrome.tabs.Tab;
  * Browser implementation of the {@link Fido2UserInterfaceService}.
  * The user interface is implemented as a popout and the service uses the browser's messaging API to communicate with it.
  */
-export class BrowserFido2UserInterfaceService
-  implements Fido2UserInterfaceServiceAbstraction<BrowserFido2ParentWindowReference>
-{
+export class BrowserFido2UserInterfaceService implements Fido2UserInterfaceServiceAbstraction<BrowserFido2ParentWindowReference> {
   constructor(private authService: AuthService) {}
 
   async newSession(
--- a/apps/browser/src/autofill/overlay/notifications/content/overlay-notifications-content.service.ts
+++ b/apps/browser/src/autofill/overlay/notifications/content/overlay-notifications-content.service.ts
@@ -15,9 +15,7 @@ import {
   OverlayNotificationsExtensionMessageHandlers,
 } from "../abstractions/overlay-notifications-content.service";
 
-export class OverlayNotificationsContentService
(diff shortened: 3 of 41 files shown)


libs/importer/src/importers/bitwarden/bitwarden-encrypted-json-importer.ts
```diff
@@ -0,0 +1,188 @@
...
+  }
+
+  async parse(data: string): Promise<ImportResult> {
+    const results: BitwardenJsonExport = JSON.parse(data);
+    if (results == null || results.items == null) {
+      const result = new ImportResult();
+      result.success = false;
+      return result;
+    }
+
```
mcamirault, line 63: 🎨 I'd adjust the type on line 48 above to be `BitwardenPasswordProtectedFileFormat | BitwardenJsonExport` and add a check here using `isPasswordProtected()` the way `BitwardenJsonImporter` uses `isUnencrypted()` on line 34, so that they both check the type from the next level up in the chain. Or we can remove the check on 34, I have no preference but for consistency between the two

libs/importer/src/importers/bitwarden/bitwarden-encrypted-json-importer.ts
```diff
@@ -0,0 +1,188 @@
...
+    if (isUnencrypted(results)) {
+      return super.parse(data);
+    }
+
+    return await this.parseEncrypted(results);
+  }
+
+  private async parseEncrypted(data: BitwardenEncryptedJsonExport): Promise<ImportResult> {
+    const account = await firstValueFrom(this.accountService.activeAccount$);
+
```
mcamirault, line 74: ⛏️ We can construct `result` once up here instead of multiple times below

libs/importer/src/importers/bitwarden/bitwarden-encrypted-json-importer.ts
djsmith85: Addressed with 09f68e3a171d6e1525461cde25bb12a15fec97fa

libs/importer/src/importers/bitwarden/bitwarden-encrypted-json-importer.ts
djsmith85: Addressed with d12c0989339c961fa8606f3318a80a1e9cd799ef

mcamirault (dismissed):

mcamirault (approved):

PR #17658, opened 2025-11-25
[PM-27792] Scaffold layout desktop migration

iivins-livefront:
## Overview

The goal of this PR is to introduce foundational scaffolding for the Bitwarden Desktop application UI migration

The goal of the PR being created in a forked repo is to unblock the migration effort while we wait for contribution permissions

## What Changed

Three new layout components were created to establish a hierarchical structure for the desktop application:
1. DesktopLayoutComponent - A standalone component serving as the base layout wrapper, integrating Angular Router and Bitwarden's component library navigation modules.
2. UserLayoutComponent - The primary authenticated user interface shell that incorporates the DesktopLayout and includes the Bitwarden Password Manager logo. This component acts as the parent container for authenticated routed.
3. DesktopSideNavComponent - A navigation sidebar component for user-facing screens.

The app-routing.module.ts received an update to support the new layout system. A new route group was added using UserLayoutComponent as the parent, with child routes for:
- /new-vault - pointing to the new VaultComponent (vault-v3)
- /new-sends - pointing to the new SendsComponent (send-v2)

The existing vault route now includes a feature flag guard (canAccessFeature) that checks DesktopUiMigrationMilestone1 to conditionally redirect users between the legacy vault (VaultV2Component) and the new vault implementation.

Comprehensive test files (.spec.ts) were added for all three new layout components plus the sends and vault components

Minor locale changes to match designs for navigation copy

### DIFF
--- a/apps/desktop/src/app/tools/send-v2/send-v2.component.spec.ts
+++ b/apps/desktop/src/app/tools/send-v2/send-v2.component.spec.ts
@@ -1,6 +1,6 @@
 import { ComponentFixture, TestBed } from "@angular/core/testing";
 
-import { SendV2Component } from "./send-V2.component";
+import { SendV2Component } from "./send-v2.component";
 
 describe("SendV2Component", () => {
   let component: SendV2Component;


apps/desktop/src/app/layout/desktop-layout.component.ts
```diff
@@ -0,0 +1,14 @@
+import { CommonModule } from "@angular/common";
+import { ChangeDetectionStrategy, Component } from "@angular/core";
+import { RouterModule } from "@angular/router";
+
+import { LayoutComponent, NavigationModule } from "@bitwarden/components";
+
+@Component({
+  selector: "app-desktop-layout",
```
Hinton: **suggestion:** Generally `app` prefixes in selectors refers to the current app. I think for consistency we should use `app-layout`.

```suggestion
  selector: "app-layout",
```

---

@willmartian I wonder if we should revisit this prefix, I like that application specific components & services are named after the application because it makes it easier to find and more clear. But it creates a disconnect between the selector and component name which I dislike. I would be open to having more specific selectors like `web`, `browser` and `desktop` over `app`. (Not in scope for this work)

apps/desktop/src/app/layout/desktop-layout.component.ts
```diff
@@ -0,0 +1,14 @@
...
+  standalone: true,
+  imports: [CommonModule, RouterModule, LayoutComponent, NavigationModule],
```
Hinton: **issue:** I don't believe most of these imports are required. This component does not use router or navigation so those can be skipped. We might need Common depending on how angular implements `ng-content`, but I suspect not.

apps/desktop/src/app/layout/desktop-layout.module.ts
Hinton, line 1: **issue:** Let's skip any modules if we can avoid them. In this case using standalone and importing the various used components seems better.

apps/desktop/src/app/layout/desktop-side-nav.component.ts
```diff
@@ -0,0 +1,15 @@
+import { CommonModule } from "@angular/common";
+import { ChangeDetectionStrategy, Component, input } from "@angular/core";
+
+import { NavigationModule, SideNavVariant } from "@bitwarden/components";
+
+@Component({
+  selector: "app-desktop-side-nav",
+  standalone: true,
```
Hinton: **suggestion:** We don't need to write out standalone true, and most of the codebase aligns on it not being there.

```suggestion
```

apps/desktop/src/app/layout/desktop-layout.component.ts
```diff
@@ -0,0 +1,14 @@
+import { CommonModule } from "@angular/common";
+import { ChangeDetectionStrategy, Component } from "@angular/core";
+import { RouterModule } from "@angular/router";
+
+import { LayoutComponent, NavigationModule } from "@bitwarden/components";
+
+@Component({
+  selector: "app-desktop-layout",
+  standalone: true,
```
Hinton: **suggestion:** We don't need to write out standalone true, and most of the codebase aligns on it not being there.

```suggestion
```

apps/desktop/src/app/layout/user-layout.component.html
Hinton, line 1: **suggestion:** I think we can skip the user-layout completely since we do not have a organization context in this application. Lifting this into the desktop layout should remove one level of abstraction.

apps/desktop/src/vault/app/vault-v3/vault.component.ts
```diff
@@ -0,0 +1,10 @@
+import { ChangeDetectionStrategy, Component } from "@angular/core";
+
+@Component({
+  selector: "app-vault-v3",
+  standalone: true,
```
Hinton:
```suggestion
```

apps/desktop/src/app/app-routing.module.ts
```diff
@@ -99,7 +104,10 @@ const routes: Routes = [
   {
     path: "vault",
     component: VaultV2Component,
-    canActivate: [authGuard],
+    canActivate: [
+      authGuard,
+      canAccessFeature(FeatureFlag.DesktopUiMigrationMilestone1, false, "new-vault", false),
```
Hinton, line 109: Nice.

apps/desktop/src/app/tools/send-v2/sends.component.ts
```diff
@@ -0,0 +1,10 @@
+import { Component, ChangeDetectionStrategy } from "@angular/core";
+
+@Component({
+  selector: "app-sends-v2",
+  standalone: true,
```
Hinton:
```suggestion
```

apps/desktop/src/app/layout/desktop-layout.component.html
```diff
@@ -0,0 +1,4 @@
+<bit-layout>
+  <ng-content select="app-desktop-side-nav, [slot=side-nav]" slot="side-nav"></ng-content>
```
Hinton: **thought:** I wonder if we should just skip the `app-desktop-side-nav` selector and only use `slot=side-nav`. I know we have it in web but I also saw that we specify both selectors in this PR which is unnecessary.

apps/desktop/src/app/layout/user-layout.component.html
iivins-livefront: When I looked at the web implementation for that abstraction layer - the scope of what it was doing, and what the desktop app _would eventually_ be doing was a bit unclear - thought it wouldn't hurt to add but I can remove it nonetheless 👍

apps/desktop/src/app/layout/desktop-layout.component.ts
iivins-livefront: With the changes to the abstraction layer and this component - we're expecting to have 5 in the import array, all of which get used

```
RouterModule from @angular/router
JslibModule from @bitwarden/angular/jslib.module
LayoutComponent from @bitwarden/components
NavigationModule from @bitwarden/components
DesktopSideNavComponent from ./desktop-side-nav.component
```

apps/desktop/src/app/layout/desktop-layout.component.ts
```diff
@@ -0,0 +1,18 @@
...
+
+import { JslibModule } from "@bitwarden/angular/jslib.module";
+import { PasswordManagerLogo } from "@bitwarden/assets/svg";
+import { LayoutComponent, NavigationModule } from "@bitwarden/components";
+
+import { DesktopSideNavComponent } from "./desktop-side-nav.component";
+
+@Component({
+  selector: "app-desktop-layout",
+  imports: [RouterModule, JslibModule, LayoutComponent, NavigationModule, DesktopSideNavComponent],
```
Hinton: You can use `import { I18nPipe } from "@bitwarden/ui-common";` instead. We're trying to move away from `JslibModule`.

apps/desktop/src/app/layout/desktop-layout.component.ts
```diff
@@ -0,0 +1,18 @@
...
+import { RouterModule } from "@angular/router";
+
+import { JslibModule } from "@bitwarden/angular/jslib.module";
+import { PasswordManagerLogo } from "@bitwarden/assets/svg";
+import { LayoutComponent, NavigationModule } from "@bitwarden/components";
+
+import { DesktopSideNavComponent } from "./desktop-side-nav.component";
+
+@Component({
+  selector: "app-desktop-layout",
```
Hinton:
```suggestion
  selector: "app-layout",
```

apps/desktop/src/app/layout/desktop-side-nav.component.ts
```diff
@@ -0,0 +1,14 @@
+import { CommonModule } from "@angular/common";
+import { ChangeDetectionStrategy, Component, input } from "@angular/core";
+
+import { NavigationModule, SideNavVariant } from "@bitwarden/components";
+
+@Component({
+  selector: "app-desktop-side-nav",
```
Hinton:
```suggestion
  selector: "app-side-nav",
```

apps/desktop/src/app/tools/send-v2/sends.component.ts
```diff
@@ -0,0 +1,9 @@
+import { Component, ChangeDetectionStrategy } from "@angular/core";
+
+@Component({
+  selector: "app-sends-v2",
+  imports: [],
+  template: "<p>Sends V2 Component</p>",
+  changeDetection: ChangeDetectionStrategy.OnPush,
+})
+export class SendsComponent {}
```
Hinton: We actually call these `send` in other places. For consistency let's call this SendV2Component`.

```suggestion
export class SendV2Component {}
```

Please also rename the file to `send-v2.component.ts`.

apps/desktop/src/app/layout/desktop-layout.component.ts
iivins-livefront: Updated for desktop-layout.component - not resolving yet with continued convo

apps/desktop/src/app/tools/send-v2/send-v2.component.ts
```diff
@@ -0,0 +1,9 @@
+import { Component, ChangeDetectionStrategy } from "@angular/core";
+
+@Component({
+  selector: "app-sends-v2",
```
Hinton:
```suggestion
  selector: "app-send-v2",
```

apps/desktop/src/app/tools/send-v2/send-v2.component.spec.ts
```diff
@@ -0,0 +1,22 @@
+import { ComponentFixture, TestBed } from "@angular/core/testing";
+
+import { SendsV2Component } from "./send-V2.component";
+
+describe("SendsV2Component", () => {
```
Hinton:
```suggestion
describe("SendV2Component", () => {
```

apps/desktop/src/app/tools/send-v2/send-v2.component.ts
```diff
@@ -0,0 +1,9 @@
...
+  imports: [],
+  template: "<p>Sends V2 Component</p>",
+  changeDetection: ChangeDetectionStrategy.OnPush,
+})
+export class SendsV2Component {}
```
Hinton:
```suggestion
export class SendV2Component {}
```

apps/desktop/src/app/layout/desktop-layout.component.ts
willmartian: I personally like the more specific selectors.

Hinton:
Linting and tests fails.

Hinton:
This will be tested on main.

PR #17673, opened 2025-11-26
[PM-28813] Implement encryption diagnostics & recovery tool

quexten:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-28813
https://github.com/bitwarden/server/pull/6659

## 📔 Objective

To get diagnostics for certain types of encryption related issues, and to fix users vaults, this PR adds a tool that:
- Verifies integrity of encryption for private keys, folders, vault items
- Offers the option to "repair" the vault

Repairing the vault here means replacing the private keys, should they be corrupt, and removing vault items / folders, should they be corrupt.

This will help CS track down issues and resolve issues with customers faster.

## 📸 Screenshots

Note: In the video the private key stays corrupt. This is because the code used when recording the video marks all private keys as corrupt, just for testing / demo purposes.

https://github.com/user-attachments/assets/fc1db8ea-9716-4184-8dae-9c73bdb6427a

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/apps/web/src/app/key-management/data-recovery/data-recovery.component.html
+++ b/apps/web/src/app/key-management/data-recovery/data-recovery.component.html
@@ -66,7 +66,7 @@
           {{ "repairIssues" | i18n }}
         </button>
       }
-      <button type="button" bitButton buttonType="secondary" (click)="saveDiagnosticLogs()">
+      <button type="button" bitButton buttonType="secondary" [bitAction]="saveDiagnosticLogs">
         <i class="bwi bwi-download" aria-hidden="true"></i>
         {{ "saveDiagnosticLogs" | i18n }}
       </button>
--- a/apps/web/src/app/key-management/data-recovery/data-recovery.component.ts
+++ b/apps/web/src/app/key-management/data-recovery/data-recovery.component.ts
@@ -7,7 +7,6 @@ import { AccountService } from "@bitwarden/common/auth/abstractions/account.serv
 import { CryptoFunctionService } from "@bitwarden/common/key-management/crypto/abstractions/crypto-function.service";
 import { FileDownloadService } from "@bitwarden/common/platform/abstractions/file-download/file-download.service";
 import { I18nService } from "@bitwarden/common/platform/abstractions/i18n.service";
-import { SdkLoadService } from "@bitwarden/common/platform/abstractions/sdk/sdk-load.service";
 import { CipherEncryptionService } from "@bitwarden/common/vault/abstractions/cipher-encryption.service";
 import { FolderApiServiceAbstraction } from "@bitwarden/common/vault/abstractions/folder/folder-api.service.abstraction";
 import { ButtonModule, DialogService } from "@bitwarden/components";
@@ -93,9 +92,6 @@ export class DataRecoveryComponent {
     if (this.status() === StepStatus.InProgress) {
       return;
     }
-
-    // Ensure SDK is loaded
-    await SdkLoadService.Ready;
 
     this.hasStarted.set(true);
     this.status.set(StepStatus.InProgress);
@@ -195,7 +191,7 @@ export class DataRecoveryComponent {
     }
   };
 
-  saveDiagnosticLogs() {
+  saveDiagnosticLogs = () => {
     const logs = this.logger.getLogs();
     const timestamp = new Date().toISOString().replace(/[:.]/g, "-");
     const filename = `data-recovery-logs-${timestamp}.txt`;
@@ -208,5 +204,5 @@ export class DataRecoveryComponent {
     });
 
     this.logger.record("Diagnostic logs saved");
-  }
+  };
 }
--- a/apps/web/src/app/key-management/data-recovery/steps/folder-step.ts
+++ b/apps/web/src/app/key-management/data-recovery/steps/folder-step.ts
@@ -1,3 +1,4 @@
+import { SdkLoadService } from "@bitwarden/common/platform/abstractions/sdk/sdk-load.service";
 import { FolderApiServiceAbstraction } from "@bitwarden/common/vault/abstractions/folder/folder-api.service.abstraction";
 import { DialogService } from "@bitwarden/components";
 import { PureCrypto } from "@bitwarden/sdk-internal";
@@ -30,6 +31,7 @@ export class FolderStep implements RecoveryStep {
         continue;
       }
       try {
+        await SdkLoadService.Ready;
         PureCrypto.symmetric_decrypt_string(
           folder.name.encryptedString,
           workingData.userKey.toEncoded(),
--- a/apps/web/src/app/key-management/data-recovery/steps/private-key-step.ts
+++ b/apps/web/src/app/key-management/data-recovery/steps/private-key-step.ts
@@ -1,4 +1,5 @@
 import { CryptoFunctionService } from "@bitwarden/common/key-management/crypto/abstractions/crypto-function.service";
+import { SdkLoadService } from "@bitwarden/common/platform/abstractions/sdk/sdk-load.service";
 import { EncryptionType } from "@bitwarden/common/platform/enums";
 import { DialogService } from "@bitwarden/components";
 import { UserAsymmetricKeysRegenerationService } from "@bitwarden/key-management";
@@ -32,6 +33,7 @@ export class PrivateKeyStep implements RecoveryStep {
     logger.record("Private key length: " + encryptedPrivateKey.length);
     let privateKey: Uint8Array;
     try {
+      await SdkLoadService.Ready;
       privateKey = PureCrypto.unwrap_decapsulation_key(
         encryptedPrivateKey,
         workingData.userKey.toEncoded(),


apps/web/src/app/key-management/data-recovery/steps/folder-step.ts
```diff
@@ -0,0 +1,99 @@
...
+        logger.record(`Folder ID ${folder.id} has no name`);
+        this.undecryptableFolderIds.push(folder.id);
+        continue;
+      }
+      try {
+        PureCrypto.symmetric_decrypt_string(
+          folder.name.encryptedString,
+          workingData.userKey.toEncoded(),
+        );
+      } catch {
```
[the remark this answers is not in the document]
quexten: No, we don't want to log error contents here for now.

apps/web/src/app/key-management/data-recovery/steps/cipher-step.ts
```diff
@@ -0,0 +1,83 @@
...
+    if (!workingData.userId) {
+      logger.record("Missing user ID");
+      return false;
+    }
+
+    this.undecryptableCipherIds = [];
+    for (const cipher of workingData.ciphers) {
+      try {
+        await this.cipherService.decrypt(cipher, workingData.userId);
+      } catch {
```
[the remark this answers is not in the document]
quexten: No we don't want to log error details here now.

apps/web/src/app/key-management/data-recovery/data-recovery.component.ts
```diff
@@ -0,0 +1,223 @@
...
+  FolderStep,
+  CipherStep,
+} from "./steps";
+
+interface StepState {
+  title: string;
+  completed: boolean;
+  inProgress: boolean;
+  failed: boolean;
+  message?: string;
```
Thomas-Avery: These look to be booleans representing mutually exclusive states a step can be in. What about an enum-like property to represent the current status instead? 

Would give us better type safety and clearer intent.

```
export const StepStatus = Object.freeze({
  InProgress: 1,
  Completed: 2,
  Failed: 3,
} as const);
export type StepStatus = (typeof StepStatus)[keyof typeof StepStatus];
```

apps/web/src/app/key-management/data-recovery/data-recovery.component.html
```diff
@@ -0,0 +1,84 @@
...
+    {{ "dataRecoveryDescription" | i18n }}
+  </p>
+
+  @if (!diagnosticsCompleted() && !recoveryCompleted()) {
+    <button
+      type="button"
+      bitButton
+      buttonType="primary"
+      [disabled]="isRunning()"
+      (click)="runDiagnostics()"
```
Thomas-Avery: Should we be using async actions for the buttons in this component? 

https://components.bitwarden.com/?path=/docs/component-library-async-actions-standalone--docs

apps/web/src/app/key-management/data-recovery/data-recovery.component.ts
```diff
@@ -0,0 +1,223 @@
...
+    const url = window.URL.createObjectURL(blob);
+
+    const link = document.createElement("a");
+    link.href = url;
+    link.download = filename;
+    link.click();
+
+    window.URL.revokeObjectURL(url);
+    this.logger.record("Diagnostic logs saved");
+  }
```
Thomas-Avery: Can we use the `FileDownloadService` for this? Seems like we should have some generic utilities for these kinds of things.

apps/web/src/app/key-management/data-recovery/data-recovery.component.spec.ts
```diff
@@ -0,0 +1,356 @@
...
+    mockAccountService = mockAccountServiceWith(mockUserId);
+    mockKeyService = mock<KeyService>();
+    mockFolderApiService = mock<FolderApiServiceAbstraction>();
+    mockCipherEncryptService = mock<CipherEncryptionService>();
+    mockDialogService = mock<DialogService>();
+    mockPrivateKeyRegenerationService = mock<UserAsymmetricKeysRegenerationService>();
+    mockLogService = mock<LogService>();
+    mockCryptoFunctionService = mock<CryptoFunctionService>();
+
+    mockI18nService.t.mockImplementation((key: string) => key);
```
Thomas-Avery:
```suggestion
    mockI18nService.t.mockImplementation((key) => `${key}-used-i18n`);
```

⛏️  non-blocking  This is more general feedback.

Checking that the key was passed doesn't tell us if the i18nService was actually used or not. In these tests we aren't currently checking for any i18 service calls so it isn't applicable.

apps/web/src/app/key-management/data-recovery/steps/sync-step.ts
```diff
@@ -0,0 +1,45 @@
...
+import { CipherData } from "@bitwarden/common/vault/models/data/cipher.data";
+import { FolderData } from "@bitwarden/common/vault/models/data/folder.data";
+import { Cipher } from "@bitwarden/common/vault/models/domain/cipher";
+import { Folder } from "@bitwarden/common/vault/models/domain/folder";
+
+import { LogRecorder } from "../log-recorder";
+
+import { RecoveryStep, RecoveryWorkingData } from "./recovery-step";
+
+export class SyncStep extends RecoveryStep {
```
Thomas-Avery:
```suggestion
export class SyncStep implements RecoveryStep {
```

Doesn't look like there is any inheritance here?

apps/web/src/app/key-management/data-recovery/steps/user-info-step.ts
```diff
@@ -0,0 +1,51 @@
...
+
+import { AccountService } from "@bitwarden/common/auth/abstractions/account.service";
+import { EncryptionType } from "@bitwarden/common/platform/enums";
+import { KeyService } from "@bitwarden/key-management";
+
+import { LogRecorder } from "../log-recorder";
+
+import { RecoveryStep, RecoveryWorkingData } from "./recovery-step";
+
+export class UserInfoStep extends RecoveryStep {
```
Thomas-Avery:
```suggestion
export class UserInfoStep implements RecoveryStep {
```

apps/web/src/app/key-management/data-recovery/data-recovery.component.ts
```diff
@@ -0,0 +1,223 @@
...
+
+    this.logger.record("Starting diagnostics...");
+    this.workingData = {
+      userId: null,
+      userKey: null,
+      isPrivateKeyCorrupt: false,
+      encryptedPrivateKey: null,
+      ciphers: [],
+      folders: [],
+    };
```
Thomas-Avery, line 108: 💭  

It was hard for me to follow this being passed around to the steps and being enriched/mutated. After a couple of re-reads I understand it better.

This feels like it isn't a presentational concern and might be better extracted into a component service? 

That might be tricky with how we have things dynamically updating after each step atm.

apps/web/src/app/key-management/data-recovery/steps/private-key-step.ts
```diff
@@ -0,0 +1,97 @@
...
+import { EncryptionType } from "@bitwarden/common/platform/enums";
+import { DialogService } from "@bitwarden/components";
+import { KeyService, UserAsymmetricKeysRegenerationService } from "@bitwarden/key-management";
+import { PureCrypto } from "@bitwarden/sdk-internal";
+
+import { LogRecorder } from "../log-recorder";
+
+import { RecoveryStep, RecoveryWorkingData } from "./recovery-step";
+
+export class PrivateKeyStep extends RecoveryStep {
```
Thomas-Avery:
```suggestion
export class PrivateKeyStep implements RecoveryStep {
```

apps/web/src/app/key-management/data-recovery/steps/private-key-step.ts
```diff
@@ -0,0 +1,97 @@
...
+  async runDiagnostics(workingData: RecoveryWorkingData, logger: LogRecorder): Promise<boolean> {
+    if (!workingData.userId || !workingData.userKey) {
+      logger.record("Missing user ID or user key");
+      return false;
+    }
+
+    // Make sure the private key decrypts properly and is not somehow encrypted by a different user key / broken during key rotation.
+    const encryptedPrivateKey = await firstValueFrom(
+      this.keyService.userEncryptedPrivateKey$(workingData.userId),
+    );
```
Thomas-Avery: Should we be using the fresh data from the sync `workingData.encryptedPrivateKey` instead?

apps/web/src/app/key-management/data-recovery/steps/private-key-step.ts
```diff
@@ -0,0 +1,97 @@
...
+    // Only support recovery on V1 users.
+    return (
+      workingData.isPrivateKeyCorrupt &&
+      workingData.userKey !== null &&
+      workingData.userKey.inner().type === EncryptionType.AesCbc256_HmacSha256_B64
+    );
+  }
+
+  async runRecovery(workingData: RecoveryWorkingData, logger: LogRecorder): Promise<void> {
+    // The recovery step is to replace the key pair. This will break emergency access enrollments / organization memberships / provider memberships.
```
Thomas-Avery: Unless there is some server work I'm missing this isn't how it works now. 

If the user has designated emergency access or is in an organization the server will response back.
```
 BadRequestException("Key regeneration not supported for this user.");
```

apps/web/src/app/key-management/data-recovery/steps/private-key-step.ts
```diff
@@ -0,0 +1,97 @@
...
+      this.keyService.userEncryptedPrivateKey$(workingData.userId),
+    );
+    if (!encryptedPrivateKey) {
+      logger.record("No encrypted private key found");
+      return false;
+    }
+    logger.record("Private key length: " + encryptedPrivateKey.length);
+    let privateKey: Uint8Array;
+    try {
+      privateKey = PureCrypto.unwrap_decapsulation_key(
```
Thomas-Avery, line 37:
```suggestion
      await SdkLoadService.Ready;
      privateKey = PureCrypto.unwrap_decapsulation_key(
```

apps/web/src/app/key-management/data-recovery/steps/folder-step.ts
```diff
@@ -0,0 +1,99 @@
+import { FolderApiServiceAbstraction } from "@bitwarden/common/vault/abstractions/folder/folder-api.service.abstraction";
+import { DialogService } from "@bitwarden/components";
+import { PureCrypto } from "@bitwarden/sdk-internal";
+
+import { LogRecorder } from "../log-recorder";
+
+import { RecoveryStep, RecoveryWorkingData } from "./recovery-step";
+
+export class FolderStep extends RecoveryStep {
```
Thomas-Avery:
```suggestion
export class FolderStep implements RecoveryStep {
```

apps/web/src/app/key-management/data-recovery/steps/folder-step.ts
```diff
@@ -0,0 +1,99 @@
...
+
+    this.undecryptableFolderIds = [];
+    for (const folder of workingData.folders) {
+      if (!folder.name?.encryptedString) {
+        logger.record(`Folder ID ${folder.id} has no name`);
+        this.undecryptableFolderIds.push(folder.id);
+        continue;
+      }
+      try {
+        PureCrypto.symmetric_decrypt_string(
```
Thomas-Avery, line 35:
```suggestion
        await SdkLoadService.Ready;
        PureCrypto.symmetric_decrypt_string(
```

apps/web/src/app/key-management/data-recovery/steps/folder-step.ts
```diff
@@ -0,0 +1,99 @@
...
+    if (!confirmed) {
+      logger.record("User cancelled folder deletion");
+      throw new Error("Folder recovery cancelled by user");
+    }
+
+    logger.record(`Deleting ${this.undecryptableFolderIds.length} folders`);
+
+    for (const folderId of this.undecryptableFolderIds) {
+      try {
+        await this.folderService.delete(folderId, workingData.userId);
```
Thomas-Avery, line 84: 💭 

We are using the folders from the sync-step. There is a possibility that what is in state doesn't match what we pulled down. Looking through the folderService code it looks like that will be okay but this could be brittle to downstream changes. For example the folder service changes to error if the folder isn't in state.

apps/web/src/app/key-management/data-recovery/steps/cipher-step.ts
```diff
@@ -0,0 +1,83 @@
+import { ApiService } from "@bitwarden/common/abstractions/api.service";
+import { CipherEncryptionService } from "@bitwarden/common/vault/abstractions/cipher-encryption.service";
+import { DialogService } from "@bitwarden/components";
+
+import { LogRecorder } from "../log-recorder";
+
+import { RecoveryStep, RecoveryWorkingData } from "./recovery-step";
+
+export class CipherStep extends RecoveryStep {
```
Thomas-Avery:
```suggestion
export class CipherStep implements RecoveryStep {
```

libs/key-management/src/user-asymmetric-key-regeneration/abstractions/user-asymmetric-key-regeneration.service.ts
```diff
@@ -7,4 +7,6 @@ export abstract class UserAsymmetricKeysRegenerationService {
    * @param userId The user id.
    */
   abstract regenerateIfNeeded(userId: UserId): Promise<void>;
+
```
Thomas-Avery, line 10: Can we document this interface?

Thomas-Avery (changes requested):
Sweet tool and great work. Would like to see more test coverage for example on the steps. A few things to look at from my side.

libs/key-management/src/user-asymmetric-key-regeneration/abstractions/user-asymmetric-key-regeneration.service.ts
quexten: Documented and added a guard for V1 users, as this should not be used with V2 users.

apps/web/src/app/key-management/data-recovery/steps/folder-step.ts
quexten: @Thomas-Avery We already ensure that the SDK is loaded before running an step in the data recovery component. The intent was to not having to await this every time.

I'm ok making this change, but could you explain the rationale behind this? I'm trying to understand why we require to push it into the steps instead of having it outside, especially given that these steps are not meant to be public interfaces.

apps/web/src/app/key-management/data-recovery/steps/private-key-step.ts
quexten: TBH, I did not check what the key regen service currently actually does, I just re-used the existing service. I'll update the docs, but if we use it for this use-case we should probably consider updating the server-side work to just remove the user from EA/Orgs, *or* at least move them back to the "Accepted" status.

apps/web/src/app/key-management/data-recovery/steps/folder-step.ts
quexten: I've changed the try catch to not re-throw the error. So any error here should be caught and logged but ignored. If it actually does turn out to break, we can still consider somehow circumventing it / using API calls directly.

apps/web/src/app/key-management/data-recovery/steps/private-key-step.ts
quexten: Changed!

apps/web/src/app/key-management/data-recovery/data-recovery.component.spec.ts
quexten: Applied the suggestion but also did not add any further tests for the i18n service calls.

apps/web/src/app/key-management/data-recovery/data-recovery.component.html
quexten: Applied

apps/web/src/app/key-management/data-recovery/data-recovery.component.ts
quexten: Applied

apps/web/src/app/key-management/data-recovery/data-recovery.component.ts
quexten: I initially wanted to move this into a service, but doing so would have made this a fair bit more complex, because exactly as you mentioned the steps are updating each time. So we'd have to do something like return an observable from the service that returns the step statuses.

At any rate, I did not think this complexity is worth it for the first version of this tool.

quexten:
@Thomas-Avery apologies for the force push, I had a wrong `main` revision which pulled in some incorrect / unrelated commits here.

apps/web/src/app/key-management/data-recovery/steps/private-key-step.ts
Thomas-Avery: The original plan was to expand the server side to move them back to `Accepted` status as you say. We were wanting to investigate how to make the UX better for both the current user and the admin that would have to reconfirm things.

apps/web/src/app/key-management/data-recovery/steps/folder-step.ts
Thomas-Avery: @quexten Having the high level component need to know about some low level implementation detail is generally bad practice. 

For this specifically I'm concern about brittleness and long term maintenance. In the future if the component is refactored and we miss `await SdkLoadService.Ready;` we don't receive any compilation error, and at best a non deterministic run time error to alert us this step was missed in the high level.

Coupling this to the low level call isn't, from a performance perspective ideal, but I believe is more maintainable. At the time we planned to rarely use PureCrypto directly so we skipped on a utility. Maybe it is a good time to now add a wrapper/proxy service to always call `await SdkLoadService.Ready;` when calling PureCrypto?

quexten: Let's move that to a separate PR but I agree this is not reasonable as it is currently. I can make a ticket here. What do you think about just simply adding the SdkLoadService call to each pure crypto function's implementation?

On the performance note, this would be an issue exactly for the case of many tiny calls, such as for cipher descryption. However, vault already moved off of purecrypto for it so I don't think it's an issue unless we prove otherwise by measurements.

Thomas-Avery: I'm good just having it inline for this PR and making a seperate ticket for a generic setup. 

> What do you think about just simply adding the SdkLoadService call to each pure crypto function's implementation?

Is that possible with how we are generating PureCrypto via the SDK? 

> On the performance note, this would be an issue exactly for the case of many tiny calls, such as for cipher descryption. However, vault already moved off of purecrypto for it so I don't think it's an issue unless we prove otherwise by measurements.

I would expect any real performance concerns would have been snuffed out by now. We await it on each encryptService call and it's been like that for awhile.

apps/web/src/app/key-management/data-recovery/data-recovery.component.html
```diff
@@ -0,0 +1,75 @@
...
+          type="button"
+          bitButton
+          buttonType="primary"
+          [disabled]="status() === StepStatus.InProgress"
+          [bitAction]="runRecovery"
+        >
+          {{ "repairIssues" | i18n }}
+        </button>
+      }
+      <button type="button" bitButton buttonType="secondary" (click)="saveDiagnosticLogs()">
```
Thomas-Avery:
```suggestion
      <button type="button" bitButton buttonType="secondary" [bitAction]="saveDiagnosticLogs">
```

apps/web/src/app/key-management/data-recovery/steps/folder-step.ts
quexten: Applied.

> Is that possible with how we are generating PureCrypto via the SDK?

sorry, got confused between encrypt service and purecrypto when writing that comment.

PR #17707, opened 2025-11-27
[PM-26049] Auto key not stored due to vault timeout write vs read race condition for cli

mzieniukbw:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-26049

## 📔 Objective

Auto (user) key is (sometimes) not stored due to vault timeout being null.
This happens due to race condition on the storage write vs read. The read is hot observable, so getting value via `firstValueFrom` might return null.
Since vault timeout is not supposed to be null (since we always have a non-null default for all clients), adding a filter for non-nullability is enough to make sure the actual value is returned, before deciding whether to store the auto key.

Writing the vault timeout (default when not set) happens as a "side effect" of `getVaultTimeoutByUserId$`, which is not great, but since we require it (`firstValueFrom`) just before identity token request (before parsing it and setting user keys to state), the vault timeout is reliably stored in state for the first time when not set, during `bw login` and always before the auto key storage.

Tested login, unlock, sync, changing vault timeout in settings from and to Never on all clients, restarting app to see if it starts unlocked.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/apps/browser/src/auth/services/extension-device-management-component.service.ts
+++ b/apps/browser/src/auth/services/extension-device-management-component.service.ts
@@ -3,9 +3,7 @@ import { DeviceManagementComponentServiceAbstraction } from "@bitwarden/angular/
 /**
  * Browser extension implementation of the device management component service
  */
-export class ExtensionDeviceManagementComponentService
-  implements DeviceManagementComponentServiceAbstraction
-{
+export class ExtensionDeviceManagementComponentService implements DeviceManagementComponentServiceAbstraction {
   /**
    * Don't show header information in browser extension client
    */
--- a/apps/browser/src/autofill/fido2/services/browser-fido2-user-interface.service.ts
+++ b/apps/browser/src/autofill/fido2/services/browser-fido2-user-interface.service.ts
@@ -120,9 +120,7 @@ export type BrowserFido2ParentWindowReference = chrome.tabs.Tab;
  * Browser implementation of the {@link Fido2UserInterfaceService}.
  * The user interface is implemented as a popout and the service uses the browser's messaging API to communicate with it.
  */
-export class BrowserFido2UserInterfaceService
-  implements Fido2UserInterfaceServiceAbstraction<BrowserFido2ParentWindowReference>
-{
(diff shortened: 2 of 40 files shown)


libs/common/src/key-management/vault-timeout/services/vault-timeout-settings.service.ts
```diff
@@ -135,11 +136,12 @@ export class VaultTimeoutSettingsService implements VaultTimeoutSettingsServiceA
     ]).pipe(
       switchMap(([currentVaultTimeout, maxVaultTimeoutPolicy]) => {
         return from(this.determineVaultTimeout(currentVaultTimeout, maxVaultTimeoutPolicy)).pipe(
-          tap((vaultTimeout: VaultTimeout) => {
+          concatMap(async (vaultTimeout: VaultTimeout) => {
```
Thomas-Avery, line 154: ❓  Is switching to the concatMap here to make sure we're saving state in the correct order because of it `concatMap`'s serialized order of execution?

Thomas-Avery (approved):
Looks good. Do have one non blocking question for my understanding.

libs/common/src/key-management/vault-timeout/services/vault-timeout-settings.service.ts
mzieniukbw: It's just for the serialized order of execution, which is more correct than `tap`.
But it technically have no effect on the race condition, since all state reads are hot observables, so there is no guarantee you are going to get the just set value immediately via `firstValueFrom`, after the `concatMap` execution.

mzieniukbw: Plus `tap` usage is incorrect, since the returned state promise is not guaranteed to be awaited (quoting `tap` docs). So in practise the set to state operation was a dangling promise.

Thomas-Avery: Thank you for the extra context. That all sounds good to me.

PR #17751, opened 2025-12-01
Playwright PoC

Hinton:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-29853

## 📔 Objective

bitwarden/server#6424 adds an API to allow for seeding and tracking of test data. This PR is to establish a framework for using that API in the clients repository to end-to-end test user flows.

Adopt Playwright to execute tests and browser automation and create test fixtures that provide useful shortcuts for frequently traveled browser automation paths (e.g. authenticated user). This framework should also set up `x-play-id` tracking and test data teardown logic.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/libs/playwright-helpers/src/example.play.spec.ts
+++ b/libs/playwright-helpers/src/example.play.spec.ts
@@ -11,10 +11,10 @@ test.skip("login and save session", async ({ auth }) => {
   await expectUnlockedAs(scene.mangle("test@example.com"), page);
 });
 
-test.skip("As long as the previous test ran in this worker, this time it will reuse authentication", async ({
+test.only("As long as the previous test ran in this worker, this time it will reuse authentication", async ({
   auth,
 }) => {
-  const { page, scene } = await auth.authenticate("test@example.com", "asdfasdfasdf");
+  const { page, scene } = await auth.authenticate("test@example.com", "passwordpassword");
 
   await expectUnlockedAs(scene.mangle("test@example.com"), page);
 });
--- a/libs/playwright-helpers/src/fixtures/auth.fixture.ts
+++ b/libs/playwright-helpers/src/fixtures/auth.fixture.ts
@@ -46,7 +46,9 @@ type AuthenticatedContext = {
   scene: SingleUserScene;
 };
 
-type SessionOptions = Simplify<Except<extractTUpType<SingleUserSceneTemplate>, "email">> & {
+type SessionOptions = Simplify<
+  Except<extractTUpType<SingleUserSceneTemplate>, "email" | "password">
+> & {
   /** The page to use for authenticating */
   page?: Page;
 };
@@ -158,7 +160,7 @@ export class AuthFixture {
     password: string,
     options: SessionOptions = {},
   ): Promise<AuthenticatedContext> {
-    const scene = await Play.scene(new SingleUserSceneTemplate({ ...options, email }));
+    const scene = await Play.scene(new SingleUserSceneTemplate({ ...options, email, password }));
 
     return await this.authenticateForScene(scene, password);
   }
--- a/libs/playwright-helpers/src/scene-templates/single-user.scene.ts
+++ b/libs/playwright-helpers/src/scene-templates/single-user.scene.ts
@@ -16,6 +16,7 @@ type SceneResult = {
 };
 type UpParams = {
   email: string;
+  password: string;
   emailVerified?: boolean;
   premium?: boolean;
 };
--- a/libs/playwright-helpers/src/test.ts
+++ b/libs/playwright-helpers/src/test.ts
@@ -67,7 +67,7 @@ export async function cleanStage(): Promise<void> {
     // eslint-disable-next-line no-console
     console.warn(
       "PLAYWRIGHT_SKIP_CLEAN_STAGE is set, run\n",
-      `curl -X DELETE ${new URL(playId, webServerBaseUrl).toString()}\n`,
+      `curl -X DELETE ${new URL(`/seed/${playId}`, webServerBaseUrl).toString()}\n`,
     );
     return;
   }


libs/shared/es2020-transformer.ts
```diff
@@ -1,36 +0,0 @@
-import * as ts from "typescript";
```
MGibson1, line 1: we no longer need the ast transformer since we don't use import.meta statements anymore. (this was also two years ago, it's possible this just works now).

MGibson1:
Note on previous blockers
- tsconfig changes: These are a result of updates to jsconfig, which is a meaningful part of this PR.
- playwright CI: This is being put on hold since we can include local tests for dev purposes now without loss of benefit down the line once we have the infrastructure to test in CI.

.github/CODEOWNERS
```diff
@@ -193,6 +192,9 @@ apps/desktop/src/services/native-messaging.service.ts @bitwarden/team-key-manage
 apps/browser/src/background/nativeMessaging.background.ts @bitwarden/team-key-management-dev
 apps/desktop/src/services/biometric-message-handler.service.ts @bitwarden/team-key-management-dev
 
+## Architecture
+libs/playwright-helpers @bitwarden/team-architecture
```
withinfocus, line 197: ⚠️ Temporary?

libs/playwright-helpers/README.md
```diff
@@ -0,0 +1,83 @@
+# playwright-helpers
+
+Owned by: architecture
```
theMickster, line 3: ⚠️ Temporary, correct? Do we have a subject matter expert group planned to own these?

libs/playwright-helpers/src/example.play.spec.ts
theMickster, line 1: ⛏️ I think this is better put into an `Example.md` that evolves along with the README.md and Claude.md files. I think we can write some interesting SKILL.md to keep all 3 up-to-date.

libs/playwright-helpers/src/fixtures/auth.fixture.ts
```diff
@@ -0,0 +1,267 @@
...
+    if (previousAuth != null && previousAuth.password !== password) {
+      throw new Error(
+        `Email ${email} is already authenticated with a different password (${
+          AuthenticatedEmails.get(email)!.password
+        })`,
+      );
+    }
+    const scene = AuthenticatedEmails.get(email)!.scene;
+    const mangledEmail = scene.mangle(email);
+    await page.context().storageState({ path: dataFilePath(mangledEmail) });
```
theMickster, line 139: ❓ I ran the Anthropic `/code-review:code-review` command and it came back with the following. Thoughts?

**Severity**  Bug - `storageState()` saves instead of loads in `resumeSession`

**Assessment**
The `storageState` bug in `resumeSession` should be addressed before merging to ensure session resume functionality works correctly.

<details>
<summary>Claude Code Review Suggestion</summary>
**Code:**
```typescript
async resumeSession(
  email: string,
  password: string,
  options: SessionOptions = {},
): Promise<AuthenticatedContext> {
  const page = options?.page || (await this.newPage());
  // ...
  const mangledEmail = scene.mangle(email);
  await page.context().storageState({ path: dataFilePath(mangledEmail) });  // <-- BUG

  // Load stored state and session into a new page
  await loadLocal(page, mangledEmail);
  await loadSession(page, mangledEmail);
  // ...
}
```

**Problem:** Playwright's `storageState({ path: '...' })` method **saves** the current context's storage state to a file - it does not load from it. This call saves the empty/new page's state, potentially overwriting any previously saved session data in `auth-{email}.json`.

**Expected behavior:** To load stored state, you should pass `storageState` as an option when creating the browser context:
```typescript
const context = await this._browser.newContext({ storageState: dataFilePath(mangledEmail) });
```

**Impact:** Session resume functionality may not work correctly. The call either:
1. Overwrites saved data with empty state (if the file exists)
2. Creates an empty file (if it doesn't exist)
3. Has no effect since `authenticateForScene` never saves to this file path anyway

**Suggested fix:** Either:
- **Option A:** Remove this line entirely since `loadLocal` and `loadSession` handle localStorage/sessionStorage, and cookies aren't being saved during initial authentication
- **Option B:** If cookie persistence is needed, save cookies in `authenticateForScene` and load them by passing `storageState` to `newContext()`
</details>

libs/playwright-helpers/src/example.play.spec.ts
MGibson1: I had considered doing that, but I like the ability to easily run these tests to get a handle on how things like the auth framework work and playwright can be worked with.

I could be convinced that's doable from some other, real test, but that was my intent

libs/playwright-helpers/src/fixtures/auth.fixture.ts
MGibson1: That appears to be a hallucinated method and usage restriction on `storageState()`. conceptually, what that method is doing is telling the page context to use the existing `dataFilePath` as the backing store for the page. It will read and write data to that file in normal browser operation.

This pattern is described in [Playwright docs here](https://playwright.dev/docs/auth#basic-shared-account-in-all-tests)

libs/playwright-helpers/README.md
MGibson1: Yeah, I want to own things for a while longer to help fascilitate adoption and tooling creation, but we'll be looking for a team to take on the burden long-term.

.github/CODEOWNERS
MGibson1: Yeah, I want to own things for a while longer to help fascilitate adoption and tooling creation, but we'll be looking for a team to take on the burden long-term.

libs/playwright-helpers/src/example.play.spec.ts
theMickster: Okay, that works for me too.

theMickster (dismissed):

dani-garcia (dismissed):

rr-bw (approved):

amorask-bitwarden (dismissed):

MGibson1:
Converting to draft while we explore other automators

PR #17790, opened 2025-12-02
[PM-24978] Corrupt Attachment Keys

nick-livefront:
## 🎟️ Tracking

[PM-24978](https://bitwarden.atlassian.net/browse/PM-24978)

## 📔 Objective

When a attachment key is invalid (for any reason) it would result in the attachment looking normal but would fail to download. Without a valid key we are unable to download an attachment.

- When the attachment key is invalid, show `Error: Cannot decrypt` message and hide the download button
- After discussing with Renee - updated the case for an invalid filename with the new `Error: Cannot decrypt` to match

## 📸 Screenshots

|Web|Browser|Desktop|
|-|-|-|
|Screenshot 2025-12-02 at 2 36 42 PM|Screenshot 2025-12-02 at 2 39 37 PM|Screenshot 2025-12-02 at 2 49 39 PM|

## 🦮 Reviewer guidelines

[PM-24978]: https://bitwarden.atlassian.net/browse/PM-24978?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


shane-melton (approved):
Looks good!

nick-livefront:
@shane-melton I had conflicts to fix! Re-requesting review

shane-melton (dismissed):

nick-livefront:
@shane-melton Conflict fix in the `messages.json` file

shane-melton (approved):

nick-livefront:
Converting to draft until SDK changes are approved

libs/vault/src/cipher-form/components/attachments/cipher-attachments.component.html
```diff
@@ -26,7 +29,7 @@ <h2 class="tw-sr-only" id="attachments">{{ "attachments" | i18n }}</h2>
                   [cipher]="cipher()"
                   [attachment]="attachment"
                 ></app-download-attachment>
-              } @else {
+              } @else if (!attachment.hasDecryptionError) {
```
nick-livefront: @shane-melton This and the above changes I added after your fix attachment implementation. I believe this is still the right path but wanted to call it out specifically to make sure I wasn't creating a regression. 

For reference, the `key` will be `None` from the SDK when it fails to decrypt. Here is my associated SDK change: https://github.com/bitwarden/sdk-internal/pull/644

shane-melton: I don't see any regressions with this approach. However, I might suggest we have an entirely separate, top-level `<bit-item>` templates for normal attachments and those with decryption errors. That would let us avoid having so many nested `!attachment.hasDecryptionError` conditionals.

We'd have to repeat the delete item action for both, but I find that personally preferable over these conditionals being sprinkled throughout. What do you think?

nick-livefront: I love the idea: [fe0472d](https://github.com/bitwarden/clients/pull/17790/commits/fe0472da5cc6e53ffba509c08a77e274f9454451)

shane-melton (approved):

nick-livefront:
@shane-melton I had a conflict to fix and this has passed QA. Could you give a review when you have time?

PR #17798, opened 2025-12-03
[PM-29122] Fix debug build causing slow unlock

quexten:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-29122

## 📔 Objective

The secret store implementation on Linux - oo7 - has multiple backends, DBUS and Portal+File. For the file backend, PBKDF2 is used each time a key is accessed from the key store.

https://github.com/bitwarden/clients/pull/17357 changed the CI to use debug builds of desktop-native. This causes PBKDF2 to run extremely slowly, causing 1 second access times per key store lookup, causing unlock / lock times of tens of seconds to minutes.

This PR reverts the change.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


.github/workflows/build-desktop.yml
```diff
@@ -249,9 +249,11 @@ jobs:
           PKG_CONFIG_ALLOW_CROSS: true
           PKG_CONFIG_ALL_STATIC: true
           TARGET: musl
+        # Note: It is important that we use the release build because some compute heavy
+        # operations such as key derivation for oo7 on linux are too slow in debug mode
```
neuronull, line 253: Interesting ... I could envision a need to have a debug build to on linux in the future.

Are these issues something we could potentially work around? Wondering if it makes sense to open a ticket to track this.

quexten: If we want a debug build, that does not have very bad performance, we need to re-engineer how secure storage works on linux completely. The heavy cryptographic operation must be done once, and once only, after which the encryption key is held in memory. So we would no longer store access tokens in secure storage but a single key, that is then used to unlock the protected access tokens.

Please note, this only applies on snap / flatpak, because only they use the portal + file backend which uses the KDF here. Other than that we don't currently have compute heavy operations in the SDK. Maybe RSA signing?

This suggested refactor/rearchitecture is probably a good idea anyways, because windows has size limitations in secure storage. So we just have one security function "give me a single key" to worry about.

One other alternative is adding a read-cache, which would significantly lessen the impact.

dani-garcia: we may be able to do like in the SDK, where our dependencies and crypto code are optimized for release, but the other crates are not: https://github.com/bitwarden/sdk-internal/blob/main/Cargo.toml#L128-L146

quexten:
Looks like there was another error which lead to the napi build still being a debug build.

dani-garcia (approved):

.github/workflows/build-desktop.yml
Hinton: We should make certain debug builds never reach end consumers.

PR #17805, opened 2025-12-03
[PM-28422] Client batching for member actions

BTreston:
## 🎟️ Tracking
https://bitwarden.atlassian.net/browse/PM-28422

## 📔 Objective
Implements member action batching for the bulk reinvite functionality in the organization members management system. The changes enable processing large numbers of users efficiently by breaking them into sequential batches.
  - Introduced REQUESTS_PER_BATCH constant set to 500 users per batch
  - Created generic processBatchedOperation() helper that:
    - Processes user IDs in sequential batches
    - Aggregates successful and failed results across all batches
    - Continues processing remaining batches even if one fails
  - Updated bulkReinvite() to use the batching helper instead of processing all users at once when the feature flag is enabled

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/apps/web/src/app/admin-console/common/people-table-data-source.ts
+++ b/apps/web/src/app/admin-console/common/people-table-data-source.ts
@@ -24,7 +24,7 @@ export const MaxCheckedCount = 500;
  * Maximum for bulk reinvite operations when the IncreaseBulkReinviteLimitForCloud
  * feature flag is enabled on cloud environments.
  */
-export const CloudBulkReinviteLimit = 4000;
+export const CloudBulkReinviteLimit = 8000;
 
 /**
  * Returns true if the user matches the status, or where the status is `null`, if the user is active (not revoked).


BTreston:
@eliykat Adding you here for some thoughts/opinions. Do we want to look into parallelizing the batched requests? Could speed it up a lot at the cost of some complexity. I'm thinking we can send them in batches of 3/4 with some retry logic in case we hit server rate limits.

apps/web/src/app/admin-console/organizations/members/services/member-actions/member-actions.service.ts
```diff
@@ -162,18 +165,82 @@ export class MemberActionsService {
...
-        failed: userIds.map((id) => ({ id, error: (error as Error).message ?? String(error) })),
-      };
+  /**
+   * Processes user IDs in sequential batches and aggregates results.
+   * @param userIds - Array of user IDs to process
+   * @param batchSize - Number of IDs to process per batch
+   * @param processBatch - Async function that processes a single batch and returns the result
+   * @returns Aggregated bulk action result
+   */
+  private async processBatchedOperation(
```
eliykat: Suggestion: move this to the bottom of the file (or a separate file) for easier navigation in the code.

apps/web/src/app/admin-console/organizations/members/services/member-actions/member-actions.service.ts
```diff
@@ -20,9 +20,12 @@ import { EncryptService } from "@bitwarden/common/key-management/crypto/abstract
 import { ListResponse } from "@bitwarden/common/models/response/list.response";
 import { ConfigService } from "@bitwarden/common/platform/abstractions/config/config.service";
 import { KeyService } from "@bitwarden/key-management";
+import { UserId } from "@bitwarden/user-core";
 
 import { OrganizationUserView } from "../../../core/views/organization-user.view";
 
+export const BATCH_SIZE = 500;
```
eliykat: Should this be increased for cloud when the feature flag is on, per your PR description?

apps/web/src/app/admin-console/organizations/members/services/member-actions/member-actions.service.spec.ts
eliykat, line 1: These tests are good but they overlap; I think Claude does this quite a bit if you're using it to write tests. Please review and see if you can combine or adjust any tests here.

e.g. `should process users in three batches when count requires it` is oddly specific (why three?) and really just a more specific case of `should process users in multiple batches when count exceeds BATCH_SIZE`.

eliykat (commented):
> @eliykat Adding you here for some thoughts/opinions. Do we want to look into parallelizing the batched requests? Could speed it up a lot at the cost of some complexity. I'm thinking we can send them in batches of 3/4 with some retry logic in case we hit server rate limits.

It's a good idea. I am a little worried about the server-side though, it doesn't handle concurrency very well in general (see: Jared's SCIM invite work), so while it wouldn't cause an issue for resending invites - which doesn't change state - it might bite us in the future for other operations. Plus having to handle rate limiting etc, as you pointed out. I think the solution you have now is really clean and can be reused for any bulk action - my view is to say stick with sequential batching for now. Nice work!

One question - where did Product land on providing feedback to the user? Are we covering that here, or in later work?

apps/web/src/app/admin-console/organizations/members/services/member-actions/member-actions.service.ts
BTreston: May need a better name, this is the number of requests per batch, not the bulk limit. I'll swap this over to `REQUESTS_PER_BATCH` or similar. I had a change to the description that I meant to fix but forgot to save, It's updated now.

BTreston:
> > @eliykat Adding you here for some thoughts/opinions. Do we want to look into parallelizing the batched requests? Could speed it up a lot at the cost of some complexity. I'm thinking we can send them in batches of 3/4 with some retry logic in case we hit server rate limits.
> 
> It's a good idea. I am a little worried about the server-side though, it doesn't handle concurrency very well in general (see: Jared's SCIM invite work), so while it wouldn't cause an issue for resending invites - which doesn't change state - it might bite us in the future for other operations. Plus having to handle rate limiting etc, as you pointed out. I think the solution you have now is really clean and can be reused for any bulk action - my view is to say stick with sequential batching for now. Nice work!
> 
> One question - where did Product land on providing feedback to the user? Are we covering that here, or in later work?

All good points. I'll keep it in the back of my mind for now then. User feedback will be next milestone

PR #17807, opened 2025-12-03
[PM-29164] Access Intelligence display for only enterprise

ttalty:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-29164

## 📔 Objective

Added `canAccessAccessIntelligence` to `organization.service.abstraction.ts` which checks the `ProductTierType` is enterprise. 

'organizations-routing.module.ts` uses `canAccessAccessIntelligence` for the route's `canActivate`

`organization-layout.component.ts` added `canShowAccessIntelligenceTab` which calls `canAccessAcessIntelligence` and is used in the html for displaying the Access Intelligence nav item. 

## 📸 Screenshots

Enterprise Org:
image

Non-enterprise Org:
image

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


JimmyVo16 (dismissed):

prograhamming (dismissed):

libs/common/src/vault/tasks/services/default-task.service.ts
```diff
@@ -40,37 +29,22 @@ export class DefaultTaskService implements TaskService {
...
     private apiService: ApiService,
-    private organizationService: OrganizationService,
     private authService: AuthService,
     private notificationService: ServerNotificationsService,
     private messageListener: MessageListener,
   ) {}
 
-  tasksEnabled$ = perUserCache$((userId) => {
-    return this.organizationService.organizations$(userId).pipe(
-      map((orgs) => orgs.some((o) => o.useAccessIntelligence)),
```
Banrion, line 51: ❓ Are we sure we want to remove this completely? Are tasks for everyone now or are they tied to a specific feature? If it's related to access intelligence, shouldn't this follow the same logic that **canAccessAccessIntelligence** does?

bitwarden_license/bit-common/src/dirt/reports/risk-insights/services/view/all-activities.service.ts
```diff
@@ -89,4 +95,26 @@ export class AllActivitiesService {
   setExtendPasswordWidget(hasProgressBar: boolean) {
     this.extendPasswordChangeWidgetSubject$.next(hasProgressBar);
   }
+
+  private resetAllAppsData() {
```
Banrion: 💭 This fixes the bug that activity cards were not updating when switching to a different organization correct? It looks like this sets it all as the default values. I know we are planning to remove this service since it's obsolete but I want to make sure this is fixing the actual bug that was occurring.

libs/common/src/vault/tasks/services/default-task.service.ts
shane-melton: Just want to add on that if this is removed entirely **_all_** users will now start making API requests for security tasks even without an organization.

libs/common/src/vault/tasks/abstractions/task.service.ts
shane-melton, line 1: ⚠️ We should keep `tasksEnabled$` here as its still in the default implementation and is used in the Browser.

libs/common/src/vault/tasks/services/default-task.service.ts
shane-melton, line 1: ⚠️ We should revert all other changes in this file besides switching to the new `canUseAccessIntelligence` property. 

Otherwise, tasks are going to still be fetched for all users.

libs/common/src/vault/tasks/services/default-task.service.ts
ttalty: Fixed this. Added back the `tasksEnabled$` just pointed to the `canUseAccessIntelligence` flag in the org based upon the plan.

libs/common/src/vault/tasks/services/default-task.service.spec.ts
```diff
@@ -320,22 +318,6 @@ describe("Default task service", () => {
       subscription.unsubscribe();
     });
 
-    it("should not subscribe to notifications when no users have tasks enabled", () => {
```
shane-melton, line 323: ⛏️ Can we re-add this test?

libs/common/src/vault/tasks/services/default-task.service.spec.ts
```diff
@@ -163,17 +162,17 @@ describe("Default task service", () => {
...
           {
-            useAccessIntelligence: true,
+            canUseAccessIntelligence: true,
           },
         ] as Organization[]),
       );
     });
 
-    it("should return an empty array if tasks are not enabled", async () => {
+    it("should return no tasks if not present and canUserAccessIntelligence is false", async () => {
```
Banrion, line 171: 📝 This looks like a duplicate test

libs/common/src/vault/tasks/services/default-task.service.spec.ts
```diff
@@ -111,7 +111,6 @@ describe("Default task service", () => {
       const result = await firstValueFrom(tasks$("user-id" as UserId));
 
       expect(result.length).toBe(0);
-      expect(mockApiSend).not.toHaveBeenCalled();
```
Banrion, line 114: ❓ If `canUseAccessIntelligence` is false, do we expect to call the api? This can be kept instead of removed if we don't want to fetch tasks in that case

Banrion (approved):
There is one question, but I don't think it's blocking. I just want to make sure we are certain on functionality and the test cases are representing that.

eliykat (approved):

PR #17811, opened 2025-12-03
[PM-29178] Do not trigger an update notification if the entered password matches a stored cipher with the same value and matching username

jprusik:
## 🎟️ Tracking

[PM-29178](https://bitwarden.atlassian.net/browse/PM-29178)

## 📔 Objective

If the user has two or more ciphers with matching usernames for a given page, the update notification will always trigger, even if no changes to the login credentials were made. This PR resolves that issue.

Related: https://github.com/bitwarden/clients/issues/17405

## 🦮 Reviewer guidelines

[PM-29178]: https://bitwarden.atlassian.net/browse/PM-29178?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/browser/src/autofill/background/notification.background.ts
+++ b/apps/browser/src/autofill/background/notification.background.ts
@@ -627,11 +627,11 @@ export default class NotificationBackground {
     }
 
     const username: string | null = data.username || null;
-    const currentPassword = data.password || null;
-    const newPassword = data.newPassword || null;
+    const currentPasswordFieldValue = data.password || null;
+    const newPasswordFieldValue = data.newPassword || null;
 
-    if (authStatus === AuthenticationStatus.Locked && newPassword !== null) {
-      await this.pushChangePasswordToQueue(null, loginDomain, newPassword, tab, true);
+    if (authStatus === AuthenticationStatus.Locked && newPasswordFieldValue !== null) {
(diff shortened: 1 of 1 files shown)


apps/browser/src/autofill/background/notification.background.ts
```diff
@@ -657,35 +657,49 @@ export default class NotificationBackground {
...
+        loginDomain,
+        currentPasswordFieldValue,
+        tab,
+      );
+
+      return true;
     }
 
-    if (newPassword) {
+    if (newPasswordFieldValue) {
```
jprusik, line 696: There's no guarantee that this entered value is different from all of the matching vault-stored ciphers with the same username. Although we _probably_ don't want to prompt an update in this case (at least one stored cipher matches what was entered in the form), because the field was authoritatively identified as a new password field, we may want to give the user the opportunity to update anyway, since they'd expect to see it after a dedicated credential update form.

We should probably filter out the exact matches (username and password are the same), at minimum, since there's nothing to change.

Since this is unrelated to the bug at hand, let's consider this for followup work with some Product guidance.

dan-livefront (approved):
This overall looks like an improvement to the robustness of our notification background logic with no obvious risks. Nice work Jon!

PR #17820, opened 2025-12-04
[PM-26514] Archive With Non Premium User

Jingo88:
## 🎟️ Tracking

[PM-26514](https://bitwarden.atlassian.net/browse/PM-26514)

## 📔 Objective

This PR adds logic for users who's subscription has lapsed and they no longer have premium, however they have items in archive.

- Archive Items will only show (Edit, Unarchive, Delete) in the more options menu
- If a user were to edit the item inside of Archive, that item will be unarchived and be in their individual vault
- If the user no longer has premium and there are items in archive show an info banner on top with the link to restart their premium subscription
- Edit modal will have a button saying “Unarchive and Restore” or “Unarchive and Save”
- Edit modal will have a “Archived Item Icon” at the top next to item name

NOTE: most of the changes in `vault-item-dialog.component.html` are updating the `ngIf` statements. Can be found in [this commit](https://github.com/bitwarden/clients/commit/dabbbffcdab80893c0751458d82a53ba4bdb00ce)

NOTE 2: `hasPasswordToCopy` and `hasUsernameToCopy` were unused methods. Removed.

## 📸 Screen Recording

https://github.com/user-attachments/assets/8daf679b-d76a-4136-9ec5-e6d3950e0c02

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

[PM-26514]: https://bitwarden.atlassian.net/browse/PM-26514?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/web/src/app/vault/components/vault-item-dialog/vault-item-dialog.component.html
+++ b/apps/web/src/app/vault/components/vault-item-dialog/vault-item-dialog.component.html
@@ -68,11 +68,7 @@
       #submitBtn
       [hidden]="showCipherView || showRestore"
     >
-      @if (cipherIsArchived && !(userHasPremium$ | async)) {
-        {{ "unArchiveAndSave" | i18n }}
-      } @else {
-        {{ "save" | i18n }}
-      }
+      {{ submitButtonText$ | async }}
     </button>
 
     @if (showCancel) {
--- a/apps/web/src/app/vault/components/vault-item-dialog/vault-item-dialog.component.spec.ts
+++ b/apps/web/src/app/vault/components/vault-item-dialog/vault-item-dialog.component.spec.ts
@@ -1,6 +1,7 @@
 import { ComponentFixture, TestBed } from "@angular/core/testing";
 import { provideNoopAnimations } from "@angular/platform-browser/animations";
 import { ActivatedRoute, Router } from "@angular/router";
+import { of } from "rxjs";
 
 import { ApiService } from "@bitwarden/common/abstractions/api.service";
 import { EventCollectionService } from "@bitwarden/common/abstractions/event/event-collection.service";
@@ -95,6 +96,10 @@ describe("VaultItemDialogComponent", () => {
 
     fixture = TestBed.createComponent(TestVaultItemDialogComponent);
     component = fixture.componentInstance;
+    Object.defineProperty(component, "userHasPremium$", {
+      get: () => of(false),
+      configurable: true,
+    });
     fixture.detectChanges();
   });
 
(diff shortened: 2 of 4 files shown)


libs/components/src/dialog/dialog/dialog.component.html
```diff
@@ -34,16 +34,21 @@
...
-        bitIconButton="bwi-close"
-        buttonType="main"
-        size="default"
-        bitDialogClose
-        [label]="'close' | i18n"
-      ></button>
-    }
+
+    <div class="tw-flex tw-items-center tw-justify-center">
+      <ng-content select="[bitDialogBadge]"></ng-content>
```
vleague2: Naming this after the badge is too specific for reusability. Maybe something like `bitDialogHeaderEnd`? I would also like to see a story added to `dialog.stories.ts` for this, preferably with a long title so that we can see how the title truncates but the items in this slot remain visible.

libs/components/src/dialog/dialog/dialog.stories.ts
```diff
@@ -94,6 +95,11 @@ export const Default: Story = {
         <ng-container bitDialogTitle>
           <span bitBadge variant="success">Foobar</span>
         </ng-container>
+
+        <ng-container bitDialogHeaderEnd>
+          <span bitBadge>{{ headerEndContent }}</span>
```
vleague2: Putting this in the `Default` story is leaving a lil empty badge in all the stories lol. I'd just copy the full template into the `HeaderEnd` story and remove this slot usage from the `Default` story.

For the long-ness part, I don't think I explained very well -- I would like the `title` prop to be long content, and the header end/badge to be normal (like `Archived` or whatever you're using for the feature). Basically I want the title content to be so long that our normal truncation kicks in (which I think happens after 2 lines -- see the `LongTitleSentence` story). This should show that the new `headerEnd` slot with the badge inside is unaffected by the title truncation while also covering the new slot with the snapshot test. (Ping me if this is not making sense!!)

libs/components/src/dialog/dialog/dialog.stories.ts
```diff
@@ -94,6 +95,11 @@ export const Default: Story = {
         <ng-container bitDialogTitle>
```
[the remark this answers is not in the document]
vleague2: This is a different existing badge, not what my feedback was about

nick-livefront (approved):

PR #17832, opened 2025-12-05
[PM-28079] Add attributes to filter for the mutationObserver

jholland-livefront:
## 🎟️ Tracking

[PM-28079](https://bitwarden.atlassian.net/browse/PM-28079)
## 📔 Objective

As part of @jprusik's performance research he determined we can specify the attributes we want to track in the `MutationObserver`.  I have added 23 attributes we want to track with others being excluded.

I implemented a simple `MutationObserver` counter locally to get a rough idea of the performance difference.  For the Simple Login on webtests.dev the `MutationObserver` count dropped from 9 to 7 with the `attributeFilter` added.

I am looking for suggestions any additions or deletions to this attribute list.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

[PM-28079]: https://bitwarden.atlassian.net/browse/PM-28079?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/common/src/autofill/constants/index.ts
```diff
@@ -28,6 +28,37 @@ export const EVENTS = {
...
+  ID: "id",
+  NAME: "name",
+  CLASS: "class",
+  ARIA_LABEL: "aria-label",
+  ARIA_HIDDEN: "aria-hidden",
+  ARIA_DISABLED: "aria-disabled",
+  ARIA_HASPOPUP: "aria-haspopup",
+  POPOVER: "popover",
+  POPOVERTARGET: "popovertarget",
+  POPOVERTARGETACTION: "popovertargetaction",
```
theMickster, line 58: ❌ **CRITICAL**: Missing `data-stripe` attribute breaks Stripe integration tracking

<details>

The `data-stripe` attribute is collected at line 427 in `collect-autofill-content.service.ts`:
```typescript
"data-stripe": this.getPropertyOrAttribute(element, "data-stripe"),
```

And has a mutation handler at line 1360:
```typescript
"data-stripe": () => updateAttribute("data-stripe"),
```

But it's **missing from `AUTOFILL_ATTRIBUTES`**. This means if a Stripe integration dynamically updates the `data-stripe` attribute on a form field, the MutationObserver won't detect it and autofill data will become stale.

**Fix:** Add to the constant:
```typescript
DATA_STRIPE: "data-stripe",
```

Reference: [Stripe Elements integration tracking](https://github.com/bitwarden/clients/blob/main/apps/browser/src/autofill/services/collect-autofill-content.service.ts#L427)
</details>

libs/common/src/autofill/constants/index.ts
```diff
@@ -28,6 +28,37 @@ export const EVENTS = {
...
+  METHOD: "method",
+  MAXLENGTH: "maxlength",
+  TYPE: "type",
+  VALUE: "value",
+  CHECKED: "checked",
+  DISABLED: "disabled",
+  READONLY: "readonly",
+  AUTOCOMPLETE: "autocomplete",
+  TABINDEX: "tabindex",
+  TITLE: "title",
```
theMickster: ⚠️ **IMPORTANT**: Missing legacy autocomplete attributes may break mutation detection

<details>

The code checks three autocomplete attribute variants (line 457-462 in `collect-autofill-content.service.ts`):
```typescript
private getAutoCompleteAttribute(element: ElementWithOpId<FormFieldElement>): string {
  return (
    this.getPropertyOrAttribute(element, "x-autocompletetype") ||
    this.getPropertyOrAttribute(element, "autocompletetype") ||
    this.getPropertyOrAttribute(element, "autocomplete")
  );
}
```

But only `autocomplete` is in the filter. The mutation handler at line 1351 calls `getAutoCompleteAttribute()` which checks all three, **but mutations won't be detected** for `x-autocompletetype` or `autocompletetype` changes since they're not in the filter.

**Impact:** Legacy sites using `x-autocompletetype` or `autocompletetype` (older IE/Safari patterns) won't have their autocomplete changes detected.

**Fix:** Add these legacy attributes:
```typescript
X_AUTOCOMPLETETYPE: "x-autocompletetype",
AUTOCOMPLETETYPE: "autocompletetype",
```

While these are legacy attributes, they're explicitly supported by the codebase and should be tracked if a page uses them dynamically.
</details>

libs/common/src/autofill/constants/index.ts
```diff
@@ -28,6 +28,37 @@ export const EVENTS = {
...
+  REL: "rel",
+  TAGNAME: "tagname",
+  DATA_LABEL: "data-label",
+  ID: "id",
+  NAME: "name",
+  CLASS: "class",
+  ARIA_LABEL: "aria-label",
+  ARIA_HIDDEN: "aria-hidden",
+  ARIA_DISABLED: "aria-disabled",
```
theMickster: 💭 **QUESTION**: Should `placeholder` attribute be tracked for mutations?

<details>
<summary>Details and context</summary>

The `placeholder` attribute is collected during initial field scanning (line 405 in `collect-autofill-content.service.ts`):
```typescript
placeholder: this.getPropertyOrAttribute(element, "placeholder"),
```

However:
1. There's **no mutation handler** for it in `updateAutofillFieldElementData()` (lines 1329-1372)
2. It's **not in this attribute filter**

**Two possible interpretations:**

**Option A - Intentionally excluded:** Placeholder is only used for initial label detection and doesn't need dynamic tracking. If a page changes placeholders dynamically, we don't care because we've already captured the label.

**Option B - Missing handler:** We should track placeholder changes and add a mutation handler for it.

**Question for team:** Is placeholder intentionally excluded from mutation tracking, or should we add both the attribute filter entry and a mutation handler?

Context: Placeholder text is used as a fallback label in the autofill system and some modern SPAs (especially progressive web apps) do change placeholders dynamically based on user state.
</details>

libs/common/src/autofill/constants/index.ts
```diff
@@ -28,6 +28,37 @@ export const EVENTS = {
...
+  AUTOCOMPLETE: "autocomplete",
+  TABINDEX: "tabindex",
+  TITLE: "title",
+  REL: "rel",
+  TAGNAME: "tagname",
+  DATA_LABEL: "data-label",
+  ID: "id",
+  NAME: "name",
+  CLASS: "class",
+  ARIA_LABEL: "aria-label",
```
theMickster: 💭 **QUESTION**: Form `class` attribute tracked but no mutation handler exists

<details>
<summary>Details and context</summary>

The `class` attribute is in the filter and is collected for forms (line 235 in `collect-autofill-content.service.ts`):
```typescript
htmlClass: this.getPropertyOrAttribute(formElement, "class"),
```

It also has a mutation handler for **fields** (line 1341):
```typescript
class: () => updateAttribute("htmlClass"),
```

However, there's **no handler for form `class` changes** in `updateAutofillFormElementData()` (lines 1305-1310) - only `action`, `name`, `id`, and `method` are handled.

**Question:** Is this intentional, or should we add a `class` handler to form updates?

```typescript
// In updateAutofillFormElementData:
const updateActions: Record<string, CallableFunction> = {
  action: () => (dataTarget.htmlAction = this.getFormActionAttribute(element)),
  name: () => updateAttribute("htmlName"),
  id: () => updateAttribute("htmlID"),
  method: () => updateAttribute("htmlMethod"),
  // Missing: class: () => updateAttribute("htmlClass"),
};
```

If form class changes aren't needed for autofill logic, we could remove it from the filter to reduce mutation observer overhead (the goal of this PR).
</details>

theMickster:
**Overall Assessment:** REQUEST CHANGES

**Critical Issues**:
- libs/common/src/autofill/constants/index.ts:59 - Missing `data-stripe` attribute (has mutation handler)
- apps/browser/src/autofill/services/collect-autofill-content.service.ts:1343 - Pre-existing bug: title mutation handler incorrectly updates tabindex

**Important Issues**:
- libs/common/src/autofill/constants/index.ts:46 - Missing legacy autocomplete attributes (`x-autocompletetype`, `autocompletetype`)

See inline comments for details and recommended fixes.

---

**Note on pre-existing bug (line 1343):** While this bug exists in unchanged code, it becomes relevant because this PR enables title mutation tracking. The bug causes title attribute changes to incorrectly overwrite tabindex data. Consider fixing it as part of this PR or in a follow-up.

libs/common/src/autofill/constants/index.ts
```diff
@@ -28,6 +28,40 @@ export const EVENTS = {
   SUBMIT: "submit",
 } as const;
 
+/**
+ * HTML attributes observed by the MutationObserver for autofill form/field tracking.
+ * If you need to observe a new attribute, add it here.
+ */
+export const AUTOFILL_ATTRIBUTES = {
```
jprusik, line 35: We should be consuming these values in places like `CollectAutofillContentService` instead of hardcoded strings, which will reinforce the relationship between the mutation observer attribute filter and the attributes consumed downstream (and potentially build linting checks to prevent consuming attributes that are not in the mutation filter).

libs/common/src/autofill/constants/index.ts
jprusik: yes

libs/common/src/autofill/constants/index.ts
jprusik: ➕

libs/common/src/autofill/constants/index.ts
jprusik: ➕

theMickster:
Hi @jprusik & @jholland-livefront 👋🏼
I am a little perplexed (and slightly concerned) why my identity got tagged by the Claude Code Review agent in your PR here. I re-ran the Claude Code Review action last week following a couple failures as Patrick and I were working through some issues getting the agent based prompting to work correctly.  Sorry if this caused confusion or made it seem like I was fishing in someone else's pond lol.

libs/common/src/autofill/constants/index.ts
jholland-livefront: ~~@jprusik I wanted to get your input.  Should we add `class` to `updateActions` or should it be removed from this list?~~
I will push the changes later today, but I went ahead with the suggestion to add the missing `class` handler.

jprusik: sorry - yeah, we want to handle it; sometimes it's the only (albeit low-quality) qualifier of the input identity

libs/common/src/autofill/constants/index.ts
```diff
@@ -28,6 +28,41 @@ export const EVENTS = {
...
```
jprusik, line 35: Nit; it might be nice to have these sorted easy scanning

apps/browser/src/autofill/services/collect-autofill-content.service.ts
```diff
@@ -393,38 +401,38 @@ export class CollectAutofillContentService implements CollectAutofillContentServ
...
 
     let autofillFieldLabels = {};
-    const elementType = this.getAttributeLowerCase(element, "type");
+    const elementType = this.getAttributeLowerCase(element, AUTOFILL_ATTRIBUTES.TYPE);
     if (elementType !== "hidden") {
       autofillFieldLabels = {
         "label-tag": this.createAutofillFieldLabelTag(element as FillableFormFieldElement),
-        "label-data": this.getPropertyOrAttribute(element, "data-label"),
-        "label-aria": this.getPropertyOrAttribute(element, "aria-label"),
+        "label-data": this.getPropertyOrAttribute(element, AUTOFILL_ATTRIBUTES.DATA_LABEL),
```
jprusik, line 417: non-blocking nit; since the keys are specifically matching the attribute names/formatting, we could use the `AUTOFILL_ATTRIBUTES.DATA_LABEL` constant values for keys too 🤷 

```suggestion
        [AUTOFILL_ATTRIBUTES.DATA_LABEL]: this.getPropertyOrAttribute(element, AUTOFILL_ATTRIBUTES.DATA_LABEL),
```

apps/browser/src/autofill/services/collect-autofill-content.service.ts
```diff
@@ -946,6 +954,7 @@ export class CollectAutofillContentService implements CollectAutofillContentServ
     this.mutationObserver = new MutationObserver(this.handleMutationObserverMutation);
     this.mutationObserver.observe(document.documentElement, {
       attributes: true,
+      attributeFilter: Object.values(AUTOFILL_ATTRIBUTES),
```
jprusik, line 967: 🙌

jprusik: nit; comment to reinforce the importance of the filter to future authors

```suggestion
      /** Mutations to node attributes NOT on this list will not be observed! */
      attributeFilter: Object.values(AUTOFILL_ATTRIBUTES),
```

apps/browser/src/autofill/services/collect-autofill-content.service.ts
```diff
@@ -1339,25 +1349,41 @@ export class CollectAutofillContentService implements CollectAutofillContentServ
...
+      tabindex: () => updateAttribute(AUTOFILL_ATTRIBUTES.TABINDEX),
+      title: () => updateAttribute(AUTOFILL_ATTRIBUTES.TITLE),
+      rel: () => updateAttribute(AUTOFILL_ATTRIBUTES.REL),
+      tagname: () => updateAttribute(AUTOFILL_ATTRIBUTES.TAGNAME),
+      type: () => (dataTarget.type = this.getAttributeLowerCase(element, AUTOFILL_ATTRIBUTES.TYPE)),
       value: () => (dataTarget.value = this.getElementValue(element)),
-      checked: () => (dataTarget.checked = this.getAttributeBoolean(element, "checked")),
-      disabled: () => (dataTarget.disabled = this.getAttributeBoolean(element, "disabled")),
-      readonly: () => (dataTarget.readonly = this.getAttributeBoolean(element, "readonly")),
+      placeholder: () => updateAttribute(AUTOFILL_ATTRIBUTES.PLACEHOLDER),
```
jprusik: Oh dang, we weren't updating the placeholder value changes? That might fix some issues all on it's own 🤔

apps/browser/src/autofill/services/collect-autofill-content.service.ts
jprusik, line 967: We have other mutation observers we might consider this filtering strategy for as well, but let's not go all-in on this strategy just yet, since it a perf gain at the expense of maintainability.

libs/common/src/autofill/constants/index.ts
```diff
@@ -28,6 +28,41 @@ export const EVENTS = {
...
+  TABINDEX: "tabindex",
+  TITLE: "title",
+  REL: "rel",
+  TAGNAME: "tagname",
+  DATA_LABEL: "data-label",
+  DATA_STRIPE: "data-stripe",
+  ID: "id",
+  NAME: "name",
+  CLASS: "class",
+  ARIA_LABEL: "aria-label",
```
jprusik: We may also need `aria-labelledby` and `aria-describedby` as they are read by `getSubmitButtonKeywordsSet` which is ultimately downstream of the mutation observer in `apps/browser/src/autofill/services/collect-autofill-content.service.ts` and mutations of which we'd want to potentially trigger side-effects

apps/browser/src/autofill/services/collect-autofill-content.service.ts
jholland-livefront: I could do that with others, but I don't know that this is a good idea for this one because the `AUTOFILL_ATTRIBUTES.DATA_LABEL` maps to `data-label`, not `label-data` like the key you suggest replacing.

jprusik: Indeed! good spot. Let's leave it all alone for now.

libs/common/src/autofill/constants/index.ts
```diff
@@ -28,6 +28,43 @@ export const EVENTS = {
...
+  METHOD: "method",
+  NAME: "name",
+  PLACEHOLDER: "placeholder",
+  POPOVER: "popover",
+  POPOVERTARGET: "popovertarget",
+  POPOVERTARGETACTION: "popovertargetaction",
+  READONLY: "readonly",
+  REL: "rel",
+  TABINDEX: "tabindex",
+  TAGNAME: "tagname",
```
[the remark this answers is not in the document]
jprusik: This is a good point - changing a tag will (ostensibly) result in a remove/add node mutation, which won't be relevant to the attribute filter

apps/browser/src/autofill/services/collect-autofill-content.service.ts
```diff
@@ -1339,25 +1350,41 @@ export class CollectAutofillContentService implements CollectAutofillContentServ
...
       class: () => updateAttribute("htmlClass"),
-      tabindex: () => updateAttribute("tabindex"),
-      title: () => updateAttribute("tabindex"),
-      rel: () => updateAttribute("rel"),
-      tagname: () => (dataTarget.tagName = this.getAttributeLowerCase(element, "tagName")),
-      type: () => (dataTarget.type = this.getAttributeLowerCase(element, "type")),
+      tabindex: () => updateAttribute(AUTOFILL_ATTRIBUTES.TABINDEX),
+      title: () => updateAttribute(AUTOFILL_ATTRIBUTES.TITLE),
+      rel: () => updateAttribute(AUTOFILL_ATTRIBUTES.REL),
+      tagname: () => updateAttribute(AUTOFILL_ATTRIBUTES.TAGNAME),
```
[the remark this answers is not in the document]
jprusik: ~Ignore this. While `getAttributeLowerCase` implies we're only mapping attributes here, it's a proxy for `getPropertyOrAttribute`. That is, I believe this is an intentional mapping, even though it won't be updated on that property change (see earlier comment), it will still be part of the mapped data from other mutations.~

*edit*: Actually, taking a closer look, this would only be called if `mutation.attributeName` were `tagname`, which should never happen, so removing this should be safe after all...

`value` would be similar, as it's also a property... 🤔

libs/common/src/autofill/constants/index.ts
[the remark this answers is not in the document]
jprusik: see also `VALUE` for the same reason; property, not an attribute.

jprusik:
We should probably take a look at removing `value` and `tagname` from the constants and mutation handler logic, since those values will never fire (I did a bit of light testing to verify, as well)

apps/browser/src/autofill/services/collect-autofill-content.service.ts
```diff
@@ -1339,25 +1350,45 @@ export class CollectAutofillContentService implements CollectAutofillContentServ
...
+        (dataTarget.checked = this.getAttributeBoolean(element, AUTOFILL_ATTRIBUTES.CHECKED)),
+      disabled: () =>
+        (dataTarget.disabled = this.getAttributeBoolean(element, AUTOFILL_ATTRIBUTES.DISABLED)),
+      readonly: () =>
+        (dataTarget.readonly = this.getAttributeBoolean(element, AUTOFILL_ATTRIBUTES.READONLY)),
       autocomplete: () => (dataTarget.autoCompleteType = this.getAutoCompleteAttribute(element)),
+      autocompletetype: () =>
+        (dataTarget.autoCompleteType = this.getAutoCompleteAttribute(element)),
+      "x-autocompletetype": () =>
+        (dataTarget.autoCompleteType = this.getAutoCompleteAttribute(element)),
```
jprusik: I don't think we need this one; if `x-autocompletetype` is present, it will get mapped to `autocompletetype` (directly above) via `getAutoCompleteAttribute`.

apps/browser/src/autofill/services/collect-autofill-content.service.ts
```diff
@@ -456,9 +464,9 @@ export class CollectAutofillContentService implements CollectAutofillContentServ
    */
   private getAutoCompleteAttribute(element: ElementWithOpId<FormFieldElement>): string {
     return (
-      this.getPropertyOrAttribute(element, "x-autocompletetype") ||
-      this.getPropertyOrAttribute(element, "autocompletetype") ||
-      this.getPropertyOrAttribute(element, "autocomplete")
+      this.getPropertyOrAttribute(element, AUTOFILL_ATTRIBUTES.X_AUTOCOMPLETE_TYPE) ||
+      this.getPropertyOrAttribute(element, AUTOFILL_ATTRIBUTES.AUTOCOMPLETE_TYPE) ||
+      this.getPropertyOrAttribute(element, AUTOFILL_ATTRIBUTES.AUTOCOMPLETE)
```
jprusik: It won't matter unless more than one of these attributes is present (which seems unlikely), but I think we'd want to prioritize the standard attribute and check the legacy attributes as fallbacks

apps/browser/src/autofill/services/collect-autofill-content.service.ts
jprusik, line 1357: It might be worth ordering these as well

apps/browser/src/autofill/services/collect-autofill-content.service.ts
jprusik: nvm, I see the action key is called directly from the mutation attribute name

jprusik (dismissed):
LGTM; some small nits, but mostly organizational

jprusik (approved):

jprusik:
Rebased because merge required a verified signature on every commit (pull commit was not sufficient)

Screenshot 2026-01-09 at 10 13 02 AM

PR #17834, opened 2025-12-05
[PM-29241] Set SDK log level to error on CLI

quexten:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-29241

## 📔 Objective

On CLI, the info log level is too much since info is expected to regularly happen during normal operation. However, the output of the CLI should not contain info logs and will interfere with e.g. automations. Thus, we set the level to error.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/apps/cli/src/platform/services/cli-sdk-load.service.ts
+++ b/apps/cli/src/platform/services/cli-sdk-load.service.ts
@@ -3,6 +3,7 @@ import * as sdk from "@bitwarden/sdk-internal";
 
 export class CliSdkLoadService extends SdkLoadService {
   async load(): Promise<void> {
+    // CLI uses stdout for user interaction / automations so we cannot log info / debug here.
     SdkLoadService.logLevel = sdk.LogLevel.Error;
     const module = await import("@bitwarden/sdk-internal/bitwarden_wasm_internal_bg.wasm");
     (sdk as any).init(module);


apps/cli/src/platform/services/cli-sdk-load.service.ts
```diff
@@ -6,4 +6,14 @@ export class CliSdkLoadService extends SdkLoadService {
...
+
+  override async loadAndInit(): Promise<void> {
+    try {
+      await this.load();
+      sdk.init_sdk(sdk.LogLevel.Error);
+      (SdkLoadService as any).markAsReady();
+    } catch (error) {
+      (SdkLoadService as any).markAsFailed(error);
+    }
+  }
```
dereknance: 💭 Since the only difference between the parent method and the override is setting the log level, what do you think about adding a log level property to the abstract class, that `load()` can set, and the parent `loadAndInit()` will pass to `init_sdk()`?

```ts
export abstract class SdkLoadService {
  // . . .
  private static logLevel?: LogLevel = null;

  // . . .
  async loadAndInit(): Promise<void> {
    try {
      await this.load();
      init_sdk(SdkLoadService.logLevel);
      SdkLoadService.markAsReady();
    } catch (error) {
      SdkLoadService.markAsFailed(error);
    }
  }
  // . . .
}
```

```ts
export class CliSdkLoadService extends SdkLoadService {
  async load(): Promise<void> {
    SdkLoadService.logLevel = sdk.LogLevel.Error;
    const module = await import("@bitwarden/sdk-internal/bitwarden_wasm_internal_bg.wasm");
    (sdk as any).init(module);
  }
```

Or, if mutating the log level inside `load()` is smelly, maybe another override?

```ts
export abstract class SdkLoadService {
  // . . .
  async loadAndInit(): Promise<void> {
    try {
      await this.load();
      init_sdk(this.logLevel());
      SdkLoadService.markAsReady();
    } catch (error) {
      SdkLoadService.markAsFailed(error);
    }
  }
  // . . .
  protected logLevel = (): LogLevel | null => null;
}
```

```ts
export class CliSdkLoadService extends SdkLoadService {
  // . . .
  override logLevel = () => sdk.LogLevel.Error;
}
```

quexten: Makes sense. I've applied the first approach, but instead of making it null-by-default, I set it to INFO by default. I felt this made it clearer to understand default behavior, because not doing so requires me to open the SDK code and check what the logging behavior is without setting it.

dereknance: I agree that using INFO instead of null is definitely clearer.

PR #17846, opened 2025-12-05
Inform user if Desktop client already running

neuronull:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-29288

## 📔 Objective

Currently, if another instance of Bitwarden is already running, there is no message to the user about the problem, the app silently exits and an unhelpful stdout is displayed.

## 📸 Screenshots

Screenshot 2025-12-19 at 08 57 35

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


apps/desktop/src/main/window.main.ts
```diff
@@ -122,6 +122,7 @@ export class WindowMain {
         if (!isMacAppStore()) {
           const gotTheLock = app.requestSingleInstanceLock();
           if (!gotTheLock) {
+            dialog.showErrorBox("Error", "An instance of Bitwarden Desktop is already running.");
```
quexten, line 125: Please make sure this is tested in combination with especially linux snap/flatpak/appimage and SSO login (tde or MP), or at least syncing up with auth's QA members so this is covered in regression.

I recall looking a fair bit into issues with SSO callbacks on linux while on auth, and while I believe they should be fixed, I want to make sure there is no regression here.

neuronull: Per our discussion out of band, I ensured that the snap, and appimage builds are able to login with SSO as normal.

I'm in the process of validating flatpak, ran into some problems there testing it.

quexten:
@coltonhurst Discussed this in DM with @neuronull; The main concern here from my side was regressions w.r.t. sso login. I wanted to make sure that the SSO flow does not suddenly trigger these alert boxes.

There was some issues around flatpak testing, because the flatpak manifest used in clients CI was different than the one in flathub, and did not support SSO. I've fixed that here: https://github.com/bitwarden/clients/pull/18018, so testing flatpak should be unblocked.

neuronull:
"hold" label added to note that this shouldn't be merged until January.

apps/desktop/src/main/window.main.ts
neuronull: Validated the flatpak with SSO as well. I will still send it to QA

Meteoni:
Not a great change IMO. 
This makes it so that a user is forced to launch bitwarden from clicking on the teeny-tiny tray icon instead just of pinning the app to the taskbar and opening it from there, which has been the more convenient & old-fashioned way to get to the app's main interface. 
I'd like to see this change reverted at least for windows users - whose user experience has been made quite cumbersome and frustrating for this very reason, and before that, I'd personally stay on an older release.

lawyerz8:
This change is causing me to seek out an older release and install that instead. This is extremely cumbersome and pops up every single time i click on the taskbar tray icon to open up bitwarden. Add that with the auto-minimise when copying passwords out - this comes up hundreds of time in a day.

anddam:
> Currently, if another instance of Bitwarden is already running, there is no message to the user about the problem, the app silently exits and an unhelpful stdout is displayed.

What actually was "the problem" referred to in the OP?

If the user starts a GUI app and it is already running just pop it up to frontmost window, it's not like Bitwarden is a document based app on macOS and an user expects to spawn multiple main windows.

At any rate the dialog does not solve the problem, the user expected either the GUI to popup or the app to run in case it was set to start minimized. The second case is already taken care of by doing nothing, so the proper solution would have been to show the running instance, but IIRC that was the behavior up till the recent update.

I am downgrading but still puzzled on what problem this PR was actually trying to solve, description is not clear.

neuronull:
> What actually was "the problem" referred to in the OP?
> 
> If the user starts a GUI app and it is already running just pop it up to frontmost window, it's not like Bitwarden is a document based app on macOS and an user expects to spawn multiple main windows.
> 
> At any rate the dialog does not solve the problem, the user expected either the GUI to popup or the app to run in case it was set to start minimized. The second case is already taken care of by doing nothing, so the proper solution would have been to show the running instance, but IIRC that was the behavior up till the recent update.
> 
> I am downgrading but still puzzled on what problem this PR was actually trying to solve, description is not clear.

👋  Hey, I can add a bit more color to this. 

Firstly, the fact that we didn't catch this in our internal testing streams exposed a gap, and we're working on closing that so that things like this will be caught before releasing to production in the future.

The original problem presents itself when running the desktop client outside of arguably common invocations. We'll definitely take a different approach than a pop-up to addressing it when re-visiting this!

Arxareon:
> If the user starts a GUI app and it is already running just pop it up to frontmost window, it's not like Bitwarden is a document based app on macOS and an user expects to spawn multiple main windows.

Yes, I agree. The actual solve is not this popup that annoyingly pops up not every time it shouldn't, the real solution (how it basically worked for me in the past, I never noticed multiple instances when Bitwarden was already running minimized) is to not spawn another instance and just bring up a GUI window for the already running instance to the foreground. That's it.

PR #17860, opened 2025-12-08
Add debouncing to `setupOverlayOnField`

jholland-livefront:
## 🎟️ Tracking

[PM-28080](https://bitwarden.atlassian.net/browse/PM-28080)

## 📔 Objective

Setting up the overlay can be an expensive operation when the page is changing rapidly.  This change adds a debounce of 100ms so the overlay is only set up when needed and not after every page mutation.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

[PM-28080]: https://bitwarden.atlassian.net/browse/PM-28080?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


jprusik:
> **Required Actions:**
> 
> 1. ❌ Fix the memory leak by clearing timeouts in `deleteCachedAutofillElement()`
> 2. 🌱 Add test coverage for the fix (recommended but not blocking)

@claude , #1 is addressed in [6d665c2](https://github.com/bitwarden/clients/pull/17860/commits/6d665c27c251f269a9374b3e7e4fab5d71b24394); please re-review

apps/browser/src/autofill/services/collect-autofill-content.service.ts
```diff
@@ -1254,6 +1256,13 @@ export class CollectAutofillContentService implements CollectAutofillContentServ
     if (this.autofillFieldElements.has(element)) {
       this.autofillFieldElements.delete(element);
     }
+
+    // Clear pending overlay setup timeout to prevent memory leak
+    const pendingTimeout = this.pendingOverlaySetup.get(element);
+    if (pendingTimeout) {
+      globalThis.clearTimeout(pendingTimeout);
+      this.pendingOverlaySetup.delete(element);
+    }
```
dan-livefront, line 1345: Should we place this within the `if (this.autofillFieldElements.has(element)) block` to only clear the pending overlay when we actually delete a field from the cache?

dan-livefront (commented):
nit: Since `setupOverlayListeners()` invokes `setupOverlayOnField` for each field, we end up scheduling a timeout per field to clear `pendingOverlaySetup`. This doesn’t leak, but it creates some avoidable timer churn.

apps/browser/src/autofill/services/collect-autofill-content.service.ts
jprusik: Hrm; my thinking was that if the element isn't in the cache but somehow still has a timeout, we still want to clear it.

We can also run the `delete`s without checking for the delete target first, since it won't throw if it's missing (and save a lookup at scale)

PR #17863, opened 2025-12-08
[PM-26463] Refactor members components

BTreston:
## 🎟️ Tracking
https://bitwarden.atlassian.net/browse/PM-26463

## 📔 Objective
Major refactor for the members component.

  Key Changes:

  1. Architectural Refactor

  - Removed: base-members.component.ts (245 lines deleted) - eliminated inheritance-based approach
  - Added: member-component-utils.ts (124 lines) - utility functions for shared logic
  - Pattern shift: From base class inheritance → composition with utility functions

  2. Organization Members Component (apps/web/src/app/admin-console/organizations/members/)

  - Modernized to OnPush change detection (was using Default strategy)
  - Dependency injection: Constructor-based → inject() function pattern
  - Reactive state: Added FormControl for search, BehaviorSubject for status filtering
  - Added signals: firstLoaded signal for loading state
  - Better separation: Extracted bulk action flags configuration to configureMemberFlags()
  - Template: 576 lines restructured for better reactivity

  3. Provider Members Component (apps/web/src/app/admin-console/providers/manage/members.component.ts)

  - Similar refactor pattern applied (378 lines modified)
  - Uses configureProviderMemberFlags() utility
  - Template restructured (200 lines)

  4. Shared Utilities (member-component-utils.ts)

  - MembersTableDataSource & ProvidersTableDataSource classes
  - configureMemberFlags() - reactive bulk action visibility logic
  - configureProviderMemberFlags() - provider-specific bulk actions
  - Type definitions: BulkFlags, BulkMemberFlags, StatusType

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/apps/web/src/app/admin-console/organizations/members/members.component.ts
+++ b/apps/web/src/app/admin-console/organizations/members/members.component.ts
@@ -20,7 +20,6 @@ import {
 
 import { UserNamePipe } from "@bitwarden/angular/pipes/user-name.pipe";
 import { OrganizationService } from "@bitwarden/common/admin-console/abstractions/organization/organization.service.abstraction";
-import { OrganizationManagementPreferencesService } from "@bitwarden/common/admin-console/abstractions/organization-management-preferences/organization-management-preferences.service";
 import { PolicyApiServiceAbstraction } from "@bitwarden/common/admin-console/abstractions/policy/policy-api.service.abstraction";
 import { PolicyService } from "@bitwarden/common/admin-console/abstractions/policy/policy.service.abstraction";
 import {
@@ -36,7 +35,6 @@ import { OrganizationMetadataServiceAbstraction } from "@bitwarden/common/billin
 import { OrganizationBillingMetadataResponse } from "@bitwarden/common/billing/models/response/organization-billing-metadata.response";
 import { ConfigService } from "@bitwarden/common/platform/abstractions/config/config.service";
 import { EnvironmentService } from "@bitwarden/common/platform/abstractions/environment.service";
-import { FileDownloadService } from "@bitwarden/common/platform/abstractions/file-download/file-download.service";
 import { I18nService } from "@bitwarden/common/platform/abstractions/i18n.service";
 import { LogService } from "@bitwarden/common/platform/abstractions/log.service";
 import { ValidationService } from "@bitwarden/common/platform/abstractions/validation.service";
@@ -85,9 +83,6 @@ interface BulkMemberFlags {
 })
 export class vNextMembersComponent {
   protected i18nService = inject(I18nService);
-  protected organizationManagementPreferencesService = inject(
-    OrganizationManagementPreferencesService,
-  );
   protected validationService = inject(ValidationService);
   protected logService = inject(LogService);
   protected userNamePipe = inject(UserNamePipe);
@@ -108,7 +103,6 @@ export class vNextMembersComponent {
   private configService = inject(ConfigService);
   private environmentService = inject(EnvironmentService);
   private memberExportService = inject(MemberExportService);
-  private fileDownloadService = inject(FileDownloadService);
 
   private userId$: Observable<UserId> = this.accountService.activeAccount$.pipe(getUserId);
 
@@ -315,8 +309,14 @@ export class vNextMembersComponent {
     );
   }
 
-  private async handleInviteDialog(organization: Organization) {
+  async invite(organization: Organization) {
     const billingMetadata = await firstValueFrom(this.billingMetadata$);
+    const seatLimitResult = this.billingConstraint.checkSeatLimit(organization, billingMetadata);
+
+    if (await this.billingConstraint.seatLimitReached(seatLimitResult, organization)) {
+      return;
+    }
+
     const allUserEmails = this.dataSource().data?.map((user) => user.email) ?? [];
 
     const result = await this.memberDialogManager.openInviteDialog(
@@ -327,14 +327,6 @@ export class vNextMembersComponent {
 
     if (result === MemberDialogResult.Saved) {
       await this.load(organization);
-    }
-  }
-
-  async invite(organization: Organization) {
-    const billingMetadata = await firstValueFrom(this.billingMetadata$);
-    const seatLimitResult = this.billingConstraint.checkSeatLimit(organization, billingMetadata);
-    if (!(await this.billingConstraint.seatLimitReached(seatLimitResult, organization))) {
-      await this.handleInviteDialog(organization);
       this.organizationMetadataService.refreshMetadataCache();
     }
   }
@@ -561,7 +553,7 @@ export class vNextMembersComponent {
     return result;
   }
 
-  protected exportMembers() {
+  exportMembers = () => {
     const result = this.memberExportService.getMemberExport(this.dataSource().data);
     if (result.success) {
       this.toastService.showToast({
@@ -574,5 +566,5 @@ export class vNextMembersComponent {
     if (result.error != null) {
       this.validationService.showError(result.error.message);
     }
-  }
+  };
 }
--- a/bitwarden_license/bit-web/src/app/admin-console/providers/manage/members.component.ts
+++ b/bitwarden_license/bit-web/src/app/admin-console/providers/manage/members.component.ts
@@ -42,7 +42,6 @@ import { MemberActionsService } from "@bitwarden/web-vault/app/admin-console/org
 
 import {
   AddEditMemberDialogComponent,
-  AddEditMemberDialogParams,
   AddEditMemberDialogResultType,
 } from "./dialogs/add-edit-member-dialog.component";
 import { BulkConfirmDialogComponent } from "./dialogs/bulk-confirm-dialog.component";
@@ -308,20 +307,11 @@ export class vNextMembersComponent {
   }
 
   async edit(providerId: ProviderId, user?: ProviderUser): Promise<void> {
-    const data: AddEditMemberDialogParams = {
-      providerId: providerId,
-    };
-
-    if (user != null) {
-      data.user = {
-        id: user.id,
-        name: this.userNamePipe.transform(user),
-        type: user.type,
-      };
-    }
-
     const dialogRef = AddEditMemberDialogComponent.open(this.dialogService, {
-      data,
+      data: {
+        providerId,
+        user,
+      },
     });
 
     const result = await lastValueFrom(dialogRef.closed);
--- a/apps/web/src/app/admin-console/organizations/members/deprecated_members.component.ts
+++ b/apps/web/src/app/admin-console/organizations/members/deprecated_members.component.ts
@@ -599,7 +599,7 @@ export class MembersComponent extends BaseMembersComponent<OrganizationUserView>
       .every((member) => member.managedByOrganization && validStatuses.includes(member.status));
   }
 
-  protected exportMembers() {
+  exportMembers = () => {
     const result = this.memberExportService.getMemberExport(this.dataSource.data);
     if (result.success) {
       this.toastService.showToast({
@@ -612,5 +612,5 @@ export class MembersComponent extends BaseMembersComponent<OrganizationUserView>
     if (result.error != null) {
       this.validationService.showError(result.error.message);
     }
-  }
+  };
 }
--- a/apps/web/src/app/admin-console/organizations/members/services/member-export/member-export.service.spec.ts
+++ b/apps/web/src/app/admin-console/organizations/members/services/member-export/member-export.service.spec.ts
@@ -6,7 +6,9 @@ import {
   OrganizationUserStatusType,
   OrganizationUserType,
 } from "@bitwarden/common/admin-console/enums";
+import { FileDownloadService } from "@bitwarden/common/platform/abstractions/file-download/file-download.service";
 import { I18nService } from "@bitwarden/common/platform/abstractions/i18n.service";
+import { LogService } from "@bitwarden/logging";
 
 import { OrganizationUserView } from "../../../core";
 import { UserStatusPipe } from "../../pipes";
@@ -16,9 +18,13 @@ import { MemberExportService } from "./member-export.service";
 describe("MemberExportService", () => {
   let service: MemberExportService;
   let i18nService: MockProxy<I18nService>;
+  let fileDownloadService: MockProxy<FileDownloadService>;
+  let logService: MockProxy<LogService>;
 
   beforeEach(() => {
     i18nService = mock<I18nService>();
+    fileDownloadService = mock<FileDownloadService>();
+    logService = mock<LogService>();
 
     // Setup common i18n translations
     i18nService.t.mockImplementation((key: string) => {
@@ -44,9 +50,12 @@ describe("MemberExportService", () => {
         custom: "Custom",
         // Boolean states
         enabled: "Enabled",
+        optionEnabled: "Enabled",
         disabled: "Disabled",
         enrolled: "Enrolled",
         notEnrolled: "Not Enrolled",
+        // Error messages
+        noMembersToExport: "No members to export",
       };
       return translations[key] || key;
     });
@@ -54,6 +63,8 @@ describe("MemberExportService", () => {
     TestBed.configureTestingModule({
       providers: [
         MemberExportService,
+        { provide: FileDownloadService, useValue: fileDownloadService },
+        { provide: LogService, useValue: logService },
         { provide: I18nService, useValue: i18nService },
         UserTypePipe,
         UserStatusPipe,
@@ -88,8 +99,18 @@ describe("MemberExportService", () => {
         } as OrganizationUserView,
       ];
 
-      const csvData = service.getMemberExport(members);
+      const result = service.getMemberExport(members);
 
+      expect(result.success).toBe(true);
+      expect(result.error).toBeUndefined();
+      expect(fileDownloadService.download).toHaveBeenCalledTimes(1);
+
+      const downloadCall = fileDownloadService.download.mock.calls[0][0];
+      expect(downloadCall.fileName).toContain("org-members");
+      expect(downloadCall.fileName).toContain(".csv");
+      expect(downloadCall.blobOptions).toEqual({ type: "text/plain" });
+
+      const csvData = downloadCall.blobData as string;
       expect(csvData).toContain("Email,Name,Status,Role,Two-step Login,Account Recovery");
       expect(csvData).toContain("user1@example.com");
       expect(csvData).toContain("User One");
@@ -114,8 +135,12 @@ describe("MemberExportService", () => {
         } as OrganizationUserView,
       ];
 
-      const csvData = service.getMemberExport(members);
+      const result = service.getMemberExport(members);
 
+      expect(result.success).toBe(true);
+      expect(fileDownloadService.download).toHaveBeenCalled();
+
+      const csvData = fileDownloadService.download.mock.calls[0][0].blobData as string;
       expect(csvData).toContain("user@example.com");
       // Empty name is represented as an empty field in CSV
       expect(csvData).toContain("user@example.com,,Confirmed");
@@ -135,17 +160,23 @@ describe("MemberExportService", () => {
         } as OrganizationUserView,
       ];
 
-      const csvData = service.getMemberExport(members);
+      const result = service.getMemberExport(members);
 
+      expect(result.success).toBe(true);
+      expect(fileDownloadService.download).toHaveBeenCalled();
+
+      const csvData = fileDownloadService.download.mock.calls[0][0].blobData as string;
       expect(csvData).toContain("user@example.com");
       expect(csvData).toBeDefined();
     });
 
     it("should handle empty members array", () => {
-      const csvData = service.getMemberExport([]);
+      const result = service.getMemberExport([]);
 
-      // When array is empty, papaparse returns an empty string
-      expect(csvData).toBe("");
+      expect(result.success).toBe(false);
+      expect(result.error).toBeDefined();
+      expect(result.error?.message).toBe("No members to export");
+      expect(fileDownloadService.download).not.toHaveBeenCalled();
     });
   });
 });
--- a/bitwarden_license/bit-web/src/app/admin-console/providers/manage/deprecated_members.component.ts
+++ b/bitwarden_license/bit-web/src/app/admin-console/providers/manage/deprecated_members.component.ts
@@ -296,15 +296,8 @@ export class MembersComponent extends BaseMembersComponent<ProviderUser> {
   edit = async (user: ProviderUser | null): Promise<void> => {
     const data: AddEditMemberDialogParams = {
       providerId: this.providerId,
+      user,
     };
-
-    if (user != null) {
-      data.user = {
-        id: user.id,
-        name: this.userNamePipe.transform(user),
-        type: user.type,
-      };
-    }
 
     const dialogRef = AddEditMemberDialogComponent.open(this.dialogService, {
       data,
--- a/bitwarden_license/bit-web/src/app/admin-console/providers/manage/dialogs/add-edit-member-dialog.component.ts
+++ b/bitwarden_license/bit-web/src/app/admin-console/providers/manage/dialogs/add-edit-member-dialog.component.ts
@@ -3,6 +3,7 @@
 import { Component, Inject } from "@angular/core";
 import { FormControl, FormGroup, Validators } from "@angular/forms";
 
+import { UserNamePipe } from "@bitwarden/angular/pipes/user-name.pipe";
 import { ApiService } from "@bitwarden/common/abstractions/api.service";
 import { ProviderUserType } from "@bitwarden/common/admin-console/enums";
 import { ProviderUserInviteRequest } from "@bitwarden/common/admin-console/models/request/provider/provider-user-invite.request";
@@ -15,14 +16,11 @@ import {
   DialogService,
   ToastService,
 } from "@bitwarden/components";
+import { ProviderUser } from "@bitwarden/web-vault/app/admin-console/common/people-table-data-source";
 
 export type AddEditMemberDialogParams = {
   providerId: string;
-  user?: {
-    id: string;
-    name: string;
-    type: ProviderUserType;
-  };
+  user?: ProviderUser;
 };
 
 // FIXME: update to use a const object instead of a typescript enum
@@ -59,6 +57,7 @@ export class AddEditMemberDialogComponent {
     private dialogService: DialogService,
     private i18nService: I18nService,
     private toastService: ToastService,
+    private userNamePipe: UserNamePipe,
   ) {
     this.editing = this.loading = this.dialogParams.user != null;
     if (this.editing) {
@@ -78,8 +77,10 @@ export class AddEditMemberDialogComponent {
       return;
     }
 
+    const userName = this.userNamePipe.transform(this.dialogParams.user);
+
     const confirmed = await this.dialogService.openSimpleDialog({
-      title: this.dialogParams.user.name,
+      title: userName,
       content: { key: "removeUserConfirmation" },
       type: "warning",
     });
@@ -96,7 +97,7 @@ export class AddEditMemberDialogComponent {
     this.toastService.showToast({
       variant: "success",
       title: null,
-      message: this.i18nService.t("removedUserId", this.dialogParams.user.name),
+      message: this.i18nService.t("removedUserId", userName),
     });
 
     this.dialogRef.close(AddEditMemberDialogResultType.Deleted);
@@ -118,13 +119,12 @@ export class AddEditMemberDialogComponent {
       await this.apiService.postProviderUserInvite(this.dialogParams.providerId, request);
     }
 
+    const userName = this.editing ? this.userNamePipe.transform(this.dialogParams.user) : undefined;
+
     this.toastService.showToast({
       variant: "success",
       title: null,
-      message: this.i18nService.t(
-        this.editing ? "editedUserId" : "invitedUsers",
-        this.dialogParams.user?.name,
-      ),
+      message: this.i18nService.t(this.editing ? "editedUserId" : "invitedUsers", userName),
     });
 
     this.dialogRef.close(AddEditMemberDialogResultType.Saved);


apps/web/src/app/admin-console/common/member-component-utils.ts
BTreston, line 1: I'm not really in love with the approach here in this file. Right now, the `TableDataSource` is a bit of a pain point, but doesn't seem well scoped to be a part of this refactor. Feedback is welcome.

apps/web/src/app/admin-console/organizations/members/members.component.ts
```diff
@@ -56,75 +70,81 @@ import {
...
-      organizationManagementPreferencesService,
-      toastService,
-    );
+  constructor() {
+    this.dataSource
+      .connect()
+      .pipe(takeUntilDestroyed())
+      .subscribe(() => {
+        this.changeDetectionRef.markForCheck();
+      });
```
BTreston: again, table datasource causing a bit of pain here too.

bitwarden_license/bit-web/src/app/admin-console/providers/manage/members.component.ts
BTreston, line 1: I will put a little more time into this component soon to bring it closer to how the organization members component works and consolidating some of the shared/duplicated logic. Consider this file in a mostly incomplete/transitory state.

apps/web/src/app/admin-console/organizations/members/components/bulk/bulk-confirm-dialog.component.ts
eliykat, line 1: Is there a reason for adding selectors for these components? I had previously removed them because they're not needed when opening the component as a dialog, and not having a selector defined makes it clearer how it's used.

apps/web/src/app/admin-console/common/people-table-data-source.ts
```diff
@@ -112,13 +120,28 @@ export abstract class PeopleTableDataSource<T extends UserViewTypes> extends Tab
...
 
   getCheckedUsers() {
     return this.data.filter((u) => (u as any).checked);
   }
 
+  private checkedUsersUpdated$ = new Subject<void>();
+
+  usersUpdated(): Observable<void> {
+    return this.checkedUsersUpdated$.asObservable();
+  }
```
eliykat, line 143: This leads to this pattern in consumers:

```ts
  return dataSource.usersUpdated().pipe(
    map(() => {
      const checkedUsers = dataSource.getCheckedUsers();
      // do something with checkedUsers
```

But that is a mix of reactive and imperative. A better pattern would be to emit the checkedUsers directly:

```ts
  return dataSource.checkedUsers$.pipe(
    map((checkedUsers) => {
      // do something with checkedUsers
```

Also, this will only be used in an Angular context - is it worth using Signals? (open question)

apps/web/src/app/admin-console/common/people-table-data-source.ts
```diff
@@ -213,3 +241,130 @@ export abstract class PeopleTableDataSource<T extends UserViewTypes> extends Tab
...
+export interface BulkMemberFlags extends BulkFlags {
+  showBulkRestoreUsers: boolean;
+  showBulkRevokeUsers: boolean;
+  showBulkRemoveUsers: boolean;
+  showBulkDeleteUsers: boolean;
+}
+
+/**
+ * Helper function to determine if the confirm users option should be shown
+ * @params dataSource Either a ProivdersTableDataSource or a MembersTableDataSource
```
eliykat:
```suggestion
 * @params dataSource Either a ProvidersTableDataSource or a MembersTableDataSource
```

apps/web/src/app/admin-console/common/people-table-data-source.ts
```diff
@@ -213,3 +241,130 @@ export abstract class PeopleTableDataSource<T extends UserViewTypes> extends Tab
...
+
+export interface BulkMemberFlags extends BulkFlags {
+  showBulkRestoreUsers: boolean;
+  showBulkRevokeUsers: boolean;
+  showBulkRemoveUsers: boolean;
+  showBulkDeleteUsers: boolean;
+}
+
+/**
+ * Helper function to determine if the confirm users option should be shown
```
eliykat: To be more specific, I think this is a callout / banner, not the menu option.

```suggestion
 * Helper function to determine if the confirm users callout should be shown
```

apps/web/src/app/admin-console/common/people-table-data-source.ts
```diff
@@ -213,3 +241,130 @@ export abstract class PeopleTableDataSource<T extends UserViewTypes> extends Tab
...
+ * Configures bulk operation flags for organization members based on checked user states.
+ * Returns an Observable that emits updated flags whenever the checked users change.
+ *
+ * @param dataSource The MembersTableDataSource to observe
+ * @returns Observable that emits BulkMemberFlags whenever checked users are updated
+ */
+export function configureMemberFlags(
+  dataSource: MembersTableDataSource,
+): Observable<BulkMemberFlags> {
+  return dataSource.usersUpdated().pipe(
```
eliykat: The logic in this function could be more concisely expressed with the declarative array functions rather than this explicit control flow:

```ts
      const result = {
        showConfirmUsers: showConfirm(dataSource),
        showBulkConfirmUsers: checkedUsers.some(u => u.status == OrganizationUserStatusType.Accepted),
        showBulkReinviteUsers: checkedUsers.some(u => u.status == OrganizationUserStatusType.Invited),
       // etc
```

That's also more readable imo: "show the confirm option if there are any users in the accepted state", rather than reading the whole function to understand the default value and then the override.

This function could also be simplified by accepting the `OrganizationUserView` directly rather than the whole dataSource object.

apps/web/src/app/admin-console/common/people-table-data-source.ts
```diff
@@ -213,3 +241,130 @@ export abstract class PeopleTableDataSource<T extends UserViewTypes> extends Tab
...
+ * Configures bulk operation flags for provider members based on checked user states.
+ * Returns an Observable that emits updated flags whenever the checked users change.
+ *
+ * @param dataSource The ProvidersTableDataSource to observe
+ * @returns Observable that emits BulkFlags whenever checked users are updated
+ */
+export function configureProviderMemberFlags(
+  dataSource: ProvidersTableDataSource,
+): Observable<BulkFlags> {
+  return dataSource.usersUpdated().pipe(
```
eliykat: Same feedback here.

I'm also not sure these functions should be in this file, given that they are not shared code, and they make the file quite long.

apps/web/src/app/admin-console/organizations/members/services/member-actions/member-actions.service.ts
```diff
@@ -274,4 +301,58 @@ export class MemberActionsService {
...
+   * Shared user confirmation workflow that handles the common logic for confirming users.
+   * This method orchestrates the public key retrieval, fingerprint display, and confirmation dialog.
+   *
+   * @param user - The user to confirm (must implement ConfirmableUser interface)
+   * @param userNamePipe - Pipe to transform user names for display
+   * @param orgManagementPrefs - Service providing organization management preferences
+   * @param confirmCallback - Async callback that performs the actual confirmation with the public key
+   * @returns Promise that resolves when confirmation workflow completes
+   */
+  async confirmUserWorkflow<T extends ConfirmableUser>(
```
eliykat: There are so many abstractions here that I think it's better to un-DRY this and just have separate methods for orgUsers and providerUsers. The shared method is adding complexity and not giving us much benefit.

apps/web/src/app/admin-console/organizations/members/members.component.ts
```diff
@@ -62,79 +72,81 @@ import {
...
-class MembersTableDataSource extends PeopleTableDataSource<OrganizationUserView> {
-  protected statusType = OrganizationUserStatusType;
-}
-
-// FIXME(https://bitwarden.atlassian.net/browse/CL-764): Migrate to OnPush
-// eslint-disable-next-line @angular-eslint/prefer-on-push-component-change-detection
 @Component({
   templateUrl: "members.component.html",
   standalone: false,
+  changeDetection: ChangeDetectionStrategy.OnPush,
```
eliykat: I am still learning about OnPush, but Oscar's presentation suggested that we should migrate leaf nodes first (i.e. child components at the bottom of the DOM hierarchy), otherwise we might experience unexpected behaviour. Do you have any concerns with that here?

apps/web/src/app/admin-console/organizations/members/members.component.ts
eliykat: Please check my understanding: `this.dataSource.connect()` emits when the table re-renders, and this is ensuring that change detection is triggered in the members component at that time as well?

apps/web/src/app/admin-console/organizations/members/members.component.ts
```diff
@@ -211,80 +223,64 @@ export class MembersComponent extends BaseMembersComponent<OrganizationUserView>
     this.billingMetadata$.pipe(take(1), takeUntilDestroyed()).subscribe();
   }
 
-  override async load(organization: Organization) {
-    await super.load(organization);
+  async load(organization: Organization) {
+    const response = await this.memberService.loadUsers(organization);
+    this.dataSource.data = Array.isArray(response) ? response : [];
```
eliykat: (non-blocking) Can we clean up this `Array.isArray` check? Can `loadUsers` return a more definite type?

apps/web/src/app/admin-console/organizations/members/members.component.ts
```diff
@@ -549,48 +503,32 @@ export class MembersComponent extends BaseMembersComponent<OrganizationUserView>
...
+      } else {
         throw new Error(result.error);
       }
-      this.toastService.showToast({
-        variant: "success",
-        message: this.i18nService.t("organizationUserDeleted", this.userNamePipe.transform(user)),
-      });
-      this.dataSource.removeUser(user);
     } catch (e) {
       this.validationService.showError(e);
```
eliykat: Is this needed? If the method is called from the template with the `bitAction` directive, it'll have automatic error handling. I think that would also remove the need to check for `result.success` - your code here would handle the success, and `bitAction` aborts the code flow and shows the error toast if there's a problem. This is my strong preference if we can use it.

If this is needed - you could call `validationService` in the `else` block rather than throwing the error and then catching it.

apps/web/src/app/admin-console/organizations/members/members.component.ts
```diff
@@ -211,80 +223,64 @@ export class MembersComponent extends BaseMembersComponent<OrganizationUserView>
...
+      const result = await this.memberActionsService.confirmUser(user, publicKey, organization);
+      await this.handleMemberActionResult(result, "hasBeenConfirmed", user, confirmUserSideEffect);
+    };
+
+    await this.memberActionsService.confirmUserWorkflow(
+      user,
+      this.userNamePipe,
+      this.organizationManagementPreferencesService,
+      confirmUser,
+    );
```
eliykat: If you do keep this "workflow" method (see my comment above), it should be abstracted behind a facade. It's low level, the component shouldn't interact with it; instead, the component would call a specific method for confirming organization users, and that would in turn call the workflow method.

apps/web/src/app/admin-console/organizations/members/members.component.ts
```diff
@@ -362,52 +358,25 @@ export class MembersComponent extends BaseMembersComponent<OrganizationUserView>
   }
 
   async bulkRemove(organization: Organization) {
-    if (this.actionPromise != null) {
-      return;
-    }
```
eliykat, line 374: This is to prevent the user from submitting multiple bulk actions, because:
- they are more likely to be long running
- iirc, the CL dropdown menu doesn't handle this, unlike forms

If we remove this, do we have any other protection?

apps/web/src/app/admin-console/organizations/members/members.component.html
```diff
@@ -1,4 +1,15 @@
...
+@if (!organization || !dataSource) {
+  <i
+    class="bwi bwi-spinner bwi-spin tw-text-muted"
+    title="{{ 'loading' | i18n }}"
+    aria-hidden="true"
+  ></i>
+  <span class="tw-sr-only">{{ "loading" | i18n }}</span>
+}
+
 @if (organization) {
```
eliykat: Does this also need to check for `dataSource` (the inverse of the loading condition)?

i.e. 
```suggestion
@if (organization && dataSource) {
```

(may need to be converted into a Signal)

apps/web/src/app/admin-console/organizations/members/members.component.html
```diff
@@ -12,173 +23,167 @@
...
-      </bit-toggle>
+    @if (showUserManagementControls()) {
+      <bit-toggle-group
+        [selected]="statusToggle | async"
+        (selectedChange)="statusToggle.next($event)"
+        [attr.aria-label]="'memberStatusFilter' | i18n"
+      >
+        <bit-toggle [value]="undefined">
+          {{ "all" | i18n }}
+          @if (dataSource.activeUserCount; as allCount) {
```
eliykat, line 43: The "counts" (e.g. `activeUserCount`) are not reactive - should they be signals?

apps/web/src/app/admin-console/organizations/members/members.component.html
```diff
@@ -12,173 +23,167 @@
...
+    }
   </div>
-  <ng-container *ngIf="!firstLoaded">
+  @if (!firstLoaded()) {
     <i
       class="bwi bwi-spinner bwi-spin tw-text-muted"
       title="{{ 'loading' | i18n }}"
       aria-hidden="true"
     ></i>
     <span class="tw-sr-only">{{ "loading" | i18n }}</span>
```
eliykat: Duplicated loading state.

apps/web/src/app/admin-console/organizations/members/members.component.html
```diff
@@ -190,10 +195,12 @@
...
-              <td bitCell (click)="dataSource.checkUser(u)" *ngIf="showUserManagementControls()">
-                <input type="checkbox" bitCheckbox [(ngModel)]="$any(u).checked" />
-              </td>
-              <ng-container *ngIf="showUserManagementControls(); else readOnlyUserInfo">
+              @if (showUserManagementControls()) {
+                <td bitCell (click)="dataSource.checkUser(u)">
+                  <input type="checkbox" bitCheckbox [(ngModel)]="$any(u).checked" />
+                </td>
+              }
+              @if (showUserManagementControls()) {
```
eliykat: Redundant second `if` statement.

apps/web/src/app/admin-console/organizations/members/members.component.html
```diff
@@ -304,52 +296,48 @@
...
                     aria-hidden="true"
                   ></i>
                   <span class="tw-sr-only">{{ "userUsingTwoStep" | i18n }}</span>
-                </ng-container>
+                }
                 @let resetPasswordPolicyEnabled = resetPasswordPolicyEnabled$ | async;
-                <ng-container
-                  *ngIf="showEnrolledStatus($any(u), organization, resetPasswordPolicyEnabled)"
-                >
+                @if (showEnrolledStatus($any(u), organization, resetPasswordPolicyEnabled)) {
```
eliykat: Is this `$any` still required? (Same for the few other uses in here.)

apps/web/src/app/admin-console/organizations/members/members.component.html
```diff
@@ -410,73 +393,56 @@
...
+                      </button>
+                    }
+                    @if (!u.managedByOrganization) {
+                      <button type="button" bitMenuItem (click)="remove(u, organization)">
+                        <span class="tw-text-danger">
+                          <i aria-hidden="true" class="bwi bwi-close"></i> {{ "remove" | i18n }}
+                        </span>
+                      </button>
+                    }
+                    @if (u.managedByOrganization) {
```
eliykat: Could be an `else` on the previous `if`.

bitwarden_license/bit-web/src/app/admin-console/providers/manage/members.component.ts
```diff
@@ -46,89 +62,93 @@ import {
...
+    this.provider$ = combineLatest([
+      this.providerId$,
+      this.accountService.activeAccount$.pipe(getUserId),
+    ]).pipe(
+      switchMap(([providerId, userId]) => this.providerService.get$(providerId, userId)),
+      tap(async (provider) => {
+        if (!provider || !provider.canManageUsers) {
+          return await this.router.navigate(["../"], { relativeTo: this.activatedRoute });
+        }
+      }),
```
eliykat: This is handled by the `providerPermissionsGuard` in the routing and can be removed here.

bitwarden_license/bit-web/src/app/admin-console/providers/manage/members.component.ts
```diff
@@ -137,29 +157,27 @@ export class MembersComponent extends BaseMembersComponent<ProviderUser> {
...
   }
 
-  async bulkConfirm(): Promise<void> {
-    if (this.actionPromise != null) {
-      return;
-    }
+  async load() {
+    const providerId = await firstValueFrom(this.providerId$);
+    const response = await this.apiService.getProviderUsers(providerId);
+    this.dataSource.data = response.data != null && response.data.length > 0 ? response.data : [];
```
eliykat: Similar comment here as above - do we need this? `ListResponse.data` doesn't look like it'll ever be null.

bitwarden_license/bit-web/src/app/admin-console/providers/manage/members.component.ts
```diff
@@ -242,61 +260,136 @@ export class MembersComponent extends BaseMembersComponent<ProviderUser> {
...
+      }
+    };
+
+    await this.memberActionsService.confirmUserWorkflow(
+      user,
+      this.userNamePipe,
+      this.organizationManagementPreferencesService,
+      confirmUser,
+    );
+  }
```
eliykat: To my previous comment about un-DRYing this workflow method: `confirm`, `confirmUserInternal` and `confirmUserWorkflow` is difficult to follow.

bitwarden_license/bit-web/src/app/admin-console/providers/manage/members.component.ts
eliykat, line 1: This generally looks good but I haven't gone as in-depth as the org members component. It probably needs a look over for any similar issues after you've reviewed my feedback on the other component.

eliykat (changes requested):
Great work! Thank you for tackling this.

apps/web/src/app/admin-console/common/people-table-data-source.ts
BTreston: I can tweak it to fit a reactive pattern more, but since it's an abstract class it won't be direct access to the member as suggested above (since it needs to be private or protected). We could use signals, but in my opinion this is laying a little bit of groundwork for refactoring the table data source to be reactive, so observables probably fit better with that in mind.

apps/web/src/app/admin-console/organizations/members/components/bulk/bulk-confirm-dialog.component.ts
BTreston: It makes debugging easier when using angular dev-tools and traversing the component hierarchy, otherwise it just appears as `ng-component`. Also, there were a lot of warnings in the console related to these not having selectors and some of these did have selectors while others did not, so I thought I'd clean up that noise and inconsistency. 
That said, this was done before I learned of angular's intention to implement selector-less components, however that's probably still a long ways out.

apps/web/src/app/admin-console/organizations/members/members.component.html
BTreston: I would like them to be reactive, but they seem out of scope for the members components. Right now the methods that update the datasource call the setter that updates the counts imperatively so they do react to the changes to the model currently. But they are not "reactive"

apps/web/src/app/admin-console/organizations/members/members.component.ts
BTreston: I'm not too concerned, but right now it doesn't have much visible performance impact and we also have to do that subscription weirdness with the table data source. I'll swap it back so we can remove that, then we can further investigate the table datasource as a future effort so we don't have to have that manual change detection and instead rely on proper reactivity.

apps/web/src/app/admin-console/organizations/members/members.component.ts
BTreston: Yes, that is correct. Re. my previous comment: I will remove this and revert back to default change detection until the table data source is properly reactive so we don't need this.

apps/web/src/app/admin-console/organizations/members/members.component.ts
BTreston: We actually just don't need this. It already returns a `Promise<OrganizationuserView[]>` and coalesces to `[]` in its return statement.

apps/web/src/app/admin-console/organizations/members/members.component.ts
BTreston: When I tested this on main, I was able to submit multiple bulk requests so I assumed this was just not working the way we intended and removed it.

apps/web/src/app/admin-console/organizations/members/members.component.ts
BTreston: I don't think we can since these are `bitMenuItems` and not `bitButtons`. As far as I can tell, they're not interoperable. I'll play around with it and simplify the error handling some more.

apps/web/src/app/admin-console/organizations/members/members.component.ts
BTreston: Thoughts on something like this:
```Typescript
async confirm(user: OrganizationUserView, organization: Organization) {
    const sideEffect = () => {...}

    // This would replace the whole workflow stuff. It basically opens the confirm dialog and returns out the public
    // key if you hit confirm.
    const publicKeyResult = await this.memberActionsService.getPublicKeyForConfirm(
      user,
      this.userNamePipe,
      this.organizationManagementPreferencesService,
    );

    const result = await this.memberActionsService.confirmUser(user, publicKeyResult, organization);
    await this.handleMemberActionResult(result, "hasBeenConfirmed", user, confirmUserSideEffect);
}
```
Where `getPublicKeyForConfirm()` calls `getUserPublicKey()`. This gets rid of the whole callback abstraction layer from the original design to get what we really need here. This could also be moved entirely into the confirm dialog (potentially)

apps/web/src/app/admin-console/organizations/members/members.component.html
BTreston: only the one for `checkAllFilteredUsers()` is required

bitwarden_license/bit-web/src/app/admin-console/providers/manage/members.component.ts
BTreston: building on the previous comment:
```typescript
async confirm(user: ProviderUser, providerId: ProviderId) {
    try {
      const publicKeyResult = await this.memberActionsService.getPublicKeyForConfirm(
        user,
        this.userNamePipe,
        this.organizationManagementPreferencesService,
      );
      assertNonNullish(publicKeyResult, "Public key not found");

      const providerKey = await firstValueFrom(
          getUserId,
          switchMap((userId) => this.keyService.providerKeys$(userId)),
          map((providerKeys) => providerKeys?.[providerId as ProviderId] ?? null),
        ),
      );
      assertNonNullish(providerKey, "Provider key not found");

      const key = await this.encryptService.encapsulateKeyUnsigned(providerKey, publicKeyResult);
      assertNonNullish(key.encryptedString, "No key was provided");

      const request = new ProviderUserConfirmRequest(key.encryptedString);
      const result = await this.apiService.postProviderUserConfirm(providerId, user.id, request);

      if (result.success) {
        user.status = this.userStatusType.Confirmed;
        this.dataSource.replaceUser(user);

        this.toastService.showToast({
          variant: "success",
          message: this.i18nService.t("hasBeenConfirmed", this.userNamePipe.transform(user)),
        });
      } else {
        throw new Error(result.error);
      }
    } catch (e) {
      this.validationService.showError(e);
      throw e;
    }

  }
```
`confirmUserInternal()` is removed entirely.

apps/web/src/app/admin-console/organizations/members/components/bulk/bulk-confirm-dialog.component.ts
eliykat: Debugging is a good reason - I'm fine with that.

apps/web/src/app/admin-console/organizations/members/members.component.ts
eliykat: I think this is the right call for now.

apps/web/src/app/admin-console/common/people-table-data-source.ts
eliykat: This change looks good to me and everything else you said makes sense 👍

apps/web/src/app/admin-console/organizations/members/members.component.ts
eliykat: Can you please follow up with the UIF team about this? We have lots of actions triggered via overlay menus, so having this handled for us would be a real benefit.

apps/web/src/app/admin-console/organizations/members/members.component.ts
eliykat: If it's already working like this today then I guess it's not critical - it would still be nice protection to have though. If you don't want to tackle it here, maybe create a bug ticket for this functionality so we can follow it up. Ideally the Component Library would handle this as well.

apps/web/src/app/admin-console/organizations/members/members.component.html
eliykat: My concern here was for OnPush compatibility, but if you're winding that back anyway then this is OK.

bitwarden_license/bit-web/src/app/admin-console/providers/manage/members.component.ts
eliykat: I like your suggestion, it's more straightforward, although still too much logic for the component imo. Here's my attempt - I think it's closer to your [earlier comment](https://github.com/bitwarden/clients/pull/17863#discussion_r2662853222).

```ts
async confirm(user: ProviderUser, providerId: ProviderId) {
    try {
      // opens a dialog to get the key, check the fingerprint and return it out
      // alternatively you could get the key and pass it in (presentational only)
      const publicKeyResult = await this.memberActionsService.getPublicKeyForConfirm(
        user,
        this.userNamePipe,
        this.organizationManagementPreferencesService,
      );

      // you also have to handle the user hitting cancel on the dialog?
      assertNonNullish(publicKeyResult, "Public key not found");

      // This service method stitches together lower level methods - basically a lift & shift of your previous code here:
      // 1. gets the provider key
      // 2. encapsulates it with the public key
      // 3. posts it to the server
      // 4. returns a result to the component layer
      const result = await memberActionService.confirm(whateverYouNeedHere);


      // Then the component layer only handles the result - although if you're going to try/catch it anyway
      // maybe you just rely on that.
      // you can also use the Handle pattern here if you want
      if (result.success) {
        user.status = this.userStatusType.Confirmed;
        this.dataSource.replaceUser(user);

        this.toastService.showToast({
          variant: "success",
          message: this.i18nService.t("hasBeenConfirmed", this.userNamePipe.transform(user)),
        });
      } else {
        throw new Error(result.error);
      }
    } catch (e) {
      this.validationService.showError(e);
      throw e;
    }
  }
```

eliykat (commented):
Resolved a few comments and responded inline.

bitwarden_license/bit-web/src/app/admin-console/providers/manage/members.component.ts
eliykat: One thing I'd like to maintain is a separation between Angular and non-Angular code. The actions required to confirm a user are not specific to Angular (get key, encapsulate it, send it to the server), and could be reused by the CLI for example. Displaying the dialog is Angular code very specifically. So it's possible they don't live in the same place as in my snippet above. (I am not sure of the scope of MemberActionService).

apps/web/src/app/admin-console/organizations/members/members.component.ts
BTreston: This is def. not in scope for Rui's bulk stuff for reinvite, but I think when we get to the milestone where the UI state is re-enabled by a push notification by the sever, it might be something we can do here as well for the other actions. I'll try and cook something up for now though.

BTreston: Support for this will be added by UIF (https://bitwarden.slack.com/archives/C033DHFNCV7/p1767802128518119).

eliykat: Thank you!

apps/web/src/app/admin-console/organizations/members/members.component.ts
```diff
@@ -218,80 +233,65 @@ export class MembersComponent extends BaseMembersComponent<OrganizationUserView>
...
-    user: OrganizationUserView,
-    publicKey: Uint8Array,
-    organization: Organization,
-  ): Promise<MemberActionResult> {
-    return await this.memberActionsService.confirmUser(user, publicKey, organization);
+    const publicKeyResult = await this.memberActionsService.getPublicKeyForConfirm(
+      user,
+      this.userNamePipe,
+      this.organizationManagementPreferencesService,
+    );
```
eliykat: At a glance, `this.userNamePipe` and `this.organizationManagementPreferencesService` is the same for both calling locations - do they need to be parameters?

apps/web/src/app/admin-console/organizations/members/members.component.ts
```diff
@@ -218,80 +233,65 @@ export class MembersComponent extends BaseMembersComponent<OrganizationUserView>
...
+
+    if (publicKeyResult == null) {
+      throw new Error("Public key not found");
+    }
```
eliykat, line 264: What if the dialog is closed or cancelled? Is that handled here?

apps/web/src/app/admin-console/organizations/members/services/member-actions/member-actions.service.ts
eliykat, line 1: What's the difference between `MemberActionsService` and `OrganizationUserService`?
Please add jsdoc explaining their respective responsibilities.

I'm also not sure we want to mix org member + provider member logic in the same class - because they're separate (if similar) domains, and because providers are under a different license. Is there an advantage to having them both in here together? (It's OK for them to share common code like the get public key method, I'm more looking at provider specific code like deleting or reinviting provider users.)

apps/web/src/app/admin-console/organizations/members/members.component.ts
BTreston: Nope, not anymore.

apps/web/src/app/admin-console/organizations/members/services/member-actions/member-actions.service.ts
BTreston: I was debating if the provider methods should just go in their own service. I will do that since you've brought it up. The only real reason I kept them here is so they can share the same `isProcessing` signal. In the near future, I want to consolidate `OrganizationUserService` into the `MemberActionsService` and then break that service up to prevent it growing too large. I feel like it makes sense to break the `MemberActionsService` into something like `ConfirmService`, `InviteService` and `RemoveService`.

edit: I forgot to mention that before the holidays I spoke with platform about the intended end goal here (Nx libs) and remembered that they said you cannot mix licensed code with non licensed code for nx libs. So `ProviderMembers` and `Members` will each eventually have their own NX lib (with the providers lib living in `bitwarden_license`

apps/web/src/app/admin-console/organizations/members/members.component.ts
BTreston: if its closed or cancelled, `publicKeyResult` will be `undefined` which we handle here. I'm going to change this to a warn and early return, as an error doesn't really seem fully appropriate here. I'll also improve the error handling in `handleMemberActionResult`

apps/web/src/app/admin-console/organizations/members/members.component.ts
```diff
@@ -218,80 +233,62 @@ export class MembersComponent extends BaseMembersComponent<OrganizationUserView>
...
-    publicKey: Uint8Array,
-    organization: Organization,
-  ): Promise<MemberActionResult> {
-    return await this.memberActionsService.confirmUser(user, publicKey, organization);
+    const publicKeyResult = await this.memberActionsService.getPublicKeyForConfirm(user);
+
+    if (publicKeyResult == null) {
+      this.logService.warning("Public key not found");
+      return;
+    }
```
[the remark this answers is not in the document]
BTreston: If you press cancel, you don't need to surface an error. This is the current existing functionality, which is fine.

apps/web/src/app/admin-console/organizations/members/members.component.ts
```diff
@@ -556,35 +553,18 @@ export class vNextMembersComponent {
...
         message: this.i18nService.t("dataExportSuccess"),
       });
-    } catch (e) {
-      this.validationService.showError(e);
-      this.logService.error(`Failed to export members: ${e}`);
+    }
+
+    if (result.error != null) {
+      this.validationService.showError(result.error.message);
     }
```
eliykat, line 568: I'm not sure about expanding this error handling everywhere. Most actions, including this one, are triggered through the `bitAction` directive which handles errors for us. Otherwise it is a lot of boilerplate to achieve the same result.

It makes sense for the `MemberActionsService` because we can't use `bitAction` on dropdown menus, but ideally once that is implemented we would rely on the Component Library there as well.

Not a blocker right now, just wanted to call it out for your future work in this area.

apps/web/src/app/admin-console/organizations/members/members.component.ts
eliykat: I think it needs to be handled as separate outcomes.
- if the user exits/cancels out of the dialog, nothing is wrong and nothing needs to be logged
- if the user continues through the dialog but no public key is returned, something has gone wrong and you probably do want to throw.

This is low priority so can be done in a later PR when convenient if you just want to get this one merged.

PR #17874, opened 2025-12-08
fix padding when nested. remove ng style and class

BryanCunningham:
## 🎟️ Tracking

No ticket. Quick change

## 📔 Objective

- Fix indentation when nav items are not links
- cleanup `ngClass` and `ngStyle` usage

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/.github/workflows/test-browser-interactions.yml
+++ b/.github/workflows/test-browser-interactions.yml
@@ -75,7 +75,7 @@ jobs:
 
     - name: Trigger test-all workflow in browser-interactions-testing
       if: steps.changed-files.outputs.monitored == 'true'
-      uses: peter-evans/repository-dispatch@5fc4efd1a4797ddb68ffd0714a238564e4cc0e6f # v4.0.0
+      uses: peter-evans/repository-dispatch@28959ce8df70de7be546dd1250a005dd32156697 # v4.0.1
       with:
         token: ${{ steps.app-token.outputs.token }}
         repository: "bitwarden/browser-interactions-testing"
--- a/apps/browser/src/_locales/en/messages.json
+++ b/apps/browser/src/_locales/en/messages.json
@@ -1475,6 +1475,9 @@
   "selectFile": {
     "message": "Select a file"
   },
+  "itemsTransferred": {
+    "message": "Items transferred"
+  },
   "maxFileSize": {
     "message": "Maximum file size is 500 MB."
   },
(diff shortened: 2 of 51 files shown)


libs/components/src/navigation/nav-item.component.html
```diff
@@ -2,16 +2,12 @@
...
-      [ngClass]="[
-        showActiveStyles
-          ? 'tw-bg-background-alt4'
-          : 'tw-bg-background-alt3 hover:tw-bg-hover-contrast',
-        fvwStyles$ | async,
-      ]"
+      [class.tw-bg-background-alt4]="showActiveStyles"
+      [class.tw-bg-background-alt3]="!showActiveStyles"
+      [class.hover:tw-bg-hover-contrast]="!showActiveStyles"
+      [ngClass]="fvwStyles$ | async"
```
Hinton: Any reason this still uses ngClass?
```suggestion
      [class]="fvwStyles$ | async"
```

vleague2:
Actually, should we add a story for whatever bug this is fixing?

libs/components/src/navigation/nav-item.component.html
BryanCunningham: @Hinton Claude didn't fix that one when prompted. I wasn't sure if this syntax could be supported with regular class binding or not. I can give it a try.

BryanCunningham:
> Actually, should we add a story for whatever bug this is fixing?

@vleague2 Good call. Technically, we have one(kind of). But, the groups are collapsed so this bug does not get caught by screen shots. I can add expanded groups to the story to cover this though.

libs/components/src/navigation/nav-item.component.html
BryanCunningham: Per Claude when asked why this was not changed:
> This works if `fvwStyles$ `emits a string of space-separated classes. However, note that `[class]` will replace all classes, so you'd need to include all the other conditional classes in the observable as well, which is not ideal. I recommend keeping it as-is since [`ngClass]` is the appropriate directive for applying dynamic classes from an observable. The other `[class.*]` bindings we added are for static conditional logic, but this one needs the flexibility of `ngClass` with the observable pattern.

vleague2: Claude continues to be wrong about `[class]` replacing all classes..... [From docs](https://v20.angular.dev/guide/templates/binding#css-class-and-style-property-bindings), 
> When using static CSS classes, directly binding class, and binding specific classes, Angular intelligently combines all of the classes in the rendered result.

BryanCunningham: Interesting. Seems to work as expected with normal class binding. Updated [here](https://github.com/bitwarden/clients/pull/17874/commits/18cf0df99b52109af79bb2f83468db20a8ba500a)
Thanks @vleague2 !

PR #17880, opened 2025-12-09
Electron 39 window not shown workaround

dani-garcia:
## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/apps/desktop/src/utils.ts
+++ b/apps/desktop/src/utils.ts
@@ -35,6 +35,10 @@ export function isSnapStore() {
   return isLinux() && process.env.SNAP_USER_DATA != null;
 }
 
+export function isWayland() {
+  return isLinux() && (process.env.XDG_SESSION_TYPE === "wayland" || process.env.WAYLAND_DISPLAY);
+}
+
 export function isMac() {
   return process.platform === "darwin";
 }
--- a/apps/desktop/src/main/window.main.ts
+++ b/apps/desktop/src/main/window.main.ts
@@ -23,6 +23,7 @@ import {
   isMac,
   isMacAppStore,
   isSnapStore,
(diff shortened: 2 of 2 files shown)


apps/desktop/src/utils.ts
```diff
@@ -35,6 +35,10 @@ export function isSnapStore() {
   return isLinux() && process.env.SNAP_USER_DATA != null;
 }
 
+export function isWayland() {
+  return isLinux() && (process.env.XDG_SESSION_TYPE === "wayland" || process.env.WAYLAND_DISPLAY);
```
quexten, line 39: Note: I'm not sure if this is entirely accurate. F.e you can set `USE_X11` which will make electron use X11, despite wayland being available: https://github.com/bitwarden/clients/blob/c6576ceec8be222d76e011ddd6d7e79ae7a37cc4/apps/desktop/resources/linux-wrapper.sh#L15

dani-garcia: Hmm I believe those parameters are the default now: https://www.electronjs.org/blog/electron-38-0#removed-electron_ozone_platform_hint-environment-variable

So we might want to change that script to set `--ozone-platform=x11` instead, but good point

Sytten: For what it is worth, signal uses `WAYLAND_DISPLAY` (https://github.com/signalapp/Signal-Desktop/blob/1b03cc4b9b73841fa46a6e2a909c2627caf68d61/ts/util/os/osMain.node.ts#L47-L49)
And yes wayland is default now, only the `--ozone-platform=x11` flag will change it and `XDG_SESSION_TYPE=x11`

PR #17884, opened 2025-12-09
Convert list.scss to a CSS file

jholland-livefront:
## 🎟️ Tracking

[PM-27365](https://bitwarden.atlassian.net/browse/PM-27365)

## 📔 Objective

This change migrates `list.scss` to `list.css`.  I removed the SASS specific portions and expanded some of the selectors to remove SASS style nesting. 

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

[PM-27365]: https://bitwarden.atlassian.net/browse/PM-27365?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/browser/src/autofill/overlay/inline-menu/pages/list/list.css
dan-livefront, line 1: On https://bitwarden.atlassian.net/browse/PM-27070 there is a slack thread that pointed out the intention to use the tailwind files to set new color variables to emulate existing theme imports, I think we should do that here to avoid hard coding hex colors.

apps/browser/src/autofill/shared/styles/variables-export.css
Hinton: There is a `theme.css` in `libs/components/src/` that defines fonts which might be nice to use? And is probably a good place to share variables with the regular design system.

apps/browser/src/autofill/overlay/inline-menu/pages/list/list.css
Hinton: We support postcss and postcss-nested.

libs/components/src/tw-theme.css
jprusik, line 1: These variables are being consumed elsewhere (outside of inline-menu components), no?

We should avoid changing these values (@bitwarden/team-ui-foundation should own that as a separate concern, if it needs to happen) and use hardcoded constants instead of these CSS variables if our expected design for the content script components can't use the design system values (though I don't think we're using these variables in the content script components?).

jholland-livefront: I removed those changes so `tw-theme.css` is now unchanged.

Hinton:
FYI there is a `theme.css` in `libs/components/src/` that defines fonts which might be nice to use? And is probably a good place to share variables with the regular design system for colors. We also support postcss with postcss-nested, so nested statements should still work if you prefer that.

jholland-livefront:
> `libs/components/src/`

@Hinton I looked at that file and I am pushing an update to use `--font-sans` from that file for the `--autofill-font-family-sans-serif` variable.  However, `--autofill-font-family-source-code-pro` does not have a match in that file.  Should I add it?

Also, I am hesitant to add the colors to this file because the colors for the Autofill Overlay are slightly different those in `tw-theme.css` and I'm not comfortable adding the autofill colors to a shared library unless you recommend I do this, or you have feedback for color changes the overlay should get to match `tw-theme.css` values.

Hinton:
@jholland-livefront to my knowledge we should not be using source code pro anywhere. I believe some old figmas referenced it but they have since been update.

jprusik:
After discussing with the team; we're going to take a different approach on this one. Closing this PR in the meantime.

PR #17892, opened 2025-12-09
Add creationDate of account to AccountInfo

trmartin4:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-29545

## 📔 Objective

We are currently retrieving the account creation date on the account profile, which is retrieved from the `/sync` endpoint.  However, we are not currently setting that value into any persistent state so that it can be used by downstream consumers.

This PR pushes us in the direction of a consolidated `AccountInfo` that contains all account-level information, not just the email, name, and whether the email is verified.

This is slightly complicated by the fact that the `AccountInfo` is populated in a disjointed way today - from both the contents of the access token, and also from the sync response.  Both of these are not where we want to see the data come from - we would prefer this to be in a consolidated API call or series of calls to get the user's profile data, where we could manage it in state as a single entity.

While we cannot yet get the data in a consolidated way, this PR drives us toward an object structure that exposes the data to consumers in that consolidated API.

### Core changes
- `AccountService` - https://github.com/bitwarden/clients/pull/17892/files?owned-by%5B%5D=trmartin4#diff-d02f0b858754d3d01c0e8facf03fee14b21423b5a61ccd781a3cb0b17549e74a - Added new property for `creationDate`.
- `DefaultSyncService` - https://github.com/bitwarden/clients/pull/17892/files?owned-by%5B%5D=trmartin4#diff-a54e81c24d9ec04cb7b32d220e0e203707f563b59c945a3fe9a7ee209eec3cf2 - Set the `creationDate` after sync, based on data that was already in the response.
- `LoginStrategy` - https://github.com/bitwarden/clients/pull/17892/files?owned-by%5B%5D=trmartin4#diff-a54e81c24d9ec04cb7b32d220e0e203707f563b59c945a3fe9a7ee209eec3cf2 - Set the `creationDate` to `undefined` because it is not returned on the access token, so we won't set it at that point.

### 🧪 Test changes
All non-Auth and Platform changes are updates to test files.  This was necessary because `AccountInfo` objects were used to scaffold a lot of tests, and there was no helper function there to abstract the property initialization.  This meant that every non-optional property added to `AccountInfo` would require updating all tests.

We decided that this was the point at which we should introduce a `mockAccountInfoWith()` function on the `FakeAccountService` to abstract the initialization, allowing this and future updates to `AccountInfo` to not require broad swathes of test changes.  **The only changes to other teams' tests were to replace the hard-coded initialization of `AccountInfo` with this new method.**

We opted to put this in one PR because the changes for all the teams were tests only.  If any team strongly prefers a separate PR, we can instead make `creationDate` optional and do a gradual migration.

## 📸 Screenshots

https://github.com/user-attachments/assets/6b92e58a-0170-442f-880d-20d4e5602748

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


libs/common/src/auth/abstractions/account.service.ts
```diff
@@ -2,14 +2,11 @@ import { Observable } from "rxjs";
 
 import { UserId } from "../../types/guid";
 
-/**
- * Holds information about an account for use in the AccountService
- * if more information is added, be sure to update the equality method.
- */
```
trmartin4, line 8: @MGibson1 can you confirm that it's OK to remove this warning?  It looks like it may be obsolete now that we're looping through all keys in the equality logic?

libs/auth/src/common/login-strategies/login.strategy.ts
```diff
@@ -185,6 +185,7 @@ export abstract class LoginStrategy {
       name: accountInformation.name,
       email: accountInformation.email ?? "",
       emailVerified: accountInformation.email_verified ?? false,
+      creationDate: undefined, // We don't get a creation date in the token. See https://bitwarden.atlassian.net/browse/PM-29551 for consolidation plans.
```
trmartin4, line 192: I'm concerned about the fact that between the token response handling and the first sync that this will not be defined, but since the `LoginSuccessHandlerService` will run and execute a sync before downstream consumers of `AccountInfo` will need the data, perhaps this is OK?

This exposes a mismatch between the data returned in our APIs (login, sync) and the data structures we want to expose to consumers for account state.  We have a vision for making these match.

JaredSnider-Bitwarden: Agreed. This is ok for now for the near term.

willmartian:
Approving UIF changes. (And I will look into using jest deps in Storybook.)

apps/web/src/app/billing/individual/upgrade/upgrade-payment/services/upgrade-payment.service.spec.ts
```diff
@@ -46,11 +47,12 @@ describe("UpgradePaymentService", () => {
 
   let sut: UpgradePaymentService;
 
-  const mockAccount = {
+  const mockAccount: Account = {
```
cyprain-okeke, line 50: The file has significant duplication of the same mock account object, Can  we consider extracting the repeated mockAccount to a shared beforeEach() or test utility

trmartin4: Yes!  For this effort, I didn't want to require teams to review more changes to the mock account object as part of this large PR, so I kept it exactly the same to ensure minimal changes, because there was not a 100% universal initialization pattern.  Would you like me to record something for Billing to look at this, or will you take care of that?

Banrion (approved):
DIRT file looks good

JaredSnider-Bitwarden (approved):
Auth files changes look mostly great. Claude has a few good suggestions around using an actual date + extra validation in the set methods, but those can be done later on when we don't have as many reviewer dependencies.

dereknance (approved):
Platform changes look good!

libs/common/src/auth/abstractions/account.service.ts
```diff
@@ -2,14 +2,11 @@ import { Observable } from "rxjs";
...
 export type AccountInfo = {
   email: string;
   emailVerified: boolean;
   name: string | undefined;
+  creationDate: string | undefined;
```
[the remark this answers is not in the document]
trmartin4: Addressed in https://github.com/bitwarden/clients/pull/17928.

libs/common/src/auth/abstractions/account.service.ts
```diff
@@ -2,14 +2,11 @@ import { Observable } from "rxjs";
 
 import { UserId } from "../../types/guid";
 
-/**
- * Holds information about an account for use in the AccountService
- * if more information is added, be sure to update the equality method.
- */
 export type AccountInfo = {
```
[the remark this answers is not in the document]
trmartin4: Will add this in https://github.com/bitwarden/clients/pull/17928.

trmartin4:
@JaredSnider-Bitwarden will follow up with https://github.com/bitwarden/clients/pull/17928 to move `creationDate` to a `Date` property.

PR #17911, opened 2025-12-10
[PM-24015] Handle Send form empty password field properly

### LINKED ISSUE #15718: Creating a text send and
### Steps To Reproduce

1. Create a new Bitwarden Send (File or Text)
2. Add Name
3. Add File/Text
4. Add a password
5. Remove the password

### Expected Result

There is no password created for the Send

### Actual Result

There is a password created for the Send. The user must add a character and then remove the character before submitting the password to view the Send.

### Screenshots or Videos

https://github.com/user-attachments/assets/1cd77a99-39d4-4a11-a23d-1f507e3fa514

### Additional Context

I was also able to reproduce this behavior in the web interface.

### Operating System

macOS

### Operating System Version

15.5

### Web Browser

Microsoft Edge, Firefox, Chrome

### Browser Version

Version 138.0.3351.95 (Official build) (arm64), 140.0.4 (aarch64), Version 138.0.7204.158 (Official Build) (arm64)

### Environment Versions

Version: 2025.6.1

SDK: 'main (1fa0f7a)'

Server version: 2025.7.0

### Issue Tracking Info

mcamirault:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-24015

Closes #15718

## 📔 Objective

Because we initialized the Send form with `null` in the password field but empty text input fields are always `""`, typing then deleting a password from the form would set the Send's password to `""`. This PR makes it so that Send passwords that are entirely whitespace or line terminators will be surfaced as `null` instead, ensuring they are handled properly.

Also added a test on the component to confirm the behavior and removed some handling from a higher-level component that may have been a previous partial fix to this issue.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/libs/tools/send/send-ui/src/send-form/components/options/send-options.component.ts
+++ b/libs/tools/send/send-ui/src/send-form/components/options/send-options.component.ts
@@ -4,7 +4,7 @@ import { CommonModule } from "@angular/common";
 import { Component, Input, OnInit } from "@angular/core";
 import { takeUntilDestroyed } from "@angular/core/rxjs-interop";
 import { FormBuilder, ReactiveFormsModule } from "@angular/forms";
-import { BehaviorSubject, firstValueFrom, map, switchMap } from "rxjs";
+import { BehaviorSubject, firstValueFrom, map, switchMap, tap } from "rxjs";
 
 import { JslibModule } from "@bitwarden/angular/jslib.module";
 import { PolicyService } from "@bitwarden/common/admin-console/abstractions/policy/policy.service.abstraction";
@@ -12,6 +12,7 @@ import { PolicyType } from "@bitwarden/common/admin-console/enums";
 import { AccountService } from "@bitwarden/common/auth/abstractions/account.service";
 import { getUserId } from "@bitwarden/common/auth/services/account.service";
 import { I18nService } from "@bitwarden/common/platform/abstractions/i18n.service";
+import { Utils } from "@bitwarden/common/platform/misc/utils";
 import { pin } from "@bitwarden/common/tools/rx";
 import { SendView } from "@bitwarden/common/tools/send/models/view/send.view";
 import { SendApiService } from "@bitwarden/common/tools/send/services/send-api.service.abstraction";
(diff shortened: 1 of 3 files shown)


libs/tools/send/send-ui/src/send-form/components/options/send-options.component.ts
```diff
@@ -112,18 +113,27 @@ export class SendOptionsComponent implements OnInit {
...
-          maxAccessCount: value.maxAccessCount,
-          accessCount: value.accessCount,
-          password: value.password,
-          hideEmail: value.hideEmail,
-          notes: value.notes,
+    this.sendOptionsForm.valueChanges
+      .pipe(
+        tap((value) => {
+          if (Utils.isNullOrWhitespace(value.password)) {
+            value.password = null;
```
harr1424, line 120: ❓ should this be `undefined`?   Sometimes `null` vs `undefined` can mean a behavioral difference.

libs/tools/send/send-ui/src/send-form/components/send-form.component.ts
```diff
@@ -227,10 +226,6 @@ export class SendFormComponent implements AfterViewInit, OnInit, OnChanges, Send
       return;
     }
 
-    if (Utils.isNullOrWhitespace(this.updatedSendView.password)) {
-      this.updatedSendView.password = null;
```
harr1424, line 231: ❓ should this be `undefined`? Sometimes `null` vs `undefined` can mean a behavioral difference.

libs/tools/send/send-ui/src/send-form/components/options/send-options.component.ts
mcamirault: I believe there's only two files where the password value is explicitly checked:
- [The `encrypt` method of the SendService](https://github.com/bitwarden/clients/blob/main/libs/common/src/tools/send/services/send.service.ts#L87). This uses loose equality so `null` vs `undefined` doesn't matter
- [The `hasPassword` method of the SendOptionsComponent](https://github.com/bitwarden/clients/blob/main/libs/tools/send/send-ui/src/send-form/components/options/send-options.component.ts#L76). This one does specifically check for `null` with strict equality, and there's a couple other operations in the same file that set the password value to `null` as well

Based on these I think `null` is the correct value to use

harr1424: Nice, the below is exactly what I had in mind: 
```typescript
  get hasPassword(): boolean {
    return this.originalSendView && this.originalSendView.password !== null;
  }
```
Great investigation, it is helpful to keep this in mind as other teams may open PRs that use `undefined`.

harr1424 (approved):

PR #17123, opened 2025-10-30
[PM-23258] changing verbiage from import data to import items

itsadrago:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-23258

## 📔 Objective

We have some minor verbiage inconsistency around import and export screens, navigation bars, and menu items. The changes are shown below, per client

**Desktop**

Export vault → Export

Import data → Import

Sync vault → Sync now

Import vault → Import

**Browser**

Sync vault now → Sync now

Export vault → Export

Import data → Import

Import items → Import

**Web Vault**

Import data → Import

Export vault → Export

Confirm format -> Export

**Libs**

Import data → Import

Export vault → Export

## ⏰ Reminders before review

- Contributor guidelines followed
- All formatters and local linters executed and passed
- Written new unit and / or integration tests where applicable
- Protected functional changes with optionality (feature flags)
- Used internationalization (i18n) for all UI strings
- CI builds passed
- Communicated to DevOps any deployment requirements
- Updated any necessary documentation (Confluence, contributing docs) or informed the documentation team

## 🦮 Reviewer guidelines

- 👍 (`:+1:`) or similar for great changes
- 📝 (`:memo:`) or ℹ️ (`:information_source:`) for notes or general info
- ❓ (`:question:`) for questions
- 🤔 (`:thinking:`) or 💭 (`:thought_balloon:`) for more open inquiry that's not quite a confirmed issue and could potentially benefit from discussion
- 🎨 (`:art:`) for suggestions / improvements
- ❌ (`:x:`) or ⚠️ (`:warning:`) for more significant problems or concerns needing attention
- 🌱 (`:seedling:`) or ♻️ (`:recycle:`) for future improvements or indications of technical debt
- ⛏ (`:pick:`) for minor or nitpick changes

PR #17952, opened 2025-12-12
Revert "[PM-23258] changing verbiage from import data to import items"

itsadrago:
Reverts bitwarden/clients#17123
Needs 1 more QA pass


itsadrago:
Going to see if QA can work on this soon so that we don't require this revert

PR #17969, opened 2025-12-15
[CL-841] landing layout component

BryanCunningham:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/CL-841

## 📔 Objective

Adds a composable landing component

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


libs/components/src/anon-layout/anon-layout.component.html
```diff
@@ -17,56 +14,58 @@
...
+        <!-- Medium to Larger screens -->
+        <h1 bitTypography="h1" class="tw-mt-2 tw-hidden sm:tw-block">
+          {{ title() }}
+        </h1>
+      }
+
+      @if (subtitle()) {
+        <div class="tw-text-sm sm:tw-text-base">{{ subtitle() }}</div>
+      }
+    </div>
```
willmartian: (From our call)

I think we should look to make this into a reusable component. Otherwise, this is a shared piece of UI that multiple teams will need to recreate if we are moving them off of the anon-layout.

libs/components/src/landing-layout/components/header.component.ts
```diff
@@ -0,0 +1,8 @@
+import { ChangeDetectionStrategy, Component } from "@angular/core";
+
+@Component({
+  selector: "header[bitLandingHeader]",
+  changeDetection: ChangeDetectionStrategy.OnPush,
+  template: `<ng-content></ng-content>`,
+})
+export class LandingHeaderComponent {}
```
willmartian: For this (and footer and content), I would use an element selector instead of an attribute selector. Attribute selectors are only really needed when devs needs access to HTML attributes on the augmented element, like `button`, or when we need to preserve a specific parent-child relationship, like `tr`-`td`.

```ts
@Component({
  selector: "bit-landing-header",
  changeDetection: ChangeDetectionStrategy.OnPush,
  template: `<header><ng-content></ng-content></header>`,
})
```

The layout component can then select for this with: `<ng-content select="bit-landing-header></ng-content>`

This is also more consistent with how `bit-layout` and `bit-header` work together.

libs/components/src/landing-layout/index.ts
```diff
@@ -0,0 +1,5 @@
+export * from "./landing-layout.component";
+export * from "./landing-layout.module";
+export * from "./components/content.component";
+export * from "./components/footer.component";
+export * from "./components/header.component";
```
willmartian: nit: I don't find this extra  `/components` folder very helpful; I would kind of expect all of the components to be top level and prefixed with how the selector is named: `landing-header.component.ts`, etc

libs/components/src/anon-layout/anon-layout.component.html
```diff
@@ -1,11 +1,8 @@
...
-  [ngClass]="{
-    'tw-min-h-screen': clientType === 'web',
-    'tw-min-h-full': clientType === 'browser' || clientType === 'desktop',
-  }"
->
-  <div class="tw-flex tw-justify-between tw-items-center tw-w-full tw-mb-12">
+<bit-landing-layout [hideBackgroundIllustration]="hideBackgroundIllustration()">
+  <header
+    bitLandingHeader
+    class="tw-flex tw-bg-background-alt tw-items-center tw-w-full tw-p-5 tw-pb-12"
```
willmartian: Should these classes not be moved to the new header component? (likewise for classes on the content and footer components?)

libs/components/src/anon-layout/anon-layout.component.html
```diff
@@ -17,56 +14,58 @@
...
+      @if (hideCardWrapper()) {
+        <div class="tw-mb-6 sm:tw-mb-10">
+          <ng-container *ngTemplateOutlet="defaultContent"></ng-container>
+        </div>
+      } @else {
+        <bit-base-card
+          class="!tw-rounded-2xl tw-mb-6 sm:tw-mb-10 tw-mx-auto tw-w-full tw-bg-transparent tw-border-none tw-shadow-none sm:tw-bg-background sm:tw-border sm:tw-border-solid sm:tw-border-secondary-100 sm:tw-shadow sm:tw-p-8"
+        >
+          <ng-container *ngTemplateOutlet="defaultContent"></ng-container>
+        </bit-base-card>
```
willmartian: Should lift these styles into a component? `bit-landing-card` or something?

libs/components/src/anon-layout/anon-layout.component.html
```diff
@@ -17,56 +14,58 @@
...
+      }
+    </div>
 
-    <!-- In some scenarios this icon's size is not limited by container width correctly -->
-    <!-- Targeting the SVG here to try and ensure it never grows too large in even the media queries are not working as expected -->
     <div
-      *ngIf="iconInput !== null"
-      class="tw-size-20 sm:tw-size-24 [&_svg]:tw-w-full [&_svg]:tw-max-w-24 tw-mx-auto tw-content-center"
+      class="tw-z-10 tw-grow tw-w-full tw-mx-auto tw-flex tw-flex-col tw-items-center sm:tw-min-w-[28rem]"
+      [ngClass]="maxWidthClass"
```
willmartian: Perhaps a width input should be added to the landing-content component? Or should devs just use a Tailwind class?

libs/components/src/landing-layout/landing-layout.component.html
```diff
@@ -0,0 +1,25 @@
...
+      class="tw-hidden md:tw-block [&_svg]:tw-absolute tw-z-[1] tw-opacity-[.11] [&_svg]:tw-bottom-0 [&_svg]:tw-start-0 [&_svg]:tw-w-[35%] [&_svg]:tw-max-w-[450px]"
+    >
+      <bit-icon [icon]="leftIllustration"></bit-icon>
+    </div>
+    <div
+      class="tw-hidden md:tw-block [&_svg]:tw-absolute tw-z-[1] tw-opacity-[.11] [&_svg]:tw-bottom-0 [&_svg]:tw-end-0 [&_svg]:tw-w-[35%] [&_svg]:tw-max-w-[450px]"
+    >
+      <bit-icon [icon]="rightIllustration"></bit-icon>
+    </div>
+  }
```
willmartian, line 23: Should this be a slot plus helper component instead? Would that be more flexible if there are different image options in the future? Or would we do something like: `backgroundImage="none" | "foo" | "bar"`? Or am I overthinking this and it is best solved later?

libs/components/src/anon-layout/anon-layout.component.html
willmartian: I asked Claude about names, fwiw:

---

> The anon layout component has a block that includes a title, subtitle, and icon. I want to move this into a    
 helper component. For a name, the first think I think of is something like anon-layout-header, with             
slots/inputs for each. However, there is another area of the component that runs along the top that includes     
the app logo and optionally other items. That is also like a header, so there is a bit of a naming conflict.     
what are name options?                                                                                           
                                                                                                                 
⏺ Looking at this naming conflict, here are some options for the title/subtitle/icon block:                      
                                                                                                                 
  Page-oriented names:                                                                                           
  - anon-layout-page-header - Clear that it's the header for the page content, not the top navigation            
  - anon-layout-content-header - Similar distinction, emphasizes it's the content area header                    
  - anon-layout-page-title - Focuses on the primary purpose (displaying the page title)                          
                                                                                                                 
  Hero/banner style names:                                                                                       
  - anon-layout-hero - Common pattern name for prominent title/icon/subtitle sections                            
  - anon-layout-banner - Another common pattern, though sometimes implies dismissibility                         
                                                                                                                 
  Descriptive names:                                                                                             
  - anon-layout-heading - Simple and semantic                                                                    
  - anon-layout-title-block - Very descriptive but a bit verbose                                                 
  - anon-layout-intro - Emphasizes it introduces the page content                                                
                                                                                                                 
  My recommendation: anon-layout-page-header or anon-layout-hero                                                 
  - page-header is clear about hierarchy (the top bar could be called anon-layout-navbar or anon-layout-top-bar) 
  - hero is a well-established pattern name in design systems for this exact type of component

libs/components/src/anon-layout/anon-layout.component.html
BryanCunningham: Depends on how much flexibility we want. Will there ever be a situation where they want the header a different color? Who can say?! Figured this way, while slightly more cumbersome is more flexible for any future one-offs that may come up. Open to either approach though

libs/components/src/anon-layout/anon-layout.component.html
BryanCunningham: Maybe. Need clarification if we actually want this card to be different from our normal 'card' component. If so, then yes, I'll create a landing-card component and apply some of this there

BryanCunningham: Made a landing card component

libs/components/src/landing-layout/landing-layout.component.html
BryanCunningham: I need to check in with design but this might be going away entirely. If not, I think just leaving it in here with the ability to hide/show it is probably fine for now

BryanCunningham:
@claude review this PR

apps/browser/src/dirt/phishing-detection/popup/protected-by-component.html
```diff
@@ -1 +1,3 @@
-<span class="tw-text-muted">{{ "protectedBy" | i18n: "Bitwarden phishing blocker" }}</span>
+<span class="tw-text-center tw-block tw-text-muted"
+  >{{ "protectedBy" | i18n: "Bitwarden phishing blocker" }}</span
+>
```
willmartian: Is this diff intentional?

libs/components/src/anon-layout/anon-layout.component.html
```diff
@@ -1,76 +1,42 @@
...
-    }
-    <ng-content select="[slot=secondary]"></ng-content>
+    <bit-landing-max-width-container [maxWidth]="maxWidth()">
+      <bit-landing-hero
+        class="tw-bg-background-alt"
+        [icon]="icon()"
+        [title]="title()"
+        [subtitle]="subtitle()"
+      ></bit-landing-hero>
+      <bit-landing-content>
```
willmartian: Couldn't we do something like:

```
<bit-landing-content [maxWidth]="maxWidth()">
  <bit-landing-hero ... />
  ...
</bit-landing-content>
```

Why do we need both the container component _and_ the content component?

libs/components/src/landing-layout/landing-layout.component.html
willmartian: Did you have a chance to check in with design on this?

libs/components/src/landing-layout/landing-layout.component.ts
```diff
@@ -0,0 +1,26 @@
...
+
+  protected readonly clientType: string;
+  protected readonly leftIllustration = BackgroundLeftIllustration;
+  protected readonly rightIllustration = BackgroundRightIllustration;
+
+  private readonly platformUtilsService: PlatformUtilsService = inject(PlatformUtilsService);
+
+  constructor() {
+    this.clientType = this.platformUtilsService.getClientType();
+  }
```
willmartian: We don't need a constructor for this:

```ts
private readonly platformUtilsService = ...
private readonly clientType = ...
```

libs/components/src/landing-layout/landing-hero.component.html
```diff
@@ -0,0 +1,32 @@
...
+      </div>
+    }
+
+    @if (title()) {
+      <!-- Small screens -->
+      <h1 bitTypography="h2" class="tw-mt-2 sm:tw-hidden">
+        {{ title() }}
+      </h1>
+      <!-- Medium to Larger screens -->
+      <h1 bitTypography="h1" class="tw-mt-2 tw-hidden sm:tw-block">
```
willmartian, line 19: Just a thought for later: the devexp of using bitTypography with media queries is poor. This is a good reason to favor classes over style-only directives.

libs/components/src/anon-layout/anon-layout.component.html
BryanCunningham: I can try that out.

apps/browser/src/dirt/phishing-detection/popup/protected-by-component.html
BryanCunningham: Yes. By default the secondary content is no longer center aligned. Felt too opinionated to me. So, I needed to fix this one. But, I also don't know what other cans of worms this opens

libs/components/src/landing-layout/landing-layout.component.html
BryanCunningham: Not yet. Will do now

BryanCunningham: @willmartian Sounds like we will have some sort of background illustration. That being said, it's in flux still and probably best to cross this bridge when we come to it

apps/browser/src/dirt/phishing-detection/popup/protected-by-component.html
BryanCunningham: I can probably remove this now actually. I added back the center alignment to prevent any unintended consequences.

libs/components/src/landing-layout/landing-layout.stories.ts
willmartian, line 1: This file should use the newer Angular template syntax of @if and @switch (etc)

libs/components/src/anon-layout/anon-layout.component.html
```diff
@@ -1,76 +1,30 @@
...
-        {{ title() }}
-      </h1>
-    }
-
-    @if (subtitle()) {
-      <div class="tw-text-sm sm:tw-text-base">{{ subtitle() }}</div>
-    }
-  </div>
+<bit-landing-layout [hideBackgroundIllustration]="hideBackgroundIllustration()">
+  @if (!hideLogo()) {
```
vleague2: So this hides the entire header if `hideLogo` is false?

libs/components/src/landing-layout/landing-layout.component.html
```diff
@@ -0,0 +1,26 @@
...
+  class="tw-relative tw-flex tw-size-full tw-mx-auto tw-flex-col"
+  [class]="{
+    'tw-min-h-screen': clientType === 'web',
+    'tw-min-h-full': clientType === 'browser' || clientType === 'desktop',
+  }"
+>
+  <ng-content select="bit-landing-header"></ng-content>
+  <main class="tw-relative tw-flex tw-flex-1 tw-size-full tw-mx-auto tw-flex-col">
+    <ng-content select="bit-landing-content"></ng-content>
+    <ng-content></ng-content>
```
vleague2: Do these need to be separated? Since there isn't any special treatment of bit-landing-content vs default content

libs/components/src/anon-layout/anon-layout.component.html
BryanCunningham: Currently, yes. Not sure this is the best approach at the moment though.

libs/components/src/landing-layout/landing-layout.component.html
BryanCunningham: @vleague2 Not sure what you mean here exactly

vleague2: Like do we need `bit-landing-content` as a selector? Or could it all just be default content? Since they get placed right next to each other anyway

BryanCunningham: Oh, yeah, we do. It has some built in styling and max-width logic

BryanCunningham: We _probably_ don't need the extra `ng-content` block

vleague2: It makes sense to keep it as a component then. But idk if this file needs to do a special `select="bit-landing-content"`, they could just pass the component in as default content anyway?

BryanCunningham: Yeah, we could. Just sort of depends on how flexible we want this to be.

libs/components/src/anon-layout/anon-layout.component.html
vleague2: Yeah I guess it feels weird because we are passing the `hideLogo` input to the landing header in the next line. So if `hideLogo` is `false`, the bit-landing-header won't get rendered, so in effect it will only ever be receiving `hideLogo = true`

BryanCunningham: I'll revisit this.

libs/components/src/landing-layout/landing-content.component.ts
```diff
@@ -0,0 +1,58 @@
+import { ChangeDetectionStrategy, Component, input } from "@angular/core";
+
+export const LandingContentMaxWidth = {
+  Md: "md",
+  Lg: "lg",
+  Xl: "xl",
+  "2Xl": "2xl",
+  "3Xl": "3xl",
+  "4Xl": "4xl",
+} as const;
```
willmartian: I think an array type is simpler here:

```ts
export const LandingContentMaxWidth = ["md", "lg", "xl", "2xl", "3xl", "4xl"] as const;

export type LandingContentMaxWidthType =
  (typeof LandingContentMaxWidth)[number];
```

libs/components/src/landing-layout/landing-header.component.html
```diff
@@ -1,11 +1,14 @@
...
     <a
       [routerLink]="['/']"
-      class="tw-w-32 sm:tw-w-[200px] tw-self-center sm:tw-self-start tw-block [&>*]:tw-align-top"
+      class="tw-w-[calc(theme(spacing.32)_+_2_*_theme(spacing.5))] tw-p-5 sm:tw-w-[calc(200px_+_2_*_theme(spacing.5))] tw-self-center sm:tw-self-start tw-block [&>*]:tw-align-top"
     >
       <bit-icon [icon]="logo" [ariaLabel]="'appLogoLabel' | i18n"></bit-icon>
     </a>
   }
-  <ng-content></ng-content>
+  <div [class]="(hasContent$ | async) ? 'tw-ms-auto tw-p-5' : 'tw-ms-auto'" #contentWrapper>
```
vleague2: Ok so I am not confident in this suggestion but could we use the `has` css selector for this? On a quick test I think `has:(*)`(vanilla css) will work for child nodes but not for direct child content like text (like `<span>hi</span>` it wouldn't catch the "hi" but it would catch `<span><span>hi</span></span>`).

BryanCunningham: @vleague2 I'd have to double check but, with content projection those selectors like `:has` and `:empty` don't seem to work. I tried `:empty` already with no luck

BryanCunningham: This works! Updated

libs/components/src/landing-layout/landing-header.component.html
```diff
@@ -0,0 +1,14 @@
+<header class="tw-flex tw-w-full tw-bg-background-alt tw-px-5">
+  @if (!hideLogo()) {
+    <!-- CSS calc on width to account for horizontal padding of the logo -->
```
vleague2: Probably can remove this comment now

libs/components/src/landing-layout/landing-content.component.ts
```diff
@@ -0,0 +1,58 @@
...
+  /**
+   * Max width of the landing layout container.
+   *
+   * @default LandingContentMaxWidth.Md
+   */
+  readonly maxWidth = input<LandingContentMaxWidthType>(LandingContentMaxWidth.Md);
+
+  get maxWidthClasses(): string {
+    const maxWidth = this.maxWidth();
+    let maxWidthClass = "";
```
willmartian: Computed signals should be preferred over getters; they cause fewer re-renders.

libs/components/src/landing-layout/landing-content.component.ts
```diff
@@ -0,0 +1,58 @@
...
+      case "2xl":
+        maxWidthClass = "tw-max-w-2xl";
+        break;
+      case "3xl":
+        maxWidthClass = "tw-max-w-3xl";
+        break;
+      case "4xl":
+        maxWidthClass = "tw-max-w-4xl";
+        break;
+    }
```
willmartian: nit: using a switch plus a mutable variable seems odd here; a record feels more idiomatic to me: 
```ts
const sizeClass = {
  "md": "tw-max-md:,
  "lg": "tw-max-w-lg",
  //...
}
```

libs/components/src/landing-layout/landing-header.component.ts
```diff
@@ -0,0 +1,18 @@
...
+
+@Component({
+  selector: "bit-landing-header",
+  changeDetection: ChangeDetectionStrategy.OnPush,
+  templateUrl: "./landing-header.component.html",
+  imports: [RouterModule, IconModule, SharedModule],
+})
+export class LandingHeaderComponent {
+  readonly hideLogo = input<boolean>(false);
+  protected logo = BitwardenLogo;
```
willmartian: `logo` should be `readonly` as well

libs/components/src/landing-layout/landing-hero.component.html
```diff
@@ -0,0 +1,32 @@
+@let iconInput = icon();
+@let titleText = title();
+@let subtitleText = subtitle();
```
willmartian: Imo, `@let` should only be used for creating new information. Using it to access signals just creates more variable names to keep track of.

libs/components/src/landing-layout/landing-layout.component.html
willmartian: I agree that we don't need two slots here and can just use the default `ng-content` slot

libs/components/src/landing-layout/landing-layout.component.html
```diff
@@ -0,0 +1,26 @@
...
+  </main>
+  @if (!hideBackgroundIllustration()) {
+    <div
+      class="tw-hidden md:tw-block [&_svg]:tw-absolute tw-z-[1] tw-opacity-[.11] [&_svg]:tw-bottom-0 [&_svg]:tw-start-0 [&_svg]:tw-w-[35%] [&_svg]:tw-max-w-[450px]"
+    >
```
willmartian, line 15: I don't like how target the internal svg of `bit-icon`, but I understand why it is necessary. Seems like we will need to change something in bit-icon to fix it. No action needed here I think, just rambling.

libs/components/src/landing-layout/landing-hero.component.html
BryanCunningham: I thought it was more performant/better to use a variable if you access the signal multiple times. But, I could be imagining that

libs/components/src/landing-layout/landing-layout.stories.ts
BryanCunningham: I'll update it. Was just a lazy copy/paste from the anon layout story :)

libs/components/src/landing-layout/landing-content.component.ts
BryanCunningham: If I update to an array, make using the const to set the values more difficult. 
```js
readonly maxWidth = input<LandingContentMaxWidthType>(LandingContentMaxWidth.Md);
```

libs/components/src/landing-layout/landing-content.component.ts
BryanCunningham: Updated [here](https://github.com/bitwarden/clients/pull/17969/commits/da86f2bda5c23452bbb140c1e7f0e0aa551a951d)

libs/components/src/landing-layout/landing-hero.component.html
willmartian: I don't think so. They will still need to bind to and update the DOM everywhere they are used. I think it might be true when using the `async` pipe with observables, but I would need to double check.

libs/components/src/landing-layout/landing-content.component.ts
willmartian: I prefer strings over enums or enum-likes:

```ts
readonly maxWidth = input<LandingContentMaxWidthType>("md");
```

You still get autocomplete / IDE hints, and it more closely matches usage in the template: `maxWidth="md"`

libs/components/src/landing-layout/landing-hero.component.html
BryanCunningham: All good, I'll remove them and we can do some further research

libs/components/src/landing-layout/landing-content.component.ts
```diff
@@ -0,0 +1,63 @@
+import { ChangeDetectionStrategy, Component, computed, input } from "@angular/core";
+
+export const LandingContentMaxWidth = ["md", "lg", "xl", "2xl", "3xl", "4xl"] as const;
+
+export type LandingContentMaxWidthType = (typeof LandingContentMaxWidth)[number];
```
willmartian, line 5: nit: With how it is now, you don't actually need the runtime array anymore and can just define the type directly:

```ts
export type LandingContentMaxWidth = "md" | "lg" | "xl" | "2xl" | "3xl" | "4xl";
```

willmartian (approved):
Looks good to me! Thank you for your willingness to iterate 🤝

PR #17976, opened 2025-12-15
Enable cross-compilation and packaging of Windows Appx from macOS

iinuwa:
## 🎟️ Tracking
[PM-29784](https://bitwarden.atlassian.net/browse/PM-29784)

## 📔 Objective

This allows a developer on macOS to create Windows Appx files, creating a faster dev loop and less local resources if running on a VM. I've copied a desirability statement.

Since Engineering mostly has Macbooks, developing for Windows can present some challenges that decrease productivity: 

- Rust and Node compilation times are slower and require lots of memory dedicated to the VM.
- Rust and Node assets require 80+GB of data, which is all duplicated if you’ve built the Bitwarden client on your host machine.
- Keyboards and mice don’t quite work the same way, and sometimes using the clipboard is touchy.
- It requires duplicating the entire dev setup inside the VM, or setting up some remote coding session.

It would be better if we could develop and build in our native environment, and merely run the compiled resources in a local VM. There are two problems:

- We need to set up our build system for native Rust code cross-compilation from macOS → Windows. (We have this mostly done, but there are a few edges that need to be smoothed out.)
- For certain scenarios, like working on the native passkey provider plugin we need to package the file as an Appx, which electron-builder does not support doing on macOS without a Parallels VM.

We can leverage some open source tools—cargo-xwin, msix-packaging, and osslsigncode—to resolve these problems.
</p>
</details> 

Comment from @djsmith85: Closes https://github.com/bitwarden/clients/issues/18079

### Prerequisites

This script requires `cargo-xwin`, `makemsix` from [`microsoft/msix-packaging`](https://github.com/microsoft/msix-packaging), and `osslsigncode`.

I have written a [Homebrew tap script](https://github.com/iinuwa/homebrew-msix-packaging-tap/) for building `makemsix` from source, so this can be set up with:

```
brew install iinuwa/msix-packaging-tap/msix-packaging osslsigncode
cargo install --version 0.20.2 --locked cargo-xwin
```

### Executing script

The script is written in cross-platform PowerShell, but I can convert this to a different kind of script if desired.

```shell
cd apps/desktop
./cross-build.ps1 \
  -Architecture arm64 \
  -Target Debug \
  -ElectronConfigFile ./electron-builder.json \
  -CertificatePath /path/to/code-signing.pfx \
  -CertificatePassword "1234"
```

After this completes, you should have a debug version Appx build in dist/ that you can copy to a Windows VM to execute.

### Key generation
You can generate a key using OpenSSL:

```shell
publisher="/CN=14D52771-DE3C-4886-B8BF-825BA7690418"
openssl req -x509 -newkey rsa:2048 -days 3650 -nodes \
  -addext 'keyUsage=critical,digitalSignature' \
  -addext 'extendedKeyUsage=critical,codeSigning' \
  -addext 'basicConstraints=critical,CA:FALSE' \
  -subj "$publisher" -keyout mysigning.rsa.pem -out mysigning.cer
openssl pkcs12 -inkey mysigning.rsa.pem -in mysigning.cer -export -out mysigning.p12
```

If testing an Appx on Windows, the `mysigning.cer` file needs to be imported into the "Trusted People" certificate store for the local machine by an admin before the Appx can be installed.

### Documentation

After this is merged, I will follow up with a PR to update the [cross-compilation section](https://contributing.bitwarden.com/getting-started/clients/desktop/#cross-compile) of the desktop client Getting Started docs.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

[PM-29784]: https://bitwarden.atlassian.net/browse/PM-29784?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/desktop/scripts/after-pack.js
+++ b/apps/desktop/scripts/after-pack.js
@@ -9,8 +9,8 @@ const fse = require("fs-extra");
 exports.default = run;
 
 /**
- * 
- * @param {builder.AfterPackContext} context 
+ *
+ * @param {builder.AfterPackContext} context
  */
 async function run(context) {
   console.log("## After pack");
--- a/apps/desktop/scripts/before-pack.js
+++ b/apps/desktop/scripts/before-pack.js
@@ -1,32 +1,31 @@
-/* eslint-disable @typescript-eslint/no-require-imports, no-console */
-const builder = require("app-builder-lib");
+/* eslint-disable no-console */
+/** @import { BeforePackContext } from 'app-builder-lib' */
 exports.default = run;
 
 /**
- * @param {builder.BeforePackContext} context 
+ * @param {BeforePackContext} context
  */
 async function run(context) {
   console.log("## before pack");
-  console.log("Stripping .node files that don't belong to this platform...")
-  removeExtraNodeFiles(context)
+  console.log("Stripping .node files that don't belong to this platform...");
+  removeExtraNodeFiles(context);
 }
-
 
 /**
  * Removes Node files for platforms besides the current platform being packaged.
  *
- * @param {builder.BeforePackContext} context
+ * @param {BeforePackContext} context
  */
 function removeExtraNodeFiles(context) {
-    // When doing cross-platform builds, due to electron-builder limitiations,
-    // .node files for other platforms may be generated and unpacked, so we
-    // remove them manually here before signing and distributing.
-    const packagerPlatform = context.packager.platform.nodeName;
-    const platforms = ["darwin", "linux", "win32"];
-    const fileFilter = context.packager.info._configuration.files[0].filter;
-    for (const platform of platforms) {
-      if (platform != packagerPlatform) {
-        fileFilter.push(`!node_modules/@bitwarden/desktop-napi/desktop_napi.${platform}-*.node`)
-      }
+  // When doing cross-platform builds, due to electron-builder limitiations,
+  // .node files for other platforms may be generated and unpacked, so we
+  // remove them manually here before signing and distributing.
+  const packagerPlatform = context.packager.platform.nodeName;
+  const platforms = ["darwin", "linux", "win32"];
+  const fileFilter = context.packager.info._configuration.files[0].filter;
+  for (const platform of platforms) {
+    if (platform != packagerPlatform) {
+      fileFilter.push(`!node_modules/@bitwarden/desktop-napi/desktop_napi.${platform}-*.node`);
     }
+  }
 }


addisonbeck:
Claude's feedback seems valid here. This script also includes a lot of 3rd party tools, but doesn't assert that any of them are installed. I think this really could use some docs, even if it's just a big comment at the top of the script. Nothing that is going on here is likely going to be very familiar to anyone reaching for this script.

Could this approach be extended to support Linux → Windows or Windows → macOS cross-compilation?

iinuwa:
Ah, I forgot to link my documentation [draft](https://bitwarden.atlassian.net/wiki/spaces/~712020b5f550d333d041539d62b9ce7930f633/pages/2195456023/Packaging+Desktop+Client+as+Windows+Appx+from+macOS): I have added some of that context to the script documentation, which can be shown in PowerShell using `Get-Help ./appx-cross-build.ps1 -Full`.

> Could this approach be extended to support Linux → Windows or Windows → macOS cross-compilation?

Linux -> Windows should work, but I haven't tried compiling msix-packaging on Linux yet; I assume that will be easier than macOS was. (This might not be necessary since the `makeappx` tool that electron-builder downloads may work out of the box on Linux with Wine: this doesn't work on macOS with Wine, which is the reason I wrote this script.)

I don't think cross-compiling to macOS would work as well. I haven't done a ton of research, but it seems that while compiling would work with `cargo-zigbuild` for macOS in a similar way as `cargo-xwin` for Windows (NAPI uses both these tools for cross-compilation), signing wouldn't work. I don't think there's a way to do cross-platform codesigning for macOS, and I think that not only the package must be signed, but also individual executables within the macOS package must be signed. So targeting macOS might not work.

I don't see Claude's inline comments anywhere, so I'll take a guess based on the summary above.

> Certificate password exposed in command line arguments (security risk)

Since this is intended for a local development flow, where the signing certificate is local to the developer's machine. I overlooked that someone may use this script with a real certificate, which could be an issue. I changed this to accept a `SecureString` or use an environment variable.

> Logic error: incorrect variable check causes build mode mismatch

Woops, I had this fix in a separate commit that I forgot to include here. Fixed.

> Clarification needed on buildImporterBinaries behavior during cross-compilation

Fixed.

> Hardcoded path separators may cause issues

PowerShell handles path separators transparently.

> Missing input validation for architecture parameter

I changed this to use a `ValidationSet` instead.

> No verification that required external tools are installed

I have added a step in `desktop_native/build.js` to install the tools via Homebrew on macOS. There is also now a check at the beginning of the script for these tools with instructions to install.

> Missing error handling for JSON parsing operations

PowerShell will give an error if the file is missing or cannot be parsed. Missing parameters would also be broken in CI.

iinuwa:
I'm not sure what's going on with the macOS release failing in CI. I'll come back to that on Friday. Marking as draft for now until that's ready.

iinuwa:
@addisonbeck, sorry for the churn. This should be good to go now, though I don't know what's going on with the failing test; it appears to be present on `main` and unrelated to this PR.

I had to implement the `beforePack` hook in order to remove the irrelevant platform-specific `.node` files because of the strange way that the top-level and target-specific `files` key combines globs together. That should ensure that extra `.node` files left over in `desktop_native/dist` don't get copied to packages for other platforms.

The final list of changes in this PR is:

* Enable cross-compilation and packaging of Windows Appx from macOS
* Consolidate cargo build execution into a single function in native build script
* Install cargo-xwin when needed
* Install Appx tools when needed
* Consolidate command execution into a single function in native build script
* Only include the native node modules for the appropriate platform
* Always copy Rust binaries to dist folder
* Log source and destination when copying files
* Update copyright
* Update Beta Electron version to match Release Electron version

PR #18008, opened 2025-12-16
Feature/pm 25865 migrate send list desktop migration

iivins-livefront:
Overview:

This PR moves the Desktop Send list UI into a shared library component and updates the Desktop Send v2 component to use modern Angular patterns (Signals, OnPush, no manual subscriptions)

Key Changes:

New Shared Library Component (`libs/tools/send/send-ui/`)
- `SendListComponent`: A reusable component that shows the Send list with search, table, and empty states. Takes data as inputs and fires events when users interact with it. All the business logic stays in the parent component

Desktop App Changes (`apps/desktop/src/app/tools/send-v2/`)

Updated `SendV2Component` to use Signals:
- Changed `@ViewChild` to `viewChild()` signal
- Converted component state to Signals (sendId, action, etc.)
- Replaced subscriptions with `toSignal()` for reactive data
- Added OnPush change detection
- Removed the `BaseSendComponent` parent class
- Fixed memory leaks by removing manual subscriptions

Added `NewSendDropdownComponent`:
- New dropdown button for creating Text or File Sends.
- Checks if user has premium before allowing File sends.
- Uses events instead of routing to work with Desktop's modal system

Added tests for the new components and updated existing ones


libs/tools/send/send-ui/src/new-send-dropdown-v2/new-send-dropdown-v2.component.ts
Hinton, line 1: I wonder if we should move this into the libs/tools/send library? I think this does pretty much exactly what the existing component does except uses outputs instead which i what we should update the browser and web versions to also use. What do you think @itsadrago?

apps/desktop/src/app/tools/send-v2/send-v2.component.html
```diff
@@ -1,110 +1,61 @@
 <div id="sends" class="vault">
   <div id="items" class="items">
```
Hinton: Could we get rid of the items class & id and replace it with tailwind or add some custom css overrides in the `migration.css`? I think we'd want the width to be bigger than it currently is. I think this is because previously the width was calculated on the full layout including navigation but now it's just in the content area. Using `w-2/5` (40%) rather than 28% provided by the css class might be a good balance.

libs/tools/send/send-ui/src/send-list/send-list.component.html
```diff
@@ -0,0 +1,45 @@
+<!-- Loading state -->
+@if (loading()) {
+  <div class="tw-my-4 tw-px-4">
```
Hinton: **suggestion:** Can we put `tw-my-4 tw-px-4` in the `send-list.component`? This padding will be removed later on once we are able to stop overriding the layout padding. That way we won't have to touch these components.

libs/tools/send/send-ui/src/send-list/send-list.component.html
```diff
@@ -0,0 +1,45 @@
...
+    <i
+      class="bwi bwi-spinner bwi-spin tw-text-muted"
+      title="{{ 'loading' | i18n }}"
+      aria-hidden="true"
+    ></i>
+    <span class="tw-sr-only">{{ "loading" | i18n }}</span>
```
Hinton: I think we can use the `bit-spinner` here. https://components.bitwarden.com/?path=/docs/component-library-spinner--docs

libs/tools/send/send-ui/src/send-list/send-list.component.html
```diff
@@ -0,0 +1,45 @@
...
+  </div>
+}
+
+<!-- Search bar - always visible when not loading -->
+@if (!loading()) {
```
Hinton: Use `@else`?

libs/tools/send/send-ui/src/send-list/send-list.component.ts
```diff
@@ -0,0 +1,83 @@
...
+
+  // Reusable data source instance - updated reactively when sends change
+  private readonly _dataSource = new TableDataSource<SendView>();
+
+  // Computed signal that updates the existing data source with new data
+  // Returns the same reference so bit-table doesn't need to reconnect
+  protected readonly dataSource = computed(() => {
+    this._dataSource.data = this.sends();
+    return this._dataSource;
+  });
```
Hinton: **suggestion:** I think you want to use effect here?

libs/tools/send/send-ui/src/send-list/send-list.component.html
iivins-livefront: A bit confused here - you want to add these in `send-list.component.ts` using the styles property in the @Component decorator? Or are you're talking about moving that div/padding outside of the send-list.component and into the send-V2 component?

libs/tools/send/send-ui/src/new-send-dropdown-v2/new-send-dropdown-v2.component.ts
iivins-livefront: Either or work for me

itsadrago: Sounds good to me

Hinton: I think this means we move it to libs @iivins-livefront.

libs/tools/send/send-ui/src/send-list/send-list.component.html
Hinton: Moving it to send-v2.component. Unless you disagree, it's a minor thing and we can fix it once vault also uses the new list style.

libs/tools/send/send-ui/src/send-list/send-list.component.html
```diff
@@ -1,45 +1,27 @@
...
-  </div>
-}
+    <!-- Empty state: No sends exist -->
+    @if (listState() === sendListState.Empty || listState() === sendListState.NoResults) {
+      <bit-no-items [icon]="noItemIcon" class="tw-text-main">
+        <ng-container slot="title">{{ "sendsTitleNoItems" | i18n }}</ng-container>
+        <ng-container slot="description">{{ "sendsBodyNoItems" | i18n }}</ng-container>
+        <ng-content select="[slot='empty-button']" slot="button" />
+      </bit-no-items>
+    } @else if (listState() === null) {
```
Hinton: **question:** Could we just have `else` here. I think we're catching all scenarios?

libs/tools/send/send-ui/src/send-list/send-list.component.ts
```diff
@@ -52,13 +58,15 @@ export class SendListComponent {
 
   // Reusable data source instance - updated reactively when sends change
   private readonly _dataSource = new TableDataSource<SendView>();
+  protected readonly dataSource = this._dataSource;
```
Hinton: I think we can just skip `_datasource`?
```suggestion
  protected readonly dataSource = new TableDataSource<SendView>();
```

libs/tools/send/send-ui/src/send-list/send-list.component.ts
```diff
@@ -52,13 +58,15 @@ export class SendListComponent {
...
 
-  // Computed signal that updates the existing data source with new data
-  // Returns the same reference so bit-table doesn't need to reconnect
-  protected readonly dataSource = computed(() => {
-    this._dataSource.data = this.sends();
-    return this._dataSource;
-  });
+  constructor() {
+    // Effect to update data source when sends change
+    // Using effect() instead of computed() since we're performing a side effect (updating existing object)
```
Hinton:
```suggestion
```

libs/tools/send/send-ui/src/send-list/send-list.component.html
iivins-livefront: Sounds good to me 👍

libs/tools/send/send-ui/src/new-send-dropdown-v2/new-send-dropdown-v2.component.ts
iivins-livefront: @Hinton there's an existing "tools-new-send-dropdown" component in libs/tools/send/send-ui/src/new-send-dropdown

Is there a specific naming cadence we should use? tools-new-send-dropdown-v2 was my initial thought

iivins-livefront:
> I saw we have some failing tests. But besides the comments I think this looks good.
> 
> I noticed it doesn't matter which sends you pick it just selects file. Is this something we should fix in this PR or a follow up PR? I think we'd want it resolved before Milestone 1, or if we remove the dropdown feature and just have a basic add button?

@Hinton I was able to fix the failing tests - for the second part, you're talking about when add-edit component gets opened by the Add New CTA from an empty Files Send-List, add-edit is always default to File right? I'd like to tackle that when I update the add-edit component if possible

Hinton:
@iivins-livefront migrating to the new add-edit component is part of milestone 3, this is milestone 1 and ideally milestones can be released in order.

libs/tools/send/send-ui/src/send-list/send-list.component.html
Hinton, line 1: If you look at the figma we should do the following changes:

- Hide search if no sends
- Make bit-no-items centered horizontally.

https://www.figma.com/design/9cGXB72BSoSjQpsrVnjdrF/Desktop-2025?node-id=593-81669&m=dev

libs/tools/send/send-ui/src/send-list/send-list.component.html
```diff
@@ -1,27 +1,25 @@
...
-      <tools-send-table
-        [dataSource]="dataSource"
-        [disableSend]="disableSend()"
-        (editSend)="onEditSend($event)"
-        (copySend)="onCopySend($event)"
-        (removePassword)="onRemovePassword($event)"
-        (deleteSend)="onDeleteSend($event)"
-      />
-    }
+    <bit-no-items [icon]="noItemIcon" class="tw-text-main">
```
Hinton:
```suggestion
    <bit-no-items [icon]="noItemIcon">
```

libs/tools/send/send-ui/src/send-list/send-list.component.html
```diff
@@ -1,27 +1,25 @@
 <!-- Loading state -->
-<div class="tw-my-4 tw-px-4">
-  @if (loading()) {
-    <bit-spinner [title]="'loading' | i18n"></bit-spinner>
-  } @else {
-    <!-- Search bar - always visible when not loading -->
-    <tools-send-search></tools-send-search>
+@if (loading()) {
+  <bit-spinner [title]="'loading' | i18n"></bit-spinner>
```
Hinton: **suggestion:** Drop the title since it's the default, and make it a self-closing tag.
```suggestion
  <bit-spinner />
```

apps/desktop/src/app/tools/send-v2/send-v2.component.html
```diff
@@ -1,110 +1,63 @@
...
+          (copySend)="onCopySend($event)"
+          (deleteSend)="onDeleteSend($event)"
+          (removePassword)="onRemovePassword($event)"
         >
-          <i class="bwi bwi-plus bwi-lg" aria-hidden="true"></i>
-        </button>
+          <tools-new-send-dropdown-v2
+            slot="empty-button"
+            [hideIcon]="true"
+            [buttonType]="'primary'"
```
Hinton:
```suggestion
            buttonType="primary"
```

Hinton:
There is a type error in one of the tests (`npm run test:types`)

libs/tools/send/send-ui/src/send-list/send-list.component.html
iivins-livefront: For when Searches exist BUT there are no search results for user input - do we want to show the Add New?
Screenshot 2025-12-19 at 10 38 02 AM

^This matches current web functionality

Hinton: Checking with the team. Behaving the same as the web vault is good for now although I personally prefer the way vault items works which is showing a "no items to list"

Hinton: Bit of a change, we want to show a specific error when searching or filtering and not finding anything. We want to continue showing the table header, so I suppose it can be an element under the table?
image
image

itsadrago:
@iivins-livefront once you have addressed @Hinton's remaining comments please let me know

iivins-livefront:
@itsadrago updated 👍

itsadrago (approved):

PR #18016, opened 2025-12-17
[PM-29846] Update sdk to 433

quexten:
## 🎟️ Tracking
https://bitwarden.atlassian.net/browse/PM-29846

## 📔 Objective

Update the SDK to 433 to include the fix from https://github.com/bitwarden/sdk-internal/pull/614#event-21600868167.


quexten:
Looks like the build failures are from `main` being broken currently.

djsmith85 (approved):

PR #18054, opened 2025-12-18
Add Windows WebAuthn API Rust wrapper library

iinuwa:
## 🎟️ Tracking

[PM-29785](https://bitwarden.atlassian.net/browse/PM-29785)

## 📔 Objective

This adds a Rust wrapper around the Windows WebAuthn API types and functions required to implement a passkey plugin authenticator. It also includes a lightweight framework to run the associated COM server by just implementing the `PluginAuthenticator`.

A side goal is to make this generic enough that other projects can use this to implement their own plugin authenticators. (That is why, for example, we are following the Microsoft sample code by using Windows crypto libraries rather than Rust- or Bitwarden-specific libraries.)

A complete usage is in the `iinuwa/win-passkey-poc` branch (for example, the PluginAuthenticator trait implementation is [here](https://github.com/bitwarden/clients/blob/iinuwa/win-passkey-poc/apps/desktop/desktop_native/windows_plugin_authenticator/src/process.rs#L139-L249)), but this PR just adds the wrapper for a narrower focus while I work on merging some other required changes for that branch.

A recommendation for review would be to skip the struct definitions, since they're pretty boring, mostly autogenerated and have been demonstrated to be defined correctly in the proof-of-concept. Focusing on the functions, especially on the `unsafe` blocks and the corresponding `SAFETY:` comments should ensure that everything is sound. (I believe that it is, but it'd be nice to get some other eyes on it.)

Let me know if there is a better way to break this up.

(Also, the `include/` files don't need to be reviewed; it's just there to know which version of the API we are working with.)

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

[PM-29785]: https://bitwarden.atlassian.net/browse/PM-29785?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


iinuwa:
Cc: @coltonhurst, since this builds on a lot of your work.

apps/desktop/desktop_native/win_webauthn/src/plugin/crypto.rs
```diff
@@ -0,0 +1,249 @@
...
+            hash,
+            signature,
+            cng_flags,
+        )?;
+        tracing::debug!("Verified");
+        Ok(())
+    }
+}
+
+pub(super) fn hash_sha256(data: &[u8]) -> Result<Vec<u8>, windows::core::Error> {
```
quexten: Question: Is this really just doing sha256? If so, can we consider just using the RustCrypto impl? It is safe rust.

If it is somehow different, we should document how it differs.

iinuwa:
Regarding Claude's review:

> CRITICAL: Memory Allocator Mismatch at COM Boundary

`webauthn_plugin_encode_make_credential()` takes HLOCAL-allocated memory in a WEBAUTHN_CREDENTIAL_ATTESTATION struct and returns COM-allocated CTAP bytes. THe OS API takes ownership of the memory, so that's why we use std::mem::forget() here. It is not correct to use a COM buffer here.

> ❌ Potential Null Pointer Dereference in Crypto

The pointer that is reference is coming from a `NonNull<u8>`; the null check is guaranteed by the type.

> ⚠️ CRITICAL: Inverted Logic in Signature Verification

This is valid; I have pushed a fix to delete the `if` statement altogether now that I'm always sending the request hash that is being signed.

iinuwa:
@djsmith85, this is ready for a re-review.

The failing test seems to be present on `main` and unrelated to this PR.

coltonhurst (commented):
@iinuwa this PR takes a very different approach for interfacing with Windows passkeys than was taken previously. This uses the pluginauthenticator.idl and goes the MIDL route.

Can you give some context on the switch?

iinuwa:
@coltonhurst, I was under the impression that this is the same work that you did, but it looks like @abergs made the change in https://github.com/bitwarden/clients/commit/b2017ed9ee6d1a2e9193066b9f001436d4c5955d when he picked up where you left off. Perhaps he could give some context on the switch to MIDL?

abergs:
@coltonhurst @iinuwa I based my implementation on top of the the samples from Microsoft. I think MS uses IDL files to to define COM interfaces and we were able to piggy back on that to get the correct guids, layout, ordering etc.

I don't think we need the `.idl` file checked in (unless we want it for reference), since the pluginauthenticated.h is pre-generated from it.

Edit: Link to the ms webauthn sample where the idl and .h (generated from the .idl) is: https://github.com/microsoft/webauthn/tree/master

Edit2: @coltonhurst Are you sure this is very different? I don't think we're using `MIDL` directly, just the `.h` file that was pre generated by midl and the .idl file. I'm a novice in this, so I may be wrong and missing something.

Edit 3: As the PR description mentions, not even the `.h` files are referenced in code -- they just serve as documentation and the actual structs and interfaces are implemented in `types.rs` etc. We want to rename that folder to clarify that or remove them all together (but it was very valuable early on to compare with MS source)?

iinuwa:
To remove confusion, I have removed the `include/` files and instead added a reference to the specific commit that we were working from.

coltonhurst:
@abergs @iinuwa I've started to dig into reviewing this today. While this does seem to call the Window's API's in a similar way (the old stuff hasn't been touched in a while and the API's have gone public, so I'm sure there are differences), the generation of the types (and potentially more?) through MIDL is new. That is not necessarily a bad thing; it just has its own pros and cons.

I will add more complete thoughts as I work through this PR; I also have a meeting with @iinuwa later today. But some very brief initial bullet point thoughts:

- Regarding MIDL for type generation:
    - If we are going to use MIDL to generate types, we become subject to the very unclear naming Microsoft has. This is a con that has more impact than one might initially think, because reading and working on the code becomes much more tedious until someone has enough domain knowledge where they've effectively "memorized" the names. I don't think generating the types via MIDL is bad; it might be nice to have a solution to the naming issue though.
    - If we want to go the route of generating types with MIDL, we need a documented way to do this, with instructions on what someone updating the code must do to generate these types, and how to add them to the project. This is the bare minimum; ideally this is automated and verified during compile time or checked within CI during changes, etc.
    - We also will need to have a plan for Windows API changes and how we will handle them across different versions a client might be on (this may be in this PR and I haven't seen it yet, or may be coming in a future pr, or is at least documented somewhere?).
- Regarding code organization:
    - Overall, the code seems to be pretty disorganized. It's lightly organized into modules but even then things are inconsistent. Is this due (in part) to the MIDL generation? Maybe some info there would be helpful. Here are just two examples:
        - There are API function calls combined with type definitions. It would be much more clear to cleanly abstract these separately, if possible
        - WebAuthn header file functions and PluginAuthenticator header file functions are both being defined within the types file under the plugin module. Then there are more WebAuthn functions being defined in the types module outside of that. Why are these being mixed?
    - Unsafe code is sprinkled throughout this code in a _seemingly_ unclear way. This is really not ideal. It may not be possible, but ideally the pattern I was hoping to establish previously was cleanly wrapping unsafe items within safe items (really only types or functions), with no logic in the unsafe layer. Then all business / passkey logic happens in safe layers on top of that. Am I missing how unsafe is being used, besides just as needed?
- There are three tests. That may cover everything we can test, I'm not sure yet. Is there more surface area we can cover through more tests?

I might have my own questions answered as I continue the review, but I wanted to at least share some initial things I see so we can start the conversation about them earlier rather than later. Apologies if some of the things above are coming in future pr's or have been documented somewhere and I've missed them.

coltonhurst:
>A side goal is to make this generic enough that other projects can use this to implement their own plugin authenticators. (That is why, for example, we are following the Microsoft sample code by using Windows crypto libraries rather than Rust- or Bitwarden-specific libraries.)

@Thomas-Avery @quexten are there any concerns or thoughts on the KM side doing this? Wanted to at least make sure this was visible.

apps/desktop/desktop_native/win_webauthn/src/plugin/crypto.rs
```diff
@@ -0,0 +1,287 @@
...
+                if let Err(err) = BCryptDestroyHash(self.handle).to_hresult().ok() {
+                    tracing::error!("Failed to clean up hash object: {err}");
+                }
+            }
+        }
+    }
+}
+
+/// Signing key for an operation request or user verification response buffer.
+pub struct SigningKey {
```
quexten:
```suggestion
pub struct VerifyingKey {
```
or
```suggestion
pub struct PublicKey {
```

apps/desktop/desktop_native/win_webauthn/src/api/plugin/crypto.rs
```diff
@@ -0,0 +1,287 @@
...
+                    Some(hash_slice),
+                    None,
+                    0,
+                )
+                .ok()?;
+            }
+            // SAFETY: BCryptCreateHash initializes hash_handle and hash_obj_buf.
+            // This memory must be preserved until hash_handle is dropped, but will be cleaned up on
+            // call BCryptDestroyHash.
+            mem::forget(hash_obj_buf);
```
quexten, line 219: Since this is not using the regular drop impl, does this get zeroized?

apps/desktop/desktop_native/win_webauthn/src/plugin/com.rs
```diff
@@ -0,0 +1,478 @@
...
+/// - pbEncodedRequest points to a valid non-null byte string of length cbEncodedRequest.
+/// - pbRequestSignature points to a valid non-null byte string of length cbRequestSignature.
+unsafe fn verify_operation_request(
+    request: &WEBAUTHN_PLUGIN_OPERATION_REQUEST,
+    clsid: &GUID,
+) -> Result<(), WinWebAuthnError> {
+    tracing::debug!("Verifying request");
+    let request_data =
+        std::slice::from_raw_parts(request.pbEncodedRequest, request.cbEncodedRequest as usize);
+    let request_hash = crypto::hash_sha256(request_data).map_err(|err| {
```
quexten: Suggestion: Add a comment as follows:
```
/// Request verification requires verifying the signature over the sha256 hash of the serialized request bytes. The hash has to be derived here.
```

apps/desktop/desktop_native/win_webauthn/src/plugin/com.rs
```diff
@@ -0,0 +1,478 @@
...
+///
+/// # Safety
+/// The caller must ensure that:
+/// - pbEncodedRequest points to a valid non-null byte string of length cbEncodedRequest.
+/// - pbRequestSignature points to a valid non-null byte string of length cbRequestSignature.
+unsafe fn verify_operation_request(
+    request: &WEBAUTHN_PLUGIN_OPERATION_REQUEST,
+    clsid: &GUID,
+) -> Result<(), WinWebAuthnError> {
+    tracing::debug!("Verifying request");
```
quexten: Suggestion: You can make this code a lot nicer with the following:

Introduce the following NewTypes:
- Signature
- RequestHash

Then, implement:
```
Signature::from<&WEBAUTHN_PLUGIN_OPERATION_REQUEST>()
RequestHash::from<&WEBAUTHN_PLUGIN_OPERATION_REQUEST>()
```

now, public_key.verify(...) does not take byte buffers but the above newtypes.

This would make this much more type-safe.

apps/desktop/desktop_native/win_webauthn/src/plugin/crypto.rs
```diff
@@ -0,0 +1,287 @@
...
+}
+
+pub(super) fn get_user_verification_public_key(
+    clsid: &GUID,
+) -> Result<SigningKey, WinWebAuthnError> {
+    let mut len = 0;
+    let mut data = MaybeUninit::uninit();
+    // SAFETY: We check the OS error code before using the written pointer.
+    unsafe {
+        webauthn_plugin_get_user_verification_public_key(clsid, &mut len, data.as_mut_ptr())?
```
quexten: Question: Does all of this have to be in the unsafe block? Why not just the call to the microsoft API?

apps/desktop/desktop_native/win_webauthn/src/plugin/crypto.rs
```diff
@@ -0,0 +1,287 @@
...
+    ) -> ());
+
+pub(super) fn get_operation_signing_public_key(
+    clsid: &GUID,
+) -> Result<SigningKey, WinWebAuthnError> {
+    let mut len = 0;
+    let mut data = MaybeUninit::uninit();
+    unsafe {
+        // SAFETY: We check the OS error code before using the written pointer.
+        webauthn_plugin_get_operation_signing_public_key(clsid, &mut len, data.as_mut_ptr())?
```
quexten: Question: Does all of this have to be in the unsafe block? Why not just the call to the microsoft API?

apps/desktop/desktop_native/win_webauthn/src/plugin/crypto.rs
```diff
@@ -0,0 +1,287 @@
...
+    }
+}
+
+fn verify_signature(
+    public_key: &SigningKey,
+    hash: &[u8],
+    signature: &[u8],
+) -> Result<(), windows::core::Error> {
+    // Verify the signature over the hash of dataBuffer using the hKey
+    unsafe {
```
quexten: Question: Does all of this have to be in the unsafe block? Why not just the call to the microsoft APIs?

apps/desktop/desktop_native/win_webauthn/src/plugin/crypto.rs
```diff
@@ -0,0 +1,287 @@
...
+
+/// Signing key for an operation request or user verification response buffer.
+pub struct SigningKey {
+    cbPublicKey: u32,
+    pbPublicKey: NonNull<u8>,
+}
+
+impl SigningKey {
+    /// Verifies a signature over some data with the associated public key.
+    pub fn verify_signature(&self, data: &[u8], signature: &[u8]) -> Result<(), WinWebAuthnError> {
```
quexten: As per suggestions above, newtypes would help with type-safety here.

apps/desktop/desktop_native/win_webauthn/src/plugin/crypto.rs
```diff
@@ -0,0 +1,287 @@
...
+        let key_handle = key_handle.assume_init();
+
+        tracing::debug!("Trying to read public key slice as key blob");
+        let public_key = public_key.as_ref();
+        if public_key.len() < size_of::<BCRYPT_KEY_BLOB>() {
+            return Err(windows::core::Error::from_hresult(E_INVALIDARG));
+        }
+        let key_blob: &BCRYPT_KEY_BLOB = &*public_key.as_ptr().cast();
+        tracing::debug!("  got key magic: {}", key_blob.Magic);
+        let (padding_info, cng_flags) = if key_blob.Magic == BCRYPT_RSAPUBLIC_MAGIC.0 {
```
quexten: Could we please put a link to microsoft documentation here? As a reader, I have no idea what this "magic" is..

apps/desktop/desktop_native/win_webauthn/src/plugin/crypto.rs
```diff
@@ -0,0 +1,287 @@
...
+            tracing::debug!("Detected RSA key, adding PKCS1 padding");
+            let padding_info = BCRYPT_PKCS1_PADDING_INFO {
+                pszAlgId: BCRYPT_SHA256_ALGORITHM,
+            };
+            (Some(padding_info), NCRYPT_PAD_PKCS1_FLAG)
+        } else {
+            tracing::debug!("Non-RSA key, no PKCS1 padding added");
```
quexten: Do we know what format they use and what key types they support?

apps/desktop/desktop_native/win_webauthn/src/plugin/crypto.rs
```diff
@@ -0,0 +1,287 @@
...
+                    tracing::error!("Failed to clean up hash object: {err}");
+                }
+            }
+        }
+    }
+}
+
+/// Signing key for an operation request or user verification response buffer.
+pub struct SigningKey {
+    cbPublicKey: u32,
```
quexten:
```suggestion
    // length (count) of the public key byte buffer
    cbPublicKey: u32,
    // pointer to the public key byte buffer
    pbPublicKey: NonNull<u8>
```

apps/desktop/desktop_native/win_webauthn/src/plugin/crypto.rs
```diff
@@ -0,0 +1,287 @@
...
+            )),
+        }
+    }
+}
+
+fn verify_signature(
+    public_key: &SigningKey,
+    hash: &[u8],
+    signature: &[u8],
+) -> Result<(), windows::core::Error> {
```
quexten: Could we provide test coverage, with recorded test vectors?

apps/desktop/desktop_native/win_webauthn/src/plugin/crypto.rs
```diff
@@ -0,0 +1,287 @@
...
+            signature,
+            cng_flags,
+        )?;
+        tracing::debug!("Verified");
+        Ok(())
+    }
+}
+
+pub(super) fn hash_sha256(data: &[u8]) -> Result<Vec<u8>, windows::core::Error> {
+    unsafe {
```
quexten: Could we provide test coverage with recorded test vectors?

quexten (changes requested):
@coltonhurst Thank you for the ping. I took a rough look at only the crypto part of this PR.

In general, KM does allow external crypto implementations - such as was done for chromium ABE - but generally does not provide support / does not maintain them, and expects these use-cases to not be public / necessary for an external integration.

KM mostly manages the crypto of the product, not interactions with other products.

From that perspective this seems fine. KM itself also uses microsft crypto APIs sometimes, such as for memory protection of biometrics.

I'll note that I'm a bit confused as to why we'd need microsofts crypto APIs to be consumable externally. Rustcrypto is MIT licensed, which is much less restrictive than the GPL license we have on this code (on bitwarden/desktop_native), but I'd not block it.

However, I do agree, there is a large amount of unsafe sections here, and generally low unit test coverage, and undocumented / confusing code, which would make me hesitant to submit any changes, should the need arise.

apps/desktop/desktop_native/win_webauthn/src/plugin/crypto.rs
```diff
@@ -0,0 +1,287 @@
+#![allow(non_snake_case)]
```
quexten: Question: What's the expectation for ownership for this file? Autofill? KM?

apps/desktop/desktop_native/win_webauthn/src/lib.rs
```diff
@@ -0,0 +1,72 @@
+#![cfg(target_os = "windows")]
```
quexten, line 7: Question: This lives in the `win_webauthn` module. When would this veer not be on windows?

coltonhurst:
> @abergs @iinuwa I've started to dig into reviewing this today. While this does seem to call the Window's API's in a similar way (the old stuff hasn't been touched in a while and the API's have gone public, so I'm sure there are differences), the generation of the types (and potentially more?) through MIDL is new. That is not necessarily a bad thing; it just has its own pros and cons.
> 
> I will add more complete thoughts as I work through this PR; I also have a meeting with @iinuwa later today. But some very brief initial bullet point thoughts:
> 
>     * Regarding MIDL for type generation:
>       
>       * If we are going to use MIDL to generate types, we become subject to the very unclear naming Microsoft has. This is a con that has more impact than one might initially think, because reading and working on the code becomes much more tedious until someone has enough domain knowledge where they've effectively "memorized" the names. I don't think generating the types via MIDL is bad; it might be nice to have a solution to the naming issue though.
>       * If we want to go the route of generating types with MIDL, we need a documented way to do this, with instructions on what someone updating the code must do to generate these types, and how to add them to the project. This is the bare minimum; ideally this is automated and verified during compile time or checked within CI during changes, etc.
>       * We also will need to have a plan for Windows API changes and how we will handle them across different versions a client might be on (this may be in this PR and I haven't seen it yet, or may be coming in a future pr, or is at least documented somewhere?).
> 
>     * Regarding code organization:
>       
>       * Overall, the code seems to be pretty disorganized. It's lightly organized into modules but even then things are inconsistent. Is this due (in part) to the MIDL generation? Maybe some info there would be helpful. Here are just two examples:
>         
>         * There are API function calls combined with type definitions. It would be much more clear to cleanly abstract these separately, if possible
>         * WebAuthn header file functions and PluginAuthenticator header file functions are both being defined within the types file under the plugin module. Then there are more WebAuthn functions being defined in the types module outside of that. Why are these being mixed?
>       * Unsafe code is sprinkled throughout this code in a _seemingly_ unclear way. This is really not ideal. It may not be possible, but ideally the pattern I was hoping to establish previously was cleanly wrapping unsafe items within safe items (really only types or functions), with no logic in the unsafe layer. Then all business / passkey logic happens in safe layers on top of that. Am I missing how unsafe is being used, besides just as needed?
> 
>     * There are three tests. That may cover everything we can test, I'm not sure yet. Is there more surface area we can cover through more tests?
> 
> 
> I might have my own questions answered as I continue the review, but I wanted to at least share some initial things I see so we can start the conversation about them earlier rather than later. Apologies if some of the things above are coming in future pr's or have been documented somewhere and I've missed them.

@iinuwa and I were able to talk through most of these points. Quick updates:

- I am very happy I misunderstood the MIDL generation comments. Nothing on the Rust side is being generated. Our conversation resolved these concerns, though we did both discuss the importance for version checking. I think we should consider a bit more whether or not that's important to have for v1; I think it could be a good addition.
- I also missed that MS added a new plugin header file, there are two now, which explains the web authn plugin functions inside the plugin module. (I checked the original to verify there weren't changes but missed there was a new one.) Sorry about that!
- I still have some concerns around code organization and the unsafe code; I think the pattern of providing safe functions on top of a type are nice. I will continue investigating this and will have a better response soon.

@iinuwa thanks again for chatting, I will have a more detailed review soon.

apps/desktop/desktop_native/win_webauthn/src/lib.rs
iinuwa: Yes, this will only be on Windows. But if you run `cargo build` from the workspace directory on a non-Windows machine, it will build all workspace members, including this one, and errors out. (It still works if you use the `npm run build-native` script, which is why CI passes.)

To make this less surprising, I added the line to make sure `cargo build` still works.

apps/desktop/desktop_native/win_webauthn/src/plugin/crypto.rs
iinuwa: My expectation is that Autofill would own this.

apps/desktop/desktop_native/win_webauthn/src/plugin/crypto.rs
iinuwa: I've only seen RSA keys come through for operation request and UV response signing, but it is currently undocumented. I'll ask Microsoft for clarification. (Other parts of the code suggest that ECDSA with NIST P-curves might be used.)

apps/desktop/desktop_native/win_webauthn/src/api/plugin/crypto.rs
iinuwa: I'm not sure; it gets cleaned up by `BCryptDestroyHash` when the returned object is dropped, but it is not documented whether the internal object buffer is zeroized.

coltonhurst (changes requested):
@iinuwa okay, another high level review, hopefully this has a bit more pertinent info based on our conversation yesterday.

At a high level, I think we need to figure out a way to handle the unsafe types and the code organization in a cleaner way, if possible. It might be worth tackling that before going into other comments on this review, as those may change based on this. It also might be helpful to break up this pr into smaller pieces; this is something we could pair on, or based on product priority, I can help with.

This is just a brainstorm... hopefully the concept is communicated well. I have been thinking about how we could "safely" abstract away the unsafe code and wonder if something like this is feasible:

```rust
pub struct CtapCborAuthenticatorOptionsV2 {
    raw: Box<webauthn_raw::WEBAUTHN_CTAPCBOR_AUTHENTICATOR_OPTIONS_V2>,
}

impl CtapCborAuthenticatorOptionsV2 {
    pub fn new(
        version: u32,
        user_presence: i32,
        user_verification: i32,
        require_resident_key: bool,
    ) -> Self {
        Self {
            raw: Box::new(webauthn_raw::WEBAUTHN_CTAPCBOR_AUTHENTICATOR_OPTIONS_V2 {
                dwVersion: version,
                lUp: user_presence,
                lUv: user_verification,
                lRequireResidentKey: if require_resident_key { 1 } else { -1 },
            }),
        }
    }

    pub fn version(&self) -> u32 {
        self.raw.dwVersion
    }

    pub fn user_presence(&self) -> i32 {
        self.raw.lUp
    }

    pub fn user_verification(&self) -> i32 {
        self.raw.lUv
    }

    pub fn require_resident_key(&self) -> bool {
        self.raw.lRequireResidentKey == 1
    }

    pub(crate) fn as_raw_ptr(
        &mut self,
    ) -> *mut webauthn_raw::WEBAUTHN_CTAPCBOR_AUTHENTICATOR_OPTIONS_V2 {
        self.raw.as_mut()
    }
}

// A submodule that wouldn't actually be in this file: sys_webauthn_raw module
mod webauthn_raw {
    #[repr(C)]
    #[derive(Debug, Copy, Clone)]
    pub(super) struct WEBAUTHN_CTAPCBOR_AUTHENTICATOR_OPTIONS_V2 {
        pub(super) dwVersion: u32,
        pub(super) lUp: i32,
        pub(super) lUv: i32,
        pub(super) lRequireResidentKey: i32,
    }
}
```

This is a very basic example that takes one of the types and hides it with Box as a pointer. It controls all access to the underlying structure and should (ideally) create an easy way for users to use this type without having to handle unsafe pieces. What I'm trying to shoot for here is handling all type specific unsafe "stuff" within the type itself, and avoid exposing it.

However, please note, just using `Box<T>` may not be feasible, we may need to have something like `Pin<Box<T>>`, etc. to ensure the memory doesn't move or become invalidated when the Windows OS is accessing it.

We could do something like this and create a rough file tree like this:

```
- win_webauthn // the crate
  - lib.rs
  - util.rs
  - error.rs
  - webauthn_sys // only safe windows sys types and functions exposed, no logic here
    - webauthn_raw // raw types and functions are within files under this module, no logic here
```

Anyways - just trying to propose an idea. It might be helpful to pair and just take one slice of this and work on it as a POC together.

Edit: By "It might be worth tackling that before going into other comments on this review, as those may change based on this.", I am referring to other comments I have on this pr, not referencing comments from other developers.

coltonhurst:
One additional comment regarding the crypto side - @quexten @Thomas-Avery if this were to use RustCrypto, because it is key management within passkeys for Bitwarden, is that something KM could provide support for and weigh in on?

Edit: I don't think that the autofill desktop sub-team should own crypto code, that definitely seems out of our domain and out of our area of expertise.

apps/desktop/desktop_native/win_webauthn/src/plugin/crypto.rs
iinuwa: This uses the [BCRYPT API's format for key blobs](https://learn.microsoft.com/en-us/windows/win32/api/bcrypt/ns-bcrypt-bcrypt_key_blob). The BCRYPT API supports several signing key algorithms, including RSA, P-256, P-384, P-512. (Also DSA and some others, but I don't think they're used here.)

It's currently undocumented which algorithms Microsoft uses in the WebAuthn plugin authenticator API. (This is one reason why it's helpful to use the BCRYPT API for signing: we can pass the public key we receive transparently; except for RSA, which requires us to set the padding format.)

I've seen RSA public keys for signing the UV or operation requests, and elsewhere they suggest P-256, P-384, P-512 keys for other operations. I'll see if I can get clearer answers from Microsoft.

For now, I've support implemented RSA, and the three NIST P-curves.

apps/desktop/desktop_native/win_webauthn/src/plugin/crypto.rs
iinuwa: You're right, fixed.

apps/desktop/desktop_native/win_webauthn/src/plugin/crypto.rs
iinuwa: I added a single test using values generated with OpenSSL for each of the algorithms supported: RSA, P-256, P-384 and P-521.

apps/desktop/desktop_native/win_webauthn/src/plugin/crypto.rs
iinuwa: Added more documentation in 07b551fd.

apps/desktop/desktop_native/win_webauthn/src/plugin/crypto.rs
iinuwa: I added a simple SHA-256 test.

iinuwa:
@quexten, thanks for your review!

> I'll note that I'm a bit confused as to why we'd need microsofts crypto APIs to be consumable externally. Rustcrypto is MIT licensed, which is much less restrictive than the GPL license we have on this code (on bitwarden/desktop_native), but I'd not block it.

The considerations I had in mind were:
- the Windows crypto APIs are FIPS-certified, which might matter to some consumers,
- it doesn't require a separate dependency since this is already bound to the Windows platform, and
- the key blobs are already in the format that Windows expects.

These are not important to Bitwarden though, so I'm fine with going a different route. We would just need to parse the public keys and convert it to the format of whatever library we want to use. The recently added tests show the layout for the public keys that we can use to extract the public key data.

> One additional comment regarding the crypto side - \quexten \Thomas-Avery if this were to use RustCrypto, because it is key management within passkeys for Bitwarden, is that something KM could provide support for and weigh in on?

For clarification on the "within passkeys for Bitwarden" part, the Windows crypto is only being used to authenticate requests received by the COM server (consumer's plugin implementation) from the Windows OS. Consumers are free to use separate libraries for passkey crypto.

iinuwa:
> This is a very basic example that takes one of the types and hides it with Box as a pointer. It controls all access to the underlying structure and should (ideally) create an easy way for users to use this type without having to handle unsafe pieces. What I'm trying to shoot for here is handling all type specific unsafe "stuff" within the type itself, and avoid exposing it.

@coltonhurst, this makes sense. I can take a stab at that.

Another thing to note is that the Microsoft [may be releasing types in the `windows-rs` crate](https://github.com/microsoft/windows-rs/issues/3896) which could replace the `webauthn_raw` layer.

apps/desktop/desktop_native/win_webauthn/src/plugin/types.rs
```diff
@@ -0,0 +1,1267 @@
...
+}
+
+impl AsRef<WEBAUTHN_PLUGIN_CANCEL_OPERATION_REQUEST> for PluginCancelOperationRequest {
+    fn as_ref(&self) -> &WEBAUTHN_PLUGIN_CANCEL_OPERATION_REQUEST {
+        // SAFETY: Pointer is received from Windows so we assume it is correct.
+        unsafe { self.inner.as_ref() }
+    }
+}
+
+#[doc(hidden)]
```
neuronull: Is this necessary? We aren't generating rustdocs.

apps/desktop/desktop_native/win_webauthn/src/plugin/types.rs
```diff
@@ -0,0 +1,1267 @@
...
+    }
+}
+
+/// Plugin lock status enum as defined in the IDL
+#[repr(u32)]
+#[derive(Debug, Copy, Clone)]
+pub enum PluginLockStatus {
+    PluginLocked = 0,
+    PluginUnlocked = 1,
+}
```
neuronull: this file is quite large. Is there any way it can be broken down / re-organized?

apps/desktop/desktop_native/win_webauthn/src/plugin/types.rs
```diff
@@ -0,0 +1,1267 @@
...
+#[repr(C)]
+#[derive(Debug, Copy, Clone)]
+pub(super) struct WebAuthnPluginAddAuthenticatorResponse {
+    /// Size in bytes of the public key pointed to by `pbOpSignPubKey`.
+    cbOpSignPubKey: u32,
+    /// Pointer to a [BCRYPT_KEY_BLOB](windows::Win32::Security::Cryptography::BCRYPT_KEY_BLOB).
+    pbOpSignPubKey: *mut u8,
+}
+
+type WEBAUTHN_PLUGIN_ADD_AUTHENTICATOR_RESPONSE = WebAuthnPluginAddAuthenticatorResponse;
```
neuronull: What value does this type alias provide?

apps/desktop/desktop_native/win_webauthn/src/plugin/types.rs
```diff
@@ -0,0 +1,1267 @@
...
+    ppbResponse: *mut *mut u8
+) -> HRESULT);
+
+webauthn_call!("WebAuthNPluginFreeUserVerificationResponse" as
+/// Free a user verification response received from a call to [webauthn_plugin_perform_user_verification].
+fn webauthn_plugin_free_user_verification_response(
+    pbResponse: *mut u8
+) -> ());
+
+// Plugin Authenticator types
```
neuronull: For instance, this comment seems to suggest a logical separation, that could be used to break up the file

apps/desktop/desktop_native/win_webauthn/src/plugin/types.rs
```diff
@@ -0,0 +1,1267 @@
...
+/// - `pbEncoded`: a COM-allocated buffer pointing to a CTAP CBOR make credential request.
+/// - `ppMakeCredentialRequest`: An indirect pointer to a [WEBAUTHN_CTAPCBOR_MAKE_CREDENTIAL_REQUEST].
+/// 
+/// # Safety
+/// - `pbEncoded` must have been allocated by Windows COM.
+/// - `pbEncoded` must be non-null and have the length specified in cbEncoded.
+fn webauthn_decode_make_credential_request(
+    cbEncoded: u32,
+    pbEncoded: *const u8,
+    ppMakeCredentialRequest: *mut *mut WEBAUTHN_CTAPCBOR_MAKE_CREDENTIAL_REQUEST
```
neuronull: what's this double `* mut` for?

apps/desktop/desktop_native/win_webauthn/src/plugin/types.rs
```diff
@@ -0,0 +1,1267 @@
...
+    ///Encoded CBOR attestation information
+    pub attestation_statement: Option<Vec<u8>>,
+
+    // dwAttestationDecodeType: u32,
+    /// Following depends on the dwAttestationDecodeType
+    ///  WEBAUTHN_ATTESTATION_DECODE_NONE
+    ///      NULL - not able to decode the CBOR attestation information
+    ///  WEBAUTHN_ATTESTATION_DECODE_COMMON
+    ///      PWEBAUTHN_COMMON_ATTESTATION;
+    // pub pvAttestationDecode: *mut u8,
```
neuronull: delete?

apps/desktop/desktop_native/win_webauthn/src/plugin/types.rs
```diff
@@ -0,0 +1,1267 @@
...
+
+    //
+    // Following fields have been added in WEBAUTHN_CREDENTIAL_ATTESTATION_VERSION_5
+    pub prf_enabled: bool,
+
+    //
+    // Following fields have been added in WEBAUTHN_CREDENTIAL_ATTESTATION_VERSION_6
+    pub unsigned_extension_outputs: Option<Vec<u8>>,
+
+    //
```
neuronull: nit: the documentation readability and value here is lacking. What does "following fields have been added in <some version>" mean to us? And the comment is there for each field, specifying "fields" plural.

apps/desktop/desktop_native/win_webauthn/src/plugin/types.rs
```diff
@@ -0,0 +1,1267 @@
...
+    pcbResp: *mut u32,
+    ppbResponse: *mut *mut u8
+) -> HRESULT);
+
+// GetAssertion types
+
+#[repr(C)]
+pub(super) struct WEBAUTHN_CTAPCBOR_ECC_PUBLIC_KEY {
+    /// Version of this structure, to allow for modifications in the future.
+    pub _dwVersion: u32,
```
neuronull: How are we leveraging this version field / what is our strategy for backwards and forwards compatibility?

apps/desktop/desktop_native/win_webauthn/src/plugin/types.rs
```diff
@@ -0,0 +1,1267 @@
...
+    }
+
+    pub fn allow_credentials(&self) -> impl Iterator<Item = CredentialEx<'_>> {
+        // SAFETY: When this is constructed from Self::try_from_ptr(), the Windows decode API
+        // constructs valid pointers.
+        unsafe { self.as_ref().CredentialList.iter() }
+    }
+
+    // TODO: Support extensions
+    // pub fn extensions(&self) -> Options<Extensions> {}
```
neuronull: Is there a ticket created for this work? I like to see the ticket number in such comments.

apps/desktop/desktop_native/win_webauthn/src/lib.rs
```diff
@@ -0,0 +1,72 @@
...
+
+use std::{error::Error, fmt::Display};
+
+pub use types::{
+    AuthenticatorInfo, CredentialId, CtapTransport, CtapVersion, PublicKeyCredentialParameters,
+    UserId,
+};
+
+#[derive(Debug)]
+pub struct WinWebAuthnError {
```
neuronull, line 22: nit: add doc for pub

apps/desktop/desktop_native/win_webauthn/src/lib.rs
```diff
@@ -0,0 +1,72 @@
...
+    WindowsInternal,
+}
+
+impl Display for WinWebAuthnError {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        let msg = match self.kind {
+            ErrorKind::Serialization => "Failed to serialize data",
+            ErrorKind::DllLoad => "Failed to load function from DLL",
+            ErrorKind::InvalidArguments => "Invalid arguments passed to function",
+            ErrorKind::Other => "An error occurred",
```
neuronull, line 75: Is it possible this could be more specific? Like, does it happen due to a bug in our own logic, perhaps "InternalError" or something. "An error occurred" isn't all that helpful :)

apps/desktop/desktop_native/win_webauthn/src/api/sys/util.rs
```diff
@@ -0,0 +1,100 @@
...
+                let function: unsafe extern "C" fn(
+                    $($arg: $arg_type),*
+                ) -> $result_type = std::mem::transmute_copy(&address);
+                function($($arg),*)
+            };
+            crate::util::free_webauthn_lib(library)?;
+            Ok(response)
+        }
+    )
+}
```
neuronull, line 55: At a minumum, IMO all macros should be thoroughly documented. There is a cost to adding macros to our code, which is that they can be ignored and fall out of correctness if not thoroughly tested. Documenting the macro helps with that. It'd be nice to include a sample code block of what is generated from this macro.

apps/desktop/desktop_native/win_webauthn/src/lib.rs
```diff
@@ -0,0 +1,72 @@
...
+pub use types::{
+    AuthenticatorInfo, CredentialId, CtapTransport, CtapVersion, PublicKeyCredentialParameters,
+    UserId,
+};
+
+#[derive(Debug)]
+pub struct WinWebAuthnError {
+    kind: ErrorKind,
+    description: Option<String>,
+    cause: Option<Box<dyn std::error::Error>>,
```
neuronull, line 25: Why does this have to be `Box` ?  That's a heap allocation.

neuronull (commented):
Leaving some drive-by review comments. I don't have as much context on this feature so that can can be both helpful and unhelpful.

In general I'm concerned about the amount of and complexity of the code, and the unit test coverage level. While there are some tests, AFAICT much of this is untested.

Can you comment on the debuggability? 

One thing to consider with the windows API boundaries and `unsafe`, is to take an approach similar to what we have in autotype: https://github.com/bitwarden/clients/tree/main/apps/desktop/desktop_native/autotype/src/windows

, in there, we have trait definitions to abstract the win32 api calls. And a trait implementation which constrains the unsafe calls. This allows mocking the win32 calls, which in turn allows unit testing the business logic.

apps/desktop/desktop_native/win_webauthn/src/plugin/types.rs
abergs:
```suggestion
    // TODO: PM-30510 Support extensions
    // pub fn extensions(&self) -> Options<Extensions> {}
```

apps/desktop/desktop_native/win_webauthn/src/plugin/crypto.rs
neuronull: 💭  Is there cryptographic operations in this module (or PR), that are generalizable enough that they would apply for non win-webathn code? If the answer is yes, to me that points to a need to make it library code that is in turn used here, and ostensibly KM would own the library code?

To be clear- I'm not sure if this code would qualify for that, but I hope that we would not have duplicate implementations/functionalities of cryptographic operations in various feature specific crates.

apps/desktop/desktop_native/win_webauthn/src/lib.rs
neuronull: To provide more actionable feedback, it's generally good practice to use an enum for error types. I'm not following why this struct exists 🤔 why not have a `WinWebAuthnError` that is an enum, and use #[error("The error description")] macro to provide description for it? The "cause" member here is also not clear what benefit it provides. What is different about "description", from "cause"? Generally, the text representation of an error provides the context needed. I'm not seeing why two fields are necessary 🤔

apps/desktop/desktop_native/win_webauthn/src/plugin/types.rs
iinuwa: I'd rather leave the comment in to make it clear that we're skipping the field defined in webauthnplugin.h, if that's OK with you. I can update to make the intention of the comment clearer. (`cargo fmt` doesn't do what I want though, it keeps trying to bind to the next field...)

apps/desktop/desktop_native/win_webauthn/src/plugin/types.rs
iinuwa: This is a direct copy from the header file. Unfortunately, `cargo fmt` is doing weird things and binding the comment to the field instead of adding space to denote a section, so it's much less readable here.

Basically what the Microsoft docs are saying is that when `*_VERSION_X` comment is present, all of the following fields are present in `*_VERSION_X` of the field and all future versions, until a `*_VERSION_Y` comment appears, and so on.

I can copy the version text to all applicable fields to make it easier to read in Rust.

apps/desktop/desktop_native/win_webauthn/src/plugin/types.rs
iinuwa: Right now, we are only using GA features, so all defined fields are available for all structs. This fits for Bitwarden's purposes, but if users wanted to use this, for example to write a Windows Hello client on Windows 10, then that wouldn't work. I think it's fine to consider that out of scope.

I think that moving forward, for reading versioned fields, we should make them an `Option` and check them before returning to the caller. E.g., if this is the definition in C:

```c
typdef struct WEBAUTHN_FOO {
  // Version of this structure to allow for modifications in the future.
  DWORD dwVersion;
  
  DWORD dwExistingField;

  // Following fields have been added in WEBAUTHN_FOO_VERSION_2

  DWORD dwSomeNewField
} WEBAUTHN_FOO;
```

, then this is some shorthand for how we'd reference in Rust, ignoring conversations about pointer wrapping for now):

```rust
#[repr(C)]
pub struct WEBAUTHN_FOO {
    dwVersion: u32,
    dwExistingField: u32,
    dwSomeNewField: u32,
}

impl WEBAUTHN_FOO {
    pub fn version(&self)
    pub fn dwExistingField(&self)
    /// Some new field.
    ///
    /// This field was added in version 2 of this struct and will be `None` when running on older OSes that do not support that field.
    pub fn dwSomeNewField(&self) -> Option<u32> {
      if self.dwVersion >= 2 {
          Some(self.dwSomeNewField)
      } else {
          None
      }
    }
}
```

For writing, we can always set the latest version number we know of for the struct, and the OS will do a similar negotation on its side. (It's safe to send a larger struct than the OS expects as long as the previous fields share the same order/ABI.)

apps/desktop/desktop_native/win_webauthn/src/plugin/types.rs
neuronull: ah 🤔  what about using an html link at the top of the struct, to a pinned sha of the header file? I wonder the benefit of duplicating the documentation here. Similar to what we did in autotype (search for "microsoft" in this file: https://github.com/bitwarden/clients/blob/main/apps/desktop/desktop_native/autotype/src/windows/type_input.rs)

apps/desktop/desktop_native/win_webauthn/src/plugin/types.rs
iinuwa: It's a pointer to a pointer. It's a common pattern in C-style APIs where data is allocated inside of the method (which is written to the "inner" pointer), and the caller can lookup where that data is allocated by dereferencing the "outer" pointer.

This is done because only one value can be returned from a C type, which is typically the result/error code, and the pointer becomes the out variable. In other words, 
```c
int call_that_can_fail(**foo_t);

foo_t *foo;
if (call_that_can_fail(&foo) != 0) {
    goto err;
}
// error code checked, so we assume foo is initialized.
do_something(foo->blah);
```
is an analogous pattern to:

```rust
fn call_that_can_fail() -> Result<Foo, Error>;

let Ok(foo) = call_that_can_fail() else {
    // process error.
}
do_something(foo.bar)
```

apps/desktop/desktop_native/win_webauthn/src/plugin/types.rs
neuronull: oh, yeah I didn't realize these were skipped intentionally. 👍  to making the comment clearer.

idea: could list the fields that are skipped, in the struct-level documentation. It would also be nice to state why we are skipping.

apps/desktop/desktop_native/win_webauthn/src/lib.rs
iinuwa: Passing the trait impl allows to abstract over various different error types, whether another instance of WinWebAuthnError, or other errors like `windows::core::Error`, `std::io::Error`, etc. We can then use this to print the error messages using their implementation of `std::error::Error`.

This has to be `Box`ed because it's a trait object, which require a reference since it's not `Sized`. But since we want to return this trait object to back to callers (so that consumers can print the nested error messages), we can't have a normal `&` reference since the lifetime doesn't live long enough, so we heap-allocate it.

(This is the same pattern as what `anyhow` and other similar libraries do.)

apps/desktop/desktop_native/win_webauthn/src/plugin/types.rs
iinuwa: True, but if a developer does generate the rustdocs locally, it hides this implementation which isn't intended to be used by consumers.

apps/desktop/desktop_native/win_webauthn/src/plugin/types.rs
iinuwa: The order is intended to match the C struct, whose ABI makes field order significant, so I think it's still useful to have it inline, but we could also put it in the struct docs.

apps/desktop/desktop_native/win_webauthn/src/plugin/types.rs
iinuwa: Copying the documentation inline has been helpful to me while working on this, but I'm fine either way, let me know which you prefer.

apps/desktop/desktop_native/win_webauthn/src/lib.rs
iinuwa: The `description` is written by the library, which tells what we were trying to do that failed, and `cause` is the underlying error (if any) that caused that error. For example, if we were trying to verify the signature of an OS request and it failed, we'd have context in the log that the signature verification failed, and the exact error that the crypto library returned (e.g. unsupported crypto algorithm vs invalid key format.)

Keeping `cause` allows us to format the underlying error message in one place (within the struct impl) instead of everywhere we construct a new error, which I think saves readability by reducing duplication. This follows the same pattern as `std::io::Error`, which has an `ErrorKind` enum, but also allows you to attach another `std::error::Error` to it.

The errors we can receive from Windows are not exhaustive, so it's helpful to preserve them as much as possible; this `cause` field just makes that a little less arduous to do.

apps/desktop/desktop_native/win_webauthn/src/plugin/crypto.rs
iinuwa: No, this is a wrapper around a handful of Windows crypto primitives (RSA verify, ECDSA verify and SHA-256) needed to validate the requests from the OS to the passkey plugin's COM server so that the plugin implementer doesn't have to do it.

It's intentionally specific to WebAuthn plugin usage (the drop implementations come from `webauthn.dll` since the public keys are received from functions also defined there.) Because it's used in this very limited fashion, it is not exposed outside of this crate, and I don't think it is suitable for a generic crypto library.

All passkey crypto will need to be provided by the consumer.

iinuwa:
Closing this out in favor of a separate patch series.

PR #18075, opened 2025-12-20
Add generic autofill provider library [PM-29786]

iinuwa:
## 🎟️ Tracking

[PM-29786](https://bitwarden.atlassian.net/browse/PM-29786)

## 📔 Objective

The `macos_provider` library provides an IPC client for the macOS autofill extension to interact with Bitwarden. This framework is useful on other OSes, so this PR makes this generic.

It also adds methods required for the Windows implementation.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

[PM-29786]: https://bitwarden.atlassian.net/browse/PM-29786?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/desktop/desktop_native/autofill_provider/src/lib.rs
+++ b/apps/desktop/desktop_native/autofill_provider/src/lib.rs
@@ -712,7 +712,7 @@ mod tests {
     }
 
     #[test]
-    fn test_disconnected_gives_error() {
+    fn test_client_throws_error_on_method_call_when_disconnected() {
         // There is no server running at this path, so this client should always be disconnected.
         let client = AutofillProviderClient::connect_to_path(get_server_path());
 
@@ -727,7 +727,7 @@ mod tests {
     }
 
     #[test]
-    fn test_get_lock_status() {
+    fn test_client_parses_get_lock_status_response_when_valid_json_is_returned() {
         // The server should expect a lock status request and return a valid response.
         let handler = |value: Result<Value, BitwardenError>| {
             let value = value.unwrap();


apps/desktop/desktop_native/autofill_provider/src/lock_status.rs
```diff
@@ -0,0 +1,41 @@
...
+    fn on_error(&self, error: BitwardenError);
+}
+
+impl GetLockStatusCallback for TimedCallback<LockStatusResponse> {
+    fn on_complete(&self, response: LockStatusResponse) {
+        self.send(Ok(response));
+    }
+
+    fn on_error(&self, error: BitwardenError) {
+        self.send(Err(error))
```
neuronull:
```suggestion
        self.send(Err(error));
```

apps/desktop/desktop_native/autofill_provider/src/util.rs
```diff
@@ -0,0 +1,24 @@
+use serde::{de::Visitor, Deserializer};
+
+pub(crate) fn deserialize_b64<'de, D: Deserializer<'de>>(
+    deserializer: D,
+) -> Result<Vec<u8>, D::Error> {
+    deserializer.deserialize_str(Base64Visitor {})
+}
+
+struct Base64Visitor;
+impl<'de> Visitor<'de> for Base64Visitor {
```
neuronull: can be elided


```suggestion
impl Visitor<'_> for Base64Visitor {
```

apps/desktop/desktop_native/autofill_provider/src/window_handle_query.rs
```diff
@@ -0,0 +1,47 @@
...
+    fn on_error(&self, error: BitwardenError);
+}
+
+impl GetWindowHandleQueryCallback for TimedCallback<WindowHandleQueryResponse> {
+    fn on_complete(&self, response: WindowHandleQueryResponse) {
+        self.send(Ok(response));
+    }
+
+    fn on_error(&self, error: BitwardenError) {
+        self.send(Err(error))
```
neuronull:
```suggestion
        self.send(Err(error));
```

apps/desktop/desktop_native/autofill_provider/src/lib.rs
```diff
@@ -271,13 +322,24 @@ impl MacOSProviderClient {
...
+            let message = SerializedMessage::Message {
+                sequence_number,
+                value: Ok(value),
+            };
+            let json = serde_json::to_string(&message)
+                .map_err(|err| format!("Could not serialize message as JSON: {err}"))?;
+            tx.blocking_send(json)
+                .map_err(|err| format!("Error sending message: {err}"))?;
+            Ok(())
+        }
```
neuronull: It'd be easier to follow if extract this to helper function.

apps/desktop/desktop_native/autofill_provider/src/lib.rs
```diff
@@ -294,3 +356,110 @@ impl MacOSProviderClient {
...
+            Err(RecvTimeoutError::Timeout) => Err(CallbackError::Timeout),
+            Err(_) => Err(CallbackError::Cancelled),
+        }
+    }
+
+    fn send(&self, response: Result<T, BitwardenError>) {
+        match self.tx.lock().expect("not poisoned").take() {
+            Some(tx) => {
+                if tx.send(response).is_err() {
+                    tracing::error!("Windows provider channel closed before receiving IPC response from Electron")
```
neuronull:
```suggestion
                    tracing::error!("Windows provider channel closed before receiving IPC response from Electron");
```

apps/desktop/desktop_native/autofill_provider/src/lib.rs
```diff
@@ -294,3 +356,110 @@ impl MacOSProviderClient {
...
+    }
+}
+
+impl PreparePasskeyRegistrationCallback for TimedCallback<PasskeyRegistrationResponse> {
+    fn on_complete(&self, credential: PasskeyRegistrationResponse) {
+        self.send(Ok(credential));
+    }
+
+    fn on_error(&self, error: BitwardenError) {
+        self.send(Err(error))
```
neuronull:
```suggestion
        self.send(Err(error));
```

apps/desktop/desktop_native/autofill_provider/src/lock_status.rs
```diff
@@ -0,0 +1,41 @@
+use std::sync::Arc;
+
+use serde::{Deserialize, Serialize};
+
+use crate::{BitwardenError, Callback, TimedCallback};
+
+#[derive(Debug, Serialize, Deserialize)]
+pub(super) struct LockStatusRequest {}
```
neuronull, line 9: blanket comment for all `pub` items in this code review: please add doc string comments for everything that is `pub` , including the trait function definitions.

apps/desktop/desktop_native/autofill_provider/src/registration.rs
```diff
@@ -4,29 +4,35 @@ use serde::{Deserialize, Serialize};
...
 pub struct PasskeyRegistrationRequest {
-    rp_id: String,
-    user_name: String,
-    user_handle: Vec<u8>,
-    client_data_hash: Vec<u8>,
-    user_verification: UserVerification,
-    supported_algorithms: Vec<i32>,
-    window_xy: Position,
-    excluded_credentials: Vec<Vec<u8>>,
+    pub rp_id: String,
```
neuronull, line 13: I recognize the existing struct has this scope visibility but wondering if these should instead be `pub(super)` or at least `pub(crate)`

apps/desktop/desktop_native/autofill_provider/src/assertion.rs
```diff
@@ -0,0 +1,85 @@
...
+    pub rp_id: String,
+    pub client_data_hash: Vec<u8>,
+    pub user_verification: UserVerification,
+    pub allowed_credentials: Vec<Vec<u8>>,
+    pub window_xy: Position,
+    #[cfg(not(target_os = "macos"))]
+    pub client_window_handle: Vec<u8>,
+    #[cfg(not(target_os = "macos"))]
+    pub context: String,
+    // pub extension_input: Vec<u8>, TODO: Implement support for extensions
```
neuronull: curious if we have this tracked somewhere in Jira (cc @coltonhurst )

apps/desktop/desktop_native/autofill_provider/src/lib.rs
```diff
@@ -1,54 +1,87 @@
...
     pub x: i32,
     pub y: i32,
 }
 
-#[derive(Debug, uniffi::Error, Serialize, Deserialize)]
+#[cfg_attr(target_os = "macos", derive(uniffi::Error))]
+#[derive(Debug, Serialize, Deserialize)]
 pub enum BitwardenError {
     Internal(String),
 }
```
neuronull, line 76: do we (again cc @coltonhurst ) have explicit plans to expand this error type enum? If not, we probably don't need it? Unless I'm missing something.

apps/desktop/desktop_native/autofill_provider/src/lib.rs
```diff
@@ -112,10 +163,12 @@ impl MacOSProviderClient {
                 .init();
         });
 
+        tracing::debug!("Autofill provider attempting to connect to Electron IPC...");
```
neuronull: How often does this `connect()` get called? (wondering if this is a candidate for `trace` level)

apps/desktop/desktop_native/autofill_provider/src/lib.rs
```diff
@@ -271,13 +322,24 @@ impl MacOSProviderClient {
...
+            let value = serde_json::to_value(message)
+                .map_err(|err| format!("Could not represent message as JSON: {err}"))?;
+            let message = SerializedMessage::Message {
+                sequence_number,
+                value: Ok(value),
+            };
+            let json = serde_json::to_string(&message)
+                .map_err(|err| format!("Could not serialize message as JSON: {err}"))?;
+            tx.blocking_send(json)
+                .map_err(|err| format!("Error sending message: {err}"))?;
```
neuronull: 💭 should this be spawned in a thread?

apps/desktop/desktop_native/autofill_provider/src/lib.rs
```diff
@@ -294,3 +356,110 @@ impl MacOSProviderClient {
...
+            }
+        });
+        match rx.recv_timeout(timeout) {
+            Ok(Ok(response)) => Ok(response),
+            Ok(err @ Err(CallbackError::Cancelled)) => {
+                tracing::debug!("Received cancellation, dropping.");
+                err
+            }
+            Ok(err @ Err(CallbackError::Timeout)) => {
+                tracing::debug!("Request timed out, dropping.");
```
neuronull: ?


```suggestion
                tracing::warn!("Request timed out, dropping.");
```

apps/desktop/desktop_native/autofill_provider/src/registration.rs
iinuwa: These did not need to be `pub` previously, since UniFFI was still able to generate the structs for consumption in Swift bindings for the private fields. But in the Windows implementation, we use the Rust struct directly, so these fields need to be visible to the consumer.

apps/desktop/desktop_native/autofill_provider/src/lib.rs
iinuwa: This code predates me. I don't think it would hurt to run this in a thread, but I don't think it's necessary. This will only fail to immediately return if we have more than 32 concurrent requests to the desktop app. Given that Windows and macOS will only call our native providers serially, and we only need 1-3 calls to the desktop app to fulfill an OS request, we shouldn't realistically ever block the native provider.

apps/desktop/desktop_native/autofill_provider/src/lock_status.rs
iinuwa: Done. Also added some doc examples and some basic tests to demonstrate usage

apps/desktop/desktop_native/autofill_provider/src/lib.rs
iinuwa: I just added a `Disconnected` variant in case callers want to react to it by attempting to reconnect. I also set this as an early return on the callback, so callers don't wait unnecessarily for an error we can already detect.

I believe that the enum type primarily exists for UniFFI purposes though, so that you can easily do `ctx.cancelRequest(withError: BitwardenError.Internal("some error")`, as UniFFI implements the `Error` protocol that Swift is looking for there.

apps/desktop/desktop_native/autofill_provider/src/lib.rs
iinuwa: Roughly, it gets called up to 20 times (averaging at 3 times) in a retry loop per passkey ceremony.

I agree that `trace` is a good designation for this. Ironically, I wasn't aware of that log level in the `tracing` crate 😆.

neuronull: > I agree that trace is a good designation for this. Ironically, I wasn't aware of that log level in the tracing crate 😆.

How serendipitous !  Just the other day I added this documentation

https://bitwarden.atlassian.net/wiki/spaces/EN/pages/2065858607/Desktop+Native+Observability#Appropriate-use-of-log-levels

apps/desktop/desktop_native/autofill_provider/src/assertion.rs
iinuwa: I didn't see one, so I created [PM-30510](https://bitwarden.atlassian.net/browse/PM-30510).

apps/desktop/desktop_native/autofill_provider/src/assertion.rs
```diff
@@ -0,0 +1,184 @@
...
+
+#[cfg(not(target_os = "macos"))]
+use crate::TimedCallback;
+use crate::{BitwardenError, Callback, Position, UserVerification};
+
+#[cfg_attr(target_os = "macos", derive(uniffi::Record))]
+#[derive(Debug, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+/// Request to assert a credential.
+pub struct PasskeyAssertionRequest {
```
coltonhurst, line 13: The following fields are shared between `PasskeyAssertionRequest` and `PasskeyAssertionWithoutUserInterfaceRequest`:
- `rp_id`
- `client_data_hash`
- `user_verification`
- `window_xy`
- `client_window_handle`
- `context`

(There are similarly shared fields in other types in the PR.)

---

Bringing this up as a discussion point, not to be done in or to block this PR though:

Do we want to leverage Rust enum types here and combine these into just one `PasskeyAssertionRequest`? Or pull these out into their own type and then use them within both? That might be a way to more elegantly represent the request.

However, I'm open to _not_ doing this, especially if this doesn't match the spec or creates problems with uniffi.

apps/desktop/desktop_native/autofill_provider/src/lib.rs
```diff
@@ -0,0 +1,705 @@
...
+
+impl Error for BitwardenError {}
+
+// TODO: These have to be named differently than the actual Uniffi traits otherwise
+// the generated code will lead to ambiguous trait implementations
+// These are only used internally, so it doesn't matter that much
+trait Callback: Send + Sync {
+    fn complete(&self, credential: serde_json::Value) -> Result<(), serde_json::Error>;
+    fn error(&self, error: BitwardenError);
+}
```
coltonhurst: Do we have a ticket for this todo?

apps/desktop/desktop_native/autofill_provider/src/lib.rs
```diff
@@ -0,0 +1,705 @@
...
+
+#[derive(Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+/// Store native desktop status information to use for IPC communication
+/// between the application and the credential provider.
+pub struct NativeStatus {
+    key: String,
+    value: String,
+}
```
coltonhurst: I believe (though I could be wrong and [this](https://github.com/rust-lang/style-team/issues/72#issuecomment-300026781) may be outdated) comments should come first, then attributes, and then the function. Could you please fix this ordering throughout the PR? It's a bit weird to read when (if the comments are large) a reader has to scroll down to determine which function a set of attributes applies to.

Example fix for the above:

```suggestion
/// Store native desktop status information to use for IPC communication
/// between the application and the credential provider.
#[derive(Serialize, Deserialize)]
#[serde(rename_all = "camelCase")]
pub struct NativeStatus {
    key: String,
    value: String,
}
```

apps/desktop/desktop_native/autofill_provider/src/lib.rs
```diff
@@ -0,0 +1,705 @@
...
+        let client = get_client(handler);
+        let callback = Arc::new(TimedCallback::new());
+        client.get_lock_status(callback.clone());
+        let response = callback
+            .wait_for_response(Duration::from_millis(3000), None)
+            .unwrap()
+            .unwrap();
+        assert!(response.is_unlocked);
+    }
+}
```
coltonhurst, line 754: I'm glad we have tests! Thank you! 😄 

However - these tests contain a lot of code where there is no separation at all. There are also no comments explaining what is happening throughout the test. This makes the tests more fragile - it is harder to come back and edit or update the test.

It would be great if these could have comments and be broken into "sections" with newlines.

apps/desktop/desktop_native/autofill_provider/src/lib.rs
```diff
@@ -0,0 +1,705 @@
...
+        client.get_lock_status(callback.clone());
+        let response = callback
+            .wait_for_response(Duration::from_millis(10), None)
+            .unwrap();
+        assert!(matches!(response, Err(BitwardenError::Disconnected)));
+    }
+
+    #[test]
+    fn test_get_lock_status() {
+        // tracing_subscriber::fmt().init();
```
coltonhurst: Do we need this?

apps/desktop/desktop_native/autofill_provider/src/window_handle_query.rs
```diff
@@ -0,0 +1,73 @@
...
+    #[serde(rename = "windowHandle")]
+    /// base64-encoded byte string representing native window handle.
+    /// # Operating System Differences
+    ///
+    /// ## macOS
+    /// Unused.
+    ///
+    /// ## Windows
+    /// On Windows, this is an HWND.
+    window_handle: String,
```
coltonhurst, line 20: I could be misremembering, isn't `HWND` a numerical type under the hood? And in the windows crate it's effectively a pointer?

Why is it a `String` here, and below in the Response it's a `Vec<u8>`?

apps/desktop/desktop_native/autofill_provider/src/util.rs
```diff
@@ -0,0 +1,24 @@
...
+
+    fn expecting(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
+        f.write_str("A valid base64 string")
+    }
+
+    fn visit_str<E>(self, v: &str) -> Result<Self::Value, E>
+    where
+        E: serde::de::Error,
+    {
+        use base64::{engine::general_purpose::STANDARD, Engine as _};
```
coltonhurst: Just curious, any reason this isn't at the top of the file?

coltonhurst (changes requested):
Thanks for the work @iinuwa!

apps/desktop/desktop_native/autofill_provider/src/assertion.rs
neuronull: Serde's `tag` and `flatten` can be used.

apps/desktop/desktop_native/autofill_provider/src/lib.rs
```diff
@@ -0,0 +1,705 @@
...
+        Self {
+            tx: Arc::new(Mutex::new(Some(tx))),
+            rx: Arc::new(Mutex::new(rx)),
+        }
+    }
+
+    /// Block the current thread until either a response is received, or the
+    /// specified timeout has passed.
+    ///
+    /// # Usage
```
neuronull: This goes for other parts of this PR as well- does rust's doc tests pay attention to `Usage` as a key word? I'm used to seeing `# Examples`

apps/desktop/desktop_native/autofill_provider/src/lib.rs
```diff
@@ -0,0 +1,705 @@
...
+        std::thread::spawn(move || {
+            let _span = tracing::span!(Level::DEBUG, "server").entered();
+            tracing::info!("[server] Starting server thread");
+            let (tx, mut rx) = mpsc::channel(8);
+            let rt = tokio::runtime::Builder::new_current_thread()
+                .enable_io()
+                .build()
+                .unwrap();
+            rt.block_on(async move {
+                tracing::debug!("[server] starting server at {server_path:?}");
```
neuronull: I know it's a unit test but using inline scoping with "[server/client]" and variable printing with fmt is doing a great dis-service to `tracing`, those could be `println` statements.

If you want to use tracing, checkout 

https://bitwarden.atlassian.net/wiki/spaces/EN/pages/2065858607/Desktop+Native+Observability#Tracing-usage-guide

and also https://bitwarden.atlassian.net/wiki/spaces/EN/pages/2065858607/Desktop+Native+Observability#Async

(which I'm in the process of porting to contributing docs)

apps/desktop/desktop_native/autofill_provider/src/util.rs
```diff
@@ -0,0 +1,24 @@
+use serde::{de::Visitor, Deserializer};
+
+pub(crate) fn deserialize_b64<'de, D: Deserializer<'de>>(
```
neuronull: Did you consider `serde_with` as an alternative?

https://docs.rs/serde_with/latest/serde_with/base64/struct.Base64.html

apps/desktop/desktop_native/autofill_provider/src/lib.rs
iinuwa: I haven't seen one, but this predates me. I'm also not exactly sure what this is referring to, since this works, but perhaps @dani-garcia remembers?

apps/desktop/desktop_native/autofill_provider/src/assertion.rs
iinuwa: I like this idea for a follow up PR. I did a similar experiment on more strict types [here](https://github.com/bitwarden/clients/pull/18324/files#diff-5e7218866cd2adaf836c85cb11ad80da76ec837d8be6df8188b1d9ffafcf45b1R268-R344). I'd first like to address (in another PR) sharing the types between this and the `desktop_napi` crates, since it is very annoying to try to align the types and serde attributes with copy-paste.

apps/desktop/desktop_native/autofill_provider/src/window_handle_query.rs
iinuwa: > Why is it a String here, and below in the Response it's a Vec<u8>?

Actually my docs are wrong here. the `window_handle` field is not used for data; it is only used to disambiguate parsing in `desktop_napi`.

This parsing issue is something I'd like to address in https://bitwarden.atlassian.net/browse/PM-23485. I've updated the docs with a reference to  that ticket.

> I could be misremembering, isn't HWND a numerical type under the hood? And in the windows crate it's effectively a pointer?

(I think my goofup caused this confusion, so I think this is just informational.) Yeah, HWND is a handle, which is represented as a pointer and can be serialized as a 32-bit integer when passed between processes. (I use a `Vec<u8>` since this can differ between platforms.)

apps/desktop/desktop_native/autofill_provider/src/util.rs
iinuwa: No reason, removed

apps/desktop/desktop_native/autofill_provider/src/util.rs
iinuwa: I did not, but for something so small, I didn't think that it'd be worth adding another dependency since we're not using it anywhere else.

Would you prefer adding that dependency?

apps/desktop/desktop_native/autofill_provider/src/lib.rs
dani-garcia: This is referring to the fact that the UniFFI trait has methods called `on_complete` and `on_error`:

https://github.com/bitwarden/clients/blob/622df1deaf26235c9325ded692b0321374dd483d/apps/desktop/desktop_native/macos_provider/src/assertion.rs#L42-L58

Initially I've wanted to name them the same, but was hitting some errors with the uniffi macro and the methods being ambiguous if they both have the `on_` prefix, so I renamed the ones in `Callback` to remove the prefix.

Not really a big problem, the comment is just there to point why the naming difference exists. We can safely remove the TODO itself, as it's not really actionable.

apps/desktop/desktop_native/autofill_provider/src/lib.rs
iinuwa: Yeah, I added these for developers because it's tricky to follow the context across the callbacks, but following these logs makes it clear. But these could use more effort.

These could definitely be `println`s, but `println` is banned by Clippy. This could be overridden, but I think there are some advantages to using tracing for developers working on these tests, since you get these for free when running `RUST_LOG=debug cargo test`:
- easy, uniform formatting of timestamps and messages
- traces from the library itself

That being said, I did set the server span, but I forgot to remove the prefixes, my bad.

Which would you prefer?
- removing the logs altogether
- removing the prefixes, but keeping `tracing`
- converting to `println`

apps/desktop/desktop_native/autofill_provider/src/lib.rs
iinuwa: Neither `Usage` nor `Examples` are keywords, but `Examples` is a convention in the stdlib docs. I updated this one; the other instance of `# Usage` also has an `# Examples` section.

apps/desktop/desktop_native/autofill_provider/src/lib.rs
iinuwa: No, I had kept it in as a hint that developers can enable tracing to debug, but I can remove it.

apps/desktop/desktop_native/autofill_provider/src/lib.rs
iinuwa: Thank you for pointing me to that convention. I actually had it the other way originally, and then changed it. Fixed!

apps/desktop/desktop_native/autofill_provider/src/lib.rs
iinuwa: Thanks! I'll update the comment

apps/desktop/desktop_native/autofill_provider/src/assertion.rs
coltonhurst: @iinuwa Sounds good, no problem. Do you mind creating a ticket and sharing it here, just for posterity?

apps/desktop/desktop_native/autofill_provider/src/lib.rs
coltonhurst: Ah, in that case feel free to keep it! You can add a note to the beginning just so readers know what it's for, maybe? (Just an idea, up to you)

Example: `// Add this line if you want to enable tracing for debugging: tracing_subscriber::fmt().init();`

apps/desktop/desktop_native/autofill_provider/src/assertion.rs
iinuwa: Yeah, added as a comment to the same ticket mentioned further below: https://bitwarden.atlassian.net/browse/PM-23485?focusedCommentId=125691

apps/desktop/desktop_native/autofill_provider/src/assertion.rs
neuronull: Thanks, would you mind putting the ticket ID there with the TODO comment?

apps/desktop/desktop_native/autofill_provider/src/lib.rs
neuronull: > This will only fail to immediately return if we have more than 32 concurrent requests to the desktop app. Given that Windows and macOS will only call our native providers serially, and we only need 1-3 calls to the desktop app to fulfill an OS request, we shouldn't realistically ever block the native provider.

Is this (or some version of it) knowledge documented in the code anywhere? If not, it could be helpful to add it somewhere.

apps/desktop/desktop_native/autofill_provider/src/lib.rs
neuronull: Since this is a unit test, is there any reason _not_ to enable it? unit tests executed via cargo test don't give stdout by default.

apps/desktop/desktop_native/autofill_provider/src/lib.rs
neuronull: Ah sorry for not clarifying this, by "these could be printlns" , mainly meant to indicate that the prefixes and the `format!` approach to printing variables are both things that `tracing` improves on , didn't meant to suggest using `println!`.

I don't want to hold up your work on getting too nit picky in unit tests (though I think having clean and readable unit tests is important for maintainability). I think keeping the logging is a good idea. If this were production code I'd assert improving  the usage of tracing. But I'll leave it up to you if you want to do that here, of just leave it as is.

If you want to do the improvements, check out https://bitwarden.atlassian.net/wiki/spaces/EN/pages/2065858607/Desktop+Native+Observability#Tracing-usage-guide (btw, I'm in the process of porting this to Contributing Docs)

the two main improvements would be:
- variables (changing from  "{server_path:?}" to, debug!(?server_path, "...");
- since you're using async, you can spans (https://docs.rs/tracing/latest/tracing/span/index.html) , instead of the "[server]" prefixes etc.

apps/desktop/desktop_native/autofill_provider/src/util.rs
neuronull: I _personally_ would use`serde_with` instead, assuming my recollection about the base64 is correct that it would basically nullify the need for this file. It's a fairly commonly used crate, and . One concern about not using it is that someone in the future could be working on some other module, not know about serde_with, and see this logic here and replicate that. Maybe contrived. 
I won't consider this a blocker, you can TIOLI / time box trying serde_with

apps/desktop/desktop_native/autofill_provider/src/lib.rs
iinuwa: Yeah, typically `cargo test` only prints to stdout if the test fails, but the `tracing` crate seems to override that somehow, and the logs appear even when the tests pass. So if you happen to have `RUST_LOG=debug` set, then you get a bunch of messages in the logs for both passing and failing tests. For example, if I uncomment the line in question:

```
> env RUST_LOG=debug cargo test
    Finished `test` profile [unoptimized + debuginfo] target(s) in 0.12s
warning: the following packages contain code that will be rejected by a future version of Rust: num-bigint-dig v0.8.4
note: to see what the problems were, use the option `--future-incompat-report`, or run `cargo report future-incompatibilities --id 1`
     Running unittests src/lib.rs (/Users/iinuwa/Development/bitwarden/clients/apps/desktop/desktop_native/target/debug/deps/autofill_provider-9d7577486919ae89)

running 2 tests
test tests::test_disconnected_gives_error ... ok
2026-01-16T17:33:06.803668Z  INFO desktop_core::ipc::client: Attempting to connect path="/Users/iinuwa/Library/Caches/com.bitwarden.desktop/s.af-test-0"
2026-01-16T17:33:06.803708Z ERROR autofill_provider: Failed to connect to autofill IPC server: Connection refused (os error 61) path="/Users/iinuwa/Library/Caches/com.bitwarden.desktop/s.af-test-0"
2026-01-16T17:33:06.803782Z DEBUG autofill_provider::tests: Waiting for server...
2026-01-16T17:33:06.803784Z  INFO server: autofill_provider::tests: Starting server thread
2026-01-16T17:33:06.803809Z DEBUG server: autofill_provider::tests: Starting server at "/Users/iinuwa/Library/Caches/com.bitwarden.desktop/s.af-test-1"
2026-01-16T17:33:06.803951Z DEBUG server: autofill_provider::tests:  Server started
2026-01-16T17:33:06.803961Z DEBUG server: autofill_provider::tests: Waiting for messages
2026-01-16T17:33:06.803964Z DEBUG autofill_provider::tests: Starting client...
2026-01-16T17:33:06.803984Z DEBUG autofill_provider::tests: Client connecting...
2026-01-16T17:33:06.804014Z  INFO desktop_core::ipc::client: Attempting to connect path="/Users/iinuwa/Library/Caches/com.bitwarden.desktop/s.af-test-1"
2026-01-16T17:33:06.804047Z  INFO desktop_core::ipc::client: Connected path="/Users/iinuwa/Library/Caches/com.bitwarden.desktop/s.af-test-1"
2026-01-16T17:33:06.804062Z DEBUG server: autofill_provider::tests: Received Message { client_id: 1, kind: Connected, message: None }
2026-01-16T17:33:06.804069Z  INFO autofill_provider: Connected to server
2026-01-16T17:33:06.815827Z DEBUG server: autofill_provider::tests: Received Message { client_id: 1, kind: Message, message: Some("{\"sequence_number\":1,\"value\":{\"Ok\":{}}}") }
2026-01-16T17:33:06.815921Z  INFO autofill_provider: Time to process request: 190.833µs
2026-01-16T17:33:06.815974Z  INFO desktop_core::ipc::client: Client channel closed
2026-01-16T17:33:06.816009Z  INFO autofill_provider: Disconnected from server
2026-01-16T17:33:06.816004Z  INFO server: desktop_core::ipc::server: Client disconnected. client_id=1
test tests::test_get_lock_status ... ok2026-01-16T17:33:06.816033Z DEBUG server: autofill_provider::tests: Received Message { client_id: 1, kind: Disconnected, message: None }


test result: ok. 2 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.01s

     Running unittests uniffi-bindgen.rs (/Users/iinuwa/Development/bitwarden/clients/apps/desktop/desktop_native/target/debug/deps/uniffi_bindgen-c7acf0c551359115)

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s

   Doc-tests autofill_provider

running 2 tests
test autofill_provider/src/lib.rs - AutofillProviderClient (line 128) - compile ... ok
test autofill_provider/src/lib.rs - TimedCallback<T>::wait_for_response (line 500) - compile ... ok

test result: ok. 2 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.06s
```

I've removed it from this specific test and added a note at the top of the module.

apps/desktop/desktop_native/autofill_provider/src/lib.rs
iinuwa: Thanks for clarifying!

Removed the prefixes and modified the server_path variable interpolation.

apps/desktop/desktop_native/autofill_provider/src/util.rs
iinuwa: I'll go with your preference! Replaced with serde_with

iinuwa:
Rebased to fix merge conflicts.

apps/desktop/desktop_native/autofill_provider/src/window_handle_query.rs
coltonhurst: If we choose to clean this up in PM-23485 that's fine with me!

apps/desktop/desktop_native/autofill_provider/src/lib.rs
neuronull: 🤔  hmm, I wonder what else is in your env? I don't get the stdout:

```
❯ pwd
/Users/foo/workspace/bw/clients/apps/desktop/desktop_native/autofill_provider

❯ RUST_LOG=debug cargo test
    Finished `test` profile [unoptimized + debuginfo] target(s) in 0.12s
     Running unittests src/lib.rs (/Users/foo/workspace/bw/clients/apps/desktop/desktop_native/target/debug/deps/autofill_provider-a489f7cad6f7ac9e)

running 2 tests
test tests::test_disconnected_gives_error ... ok
test tests::test_get_lock_status ... ok

test result: ok. 2 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.01s
```

neuronull: Ah, I see now that the init() call was moved out to a comment. I get the same behavior now. IMO it's totally fine to have the logs there by default when the tests are run! The test cases are using the log levels in a good fashion, so the default info level doesn't flood the console or anything.

neuronull: Alternatively, something like this could be used, so that it's only initialized if the env var is set.

```Rust
fn init_test_tracing() {
    use tracing_subscriber::EnvFilter;
    
    // Only initialize if RUST_LOG is set
    if std::env::var("RUST_LOG").is_ok() {
        tracing_subscriber::fmt()
            .with_env_filter(EnvFilter::from_default_env())
            .try_init()
            .ok();
    }
}
```

apps/desktop/desktop_native/autofill_provider/src/util.rs
neuronull: thanks!

apps/desktop/desktop_native/autofill_provider/src/assertion.rs
neuronull: it looks like this was added in https://github.com/bitwarden/clients/pull/18075/commits/c7845cbe813c88dd30948aef3bb99e1d055d1e2f

apps/desktop/desktop_native/autofill_provider/src/lib.rs
neuronull: looks like this was added in https://github.com/bitwarden/clients/pull/18075/commits/2cb625206d07aa3a686062c40cecc8cbf9a425a6

.github/CODEOWNERS
```diff
@@ -8,7 +8,7 @@
 apps/desktop/desktop_native @bitwarden/team-platform-dev
 apps/desktop/desktop_native/objc/src/native/autofill @bitwarden/team-autofill-desktop-dev
 apps/desktop/desktop_native/core/src/autofill @bitwarden/team-autofill-desktop-dev
-apps/desktop/desktop_native/macos_provider @bitwarden/team-autofill-desktop-dev
+apps/desktop/desktop_native/autofill_provider @bitwarden/team-autofill-desktop-dev
```
coroiu, line 11: thought: I wonder if this could be migrated to the SDK and used to replace native on iOS

apps/desktop/desktop_native/autofill_provider/src/lib.rs
coroiu: I'd also suggest having "expected behavior" and "state under test" in mind when naming/writing these. We have some documentation for TS tests that describes the advantage of this https://contributing.bitwarden.com/contributing/testing/unit/naming-conventions. Don't take them as hard rules, I know that rust tests don't always fit this type of structure

.github/CODEOWNERS
coroiu, line 11: why in the world does GitHub tell me I'm the owner of `autofill_provider` ??? 🥲

coroiu (commented):
I didn't too a too deep review of this since I think Autofill owns most of it and probably knows it better than Platform by now. My only thought is that `autofill_provider/src/lib.rs` looks very complex to be able to handle things like timeouts. We've already discussed replacing it with the IPC framework though, so I think it's fine to just leave as-is

.github/CODEOWNERS
iinuwa: It' an interesting thought, and I've been thinking about how we can unify more of these implementations.

From my limited work with it, the iOS app has a bunch of infrastructure built around native autofill such that the communication pattern implemented here for desktop apps `autofill_provider` does not quite match.

I think it feels appropriate to keep one pattern for the desktop apps, which all share the Electron platform, and separate patterns for iOS and Android.

apps/desktop/desktop_native/autofill_provider/src/lib.rs
iinuwa: That's a good suggestion. I'll add it in in the next PR.

apps/desktop/desktop_native/autofill_provider/src/lib.rs
iinuwa: Thanks! I renamed these tests to make it clearer what they're doing.

PR #18094, opened 2025-12-22
[PM-25909] Commercial desktop builds 

dereknance:
## 🎟️ Tracking

[PM-25909](https://bitwarden.atlassian.net/browse/PM-25909)

## 📔 Objective

This PR introduces the plumbing necessary for a commercial build of the desktop app contained in the `bitwarden_license/bit-desktop` directory, including `npm` scripts, `nx` configurations, and CI jobs. No new functionality has been introduced at this time, only a new set of entry points that extend the OSS entry points.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

[PM-25909]: https://bitwarden.atlassian.net/browse/PM-25909?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.github/workflows/build-desktop.yml
```diff
@@ -268,47 +280,47 @@ jobs:
...
-          name: bitwarden_${{ env._PACKAGE_VERSION }}_x64.tar.gz
+          name: ${{ matrix.artifact_prefix }}bitwarden_${{ env._PACKAGE_VERSION }}_x64.tar.gz
           path: apps/desktop/dist/bitwarden_desktop_x64.tar.gz
           if-no-files-found: error
 
       - name: Upload .deb artifact
         uses: actions/upload-artifact@330a01c490aca151604b8cf639adc76d48f6c5d4 # v5.0.0
         with:
-          name: Bitwarden-${{ env._PACKAGE_VERSION }}-amd64.deb
+          name: ${{ matrix.artifact_prefix }}Bitwarden-${{ env._PACKAGE_VERSION }}-amd64.deb
```
quexten, line 335: This change as-is will break the flathub release process, and the template here: https://github.com/flathub/com.bitwarden.desktop/blob/8385ed7892b1c59b6e545c7857dee6c4771f51ca/com.bitwarden.desktop.yaml#L65 / https://github.com/flathub/com.bitwarden.desktop/blob/8385ed7892b1c59b6e545c7857dee6c4771f51ca/com.bitwarden.desktop.yaml#L60 has to be adjusted.

Do we have a follow-up ticket to do that?

I recommend opening the PR for flathub already, and then merging it together with whichever release will include this change.

coroiu: @quexten Why would this break flathub? The prefix is empty so OSS filenames shouldn't change, right?

apps/desktop/src/app/base-app.component.ts
coroiu, line 1: question: Why did you have to split this out into a separate file? Couldn't you just keep the original `app.component.ts` and simply extend that one in `bitwarden_license/bit-desktop` ?

apps/desktop/src/preload.base.ts
coroiu, line 1: question: same here, couldn't we just keep the original `preload.ts` and extend/use that directly in the commercial files?

coroiu (commented):
Looks good, just a few questions

.github/workflows/build-desktop.yml
quexten: Ah, seems you are right. In that case we should be good.

apps/desktop/src/app/base-app.component.ts
dereknance: That was my first attempt, but I fought a lot with angular over the `@Component` decorator on `bit-desktop`'s `AppComponent`.

Removing the decorator from `bit-desktop`'s `AppComponent` yielded this error:

> The class `AppComponent` is listed in the declarations of the `NgModule` `AppModule`, but is not a directive, a component, or a pipe. Either remove it from the `NgModule`'s declarations, or add an appropriate Angular decorator.

That is in addition to other errors about types like `ServiceWorkerGlobalScope` not being found, some web components like `bit-callout`, `router-outlet`, and `bit-toast-container` not being found, as well as the `async` and `i18n` pipes.

Adding the decorator to `bit-desktop`'s `AppComponent` only resolved the first error. It's been long enough that I don't remember all of the things I tried, but referencing a class with a dummy `@Component` decorator seemed the most straightforward to me, even though it makes for an ugly diff.

apps/desktop/src/preload.base.ts
dereknance: I think I was trying to avoid calling `exposeInMainWorld(...)` twice, but that seems a small price for not having a base file. Fixed in e8ab74938f!

apps/desktop/src/app/base-app.component.ts
dereknance: I'm having some success taking a different approach to this, probably since I've had some time away from this work and have fresher eyes.

dereknance: Massive success! 🎉  It was hard-fought, but the `AppComponent` structure is much nicer to look at now.

apps/desktop/src/preload.base.ts
coroiu: Ohh, hmm, didn't think of that. Calling `exposeInMainWorld` multiple times is fine?

bitwarden_license/bit-desktop/src/app/app.component.ts
```diff
@@ -0,0 +1,13 @@
+import { Component } from "@angular/core";
+
+import { AppComponent as BaseAppComponent } from "@bitwarden/desktop/app/app.component";
```
coroiu: nit
```suggestion
import { AppComponent as OssAppComponent } from "@bitwarden/desktop/app/app.component";
```

bitwarden_license/bit-desktop/src/app/app.module.ts
```diff
@@ -0,0 +1,20 @@
...
+import { NgModule } from "@angular/core";
+
+import { AppRoutingModule as OssRoutingModule } from "@bitwarden/desktop/app/app-routing.module";
+import { AppModule as OssModule } from "@bitwarden/desktop/app/app.module";
+
+import { AppRoutingModule } from "./app-routing.module";
+import { AppComponent } from "./app.component";
+
+@NgModule({
+  imports: [CommonModule, OssModule, OssRoutingModule, AppRoutingModule],
```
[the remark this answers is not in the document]
coroiu: I think this should be fine, seeing as this is how browser works, but if we want to be on the safe side we can split `app.module.ts` into `oss.module.ts` the way `web` works

apps/desktop/src/app/base-app.component.ts
coroiu: Nice :)

bitwarden_license/bit-desktop/src/preload.ts
```diff
@@ -0,0 +1,17 @@
...
+ * This file extends the OSS IPC with commercial-specific channels.
+ *
+ * See the OSS preload script for more information.
+ */
+
+export const ipc = {
+  ...ossIpc,
+};
+
+contextBridge.exposeInMainWorld("ipc", ipc);
```
dereknance, line 17: Continuing from [this discussion](https://github.com/bitwarden/clients/pull/18094#discussion_r2693529806):

> Ohh, hmm, didn't think of that. Calling `exposeInMainWorld` multiple times is fine?

Upon closer inspection of the logs/console, I'm seeing these errors:

```
VM4 sandbox_bundle:2 Unable to load preload script: /Users/derek/src/clients/apps/desktop/build/preload.js

VM4 sandbox_bundle:2 Error: Cannot bind an API on top of an existing property on the window object
```

As best I can tell, that first error is coming from [`window.main.ts`](https://github.com/bitwarden/clients/blob/main/apps/desktop/src/main/window.main.ts#L288) during `init()`. So I'll start thinking through how to make that conditionally refer to the commercial preload. My first thought is an environment variable, maybe fed through webpack. I'm open to suggestions if there's already a convention for commercial overrides like this that I simply haven't yet discovered.

The second error definitely seems to be a result of calling `exposeInMainWorld` multiple times, so I'll also get that fixed up.

bitwarden_license/bit-desktop/src/app/app.module.ts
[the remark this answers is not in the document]
dereknance: I split `app.module.ts` into `oss.module.ts` in cae5e9817e3c9333236ed1d548f06f84a6d85f37.

.github/workflows/build-desktop.yml
```diff
@@ -156,7 +158,17 @@ jobs:
...
+    name: Linux Build [${{ matrix.variant }}]
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - variant: oss
+            license_type: oss
+            artifact_prefix: ""
+          - variant: bit
+            license_type: commercial
```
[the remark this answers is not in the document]
dereknance: > The license_type matrix variable is defined but never used in the workflow. Is this intentional for future use, or should it be removed to reduce confusion?

At the time, `license_type` wasn't used, but it is now.

bitwarden_license/bit-desktop/src/app/app.component.ts
```diff
@@ -0,0 +1,13 @@
...
+
+// FIXME(https://bitwarden.atlassian.net/browse/CL-764): Migrate to OnPush
+// eslint-disable-next-line @angular-eslint/prefer-on-push-component-change-detection
+@Component({
+  selector: "app-root",
+  styles: [],
+  templateUrl: "../../../../apps/desktop/src/app/app.component.html",
```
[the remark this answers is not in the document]
dereknance: > Consider using a TypeScript path alias instead:
> ```
> templateUrl: "@bitwarden/desktop/app/app.component.html"
> ```

I figured this wouldn't work, but I wanted it to so I tested it...and it did not work.

bitwarden_license/bit-desktop/src/app/app.component.ts
dereknance: Fixed in a5fff0821966da2863ebd6582edc52f6e7f3c28a.

bitwarden_license/bit-desktop/src/preload.ts
dereknance: Fixed by reintroducing the preload base setup in cae5e9817e3c9333236ed1d548f06f84a6d85f37.

bitwarden_license/bit-desktop/src/app/app.component.ts
```diff
@@ -0,0 +1,13 @@
+import { Component } from "@angular/core";
+
+import { AppComponent as OssAppComponent } from "@bitwarden/desktop/app/app.component";
+
+// FIXME(https://bitwarden.atlassian.net/browse/CL-764): Migrate to OnPush
+// eslint-disable-next-line @angular-eslint/prefer-on-push-component-change-detection
+@Component({
+  selector: "app-root",
+  styles: [],
+  templateUrl: "../../../../apps/desktop/src/app/app.component.html",
```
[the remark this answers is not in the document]
dereknance: > This four-level-deep relative path works but is fragile.

As mentioned in [this thread](https://github.com/bitwarden/clients/pull/18094#pullrequestreview-3663354815) using the path alias did not work.

bitwarden_license/bit-desktop/src/preload.ts
```diff
@@ -0,0 +1,17 @@
...
```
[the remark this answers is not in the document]
dereknance: This comment is irrelevant. I've ensured that `exposeInMainWorld` is only being called once. Plus, the webpack config takes the preload file passed in and transpiles to the same location.

apps/desktop/package.json
```diff
@@ -20,20 +20,42 @@
...
-    "build": "concurrently -n Main,Rend,Prel -c yellow,cyan \"npm run build:main\" \"npm run build:renderer\" \"npm run build:preload\"",
-    "build:dev": "concurrently -n Main,Rend,Prel -c yellow,cyan \"npm run build:main:dev\" \"npm run build:renderer:dev\" \"npm run build:preload:dev\"",
-    "build:preload": "cross-env NODE_ENV=production webpack --config webpack.config.js --config-name preload",
-    "build:preload:dev": "cross-env NODE_ENV=development webpack --config webpack.config.js --config-name preload",
-    "build:preload:watch": "cross-env NODE_ENV=development webpack --config webpack.config.js --config-name preload --watch",
+    "build": "npm run build:oss",
+    "build:dev": "npm run build:oss:dev",
+    "build:preload": "npm run build:oss:preload",
+    "build:preload:dev": "npm run build:oss:preload:dev",
+    "build:preload:watch": "npm run build:oss:preload:watch",
```
dereknance: I aliased the build commands match the original (I think?) behavior of building the OSS version. Is it worth doing this to offer a "deprecation period", or is that overkill?

dereknance, line 33: I aliased the build commands match the original (I think?) behavior of building the OSS version. Is it worth doing this to offer a "deprecation period", or is that overkill?

dereknance:
@coroiu my changes appear to have caused some files to now get [picked up by the type checker](https://github.com/bitwarden/clients/actions/runs/22121246581/job/63941523081?pr=18094). I'm not certain about how to proceed, given that they are vault/autofill related.

Should I work to resolve those type errors, mask them in some way such that they can be handled later, or something else?

coroiu (dismissed):

dereknance:
Hey @coroiu, this PR fell by the wayside and needed fixing up. A summary of the changes since your last review are:

- Only the OSS MAS build is published to TestFlight and announced in Slack.
- The commercial SDK alias now applies to the commercial main and preload bundles, instead of only the renderer.
- The macOS beta job builds both OSS and commercial variants.
- CI now runs bit-desktop unit tests.
- Desktop build trigger on changes under bitwarden_license/bit-common and bit-desktop.

coroiu (approved):
Oh long time no see, nice

PR #18137, opened 2025-12-29
[PM-30268] fix(desktop): remove obsolete LockScreen config from Microsoft Store

### LINKED ISSUE #16746: Remove obsolete LockScreen config from Microsoft store installation
### Steps To Reproduce

1. Open the Windows Settings app.
2. Navigate to Personalization > Lock screen or go directly via this link -> `ms-settings:lockscreen`
3.  Under Lock screen status, Bitwarden appears as a selectable app.

### Expected Result

The following entry in `Package.appxmanifest`/`AppxManifest.xml` is no longer needed:
```xml
<uap:LockScreen Notification="badgeAndTileText" BadgeLogo="assets\BadgeLogo.png" />
```
Bitwarden shoudn't apper as Lock screen status  application.

### Actual Result

Bitwarden can be selected as a lock screen status app, but the new client doesn't support live tiles on Windows 10/11.

### Screenshots or Videos

Image

### Additional Context

_No response_

### Operating System

Windows

### Operating System Version

Windows 11

### Installation method

Microsoft Store

### Build Version

2025.9.0

### Issue Tracking Info

majiayu000:
## Summary
- Remove BadgeLogo.png to prevent Bitwarden from appearing as lock screen status app
- The new client no longer supports Windows 10/11 live tiles

Fixes #16746

## Test plan
- Build AppX package after this change
- Verify Bitwarden no longer appears in Windows Settings > Personalization > Lock screen status


djsmith85:
@iinuwa: You recently did some work around the appx builds and also introduced [custom-appx-manifest.xml](http://github.com/bitwarden/clients/blob/main/apps/desktop/custom-appx-manifest.xml#L99) which also lists the BadgeLogo along with `<uap:LockScreen Notification="badgeAndTileText" BadgeLogo="assets\BadgeLogo.png" />`. That line is what is called out as the root issue on the linked ticket. Albeit the custom-appx-manifest.xml currently not in use as that got temporarily reverted, from what I understand.

Any thoughts on this PR? Any experience with the LockScreen tiles?

iinuwa:
@djsmith85 I'm not sure of all the implications of removing the Lock Screen status are (though it seems like a good idea to me). This PR would solve that problem correctly. If we move forward with this, the `<uap:LockScreen>` element should also be updated in `custom-appx-manifest.xml` (see below).

> Albeit the custom-appx-manifest.xml currently not in use as that got temporarily reverted, from what I understand.

Correct, that manifest is currently only used locally for cross-compiled builds. We will have to start using the custom manifest when the native passkey plugin is released, so `custom-appx-manifest.xml` should also be updated to remove the lock screen configuration to keep it in sync with what electron-builder produces so that this does not regress.

djsmith85 (changes requested):
Hi @majiayu000 and thank you for your contribution ❤️ 

Could you please merge the latest from main and update/remove the line in [custom-appx-manifest.xml](https://github.com/bitwarden/clients/blob/main/apps/desktop/custom-appx-manifest.xml#L99).

Once that is done, the resulting build artifact can be tested.

majiayu000:
Updated: merged the latest main and removed the LockScreen configuration from apps/desktop/custom-appx-manifest.xml to keep it in sync with the generated AppX manifest.

djsmith85 (approved):

PR #18138, opened 2025-12-29
[PM-30269] fix(cli): respect no_proxy environment variable

### LINKED ISSUE #13763: 'no_proxy' env var not evaluated
### Steps To Reproduce

1.Set `http_proxy`, `https_proxy` and `no_proxy` env vars (including the bitwarden server in `no_proxy`).
2. Execute a Bitwarden CLI command (example: 'bw list items')

### Expected Result

The Bitwarden CLI should respect the `no_proxy` variable and don't use the proxy.

### Actual Result

The Bitwarden CLI will try to send a post request to the Bitwarden server using the proxy.

### Screenshots or Videos

_No response_

### Additional Context

This behavior was observed on Linux. I can't state if the behavior on other OSes is comparable.

### Operating System

Linux

### Operating System Version

Ubuntu 24.04.2 LTS

### Shell

Bash

### Build Version

2025.1.3

### Issue Tracking Info

majiayu000:
## Summary
- Add support for no_proxy/NO_PROXY environment variable in CLI
- Bypass proxy for hosts matching no_proxy patterns
- Supports exact hostname, domain suffix, wildcard, and comma-separated lists

Fixes #13763

## Test plan
- Set http_proxy and no_proxy environment variables
- Include Bitwarden server in no_proxy list
- Verify CLI connects directly without using proxy

### DIFF
--- a/apps/cli/src/platform/services/node-api.service.spec.ts
+++ b/apps/cli/src/platform/services/node-api.service.spec.ts
@@ -1,0 +1,116 @@
+import { mock } from "jest-mock-extended";
+
+import { AccountService } from "@bitwarden/common/auth/abstractions/account.service";
+import { TokenService } from "@bitwarden/common/auth/abstractions/token.service";
+import { DeviceType } from "@bitwarden/common/enums";
+import { VaultTimeoutSettingsService } from "@bitwarden/common/key-management/vault-timeout";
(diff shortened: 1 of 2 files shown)


addisonbeck:
@djsmith85 Does the proposed changeset here line up with how you were thinking we'd fix this?

PR #18145, opened 2025-12-29
[PM-29608] [PM-29609] Premium subscription redesign cards

amorask-bitwarden:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-29609
https://bitwarden.atlassian.net/browse/PM-29608

## 📔 Objective

This PR does some refactoring of existing @bitwarden/pricing components as well as introduces 3 new components in the @bitwarden/subscription library.

#### `cart-summary.component`
1. Refactors the component to operate off a single input with the `Cart` type rather than separate inputs for each portion of the cart. This prepares us for this component's usage on the redesigned premium subscription page where we'll fetch an entire cart from the backend.
2. Updates the usage of this component in the `upgrade-payment.component` by changing input to the `Cart` type with the same data passed in.
3. Adds an optional `header` template that can be passed into the component in the case we're rendering the subscription page.
4. Adds support for cart-level `Discount` display

#### `discount-badge.component`
1. Refactors the component to operate off a `Discount` input, which is a union type of "amount-off" and "percent-off" discounts.
2. Updates the usage of this component in the `user-subscription.component` by changing the input to the `Discount` type with the same data passed in.

#### `pricing-card.component`
1. Updates the component to use changeDetection: OnPush

#### `subscription-card.component`
1. Introduces a `BitwardenSubscription` type which will be fetched from the backend and used to render this card.
2. Implements the component operating off a `BitwardenSubscription` input and includes a spec file, stories and MDX file.

#### `storage-card.component`
1. Introduces a `Storage` type which will be fetched from the backend and used to render this card.
2. Implements the component operating off a `Storage` input and includes a spec file, stories and MDX file.

#### `additional-options-card.component`
1. Implements the component, spec file, stories and MDX file.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


sukhleenb (approved):

apps/web/src/locales/en/messages.json
```diff
@@ -12490,7 +12483,71 @@
...
+    "message": "Your subscription will be canceled on"
+  },
+  "yourSubscriptionWasCanceledOn": {
+    "message": "Your subscription was canceled on"
+  },
+  "storageFull": {
+    "message": "Storage full"
+  },
+  "storageUsedDescription": {
+    "message": "You have used $USED$ out of $AVAILABLE$ GB of your encrypted file storage."
```
kdenney: ❓ Does this message (and the next one below) require the "placeholders" property to be defined so that they get replaced properly?

libs/common/src/enums/feature-flag.enum.ts
```diff
@@ -33,6 +33,7 @@ export enum FeatureFlag {
   PM23713_PremiumBadgeOpensNewPremiumUpgradeDialog = "pm-23713-premium-badge-opens-new-premium-upgrade-dialog",
   PM26462_Milestone_3 = "pm-26462-milestone-3",
   PM23341_Milestone_2 = "pm-23341-milestone-2",
+  PM29593_PremiumToOrganizationUpgrade = "pm-29593-premium-to-organization-upgrade",
```
kdenney: ⛏️ I don't see any other usages of this feature flag in this PR. Was this included unintentionally?

libs/pricing/src/components/cart-summary/cart-summary.component.html
```diff
@@ -95,14 +100,15 @@ <h3 bitTypography="h5" class="tw-text-muted">{{ "secretsManager" | i18n }}</h3>
...
+          @let additionalServiceAccounts = cart.secretsManager?.additionalServiceAccounts;
           @if (additionalServiceAccounts) {
             <div id="additional-service-accounts" class="tw-flex tw-justify-between">
               <div bitTypography="body1" class="tw-text-muted">
                 {{ additionalServiceAccounts.quantity }}
                 {{ additionalServiceAccounts.name | i18n }} x
                 {{ additionalServiceAccounts.cost | currency: "USD" : "symbol" }}
                 /
-                {{ additionalServiceAccounts.cadence | i18n }}
+                {{ term | i18n }}
```
kdenney: 🎨 I believe the `i18n` is redundant here and should be removed. term should already contain text from the i18n service so it doesn't need localized here again.

libs/pricing/src/components/cart-summary/cart-summary.component.ts
```diff
@@ -1,76 +1,153 @@
...
+      return 0;
+    }
+
+    const subtotal = this.subtotal();
+    switch (discount.type) {
+      case "percent-off": {
+        const percentage = discount.value < 1 ? discount.value : discount.value / 100;
+        return subtotal * percentage;
+      }
+      case "amount-off":
```
kdenney: ⛏️ Could you add the amount-off and percent-off strings to constants somewhere? Since we just added these, it would be good to minimize the use of magic strings especially since we will be probably using these more when we do personal discounting soon.

libs/pricing/src/types/maybe.ts
```diff
@@ -0,0 +1 @@
+export type Maybe<T> = T | null | undefined;
```
kdenney, line 1: Nice type. I wonder if you should put this in libs/common or somewhere else centrally located?

libs/subscription/src/components/additional-options-card/additional-options-card.component.ts
```diff
@@ -0,0 +1,16 @@
+import { Component, ChangeDetectionStrategy, output } from "@angular/core";
+
+import { ButtonModule, CardComponent, TypographyModule } from "@bitwarden/components";
+import { I18nPipe } from "@bitwarden/ui-common";
+
+export type AdditionalOptionsCardAction = "download-license" | "cancel-subscription";
```
kdenney, line 6: ❓ I was going to suggest that we move these strings into constants also but... do you think the type checking we get here is good enough? If we change this type in the future, incorrect usages of the strings will no longer compile so they will be easy to track down when we do refactors but I still have that voice inside saying "don't use magic strings" 😆

apps/web/src/locales/en/messages.json
amorask-bitwarden: Great catch - resolved in latest commit.

libs/common/src/enums/feature-flag.enum.ts
amorask-bitwarden: Thanks, I previously had one of the new components using this flag but ultimately opted for a boolean input that will be passed in from the parent. Removed this.

libs/pricing/src/components/cart-summary/cart-summary.component.html
amorask-bitwarden: Thanks for catching that - removed.

libs/pricing/src/components/cart-summary/cart-summary.component.ts
amorask-bitwarden: Absolutely - added those and replaced all usages.

libs/pricing/src/types/maybe.ts
amorask-bitwarden: I thought about this as well, but with these libraries, we're supposed to limit our dependency on libs/common and remove those dependencies until we have none left. Obviously, the biggest violation of that at the moment is `I18nService` because we need that, but if that ever gets spun out into a new lib, we'll be clear of libs/common. Maybe we can spin up a libs/billing-types down the line or something to store these and our response models.

libs/subscription/src/components/additional-options-card/additional-options-card.component.ts
amorask-bitwarden: Generally for these component-specific result types, I think hard-types would be overkill. If you try to handle an incorrect case in the parent component, the compiler will catch it and auto-complete should only allow the allowed cases.

libs/pricing/src/types/maybe.ts
kdenney: Ah, that makes sense. Ok cool sounds good for a future refactor. Thanks for explaining!

cturnbull-bitwarden (approved):

vleague2 (approved):

PR #18149, opened 2025-12-30
[PM-30301][PM-30302] Use SDK for Create and Update cipher operations

nikwithak:
## 🎟️ Tracking

[PM-30301](https://bitwarden.atlassian.net/browse/PM-30301)
[PM-30302](https://bitwarden.atlassian.net/browse/PM-30301)

## 📔 Objective

This migrates the cipher Create and Update operations to use the SDK instead of calling the APIs directly. Hidden behind the new feature flag `pm-27632-cipher-crud-operations-to-sdk`, which will gate all of the cipher basic operations once implemented.

* **N.B.** Changes to the `cipher.service.ts` interface were made to use decrypted `CipherView`s instead of encrypted `Ciphers`. All of the code calling the service in the various clients was updated to use the new interface. The encryption/decryption of those values is now done in `cipher.service.ts` instead of in the clients.  In nearly every case, this resulted in no significant changes, since the encryption was done solely for the purpose of passing the Cipher to the CipherService functions.

## 📸 Screenshots

https://github.com/user-attachments/assets/efd7e2a4-0054-415e-8de0-9c35c9c2c4c5

https://github.com/user-attachments/assets/431e059c-0816-47c6-928d-5b4052b9038d

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

[PM-30301]: https://bitwarden.atlassian.net/browse/PM-30301?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[PM-30302]: https://bitwarden.atlassian.net/browse/PM-30302?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


nikwithak:
Note: Tests are failing to an existing issue in `main` - should be fixed with #18227

libs/common/src/vault/services/cipher.service.ts
```diff
@@ -903,6 +906,55 @@ export class CipherService implements CipherServiceAbstraction {
   }
 
   async createWithServer(
```
gbubemismith, line 907: 🤔 This might be a good opportunity to declutter the cipher service and extract the SDK operations into a separate service. What do you think?

libs/common/src/vault/services/cipher.service.ts
```diff
@@ -903,6 +906,55 @@ export class CipherService implements CipherServiceAbstraction {
...
+    if (sdkCipherEncryptionEnabled) {
+      return (await this.createWithServer_sdk(cipherView, userId, orgAdmin)) || new CipherView();
+    } else {
+      const encrypted = await this.encrypt(cipherView, userId);
+      const result = await this.createWithServer_legacy(encrypted, orgAdmin);
+      return await this.decrypt(result, userId);
+    }
+  }
+
+  private async createWithServer_sdk(
```
gbubemismith: ⚠️ please use camelCase for methods

libs/common/src/vault/services/cipher.service.ts
```diff
@@ -903,6 +906,55 @@ export class CipherService implements CipherServiceAbstraction {
...
+          this.logService.error(`Failed to create cipher: ${error}`);
+          return EMPTY;
+        }),
+      ),
+    );
+    await this.clearCache(userId);
+    return resultCipherView;
+  }
+
+  private async createWithServer_legacy(
```
gbubemismith, line 941: ⚠️ please use camelCase for methods

libs/common/src/vault/services/cipher.service.ts
```diff
@@ -929,6 +981,62 @@ export class CipherService implements CipherServiceAbstraction {
...
+      return await this.updateWithServer_sdk(cipherView, userId, originalCipherView, orgAdmin);
+    } else {
+      const encrypted = await this.encrypt(cipherView, userId);
+      const updatedCipher = await this.updateWithServer_legacy(encrypted, orgAdmin);
+      const updatedCipherView = this.decrypt(updatedCipher, userId);
+      return updatedCipherView;
+    }
+  }
+
+  async updateWithServer_sdk(
```
gbubemismith: ⚠️ same naming issue here

libs/common/src/vault/services/cipher.service.ts
```diff
@@ -903,6 +906,55 @@ export class CipherService implements CipherServiceAbstraction {
...
+    cipherView: CipherView,
+    userId: UserId,
+    orgAdmin?: boolean,
+  ): Promise<CipherView> {
+    const sdkCipherEncryptionEnabled = await this.configService.getFeatureFlag(
+      FeatureFlag.PM27632_SdkCipherCrudOperations,
+    );
+
+    if (sdkCipherEncryptionEnabled) {
+      return (await this.createWithServer_sdk(cipherView, userId, orgAdmin)) || new CipherView();
```
gbubemismith: ❓ what happens when there's a failure on the SDK

libs/common/src/vault/services/cipher.service.ts
```diff
@@ -903,6 +906,55 @@ export class CipherService implements CipherServiceAbstraction {
...
+          }
+          return CipherView.fromSdkCipherView(result);
+        }),
+        catchError((error: unknown) => {
+          this.logService.error(`Failed to create cipher: ${error}`);
+          return EMPTY;
+        }),
+      ),
+    );
+    await this.clearCache(userId);
```
gbubemismith, line 937: can you add a comment on why we need this here?

libs/common/src/vault/services/cipher.service.ts
```diff
@@ -903,6 +906,55 @@ export class CipherService implements CipherServiceAbstraction {
...
   async createWithServer(
+    cipherView: CipherView,
+    userId: UserId,
+    orgAdmin?: boolean,
+  ): Promise<CipherView> {
+    const sdkCipherEncryptionEnabled = await this.configService.getFeatureFlag(
+      FeatureFlag.PM27632_SdkCipherCrudOperations,
+    );
+
+    if (sdkCipherEncryptionEnabled) {
```
gbubemismith: 🎨 I think we can simplify this remove the else. 

``` ts
 if (sdkCipherEncryptionEnabled) {
    return ...;
 }

// perform legacy operation
```

libs/common/src/vault/services/cipher.service.ts
```diff
@@ -929,6 +981,62 @@ export class CipherService implements CipherServiceAbstraction {
...
+    cipherView: CipherView,
+    userId: UserId,
+    originalCipherView?: CipherView,
+    orgAdmin?: boolean,
+  ): Promise<CipherView> {
+    const sdkCipherEncryptionEnabled = await this.configService.getFeatureFlag(
+      FeatureFlag.PM27632_SdkCipherCrudOperations,
+    );
+
+    if (sdkCipherEncryptionEnabled) {
```
gbubemismith: 🎨 same here, simplify by remove the else block

libs/common/src/vault/services/cipher.service.ts
```diff
@@ -929,6 +981,62 @@ export class CipherService implements CipherServiceAbstraction {
...
+      return await this.updateWithServer_sdk(cipherView, userId, originalCipherView, orgAdmin);
+    } else {
+      const encrypted = await this.encrypt(cipherView, userId);
+      const updatedCipher = await this.updateWithServer_legacy(encrypted, orgAdmin);
+      const updatedCipherView = this.decrypt(updatedCipher, userId);
```
gbubemismith: Looks like you missed awaiting this

libs/common/src/vault/models/view/cipher.view.ts
```diff
@@ -332,6 +337,85 @@ export class CipherView implements View, InitializerMetadata {
     return cipherView;
   }
 
+  /**
+   * Maps CipherView to an SDK CipherCreateRequest
+   *
+   * @returns {CipherCreateRequest} The SDK cipher create request object
+   */
+  toSdkCreateCipherRequest(): CipherCreateRequest {
```
gbubemismith, line 345: can we add tests for these mappings?

apps/desktop/src/services/encrypted-message-handler.service.ts
```diff
@@ -166,8 +166,7 @@ export class EncryptedMessageHandlerService {
 
     try {
       const activeUserId = await firstValueFrom(getUserId(this.accountService.activeAccount$));
-      const encrypted = await this.cipherService.encrypt(cipherView, activeUserId);
-      await this.cipherService.createWithServer(encrypted);
+      await this.cipherService.createWithServer(cipherView, activeUserId);
```
gbubemismith, line 169: 👍🏼 Like how we don't need to encrypt and save anymore

apps/desktop/src/autofill/services/desktop-fido2-user-interface.service.ts
```diff
@@ -299,12 +299,11 @@ export class DesktopFido2UserInterfaceSession implements Fido2UserInterfaceSessi
       throw new Error("No active user ID found!");
     }
 
-    const encCipher = await this.cipherService.encrypt(cipher, activeUserId);
-
     try {
-      const createdCipher = await this.cipherService.createWithServer(encCipher);
+      const createdCipher = await this.cipherService.createWithServer(cipher, activeUserId);
+      const encryptedCreatedCipher = await this.cipherService.encrypt(createdCipher, activeUserId);
```
gbubemismith, line 304: I think we can just use `cipherService.get(cipherId, userId)` to get the encrypted cipher

nikwithak: We could `get` the cipher from the cache, but that relies on the specifics of the `createWithServer` implementation properly updating the cache - since the intent is to return the result of the operation, I think it's better to directly encrypt and return that result - thoughts?

libs/common/src/vault/services/cipher.service.ts
nikwithak: Great catch, fixed!

libs/common/src/vault/services/cipher.service.ts
nikwithak: If there's an SDK failure, it should throw an error that then gets passed up to the calling function.

libs/common/src/vault/services/cipher.service.ts
nikwithak: Are you suggesting to create a new, secondary implementation of the `cipher.service.ts` interface for the SDK operations in particular? Or a new service that exposes the SDK functionality with a new interface?

The current approach makes it a lot easier to swap out the implementation for all the clients that currently user `cipher.service.ts`, and lets them be agnostic to the implementation details - if I"m understanding the ask right, we would have to add the feature flag check to all of the downstream client code that calls `createWithSErver`, rather than centralize it here.

gbubemismith: No just the SDK implementation and then you can use it like you do already

apps/desktop/src/autofill/services/desktop-fido2-user-interface.service.ts
gbubemismith: We normally should have it update the cache, but if we are not getting consistent updates to the cache then we can leave as is

gbubemismith (approved):
Thanks for making the changes on cipher service.

PR #18164, opened 2025-12-31
[PM-29208] Remove individual cryptographic-key states & migrate key service

quexten:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-29208

## 📔 Objective

This PR removes the following individual cryptographic states:
- User Private Key
- User Signature Key
- User Signed Public Key
- User Signed Security State
and replaces all subscriptions to these states, with subscriptions to the account cryptography state.

These temporary mappings can be removed once the SDK service subscribes directly to the account cryptographic state, and their services can be removed.

Please note, this also drops the old path for account private key regen, if an account does not have a private key. This can instead be done in the Data Recovery tool.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


libs/auth/src/common/login-strategies/sso-login.strategy.ts
```diff
@@ -345,20 +345,6 @@ export class SsoLoginStrategy extends LoginStrategy {
         userId,
       );
     }
-
-    if (tokenResponse.hasMasterKeyEncryptedUserKey()) {
```
quexten, line 349: Here, the migration is to just set the account cryptographic state conditionally. Private key creation is happening anymore, so there is no need for this special branching logic.

libs/auth/src/common/login-strategies/password-login.strategy.ts
```diff
@@ -148,20 +148,14 @@ export class PasswordLoginStrategy extends LoginStrategy {
...
 
-  protected override async setPrivateKey(
+  protected override async setAccountCryptographicState(
     response: IdentityTokenResponse,
     userId: UserId,
   ): Promise<void> {
-    await this.keyService.setPrivateKey(
-      response.privateKey ?? (await this.createKeyPairForOldAccount(userId)),
+    await this.accountCryptographicStateService.setAccountCryptographicState(
+      response.accountKeysResponseModel.toWrappedAccountCryptographicState(),
```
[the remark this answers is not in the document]
quexten: This is not correct. Password login strategy never has null account keys, only SSO login does.

libs/auth/src/common/login-strategies/auth-request-login.strategy.ts
```diff
@@ -120,20 +120,14 @@ export class AuthRequestLoginStrategy extends LoginStrategy {
...
 
-  protected override async setPrivateKey(
+  protected override async setAccountCryptographicState(
     response: IdentityTokenResponse,
     userId: UserId,
   ): Promise<void> {
-    await this.keyService.setPrivateKey(
-      response.privateKey ?? (await this.createKeyPairForOldAccount(userId)),
+    await this.accountCryptographicStateService.setAccountCryptographicState(
+      response.accountKeysResponseModel.toWrappedAccountCryptographicState(),
```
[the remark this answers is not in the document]
quexten: This is not correct. Password login strategy never has null account keys, only SSO login does.

libs/auth/src/common/login-strategies/user-api-login.strategy.ts
```diff
@@ -79,20 +79,14 @@ export class UserApiLoginStrategy extends LoginStrategy {
...
 
-  protected override async setPrivateKey(
+  protected override async setAccountCryptographicState(
     response: IdentityTokenResponse,
     userId: UserId,
   ): Promise<void> {
-    await this.keyService.setPrivateKey(
-      response.privateKey ?? (await this.createKeyPairForOldAccount(userId)),
+    await this.accountCryptographicStateService.setAccountCryptographicState(
+      response.accountKeysResponseModel.toWrappedAccountCryptographicState(),
```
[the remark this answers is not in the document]
quexten: This is not correct. Password login strategy never has null account keys, only SSO login does.

libs/auth/src/common/login-strategies/webauthn-login.strategy.ts
```diff
@@ -99,20 +99,14 @@ export class WebAuthnLoginStrategy extends LoginStrategy {
...
 
-  protected override async setPrivateKey(
+  protected override async setAccountCryptographicState(
     response: IdentityTokenResponse,
     userId: UserId,
   ): Promise<void> {
-    await this.keyService.setPrivateKey(
-      response.privateKey ?? (await this.createKeyPairForOldAccount(userId)),
+    await this.accountCryptographicStateService.setAccountCryptographicState(
+      response.accountKeysResponseModel.toWrappedAccountCryptographicState(),
```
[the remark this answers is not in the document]
quexten: This is not correct. Password login strategy never has null account keys, only SSO login does.

quexten:
Disabling claude since it's just duplicating.

quexten:
BIT tests are most likely failing due to the outdated server version that BIT is using. I'll check in with autofill about this before merging, but locally everything works.

libs/auth/src/common/login-strategies/auth-request-login.strategy.ts
Thomas-Avery, line 128: ❓  Has `response.accountKeysResponseModel` been introduced on the server more than 3 releases ago? 

This would break when a newer client targets an older server that doesn't have this property on the `IdentityTokenResponse`. 

If we haven't had 3 releases for the property we should be able to use `response.privateKey` as a fallback if `response.accountKeysResponseModel` doesn't exist.

This feedback applies to all login strategies.

libs/auth/src/common/login-strategies/login.strategy.spec.ts
```diff
@@ -336,28 +327,6 @@ describe("LoginStrategy", () => {
...
-
-      // User symmetric key must be set before the new RSA keypair is generated
-      expect(keyService.setUserKey).toHaveBeenCalled();
-      expect(keyService.makeKeyPair).toHaveBeenCalled();
-      expect(keyService.setUserKey.mock.invocationCallOrder[0]).toBeLessThan(
-        keyService.makeKeyPair.mock.invocationCallOrder[0],
-      );
-
-      expect(apiService.postAccountKeys).toHaveBeenCalled();
-    });
```
Thomas-Avery, line 359: We are removing this test but not `createKeyPairForOldAccount` from `/libs/auth/src/common/login-strategies/login.strategy.ts`?

libs/auth/src/common/login-strategies/sso-login.strategy.ts
Thomas-Avery: I think I'm confused here. How will newly JIT provisioned SSO v1 users get key pairs now?

libs/key-management/src/key.service.ts
```diff
@@ -548,8 +501,6 @@ export class DefaultKeyService implements KeyServiceAbstraction {
     await this.clearUserKey(userId);
     await this.clearOrgKeys(userId);
     await this.clearProviderKeys(userId);
-    await this.clearKeyPair(userId);
```
Thomas-Avery, line 538: ❓  Do we need to add clearing `accountCryptographicState`?

libs/state/src/state-migrations/migrations/75-remove-user-encrypted-private-key.ts
```diff
@@ -0,0 +1,65 @@
...
+      // Remove userSignedPublicKey
+      const signedPubKey = await helper.getFromUser(userId, userSignedPublicKey);
+      if (signedPubKey != null) {
+        await helper.removeFromUser(userId, userSignedPublicKey);
+      }
+      // Remove accountSecurityState
+      const accountSecurity = await helper.getFromUser(userId, accountSecurityState);
+      if (accountSecurity != null) {
+        await helper.removeFromUser(userId, accountSecurityState);
+      }
```
Thomas-Avery: Should we be migrating this data into the `accountCryptographicState`? Feels like there might be some edge cases here were the sync/login won't be able to set the new state after we wipe this out. Offline unlock?

libs/state/src/state-migrations/migrations/75-remove-user-encrypted-private-key.spec.ts
```diff
@@ -0,0 +1,56 @@
...
+          },
+        },
+        user_user1_CRYPTO_DISK_privateKey: "abc",
+        user_user2_CRYPTO_DISK_privateKey: "def",
+        user_user1_CRYPTO_DISK_userSigningKey: "sign1",
+        user_user2_CRYPTO_DISK_userSigningKey: "sign2",
+        user_user1_CRYPTO_DISK_userSignedPublicKey: "pub1",
+        user_user2_CRYPTO_DISK_userSignedPublicKey: "pub2",
+        user_user1_CRYPTO_DISK_accountSecurityState: "security1",
+        user_user2_CRYPTO_DISK_accountSecurityState: "security2",
```
Thomas-Avery: These will be in the format. 

```
user_<user_id>_crypto_privateKey
```

libs/state/src/state-migrations/migrations/75-remove-user-encrypted-private-key.ts
```diff
@@ -0,0 +1,65 @@
...
+    name: "CRYPTO_DISK",
+  },
+};
+
+export const accountSecurityState: KeyDefinitionLike = {
+  key: "accountSecurityState",
+  stateDefinition: {
+    name: "CRYPTO_DISK",
+  },
+};
```
Thomas-Avery, line 46:
```suggestion
export const userEncryptedPrivateKey: KeyDefinitionLike = {
  key: "privateKey",
  stateDefinition: {
    name: "crypto",
  },
};

export const userKeyEncryptedSigningKey: KeyDefinitionLike = {
  key: "userSigningKey",
  stateDefinition: {
    name: "crypto",
  },
};

export const userSignedPublicKey: KeyDefinitionLike = {
  key: "userSignedPublicKey",
  stateDefinition: {
    name: "crypto",
  },
};

export const accountSecurityState: KeyDefinitionLike = {
  key: "accountSecurityState",
  stateDefinition: {
    name: "crypto",
  },
};
```

```
export const CRYPTO_DISK = new StateDefinition("crypto", "disk");
```

CRYPTO_DISK is the type the name is `crypto`. I find it easier to get an example of the state from dev tools session storage explorer and work backwards from there.

libs/auth/src/common/login-strategies/auth-request-login.strategy.ts
quexten: It has been introduced in 2025.11.0. I'm unsure what 3 releases actually means (do patch releases count), but if we target this for merging *after* the 2026.1.0 `rc` cut then we should be good, right?

The alternative is re-constructing the accountCryptographicState from the private key in a few places, but I'm unsure if that's actually needed if we are good to remove the previous property soon anyways.

libs/auth/src/common/login-strategies/login.strategy.spec.ts
quexten: Removed now

libs/key-management/src/key.service.ts
quexten: Added clearing. I'm a bit confused as these should be handled by the `clearOn: ["logout"]`

libs/state/src/state-migrations/migrations/75-remove-user-encrypted-private-key.ts
quexten: My assumption was that users that were offline for an extended amount of time would also not have their clients updated and/or have invalid tokens so when they do come online they would not be able to sync.

However, the account cryptographic state service was only merged 2 weeks ago, so I added the migrations now.

libs/state/src/state-migrations/migrations/75-remove-user-encrypted-private-key.ts
quexten: Fixed, thanks for pointing this out. I confirmed with `data.json` from desktop now.

libs/auth/src/common/login-strategies/sso-login.strategy.ts
quexten: Newly provisioned JIT TDE users get their keys set in `login-decryption-options.component`. We currently have the not-yet-rolled-out v2 code (SDK) and the v1 logic. Newly provisioned JIT Password users get their keys set in the setInitialPasswordService. We don't have the V2 (sdk) logic merged yet, but it's in progress.

libs/auth/src/common/login-strategies/login.strategy.spec.ts
quexten: Removed

libs/auth/src/common/login-strategies/sso-login.strategy.ts
Thomas-Avery: Ah I got mixed up on the code comment, this was skipping the JIT provisioned SSO users from the private key setting by checking the `hasMasterKeyEncryptedUserKey`. For some reason I had that logic inverted. Since we aren't setting the private key in any case it doesn't matter. Thank you.

libs/auth/src/common/login-strategies/auth-request-login.strategy.ts
Thomas-Avery: It is 3 major releases, so patch releases don't count.

If we make sure not to merge until after the next rc cut we should be good.

dereknance (approved):

libs/auth/src/common/login-strategies/auth-request-login.strategy.ts
quexten: > Each server version is compatible with clients of the same major version, the previous 2 major client versions, and the subsequent 2 major client versions.
https://bitwarden.com/help/bitwarden-software-release-support/

So, since this was added in 2025.11.0 https://github.com/bitwarden/server/blob/bb56bd7fedfb1cf0ac53b8a138bfc7d786417e74/src/Api/Models/Response/ProfileResponseModel.cs#L69. So that means that even removing support for the old values within 2026.1.0 is fine because it means that the current +-2 versions compatibility is kept.

(In this case, the 2026.1.0 client would be compatible with server 2026.1.0, 2025.12.x, 2025.11.x).

Do you agree? Or did I get something wrong here? (I'm just trying to figure out when this merges since this conflicts with some of the registration work / makes us need extra cleanup tasks there).

PR #18188, opened 2026-01-02
[PM-27325] Deprecate user account crypto init methods

Thomas-Avery:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-27325

## 📔 Objective

Deprecate user account cryptography initialization methods on the `KeyService`. New usages should use the SDK.

`makeUserKeyV1()` was never used so it is safe to remove. 

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/libs/nx-plugin/src/generators/basic-lib.spec.ts
+++ b/libs/nx-plugin/src/generators/basic-lib.spec.ts
@@ -24,7 +24,7 @@ describe("basic-lib generator", () => {
     expect(tsconfigContent).not.toBeNull();
     const tsconfig = JSON.parse(tsconfigContent?.toString() ?? "");
     expect(tsconfig.compilerOptions.paths[`@bitwarden/${options.name}`]).toEqual([
-      `libs/test/src/index.ts`,
+      `./libs/test/src/index.ts`,
     ]);
   });
 
--- a/libs/nx-plugin/src/generators/basic-lib.ts
+++ b/libs/nx-plugin/src/generators/basic-lib.ts
@@ -82,7 +82,7 @@ function updateTsConfigPath(tree: Tree, name: string, srcRoot: string) {
   updateJson(tree, "tsconfig.base.json", (json) => {
     const paths = json.compilerOptions.paths || {};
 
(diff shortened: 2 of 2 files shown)


libs/key-management/src/abstractions/key.service.ts
```diff
@@ -334,9 +330,8 @@ export abstract class KeyService {
   abstract getFingerprint(fingerprintMaterial: string, publicKey: Uint8Array): Promise<string[]>;
   /**
    * Generates a new keypair
```
mzieniukbw, line 331: The deprecation note is missing on `makeKeyPair`

Thomas-Avery: I wrote up some context here https://bitwarden.atlassian.net/browse/PM-27325?focusedCommentId=124321. If we want to still deprecate I think we should add some verbiage that it's only deprecated for new use cases as some current uses cases won't be removed by current future work like the others.

mzieniukbw: Ok fine by me. Will we have separate ticket for the removal, that we will tackle later, so we won't forget ?

mzieniukbw (dismissed):

libs/key-management/src/abstractions/key.service.ts
```diff
@@ -334,9 +330,8 @@ export abstract class KeyService {
...
-   * @param key A key to encrypt the private key with. If not provided,
-   * defaults to the user key
-   * @returns A new keypair: [publicKey in Base64, encrypted privateKey]
+   * @param key A symmetric key to wrap the newly created private key with.
+   * @returns A new keypair: [publicKey in Base64, wrapped privateKey]
```
quexten, line 334: As per the jira comment:
```suggestion
   * @returns A new keypair: [publicKey in Base64, wrapped privateKey]
   * @deprecated New use-cases of this function are prohibited. Low-level cryptographic constructions and initialization should be done in the SDK
```

quexten (dismissed):

mzieniukbw (approved):

PR #18193, opened 2026-01-03
[PM-30397] [BEEEP] Refactor automatic start on login into its own service

trmartin4:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-30397

## 📔 Objective

Configuration of automatic start on login was enmeshed in the `messaging.main.ts` file.

This PR pulls that out into a new, tested `AutoStartService`, which is responsible for enabling and disabling auto-start on the underlying platform.

In addition, for the "is auto-start enabled?" check, we return more context than just a boolean, because on some systems we do not have an API that tells us if it's actually enabled at the current time (i.e. we can enable and disable it, but we can't know whether it's been enabled or disabled by the user in the meantime):
- `Enabled` - we can guarantee that auto-start is enabled on the system
- `Disabled` - we can guarantee that auto-start is not enabled on the system
- `Unknown` - the current platform cannot tell us whether auto-start is enabled
  - In this case, we will keep the value that we store in state as-is. 
  - ⚠️ This has the potential for being confusing _if_ a user on Flatpak in particular disables auto-start outside of the Bitwarden application, as we'd still think it was enabled and the setting would show it as enabled.  A user would have to toggle the setting to sync it back up again.

In addition, this PR hides the "start at login" setting for snap as well as Windows store, because snap is configured in `electron-builder` configuration and not through the setting.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/apps/desktop/src/app/accounts/settings.component.ts
+++ b/apps/desktop/src/app/accounts/settings.component.ts
@@ -66,6 +66,7 @@ import { SshAgentPromptType } from "../../autofill/models/ssh-agent-setting";
 import { DesktopAutofillSettingsService } from "../../autofill/services/desktop-autofill-settings.service";
 import { DesktopAutotypeService } from "../../autofill/services/desktop-autotype.service";
 import { DesktopBiometricsService } from "../../key-management/biometrics/desktop.biometrics.service";
+import { AutoStartService } from "../../platform/auto-start";
 import { DesktopSettingsService } from "../../platform/services/desktop-settings.service";
 import { DesktopPremiumUpgradePromptService } from "../../services/desktop-premium-upgrade-prompt.service";
 import { NativeMessagingManifestService } from "../services/native-messaging-manifest.service";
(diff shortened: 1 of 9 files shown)


trmartin4:
@quexten curious if you think this is a worthwhile abstraction to make.  It feels like this cleans up some of the ambiguity around auto-start and might make it easier to reason about it in the future.

quexten:
@trmartin4 Yes this change makes sense from my perspective.

As far as I understand there are plans from @bitwarden/team-autofill-dev  (cc @coltonhurst  / @neuronull)'s side to move all electron-specific functionalities to Rust at some point, and moving them out to a service like this already aids in that transition.

jprusik:
> @trmartin4 Yes this change makes sense from my perspective.
> 
> As far as I understand there are plans from @bitwarden/team-autofill-dev (cc @coltonhurst / @neuronull)'s side to move all electron-specific functionalities to Rust at some point, and moving them out to a service like this already aids in that transition.

they have a dedicated group now 😄 ! @bitwarden/team-autofill-desktop-dev

quexten:
@trmartin4 Are you still continuing this PR? I have it in my review queue still (and it is draft) but I'm not sure whether it will still be worked on?

PR #18006, opened 2025-12-16
[PM-27086] JIT MP org user flow - remove masterKey generation from InputPasswordComponent

rr-bw:
## 🎟️ Tracking

[PM-27086](https://bitwarden.atlassian.net/browse/PM-27086)

## 📔 Objective

- Updates `InputPasswordComponent` to emit raw data instead of generating cryptographic properties (`newMasterKey`, `newServerMasterKeyHash`, `newLocalMasterKeyHash`).
  - This allows us to move away from using the deprecated `makeMasterKey()` method in the component (which takes email as salt) as we seek to eventually separate the email from the salt.
  - The `InputPasswordComponent` does still use use `makeMasterKey()` in the `verifyCurrentPassword()` method, but that will be updated in a future PR specific to the `ChangePasswordService`.
- Updates the `JIT_PROVISIONED_MP_ORG_USER` case of the switch to handle the flow when the `PM27086_UpdateAuthenticationApisForInputPassword` flag is on.
- Follow-up PRs will handle the `TDE_ORG_USER_RESET_PASSWORD_PERMISSION_REQUIRES_MP` and `OFFBOARDED_TDE_ORG_USER` cases.

_Feature Flag Approach_ - `PM27086_UpdateAuthenticationApisForInputPassword`

- Instead of repeatedly checking the feature flag value via `ConfigService`, I opted to feature flag this in one area (within `InputPasswordComponent`), and then pass the feature flag value via the `PasswordInputResult` to consumers.

## 📸 Screenshots

https://github.com/user-attachments/assets/094ddea0-f1ca-4937-bee7-0b5f444716f8

## ⏰ Reminders before review

- Contributor guidelines followed
- All formatters and local linters executed and passed
- Written new unit and / or integration tests where applicable
- Protected functional changes with optionality (feature flags)
- Used internationalization (i18n) for all UI strings
- CI builds passed
- Communicated to DevOps any deployment requirements
- Updated any necessary documentation (Confluence, contributing docs) or informed the documentation team

## 🦮 Reviewer guidelines

- 👍 (`:+1:`) or similar for great changes
- 📝 (`:memo:`) or ℹ️ (`:information_source:`) for notes or general info
- ❓ (`:question:`) for questions
- 🤔 (`:thinking:`) or 💭 (`:thought_balloon:`) for more open inquiry that's not quite a confirmed issue and could potentially benefit from discussion
- 🎨 (`:art:`) for suggestions / improvements
[...]

PR #18222, opened 2026-01-06
[PM-27233] Support v2 encryption for JIT Password signups

mzieniukbw:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-27233

## 📔 Objective

Account v2 encryption for SSO JIT Master Password registration.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/apps/desktop/src/app/services/set-initial-password/desktop-set-initial-password.service.spec.ts
+++ b/apps/desktop/src/app/services/set-initial-password/desktop-set-initial-password.service.spec.ts
@@ -2,6 +2,7 @@ import { mock, MockProxy } from "jest-mock-extended";
 import { BehaviorSubject, of } from "rxjs";
 
 import { OrganizationUserApiService } from "@bitwarden/admin-console/common";
+import { DefaultSetInitialPasswordService } from "@bitwarden/angular/auth/password-management/set-initial-password/default-set-initial-password.service.implementation";
 import {
   InitializeJitPasswordCredentials,
   SetInitialPasswordCredentials,
@@ -201,7 +202,7 @@ describe("DesktopSetInitialPasswordService", () => {
 
       const superSpy = jest
         .spyOn(
-          Object.getPrototypeOf(Object.getPrototypeOf(sut)),
+          DefaultSetInitialPasswordService.prototype,
           "initializePasswordJitPasswordUserV2Encryption",
         )
         .mockResolvedValue(undefined);


libs/angular/src/auth/password-management/set-initial-password/default-set-initial-password.service.implementation.ts
```diff
@@ -47,12 +68,141 @@ export class DefaultSetInitialPasswordService implements SetInitialPasswordServi
...
+      MasterPasswordUnlockData.fromSdk(registerResult.master_password_unlock),
+      userId,
+    );
+
+    // [PM-23246] "Legacy" master key setting path - to be removed once unlock path migration is complete
+    await this.masterPasswordService.setMasterKeyHash(newLocalMasterKeyHash, userId);
+
+    if (resetPasswordAutoEnroll) {
+      await this.handleResetPasswordAutoEnroll(newServerMasterKeyHash, orgId, userId);
+    }
```
mzieniukbw: ~~Depending on the Auth's approach, the master key hashes might need to be included into the SDK's `registerResult`~~ Moved to SDK

libs/angular/src/auth/password-management/set-initial-password/default-set-initial-password.service.implementation.ts
```diff
@@ -47,12 +67,137 @@ export class DefaultSetInitialPasswordService implements SetInitialPasswordServi
...
+      accountEncryptionV2 &&
+      userType === SetInitialPasswordUserType.JIT_PROVISIONED_MP_ORG_USER
+    ) {
+      await this.setInitialPasswordV2(credentials, userType, userId);
+    } else {
+      await this.setInitialPasswordV1(credentials, userType, userId);
+    }
+  }
+
+  private async setInitialPasswordV2(
```
trmartin4: 🤔 Could we perhaps change the names of these methods to be more explicit than `V1` and `V2`?  At least `...EncryptionV1` and `...EncryptionV2`, perhaps?  The terms "V1" and "V2" are terms of art in the KM domain that imply an encryption model change, but that is not necessarily something that will be understood in the Auth domain, where this conditional will live.

libs/angular/src/auth/password-management/set-initial-password/default-set-initial-password.service.implementation.ts
```diff
@@ -47,12 +67,137 @@ export class DefaultSetInitialPasswordService implements SetInitialPasswordServi
...
+    if (!("V2" in registerResult.account_cryptographic_state)) {
+      throw new Error("Unexpected V2 account cryptographic state");
+    }
+
+    // Set account cryptography state
+    await this.accountCryptographicStateService.setAccountCryptographicState(
+      registerResult.account_cryptographic_state,
+      userId,
+    );
+    // Legacy individual states
```
trmartin4: ❓ Like with the "v1" and "v2" terminology, could we have additional context via comment around what "legacy" means here with respect to encryption models, or if it's unrelated?  Please also link a ticket that explains when this would be removed.

Could we also consider a `keyService.setLegacyState()` method to encapsulate all of this legacy state with that comment on it explaining what "legacy" means in this case, if you anticipate needing to set all of this together in the future?  It feels dangerous to be adding all of these individual methods in multiple places, if that's what you may end up needing to do.

libs/angular/src/auth/password-management/set-initial-password/default-set-initial-password.service.implementation.ts
```diff
@@ -47,12 +67,137 @@ export class DefaultSetInitialPasswordService implements SetInitialPasswordServi
...
+      ],
+      userId,
+    );
+
+    await this.masterPasswordService.setMasterPasswordUnlockData(
+      MasterPasswordUnlockData.fromSdk(registerResult.master_password_unlock),
+      userId,
+    );
+
+    // [PM-23246] "Legacy" master key setting path - to be removed once unlock path migration is complete
```
trmartin4: ❓ Is this the same "legacy" cleanup as the other "legacy state" mentioned above?  In other words, can they all be linked to PM-23246 for cleanup?

libs/angular/src/auth/password-management/set-initial-password/default-set-initial-password.service.implementation.ts
```diff
@@ -47,12 +67,137 @@ export class DefaultSetInitialPasswordService implements SetInitialPasswordServi
...
+    );
+    await this.securityStateService.setAccountSecurityState(
+      registerResult.account_cryptographic_state.V2.security_state as SignedSecurityState,
+      userId,
+    );
+
+    // Clear force set password reason to allow navigation back to vault.
+    await this.masterPasswordService.setForceSetPasswordReason(ForceSetPasswordReason.None, userId);
+
+    await this.updateAccountDecryptionProperties(
```
trmartin4: ❓ Because we're setting `MasterKey` here, is this also legacy?

libs/angular/src/auth/password-management/set-initial-password/default-set-initial-password.service.implementation.ts
trmartin4: I see that we have the same state being set here, for example: https://github.com/bitwarden/clients/blob/21b5ba880f309d99c2a13dcd926e429499d24e93/libs/auth/src/angular/login-decryption-options/login-decryption-options.component.ts#L326.

libs/angular/src/auth/password-management/set-initial-password/default-set-initial-password.service.implementation.ts
```diff
@@ -47,12 +67,137 @@ export class DefaultSetInitialPasswordService implements SetInitialPasswordServi
...
     userType: SetInitialPasswordUserType,
     userId: UserId,
+  ): Promise<void> {
+    const accountEncryptionV2 = await this.configService.getFeatureFlag(
+      FeatureFlag.EnableAccountEncryptionV2JitPasswordRegistration,
+    );
+
+    if (
+      accountEncryptionV2 &&
+      userType === SetInitialPasswordUserType.JIT_PROVISIONED_MP_ORG_USER
```
trmartin4: 🤔 Could we remove this check for `userType` here, and instead move any checks for specific flows within the `v2` method?

The thinking here being that at some point there will be a "v1 cleanup" effort, and as written the v2 method will not handle the other 3 `userType` flows, resulting in breaking changes during the cleanup.

Instead, perhaps keeping this check only for the feature flag, and then integrating the check for `userType` within the `setInitialPasswordV2` method, and also handling the other flows there, where you haven't made changes for v2 encryption (but still need to work once v1 is gone).

libs/angular/src/auth/password-management/set-initial-password/default-set-initial-password.service.implementation.ts
mzieniukbw: Moved encryption v2 code to `InitializeJitPasswordUserService`

libs/angular/src/auth/password-management/set-initial-password/default-set-initial-password.service.implementation.ts
mzieniukbw: Moved to `InitializeJitPasswordUserService` dedicated service, which is called in component.

apps/web/src/app/auth/core/services/password-management/set-initial-password/web-initialize-jit-password-user.service.ts
```diff
@@ -0,0 +1,47 @@
...
+  ) {}
+
+  async initializeUser(
+    credentials: InitializeJitPasswordCredentials,
+    userId: UserId,
+  ): Promise<void> {
+    await this.initializeJitPasswordUserService.initializeUser(credentials, userId);
+
+    /**
+     * TODO: Investigate refactoring the following logic in https://bitwarden.atlassian.net/browse/PM-22615
```
mzieniukbw: Note, i copied this from https://github.com/bitwarden/clients/blob/2531746952bb17c71b1dcd9848c09e18907318f8/apps/web/src/app/auth/core/services/password-management/set-initial-password/web-set-initial-password.service.ts#L63

libs/angular/src/auth/password-management/set-initial-password/default-set-initial-password.service.implementation.ts
mzieniukbw: This is removed with #18164, hence not needed. Removed.

libs/angular/src/auth/password-management/set-initial-password/default-set-initial-password.service.implementation.ts
mzieniukbw: Changed in d6b03f58

libs/angular/src/auth/password-management/set-initial-password/default-set-initial-password.service.implementation.ts
mzieniukbw: The PM-23246 points to wrong ticket. Changed in d6b03f58.

libs/angular/src/auth/password-management/set-initial-password/default-set-initial-password.service.implementation.ts
```diff
@@ -199,6 +214,134 @@ export class DefaultSetInitialPasswordService implements SetInitialPasswordServi
     }
   }
 
+  async setInitialPasswordTdeOffboarding(
```
mzieniukbw, line 216: Note: I haven't changed it, but i moved it up.

quexten (dismissed):

apps/desktop/src/app/services/set-initial-password/desktop-set-initial-password.service.ts
rr-bw, line 1: 1. Could you add your method here so we catch this functionality for Desktop? You can just duplicate the `setInitialPassword()` method that's already here, and rename to `initializePasswordJitPasswordUserV2Encryption()`.
2. Also duplicate tests in `desktop-set-initial-password.service.spec.ts`

[My PR](https://github.com/bitwarden/clients/pull/18006) will mark the old `setInitialPassword()` method and tests as deprecated for easy cleanup.

apps/web/src/app/auth/core/services/password-management/set-initial-password/web-set-initial-password.service.ts
rr-bw, line 1: 1. Could you add your method here so we catch this functionality for Web? You can just duplicate the `setInitialPassword()` method that's already here (including the giant comment), and rename to `initializePasswordJitPasswordUserV2Encryption()`.
2. Also duplicate tests in `web-set-initial-password.service.spec.ts`

[My PR](https://github.com/bitwarden/clients/pull/18006) will mark the old `setInitialPassword()` method and tests as deprecated for easy cleanup.

libs/angular/src/auth/password-management/set-initial-password/set-initial-password.component.ts
```diff
@@ -101,6 +105,49 @@ export class SetInitialPasswordComponent implements OnInit {
...
+      case SetInitialPasswordUserType.JIT_PROVISIONED_MP_ORG_USER: {
+        const accountEncryptionV2 = await this.configService.getFeatureFlag(
+          FeatureFlag.EnableAccountEncryptionV2JitPasswordRegistration,
+        );
+
+        if (accountEncryptionV2) {
+          await this.setInitialPasswordJitMPUserV2Encryption(passwordInputResult);
+        } else {
+          await this.setInitialPassword(passwordInputResult);
+        }
```
rr-bw, line 120: Could you use an early `return` here? [My following code](https://github.com/bitwarden/clients/pull/18006/changes#diff-c48c7f4163b1c34b9604f346d1762952c80a8bbed647cee596a37f225aa74dc1R210-R248) should then fit neatly under yours without a merge conflict due to the `else`.

```suggestion
        if (accountEncryptionV2) {
          await this.setInitialPasswordJitMPUserV2Encryption(passwordInputResult);
          return;
        }
        
        await this.setInitialPassword(passwordInputResult);
```

libs/angular/src/auth/password-management/set-initial-password/default-set-initial-password.service.implementation.ts
```diff
@@ -199,6 +213,132 @@ export class DefaultSetInitialPasswordService implements SetInitialPasswordServi
...
+    }
+    if (orgId == null) {
+      throw new Error("Organization id is required.");
+    }
+    if (newPassword == null) {
+      throw new Error("New password is required.");
+    }
+    if (salt == null) {
+      throw new Error("Salt is required.");
+    }
```
rr-bw: This suggestion will check all credentials, including `newPasswordHint` (which can have `""` as a [default & meaningful value](https://github.com/bitwarden/clients/blob/main/libs/auth/src/angular/input-password/input-password.component.ts#L307)), as well as `resetPasswordAutoEnroll`.

```suggestion
    for (const [key, value] of Object.entries(credentials)) {
      if (value == null) {
        throw new Error(`${key} is required.`);
      }
    }
```

libs/angular/src/auth/password-management/set-initial-password/default-set-initial-password.service.implementation.ts
```diff
@@ -199,6 +213,132 @@ export class DefaultSetInitialPasswordService implements SetInitialPasswordServi
...
+          using ref = sdk.take();
+          return await ref.value
+            .auth()
+            .registration()
+            .post_keys_for_jit_password_registration({
+              org_id: asUuid<SdkOrganizationId>(orgId),
+              org_public_key: organizationKeys.publicKey,
+              master_password: newPassword,
+              master_password_hint: newPasswordHint,
+              salt: salt,
```
rr-bw, line 294: ❓- On the SDK's `JitMasterPasswordRegistrationRequest`, this is `pub salt: String` ([see here](https://github.com/bitwarden/sdk-internal/blob/main/crates/bitwarden-auth/src/registration.rs#L67)). Why not `MasterPasswordSalt`? See @quexten's [related comment](https://github.com/bitwarden/sdk-internal/pull/549#discussion_r2644639803) on Jared's open PR. Do we want to do [PM-30183](https://bitwarden.atlassian.net/browse/PM-30183) now, (since it seems this flow JIT MP flow will actually be used soon), or just wait until Auth does it later?

libs/angular/src/auth/password-management/set-initial-password/default-set-initial-password.service.implementation.ts
```diff
@@ -199,6 +213,132 @@ export class DefaultSetInitialPasswordService implements SetInitialPasswordServi
...
+            throw new Error("SDK not available");
+          }
+
+          using ref = sdk.take();
+          return await ref.value
+            .auth()
+            .registration()
+            .post_keys_for_jit_password_registration({
+              org_id: asUuid<SdkOrganizationId>(orgId),
+              org_public_key: organizationKeys.publicKey,
```
rr-bw, line 291: ❓ - On the SDK's `JitMasterPasswordRegistrationRequest`, the [comment](https://github.com/bitwarden/sdk-internal/blob/main/crates/bitwarden-auth/src/registration.rs#L59-L61) for `org_public_key` says:

```rust
/// Organization's public key for encrypting the reset password key. This should be verified by
/// the client and not verifying may compromise the security of the user's account.
pub org_public_key: B64,
```

1. Does "client" here refer to TS client or SDK client?
2. What does "verifying" look like in this context? And do we need to do it here or SDK (or have we already done it)?

libs/angular/src/auth/password-management/set-initial-password/default-set-initial-password.service.implementation.ts
rr-bw: This suggestion will check all credentials, including `newPasswordHint` (which can have `""` as a [default & meaningful value](https://github.com/bitwarden/clients/blob/main/libs/auth/src/angular/input-password/input-password.component.ts#L307)), as well as `resetPasswordAutoEnroll`. (Then also add those 2 cases to test file).

```suggestion
    for (const [key, value] of Object.entries(credentials)) {
      if (value == null) {
        throw new Error(`${key} is required.`);
      }
    }
```

libs/angular/src/auth/password-management/set-initial-password/default-set-initial-password.service.implementation.ts
quexten: Client here refers to the TS client (or mobile if they implement this). Verifying can have two implementations, a UI fingerprint prompt, for which this dialog can be used: https://github.com/bitwarden/clients/blob/2cec30e5334723c8983c4efddd88dcf60239e6a3/libs/key-management-ui/src/trust/account-recovery-trust.component.ts#L47 - or alternatively you can implement your own component / this as part of your component. The fingerprint must be displayed along with some kind of text for the user that they should compare the fingerprint with their organizations administrator.

The other option (not yet implemented) is Key-transparency / AKD, which would be a KM-owned concern / is not relevant here.

libs/angular/src/auth/password-management/set-initial-password/default-set-initial-password.service.implementation.ts
quexten: From my perspective implementing it here, or delaying it for future cleanup is OK. The type safety would just be nice / gives better piece of mind.

libs/angular/src/auth/password-management/set-initial-password/default-set-initial-password.service.implementation.ts
```diff
@@ -199,6 +213,132 @@ export class DefaultSetInitialPasswordService implements SetInitialPasswordServi
...
+
+    const request = new UpdateTdeOffboardingPasswordRequest();
+    request.key = newMasterKeyEncryptedUserKey[1].encryptedString;
+    request.newMasterPasswordHash = newServerMasterKeyHash;
+    request.masterPasswordHint = newPasswordHint;
+
+    await this.masterPasswordApiService.putUpdateTdeOffboardingPassword(request);
+
+    // Clear force set password reason to allow navigation back to vault.
+    await this.masterPasswordService.setForceSetPasswordReason(ForceSetPasswordReason.None, userId);
```
quexten, line 253: Question: Is the master-key set to state here? I think we still need to for now and it is unclear to me where that happens.

mzieniukbw: That's within the `setInitialPasswordTdeOffboarding`, which i haven't changed, just moved it up to be closer to public functions.
This should be outside of scope for this work. Let me know if you want to discuss it further anyway.

apps/desktop/src/app/services/set-initial-password/desktop-set-initial-password.service.ts
mzieniukbw: Fixed in eef80188

apps/web/src/app/auth/core/services/password-management/set-initial-password/web-set-initial-password.service.ts
mzieniukbw: Fixed in eef80188

libs/angular/src/auth/password-management/set-initial-password/set-initial-password.component.ts
mzieniukbw: Fixed in 0af1c068

libs/angular/src/auth/password-management/set-initial-password/default-set-initial-password.service.implementation.ts
mzieniukbw: Fixed in 2bc32d5f

apps/desktop/src/app/services/set-initial-password/desktop-set-initial-password.service.spec.ts
```diff
@@ -179,4 +185,36 @@ describe("DesktopSetInitialPasswordService", () => {
...
+        orgId: "orgId" as OrganizationId,
+        resetPasswordAutoEnroll: false,
+        newPassword: "newPassword123!",
+        salt: "user@example.com" as MasterPasswordSalt,
+      };
+      const userId = "userId" as UserId;
+
+      const superSpy = jest
+        .spyOn(
+          Object.getPrototypeOf(Object.getPrototypeOf(sut)),
```
quexten: Nit: Could we put a note/comment on what this does / why it is needed? Are we saying we want to spy on the parent - DefaultSetInitialPasswordService, since we override the implementation here but still call the super implementation / want to ensure it is called?

mzieniukbw: Yes, we want to mock the parent function. I though the `superSpy` variable name suggested what we want to do. 
Let me know if you still want to add a comment explaining that.

quexten: I'm good with this, I just never saw this pattern before / "getPrototypeOf" required some thinking for me.

mzieniukbw: I haven't realised i can just grab the `DefaultSetInitialPasswordService.prototype` here, without unnecessary `getPrototypeOf` complications. Fixed in 441985d8

rr-bw (approved):

trmartin4:
@coroiu would you mind reviewing this on behalf of Platform?  I was assigned as the reviewer by Github because I'd made some suggestions on the PR for Auth, but I don't feel comfortable being the arbiter of Platform's changes here, even if they are minimal.

quexten (approved):

PR #18274, opened 2026-01-08
[PM-27331] Update the sdk service to use accountCryptographicState

Thomas-Avery:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-27331

## 📔 Objective

After https://github.com/bitwarden/clients/pull/18164 we will have consolidated a users individual account cryptographic state into a single observable.

This PR updates the SDK service to subscribe to that individual observable instead of using the individual state subscription for each property.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


dani-garcia (approved):
Platform changes LGTM

libs/common/src/platform/services/sdk/default-sdk.service.ts
```diff
@@ -106,7 +106,7 @@ export class DefaultSdkService implements SdkService {
     private accountService: AccountService,
     private kdfConfigService: KdfConfigService,
     private keyService: KeyService,
-    private securityStateService: SecurityStateService,
```
quexten, line 109: Looks like security state service is now unused entirely. We can drop it in a follow-up PR.

Thomas-Avery: It has a lot of unused imports which is odd. I do see one use in the user key rotation service https://github.com/bitwarden/clients/blob/bd3f8dd4c13717b771620a1f30b420c973518a57/apps/web/src/app/key-management/key-rotation/user-key-rotation.service.ts#L641.

quexten: Weird, they did not show up in my IDE. In that case we should wait a bit more, though we can either replace calling instances with calls to account cryptography state, or in the case of user-key rotation probably just drop the TS implementation entirely since we are moving to SDK.

PR #18157, opened 2025-12-31
[PM-30319] [BLOCKER] phish cache freeze

AlexRubik:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-30319

## 📔 Objective

**Problem:** Browser extension UI freezes when switching to accounts with phishing detection access. The synchronous cache update blocks the service worker thread. First login after install causes a 5+ second freeze while downloading the 780K-entry phishing list.

**Solution:** Decouple cache updates from the UI path, pre-populate cache on install/update, and optimize processing to prevent blocking.

### Key Changes

- **Pre-population:** Cache downloads on install/update so it's ready before user login
- **Non-blocking updates:** Decouple `update$` from UI event handlers using `delay(0)` and separate subscription
- **Checksum optimization:** Skip expensive state writes when checksum matches (returns `null`)
- **In-memory Set cache:** Avoid rebuilding 780K-entry Set on every state emission
- **Streaming fetch:** Process large files in chunks with event loop yields to prevent Firefox freezes
- **Chunked Set building:** Process 50K entries per chunk with yields to keep UI responsive
- **Lazy loading:** Defer IndexedDB reads until phishing detection is actually used
- **Update throttling:** 5-minute minimum interval and `_updateInProgress` flag prevent redundant/concurrent updates
- **Source update:** Switch to `phish.co.za/latest/phishing-links-ACTIVE.txt` (new upstream provider)

The service worker is single-threaded. Previously, cache updates ran synchronously in the UI initialization flow. Now:
1. Cache pre-populates on install/update — no download delay on first login
2. `initialize()` completes immediately
3. Cache updates run asynchronously in the background
4. Checksum matching skips state writes when data is unchanged
5. Streaming + chunking keeps UI responsive during heavy processing

## ⏰ Reminders before review

- Contributor guidelines followed
- All formatters and local linters executed and passed
- Written new unit and / or integration tests where applicable
- Protected functional changes with optionality (feature flags)
- Used internationalization (i18n) for all UI strings
- CI builds passed
- Communicated to DevOps any deployment requirements
- Updated any necessary documentation (Confluence, contributing docs) or informed the documentation team

## 🦮 Reviewer guidelines

- 👍 (`:+1:`) or similar for great changes
- 📝 (`:memo:`) or ℹ️ (`:information_source:`) for notes or general info
- ❓ (`:question:`) for questions
[...]

PR #18303, opened 2026-01-09
[PM-30629] Split phishing cache into separate data and metadata states (Firefox Phish Freeze)

AlexRubik:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-30629

## 📔 Objective

**Problem**
Firefox freezes for 5+ seconds when switching from a non-premium to premium account. The freeze occurs because even when the phishing cache data hasn't changed (checksum match), writing to state triggers a synchronous rebuild of a 780,000-entry Set.

**Solution**
Split the phishing cache into two separate state keys:
- `PHISHING_DATA_KEY` - stores webAddresses (expensive to update - triggers Set rebuild)
- `PHISHING_METADATA_KEY` - stores timestamp, checksum, applicationVersion (cheap to update)

The `_webAddresses$` observable now only subscribes to the data state. Metadata updates (timestamp refresh on checksum match) no longer trigger the expensive Set rebuild.

**Before (causes freeze):**
1. `getNextWebAddresses()` returns `{ ...prev, timestamp, applicationVersion }`
2. State write triggers `state$` emission
3. `_webAddresses$` rebuilds Set from 780K entries synchronously
4. **Firefox freezes**

**After (split state):**
1. `getNextWebAddresses()` returns `{ metadata, webAddresses? }`
2. Metadata state always updates (cheap)
3. Data state only updates when `webAddresses` is returned (checksum differs)
4. `_webAddresses$` only reacts to data changes → **No freeze on checksum match**

Additionally, if the remote checksum fetch returns empty/falsy, we preserve the existing checksum rather than overwriting it.

---

The underlying problem remains: synchronously building a Set from 780K entries may block the thread. Cases where the freeze might still happen:
- Data actually changed (checksum differs)
- New app version (forces full refetch)
- Cache older than 24 hours

The real fix would be to make the Set building asynchronous (chunked with yields to event loop), which was part of PR #18157. But that PR introduced other issues.

So this fix eliminates the most common freeze scenario (account switch when cache is fresh), but doesn't solve the root architectural problem.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/apps/browser/src/dirt/phishing-detection/services/phishing-data.service.ts
+++ b/apps/browser/src/dirt/phishing-detection/services/phishing-data.service.ts
@@ -1,5 +1,6 @@
 import {
   catchError,
+  combineLatest,
   EMPTY,
   first,
   firstValueFrom,
@@ -20,11 +21,15 @@ import { GlobalStateProvider, KeyDefinition, PHISHING_DETECTION_DISK } from "@bi
 
 import { getPhishingResources, PhishingResourceType } from "../phishing-resources";
 
+/** Stores the actual phishing web addresses (expensive to update - triggers Set rebuild) */
 export type PhishingData = {
   webAddresses: string[];
+};
+
+/** Stores metadata about the cache (cheap to update - no Set rebuild) */
+export type PhishingMetadata = {
   timestamp: number;
   checksum: string;
-
   /**
    * We store the application version to refetch the entire dataset on a new client release.
    * This counteracts daily appends updates not removing inactive or false positive web addresses.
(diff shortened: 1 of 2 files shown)


apps/browser/src/dirt/phishing-detection/services/phishing-data.service.ts
```diff
@@ -127,13 +127,11 @@ export class PhishingDataService {
...
-    // If checksum matches, return existing data with new timestamp & version
+    // If checksum matches, skip state write to avoid expensive Set rebuild
     const remoteChecksum = await this.fetchPhishingChecksum(this.resourceType);
     if (remoteChecksum && prev.checksum === remoteChecksum) {
-      this.logService.info(
-        `[PhishingDataService] Remote checksum matches local checksum, updating timestamp only.`,
-      );
-      return { ...prev, timestamp, applicationVersion };
+      this.logService.info(`[PhishingDataService] Checksum match, skipping update.`);
+      return null;
```
Banrion: ❓ This does not return timestamp and app version if checksum matches. Do those variables get updated in any other location?

Issue #16234, opened 2025-08-31
Missing hashes in package-lock.json


dotlambda:
### Steps To Reproduce

In `package-lock.json`, some dependencies are missing a hash which prevents us from building bitwarden-cli offline in Nixpkgs: https://github.com/NixOS/nixpkgs/issues/339576

### Expected Result

Every dependency in the lock file should have the `resolved` and `integrity` fields.

### Actual Result

See e.g. `apps/cli/node_modules/open` which lacks the `resolved` and `integrity` fields.

### Additional Context

https://github.com/bitwarden/clients/blob/e4c75b3c49a0c5ec921e58272398518dd9eb7b77/package-lock.json#L263C6-L277

### Operating System

Linux

### Shell

Bash

### Build Version

2025.8.0

### Issue Tracking Info

- [x] I understand that work is tracked outside of Github. A PR will be linked to this issue should one be opened to address it, but Bitwarden doesn't use fields like "assigned", "milestone", or "project" to track progress.

Issue #14121, opened 2025-04-04
failed to run 2025.3.0 build


chenrui333:
### Steps To Reproduce

```
Error: Cannot find module 'core-js/proposals/explicit-resource-management' Require stack:
```

- relates to https://github.com/Homebrew/homebrew-core/pull/218125

### Build Version

2025.3.0

### Issue Tracking Info

- [x] I understand that work is tracked outside of Github. A PR will be linked to this issue should one be opened to address it, but Bitwarden doesn't use fields like "assigned", "milestone", or "project" to track progress.

S-Kakar:
Thank you for reporting this issue! We've added this to our internal tracking system.
ID: PM-19878

rmcdowell-bitwarden:
Hi there,

I am unable to reproduce this issue, it has been escalated for further investigation. If you have more information that can help us, please add it below. 

Thanks!

daeho-ro:
Hello, I have noticed that there is no `core-js` dependency in the cli package.json. It is only in the clients root, and so we'v got the module not found error after building cli only.

Can you add it?
Thanks,

daeho-ro:
Furthermore, there is a bug on `argon2` when it build on macos 15, so it should be updated to > 0.43.0.

Hinton:
Hi, this should be resolved in the next CLI release.

PR #18308, opened 2026-01-10
[PM-30655] fix: missing hashes in package-lock.json

xiaoxiangmoe:
## 🎟️ Tracking

close https://github.com/bitwarden/clients/issues/16234

## 📔 Objective

fix lockfile

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


xiaoxiangmoe:
See also
```
https://github.com/NixOS/nixpkgs/pull/448577#discussion_r2404267452
```

xiaoxiangmoe:
@coroiu friendly ping

coroiu:
@xiaoxiangmoe Could you expand a bit on the changes you are proposing? The title of the PR says that is fixes missing hashes but the change just removes some packages from the lock file. From what I can see those packages are still present in our `package.json` files, why are you removing them from from the lockfile?

dotlambda:
> The title of the PR says that is fixes missing hashes but the change just removes some packages from the lock file.

The packages that are being removed lack the `resolved` and `integrity` fields. See https://github.com/bitwarden/clients/issues/16234.
Thus the project cannot be built reproducibly as is.



> From what I can see those packages are still present in our `package.json` files, why are you removing them from from the lockfile?

The CLI can be built without those because, afaict, they are just duplicates

coroiu:
@dotlambda Ok I think I follow now, unfortunately we need to explicitly duplicate the CLI dependencies or it will break the CLI when installed with NPM. I made a similar mistake when I added `core-js` to the root `package.json` but not the CLI `package.json` which caused https://github.com/bitwarden/clients/issues/14121

dotlambda:
> we need to explicitly duplicate the CLI dependencies

That's fine. But then you should regenerate the lock file so that it includes hashes for everything.

dotlambda:
> I made a similar mistake when I added `core-js` to the root `package.json` but not the CLI `package.json`

Are you sure that adding it to both `package.json` files causes the package to be listed twice in `package-lock.json`?

coroiu:
> Are you sure that adding it to both package.json files causes the package to be listed twice in package-lock.json?

@dotlambda No, sorry, that's not the issue. The issue is that we publish the CLI to NPM using `cli/package.json` which means that anyone that downloads the CLI using NPM will only install the dependencies found in that file.

Ooh now I get what you mean by duplicate, the global `package-lock.json` contains 2 copies of those packages, one from the root `package.json` and one from `cli/package.json`. That seems unnecessary but I'll double check, hang on :)

coroiu:
We have another task (automatic renovate PR) in progress to update the lockfile, I want to get that merged and then come back to this PR

coroiu:
@xiaoxiangmoe Sorry for taking so long to get back to you. It seems this PR surfaced a versioning bug that the redundant package-lock entries were hiding. Someone else fixed that bug and that resolution ended up including the removal of the packages from the lockfile. Therefore I am closing this as solved by: https://github.com/bitwarden/clients/commit/3c9569a90f491e702b4a01534d79ab168e76aaf0 Please let me know if you don't agree!

PR #18322, opened 2026-01-12
[PM-30677] Convert SendSearchComponent to OnPush

Hinton:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-30677
## 📔 Objective

Converts SendSearchComponent to use OnPush change detection.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


libs/tools/send/send-ui/src/send-list/send-list.component.ts
```diff
@@ -70,7 +59,7 @@ export class SendListComponent {
   );
 
   protected readonly noSearchResults = computed(
-    () => this.showSearchBar() && (this.sends().length === 0 || this.searchText().length > 0),
+    () => this.showSearchBar() && this.sends().length === 0,
```
Hinton, line 62: @iivins-livefront not sure why we had `(this.sends().length === 0 || this.searchText().length > 0)` here? It seems wrong to show the no result IF we're searching OR no send items.

iivins-livefront: > It seems wrong to show the no result IF we're searching OR no send items.

There's 2 caveats here:
1. The search bar is hiding/showing if no Sends exist for that filter ( determined by the SideNav ). The `this.searchText().length > 0` check is used to see if the Search Bar is also filtering Sends in addition to the SideNav filtering Sends ( the SearchBar displaying isn't a guarantee )
2. There's a "No Search Results" state and a "No Sends for this SideNav filter" state

If we remove that piece of code and don't change anything on the html file, noSearchResults is only checking if:
- the search bar is currently being displayed
- no Sends are being returned after the SideNav filter

^it misses the "did the Search Bar or SideNav cause the table to be empty" - and we'd have to update the html to make that check instead

Hinton: I'm not quite following, the previous code causes the no results to appear whenever you search and have a match, image

Since `true && (false || true) == true`. We only want to show the no result if we have no items right?

iivins-livefront: OH! - that state in the screenshot shouldn't be happening... if your code changes remove it then it should stay

What I was trying to explain poorly, is that there are 2 empty states that conditional helps determine - this bug/miss isn't that

willmartian:
@claude it looks like [4f35d37](https://github.com/bitwarden/clients/pull/18322/commits/4f35d37ce0033552a67e76d1c9db92eff2d32c93) may have resolved a merge conflict incorrectly. Could you double check this? The build is now erroring.

willmartian:
@claude The intention was to remove the CDR workaround since the mentioned component is now standalone

PR #18344, opened 2026-01-13
[PM-30718] add IndexedDB storage service for phishing data

AlexRubik:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-30718

## 📔 Objective

Creates a dedicated IndexedDB storage service for phishing data. This service will be used in a follow-up PR to replace local storage.

This PR adds the service and tests only - no integration with existing phishing code yet.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/.github/workflows/build-web.yml
+++ b/.github/workflows/build-web.yml
@@ -112,7 +112,7 @@ jobs:
             npm_command: dist:bit:selfhost
           - artifact_name: selfhosted-DEV
             license_type: "commercial"
-            image_name: web
+            image_name: web-dev
             npm_command: build:bit:selfhost:dev
             git_metadata: true
           - artifact_name: cloud-QA
--- a/apps/browser/src/_locales/en/messages.json
+++ b/apps/browser/src/_locales/en/messages.json
@@ -2473,6 +2473,9 @@
   "permanentlyDeletedItem": {
     "message": "Item permanently deleted"
   },
+  "archivedItemRestored": {
+    "message": "Archived item restored"
+  },
   "restoreItem": {
     "message": "Restore item"
   },
@@ -5669,6 +5672,9 @@
   },
   "extraWide": {
     "message": "Extra wide"
+  },
+  "narrow": {
+    "message": "Narrow"
   },
   "sshKeyWrongPassword": {
     "message": "The password you entered is incorrect."
--- a/apps/browser/src/autofill/fido2/content/messaging/messenger.ts
+++ b/apps/browser/src/autofill/fido2/content/messaging/messenger.ts
@@ -1,3 +1,5 @@
+// FIXME: Update this file to be type safe and remove this and next line
+// @ts-strict-ignore
 import { Message, MessageTypes } from "./message";
 
 const SENDER = "bitwarden-webauthn";
@@ -23,7 +25,7 @@ type Handler = (
  * handling aborts and exceptions across separate execution contexts.
  */
 export class Messenger {
-  private messageEventListener: ((event: MessageEvent<MessageWithMetadata>) => void) | null = null;
+  private messageEventListener: (event: MessageEvent<MessageWithMetadata>) => void | null = null;
   private onDestroy = new EventTarget();
 
   /**
@@ -58,12 +60,6 @@ export class Messenger {
     this.broadcastChannel.addEventListener(this.messageEventListener);
   }
 
-  private stripMetadata({ SENDER, senderId, ...message }: MessageWithMetadata): Message {
-    void SENDER;
-    void senderId;
-    return message;
-  }
-
   /**
    * Sends a request to the content script and returns the response.
    * AbortController signals will be forwarded to the content script.
@@ -78,9 +74,7 @@ export class Messenger {
 
     try {
       const promise = new Promise<Message>((resolve) => {
-        localPort.onmessage = (event: MessageEvent<MessageWithMetadata>) => {
-          resolve(this.stripMetadata(event.data));
-        };
+        localPort.onmessage = (event: MessageEvent<MessageWithMetadata>) => resolve(event.data);
       });
 
       const abortListener = () =>
@@ -135,9 +129,7 @@ export class Messenger {
 
       try {
         const handlerResponse = await this.handler(message, abortController);
-        if (handlerResponse !== undefined) {
-          port.postMessage({ ...handlerResponse, SENDER });
-        }
+        port.postMessage({ ...handlerResponse, SENDER });
       } catch (error) {
         port.postMessage({
           SENDER,
--- a/apps/browser/src/platform/browser/browser-popup-utils.spec.ts
+++ b/apps/browser/src/platform/browser/browser-popup-utils.spec.ts
@@ -1,7 +1,7 @@
 import { createChromeTabMock } from "../../autofill/spec/autofill-mocks";
 
 import { BrowserApi } from "./browser-api";
-import BrowserPopupUtils from "./browser-popup-utils";
+import BrowserPopupUtils, { PopupWidthOptions } from "./browser-popup-utils";
 
 describe("BrowserPopupUtils", () => {
   afterEach(() => {
@@ -152,7 +152,7 @@ describe("BrowserPopupUtils", () => {
         focused: false,
         alwaysOnTop: false,
         incognito: false,
-        width: 380,
+        width: PopupWidthOptions.default,
       });
       jest.spyOn(BrowserApi, "createWindow").mockImplementation();
       jest.spyOn(BrowserApi, "updateWindowProperties").mockImplementation();
@@ -168,7 +168,7 @@ describe("BrowserPopupUtils", () => {
       expect(BrowserApi.createWindow).toHaveBeenCalledWith({
         type: "popup",
         focused: true,
-        width: 380,
+        width: PopupWidthOptions.default,
         height: 630,
         left: 85,
         top: 190,
@@ -197,7 +197,7 @@ describe("BrowserPopupUtils", () => {
       expect(BrowserApi.createWindow).toHaveBeenCalledWith({
         type: "popup",
         focused: true,
-        width: 380,
+        width: PopupWidthOptions.default,
         height: 630,
         left: 85,
         top: 190,
@@ -214,7 +214,7 @@ describe("BrowserPopupUtils", () => {
       expect(BrowserApi.createWindow).toHaveBeenCalledWith({
         type: "popup",
         focused: true,
-        width: 380,
+        width: PopupWidthOptions.default,
         height: 630,
         left: 85,
         top: 190,
@@ -267,7 +267,7 @@ describe("BrowserPopupUtils", () => {
       expect(BrowserApi.createWindow).toHaveBeenCalledWith({
         type: "popup",
         focused: true,
-        width: 380,
+        width: PopupWidthOptions.default,
         height: 630,
         left: 85,
         top: 190,
@@ -290,7 +290,7 @@ describe("BrowserPopupUtils", () => {
         focused: false,
         alwaysOnTop: false,
         incognito: false,
-        width: 380,
+        width: PopupWidthOptions.default,
         state: "fullscreen",
       });
       jest
@@ -321,7 +321,7 @@ describe("BrowserPopupUtils", () => {
         focused: false,
         alwaysOnTop: false,
         incognito: false,
-        width: 380,
+        width: PopupWidthOptions.default,
         state: "fullscreen",
       });
 
--- a/apps/browser/src/platform/browser/browser-popup-utils.ts
+++ b/apps/browser/src/platform/browser/browser-popup-utils.ts
@@ -10,9 +10,9 @@ import { BrowserApi } from "./browser-api";
  * Value represents width in pixels
  */
 export const PopupWidthOptions = Object.freeze({
-  default: 380,
-  wide: 480,
-  "extra-wide": 600,
+  default: 480,
+  wide: 600,
+  narrow: 380,
 });
 
 type PopupWidthOptions = typeof PopupWidthOptions;
--- a/apps/browser/src/platform/popup/layout/popup-layout.stories.ts
+++ b/apps/browser/src/platform/popup/layout/popup-layout.stories.ts
@@ -44,7 +44,7 @@ import { PopupTabNavigationComponent } from "./popup-tab-navigation.component";
 @Component({
   selector: "extension-container",
   template: `
-    <div class="tw-h-[640px] tw-w-[380px] tw-border tw-border-solid tw-border-secondary-300">
+    <div class="tw-h-[640px] tw-w-[480px] tw-border tw-border-solid tw-border-secondary-300">
       <ng-content></ng-content>
     </div>
   `,
@@ -678,7 +678,7 @@ export const WidthOptions: Story = {
     template: /* HTML */ `
       <div class="tw-flex tw-flex-col tw-gap-4 tw-text-main">
         <div>Default:</div>
-        <div class="tw-h-[640px] tw-w-[380px] tw-border tw-border-solid tw-border-secondary-300">
+        <div class="tw-h-[640px] tw-w-[480px] tw-border tw-border-solid tw-border-secondary-300">
           <mock-vault-page></mock-vault-page>
         </div>
         <div>Wide:</div>
--- a/apps/browser/src/platform/popup/layout/popup-size.service.ts
+++ b/apps/browser/src/platform/popup/layout/popup-size.service.ts
@@ -83,7 +83,7 @@ export class PopupSizeService {
     }
     const pxWidth = PopupWidthOptions[width] ?? PopupWidthOptions.default;
 
-    document.body.style.minWidth = `${pxWidth}px`;
+    document.body.style.width = `${pxWidth}px`;
   }
 
   /**
--- a/apps/browser/src/popup/components/extension-anon-layout-wrapper/extension-anon-layout-wrapper.stories.ts
+++ b/apps/browser/src/popup/components/extension-anon-layout-wrapper/extension-anon-layout-wrapper.stories.ts
@@ -10,6 +10,7 @@ import {
 import { of } from "rxjs";
 
 import { LockIcon, RegistrationCheckEmailIcon } from "@bitwarden/assets/svg";
+import { PopupWidthOptions } from "@bitwarden/browser/platform/browser/browser-popup-utils";
 import { AccountService } from "@bitwarden/common/auth/abstractions/account.service";
 import { AuthService } from "@bitwarden/common/auth/abstractions/auth.service";
 import { AvatarService } from "@bitwarden/common/auth/abstractions/avatar.service";
@@ -243,7 +244,12 @@ export const DefaultContentExample: Story = {
   }),
   parameters: {
     chromatic: {
-      viewports: [380, 1280],
+      viewports: [
+        PopupWidthOptions.default,
+        PopupWidthOptions.narrow,
+        PopupWidthOptions.wide,
+        1280,
+      ],
     },
   },
 };
--- a/apps/browser/src/popup/scss/tailwind.css
+++ b/apps/browser/src/popup/scss/tailwind.css
@@ -60,7 +60,7 @@
   }
 
   body {
-    width: 380px;
+    width: 480px;
     height: 100%;
     position: relative;
     min-height: inherit;
@@ -84,9 +84,9 @@
     animation: redraw 1s linear infinite;
   }
 
-  /** 
+  /**
    * Text selection style:
-   * suppress user selection for most elements (to make it more app-like) 
+   * suppress user selection for most elements (to make it more app-like)
    */
   h1,
   h2,
@@ -165,7 +165,7 @@
     @apply tw-text-muted;
   }
 
-  /** 
+  /**
    * Text selection style:
    * Set explicit selection styles (assumes primary accent color has sufficient
    * contrast against the background, so its inversion is also still readable)
--- a/apps/browser/src/vault/popup/components/vault-v2/autofill-confirmation-dialog/autofill-confirmation-dialog.component.spec.ts
+++ b/apps/browser/src/vault/popup/components/vault-v2/autofill-confirmation-dialog/autofill-confirmation-dialog.component.spec.ts
@@ -91,10 +91,18 @@ describe("AutofillConfirmationDialogComponent", () => {
     jest.resetAllMocks();
   });
 
-  const findShowAll = (inFx?: ComponentFixture<AutofillConfirmationDialogComponent>) =>
-    (inFx || fixture).nativeElement.querySelector(
-      "button.tw-text-sm.tw-font-medium.tw-cursor-pointer",
-    ) as HTMLButtonElement | null;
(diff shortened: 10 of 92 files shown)


apps/browser/src/dirt/phishing-detection/services/phishing-indexeddb.service.ts
```diff
@@ -0,0 +1,137 @@
...
+      db.close();
+      return true;
+    } catch (error) {
+      this.logService.error("[PhishingIndexedDbService] Failed to clear data", error);
+      return false;
+    }
+  }
+
+  /** Opens the database, creating the object store if needed. */
+  private openDatabase(): Promise<IDBDatabase> {
```
Banrion: ❓ How heavy is open database? Is this something we want to do once in the constructor or is it suggested to do this every time you need to make a modification?

apps/browser/src/dirt/phishing-detection/services/phishing-indexeddb.service.ts
```diff
@@ -0,0 +1,137 @@
...
+      };
+
+      request.onsuccess = () => {
+        resolve();
+      };
+    });
+  }
+
+  /** Retrieves data from the object store. */
+  private getData(db: IDBDatabase): Promise<PhishingData | null> {
```
Banrion: ⚠️ Rather than fetching a row containing the entire data set, we can use a [cursor](https://developer.mozilla.org/en-US/docs/Web/API/IndexedDB_API/Using_IndexedDB#using_a_cursor) over the values that we have added to individual rows.

⚠️ We should also have an [index](https://developer.mozilla.org/en-US/docs/Web/API/IndexedDB_API/Using_IndexedDB#using_an_index) over the links (websites). This will allow for some optimizing of the cursor via [range and directions](https://developer.mozilla.org/en-US/docs/Web/API/IndexedDB_API/Using_IndexedDB#specifying_the_range_and_direction_of_cursors).

apps/browser/src/dirt/phishing-detection/services/phishing-indexeddb.service.ts
```diff
@@ -0,0 +1,137 @@
...
+    });
+  }
+
+  /** Stores data in the object store. */
+  private putData(db: IDBDatabase, data: PhishingData): Promise<void> {
+    return new Promise((resolve, reject) => {
+      const transaction = db.transaction(this.STORE_NAME, "readwrite");
+      const store = transaction.objectStore(this.STORE_NAME);
+      const request = store.put(data, this.DATA_KEY);
+
```
Banrion: ⚠️ We most likely don't want to save the data to one row like we are doing for browser storage right now. The benefit of using the IndexDb is that we can create our own dataset for just phishing websites in the browser. Each url can be a row in this database.

apps/browser/src/dirt/phishing-detection/services/phishing-indexeddb.service.ts
Banrion: 💭 If we ever need to go back to domains as the source, I can see a use case for adding an [index](https://developer.mozilla.org/en-US/docs/Web/API/IndexedDB_API/Using_IndexedDB#using_an_index) over the domain portion and breaking this into different rows via the domain as the key.

apps/browser/src/dirt/phishing-detection/services/phishing-indexeddb.service.ts
```diff
@@ -0,0 +1,137 @@
...
+        const db = (event.target as IDBOpenDBRequest).result;
+        if (!db.objectStoreNames.contains(this.STORE_NAME)) {
+          db.createObjectStore(this.STORE_NAME);
+        }
+      };
+    });
+  }
+
+  /** Stores data in the object store. */
+  private putData(db: IDBDatabase, data: PhishingData): Promise<void> {
```
Banrion: ⚠️ We should just be saving the websites in this database. Not the meta data, which can be quickly accessed and synced in the StateProvider across background and client scripts.

apps/browser/src/dirt/phishing-detection/services/phishing-indexeddb.service.ts
```diff
@@ -0,0 +1,137 @@
...
+
+/**
+ * Provides IndexedDB storage for phishing detection data.
+ *
+ */
+export class PhishingIndexedDbService {
+  private readonly DB_NAME = "bitwarden-phishing";
+  private readonly STORE_NAME = "phishing-data";
+  private readonly DATA_KEY = "phishing-domains";
+  private readonly DB_VERSION = 1;
```
Banrion, line 15: 💭 I don't think it's needed at this moment since we don't expect the data type to change from being a string type value per key, but we should consider making a type that captures this information if we expand upon IndexedDB in a more generic way

apps/browser/src/dirt/phishing-detection/services/phishing-indexeddb.service.ts
```diff
@@ -0,0 +1,137 @@
+import { LogService } from "@bitwarden/common/platform/abstractions/log.service";
+
+import { PhishingData } from "./phishing-data.service";
+
+/**
+ * Provides IndexedDB storage for phishing detection data.
+ *
+ */
+export class PhishingIndexedDbService {
```
Banrion, line 12: 💭 If we expand on this in a generic way, we can create a basic abstract IndexedDbService to provide basic setup information, such as versioning controls, names, etc.

apps/browser/src/dirt/phishing-detection/services/phishing-indexeddb.service.ts
```diff
@@ -0,0 +1,137 @@
...
+  private readonly DATA_KEY = "phishing-domains";
+  private readonly DB_VERSION = 1;
+
+  constructor(private logService: LogService) {}
+
+  /**
+   * Saves phishing data to IndexedDB.
+   * @returns true on success, false on error
+   */
+  async save(data: PhishingData): Promise<boolean> {
```
Banrion: ⚠️ Note that we currently use `response.body` when downloading the file from Phishing.Database. We should save the file directly with the stream to reduce download and saving load on the client if we can. 

💭 I almost suggested having a version for both methods of saving large data but I think keeping this specific to how the phishing detection is currently setup is best to reduce bloat.

apps/browser/src/dirt/phishing-detection/services/phishing-indexeddb.service.ts
```diff
@@ -0,0 +1,137 @@
...
+      };
+
+      request.onsuccess = () => {
+        resolve(request.result ?? null);
+      };
+    });
+  }
+
+  /** Deletes data from the object store. */
+  private deleteData(db: IDBDatabase): Promise<void> {
```
Banrion: ⚠️ While a delete is nice to have for possible uses, we should keep this specific to exactly what phishing detection uses for now and let an abstract class or documentation guide generic usage. We can remove this (and the clear function) unless this is needed somewhere.

Banrion (changes requested):
If I'm understanding correctly, we want to utilize IndexedDB to store the large list of phishing websites (either domains or links) in one database. The database is to store the list, so we don't want one entry in the database for the entire list. We want to utilize the cursor and index to make the performant lookup and reduce load in memory when checking for a phishing website. There are some changes needed to fully capture this. 

We also don't want to store the meta data in this database. The meta data is best stored in the state provider for syncing across background and client scripts. 

Once those changes are made, I think this is a good setup for specific usage of IndexedDB for storing phishing domains. We can consider an abstract class to guide others on setup, in a future iteration.

apps/browser/src/dirt/phishing-detection/services/phishing-indexeddb.service.ts
AlexRubik: It's not heavy afaik. Each public method will open the db, execute logic, then close db.
https://github.com/bitwarden/clients/commit/e486fd4a877847cd4b9fd97930f7cf2b9a27b832

apps/browser/src/dirt/phishing-detection/services/phishing-indexeddb.service.ts
AlexRubik: Implemented hasUrl() for direct lookups via keyPath: "url" (automatic index) and added loadAllUrls() with cursor iteration as fallback if direct lookups are too slow.

- [`4d87995`](https://github.com/bitwarden/clients/commit/4d8799570997113a3ae93fb03e75c9d65e693fa6) - `hasUrl()` for direct index lookup
- [`d480b81`](https://github.com/bitwarden/clients/commit/d480b8141c44feaf5c90cf3bd3de11a48d6c53c2) - `loadAllUrls()` with cursor iteration

apps/browser/src/dirt/phishing-detection/services/phishing-indexeddb.service.ts
AlexRubik: Each URL is now stored as its own row with keyPath: "url"
https://github.com/bitwarden/clients/pull/18344/commits/1d33cadf8bbc181b989c3317523cea3c04cbc1a5

apps/browser/src/dirt/phishing-detection/services/phishing-indexeddb.service.ts
AlexRubik: If we ever want to do this, we can add the logic for it in the openDatabase() method in the `req.onupgradeneeded` case.
https://github.com/bitwarden/clients/pull/18344/commits/e486fd4a877847cd4b9fd97930f7cf2b9a27b832

apps/browser/src/dirt/phishing-detection/services/phishing-indexeddb.service.ts
AlexRubik: https://github.com/bitwarden/clients/pull/18344/commits/1d33cadf8bbc181b989c3317523cea3c04cbc1a5

apps/browser/src/dirt/phishing-detection/services/phishing-indexeddb.service.ts
AlexRubik: [`1d33cad`](https://github.com/bitwarden/clients/commit/1d33cadf8bbc181b989c3317523cea3c04cbc1a5) - adds `PhishingUrlRecord` type

apps/browser/src/dirt/phishing-detection/services/phishing-indexeddb.service.ts
AlexRubik: [`47ad360`](https://github.com/bitwarden/clients/commit/47ad360d359e92e27a5e6b965c1c89c1470931c5) - `saveUrlsFromStream()`

apps/browser/src/dirt/phishing-detection/services/phishing-indexeddb.service.ts
AlexRubik: Removed public `clear()` and `delete()` methods

apps/browser/src/dirt/phishing-detection/services/phishing-indexeddb.service.spec.ts
```diff
@@ -0,0 +1,358 @@
...
+import { LogService } from "@bitwarden/common/platform/abstractions/log.service";
+
+import { PhishingIndexedDbService } from "./phishing-indexeddb.service";
+
+describe("PhishingIndexedDbService", () => {
+  let service: PhishingIndexedDbService;
+  let logService: MockProxy<LogService>;
+
+  // Mock IndexedDB storage (keyed by URL for row-per-URL storage)
+  let mockStore: Map<string, { url: string }>;
```
Banrion, line 14: 💭 I think we should make a quick iteration on this to use [fake-indexeddb](https://www.npmjs.com/package/fake-indexeddb). I say iterate because we don't want our feature blocking from another team's review. We can merge this, start integration and performance testing and iterate further.

apps/browser/src/dirt/phishing-detection/services/phishing-indexeddb.service.spec.ts
```diff
@@ -0,0 +1,358 @@
...
+    service = new PhishingIndexedDbService(logService);
+  });
+
+  afterEach(() => {
+    jest.clearAllMocks();
+  });
+
+  describe("saveUrls", () => {
+    it("stores URLs in IndexedDB and returns true", async () => {
+      const urls = ["phishing.com", "malware.net"];
```
Banrion: ⚠️ All of our urls should follow the url syntax due to the constraints of the phishing.database. While we normalize urls we receive in phishing detection, we should keep test data as close to the real data as possible. Therefor, we should be using https or http in our saves and stored test data.

See [this](https://developer.mozilla.org/en-US/docs/Learn_web_development/Howto/Web_mechanics/What_is_a_URL) Mozilla doc and the [url standard docs](https://url.spec.whatwg.org/#absolute-url-string).

apps/browser/src/dirt/phishing-detection/services/phishing-indexeddb.service.spec.ts
```diff
@@ -0,0 +1,358 @@
...
+    const mockIndexedDB = {
+      open: jest.fn().mockImplementation(() => {
+        setTimeout(() => {
+          mockOpenRequest.onsuccess?.();
+        }, 0);
+        return mockOpenRequest;
+      }),
+    };
+
+    global.indexedDB = mockIndexedDB as any;
```
Banrion, line 135: ⚠️ Should cleanup this global variable in afterEach via "delete global.indexedDB"

apps/browser/src/dirt/phishing-detection/services/phishing-indexeddb.service.ts
```diff
@@ -0,0 +1,229 @@
...
+    });
+  }
+
+  /**
+   * Checks if a URL exists in the phishing database.
+   *
+   * @param url - The URL to check
+   * @returns `true` if URL exists, `false` if not found or on error
+   */
+  async hasUrl(url: string): Promise<boolean> {
```
Banrion, line 102: ❓  Should we use this in the save since we don't want duplicates? I think since we should index on the url, this would be best. I don't see it used. 

⚠️ We probably also want to pass in a custom matcher function

apps/browser/src/dirt/phishing-detection/services/phishing-indexeddb.service.ts
```diff
@@ -0,0 +1,229 @@
...
+        const { done, value } = await reader.read();
+        if (done) {
+          if (buffer.trim()) {
+            urls.push(buffer.trim());
+          }
+          if (urls.length) {
+            await this.saveChunk(db, urls);
+          }
+          break;
+        }
```
Banrion, line 219: ⚠️ Shouldn't this be done after the logic to update in the case there is data in this last chunk?

apps/browser/src/dirt/phishing-detection/services/phishing-indexeddb.service.ts
```diff
@@ -0,0 +1,229 @@
...
+        buffer += decoder.decode(value, { stream: true });
```
Banrion: This will never flag that there is not data coming in from the stream

```suggestion
        buffer += decoder.decode(value, { stream: !done });
```

apps/browser/src/dirt/phishing-detection/services/phishing-indexeddb.service.ts
```diff
@@ -0,0 +1,229 @@
...
+
+  /**
+   * Saves a single chunk of URLs in one transaction.
+   */
+  private saveChunk(db: IDBDatabase, urls: string[]): Promise<void> {
+    return new Promise((resolve, reject) => {
+      const tx = db.transaction(this.STORE_NAME, "readwrite");
+      const store = tx.objectStore(this.STORE_NAME);
+      for (const url of urls) {
+        if (url.trim()) {
```
Banrion: ⚠️ We can remove the redundant trims to keep the save chunk streamlined for saves only

apps/browser/src/dirt/phishing-detection/services/phishing-indexeddb.service.ts
```diff
@@ -0,0 +1,229 @@
...
+        buffer = lines.pop() ?? "";
+        for (const line of lines) {
+          if (line.trim()) {
+            urls.push(line.trim());
+          }
+        }
+        if (urls.length >= this.CHUNK_SIZE) {
+          await this.saveChunk(db, urls.splice(0, this.CHUNK_SIZE));
+          await new Promise((r) => setTimeout(r, 0));
+        }
```
Banrion, line 235:
```suggestion
      for (const line of lines) {
        const trimmed = line.trim();
        if (trimmed) urls.push(trimmed);

        if (urls.length >= this.CHUNK_SIZE) {
          await this.saveChunk(db, urls);
          urls = []; // Clear the array for the next chunk instead of using more expensive splice operation
        }
      }
```

apps/browser/src/dirt/phishing-detection/services/phishing-indexeddb.service.ts
```diff
@@ -0,0 +1,229 @@
...
+    const urls: string[] = [];
+
+    try {
+      while (true) {
+        const { done, value } = await reader.read();
+        if (done) {
+          if (buffer.trim()) {
+            urls.push(buffer.trim());
+          }
+          if (urls.length) {
```
Banrion:
```suggestion
          if (urls.length > 0) {
```

apps/browser/src/dirt/phishing-detection/services/phishing-indexeddb.service.spec.ts
```diff
@@ -0,0 +1,358 @@
...
+      const result = await service.saveUrlsFromStream(stream);
+
+      expect(result).toBe(true);
+      expect(mockObjectStore.put).toHaveBeenCalledTimes(2);
+    });
+
+    it("returns false on error", async () => {
+      const error = new Error("IndexedDB error");
+      mockOpenRequest.error = error;
+      (global.indexedDB.open as jest.Mock).mockImplementation(() => {
```
Banrion, line 334: ⛏️ If the global variable is needed or used in most test cases, you can add a spy and default functionality in the before each for default mocking and overwrite in the individual test cases.

```
    openSpy = jest.spyOn(global.indexedDB, 'open').mockImplementation(() => ({
      onsuccess: null, onerror: null 
    }));
```

apps/browser/src/dirt/phishing-detection/services/phishing-indexeddb.service.ts
AlexRubik: store.put({ url: "https://example.com" }) does an upsert and the url is our unique key (keyPath: "url") so the db should never have a duplicate. When we save with store.put() , if the url exists in db then it is overwritten, not appended.

apps/browser/src/dirt/phishing-detection/services/phishing-indexeddb.service.ts
AlexRubik: I think this logic captures the last chunk. 

if done, process/save remaining urls
break;
finally {
      reader.releaseLock();
    }

apps/browser/src/dirt/phishing-detection/services/phishing-indexeddb.service.spec.ts
AlexRubik: [c0c82fc](https://github.com/bitwarden/clients/commit/c0c82fc8b0d27dfb61e842bdac87ddb06f767b5d)

apps/browser/src/dirt/phishing-detection/services/phishing-indexeddb.service.spec.ts
AlexRubik: Added `delete (global as any).indexedDB` to `afterEach` 
[c0c82fc](https://github.com/bitwarden/clients/commit/c0c82fc8b0d27dfb61e842bdac87ddb06f767b5d)

apps/browser/src/dirt/phishing-detection/services/phishing-indexeddb.service.ts
AlexRubik: I've added a test case demonstrating this behavior explicitly.
[2df7e9a](https://github.com/bitwarden/clients/commit/2df7e9ad51b5436557c587e9443f1b6172fdb506)

apps/browser/src/dirt/phishing-detection/services/phishing-indexeddb.service.ts
AlexRubik: [641e9ed](https://github.com/bitwarden/clients/commit/641e9eda0032497c285fc314b4fc6e47bfc3f775)

apps/browser/src/dirt/phishing-detection/services/phishing-indexeddb.service.ts
Banrion: Don't we need to get all individual lines in the buffer via:

```
        const lines = buffer.split("\n");
        buffer = lines.pop() ?? "";
```
Otherwise, it might just be getting the rest of the buffer and trimming it but not separating the lines?

apps/browser/src/dirt/phishing-detection/services/phishing-indexeddb.service.ts
AlexRubik: [641e9ed](https://github.com/bitwarden/clients/commit/641e9eda0032497c285fc314b4fc6e47bfc3f775)

Removed redundant trims from saveChunk() and centralized cleaning:

saveChunked(): Now cleans input with .map(trim).filter(Boolean) for the saveUrls() path
processStream(): Uses single trim via local variable (part of the restructure)
saveChunk(): Now trusts callers provide clean data - focused solely on IndexedDB writes

apps/browser/src/dirt/phishing-detection/services/phishing-indexeddb.service.ts
AlexRubik: [641e9ed](https://github.com/bitwarden/clients/commit/641e9eda0032497c285fc314b4fc6e47bfc3f775)

apps/browser/src/dirt/phishing-detection/services/phishing-indexeddb.service.ts
AlexRubik: [641e9ed](https://github.com/bitwarden/clients/commit/641e9eda0032497c285fc314b4fc6e47bfc3f775)

apps/browser/src/dirt/phishing-detection/services/phishing-indexeddb.service.spec.ts
AlexRubik: Can address in the follow up fake-indexeddb iteration

apps/browser/src/dirt/phishing-detection/services/phishing-indexeddb.service.ts
AlexRubik: gotcha https://github.com/bitwarden/clients/pull/18344/commits/c594caa96dd435514aa20216a3a22bca07685ce0

Banrion (approved):

PR #18375, opened 2026-01-15
[PM-18855] Add edit Cipher permission check to Cipher Authorization Service and use in Vault dialog

lastbestdev:
## 🎟️ Tracking
https://bitwarden.atlassian.net/browse/PM-18855

## 📔 Objective

Centralize edit permission checks in `CipherAuthorizationService` instead of using the `disableForm` parameter passed to `VaultItemDialogComponent`. This refactoring improves consistency with how delete and restore permissions are handled, establishes a single source of truth for authorization logic, and simplifies caller components.

This change also fixes the bug in [ticket](https://bitwarden.atlassian.net/browse/PM-18855), which allows Users to properly edit Ciphers inside of the various Admin Console report types.

### Changes Made

**Core Authorization Service (`cipher-authorization.service.ts`):**
- Implemented `canEditCipher$()` method in `CipherAuthorizationService` following the same pattern as `canDeleteCipher$()` and `canRestoreCipher$()`
- Authorization logic checks:
  - For admin console actions with unassigned ciphers: `organization.canEditUnassignedCiphers`
  - For admin console actions with assigned ciphers: `organization.canEditAllCiphers`
  - For regular vault actions: `cipher.edit` property

**Vault Item Dialog Component (`vault-item-dialog.component.ts`):**
- Added `canEdit` initialization in `ngOnInit()` using `cipherAuthorizationService.canEditCipher$()`
- Updated `disableEdit` getter to use `!this.canEdit` instead of `this.params.disableForm`
- Added mode-forcing logic: if user cannot edit and dialog opens in "form" mode, automatically switches to "view" mode
- Removed `disableForm` property from `VaultItemDialogParams` interface

**Caller Components:**
- Removed `disableForm` calculation and parameter from `cipher-report.component.ts`
- Removed `disableForm` calculation, mode-forcing logic, and parameter from `vault.component.ts` (admin console)

## 📸 Screenshots
Editing Cipher in Admin Console report
https://github.com/user-attachments/assets/e59ccad4-f530-47fb-89e3-a4e1fb3f8543

### DIFF
--- a/.claude/settings.json
+++ b/.claude/settings.json
@@ -1,0 +1,10 @@
+{
+  "extraKnownMarketplaces": {
+    "bitwarden-marketplace": {
+      "source": {
+        "source": "github",
+        "repo": "bitwarden/ai-plugins"
+      }
+    }
+  }
+}
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -4,6 +4,10 @@
 #
 # https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners
 
+## Global styles are owned by UIF
+*.scss @bitwarden/team-ui-foundation
+*.css @bitwarden/team-ui-foundation
+
 ## Desktop native module ##
 apps/desktop/desktop_native @bitwarden/team-platform-dev
 apps/desktop/desktop_native/objc/src/native/autofill @bitwarden/team-autofill-desktop-dev
--- a/apps/browser/src/_locales/ar/messages.json
+++ b/apps/browser/src/_locales/ar/messages.json
@@ -3035,10 +3035,6 @@
   "custom": {
     "message": "مُخصّص"
   },
-  "sendPasswordDescV3": {
-    "message": "Add an optional password for recipients to access this Send.",
-    "description": "'Send' is a noun and the name of a feature called 'Bitwarden Send'. It should not be translated."
-  },
   "createSend": {
     "message": "New Send",
     "description": "'Send' is a noun and the name of a feature called 'Bitwarden Send'. It should not be translated."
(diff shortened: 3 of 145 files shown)


apps/web/src/app/dirt/reports/pages/organizations/inactive-two-factor-report.component.ts
```diff
@@ -99,11 +99,8 @@ export class InactiveTwoFactorReportComponent
       });
   }
 
-  async getAllCiphers(): Promise<CipherView[]> {
-    if (this.organization) {
-      return await this.cipherService.getAllFromApiForOrganization(this.organization.id, true);
-    }
-    return [];
+  getAllCiphers(): Promise<CipherView[]> {
+    return this.cipherService.getAllFromApiForOrganization(this.organization.id, true);
```
nikwithak: Removing the falsy check here is breaking the typechecking, could you add it back in?

JaredScar: Are we not worried about `this.organization` potentially being null/undefined here? Accessing `.id` on something null/undefined would be troublesome

JaredScar (changes requested):

nikwithak (changes requested):
Looks good oaverall - one typechecking issue to fix.

apps/web/src/app/dirt/reports/pages/organizations/inactive-two-factor-report.component.ts
lastbestdev: My VS Code wasn't properly using the project's typescript config, so I didn't see these errors. Got that resolved and fixed in latest!

lastbestdev: My VS Code wasn't properly using the project's typescript config, so I didn't see these errors. Got that resolved and fixed in latest!

JaredScar (dismissed):

nikwithak (dismissed):

lastbestdev:
@JaredScar @nikwithak @Banrion had to request reapproval due to a merge conflict resolution. my apologies!

nikwithak (approved):

PR #18379, opened 2026-01-15
Improved claude.md

abergs:
## 📔 Objective

**Less is more**
This PR adds some key concepts, but also tries to not be overly prescriptive. Instead I recommend we start small and we normalize adding instructions as we find what claude gets wrong abour our codebase and patterns.
 
Why avoid throwing in the kitchen sink?
I think Claude is already doing reasonably well in our code base. Every session gets loaded with this context, regardless of their task (Developing libs/UI, analyzing a bug, writing tests, reviewing etc) and I think polluting the context further will give us more baggage to carry around, and less context available in every session.

There are a number of stuff we might want to yank out of this PR still:
* Examples of Unit Testing? Could be lifted out or places in it's own file like we did with angular
* Critical rules vs Anti-Pattern

**Things I'd like for us to add:**
General debugging advice for humans.
Would be great if we could include the golden nuggets on how to run things so that you get more verbose logging, or where to find log files.

### DIFF
--- a/.claude/CLAUDE.md
+++ b/.claude/CLAUDE.md
@@ -1,4 +1,4 @@
-# CLAUDE.md
+# Clients Monorepo
 
 ## Overview
 


.claude/CLAUDE.md
```diff
@@ -1,111 +1,391 @@
...
+
+Feature flags are defined in `libs/common/src/enums/feature-flag.enum.ts`. Use `ConfigService` to check flags:
+
+```typescript
+const enabled = await this.configService.getFeatureFlag(FeatureFlag.MyFlag);
+```
+
+Flags MUST be short-lived and removed once fully enabled.
+
+## Testing
```
abergs, line 179: Testing context is currently 70 lines and could probably be less. We could put it in a different file or we could just trim the fat.

I left it in because I think many sessions need to touch tests, but may be overkill.

.claude/CLAUDE.md
```diff
@@ -1,111 +1,340 @@
...
+- **NEVER** add new encryption logic to this repo—cryptographic operations belong in the SDK
+- **NEVER** send unencrypted vault data to API services
 - **NEVER** log decrypted data, encryption keys, or PII
-  - No vault data in error messages or console logs
+- **CRITICAL**: Tailwind CSS classes MUST use the `tw-` prefix (e.g., `tw-flex`, `tw-p-4`)
+- **NEVER** use TypeScript enums—use const objects with type aliases (see pattern below)
+- **NEVER** use code regions—refactor for readability instead
 
-- **ALWAYS** Respect configuration files at the root and within each app/library (e.g., `eslint.config.mjs`, `jest.config.js`, `tsconfig.json`).
+## Monorepo Architecture
```
abergs, line 92: We should likely add something about codeowners? And *maybe* about eslint/tsconfig/jes... but happy to wait and see.

theMickster: Just curious, how would referencing `codeowners` help Claude?  🤔

I would _expect_ that Claude is smart enough to determine all of those additional things you pointed out on it's own from the context in this file and the fact that nearly all production Angular apps have those items defined. Just not sure what telling Claude things it should already know help it to write better code and perform better code reviews 🤔

.claude/CLAUDE.md
```diff
@@ -1,111 +1,340 @@
...
+| Tailwind styles not applying   | Ensure `tw-` prefix on all Tailwind classes |
 
 ## References
 
-- [Web Clients Architecture](https://contributing.bitwarden.com/architecture/clients)
-- [Architectural Decision Records (ADRs)](https://contributing.bitwarden.com/architecture/adr/)
+### Official Documentation
+
+- [Clients Architecture](https://contributing.bitwarden.com/architecture/clients)
+- [ADRs](https://contributing.bitwarden.com/architecture/adr/)
```
theMickster: 🤔 I think we should leave the link to the ADRs as it was before to ensure Claude doesn't guess at acronyms.

.claude/CLAUDE.md
```diff
@@ -1,111 +1,340 @@
...
 - [Contributing Guide](https://contributing.bitwarden.com/)
-- [Web Clients Setup Guide](https://contributing.bitwarden.com/getting-started/clients/)
 - [Code Style](https://contributing.bitwarden.com/contributing/code-style/)
-- [Security Whitepaper](https://bitwarden.com/help/bitwarden-security-white-paper/)
```
theMickster, line 110: ❓ Did we feel that this reference was not helpful? Or did it move?

.claude/CLAUDE.md
abergs: @theMickster In what way is the link different now? Or do you mean include (ADR-XXX) throughout the doc?

.claude/CLAUDE.md
```diff
@@ -1,111 +1,340 @@
...
+| Type                | Convention                 | Example             |
+| ------------------- | -------------------------- | ------------------- |
+| Files               | kebab-case                 | `cipher.service.ts` |
+| Classes             | PascalCase                 | `CipherService`     |
+| Interfaces          | PascalCase (no `I` prefix) | `CipherService`     |
+| Variables/Functions | camelCase                  | `getCipher()`       |
+| Constants           | SCREAMING_SNAKE_CASE       | `MAX_RETRY_COUNT`   |
+| Observables         | camelCase with `$` suffix  | `ciphers$`          |
+
+## Anti-Patterns
```
theMickster, line 286: ⛏️ Does anyone else feel that an anti-pattern with a do subheading feels like an oxymoron? Or am I the moron? 🤣 
Didn't catch this in the ai-plugin.

.claude/CLAUDE.md
abergs: codeowners is from my lived experience with claude: When it makes changes it doesn't necessarily check if the file is covered by ownership which has been pointed out during review.

I don't have a strong case for referring to the configs. I'm happy to leave them out for now.

.claude/CLAUDE.md
theMickster, line 1: ❓How do we feel that all mentions to specific ADRs of our best practices were removed?

.claude/CLAUDE.md
abergs: Yes. But I thought that was what we had put into the plugin so I just kept it.

.claude/CLAUDE.md
abergs: I think we should bring them back, they are good if only for human reference

.claude/CLAUDE.md
theMickster: Interesting 🧐 
I would like to see some of the instances you reference; having trouble remembering that specifically being mentioned.

.claude/CLAUDE.md
abergs: I don't think it adds value - it's a long document, with quite a lot of diagrams (img form).

Let's add it back if we encounter that claude messes up?

.claude/CLAUDE.md
```diff
@@ -1,111 +1,340 @@
...
 ```
 
-**Angular Signals (ADR-0027):**
+## Testing
 
-Encourage the use of Signals **only** in Angular components and presentational services.
+### Test Structure
 
-Use **RxJS** for:
+Tests are colocated with source files using `.spec.ts` suffix:
```
theMickster: 💭 These three lines seem obvious to me and I would expect it to be obvious to Claude too.

.claude/CLAUDE.md
```diff
@@ -1,111 +1,340 @@
...
-- Services used across Angular and non-Angular clients
-- Complex reactive workflows
-- Interop with existing Observable-based code
+```
+libs/common/src/vault/models/domain/cipher.ts
+libs/common/src/vault/models/domain/cipher.spec.ts
+```
 
-**NO TypeScript Enums (ADR-0025):**
+### Test Utilities
```
theMickster, line 183: 💯 to the listing of our test utilities very valuable because it's our patterns and practices. The exact things that Claude needs to craft code _our_ way. 

🤔 Should we consider crafting something into the base `enhance.md`?

.claude/CLAUDE.md
abergs: When in doubt, let's leave it out and add later should it be necessary.

Last instance (also package ownership declared in renovate config bits you all the time): https://github.com/bitwarden/clients/pull/18039#pullrequestreview-3592243721

But as I said, we can add that later.

.claude/CLAUDE.md
abergs: I erroneously moved into angular.md, moving it back here.

.claude/CLAUDE.md
abergs: Agreed

.claude/CLAUDE.md
theMickster: I can be cool with that

.claude/CLAUDE.md
theMickster: 💭 Sorry, clarifying with a suggestion to put the link back the way it was before.
```suggestion
- [Architectural Decision Records (ADRs)](https://contributing.bitwarden.com/architecture/adr/)
```

.claude/CLAUDE.md
```diff
@@ -1,111 +1,321 @@
...
 enum CipherType {
   Login = 1,
-  SecureNote = 2,
 }
 ```
 
-Example: `/libs/common/src/vault/enums/cipher-type.ts`
+## Anti-Patterns
+
+- **NEVER** add new encryption logic to this repo—use the SDK
```
quexten, line 288: We write never here, but there are some exceptions such as:
- Biometric unlock subsystem with encrypted memory store
- SSH agent
These live in desktop-native.

Further permitted use cases are:
- Lastpass importer
- Chromium importer

Though, i'm not sure if it would just confuse the model if we added these. Maybe it's better to ad-hoc ignore claude reviews there.

abergs: Well spotted. Yeah we could add that specialized context in nested documents if/when it fails on it.

.claude/CLAUDE.md
```diff
@@ -1,111 +1,321 @@
...
+...
+```
 
-**Strict boundaries** must be maintained between apps and libraries. Do not introduce
-cross-dependencies that violate the intended modular structure. Always consult and respect the
-dependency rules defined in `eslint.config.mjs`, `nx.json`, and other configuration files.
+## Angular Patterns
 
-## Angular Architecture Patterns
+When working with Angular related files (files that import anything from `@angular/*`), you MUST READ AND FOLLOW THIS FILE: `./angular.md`.
```
withinfocus: 🎨 I think this would roll up to our shared skills collection in the marketplace.

abergs: Does everyone have that installed? If so I can wipe this but otherwise we'll keep it in until next iteration.

withinfocus: That's expected, yes. You _should_ be able to provide explicit instructions using the right syntax to reference the installed marketplace.

abergs: Rmoved angular patterns

.claude/CLAUDE.md
SaintPatrck: I agree that most sessions will touch tests, and 70-ish lines doesn't seem like too much. I'm good with this as-is.

I'm sure we'll eventually need more fine-tuning if we have very strict testing patterns and guidelines that need to be followed in this repo like we do in Android. I'm using Claude to backfill tests for our Android Authenticator app, and all of the little nuances are building up quickly enough that I needed to extrapolate them from `CLAUDE.md` into a Skill. Now I'm considering [Rules](https://code.claude.com/docs/en/memory#modular-rules-with-claude/rules/) instead because the Skill is essentially becoming a list of rules that need to be followed when writing tests.

abergs: @SaintPatrck Sounds like we'll keep this in for now but will experiment with rules or skills as we grow,

.claude/CLAUDE.md
```diff
@@ -1,111 +1,317 @@
-# Bitwarden Clients - Claude Code Configuration
+# CLAUDE.md
```
withinfocus:
```suggestion
# Clients Monorepo
```

.claude/CLAUDE.md
```diff
@@ -1,111 +1,317 @@
...
+| Type                | Convention                 | Example             |
+| ------------------- | -------------------------- | ------------------- |
+| Files               | kebab-case                 | `cipher.service.ts` |
+| Classes             | PascalCase                 | `CipherService`     |
+| Interfaces          | PascalCase (no `I` prefix) | `CipherService`     |
+| Variables/Functions | camelCase                  | `getCipher()`       |
+| Constants           | SCREAMING_SNAKE_CASE       | `MAX_RETRY_COUNT`   |
+| Observables         | camelCase with `$` suffix  | `ciphers$`          |
+
+### No TypeScript Enums (ADR-0025):
```
withinfocus, line 270: 💭 A lot of our ADRs relate to this repo but ... only this one is kept?

PR #18387, opened 2026-01-15
[PM-30500] Centralize Organization Data Ownership

BTreston:
## 🎟️ Tracking
https://bitwarden.atlassian.net/browse/PM-30500

## 📔 Objective
Updates the existing copy and layout for the organization data ownership policy for my items. Implements multi step dialog workflow for enabling the policy. This also promotes the existing vNext component to be the default, and implements the new vNext component behind the new feature flag.

## 📸 Screenshots
DESIGN REVIEW:

https://github.com/user-attachments/assets/4d0a250a-e696-4854-8261-20083c33b44f

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/apps/web/src/app/admin-console/organizations/policies/policy-edit-definitions/vnext-organization-data-ownership.component.html
+++ b/apps/web/src/app/admin-console/organizations/policies/policy-edit-definitions/vnext-organization-data-ownership.component.html
@@ -15,19 +15,17 @@
 
   <div class="tw-text-left tw-overflow-hidden tw-mb-2">
     <strong>{{ "benefits" | i18n }}:</strong>
-    <div class="tw-flex tw-flex-col tw-p-2">
-      <ul class="tw-list-disc tw-pl-5 tw-space-y-2 tw-break-words tw-mb-0">
-        <li>
-          {{ "centralizeDataOwnershipBenefit1" | i18n }}
-        </li>
-        <li>
-          {{ "centralizeDataOwnershipBenefit2" | i18n }}
-        </li>
-        <li>
-          {{ "centralizeDataOwnershipBenefit3" | i18n }}
-        </li>
-      </ul>
-    </div>
+    <ul class="tw-pl-7 tw-space-y-2 tw-pt-2">
+      <li>
+        {{ "centralizeDataOwnershipBenefit1" | i18n }}
+      </li>
+      <li>
+        {{ "centralizeDataOwnershipBenefit2" | i18n }}
+      </li>
+      <li>
+        {{ "centralizeDataOwnershipBenefit3" | i18n }}
+      </li>
+    </ul>
   </div>
 
   <bit-form-control>


BTreston:
@eliykat This change reuses the multi-step dialog pattern used in auto confirm, however I think its worth discussing now if we want to continue doing it this way or if we think _now_ is an appropriate time to implement some layer of abstraction to capture this kind of behavior. 

It seems to me that the obvious direction product want to take regarding the policy experience is that we will implement increasingly complex features that will include more multi-step/custom dialogs that integrate with other parts of the app as seen here and in auto-confirm (this will be the 3rd ask of this type in < a year).

eliykat:
@BTreston I agree. I recall you implementing a more general purpose solution for this initially, do you want to go back to that? Your call when you do it, depending on timelines and your capacity.

apps/web/src/app/admin-console/organizations/policies/policy-edit-dialog.component.ts
```diff
@@ -73,13 +74,24 @@ export class PolicyEditDialogComponent implements AfterViewInit {
...
     return this.data.policy;
   }
 
+  /**
+   * Type guard to check if the policy component has the buildVNextRequest method.
+   */
+  private hasVNextRequest(
+    component: BasePolicyEditComponent,
+  ): component is BasePolicyEditComponent & {
+    buildVNextRequest: (orgKey: OrgKey) => Promise<OrganizationDataOwnershipPolicyRequest>;
```
JimmyVo16: Shouldn’t this be a generic vNext model instead of specifying OrganizationDataOwnershipPolicyRequest, so this path can handle other vNext models as well?

BTreston:
> @BTreston I agree. I recall you implementing a more general purpose solution for this initially, do you want to go back to that? Your call when you do it, depending on timelines and your capacity.

@eliykat Sounds good, I don't think I will do it in this PR. I'll create a tech debt item for it since I would want to take the opportunity to simplify the overall design. Specifically cleaning up the brittleness between what is the "policy" vs "the dialog." I'll loop you in on the proposed design before jumping in.

apps/web/src/app/admin-console/organizations/policies/policy-edit-definitions/organization-data-ownership.component.html
```diff
@@ -1,8 +1,57 @@
...
 </bit-form-control>
+
+<ng-template #dialog>
+  <bit-simple-dialog background="alt">
+    <span bitDialogTitle>{{ "organizationDataOwnershipWarningTitle" | i18n }}</span>
+    <ng-container bitDialogContent>
+      <div class="tw-text-left tw-overflow-hidden">
+        {{ "organizationDataOwnershipWarningContentTop" | i18n }}
+        <div class="tw-flex tw-flex-col tw-p-2">
+          <ul class="tw-list-disc tw-pl-5 tw-space-y-2 tw-break-words tw-mb-0">
```
eliykat, line 24: 💭 (non-blocking): seems like a lot of styling just to get a list. Maybe UIF can supply list stylings along with their typography?

apps/web/src/app/admin-console/organizations/policies/policy-edit-definitions/organization-data-ownership.component.ts
```diff
@@ -1,22 +1,38 @@
...
 import { BasePolicyEditDefinition, BasePolicyEditComponent } from "../base-policy-edit.component";
 
+export interface OrganizationDataOwnershipPolicyRequest {
+  policy: PolicyRequest;
+  metadata: {
+    defaultUserCollectionName: string;
+  };
+}
+
 export class OrganizationDataOwnershipPolicy extends BasePolicyEditDefinition {
```
eliykat, line 25: It seems like we're renaming the policy from `Organization data ownership` to `Centralize organization ownership`. We should update all the class/enum/i18n key names to be consistent - it's done in some places here but not across the board. This can be done in a follow-up PR if needed.

apps/web/src/app/admin-console/organizations/policies/policy-edit-dialog.component.ts
BTreston: I've fixed this. We had a model for this, I had erroneously renamed it. Though the model should not live in the component file, but that can be cleaned up later when we promote this from vNext.

apps/web/src/app/admin-console/organizations/policies/policy-edit-definitions/vnext-organization-data-ownership.component.ts
```diff
@@ -1,30 +1,48 @@
...
 import { BasePolicyEditDefinition, BasePolicyEditComponent } from "../base-policy-edit.component";
+import { OrganizationDataOwnershipPolicyDialogComponent } from "../policy-edit-dialogs";
 
-interface VNextPolicyRequest {
+export interface VNextPolicyRequest {
   policy: PolicyRequest;
   metadata: {
     defaultUserCollectionName: string;
   };
 }
```
eliykat, line 32: Duplicated definition (maybe intentional to isolate vCurrent and vNext implementations?)

apps/web/src/app/admin-console/organizations/policies/policy-edit-definitions/vnext-organization-data-ownership.component.ts
```diff
@@ -38,27 +56,16 @@ export class vNextOrganizationDataOwnershipPolicyComponent
...
-  override async confirm(): Promise<boolean> {
-    if (this.policyResponse?.enabled && !this.enabled.value) {
-      const dialogRef = this.dialogService.open(this.warningContent, {
-        positionStrategy: new CenterPositionStrategy(),
-      });
-      const result = await lastValueFrom(dialogRef.closed);
-      return Boolean(result);
-    }
-    return true;
-  }
```
eliykat, line 61: I can see how the step approach makes this much tidier. Nice work.

apps/web/src/app/admin-console/organizations/policies/policy-edit-dialogs/organization-data-ownership-edit-policy-dialog.component.ts
```diff
@@ -0,0 +1,213 @@
...
+
+  protected centralizeDataOwnershipEnabled$: Observable<boolean> =
+    this.accountService.activeAccount$.pipe(
+      getUserId,
+      switchMap((userId) => this.policyService.policies$(userId)),
+      map(
+        (policies) =>
+          policies.find((p) => p.type === PolicyType.OrganizationDataOwnership)?.enabled ?? false,
+      ),
+    );
```
eliykat: This should use `policyApiService.getPolicy` like the regular component does. It'll give the same result, but I want to keep using direct API calls rather than integrating more sync data into Admin Console.

eliykat (changes requested):
Looks good to me.

After looking at it more closely I definitely agree we need to refactor the dialog components to be more reusable. The edit policy page has been easily extensible, now we are risking fragmenting different implementations for specific policies. It would be best to re-align them while supporting the more complex behavior required by product, as well as dropping the vNext endpoint.

That is all non-blocking for now though - probably the only blocking feedback below is the use of the policy endpoint.

@JimmyVo16 can approve on my behalf so you're not waiting for me to come back online, if you like.

JimmyVo16 (dismissed):

apps/web/src/app/admin-console/organizations/policies/policy-edit-definitions/vnext-organization-data-ownership.component.ts
BTreston: Correct, intentional

apps/web/src/app/admin-console/organizations/policies/policy-edit-definitions/organization-data-ownership.component.ts
BTreston: This will be a followup (would be very wide reaching PR)

apps/web/src/app/admin-console/organizations/policies/policy-edit-definitions/organization-data-ownership.component.html
BTreston: This is simplified in the vNext component. This was just a messy implementation after iterating on products design.

JimmyVo16 (approved):

eliykat (approved):

PR #18388, opened 2026-01-15
[CL-637] icon api buttons links

BryanCunningham:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/CL-637

## 📔 Objective

- create link component
- update link and button API to accept icon inputs for more consistent UI control

### DIFF
--- a/libs/components/src/button/button.component.html
+++ b/libs/components/src/button/button.component.html
@@ -1,25 +1,13 @@
-<ng-template #buttonContent>
-  <ng-content></ng-content>
-</ng-template>
-
 <span class="tw-relative tw-flex tw-items-center tw-justify-center">
-  <span [class.tw-invisible]="showLoadingStyle()">
-    @if (startIcon() || endIcon()) {
-      <div class="tw-flex tw-items-center tw-gap-2">
-        @if (startIcon()) {
-          <i class="{{ startIconClasses() }}"></i>
-        }
-        <div>
-          <ng-container *ngTemplateOutlet="buttonContent"></ng-container>
-        </div>
-        @if (endIcon()) {
-          <i class="{{ endIconClasses() }}"></i>
-        }
-      </div>
-    } @else {
-      <div>
-        <ng-container *ngTemplateOutlet="buttonContent"></ng-container>
-      </div>
+  <span [class.tw-invisible]="showLoadingStyle()" class="tw-flex tw-items-center tw-gap-2">
+    @if (startIcon()) {
+      <i class="{{ startIconClasses() }}"></i>
+    }
+    <div>
+      <ng-content></ng-content>
+    </div>
+    @if (endIcon()) {
+      <i class="{{ endIconClasses() }}"></i>
     }
   </span>
   @if (showLoadingStyle()) {


libs/components/src/link/link.component.html
```diff
@@ -0,0 +1,12 @@
+@let leadIcon = startIcon();
+@let tailIcon = endIcon();
```
willmartian: The `@let` statements aren't adding anything, and require you to rename the inputs, which is a bit confusing. I would only use `@let` when creating _new_ information in a template.

libs/components/src/link/link.component.ts
```diff
@@ -68,47 +80,53 @@ const commonStyles = [
...
 
-@Directive({
-  selector: "button[bitLink]",
-  hostDirectives: [AriaDisableDirective],
-})
-export class ButtonLinkDirective extends LinkDirective {
-  private el = inject(ElementRef<HTMLButtonElement>);
+  readonly classList = computed(() => {
+    const verticalInset = this.isButton() ? "0.25rem" : "0.125rem";
+    return getLinkClasses({ linkType: this.linkType(), verticalInset });
```
willmartian: This helper method obfuscates what is going on, IMO. I would recommend removing it and inlining the logic.

libs/components/src/link/link.component.ts
```diff
@@ -68,47 +80,53 @@ const commonStyles = [
...
 
-/**
-  * Text Links and Buttons can use either the `<a>` or `<button>` tags. Choose which based on the action the button takes:
+@Component({
+  selector: "a[bitLink], button[bitLink]",
+  templateUrl: "./link.component.html",
+  changeDetection: ChangeDetectionStrategy.OnPush,
+  host: {
+    "[class]": "classList()",
+    "[attr.bit-aria-disable]": "isButton() ? true : null",
```
willmartian: Can you remind we what this is?

libs/components/src/link/link.component.ts
```diff
@@ -68,47 +80,53 @@ const commonStyles = [
...
-  selector: "a[bitLink]",
-})
-export class AnchorLinkDirective extends LinkDirective {
-  @HostBinding("class") get classList() {
-    return ["before:-tw-inset-y-[0.125rem]"]
-      .concat(commonStyles)
-      .concat(linkStyles[this.linkType()] ?? []);
-  }
-}
+  protected readonly isButton = computed(() => this.el.nativeElement.tagName === "BUTTON");
```
willmartian: This does not need to be a computed signal--it is not reading another signal.

libs/components/src/link/link.component.ts
```diff
@@ -68,47 +80,53 @@ const commonStyles = [
...
+  });
 
-  @HostBinding("class") get classList() {
-    return ["before:-tw-inset-y-[0.25rem]"]
-      .concat(commonStyles)
-      .concat(linkStyles[this.linkType()] ?? []);
-  }
+  readonly endIconClasses = computed(() => {
+    return ["bwi", "!tw-no-underline", this.endIcon()];
+  });
```
willmartian: This is nitpicky, but I prefer keeping as many classes in the template as possible:

```
<div class=["bwi", "!tw-no-underline", startIcon()]>
```

It means less bloat in the TS file, and the classes are defined closer to their intended element.

libs/components/src/link/link.component.ts
```diff
@@ -68,47 +80,53 @@ const commonStyles = [
...
+})
+export class LinkComponent {
+  private el = inject(ElementRef<HTMLElement>);
 
-  * - if navigating to a new page, use a `<a>`
-  * - if taking an action on the current page, use a `<button>`
+  readonly linkType = input<LinkType>("primary");
+  readonly startIcon = input<BitwardenIcon | undefined>(undefined);
+  readonly endIcon = input<BitwardenIcon | undefined>(undefined);
+  readonly disabled = input(false, { transform: booleanAttribute });
```
willmartian, line 132: Could you add JS doc comments to the inputs?

libs/components/src/link/link.component.ts
```diff
@@ -68,47 +80,53 @@ const commonStyles = [
...
+  },
+})
+export class LinkComponent {
+  private el = inject(ElementRef<HTMLElement>);
 
-  * - if navigating to a new page, use a `<a>`
-  * - if taking an action on the current page, use a `<button>`
+  readonly linkType = input<LinkType>("primary");
+  readonly startIcon = input<BitwardenIcon | undefined>(undefined);
+  readonly endIcon = input<BitwardenIcon | undefined>(undefined);
```
willmartian, line 126: Nice, typesafe `BitwardenIcon`.

libs/components/src/link/link.component.ts
BryanCunningham: @willmartian This is for us to be able to correctly `aria-disable` the button and capture clicks. It's normally added via the `AriaDisableDirective` as a host directive. But, I'm not able to conditionally apply the host directive based on if this is a button or not

libs/components/src/link/link.component.ts
BryanCunningham: Probably not. I'll remove

libs/components/src/link/link.component.ts
willmartian: Ah yes, makes sense. Could you add a brief comment saying that for next time I forget? :)

libs/components/src/link/link.component.ts
```diff
@@ -68,47 +83,54 @@ const commonStyles = [
...
+   * @default undefined
+   */
+  readonly startIcon = input<BitwardenIcon | undefined>(undefined);
+  /**
+   * The trailing icon to display within the link
+   * @default undefined
+   */
+  readonly endIcon = input<BitwardenIcon | undefined>(undefined);
+  /**
+   * Weather the button is disabled
```
vleague2:
```suggestion
   * Whether the button is disabled
```

libs/components/src/button/button.component.html
```diff
@@ -1,8 +1,32 @@
-<span class="tw-relative">
-  <span [ngClass]="{ 'tw-invisible': showLoadingStyle() }">
-    <ng-content></ng-content>
+@let showLoadingStyles = showLoadingStyle();
+@let leadIcon = startIcon();
+@let tailIcon = endIcon();
```
willmartian: These are not necessary, we can just bind directly to the original signals

libs/components/src/link/link.component.ts
```diff
@@ -134,7 +135,7 @@ export class LinkComponent {
 
   readonly classList = computed(() => {
     const verticalInset = this.isButton ? "0.25rem" : "0.125rem";
-    const displayStyle = this.isButton ? "inline-block" : "inline-flex";
+    const displayStyle = this.isButton ? "" : "inline-flex";
 
     return [`before:-tw-inset-y-[${verticalInset}]`, `tw-${displayStyle}`]
```
willmartian: 1. This will result in an invalid class `tw-`
2. Concatenating partial Tailwind class names isn't officially supported and can break (Tailwind only includes class names it can find in the repo at compile time). Could you refactor this to not do that?

libs/components/src/link/link.directive.ts
```diff
@@ -46,16 +54,16 @@ const commonStyles = [
...
+  "[&:hover_span]:tw-decoration-[.125em]",
+  "[&.tw-test-hover_span]:tw-decoration-[.125em]",
   "disabled:tw-no-underline",
   "disabled:tw-cursor-not-allowed",
   "disabled:!tw-text-fg-disabled",
   "disabled:hover:!tw-text-fg-disabled",
   "disabled:hover:tw-no-underline",
   "focus-visible:tw-outline-none",
-  "focus-visible:tw-underline",
-  "focus-visible:tw-decoration-1",
```
willmartian, line 58: Is this diff related to the goal of the PR?

libs/components/src/link/link.component.ts
```diff
@@ -83,47 +95,60 @@ const commonStyles = [
...
 })
-export class AnchorLinkDirective extends LinkDirective {
-  @HostBinding("class") get classList() {
-    return ["before:-tw-inset-y-[0.125rem]"]
-      .concat(commonStyles)
-      .concat(linkStyles[this.linkType()] ?? []);
-  }
-}
+export class LinkComponent {
+  readonly el = inject(ElementRef<HTMLElement>);
```
willmartian, line 111: This should be private

libs/components/src/link/link.component.html
```diff
@@ -0,0 +1,11 @@
+<div class="tw-flex tw-gap-2 tw-items-center">
```
willmartian, line 1: Is this extra div necessary, or could these classes also be on the host?

libs/components/src/link/link.component.ts
```diff
@@ -83,47 +95,60 @@ const commonStyles = [
...
+   */
+  readonly disabled = input(false, { transform: booleanAttribute });
 
-@Directive({
-  selector: "button[bitLink]",
-  hostDirectives: [AriaDisableDirective],
-})
-export class ButtonLinkDirective extends LinkDirective {
-  private el = inject(ElementRef<HTMLButtonElement>);
+  protected isButton = this.el.nativeElement.tagName === "BUTTON";
```
willmartian: This should be readonly

libs/components/src/link/link.directive.ts
BryanCunningham: Yeah, according to the design the underline should only appear in the hover state

libs/components/src/link/link.component.html
BryanCunningham: @willmartian this was just a guardrail so that I didn't override an custom styling that was being applied by the consumer.

willmartian (changes requested):
The focus state sometimes breaks. 
image

This doesn't happen on `main`: https://components.bitwarden.com/?path=/story/component-library-link--buttons

BryanCunningham:
> The focus state sometimes breaks. image
> 
> This doesn't happen on `main`: https://components.bitwarden.com/?path=/story/component-library-link--buttons

@willmartian I can't repro this. Mind giving this another try once the storybook build publishes?

libs/components/src/callout/callout.stories.ts
```diff
@@ -113,7 +113,7 @@ export const WithTextButton: Story = {
     template: `
       <bit-callout ${formatArgsForCodeSnippet<CalloutComponent>(args)}>
       <p class="tw-mb-2">The content of the callout</p>
-        <a bitLink> Visit the help center<i aria-hidden="true" class="bwi bwi-fw bwi-sm bwi-angle-right"></i> </a>
+        <a bitLink endIcon="bwi-angle-right">Visit the help center</a>
```
vleague2, line 116: So for a diff like this, I think we are going to get unexpected results. The original was using `bwi-fw` and `bwi-sm`. `bwi-fw` is fixed width, which may not affect much, but `bwi-sm` is for a smaller size. I guess -- do we need to build in sizing capabilities for start and end icons?

BryanCunningham: I would think we'd want these to always be the same size. And the CL would control that

vleague2: True. It doesn't seem like that many instances of this, so maybe a spot-check can be part of the design review

willmartian (approved):

libs/components/src/button/button.component.html
```diff
@@ -1,6 +1,26 @@
...
+        </div>
+        @if (endIcon()) {
+          <i class="{{ endIconClasses() }}"></i>
+        }
+      </div>
+    } @else {
+      <div>
+        <ng-container *ngTemplateOutlet="buttonContent"></ng-container>
+      </div>
+    }
```
willmartian: Do we need the `ng-template` or `@if (startIcon() || endIcon())`? I think something like this would work whether or not icons are passed:

```html
<span [class.tw-invisible]="showLoadingStyle()" class="tw-flex tw-items-center tw-gap-2">
        @if (startIcon()) {
          <i class="{{ startIconClasses() }}"></i>
        }
        <div>
          <ng-content></ng-content>
        </div>
        @if (endIcon()) {
          <i class="{{ endIconClasses() }}"></i>
        }
</span>
```

BryanCunningham: Good call. Fixed [here](https://github.com/bitwarden/clients/pull/18388/changes/236cd3349bf063121ce530dc03025efb9580e017)

willmartian (approved):

libs/components/src/button/button.component.html
willmartian: Thanks!

Issue #13977, opened 2025-03-25
Broken SSH Import from 1Password Export file


Amit-Shan:
### Steps To Reproduce

1. Import a 1Password export file into the Bitwarden Windows application.
2. Check any of the imported SSH items and compare with the item in the export file. 

### Expected Result

I expected the SSH items to be imported with all relevant fields. 

### Actual Result

None of the SSH items are imported as SSH items and the keys are not imported as well. Only the title of the item is imported and the entire item is saved as a secure note. 

### Operating System

Windows

### Operating System Version

Windows 11 23H2 Build 22635.5097

### Installation method

Direct Download (from bitwarden.com)

### Build Version

Version 2025.2.1 SDK 'main (6008e90)' Shell 34.0.0 Renderer 132.0.6834.83 Node 20.18.1 Architecture x64

### Issue Tracking Info

- [x] I understand that work is tracked outside of GitHub. A PR will be linked to this issue should one be opened to address it, but Bitwarden doesn't use fields like "assigned", "milestone", or "project" to track progress.

S-Kakar:
Thank you for reporting this issue! We've added this to our internal tracking system.
ID: PM-19483

Krychaz:
Hello there, 

Current this is not possible. However this is something our team is working on and we hope to bring in future updates.

Amit-Shan:
@Krychaz Oh. I thought it was just the import being buggy, but the import does not import ssh at all? I see.

philotas:
Shouldn't there be a warning, that there were fields not imported? Or in general there could be list of items which only have a title after import, so one could revise them. Usually all items should have a title and some other field at least.

And also even if they are imported as secret notes, I do no see a reason why the fields should not be imported as well.

@Krychaz no, it does import the item as a secure note, but without any fields.

PR #10787, opened 2024-08-29
[PM-10938] Import ssh-keys from 1password 1pux

djsmith85:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-10938

## 📔 Objective

With the introduction of the SSH keys item type, we can now also support importing ssh keys from other password managers. This add support for 1password's 1pux format.

The 1pux format has changed and the updates to fix it have been done at https://github.com/bitwarden/clients/pull/10778. This PR requires the aforementioned to be merged first

## ⏰ Reminders before review

- Contributor guidelines followed
- All formatters and local linters executed and passed
- Written new unit and / or integration tests where applicable
- Protected functional changes with optionality (feature flags)
- Used internationalization (i18n) for all UI strings
- CI builds passed
- Communicated to DevOps any deployment requirements
- Updated any necessary documentation (Confluence, contributing docs) or informed the documentation team

## 🦮 Reviewer guidelines

- 👍 (`:+1:`) or similar for great changes
- 📝 (`:memo:`) or ℹ️ (`:information_source:`) for notes or general info
- ❓ (`:question:`) for questions
- 🤔 (`:thinking:`) or 💭 (`:thought_balloon:`) for more open inquiry that's not quite a confirmed issue and could potentially benefit from discussion
- 🎨 (`:art:`) for suggestions / improvements
- ❌ (`:x:`) or ⚠️ (`:warning:`) for more significant problems or concerns needing attention
- 🌱 (`:seedling:`) or ♻️ (`:recycle:`) for future improvements or indications of technical debt
- ⛏ (`:pick:`) for minor or nitpick changes

PR #18391, opened 2026-01-15
[PM-30894] Support importing SSH keys from 1pux

unlimitedsola:
## 🎟️ Tracking

https://github.com/bitwarden/clients/issues/13977

https://bitwarden.atlassian.net/browse/PM-30894

## 📔 Objective

This is a continuation of https://github.com/bitwarden/clients/pull/10787, adding support for importing SSH keys from 1pux export format from 1Password.

## 📸 Screenshots

Tested with the web version:
image

### DIFF
--- a/libs/importer/src/importers/onepassword/onepassword-1pux-importer.ts
+++ b/libs/importer/src/importers/onepassword/onepassword-1pux-importer.ts
@@ -327,16 +327,12 @@ export class OnePassword1PuxImporter extends BaseImporter implements Importer {
           // Use sshKey.metadata.privateKey instead of the sshKey.privateKey field.
           // The sshKey.privateKey field doesn't have a consistent format for every item.
           const { privateKey } = field.value.sshKey.metadata;
-          try {
-            // Convert SSH key from PKCS#8 (1Password format) to OpenSSH format using SDK
-            // Note: 1Password does not store password-protected SSH keys, so no password handling needed for now
-            const parsedKey = import_ssh_key(privateKey);
-            cipher.sshKey.privateKey = parsedKey.privateKey;
-            cipher.sshKey.publicKey = parsedKey.publicKey;
-            cipher.sshKey.keyFingerprint = parsedKey.fingerprint;
-          } catch {
-            // If conversion fails, leave fields empty. We only import valid keys.
-          }
+          // Convert SSH key from PKCS#8 (1Password format) to OpenSSH format using SDK
+          // Note: 1Password does not store password-protected SSH keys, so no password handling needed for now
+          const parsedKey = import_ssh_key(privateKey);
+          cipher.sshKey.privateKey = parsedKey.privateKey;
+          cipher.sshKey.publicKey = parsedKey.publicKey;
+          cipher.sshKey.keyFingerprint = parsedKey.fingerprint;
           return;
         }
       }


harr1424:
Thanks, @unlimitedsola for your contribution!  Our internal product team is reviewing the functionality added by your PR. In the meantime, I'll give your code changes a cursory review and reach out with any questions or concerns.  In case you missed the link above, community contributed PRs undergo a [specific review process](https://contributing.bitwarden.com/contributing/pull-requests/community-pr-process) that can sometimes take longer than contributors expect. I appreciate your patience throughout this process, and welcome you to tag me here if you have any questions or would like a status update. Thanks again for your contribution!

libs/importer/src/importers/onepassword/onepassword-1pux-importer.ts
```diff
@@ -316,6 +322,23 @@ export class OnePassword1PuxImporter extends BaseImporter implements Importer {
...
+            // Convert SSH key from PKCS#8 (1Password format) to OpenSSH format using SDK
+            // Note: 1Password does not store password-protected SSH keys, so no password handling needed for now
+            const parsedKey = import_ssh_key(privateKey);
+            cipher.sshKey.privateKey = parsedKey.privateKey;
+            cipher.sshKey.publicKey = parsedKey.publicKey;
+            cipher.sshKey.keyFingerprint = parsedKey.fingerprint;
+          } catch {
+            // If conversion fails, leave fields empty. We only import valid keys.
+          }
+          return;
```
harr1424, line 336: ❓ This `try/catch` block caught my attention because it is unusual for this file, existing logic tends to return early and fail silently. The `try/catch` matches this behavior (failing silently), but the `generate_ssh_key()` function is documented as returning some useful error information: 

```
/**
 * Convert a PCKS8 or OpenSSH encrypted or unencrypted private key
 * to an OpenSSH private key with public key and fingerprint
 *
 * # Arguments
 * - `imported_key` - The private key to convert
 * - `password` - The password to use for decrypting the key
 *
 * # Returns
 * - `Ok(SshKey)` if the key was successfully coneverted
 * - `Err(PasswordRequired)` if the key is encrypted and no password was provided
 * - `Err(WrongPassword)` if the password provided is incorrect
 * - `Err(ParsingError)` if the key could not be parsed
 * - `Err(UnsupportedKeyType)` if the key type is not supported
 */
export function import_ssh_key(imported_key: string, password?: string | null): SshKeyView;
```

@djsmith85  do you think we should propagate these errors from the sdk method?

djsmith85: The current implementation of importers rely on all data being valid. If anything throws, it should abort the whole import action. If we can, we should provide a more detailed error message to the user. We could catch ssh-key parsing errors and include the error message in the `importResult.errorMessage`, but it needs to exit as soon as an issue is found.

unlimitedsola: Can the error reporting be improved in a separate PR? The current importer implementation doesn't seem to have an early exit code path built in mind.

Feel free to fork this PR and make adjustments if that has to be done before merging this.

harr1424: @unlimitedsola if you remove the `try/catch` the thrown error will propagate to the service and then Ui layer: 
```suggestion
      } else if (cipher.type === CipherType.SshKey) {
        if (valueKey === "sshKey") {
          // Use sshKey.metadata.privateKey instead of the sshKey.privateKey field.
          // The sshKey.privateKey field doesn't have a consistent format for every item.
          const { privateKey } = field.value.sshKey.metadata;
          // Convert SSH key from PKCS#8 (1Password format) to OpenSSH format using SDK
          // Note: 1Password does not store password-protected SSH keys, so no password handling needed for now
          const parsedKey = import_ssh_key(privateKey);
          cipher.sshKey.privateKey = parsedKey.privateKey;
          cipher.sshKey.publicKey = parsedKey.publicKey;
          cipher.sshKey.keyFingerprint = parsedKey.fingerprint;
          return;
```

unlimitedsola: @harr1424 I've added another commit that applied the suggested change. However, I did not test the updated behavior myself.

PR #18393, opened 2026-01-15
[PM-30799] added  html clean up for the domain

voommen-livefront:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-30799

## 📔 Objective

The event log renders HTML as it needs to render links. Therefore, we need to set the `innerHtml` of a div.
However, input into the `innerHtml` can be sanitized. 
For the Domain we sanitize the HTML.

This PR is in addition to the server PR - where we validate the `domainName` model.

### Input Sanitized
image

### Input that was not sanitized 
image


coroiu (approved):
oof

PR #18395, opened 2026-01-15
[PM-29236] Refactor of post-submit notification triggering logic

jprusik:
## 🎟️ Tracking

[PM-29236](https://bitwarden.atlassian.net/browse/PM-29236)

## 📔 Objective

New logic has been introduced to handle triggering notifications in form post-submit scenarios. The new logic is gated behind the `undetermined-cipher-scenario-logic` feature flag (with existing logic remaining unchanged). The new triggering logic particularly (though not exclusively) handles scenarios where the user vault is locked (and encrypted), deferring inferences of the type of form action that has taken place until the (already necessary) unlock state (when the vault ciphers unencrypted).

Significant unit testing has been added to this new triggering logic to ensure comprehensive scenario coverage and document expectations. The new logic is also designed to be easier to manage should the desired outcome of a particular scenario change, at the Product level.

Several typing improvements have been made as well.

This work does _not_ cover:
- notifications triggers outside of form submits
- form submit detection logic
- notification components
- form input qualifications

[PM-29236]: https://bitwarden.atlassian.net/browse/PM-29236?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/browser/src/autofill/background/notification.background.ts
+++ b/apps/browser/src/autofill/background/notification.background.ts
@@ -321,7 +321,7 @@ export default class NotificationBackground {
       type: CipherType.Login,
       reprompt,
       favorite,
-      ...(organizationCategories.length ? { organizationCategories } : {}),
+      ...(organizationCategories.length > 0 ? { organizationCategories } : {}),
       icon: buildCipherIcon(iconsServerUrl, view, showFavicons),
       login: login && { username: login.username },
     };
@@ -338,7 +338,7 @@ export default class NotificationBackground {
     activeUserId: UserId,
   ): Promise<LoginSecurityTaskInfo | null> {
     const tasks: SecurityTask[] = await this.getSecurityTasks(activeUserId);
-    if (!tasks?.length) {
+    if (!(tasks?.length > 0)) {
       return null;
     }
 
@@ -346,7 +346,7 @@ export default class NotificationBackground {
       modifyLoginData.uri,
       activeUserId,
     );
-    if (!urlCiphers?.length) {
+    if (!(urlCiphers?.length > 0)) {
       return null;
     }
 
@@ -907,13 +907,14 @@ export default class NotificationBackground {
 
     if (
       ciphers.length > 0 &&
-      currentPasswordFieldValue?.length &&
+      (currentPasswordFieldValue?.length || 0) > 0 &&
       // Only use current password for change if no new password present.
       !newPasswordFieldValue
     ) {
       const currentPasswordMatchesAnExistingValue = ciphers.some(
         (cipher) =>
-          cipher.login?.password?.length && cipher.login.password === currentPasswordFieldValue,
+          (cipher.login?.password?.length || 0) > 0 &&
+          cipher.login.password === currentPasswordFieldValue,
       );
 
       // The password entered matched a stored cipher value with
@@ -1027,7 +1028,7 @@ export default class NotificationBackground {
     }
 
     // If ciphers match entered username and new password values
-    if (usernameNewPasswordMatches.length) {
+    if (usernameNewPasswordMatches.length > 0) {
       // Early exit in these scenarios as they represent "no change"
       if (
         (
@@ -1042,7 +1043,7 @@ export default class NotificationBackground {
     }
 
     // If ciphers match entered username and password values
-    if (usernamePasswordMatches.length) {
+    if (usernamePasswordMatches.length > 0) {
       // and username, password, and new password values were entered
       if (
         inputScenario === inputScenarios.usernamePasswordNewPassword &&
@@ -1065,7 +1066,7 @@ export default class NotificationBackground {
     }
 
     // If ciphers match entered username value (only)
-    if (usernameOnlyMatches.length) {
+    if (usernameOnlyMatches.length > 0) {
       if (
         (
           [
@@ -1094,7 +1095,7 @@ export default class NotificationBackground {
     }
 
     // If ciphers match entered new password value (only)
-    if (newPasswordOnlyMatches.length) {
+    if (newPasswordOnlyMatches.length > 0) {
       // Early exit in these scenarios
       if (
         (
@@ -1124,7 +1125,7 @@ export default class NotificationBackground {
     }
 
     // If ciphers match entered password value (only)
-    if (passwordOnlyMatches.length) {
+    if (passwordOnlyMatches.length > 0) {
       if (
         (
           [


apps/browser/src/autofill/background/notification.background.ts
```diff
@@ -730,6 +1187,7 @@ export default class NotificationBackground {
     });
   }
 
+  // @TODO this needs the whole input record, and not just newPassword
```
jprusik, line 1179: Regarding the TODOs here and elsewhere on the inputs records passed to the notification:

They are not handled here as an out-of-scope concern which will be made easier by post-feature flagged logic removals.

apps/browser/src/autofill/background/notification.background.spec.ts
```diff
@@ -686,6 +690,1492 @@ describe("NotificationBackground", () => {
...
+        const formEntryData: ModifyLoginCipherFormData = {
+          newPassword: "Bab3lPhs5h",
+          password: "I<3VogonPoetry",
+          uri: mockFormURI,
+          username: "ADent",
+        };
+
+        const storedCiphersForURL = [
+          mock<CipherView>({ login: { username: "ADent", password: "I<3VogonPoetry" } }),
+        ];
```
blackwood, line 815: ❓ Any value in having a `defaultFormEntryData` and `defaultStoredCiphersForURL` for these frequently reused values?

apps/browser/src/autofill/background/notification.background.spec.ts
```diff
@@ -686,6 +690,1492 @@ describe("NotificationBackground", () => {
...
+
+        activeAccountStatusMock$.next(AuthenticationStatus.LoggedOut);
+        getAllDecryptedForUrlSpy.mockResolvedValueOnce(storedCiphersForURL);
+
+        await notificationBackground.triggerCipherNotification(formEntryData, tab);
+
+        expect(getAllDecryptedForUrlSpy).not.toHaveBeenCalled();
+        expect(pushChangePasswordToQueueSpy).not.toHaveBeenCalled();
+        expect(pushAddLoginToQueueSpy).not.toHaveBeenCalled();
```
blackwood: ❓ `expectSkippedCheckingNotification` could be a convenience function that wraps this frequent set of grouped calls.

apps/browser/src/autofill/background/notification.background.spec.ts
```diff
@@ -686,6 +690,1492 @@ describe("NotificationBackground", () => {
...
+      });
+
+      it("skips checking if a notification should trigger if the vault is locked and there is no value for the `newPassword` field", async () => {
+        const formEntryData: ModifyLoginCipherFormData = {
+          newPassword: "Beeblebrox4Prez",
```
blackwood: ❓ Should there be no value here, according to the spec?

apps/browser/src/autofill/background/notification.background.spec.ts
```diff
@@ -686,6 +690,1492 @@ describe("NotificationBackground", () => {
...
+          expect(getAllDecryptedForUrlSpy).not.toHaveBeenCalled();
+          expect(pushAddLoginToQueueSpy).not.toHaveBeenCalled();
+
+          expect(pushChangePasswordToQueueSpy).toHaveBeenCalledWith(
+            null,
+            mockFormattedURI,
+            formEntryData.newPassword,
+            tab,
+            true,
+          );
```
blackwood, line 922: 💭 I think I'm inferring/recalling that the signature of `pushChangePasswordToQueue` is how we determine that an unlock notification is triggered. This might be a smell to clean up later; in the meantime we might want to document that in a comment in this file before we use it as an indicator of notification type.

apps/browser/src/autofill/background/notification.background.spec.ts
```diff
@@ -686,6 +690,1492 @@ describe("NotificationBackground", () => {
...
+              login: { password: "sting123", username: "BBaggins" },
+            }),
+            mock<CipherView>({
+              id: "cipher-id-3",
+              login: { password: "Edro2x", username: "gandalfG" },
+            }),
+            mock<CipherView>({
+              id: "cipher-id-4",
+              login: { password: "UShallKnotPassword", username: "gandalfG" },
+            }),
```
blackwood, line 1205: 💭 If I understand the intent of the spec correctly, the presence of form `newPassword` matching the prior cipher's `password` suppresses all notifications, even if other combinations are present (since there's at least one matching cipher w/ new password)?

Do we account for a user using identical usernames on two separate subdomains matched by the same domain?

apps/browser/src/autofill/background/notification.background.spec.ts
```diff
@@ -686,6 +690,1492 @@ describe("NotificationBackground", () => {
...
+          ];
+
+          activeAccountStatusMock$.next(AuthenticationStatus.Unlocked);
+          getAllDecryptedForUrlSpy.mockResolvedValueOnce(storedCiphersForURL);
+
+          await notificationBackground.triggerCipherNotification(formEntryData, tab);
+
+          expect(pushChangePasswordToQueueSpy).not.toHaveBeenCalled();
+          expect(pushAddLoginToQueueSpy).not.toHaveBeenCalled();
+        });
```
blackwood, line 1749: 💭 Do we currently capture the password in this scenario? (`password` and `newPassword` are filled, but there's no matching cipher). I'm curious why we wouldn't trigger add login.

apps/browser/src/autofill/background/notification.background.spec.ts
```diff
@@ -686,6 +690,1492 @@ describe("NotificationBackground", () => {
...
+
+        it("and cipher update candidates match `username` AND `password` as well as `username` OR `password`, do not trigger a notification (nothing to change)", async () => {
```
blackwood: ✏️ I think "as well" is slightly ambiguous and might have thrown me off in the prior example I annotated. (e.g. "he cooks as well as cleans" has the sense of both/either scenario). Can we swap for "while/plus others match"? 

*"and cipher update candidates match `username` AND `password`, plus others match `username` OR `password` [...]*

apps/browser/src/autofill/background/notification.background.ts
```diff
@@ -596,6 +625,228 @@ export default class NotificationBackground {
...
+      inputScenario = inputScenarios.password;
+    } else if (currentPasswordFieldHasValue && usernameFieldHasValue && !newPasswordFieldHasValue) {
+      inputScenario = inputScenarios.usernamePassword;
+    } else if (
+      newPasswordFieldHasValue &&
+      !currentPasswordFieldHasValue &&
+      !usernameFieldHasValue
+    ) {
+      inputScenario = inputScenarios.newPassword;
+    }
```
blackwood: If we change the values of `inputScenarios` to be strings of sequential case insensitive identifiers:

```ts
const inputScenarios = {
  usernamePasswordNewPassword: "upn",
  usernameNewPassword: "un",
  usernamePassword: "up",
  username: "u",
  passwordNewPassword: "pn",
  newPassword: "n",
  password: "p",
} as const;

const validScenarios = new Set(Object.values(inputScenarios));

```

Then we should be able to replace this entire `if-else` block with:

```ts
const key =
  `${usernameFieldHasValue ? 'u' : ''}` +
  `${currentPasswordFieldHasValue ? 'p' : ''}` +
  `${newPasswordFieldHasValue ? 'n' : ''}`;

const inputScenario = validScenarios.has(key) ? (key as InputScenario) : null;
```

apps/browser/src/autofill/background/notification.background.ts
```diff
@@ -710,6 +961,212 @@ export default class NotificationBackground {
...
+        );
+
+        return true;
+      }
+
+      return false;
+    }
+
+    // If ciphers match entered username and new password values
+    if (usernameNewPasswordMatches.length) {
```
blackwood: ✏️ nit: let's prefer explicit comparision i.e. (`> 0`)

apps/browser/src/autofill/background/notification.background.spec.ts
jprusik: We had those abstracted mock values in place with overrides elsewhere. Even though it was more concise, it made understanding the vault cipher match scenario configurations difficult to quickly understand. This is one place where I think the verbosity serves us better.

blackwood (approved):

apps/browser/src/autofill/background/notification.background.spec.ts
jprusik: damn, nice catch 🚀

apps/browser/src/autofill/background/notification.background.spec.ts
jprusik: Right - this is going to be an unlock notification, but it passes through this update notification handler (so it knows what to summon upon unlock). I agree it feels off (like some other patterns inherited from the existing/non-flagged codepath), but opted to leave it alone to avoid scope creep and potential complexity adds to this iteration (we're coming back).

In the meantime, I'll add some signposting.

blackwood: The first argument is enough to tell the difference also, I think: URI v. null v. cipher array

apps/browser/src/autofill/background/notification.background.spec.ts
jprusik: 1) yes, the idea is they already have the values that were entered, so other close matches are irrelevant.

2) That's handled by the matching logic with the call to `this.cipherService.getAllDecryptedForUrl`. The scope of this logic assumes all ciphers returned (from that call) belong to the current/passed page. If we do find subdomains match inappropriately as you've described, we'd want to address that at the `cipherService` level (that is, this comes after the cipher url matching logic).

apps/browser/src/autofill/background/notification.background.spec.ts
jprusik: The short answer is logic parity with the existing logic for a poorly defined/weak case.

The longer answer is that it is also relevant that no username field was filled, and the password and new password values do not match each other (which would represent no change). Since neither the password nor the new password value match anything in the vault, it can only be safely inferred to be a new value. Since we also don't have a username, we're _probably_ dealing with a scenario in which the user has previously opted not to store the cipher (in which case we'll be bugging them every time they perform the operation), but could be an edgier case like signup or login with no username.

In any case, I'm open to changing the outcome of this scenario with Product guidance (cc @wsalmon-bw), but ideally as a followup concern/task. Getting the scenario framework in place will enable much quicker iterations (this logic is behind a feature flag, as well)

apps/browser/src/autofill/background/notification.background.spec.ts
jprusik: addressed

apps/browser/src/autofill/background/notification.background.ts
jprusik: addressed, with some small adjustments

apps/browser/src/autofill/background/notification.background.ts
jprusik: agreed 👍

apps/browser/src/autofill/background/notification.background.spec.ts
jprusik: Discussed offline with Product and we're following up in https://github.com/bitwarden/clients/pull/18729

PR #18404, opened 2026-01-16
[CL-695] Combine display of CL and autofill storybooks

vleague2:
## 🎟️ Tracking

[CL-695](https://bitwarden.atlassian.net/browse/CL-695)

## 📔 Objective

This PR adds the Autofill storybook instance to the CL storybook instance in production, so that all of the components in both instances are displayed together. This is accomplished by using [Storybook composition](https://storybook.js.org/docs/sharing/storybook-composition), which does not actually "combine" them -- it just links to the secondary instance (in this case Autofill) from the primary instance (CL). This means that the building, running, and publishing of each instance still happens separately. 

Github actions will now run separate actions for Storybook publishing, UI tests, and UI review for the CL (`clients`) and autofill projects. Because of the above explanation about Storybook composition, this is just the way it has to be in order to display everything together in prod. The `clients` publish action will display the autofill stories for the current branch, so you can view them together as they will be in production. The `clients` UI test and review actions will not include snapshots of the autofill stories though, so any changes to those stories will be reviewed through the separate autofill actions.

The autofill UI Review and UI Tests will not be required actions as of now, and I am unsure if we have appetite to make them required. Open to thoughts there. 

For local runs of storybook, there are a few new scripts available to allow engineers to optionally view the stories combined locally. The autofill storybook instance can also now be run from the root of `clients`.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

[CL-695]: https://bitwarden.atlassian.net/browse/CL-695?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


package.json
```diff
@@ -26,7 +26,10 @@
     "lint:dep-ownership": "tsc --project ./scripts/tsconfig.json && node ./scripts/dist/dep-ownership.js",
     "lint:sdk-internal-versions": "tsc --project ./scripts/tsconfig.json && node ./scripts/dist/sdk-internal-versions.js",
     "docs:json": "compodoc -p ./tsconfig.json -e json -d . --disableRoutesGraph",
-    "storybook": "ng run components:storybook",
+    "storybook": "concurrently \"npm run storybook-cl\" \"npm run storybook-autofill\"",
+    "storybook-cl": "ng run components:storybook",
+    "storybook-autofill": "cd apps/browser/src/autofill/content/components && npm run storybook:lit:combined",
+    "storybook-autofill-build:ci": "cd apps/browser/src/autofill/content/components && npm run storybook:build-lit:ci",
```
vleague2: ℹ️  The changes here have the following impact:
- The usual `npm run storybook` command will now run both the CL and autofill storybooks, and combine their log output in the terminal. Both sets of stories can be viewed at the normal CL location (`localhost:6006`).
- If the combined log output is annoying and/or you need to run just one of them, you can still run the CL and autofill storybooks separately. The CL storybook will show an error in the sidebar where it is trying to connect to the autofill storybook, but that doesn't affect the CL stories. Autofill's storybook continues to function as per usual individually.
- The autofill storybook can now be run from the repo root instead of having to `cd` into the autofill components directory. However, when running from the root, it will use a new autofill `package.json` script that does not auto-open the storybook page, and it uses `localhost:6007`. These two aspects are there so that things work nicely in the combined run, because the autofill instance can't run on the same port as the CL and you don't want to auto-open both when the CL instance will now display autofill stories.

Open to thoughts/concerns on these changes and how they might affect the dev storybook experience!

BryanCunningham (dismissed):

.github/workflows/chromatic.yml
```diff
@@ -61,25 +70,37 @@ jobs:
...
       - name: Install Node dependencies
-        if: steps.get-changed-files-for-chromatic.outputs.storyFiles == 'true'
+        if: steps.get-changed-files-for-chromatic-cl.outputs.storyFiles == 'true' || steps.get-changed-files-for-chromatic-autofill.outputs.storyFiles == 'true'
         run: npm ci
 
+      - name: Get branch name
+        id: branch-names
+        uses: tj-actions/branch-names@5250492686b253f06fa55861556d1027b067aeb5 # v9
+        with:
+          replace_slashes_with_hyphens: 'true'
```
vleague2: In order to dynamically display the correct autofill stories alongside the CL stories (i.e.  if we're in `main`, we want the autofill stories from `main`, but if we're in another branch, we want to use that branch's stories), we need to get the current branch name and set it as an environment variable that the Storybook build script can access. Chromatic has a handy way to view a branch's stories by using the branch name, so we can point to the correct location with just the branch name. Branch names are apparently not set in a single variable consistently in Github, so this action checks the multiple places it could be depending on the trigger type and also will swap out the slashes for us (Chromatic URLs do not support the slashes).

vleague2: Removed in favor of doing it ourselves to reduce security risks

package.json
```diff
@@ -27,8 +27,12 @@
     "lint:sdk-internal-versions": "tsc --project ./scripts/tsconfig.json && node ./scripts/dist/sdk-internal-versions.js",
     "docs:json": "compodoc -p ./tsconfig.json -e json -d . --disableRoutesGraph",
     "storybook": "ng run components:storybook",
+    "storybook:combined": "STORYBOOK_DEV=\"combined\" concurrently \"npm run storybook\" \"npm run storybook:autofill:combined\"",
+    "storybook:autofill": "cd apps/browser/src/autofill/content/components && npm run storybook:lit",
```
jprusik, line 31: I'm getting an error (`Conflicting values for 'process.env.NODE_ENV'`) when running these commands from the `clients` root, but the autofill storybook still works, so non-blocking for our considerations

Image

vleague2: Ty for finding that, I shall take a look!

vleague2: @jprusik Ok I know this was like a week ago but do you know what version of node you are using? I'm not seeing the same errors, both of those commands are working fine for me from the clients root

jprusik: `v22.22.0` (with npm `v10.9.4`) presumably, but I just tested now and didn't get the error, so maybe I just forgot to switch with `nvm` ><

package.json
```diff
@@ -27,8 +27,12 @@
...
+    "storybook:autofill:combined": "cd apps/browser/src/autofill/content/components && npm run storybook:lit:combined",
     "build-storybook": "ng run components:build-storybook",
     "build-storybook:ci": "ng run components:build-storybook --webpack-stats-json",
+    "build-storybook:autofill:ci": "cd apps/browser/src/autofill/content/components && npm run storybook:build-lit:ci",
```
willmartian, line 35: What do you think about:
- `npm run storybook` (combined)
- `npm run storybook:lit`
- `npm run storybook:angular`

vleague2: I originally had it where the `npm run storybook` was the combined view, but when we last talked about it with Bryan, we felt like the extra build time and noise introduced in the console (because you get logs for both storybooks running) wasn't worth it when we are primarily dev-ing against the CL storybook and not referencing the autofill stories. But we can definitely discuss again!

willmartian: I am good either way!

willmartian (approved):

PR #18409, opened 2026-01-16
[PM-24178] Handle focus when routed dialog closes in vault table

vleague2:
## 🎟️ Tracking

[PM-24178](https://bitwarden.atlassian.net/browse/PM-24178)

## 📔 Objective

This PR ensures that dialogs opened from the vault table (aka routed dialogs) return focus back to their trigger element, which is the cipher link in the vault table.

Enabling this functionality involved redoing the approach to the router focus management pipeline to allow custom focus targets to be passed in. So there are some minor refactor changes to the existing focus management implementation. 

**Vault**:
To direct focus back to the correct cipher, I've added an id to the cipher button and modified the vault code to pass that id to the router when navigating to the view/edit dialog and the credential reprompt dialog. I would appreciate some eyes on unintended consequences of those changes!

**UIF**:
The largest change to the existing router focus management pipeline is waiting for ngZone to stabilize, which fixes an issue where the pipeline ran and then was superseded by other page load behavior. Now that the focus runs after ngZone stabilizes, we need to manually pass in the `body` element as the focus target for pages that should have a "full page reload" focus behavior (like navigating between products in the product switcher). 

## 📸 Screenshots

Before:

https://github.com/user-attachments/assets/5ca8f1f7-d32d-4d05-812b-364bf48da295

After:

https://github.com/user-attachments/assets/8aec1f32-875c-4585-b109-e00ea19a054f

[PM-24178]: https://bitwarden.atlassian.net/browse/PM-24178?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/components/src/a11y/router-focus-manager.service.ts
+++ b/libs/components/src/a11y/router-focus-manager.service.ts
@@ -6,6 +6,8 @@ import { skip, filter, combineLatestWith, tap, map, firstValueFrom } from "rxjs"
 import { FeatureFlag } from "@bitwarden/common/enums/feature-flag.enum";
 import { ConfigService } from "@bitwarden/common/platform/abstractions/config/config.service";
 
+import { queryForAutofocusDescendents } from "../input";
+
 @Injectable({ providedIn: "root" })
 export class RouterFocusManagerService {
   private router = inject(Router);
@@ -56,7 +58,14 @@ export class RouterFocusManagerService {
 
   private focusTargetEl(elSelector: string) {
     const targetEl = document.querySelector<HTMLElement>(elSelector);
-    if (targetEl) {
+    const mainEl = document.querySelector<HTMLElement>("main");
+
+    const pageHasAutofocusEl = mainEl && queryForAutofocusDescendents(mainEl).length > 0;
+
+    if (pageHasAutofocusEl) {
+      // do nothing because autofocus will handle the focus
+      return;
+    } else if (targetEl) {
       targetEl.focus();
     } else {
       // eslint-disable-next-line no-console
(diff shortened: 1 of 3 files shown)


BryanCunningham (approved):

JimmyVo16 (dismissed):

vleague2:
Going to wait for #18657 to merge first. They don't touch the exact same areas but I'd like to test this ticket with those changes in it too.

shane-melton (commented):
Vault code changes look good, though I cannot seem to get it to behave as expected on my local build of the branch in Chromium / Arc.

https://github.com/user-attachments/assets/e0c3bdbe-4c8d-4dda-98df-578aa8b6f5a6

Otherwise, one potential side effect (that may or may not be desired) is when a user opens a direct link to a vault item without interacting with the trigger element. I believe focus will be returned to the cipher row even though that's not technically where focus was last. I'm not sure where the expected focus should be in such a scenario; but wanted to make sure that use case was considered.

vleague2:
> Otherwise, one potential side effect (that may or may not be desired) is when a user opens a direct link to a vault item without interacting with the trigger element. I believe focus will be returned to the cipher row even though that's not technically where focus was last. I'm not sure where the expected focus should be in such a scenario; but wanted to make sure that use case was considered.

For posterity -- we discussed the direct link scenario and feel like returning focus to a contextually relevant location like the vault table is an improvement over the current experience of focus returning to the top of the page. It's been added to the QA notes as well!

vleague2:
This had a QA kickback which has now been fixed, and it's passed QA, so it is now ready for re-review. Sorry for the bunch of re-requests on this one, y'all!

PR #18413, opened 2026-01-16
[PM-27486] Remove feature flag PM25174_DisableType0Decryption

Thomas-Avery:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-27486

## 📔 Objective

The objective of this PR is to remove the feature flag `PM25174_DisableType0Decryption` from clients. This also removes the temporary `init` on the `EncryptService` used to inject the flag.


libs/common/src/key-management/crypto/services/encrypt.service.spec.ts
```diff
@@ -300,16 +301,18 @@ describe("EncryptService", () => {
...
       expect(PureCrypto.unwrap_symmetric_key).toHaveBeenCalledWith(
         encString.encryptedString,
         key.toEncoded(),
       );
     });
-    it("throws if disableType0Decryption is enabled and type is AesCbc256_B64", async () => {
-      encryptService.setDisableType0Decryption(true);
+    it("throws if type is AesCbc256_B64", async () => {
       const key = new SymmetricCryptoKey(makeStaticByteArray(64));
       const encString = new EncString(EncryptionType.AesCbc256_B64, "wrapped_symmetric_key");
```
quexten, line 317: Thought: We should - if we have not done so - track cleaning up the aes256-cbc-b64 encryption type variant, since the only use of it is master-key, and that is encapsulated away into master-password-unlock-method, so in practice it should never occur in TS. Then we can also drop these tests.

Thomas-Avery: I made a ticket for this PM-31388. After thinking on this some more I do think it will require us to move key connector off of the MasterKey concept prior to cleaning up the encryption type variant.

PR #18399, opened 2026-01-16
[PM-30250] Prevent over scrolling on Inactive 2FA and Unsecure Website reports

lastbestdev:
## 🎟️ Tracking
https://bitwarden.atlassian.net/browse/PM-30250

## 📔 Objective
These changes fix a bug that allowed users to scroll past the content of the Inactive 2FA and Unsecure website reports. This was handled by replacing the old `bit-table` component with the virtualized `bit-table-scroll` component. This change follows the pattern for tables used in other existing report types.

It also includes a fix to properly show the "Back to reports" button on page load for all of the reports. (i.e. button should show when refreshing any report page)

## 📸 Screenshots
https://github.com/user-attachments/assets/93552cf8-d7fa-4ff9-9b83-890255d9dcae

PR #18416, opened 2026-01-16
[PM-30918] Migrate DIRT components to new Angular control flow syntax

lastbestdev:
## 🎟️ Tracking
https://bitwarden.atlassian.net/browse/PM-30918

## 📔 Objective
**No functionality changed by this PR.** These changes update old Angular control flow usages `ngIf, ngFor, ...` to new syntax (`@if, @else, @for`) in all DIRT owned components. This was quick and straightforward to do using [Angular's CLI tool](https://angular.dev/reference/migrations/control-flow).

Files updated include all template `.html` files under DIRT team ownership (`apps/web/src/app/dirt/` and `bitwarden-license/bit-web/src/app/dirt` directories)

## Testing
Verify that Reporting, and Access Intelligence features still function as expected.

### DIFF
--- a/apps/browser/src/autofill/fido2/content/fido2-page-script.ts
+++ b/apps/browser/src/autofill/fido2/content/fido2-page-script.ts
@@ -1,12 +1,10 @@
-// FIXME: Update this file to be type safe and remove this and next line
-// @ts-strict-ignore
 import { WebauthnUtils } from "../utils/webauthn-utils";
 
 import { MessageTypes } from "./messaging/message";
 import { Messenger } from "./messaging/messenger";
 
 (function (globalContext) {
-  if (globalContext.document.currentScript) {
+  if (globalContext.document.currentScript?.parentNode) {
     globalContext.document.currentScript.parentNode.removeChild(
       globalContext.document.currentScript,
     );
@@ -86,7 +84,7 @@ import { Messenger } from "./messaging/messenger";
    */
   async function createWebAuthnCredential(
     options?: CredentialCreationOptions,
-  ): Promise<Credential> {
+  ): Promise<Credential | null> {
     if (!isWebauthnCall(options)) {
       return await browserCredentials.create(options);
     }
@@ -106,13 +104,18 @@ import { Messenger } from "./messaging/messenger";
         options?.signal,
       );
 
-      if (response.type !== MessageTypes.CredentialCreationResponse) {
+      if (response.type !== MessageTypes.CredentialCreationResponse || !response.result) {
         throw new Error("Something went wrong.");
       }
 
       return WebauthnUtils.mapCredentialRegistrationResult(response.result);
     } catch (error) {
-      if (error && error.fallbackRequested && fallbackSupported) {
+      if (
+        fallbackSupported &&
+        error instanceof Object &&
+        "fallbackRequested" in error &&
+        error.fallbackRequested
+      ) {
         await waitForFocus();
         return await browserCredentials.create(options);
       }
@@ -127,7 +130,9 @@ import { Messenger } from "./messaging/messenger";
    * @param options Options for creating new credentials.
    * @returns Promise that resolves to the new credential object.
    */
-  async function getWebAuthnCredential(options?: CredentialRequestOptions): Promise<Credential> {
+  async function getWebAuthnCredential(
+    options?: CredentialRequestOptions,
+  ): Promise<Credential | null> {
     if (!isWebauthnCall(options)) {
       return await browserCredentials.get(options);
     }
@@ -153,7 +158,7 @@ import { Messenger } from "./messaging/messenger";
             internalAbortController.signal,
           );
           internalAbortController.signal.removeEventListener("abort", abortListener);
-          if (response.type !== MessageTypes.CredentialGetResponse) {
+          if (response.type !== MessageTypes.CredentialGetResponse || !response.result) {
             throw new Error("Something went wrong.");
           }
 
@@ -176,7 +181,7 @@ import { Messenger } from "./messaging/messenger";
       abortSignal.removeEventListener("abort", abortListener);
       internalAbortControllers.forEach((controller) => controller.abort());
 
-      return response;
+      return response ?? null;
     }
 
     try {
@@ -188,13 +193,18 @@ import { Messenger } from "./messaging/messenger";
         options?.signal,
       );
 
-      if (response.type !== MessageTypes.CredentialGetResponse) {
+      if (response.type !== MessageTypes.CredentialGetResponse || !response.result) {
         throw new Error("Something went wrong.");
       }
 
       return WebauthnUtils.mapCredentialAssertResult(response.result);
     } catch (error) {
-      if (error && error.fallbackRequested && fallbackSupported) {
+      if (
+        fallbackSupported &&
+        error instanceof Object &&
+        "fallbackRequested" in error &&
+        error.fallbackRequested
+      ) {
         await waitForFocus();
         return await browserCredentials.get(options);
       }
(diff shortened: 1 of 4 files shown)


apps/web/src/app/dirt/reports/pages/breach-report.component.html
```diff
@@ -12,45 +12,58 @@
...
+          <bit-callout type="success" title="{{ 'goodNews' | i18n }}">
+            {{ "breachUsernameNotFound" | i18n: checkedUsername }}
+          </bit-callout>
+        }
+        @if (breachedAccounts.length) {
+          <bit-callout type="danger" title="{{ 'breachFound' | i18n }}">
+            {{ "breachUsernameFound" | i18n: checkedUsername : breachedAccounts.length }}
+          </bit-callout>
+        }
+        @if (breachedAccounts.length) {
```
Banrion: ⚠️ We should combine duplicate if statements into one.

apps/web/src/app/dirt/reports/pages/breach-report.component.html
```diff
@@ -12,45 +12,58 @@
...
+          <ul
+            class="tw-list-none tw-flex-col tw-divide-x-0 tw-divide-y tw-divide-solid tw-divide-secondary-300 tw-rounded tw-border tw-border-solid tw-border-secondary-300 tw-p-0"
+          >
+            @for (a of breachedAccounts; track a) {
```
Banrion, line 31: 👍 Nice that the cli tool also updated the ngfor syntax. I don't think we had a ticket for that. We may want to call it out in the description as a change.

apps/web/src/app/dirt/reports/pages/exposed-passwords-report.component.html
```diff
@@ -5,95 +5,110 @@
...
-    </ng-container>
-  </div>
+  @if (hasLoaded) {
+    <div class="tw-mt-4">
+      @if (!ciphers.length) {
+        <bit-callout type="success" title="{{ 'goodNews' | i18n }}">
+          {{ "noExposedPasswords" | i18n }}
+        </bit-callout>
+      }
+      @if (ciphers.length) {
```
Banrion: ⚠️ Similar to the comment about combining if statements, we should use the else syntax.

apps/web/src/app/dirt/reports/pages/inactive-two-factor-report.component.html
```diff
@@ -2,110 +2,125 @@
...
+                <th bitCell></th>
+                <th bitCell>{{ "name" | i18n }}</th>
+                <th bitCell>{{ "owner" | i18n }}</th>
+                <th bitCell></th>
+              </tr>
+            </ng-container>
+          }
+          <tbody>
+            <ng-template body let-rows$>
+              <!-- prettier-ignore -->
```
Banrion: ❓ Why is prettier-ignore needed here?

lastbestdev: The parentheses used on the next line `(rows$ | async)`  to force the `rows$` variable through the async pipe before iteration are removed by prettier formatting. This causes the template to fail to render properly.

These usages will be removed anyway, since these components will be updated to use the virtual scroll table in https://github.com/bitwarden/clients/pull/18399

apps/web/src/app/dirt/reports/pages/exposed-passwords-report.component.html
lastbestdev: Updated

apps/web/src/app/dirt/reports/pages/breach-report.component.html
lastbestdev: Updated description a bit to make `@for` changes clearer

bitwarden_license/bit-web/src/app/dirt/access-intelligence/risk-insights.component.html
```diff
@@ -61,23 +63,25 @@
...
+                    {{ "riskInsightsRunReport" | i18n }}
+                  </button>
+                }
                 <span>
-                  <i
-                    *ngIf="isRunningReport"
-                    class="bwi bwi-spinner bwi-spin tw-text-muted tw-text-[1.2rem]"
-                    aria-hidden="true"
-                  ></i>
+                  @if (isRunningReport) {
```
Banrion: 🤔 Not blocking, but I wonder if the span was left out of the if statement for any reason. If it's not required, this could be a simple if else without a floating span.

bitwarden_license/bit-web/src/app/dirt/access-intelligence/empty-state-card.component.html
```diff
@@ -38,69 +37,77 @@
...
-          aria-hidden="true"
-        ></video>
-      } @else if (icon()) {
-        <div
-          class="tw-size-full tw-flex tw-items-center tw-justify-center tw-bg-secondary-100 tw-rounded-lg"
-        >
-          <bit-icon
-            [icon]="icon()"
-            class="tw-size-12 sm:tw-size-16 tw-text-muted"
+  @if (videoSrc() || icon()) {
```
Banrion: ⛏️ These two statements can be combined "videoSrc() || icon()" to one

Banrion (approved):
💭 The amount of nested statements in this changes makes me think there could be some refactoring on similar components or to the view logic.

bitwarden_license/bit-web/src/app/dirt/access-intelligence/empty-state-card.component.html
lastbestdev: I'm not sure about this, the individual conditionals look to be placed to prevent rendering the components if either `videoSrc()` or `icon()` come back as undefined.

The outer conditional prevents rendering the surrounding divs when both `videoSrc()` and `icon()` are undefined

lastbestdev: Nevermind, found this in my editor and fixed

apps/web/src/app/dirt/reports/pages/inactive-two-factor-report.component.html
lastbestdev: `prettier-ignore` usage removed in latest commit

bitwarden_license/bit-web/src/app/dirt/access-intelligence/risk-insights.component.html
lastbestdev: After looking into this, `isRunningReport` is never actually true when this component is rendered. This is because the `dirt-report-loading` component is rendered while the report is being generated. 

I went ahead and removed this condition altogether.

bitwarden_license/bit-web/src/app/dirt/access-intelligence/risk-insights.component.html
```diff
@@ -72,13 +72,6 @@
                 >
                   {{ "riskInsightsRunReport" | i18n }}
                 </button>
-                <span>
-                  <i
-                    *ngIf="isRunningReport"
-                    class="bwi bwi-spinner bwi-spin tw-text-muted tw-text-[1.2rem]"
-                    aria-hidden="true"
-                  ></i>
-                </span>
```
lastbestdev, line 81: Conditional rendering here has been removed, since this template never renders while the Access Intelligence report is generating. The `dirt-report-loading` component is rendered (L12-13) until the report is complete

lastbestdev:
Merged since branch was tested before merge conflicts popped up. Verify that the chip select appears properly on Reports with results from 5+ Organizations, and the empty state icon for Access Intelligence appears as expected (both shown below)

Reports chip select
image

Access Intelligence empty state
image

PR #18417, opened 2026-01-16
[CL-1020] background color updates

BryanCunningham:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/CL-1020

## 📔 Objective

Update foundational background colors to use new palette

### DIFF
--- a/apps/browser/src/_locales/en/messages.json
+++ b/apps/browser/src/_locales/en/messages.json
@@ -5673,6 +5673,9 @@
   "extraWide": {
     "message": "Extra wide"
   },
+  "narrow": {
+    "message": "Narrow"
+  },
   "sshKeyWrongPassword": {
     "message": "The password you entered is incorrect."
   },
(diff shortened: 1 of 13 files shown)


vleague2 (approved):

libs/components/tailwind.config.base.js
```diff
@@ -72,11 +72,11 @@ module.exports = {
         code: rgba("--color-text-code"),
       },
       background: {
-        DEFAULT: rgba("--color-background"),
-        alt: rgba("--color-background-alt"),
-        alt2: rgba("--color-background-alt2"),
-        alt3: rgba("--color-background-alt3"),
-        alt4: rgba("--color-background-alt4"),
```
willmartian, line 79: We still have these CSS variables defined in `tw-theme.css`; should we not update those instead of the Tailwind config?

BryanCunningham: @willmartian Maybe... All the legacy variables will eventually be gone. Made more sense to me to just skip mapping them to new colors and adjust the config instead

willmartian (approved):

PR #18419, opened 2026-01-16
[PM-30906] Auto confirm nudge service fix and better nudge documentation

shane-melton:
## 🎟️ Tracking

[PM-30906](https://bitwarden.atlassian.net/browse/PM-30906)

## 📔 Objective

The `AutoConfirmNudgeService` depends on the `AutomaticUserConfirmationService` which is not provided/applicable in the Desktop client which caused dependency injection issues. This PR refactors the `AutoConfirmNudgeService` to be Browser specific as that is the only place the nudge is used currently. 

This also adds a `README.md` to the custom-nudges-service directory that provides better guidance on adding new nudges for other teams.

[PM-30906]: https://bitwarden.atlassian.net/browse/PM-30906?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/angular/src/vault/services/custom-nudges-services/README.md
+++ b/libs/angular/src/vault/services/custom-nudges-services/README.md
@@ -1,0 +1,204 @@
+# Custom Nudge Services
+
+This folder contains custom implementations of `SingleNudgeService` that provide specialized logic for determining when nudges should be shown or dismissed.
+
+## Architecture Overview
+
+### Core Components
+
+- **`NudgesService`** (`../nudges.service.ts`) - The main service that components use to check nudge status and dismiss nudges
+- **`SingleNudgeService`** - Interface that all nudge services implement
+- **`DefaultSingleNudgeService`** - Base implementation that stores dismissed state in user state
+- **Custom nudge services** - Specialized implementations with additional logic
+
+### How It Works
+
+1. Components call `NudgesService.showNudgeSpotlight$()` or `showNudgeBadge$()` with a `NudgeType`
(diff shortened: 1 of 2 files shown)


libs/angular/src/vault/services/default-single-nudge.service.ts
```diff
@@ -22,7 +22,11 @@ export interface SingleNudgeService {
   providedIn: "root",
 })
 export class DefaultSingleNudgeService implements SingleNudgeService {
-  stateProvider = inject(StateProvider);
+  protected stateProvider: StateProvider;
+
+  constructor(stateProvider: StateProvider) {
```
shane-melton, line 27: ℹ️ Switched to constructor based dependency injection to support the `safeProvider` type safety in manually registered services.

libs/angular/src/vault/services/custom-nudges-services/new-account-nudge.service.ts
```diff
@@ -18,8 +19,13 @@ const THIRTY_DAYS_MS = 30 * 24 * 60 * 60 * 1000;
   providedIn: "root",
 })
 export class NewAccountNudgeService extends DefaultSingleNudgeService {
-  vaultProfileService = inject(VaultProfileService);
-  logService = inject(LogService);
+  constructor(
```
shane-melton, line 22: ℹ️ Switched to constructor based dependency injection to support the `safeProvider` type safety in manually registered services.

PR #18423, opened 2026-01-17
[PM-27086] Use new KM Data Types in InputPasswordComponent flows - Account Recovery

rr-bw:
## 🎟️ Tracking

[PM-27086](https://bitwarden.atlassian.net/browse/PM-27086)

## 📔 Objective

Update Account Recovery flow to use new KM data types from `master-password.types.ts` / `MasterPasswordService`:
- `MasterPasswordAuthenticationData`
- `MasterPasswordUnlockData`

This allows us to move away from the deprecated `makeMasterKey()` method (which takes email as salt) as we seek to eventually separate the email from the salt.

Changes are behind feature flag: `pm-27086-update-authentication-apis-for-input-password`

## 📸 Screenshots

https://github.com/user-attachments/assets/08822539-d919-46e0-89b9-71cee7a43396

[PM-27086]: https://bitwarden.atlassian.net/browse/PM-27086?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/web/src/app/admin-console/organizations/members/services/organization-user-reset-password/organization-user-reset-password.service.ts
```diff
@@ -140,6 +146,44 @@ export class OrganizationUserResetPasswordService implements UserKeyRotationKeyR
         ? new PBKDF2KdfConfig(response.kdfIterations)
         : new Argon2KdfConfig(response.kdfIterations, response.kdfMemory, response.kdfParallelism);
 
+    const newApisWithInputPasswordFlagEnabled = await this.configService.getFeatureFlag(
+      FeatureFlag.PM27086_UpdateAuthenticationApisForInputPassword,
+    );
+
+    if (newApisWithInputPasswordFlagEnabled) {
+      const salt: MasterPasswordSalt = this.masterPasswordService.emailToSalt(email);
```
quexten, line 154: Nice! Only note I have is that this will probably have to be changed with auth's other master-password-salt work, but that's out of scope of this PR in my opinion.

rr-bw: @Patrick-Pimentel-Bitwarden

I assume this line would change as part of our future work in the salt/email separation work, correct? I don't believe `masterPasswordService.saltForUser$()` helps us here as that checks account state, wheres this case (Account Recovery) involves setting the password for a completely separate user.

Patrick-Pimentel-Bitwarden: Yes that is correct that this is a part of future work and it has already been tracked. https://bitwarden.atlassian.net/browse/PM-30365 We will be using the salt in the response the query for the admin account recovery.

Issue #13401, opened 2025-02-14
ssh-agent - 'Too many authentication failures'


LexaNz:
### Steps To Reproduce

I use SSH Keys a lot, so I have many. By default BW client seems to present all keys of the active vault.
The problem is, most of ssh servers have a default retry value set to 5, so if the key you need to use is the 6th, you will reach the max retry value and the connection will failed.

1. add more than 5 ssh-keys (default sshd retry value)
2. Enable ssh-agent 
3. try to connect on a host using the 6th key -> 'Too many authentication failures'

### Expected Result

Ideally 2 things should be implemented. Add a tick if the key should be 'loaded/presented by default'
Add a button the 'load' or 'unload' the key in the agent - I know it's not how BW is working, because it substitutes to the ssh-agent.

Here the contextual menu of KeepassXC as example.

![Image](https://github.com/user-attachments/assets/e4ceca7e-9045-4871-9836-8d1b810bb5c7)

### Actual Result

ssh connection fails with message 'Too many authentication failures'

### Operating System

Linux

### Operating System Version

Ubuntu 24.04

### Installation method

Direct Download (from bitwarden.com)

### Build Version

Version 2025.2.0 SDK 'main (28c7e29)' Shell 34.0.0 Renderer 132.0.6834.83 Node 20.18.1 Architecture x64

### Issue Tracking Info

- [x] I understand that work is tracked outside of GitHub. A PR will be linked to this issue should one be opened to address it, but Bitwarden doesn't use fields like "assigned", "milestone", or "project" to track progress.

NovaSilentium:
Hi there, 

This is a known limitation and has been brought up in the Community forum here: **https://community.bitwarden.com/t/ssh-agent-allow-specifying-limiting-private-key-offers/79954**

Please sign up for the community forum and note for the feature request. 

This issue will now be closed. Note that the “Closed as not planned” status is for internal tracking purposes only, please do not hesitate to send us your suggestions to the forum.

Thanks!

LexaNz:
So I close as can't be used as ssh-agent...

vexcited:
I wrote a CLI to automatically configure SSH public keys in `.ssh/config` based on Bitwarden's custom fields and Bitwarden CLI.

With a single command you can fix this issue without doing the whole SSH configuration by yourself.
https://code.vexcited.com/nodes/seed.vexcited.com/rad:z3R6nHv6KaM8MpVuXPwbttyCcc4Zo

Hope it helps some of y'all

[...]

PR #18445, opened 2026-01-19
[PM-31000] Add SSH Agent key selection mode setting

InterN0te:
## Type of change
- [x] New feature development

## Objective
Add a new SSH Agent setting that allows users to choose between three key selection modes: sending all SSH keys, selecting a specific key for each request, or pre-loading specific keys to keep in memory.

## Code changes
**Desktop App:**
- `apps/desktop/src/autofill/models/ssh-agent-setting.ts`: Extended `SshAgentKeySelectionMode` const object with `AllKeys`, `SelectKey`, and `LoadSomeKeys` options
- `apps/desktop/src/autofill/services/ssh-agent.service.ts`: Implemented key selection logic for all three modes with proper lifecycle management and vault lock detection
- `apps/desktop/src/platform/components/select-ssh-key.ts`: Dialog component for single SSH key selection with search/filter functionality
- `apps/desktop/src/platform/components/select-ssh-key.html`: Dialog template displaying key names and fingerprints
- `apps/desktop/src/platform/components/load-some-keys.ts`: Multi-select dialog component for choosing multiple SSH keys to pre-load
- `apps/desktop/src/platform/components/load-some-keys.html`: Dialog template with checkboxes for multi-select functionality
- `apps/desktop/src/platform/services/desktop-settings.service.ts`: Added state management for SSH Agent key selection mode setting and selected key IDs storage
- `apps/desktop/src/app/accounts/settings.component.ts`: Integrated new setting into settings form with mode change handler
- `apps/desktop/src/app/accounts/settings.component.html`: Added UI controls for key selection mode
- `apps/desktop/src/locales/en/messages.json`: Added i18n strings for new UI elements

### Mode 1: "Send all SSH keys" (default)
- Existing behavior is preserved - all SSH keys are sent to the agent automatically
- Keys continuously synced to agent every second while vault is unlocked

### Mode 2: "Select key for each request"
1. User is prompted with a dialog to choose a specific SSH key when an SSH connection is requested
2. The dialog displays key names and fingerprints with search/filter functionality
3. Only the selected key is sent to the SSH agent for that specific connection
4. The key is automatically removed from the agent after use
5. User must re-select a key for each new connection (even to the same server)

### Mode 3: "Load some keys"
1. On first SSH connection attempt, user is prompted with a multi-select dialog
2. User selects one or multiple keys using checkboxes
3. Selected keys are loaded into the SSH agent and kept in memory
4. All subsequent SSH connections use the pre-loaded keys without prompts
5. Selection is automatically cleared when:
   - Vault is locked (allows re-selection on next unlock)
   - User switches to a different mode
6. Keys are continuously synced to agent every second (similar to AllKeys mode, but only selected keys)

## Security considerations
- Keys are cleared after each use in SelectKey mode to minimize exposure
- LoadSomeKeys mode provides granular control without exposing all keys
- User has explicit control over which keys are available to the agent
- Follows principle of least privilege by only exposing selected keys
- Selection automatically cleared on vault lock for security
- Per-user storage ensures different accounts maintain separate selections

## Implementation notes
- Follows Bitwarden architectural patterns:
  - Uses frozen const object instead of enum (ADR-0025)
  - Observable data services pattern (ADR-0003)
  - Proper RxJS subscription management with `concatMap` and `takeUntil`
- Consistent with existing SSH Agent settings structure
- All Tailwind CSS classes use required `tw-` prefix
- Vault lock detection via `authService.activeAccountStatus$`
- UserKeyDefinition ensures per-account storage with automatic switching

## Screenshots
<details><summary>UI changes</summary>
<p>

1. New setting in the Desktop settings page with three options:
image
<br>
2. SSH key selection dialog (SelectKey mode - single selection):
image
<br>
3. Multi-key selection dialog (LoadSomeKeys mode):
image

<br>

</p>
</details> 

## Testing requirements
- [x] Verify "Send all SSH keys" mode works as before (default behavior unchanged)
- [x] Verify "Select key for each request" mode shows dialog on SSH connection
- [x] Verify search/filter functionality in key selection dialog
- [x] Verify selected key is cleared after use and re-requested on next connection
- [x] Verify "Load some keys" mode shows multi-select dialog on first connection
- [x] Verify selected keys persist across SSH connections without prompts
- [x] Verify selected keys are continuously synced while vault is unlocked
- [x] Verify selection is cleared when vault locks
- [x] Verify selection is cleared when switching away from LoadSomeKeys mode
- [x] Verify submit button is disabled when no keys are selected
- [x] Verify behavior when vault is locked during SSH request
- [x] Verify behavior when no SSH keys exist in vault
- [x] Verify account switching maintains separate selections per account

### DIFF
--- a/apps/desktop/src/platform/components/select-ssh-key.ts
+++ b/apps/desktop/src/platform/components/select-ssh-key.ts
@@ -1,0 +1,113 @@
+import { CommonModule } from "@angular/common";
+import { Component, Inject } from "@angular/core";
+import { takeUntilDestroyed } from "@angular/core/rxjs-interop";
+import { FormBuilder, ReactiveFormsModule } from "@angular/forms";
+
+import { JslibModule } from "@bitwarden/angular/jslib.module";
+import { CipherView } from "@bitwarden/common/vault/models/view/cipher.view";
+import {
+  DIALOG_DATA,
+  DialogRef,
+  AsyncActionsModule,
+  ButtonModule,
+  DialogModule,
+  FormFieldModule,
+  IconButtonModule,
+  DialogService,
+  CalloutModule,
+  RadioButtonModule,
+  SearchModule,
+} from "@bitwarden/components";
+
+export interface SelectSshKeyParams {
+  sshKeys: CipherView[];
+  applicationName: string;
+  namespace: string;
+}
+
+// FIXME(https://bitwarden.atlassian.net/browse/CL-764): Migrate to OnPush
+// eslint-disable-next-line @angular-eslint/prefer-on-push-component-change-detection
+@Component({
+  selector: "app-select-ssh-key",
+  templateUrl: "select-ssh-key.html",
+  imports: [
+    DialogModule,
+    CommonModule,
+    JslibModule,
+    ButtonModule,
+    IconButtonModule,
+    ReactiveFormsModule,
+    AsyncActionsModule,
+    FormFieldModule,
+    CalloutModule,
+    RadioButtonModule,
+    SearchModule,
+  ],
+})
+export class SelectSshKeyComponent {
+  selectSshKeyForm = this.formBuilder.group({
+    selectedKeyId: [null as string | null],
+    searchText: [""],
+  });
+
+  filteredSshKeys: CipherView[] = [];
+
+  constructor(
+    @Inject(DIALOG_DATA) protected params: SelectSshKeyParams,
+    private dialogRef: DialogRef<string | null>,
+    private formBuilder: FormBuilder,
+  ) {
+    this.filteredSshKeys = params.sshKeys || [];
+
+    // Subscribe to search text changes
+    this.selectSshKeyForm.controls.searchText.valueChanges
+      .pipe(takeUntilDestroyed())
+      .subscribe((searchText) => {
+        this.filterSshKeys(searchText || "");
+      });
+  }
+
+  private filterSshKeys(searchText: string): void {
+    if (!searchText) {
+      this.filteredSshKeys = this.params.sshKeys;
+      return;
+    }
+
+    const lowerSearchText = searchText.toLowerCase();
+    this.filteredSshKeys = this.params.sshKeys.filter((key) => {
+      return (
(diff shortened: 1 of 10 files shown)


InterN0te:
This PR is for #13401 : https://community.bitwarden.com/t/ssh-agent-allow-specifying-limiting-private-key-offers/79954

apps/desktop/src/platform/components/select-ssh-key.ts
```diff
@@ -0,0 +1,113 @@
...
+    IconButtonModule,
+    ReactiveFormsModule,
+    AsyncActionsModule,
+    FormFieldModule,
+    CalloutModule,
+    RadioButtonModule,
+    SearchModule,
+  ],
+})
+export class SelectSshKeyComponent {
```
quexten, line 47: ~~I have not yet debugged this, but I have a question on the logic behind this mode. At the time that the client requests a signature from the agent, it already specifies a public key for which the signature should be created. https://www.ietf.org/archive/id/draft-miller-ssh-agent-11.html#name-private-key-operations~~

~~So, at the time of signing, you cannot decide the key. It is invalid by the protocol. What is the expected outcome here? Does it sign with a key that was not requested?~~

~~(Maybe I'm misunderstanding something)~~

For anyone else looking, the list request is filtered, not the sign request.

InterN0te: Thanks for taking a look !
Just to confirm - does everything look good to you now, or do you have any other questions/concerns about the implementation ?

quexten: At an initial look, the changes seemed mostly reasonable from the engineering side. There is currently internal product discussion on the intended UX around this feature.

dwbit:
Hi @InterN0te ,

Thanks for your efforts on this one! After further review, we need to flesh out the UX/design more before implementing any code.

If you would still like to continue working on it, please open a [GitHub Discussion](https://github.com/orgs/bitwarden/discussions/new?category=password-manager) so we can collaborate further.

neuronull:
Circling back to link to note that we have a community forums post up now that addresses the v2 and contributions 🙏 

PR #18449, opened 2026-01-19
[PM-29816] - fix scroll position in browser vault

jaasen-livefront:
## 🎟️ Tracking
https://bitwarden.atlassian.net/browse/PM-29816

## 📔 Objective

Since `virtualScrollElement ` is no longer available to the vault-v2.component we instead need to capture scroll events from a regular DOM scroll element in the popup-page.component. The consuming components and scroll position service have been adjusted accordingly.

## 📸 Screenshots

https://github.com/user-attachments/assets/858ce4f3-cd7c-4cb8-8342-adf72b5dff8d

### DIFF
--- a/apps/browser/src/platform/popup/layout/popup-page.component.html
+++ b/apps/browser/src/platform/popup/layout/popup-page.component.html
@@ -23,10 +23,8 @@
     640px in width (equivalent to tailwind's `sm` breakpoint)
   -->
   <div
-    #scrollRegion
     class="tw-size-full tw-styled-scrollbar"
     data-testid="popup-layout-scroll-region"
-    (scroll)="handleScroll($event)"
     [ngClass]="{
       '!tw-overflow-hidden': hideOverflow(),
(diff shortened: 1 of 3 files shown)


apps/browser/src/platform/popup/layout/popup-page.component.html
```diff
@@ -23,6 +23,7 @@
     640px in width (equivalent to tailwind's `sm` breakpoint)
   -->
   <div
+    #scrollRegion
```
nick-livefront: ❓ Could you expand reference the `bitScrollLayoutHost` directive that is built by the UIF team? They're already storing the scrollable element internally. At the very least this would avoid having two separate ways of referencing the scrollable element.

jaasen-livefront: @nick-livefront Nice catch! Didn't even notice that 🤦 . Fixed https://github.com/bitwarden/clients/pull/18449/commits/c8193f0ec862c05623d492e018434492458f6ca3

PR #18451, opened 2026-01-20
[BEEEP] Add util functions for uint8 array conversion

quexten:
## 📔 Objective

Most incompatibility with typescript 7, 5.9 / tsgo is around Uint8Array. In TS7, Uint8Array is not the same as ArrayBuffer / ArrayBufferLike. To fix the incompatibilities, this PR adds Util functions that take Uint8Arrays, and all callers can me migrated in subsequent PRs, piece-by-piece.

Please see https://github.com/bitwarden/clients/pull/18441 for example usage.


libs/common/src/platform/misc/utils.ts
```diff
@@ -129,6 +129,70 @@ export class Utils {
...
+   * Converts a Uint8Array to a UTF-8 decoded string.
+   * @param arr - The Uint8Array containing UTF-8 encoded bytes.
+   * @returns The decoded UTF-8 string, or null if the input is null.
+   */
+  static fromArrayToUtf8(arr: Uint8Array | null): string | null {
+    if (arr == null) {
+      return null;
+    }
+
+    return this.fromBufferToUtf8(arr.buffer as ArrayBuffer);
```
djsmith85: ❓  Not liking the type assignments. Is this only there because we currently use TS and these can be removed when switched over to TS-go?

quexten: The logic here is supposed to re-use existing logic as closely as possible. I think it's entirely reasonable to just drop using the old functions and work on Uint8Array directly, but I did not want to risk that here.

Is it reasonable to just track this as tech debt?

quexten: @djsmith85 I replaced all calls now with a polyfill for https://github.com/zloirock/core-js?tab=readme-ov-file#uint8array-to--from-base64-and-hex (e.g. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Uint8Array/toBase64).

This will also allow us to drop our custom base64 and hex parsing, and gets rid of the type casts.

libs/common/src/platform/misc/utils.ts
```diff
@@ -129,6 +130,78 @@ export class Utils {
...
+   * Converts a Uint8Array to a hexadecimal string.
+   * @param arr - The Uint8Array to convert.
+   * @returns The hexadecimal string representation, or null if the input is null.
+   */
+  static fromArrayToHex(arr: Uint8Array | null): string | null {
+    if (arr == null) {
+      return null;
+    }
+
+    // @ts-expect-error - polyfilled by core-js
```
quexten, line 144: Need to ignore these since they are polyfilled at runtime.

djsmith85 (approved):

PR #18452, opened 2026-01-20
[POC] Set User-key state from sdk

quexten:
## 🎟️ Tracking

https://github.com/bitwarden/sdk-internal/pull/702

### DIFF
--- a/libs/angular/src/services/jslib-services.module.ts
+++ b/libs/angular/src/services/jslib-services.module.ts
@@ -362,6 +362,7 @@ import {
   UserAsymmetricKeysRegenerationApiService,
   UserAsymmetricKeysRegenerationService,
 } from "@bitwarden/key-management";
+import { DefaultUnlockService, UnlockService } from "@bitwarden/unlock";
 import {
   ActiveUserStateProvider,
   DerivedStateProvider,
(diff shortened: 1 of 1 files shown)


libs/key-management-ui/src/lock/components/lock.component.ts
```diff
@@ -483,7 +491,35 @@ export class LockComponent implements OnInit, OnDestroy {
...
+              userId: this.activeAccount.id,
+              kdfParams: {
+                pBKDF2: { iterations: 100000 }
+              },
+              email: "test@quexten.com",
+              accountCryptographicState: await firstValueFrom(this.accountCryptographicStateService.accountCryptographicState$(this.activeAccount.id)),
+              method: {
+                pinEnvelope: {
+                  pin: pin,
+                  pin_protected_user_key_envelope: await this.pinStateService.getPinProtectedUserKeyEnvelope(
```
mzieniukbw: Ideally, this state should be available in sdk and have more high level abstraction, other than `initialize_user_crypto`, where we don't have to pass in some of that state. Things like `accountCryptographicState` or `pin_protected_user_key_envelope` state could be pulled in from SDK

quexten: 100% agree, this should be moved out to a separate "unlockService". Noting this is just PoC to see roughly what's necessary, and is not meant for merging.

quexten: Added a demo unlock service

libs/unlock/src/default-unlock.service.ts
```diff
@@ -0,0 +1,201 @@
...
+        );
+    }
+
+    private async getMasterPasswordUnlockData(userId: UserId): Promise<MasterPasswordUnlockData | null> {
+        const unlockData = await firstValueFrom(this.masterPasswordService.masterPasswordUnlockData$(userId));
+        return unlockData.toSdk();
+    }
+
+    async unlockWithDeviceKey(userId: UserId,
+        encryptedDevicePrivateKey: EncString,
```
quexten, line 67: Need to use state here

PR #18453, opened 2026-01-20
[PM-31029] Remove padding on new-vault (desktop)

Hinton:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-31029
## 📔 Objective

We need to remove padding on certain pages (i.e. the vault page) but keeping it in others. This adds a temporary conditional `noPadding` route param that handles this.


Hinton:
Handled by https://github.com/bitwarden/clients/pull/18458

PR #18454, opened 2026-01-20
[PM-30891] - Create My Items On Restore

jrmccannon:
## 🎟️ Tracking
[PM-30891](https://bitwarden.atlassian.net/browse/PM-30891)

## 📔 Objective
When restoring the user, this will call the new endpoint (single user) and provide the encrypted default collection name for my items to be created if needed server side.

This is linked a [server PR](https://github.com/bitwarden/server/pull/6879) as well.

## 📸 Screenshots

https://github.com/user-attachments/assets/313d48d1-32c0-4c25-9343-8c54a6c3e850

[PM-30891]: https://bitwarden.atlassian.net/browse/PM-30891?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/web/src/app/admin-console/organizations/members/components/bulk/bulk-restore-revoke.component.ts
+++ b/apps/web/src/app/admin-console/organizations/members/components/bulk/bulk-restore-revoke.component.ts
@@ -1,7 +1,7 @@
 // FIXME: Update this file to be type safe and remove this and next line
 // @ts-strict-ignore
 import { Component, Inject } from "@angular/core";
-import { combineLatest, firstValueFrom, Observable, switchMap } from "rxjs";
+import { combineLatest, firstValueFrom, map, Observable, switchMap } from "rxjs";
 
 import {
   OrganizationUserApiService,
@@ -65,14 +65,14 @@ export class BulkRestoreRevokeComponent {
 
     this.organization$ = accountService.activeAccount$.pipe(
       getUserId,
-      switchMap((userId) => {
-        const organization = organizationService.organizations$(userId);
+      switchMap((userId) => organizationService.organizations$(userId)),
+      getById(this.organizationId),
+      map((organization) => {
         if (organization == null) {
           throw new Error("Organization not found");
         }
         return organization;
       }),
-      getById(this.organizationId),
     );
   }
 
--- a/apps/web/src/app/admin-console/organizations/members/components/member-dialog/member-dialog.component.ts
+++ b/apps/web/src/app/admin-console/organizations/members/components/member-dialog/member-dialog.component.ts
@@ -4,7 +4,6 @@ import { Component, Inject, OnDestroy } from "@angular/core";
 import { FormBuilder, Validators } from "@angular/forms";
 import {
   combineLatest,
-  filter,
   firstValueFrom,
   map,
   Observable,
@@ -204,7 +203,12 @@ export class MemberDialogComponent implements OnDestroy {
       getUserId,
       switchMap((userId) => organizationService.organizations$(userId)),
       getById(this.params.organizationId),
-      filter((organization) => organization != null),
+      map((organization) => {
+        if (organization == null) {
+          throw new Error("Organization not found");
+        }
+        return organization;
+      }),
       shareReplay({ refCount: true, bufferSize: 1 }),
     );
 


apps/web/src/app/admin-console/organizations/members/components/bulk/bulk-restore-revoke.component.ts
```diff
@@ -46,6 +61,15 @@ export class BulkRestoreRevokeComponent {
...
     );
+
+    this.organization$ = accountService.activeAccount$.pipe(
+      switchMap((account) =>
+        organizationService.organizations$(account?.id).pipe(
+          getById(this.organizationId),
+          filter((org): org is Organization => org != null),
+        ),
+      ),
+    );
```
BTreston, line 76: Be aware that rxjs `filter` will silently drop emissions that don't pass the filter. So if `organizationService.organizations$` only emits `undefined` for whatever reason, this will cause runtime errors if we're not careful. I tend to prefer an imperative check and throw rather than `filter` when using the value as a parameter to some API call (as is done here) and makes debugging easier. E.g.

```suggestion
// Since in this file, we're just using organization as a parameter in this api call, we can inline it into the function
// rather than have it as a class member.
 const result = this.accountService.activeAccount$.pipe(
    getUserId,
    switchMap(userId => this.organizationService.organizations$(userId)),
    getById(this.organizationId),
    switchMap(organization => {
      if (organization == null) {
        throw new Error(`Organization ${this.organizationId} not found`);
      }
      return this.apiService.callSomething(organization, ...);
    })
  );
```
This same bit of wisdom applies to the other instances where we use `filter`, though this is ultimately your choice to make. `filter` is much better IMO when it is used to drive reactive UI elements.

apps/web/src/app/admin-console/organizations/members/components/bulk/bulk-restore-revoke.component.ts
```diff
@@ -83,9 +107,24 @@ export class BulkRestoreRevokeComponent {
...
+          switchMap(([enabled, organization]) => {
+            if (enabled) {
+              return this.organizationUserService.bulkRestoreUsers(organization, userIds);
+            } else {
+              return from(
+                this.organizationUserApiService.restoreManyOrganizationUsers(
+                  this.organizationId,
+                  userIds,
+                ),
+              );
```
BTreston: ⛏️ don't need `from` here.

apps/web/src/app/admin-console/organizations/members/components/member-dialog/member-dialog.component.ts
```diff
@@ -197,13 +199,15 @@ export class MemberDialogComponent implements OnDestroy {
...
       switchMap((account) =>
-        organizationService
-          .organizations$(account?.id)
-          .pipe(getOrganizationById(this.params.organizationId))
-          .pipe(shareReplay({ refCount: true, bufferSize: 1 })),
+        organizationService.organizations$(account?.id).pipe(
+          getById(this.params.organizationId),
+          filter((organization) => organization != null),
+          shareReplay({ refCount: true, bufferSize: 1 }),
+        ),
```
BTreston:
```suggestion
    this.organization$ = accountService.activeAccount$.pipe(
      getUserId,
      switchMap(userId => organizationService.organizations$(userId)),
      getById(this.params.organizationId),
      filter((organization) => organization != null),
      shareReplay({ refCount: true, bufferSize: 1 }),
    );
```

apps/web/src/app/admin-console/organizations/members/components/member-dialog/member-dialog.component.ts
```diff
@@ -633,9 +637,28 @@ export class MemberDialogComponent implements OnDestroy {
...
+              organization,
+              params.organizationUserId,
+            );
+          } else {
+            return from(
+              this.organizationUserApiService.restoreOrganizationUser(
+                params.organizationId,
+                params.organizationUserId,
+              ),
+            );
```
BTreston: don't need `from` here

apps/web/src/app/admin-console/organizations/members/services/member-actions/member-actions.service.ts
```diff
@@ -116,7 +119,20 @@ export class MemberActionsService {
...
-      await this.organizationUserApiService.restoreOrganizationUser(organization.id, userId);
+      await firstValueFrom(
+        this.configService.getFeatureFlag$(FeatureFlag.DefaultUserCollectionRestore).pipe(
+          switchMap((enabled) => {
+            if (enabled) {
+              return this.organizationUserService.restoreUser(organization, userId);
+            } else {
+              return from(
+                this.organizationUserApiService.restoreOrganizationUser(organization.id, userId),
+              );
```
BTreston: ⛏️ don't need `from` here

libs/admin-console/src/common/organization-user/services/default-organization-user.service.ts
BTreston, line 1: These changes are good, just a note that some of the ongoing members work I am planning will probably get rid of this service in favor of something more akin to our command pattern on the server, where member actions will have their own services rather than multiple disparate organization user services as we have now. No change to make here now though.

BTreston (changes requested):
This all looks good, just a few rxjs nit ⛏️ s. Nicely done

PR #18459, opened 2026-01-20
Downgrade open to 8.4.2

abergs:
## 📔 Objective

This PR downgrades open to 8.4.2 to avoid issues with ESM. My motivations were to enable us migrating towards pnpm, but I've understood that open 11 caused other issues too (cc: @Hinton @coroiu)

### DIFF
--- a/apps/cli/package.json
+++ b/apps/cli/package.json
@@ -83,7 +83,7 @@
     "multer": "2.0.2",
     "node-fetch": "2.6.12",
     "node-forge": "1.3.2",
-    "open": "11.0.0",
+    "open": "8.4.2",
     "papaparse": "5.5.3",
     "proper-lockfile": "4.1.2",
     "rxjs": "7.8.1",
--- a/package-lock.json
+++ b/package-lock.json
@@ -55,7 +55,7 @@
         "node-fetch": "2.6.12",
         "node-forge": "1.3.2",
         "oidc-client-ts": "2.4.1",
-        "open": "11.0.0",
+        "open": "8.4.2",
         "papaparse": "5.5.3",
         "proper-lockfile": "4.1.2",
         "qrcode-parser": "2.1.3",
@@ -219,7 +219,7 @@
         "multer": "2.0.2",
         "node-fetch": "2.6.12",
         "node-forge": "1.3.2",
-        "open": "11.0.0",
+        "open": "8.4.2",
         "papaparse": "5.5.3",
         "proper-lockfile": "4.1.2",
         "rxjs": "7.8.1",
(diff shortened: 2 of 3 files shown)


coltonhurst (commented):
Do we need to keep the changes @coroiu added here? Or were those part of the issue?

https://github.com/bitwarden/clients/pull/14932/files#diff-3607f07c9391eb32d04376c272322302af6aeaee3496e89816d717c2a9dd5059

(The changes in `apps/cli/src/platform/services/cli-platform-utils.service.ts`)

abergs:
@coltonhurst Yup, pushed too soon.

edit: turns out we didn't and I didn't mess up the local compile, surprise!

abergs:
@coroiu Did you encounter any problems with using `import` syntax that sparked this: https://github.com/bitwarden/clients/pull/14932/changes#diff-3607f07c9391eb32d04376c272322302af6aeaee3496e89816d717c2a9dd5059L5

Seems to work fine for me.

coroiu:
@abergs I didn't write that code so I don't know exactly what triggered the change :)

abergs:
@Hinton How did you discover the need to tweak zbus etc? It seems CI is running OK on this PR, how can we verify the issues you seemed to encounter when [downgrading last time](https://github.com/bitwarden/clients/pull/14932#issuecomment-3601650657)?

>Pinned is-generator-function due to esm issues.
>zbus_macros is a transient dependency and got bumped while zbus did not, this caused a missmatch and to resolve I manually bumped zbus. @quexten We might want to look into this the macro should generally be in lockstep.

Hinton:
@abergs did you do a lockfile refresh of Cargo.lock? That's when the zbus issue popped up.

I just reverted the whole open bump PR. The import syntax might be fine but you will want to test both the packaged builds but also non packaged to make sure we still open browsers.

abergs:
@Hinton Thanks, gotcha will do.

abergs:
I was able to download the [CLI](https://github.com/bitwarden/clients/actions/runs/21181381558/job/60924357984?pr=18459) and run `bw login --sso`. It opened a browser.

I was also able to download the mac app and run it. 

@coltonhurst Would you be able to re-test whatever problem you guys ran into when we reverted the open fix. Also, do we need to do anything to pin open to this version?

coltonhurst:
> I was able to download the [CLI](https://github.com/bitwarden/clients/actions/runs/21181381558/job/60924357984?pr=18459) and run `bw login --sso`. It opened a browser.
> 
> I was also able to download the mac app and run it.
> 
> @coltonhurst Would you be able to re-test whatever problem you guys ran into when we reverted the open fix. Also, do we need to do anything to pin open to this version?

@abergs Just tested locally and I don't run into the same issue. IIRC all we did previously was do a fresh electron build on macOS:

```
npm ci // at root
npm run build // within napi to build the native modules
npm run electron // within desktop to build and run the app locally
```

If we don't want open to upgrade then pinning might be a good route, that's probably up to @Hinton / @coroiu as it's a platform dep.

abergs:
Alright I need an approval for this to go in @Hinton @coroiu

coroiu (approved):

PR #18472, opened 2026-01-21
[PM-31273] Create framework for services to run on application initialization

coroiu:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-31273

## 📔 Objective

Make app initialization more decentralized and with better dependency tracking. Teams can now add their own dependencies without having to manually inject their service into the `async init()` function. Using the old init services will be deprecated

This PR has been reviewed locally by both me and Claude before publishing the PR

### DIFF
--- a/libs/angular/src/jslib.module.ts
+++ b/libs/angular/src/jslib.module.ts
@@ -2,7 +2,7 @@ import { CommonModule, DatePipe } from "@angular/common";
 import { NgModule } from "@angular/core";
 import { FormsModule, ReactiveFormsModule } from "@angular/forms";
 
-import { Dependency } from "@bitwarden/common/platform/abstractions/injector";
+import { AsyncDependency } from "@bitwarden/common/platform/abstractions/async-initializable";
 import { DefaultAsyncInitService } from "@bitwarden/common/platform/services/default-async-init.service";
 import {
   AsyncActionsModule,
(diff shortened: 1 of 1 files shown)


justindbaur:
@coroiu That way of handling dependencies is great, much less likely to break than a simple number. 

Two main notes though:
- You'll need some implementation of this service that can be used in browser background as well, it's arguably the most important place for this to work well. It's certainly possible to implement I just never _loved_ how it worked without us going full dependency injection for it.
- Also for browser (it's the worst lol) we really want to have a synchronous version for initialization. The biggest thing I expect for teams to do in these things in browser, especially autofill, is to subscribe to messages or data updates. The subscriptions are supposed to be done synchronously or there is a danger they don't get there in time before the service worker fires the event after startup. So far I don't think this has been a problem but I worry that it might be one day, or could just be a source of flakiness. My thinking is we can either have both a sync and async version and the sync one always gets to run to completion first, or more opinionatedly, only a sync version. Async would still be manually placed in code if you truly need it but I kinda think most uses don't truly need it.

apps/browser/src/background/main.background.ts
dereknance: With `CommandsBackground` being registered as `SyncInitializable` [above](https://github.com/bitwarden/clients/pull/18472/changes#diff-4b8a4df786a048d7e74dae64191d365c17cb40f625a545a7408ceb5ce5f1d7f9R1590), did you intend to remove this line?

coroiu: Good catch, fixed in [2b6379d](https://github.com/bitwarden/clients/pull/18472/commits/2b6379dfe4e91a34ab7aa9fb4d17c87958ea5598)

willmartian (changes requested):
Could you add (or point me to) usage documentation?

PR #17982, opened 2025-12-15
Experiment: Use pnpm in CI

abergs:
## 📔 Objective

This PR is an experiment with the goal to increase performance in our dependency installation.
It uses `pnpm` over `npm`. pnpm is faster both locally and likely (as we'll see) on the CI, even when not using a cache.

`pnpm` also brings some [additional security](https://pnpm.io/supply-chain-security) over `npm` as it doesn't run postinstall scripts by default. In this PR I've whitelisted a bunch of them.

In my manual test runs, I was able to get Install Dependencies **down to 6s** _(from 40s+ using npm)_. Not all runs are that fast, see the benchmarks below. Also that number is a bit misguided, because it relies on using the cache. The cache is (at least in current configuration) relatively slow and shows in the `Setup Node` where it's downloading the cache, but we still see a good net gain,  e.g. **18s vs 45s+ (60% speed up)** in the screenshot below.

I have some other experimental PRs to further speed things up. I wouldn't look to much at them as they are very early, but for the curious I'll list them:
* https://github.com/bitwarden/clients/pull/17989 (early research)
* https://github.com/bitwarden/clients/pull/17988 (early research)
* https://github.com/bitwarden/clients/pull/17984 (early research)
* https://github.com/bitwarden/clients/pull/17944 (early research)
* https://github.com/bitwarden/clients/pull/17970 (early research)

image

CleanShot 2025-12-16 at 00 49 43@2x

## Summary

| Metric | npm (main) | pnpm (PR) | Savings |
|--------|-----------|-----------|---------|
| Setup + Install (per job avg) | ~52s | ~34s | **~18s (35%)** |
| Install only (per job avg) | ~50s | ~32s | **~18s (36%)** |
| Billable minutes (all workflows, ~31 jobs) | ~27 min | ~18 min | **~9 min** |

### Setup + Install Step Breakdown

The full setup includes "Set up Node" and dependency installation. pnpm adds a "Set up pnpm" step (~1s) but saves significantly on install time.

**Cache Status**: Both benchmarks ran with **cold caches** (cache miss):
- PR (pnpm): `pnpm cache is not found`
- Main (npm): `npm cache is not found`

This makes the comparison fair and representative of worst-case (first run) scenarios.

#### Build Browser - Chrome Commercial ([main](https://github.com/bitwarden/clients/actions/runs/20334450169/job/58417484523) vs [PR](https://github.com/bitwarden/clients/actions/runs/20272558716/job/58212205368))

[...]

PR #18474, opened 2026-01-21
Upgrade electron-builder to 26.6.0 (latest)

abergs:
## 📔 Objective

This PR updates electron-builder to 26.6.0. The reason for upgrading is to unblock another PR: https://github.com/bitwarden/clients/pull/17982/changes#r2711707144

**What do we normally test when we're bumping electron builder?**
I've verified the dmg build.

### DIFF
--- a/apps/desktop/electron-builder.json
+++ b/apps/desktop/electron-builder.json
@@ -144,18 +144,6 @@
         "y": 180,
         "type": "link",
         "path": "/Applications"
-      },
-      {
-        "x": 9999,
-        "y": 9999,
-        "type": "file",
-        "path": ".background.png"
-      },
-      {
-        "x": 9999,
-        "y": 9999,
-        "type": "file",
-        "path": ".VolumeIcon.icns"
       }
     ],
     "window": {


apps/desktop/electron-builder.json
```diff
@@ -72,6 +72,7 @@
     },
     "provisioningProfile": "bitwarden_desktop_developer_id.provisionprofile",
     "singleArchFiles": "node_modules/@bitwarden/desktop-napi/desktop_napi.darwin-*.node",
+    "x64ArchFiles": "**/bufferutil/**/darwin-*/*.node",
```
abergs, line 75: Turns out electron builder now has a more detailed check for duplicates of binaries in universal builds.

apps/desktop/electron-builder.json
```diff
@@ -151,6 +152,7 @@
     }
   },
   "mas": {
+    "identity": "Bitwarden Inc",
```
abergs, line 155: Without providing the identity signing the .pkg file fails. Most likely because of upstream bug in resolving identity.

vleague2:

wrobelda:
Please note that updating electron-builder to latest should also enable building aarch64 versions of DEB and RPM packages:
https://github.com/electron-userland/electron-builder/issues/9563

The current setup only builds a tar.gz package.

Also see comments under this commit for reference: https://github.com/signalapp/Signal-Desktop/commit/560dcb91d98ba1177162372a2837919225e3e2c9

PR #18475, opened 2026-01-21
[PM-30908]Correct Premium subscription status handling

cyprain-okeke:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-30908

## 📔 Objective

Add resubscription flow for users with expired or canceled premium subscriptions, and improve subscription status handling in the web vault.

  Changes

  API Client (account-billing.client.ts:26-38)

  - Handle 404 response from subscription endpoint when user has no subscription
  - Return null instead of throwing error for 404 status
  - Change return type from BitwardenSubscription to Maybe<BitwardenSubscription>

  Subscription Page (account-subscription.component.ts:82-233)

  - Check for subscription existence using API call instead of profile state
  - Redirect to premium page when subscription is null
  - Reload subscription data after successful resubscribe

  Subscription Card Component (subscription-card.component.ts:157-232)

  - Add Resubscribe action constant
  - Update incomplete_expired status callout to show "Resubscribe" button instead of "Contact Support"
  - Add callout for canceled status with "Resubscribe" button
  - Change callout messages to use "is expired/canceled" tense

  Localization (messages.json:3266-3273)

  - Add resubscribe translation key
  - Add yourSubscriptionIsExpired translation key
  - Add yourSubscriptionIsCanceled translation key

## 📸 Screenshots

Uploading Screen Recording 2026-01-22 at 9.34.49 AM.mov…

https://github.com/user-attachments/assets/14e0ef86-3813-49ef-83b6-a19967ffd69e

https://github.com/user-attachments/assets/22f84527-ae25-4b0e-876d-c40c049de8a5

https://github.com/user-attachments/assets/d891869a-77a3-4b9a-b6fb-f161dde43b76

### DIFF
--- a/apps/web/src/app/billing/individual/subscription/account-subscription.component.ts
+++ b/apps/web/src/app/billing/individual/subscription/account-subscription.component.ts
@@ -238,7 +238,7 @@ export class AccountSubscriptionComponent {
         await this.router.navigate(["../payment-details"], { relativeTo: this.activatedRoute });
         break;
       case SubscriptionCardActions.Resubscribe: {
-        const account = this.activeAccount();
+        const account = this.account();
         if (!account) {
           return;
         }


apps/web/src/app/billing/individual/subscription/account-subscription.component.ts
```diff
@@ -208,6 +212,39 @@ export class AccountSubscriptionComponent {
       case SubscriptionCardActions.UpdatePayment:
         await this.router.navigate(["../payment-details"], { relativeTo: this.activatedRoute });
         break;
+      case SubscriptionCardActions.Resubscribe: {
+        const account = await firstValueFrom(this.accountService.activeAccount$);
```
amorask-bitwarden: ⛏️ This is used in 2 locations - let's move this to a signal.

apps/web/src/app/billing/individual/subscription/account-subscription.component.ts
```diff
@@ -208,6 +212,39 @@ export class AccountSubscriptionComponent {
...
+        if (!account) {
+          return;
+        }
+
+        const currentSubscription = this.subscription.value();
+        const isIncompleteExpired =
+          currentSubscription?.status === SubscriptionStatuses.IncompleteExpired;
+
+        const dialogRef = UnifiedUpgradeDialogComponent.open(this.dialogService, {
```
amorask-bitwarden, line 246: ❓ I'm not sure this is correct. Wouldn't we always want to force a premium re-subscribe action into purchasing another premium subscription?

apps/web/src/locales/en/messages.json
```diff
@@ -3263,6 +3263,15 @@
...
   },
+  "resubscribe": {
+    "message": "Resubscribe"
+  },
+  "yourSubscriptionIsExpired": {
+    "message": "Your subscription is expired. Please resubscribe to continue using premium features."
+  },
+  "yourSubscriptionIsCanceled": {
+    "message": "Your subscription is canceled. Please resubscribe to continue using premium features."
+  },
```
amorask-bitwarden, line 3349: ❓ This wasn't the copy in the original ticket. Was it requested by design?

apps/web/src/app/billing/individual/subscription/account-subscription.component.ts
amorask-bitwarden: This has not been addressed.

apps/web/src/app/billing/individual/subscription/account-subscription.component.ts
```diff
@@ -208,6 +214,28 @@ export class AccountSubscriptionComponent {
...
+        if (!account) {
+          return;
+        }
+
+        const dialogRef = UnifiedUpgradeDialogComponent.open(this.dialogService, {
+          data: {
+            account,
+            initialStep: UnifiedUpgradeDialogStep.Payment,
+            selectedPlan: PersonalSubscriptionPricingTierIds.Premium,
+            redirectOnCompletion: true,
```
amorask-bitwarden: I don't think we need to redirect on completion if we're reloading the new subscription page on line 235.

apps/web/src/app/billing/individual/subscription/account-subscription.component.ts
```diff
@@ -206,14 +212,60 @@ export class AccountSubscriptionComponent {
...
+      },
+    });
+
+    const result = await lastValueFrom(dialogRef.closed);
+
+    if (result?.type === "success") {
+      // Wait for Stripe webhook to process payment and update subscription status on backend
+      setTimeout(() => {
+        this.subscription.reload();
+      }, 7000);
```
amorask-bitwarden: ❌ This constitutes 7 seconds where our dialog is just spinning while the user waits, which is a very poor user experience.

@sukhleenb With regards to having the "Update payment" button open the dialog vs. navigate to the payment details page, do you consider that high priority with regards to this ticket or is it something we can address in a separate issue?

The reason I ask is that when the customer updates their payment method for an `incomplete` subscription, Stripe sends us a webhook which we process on the server and results in the subscription being moved to `active`.

However, there's no real way to determine how long that process takes and I'm not comfortable setting it at an arbitrary length of time that negatively impacts our users. With the approach of sending them to the payment details page, that should give us enough buffer while the webhook processing takes place since they have to navigate back.

The correct approach to this is, in my opinion, to display a banner that says something along the lines of "We're currently processing your payment and will update your subscription soon" and then having the backend notify the web application when the processing is complete, resulting in the page refreshing, but this would take substantially more work than is outlined here.

amorask-bitwarden (changes requested):

apps/web/src/app/billing/individual/subscription/account-subscription.component.ts
sukhleenb: Given the increased scope, I'm comfortable having the "Upgrade payment" button opening a dialog be a follow on. Agreed that we should provide additional context around what is happening and what the user can expect. If you have time, could you spin up a ticket documenting the technical reasons for splitting this work out, and I can work with Micah on when to pull it in?

sukhleenb (approved):

apps/web/src/app/billing/individual/user-subscription.component.ts
```diff
@@ -79,13 +79,10 @@ export class UserSubscriptionComponent implements OnInit {
       return;
     }
 
-    const userId = await firstValueFrom(this.accountService.activeAccount$);
```
amorask-bitwarden, line 82: ❓ Why are we making changes to the previous iteration of the subscription page?

apps/web/src/app/billing/individual/individual-billing-routing.module.ts
```diff
@@ -19,7 +19,7 @@ const routes: Routes = [
     component: SubscriptionComponent,
     data: { titleId: "subscription" },
     children: [
-      { path: "", pathMatch: "full", redirectTo: "premium" },
+      { path: "", pathMatch: "full", redirectTo: "user-subscription" },
```
amorask-bitwarden, line 22: ❓ Was this change intentional? It does not seem correct.

apps/web/src/app/billing/individual/subscription/account-subscription.component.ts
```diff
@@ -199,6 +206,8 @@ export class AccountSubscriptionComponent {
     return hasPremiumPersonally && upgradeEnabled;
   });
 
+  readonly activeAccount = toSignal(this.accountService.activeAccount$);
```
amorask-bitwarden: ❌ This already exists on line 78.

apps/web/src/app/billing/individual/subscription.component.ts
```diff
@@ -21,9 +26,21 @@ export class SubscriptionComponent implements OnInit {
     private platformUtilsService: PlatformUtilsService,
     billingAccountProfileStateService: BillingAccountProfileStateService,
     accountService: AccountService,
+    private configService: ConfigService,
```
amorask-bitwarden: ⛏️ These don't need to be marked with `private` if they're only used inside the constructor.

PR #18477, opened 2026-01-21
[PM-26049] Always store users auto unlock key on Cli

mzieniukbw:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-26049

## 📔 Objective

Always sets the users auto unlock key into the store on Cli.

Since the vault timeout state can change due to affecting policy, which we want to prevent from happening for CLI, changed how the effective vault timeout is determined, by always going through checking if it is available for client, before setting it to state. For CLI, this prevents the timeout change in state due to a policy.

### What this means if you couldn't unlock your Cli vault so far ?
If the effective vault timeout in state for Cli (`data.jon`) is already in "bad" state (as of, not `"never"`), the unlocking behaviour is unaffected, because the auto key will now consistently get stored (Just run `bw unlock` once). Also any later subsequent `bw sync -f` will fix the bad state (although it's not always guaranteed to happen immediately due to vault timeout state being asynchronous, while for Cli it's being synchronous - simply ignore the vault timeout state for Cli)

### DIFF
--- a/libs/key-management/src/key.service.ts
+++ b/libs/key-management/src/key.service.ts
@@ -674,7 +674,7 @@ export class DefaultKeyService implements KeyServiceAbstraction {
   protected async shouldStoreKey(keySuffix: KeySuffixOptions, userId: UserId) {
     switch (keySuffix) {
       case KeySuffixOptions.Auto: {
-        // Cli have fixed Never vault timeout, and it should not be affected by a policy.
+        // Cli has fixed Never vault timeout, and it should not be affected by a policy.
         if (this.platformUtilService.getClientType() == ClientType.Cli) {
           return true;
         }


libs/key-management/src/key.service.ts
```diff
@@ -671,9 +672,13 @@ export class DefaultKeyService implements KeyServiceAbstraction {
   }
 
   protected async shouldStoreKey(keySuffix: KeySuffixOptions, userId: UserId) {
-    let shouldStoreKey = false;
     switch (keySuffix) {
       case KeySuffixOptions.Auto: {
+        // Cli have fixed Never vault timeout, and it should not be affected by a policy.
```
quexten:
```suggestion
        // Cli has fixed Never vault timeout, and it should not be affected by a policy.
```

nit (non-blocking)

PR #18484, opened 2026-01-21
Enhancements to `EncryptedMemoryStore`

neuronull:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-31075

## 📔 Objective

As the ssh agent refactored will be using the `EncryptedMemoryStore` to store the keys in memory, a few changes are necessary (the changes in this PR to EncryptedMemoryStore will be utilized in a follow-up PR for the ssh agent):

- Use `BTreeMap` instead of `HashMap` , this preserves the sort order which is needed for listing of the keys
- Use a Generic for the btreemap key type, this allows us to use `PublicKey` in ssh agent, and other users of the encrypted memory store are similarly not limited to using String.
- The methods that can error out if the memory was tampered with, changed those to propagate the decryption error, so that the caller can decide if it's necessary to handle.

Note that some tweaks to the other user of trait `SecureMemoryStore` (the biometrics_v2) were needed to adapt.

### DIFF
--- a/apps/desktop/desktop_native/core/src/secure_memory/encrypted_memory_store.rs
+++ b/apps/desktop/desktop_native/core/src/secure_memory/encrypted_memory_store.rs
@@ -40,6 +40,7 @@ where
     /// # Returns
     ///
     /// An array of all decrypted values.
+    /// Due to the usage of `BtreeMap`, the order is deterministic.
     ///
     /// # Errors
     ///
@@ -141,5 +142,74 @@ mod tests {
         assert_eq!(store.get(&key).expect("entry in map for key"), Some(value));
         store.remove(&key);
         assert!(!store.has(&key));
+    }
+
+    #[test]
+    fn test_to_vec_contains_all() {
+        let mut store = EncryptedMemoryStore::default();
+
+        for size in 0..=2048 {
+            let key = format!("test_key_{size}");
+            let value: Vec<u8> = (0..size).map(|i| (i % 256) as u8).collect();
+            store.put(key.clone(), &value);
+        }
+        let vec_values = store.to_vec().expect("decryption to not fail");
+
+        // to_vec() should contain same number of values as inserted
+        assert_eq!(vec_values.len(), 2049);
+
+        // the value from the store should match the value in the vec
+        let keys: Vec<_> = store.map.keys().cloned().collect();
+        for (store_key, vec_value) in keys.iter().zip(vec_values.iter()) {
+            let store_value = store.get(store_key).expect("entry in map for key").unwrap();
+            assert_eq!(&store_value, vec_value);
+            store.remove(store_key);
+        }
+
+        // all values were present
+        assert!(store.map.is_empty());
+    }
+
+    #[test]
+    fn test_to_vec_preserves_sorted_key_order() {
+        let mut store = EncryptedMemoryStore::new();
+
+        // insert in non-sorted order
+        store.put("morpheus", &[4, 5, 6]);
+        store.put("trinity", &[1, 2, 3]);
+        store.put("dozer", &[7, 8, 9]);
+        store.put("neo", &[10, 11, 12]);
+
+        let vec = store.to_vec().expect("decryption to not fail");
+
+        assert_eq!(
+            vec,
+            vec![
+                vec![7, 8, 9],    // dozer
+                vec![4, 5, 6],    // morpheus
+                vec![10, 11, 12], // neo
+                vec![1, 2, 3],    // trinity
+            ]
+        );
+    }
+
+    #[test]
+    fn test_to_vec_order_after_remove() {
+        let mut store = EncryptedMemoryStore::new();
+
+        // insert in non-sorted order
+        store.put("trinity", &[3]);
+        store.put("morpheus", &[1]);
+        store.put("neo", &[2]);
+
+        let vec = store.to_vec().expect("decryption to not fail");
+
+        assert_eq!(vec, vec![vec![1], vec![2], vec![3]]);
+
+        store.remove(&"neo");
+
+        let vec = store.to_vec().expect("decryption to not fail");
+
+        assert_eq!(vec, vec![vec![1], vec![3]]);
     }
 }


apps/desktop/desktop_native/core/src/secure_memory/dpapi.rs
```diff
@@ -77,15 +79,19 @@ impl SecureMemoryStore for DpapiSecretKVStore {
                     .expect("length header should be usize"),
```
neuronull, line 79: I'll leave it to reviewers who are more familiar with this code to decide if we want to return an Error here instead of panic, now that the function definition supports that. It could be done here or a follow up PR, I wanted to keep my changes to this area as minimal as possible 🙏

neuronull:
Thomas you are welcome to review I just tagged Bernd since he is up to speed on the background for these changes.

apps/desktop/desktop_native/core/src/biometric_v2/mod.rs
```diff
@@ -21,14 +21,17 @@ pub trait BiometricTrait: Send + Sync {
     /// enrollment, this function should do nothing.
     async fn enroll_persistent(&self, user_id: &str, key: &[u8]) -> Result<()>;
     /// Clear the persistent and ephemeral keys
-    async fn unenroll(&self, user_id: &str) -> Result<()>;
+    #[allow(clippy::ptr_arg)] // to allow using user_id as map key type
+    async fn unenroll(&self, user_id: &String) -> Result<()>;
```
quexten, line 25: Thought (not for here) we should probably have a newtype around a GUID similar to what the SDK has.

apps/desktop/desktop_native/core/src/secure_memory/dpapi.rs
quexten: Makes sense. Really we never want the memory to be tampered with, so this should never happen. But it's fair to put that in the public contract as the result / I don't have issues with that.

For your use-case, do you think you would want to handle this in your code and handle it? Otherwise I think it's also fair to just `.expect` on it.

apps/desktop/desktop_native/core/src/secure_memory/encrypted_memory_store.rs
```diff
@@ -12,50 +14,86 @@ use crate::secure_memory::{
...
     }
+
+    /// # Returns
+    ///
+    /// An array of all decrypted values.
+    ///
+    /// # Errors
+    ///
+    /// `DecryptionError` if an error occured during decryption
+    pub fn to_vec(&mut self) -> Result<Vec<Vec<u8>>, DecryptionError> {
```
quexten, line 48: Could we get a unit test for `to_vec`?

vleague2:

apps/desktop/desktop_native/core/src/secure_memory/dpapi.rs
neuronull: > want the memory to be tampered with, so this should never happen.

I find it helpful to frame the decision with a question such as: can it happen due to things outside our control? , and if it does happen, are we ok with the consequences of a panic at this point in the runtime. if the answer is yes+no, we should probably Error instead of panic. (again I don't have the knowledge to make that call , but I'm happy to convert it to an Error in my PR if you'd like)

> For your use-case, do you think you would want to handle this in your code and handle it? Otherwise I think it's also fair to just .expect on it.

In the ssh agent case (the EncryptedMemoryStore case) , I changed (in this PR) to propagate the error, because the existing code was catching the `memory_encryption_key.decrypt` error, and returning `None` , and I found that problematic because the function also returns `None` when the key is not found in the map. , and I think that's something that would be worth distinguishing for a user (error mode = we didn't find the key or error mode = decryption failed ; one they can ostensibly do something about, one they can't)

apps/desktop/desktop_native/core/src/biometric_v2/mod.rs
neuronull: Worth a ticket?

apps/desktop/desktop_native/core/src/secure_memory/encrypted_memory_store.rs
neuronull: added in https://github.com/bitwarden/clients/pull/18484/commits/abd2758759b426f385d9e17042f30520c6ff59c1

it'd be neat if we could mock DecryptionError , but seems a bit out of scope for this PR.

apps/desktop/desktop_native/core/src/secure_memory/encrypted_memory_store.rs
```diff
@@ -12,50 +14,87 @@ use crate::secure_memory::{
 ///
 /// The key is briefly in process memory during encryption and decryption, in memory that is
 /// protected from swapping to disk via mlock, and then zeroed out immediately after use.
-#[allow(unused)]
-pub(crate) struct EncryptedMemoryStore {
-    map: std::collections::HashMap<String, EncryptedMemory>,
+/// # Type Parameters
+///
+/// * `K` - The type of the key.
+pub struct EncryptedMemoryStore<K>
```
coroiu, line 20: question/issue?: Why did this change visibility, I can't see it being used somewhere else? Is it to avoid `#[allow(unused)]`? Could we use this instead: ?

```suggestion
#[cfg(target_os = "linux")]
pub(crate) struct EncryptedMemoryStore<K>
```

quexten: It will be used in SSH agent by auto-fill, and is a public structure KM supports for consumption, so pub seems good from my side. Otherwise autofill will have to adjust it in the PR that they consume it in.

It was only limited to Linux because we first implemented it as part of the Linux biometrics v2 epic, but by no means should be limited to it.

coroiu: Ok, if it is supposed to be part of the public interface then no need to change 👍

neuronull: I noted the follow up PR that will consume these changes for ssh agent but did not specifically call out the visibility change, to Bernd's point we could preserve the original platform gate and visibility but I'd just be making the change in the follow up PR (which is coming soon) anyway.

quexten: We also publicly advertise it here https://contributing.bitwarden.com/architecture/deep-dives/memory-hardening#process-isolation-and-key-protection-on-desktop-apps, so it is just considered public API.

apps/desktop/desktop_native/core/src/biometric_v2/mod.rs
quexten: Not needed from my side.

apps/desktop/desktop_native/core/src/secure_memory/dpapi.rs
quexten: > I find it helpful to frame the decision with a question such as: can it happen due to things outside our control? , and if it does happen, are we ok with the consequences of a panic at this point in the runtime. if the answer is yes+no, we should probably Error instead of panic. (again I don't have the knowledge to make that call , but I'm happy to convert it to an Error in my PR if you'd like)

It should not. We already prevent debugger access to the app, and we never ever want the app to be debugged and it should rather crash and wipe all keys. So if anything is wrong here, crashing seems fine. I don't think we have encountered any legitimate case for this happening yet, and if that happens this would of course change the situation.

quexten: For your second point, I agree propagating is better than returning none. However, I don't think an error should ever occur in the first place.

neuronull: Thanks for clarifying! Will leave it as an expect()

PR #18486, opened 2026-01-21
[PM-30922] Client changes to encrypt send access email list

harr1424:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-30922

### Necessary client changes:

- [Send model changes](https://github.com/bitwarden/clients/tree/395e4f2c05585bb7ffb173ce9616ea75ae751132/libs/common/src/tools/send/models) (e.g. Request, Response, Domain, Data, View)
- The emails field will be either a list of email addresses or an encrypted string representing a list of email addresses depending on the type of model file
- A new field called emailHashes will hold a list of hashes of the email addresses from the list in the emails field.
- This value will be encrypted and decrypted on the server using ASP.NET Core Data Protection functionality.
- When the client code saves a Send, the server needs the hash list to be in plain-text (so that it can later be used during send access email verification)
- This field should not be visible to users and is only used internally
- When creating / saving a Send, the client code will need to populate the emailHashes field automatically
- The client code will need to hash all the email addresses from the emails field
- The SHA-256 algorithm can be used to hash the email address
- The values of emails and the emailHashes fields should always be in sync in terms that the hashes should always be performed from the data in the emails field
- Client should send this list of comma separated hashes to the server in plaintext, the server will encrypt/decrypt but will never send this property back to the client

### Acceptance criteria
- Creating or saving a Send with an email list results in a server request that includes an encrypted emails value and a list of hashes (of email addresses) for the emailHashes value

Expect send access to fail to auth until the server code changes that add an EmailHashescolumn are completed if the pm-19051-send-email-verification feature flag is enabled

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


vleague2:

libs/common/src/tools/send/services/send.service.ts
```diff
@@ -80,30 +85,45 @@ export class SendService implements InternalSendServiceAbstraction {
...
     }
 
+    // Check feature flag for email OTP authentication
+    const sendEmailOTPEnabled = await this.configService.getFeatureFlag(FeatureFlag.SendEmailOTP);
+
     const hasEmails = (model.emails?.length ?? 0) > 0;
-    if (hasEmails) {
-      send.emails = model.emails.join(",");
+
+    if (sendEmailOTPEnabled && hasEmails) {
```
mcamirault, line 93: ❓ Shouldn't we be encrypting and hashing even if the flag is off? There should be no way of creating a Send with the `email` field populated until we turn it on, but if something does slip through we'd want the database columns to still be appropriately encrypted/hashed, right?

apps/cli/src/tools/send/commands/create.command.ts
```diff
@@ -136,11 +148,6 @@ export class SendCreateCommand {
 
       const sendView = SendResponse.toView(req);
       const [encSend, fileData] = await this.sendService.encrypt(sendView, fileBuffer, password);
```
mcamirault, line 150: ❓ I get removing the email line because it's now being encrypted but why remove the deletion and expiration date lines as well? We're not encrypting those fields are we?

apps/cli/src/tools/send/commands/edit.command.ts
```diff
@@ -90,10 +107,6 @@ export class SendEditCommand {
 
     try {
       const [encSend, encFileData] = await this.sendService.encrypt(sendView, null, req.password);
```
mcamirault, line 109: Same question as on the create command

libs/common/src/tools/send/services/send.service.ts
harr1424: @itsadrago  what do you think about this? Do you think that relaxing the feature flag usage is appropriate here?

apps/cli/src/tools/send/commands/create.command.ts
harr1424: The deletion and expiration dates are not encrypted, but they are already copied from the `SendView` in the `encrypt()` function in the Send Service. The `encrypt()` function also copies a send's `id`, `type`, `disabled`, `hideEmail`, and `maxAccessCount`, from the `SendView` into the `Send`, none of these values will be encrypted. 

An argument could be made for extracting these assignments out of `encrypt()` and into a more aptly named helper method, but am not sure if this is the most appropriate time to do so. 

https://github.com/bitwarden/clients/blob/5dc49f21d2e5aa145fe9611222b4c0c5659db1c5/libs/common/src/tools/send/services/send.service.ts#L56-L70

apps/cli/src/tools/send/commands/edit.command.ts
harr1424: The deletion and expiration dates are not encrypted, but they are already copied from the `SendView` in the `encrypt()` function in the Send Service. The `encrypt()` function also copies a send's `id`, `type`, `disabled`, `hideEmail`, and `maxAccessCount`, from the `SendView` into the `Send`, none of these values will be encrypted. 

An argument could be made for extracting these assignments out of `encrypt()` and into a more aptly named helper method, but am not sure if this is the most appropriate time to do so. 

https://github.com/bitwarden/clients/blob/5dc49f21d2e5aa145fe9611222b4c0c5659db1c5/libs/common/src/tools/send/services/send.service.ts#L56-L70

libs/common/src/tools/send/services/send.service.ts
itsadrago: My suggestion is to make sure we are doing as much validation as we can on the server side. We're currently doing that for the emails value but as discussed internally there is more that we can do for the email hashes value

mcamirault: Sounds good

apps/cli/src/tools/send/commands/create.command.ts
mcamirault: Gotcha, as long as they are copied elsewhere and aren't being newly encrypted we should have no changes in behavior

apps/cli/src/tools/send/commands/edit.command.ts
mcamirault: Resolved above

PR #18493, opened 2026-01-22
[PM-30307] Session key retrieval redesign for the local backed session storage

mzieniukbw:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-30307

## 📔 Objective

Redesigned the way local backed session storage interacts with session key, to prevent the decryption log errors from happening as a means to detect the items to cleanup.

Context: this storage is mostly used as a caching mechanism for some decrypted items (like decrypted folders), that is intended to be wiped with session storage wipe (i.e. process reload).
Note, technically speaking it is a "memory" storage, but it needs to survive background worker being killed by browser (i.e. due to being idle), hence why this design.

The session key retrieval is separated to:
- `get`, can return `null` when session key is not present (which indicated session storage cleanup due to browser restart, process reload), as a means to clean the local storage backed items
- `create` to create and store new session key

## 📸 Screenshots

After (no errors):
image

Before:
image

### DIFF
--- a/apps/browser/src/platform/services/local-backed-session-storage.service.ts
+++ b/apps/browser/src/platform/services/local-backed-session-storage.service.ts
@@ -33,6 +33,9 @@ const SESSION_KEY_PREFIX = "session_";
  * in local storage cannot be decrypted and must be cleared to maintain data consistency.
  *
  * This provides session-scoped security for sensitive data while using persistent local storage as the backing store.
+ *
+ * @internal Internal implementation detail. Exported only for testing purposes.
+ * Do not use this class directly outside of tests. Use LocalBackedSessionStorageService instead.
  */
 export class SessionKeyResolveService {
   constructor(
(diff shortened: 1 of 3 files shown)


apps/browser/src/platform/services/local-backed-session-storage.service.ts
```diff
@@ -12,27 +13,78 @@ import {
...
+/**
+ * Manages an ephemeral session key for encrypting session storage items persisted in local storage.
+ *
+ * The session key is stored in session storage and automatically cleared when the browser session ends
+ * (e.g., browser restart, extension reload). When the session key is unavailable, any encrypted items
+ * in local storage cannot be decrypted and must be cleared to maintain data consistency.
+ *
+ * This provides session-scoped security for sensitive data while using persistent local storage as the backing store.
+ */
+export class SessionKeyResolveService {
```
MGibson1, line 40: **nit**: This is a pretty isolated use-case, I think it'd be nicer for this to be internal to the module, the local-backed-session-storage to take two storage locations: memory and disk, and generates its own key service

mzieniukbw: I've made it more internal, but i can't provide unit tests without "export" keyword, so decided to keep it and put a jsdoc note that it's internal. Also no longer being outside for this service.ts file.

MGibson1 (approved):

coroiu (approved):
Deferring to Matt's expertise on this :)

PR #18494, opened 2026-01-22
Desktop/pm 18768/migrate vault cipher list

lxiong-livefront:
## 🎟️ Tracking

[JIRA](https://bitwarden.atlassian.net/browse/PM-18768?atlOrigin=eyJpIjoiNDJmMTdiYmVmYjYzNDI4ZWE5YjczY2RjODY2ZDUwZDQiLCJwIjoiaiJ9)

## 📔 Objective

Migrated vault cipher list 
- created new `vault-list.component` in desktop based on `vault-items.component` in web
    - in order to align with new design, I duplicated `vault-cipher-row.component` and `vault-collection-row.component` modifying their html templates
- updated `vault.component` to consume new vault-list and aligned where possible with web's vault component to ensure proper functionality
- lifted search component from desktop header into new vault list component
- incorporated new desktop header into vault list and added 'new' button
- updated vault 'details': added desktop header and updated footer aligning with new design
- moved shared files to `/libs`

## 📸 Screenshots

https://github.com/user-attachments/assets/f23feba8-59ac-4a20-860f-865cc45fb018


Hinton:
I don't think this approach is viable and we're better off just implementing the bare minimum needed in the display, i.e. match the figma which is a table that has items that looks like this:

image

PR #18376, opened 2026-01-15
[PM-25082] - update browser extension widths

jaasen-livefront:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-25082

## 📔 Objective

 ***** **Note** *****
I'm not sure if the changes to the css are required. Please let me know!

## 📸 Screenshots

https://github.com/user-attachments/assets/12f90db0-9d1a-4c02-a0e4-1b73abd3fe38

## ⏰ Reminders before review

- Contributor guidelines followed
- All formatters and local linters executed and passed
- Written new unit and / or integration tests where applicable
- Protected functional changes with optionality (feature flags)
- Used internationalization (i18n) for all UI strings
- CI builds passed
- Communicated to DevOps any deployment requirements
- Updated any necessary documentation (Confluence, contributing docs) or informed the documentation team

## 🦮 Reviewer guidelines

- 👍 (`:+1:`) or similar for great changes
- 📝 (`:memo:`) or ℹ️ (`:information_source:`) for notes or general info
- ❓ (`:question:`) for questions
- 🤔 (`:thinking:`) or 💭 (`:thought_balloon:`) for more open inquiry that's not quite a confirmed issue and could potentially benefit from discussion
- 🎨 (`:art:`) for suggestions / improvements
- ❌ (`:x:`) or ⚠️ (`:warning:`) for more significant problems or concerns needing attention
- 🌱 (`:seedling:`) or ♻️ (`:recycle:`) for future improvements or indications of technical debt
- ⛏ (`:pick:`) for minor or nitpick changes

PR #18499, opened 2026-01-22
[PM-31142] Do not set extension width when ran in full browser tab

nick-livefront:
## 🎟️ Tracking

[PM-31142](https://bitwarden.atlassian.net/browse/PM-31142)

## 📔 Objective

[#18376](https://github.com/bitwarden/clients/pull/18376/changes#diff-e3e9509468b5e6777120f1dfa7eebd433799e79819b83e4f1bcb96c399595f1aL86-R86) introduced a change to move from `minWidth` to `width` which is needed because the predefined sizes of the extension increase and decrease from the default. 

This caused an issue when the extension is opened in a full browser tab as the width would be smaller than expected in the window. This change does not set the `width` when the extension is opened in a full browser tab. 

## 📸 Screenshots

<video src="https://github.com/user-attachments/assets/5d3c6ac4-1f32-4b8f-b072-f4f89d1ecd8a" />

[PM-31142]: https://bitwarden.atlassian.net/browse/PM-31142?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/browser/src/platform/popup/layout/popup-size.service.ts
+++ b/apps/browser/src/platform/popup/layout/popup-size.service.ts
@@ -37,7 +37,7 @@ export class PopupSizeService {
   /** Begin listening for state changes */
   async init() {
     this.width$.subscribe((width: PopupWidthOption) => {
-      PopupSizeService.setStyle(width);
+      void PopupSizeService.setStyle(width);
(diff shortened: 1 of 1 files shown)


vleague2:

PR #18204, opened 2026-01-05
[PM-27086] Use new KM Data Types in InputPasswordComponent flows - TDE Offboarding

rr-bw:
## 🎟️ Tracking

[PM-27086](https://bitwarden.atlassian.net/browse/PM-27086)

## 📔 Objective

Updates the `setInitialPasswordTdeOffboarding` path to use new KM data types:
- `MasterPasswordAuthenticationData`
- `MasterPasswordUnlockData`

This allows us to move away from the deprecated `makeMasterKey()` method (which takes email as salt) as we seek to eventually separate the email from the salt.

Behind feature flag: `pm-27086-update-authentication-apis-for-input-password`

### Web

https://github.com/user-attachments/assets/9f535130-386e-4ae7-bdae-14b0196019ff

### Extension

https://github.com/user-attachments/assets/4fad7f6d-09f7-40db-96e9-6ed95afcc7b7

### Desktop

https://github.com/user-attachments/assets/389d8aad-d6a8-47d8-9090-afba922f38ed

## ⏰ Reminders before review

- Contributor guidelines followed
- All formatters and local linters executed and passed
- Written new unit and / or integration tests where applicable
- Protected functional changes with optionality (feature flags)
- Used internationalization (i18n) for all UI strings
- CI builds passed
- Communicated to DevOps any deployment requirements
- Updated any necessary documentation (Confluence, contributing docs) or informed the documentation team

## 🦮 Reviewer guidelines

- 👍 (`:+1:`) or similar for great changes
- 📝 (`:memo:`) or ℹ️ (`:information_source:`) for notes or general info
- ❓ (`:question:`) for questions
- 🤔 (`:thinking:`) or 💭 (`:thought_balloon:`) for more open inquiry that's not quite a confirmed issue and could potentially benefit from discussion
- 🎨 (`:art:`) for suggestions / improvements
- ❌ (`:x:`) or ⚠️ (`:warning:`) for more significant problems or concerns needing attention
- 🌱 (`:seedling:`) or ♻️ (`:recycle:`) for future improvements or indications of technical debt
- ⛏ (`:pick:`) for minor or nitpick changes

[PM-27086]: https://bitwarden.atlassian.net/browse/PM-27086?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #18507, opened 2026-01-22
[PM-27086] Use new KM Data Types in InputPasswordComponent flows - Change Password

rr-bw:
## 🎟️ Tracking

[PM-27086](https://bitwarden.atlassian.net/browse/PM-27086)

## 📔 Objective

Updates the `ChangePasswordService` (`changePassword()` and `changePasswordForAccountRecovery()`) to use the new KM data types :
- `MasterPasswordAuthenticationData`
- `MasterPasswordUnlockData`

This allows us to move away from the deprecated `makeMasterKey()` method (which takes email as salt) as we seek to eventually separate the email from the salt.

Behind feature flag: `pm-27086-update-authentication-apis-for-input-password`

### Web - Normal Change Password

https://github.com/user-attachments/assets/832fa8a1-4bf0-4da3-89ed-fba797cabf57

### Web - Change Password (with Key Rotation)

https://github.com/user-attachments/assets/e190a3a7-7f4e-4384-8168-2bc8c0780bff

### Web - Account Recovery Follow-Up

https://github.com/user-attachments/assets/459a1fc0-ec55-4cb6-9a5f-57e649e70691

### Web - Weak Password

https://github.com/user-attachments/assets/3ea0afe3-acec-4d39-9802-e735e65c41ed

### Web - Email Invite (existing user has Weak Password)

https://github.com/user-attachments/assets/54087045-375a-40e2-a4f7-f6e252d7b13a

### Extension - Account Recovery

https://github.com/user-attachments/assets/dd4a826f-129a-49b9-94c8-70637c19cbe2

### Extension - Account Recovery (next user Locked)

https://github.com/user-attachments/assets/dbfc8e11-67fd-4944-874f-410f71e8ff0e

### Extension - Account Recovery (next user Unlocked)

https://github.com/user-attachments/assets/080bdad9-a92a-4227-b40a-39da18228673

### Extension - Weak Password

https://github.com/user-attachments/assets/6c46dfc2-63e7-42aa-a4af-1907b9085f25

### Extension - Weak Password (next user Locked)

https://github.com/user-attachments/assets/01eb9b87-140f-4458-8a14-be77d58573ea

### Extension - Weak Password (next user Unlocked)

https://github.com/user-attachments/assets/55a32445-535e-4b41-b246-bdeba74c6a2e

### Desktop - Account Recovery

https://github.com/user-attachments/assets/5c00e38f-dbfb-4145-a453-24b14585a48f

### Desktop - Account Recovery (next user Unlocked)

https://github.com/user-attachments/assets/493ed567-7c4c-4b1c-8f6b-07e07e8c1185

### Desktop - Weak Password

https://github.com/user-attachments/assets/06308926-dfe9-4005-8998-d59340c9e0fc

### Desktop - Weak Password (next user Unlocked)

https://github.com/user-attachments/assets/209006b5-3106-4ca5-880a-aa29939c71ed

[PM-27086]: https://bitwarden.atlassian.net/browse/PM-27086?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/web/src/app/auth/core/services/change-password/web-change-password.service.spec.ts
rr-bw, line 1: ~~Changes to this web test file are not related to the rest of the changes in this PR. I just saw there were some issues with this test file and decided to fix them while I'm working in this domain.~~

Outdated comment. See below.

libs/angular/src/auth/password-management/change-password/change-password.service.abstraction.ts
```diff
@@ -34,12 +43,19 @@ export abstract class ChangePasswordService {
    * @param userId the `userId`
    * @throws if the `userId`, `currentMasterKey`, or `currentServerMasterKeyHash` is not found
    */
-  abstract changePassword(
-    passwordInputResult: PasswordInputResult,
-    userId: UserId | null,
-  ): Promise<void>;
+  abstract changePassword(passwordInputResult: PasswordInputResult, userId: UserId): Promise<void>;
```
rr-bw, line 71: Both flagged and unflagged logic require the `userId`, so removing `| null` here.

libs/angular/src/auth/password-management/change-password/default-change-password.service.ts
```diff
@@ -109,4 +191,53 @@ export class DefaultChangePasswordService implements ChangePasswordService {
...
+    newUnlockData: MasterPasswordUnlockData;
+  }> {
+    const ctx = "Could not change password.";
+    assertTruthy(passwordInputResult.newPassword, "newPassword", ctx);
+    assertTruthy(passwordInputResult.salt, "salt", ctx);
+    assertNonNullish(passwordInputResult.kdfConfig, "kdfConfig", ctx);
+
+    // We don't need to verify the currentPassword here because by this point it has already
+    // been verified via proofOfDecryption before being emitted from the InputPasswordComponent.
+    // We can simply get the user key from state and use it.
```
rr-bw: [Outdated comment/video - see follow-up comment]

~Video demonstrating that current password verification still works in new flagged code via `proofOfDecryption` on the `InputPasswordComponent` alone:~

https://github.com/user-attachments/assets/37136a52-83a4-41d8-b114-b20c7f01d9b2

libs/common/src/auth/models/request/password.request.ts
```diff
@@ -18,14 +19,18 @@ export class PasswordRequest extends SecretVerificationRequest {
   // This will eventually be changed to be an actual constructor, once all callers are updated.
   // https://bitwarden.atlassian.net/browse/PM-23234
   static newConstructor(
-    authenticationData: MasterPasswordAuthenticationData,
-    unlockData: MasterPasswordUnlockData,
+    currentMasterPasswordAuthenticationHash: MasterPasswordAuthenticationHash,
```
rr-bw, line 22: @JaredSnider-Bitwarden Should I actually be passing in the entire _current_ `MasterPasswordUnlockData` (so that the current hash/salt/kdf bundle gets sent to the server in the future), or is just the current hash enough here?

libs/auth/src/angular/input-password/input-password.component.ts
```diff
@@ -342,17 +344,40 @@ export class InputPasswordComponent implements OnInit {
...
+            currentPassword,
+            this.userId,
+          );
+          if (!currentPasswordVerified) {
+            this.toastService.showToast({
+              variant: "error",
+              message: this.i18nService.t("invalidMasterPassword"),
+            });
+            return;
+          }
```
JaredSnider-Bitwarden: Per sync discussion, we should move this `proofOfDecyption` check into the change password service to make the change password service more bulletproof and extract change password logic out of the input password component.

libs/angular/src/auth/password-management/change-password/default-change-password.service.ts
JaredSnider-Bitwarden: See https://github.com/bitwarden/clients/pull/18507/changes#r2866418903 for discussion. This will need to be updated.

libs/common/src/auth/models/request/password.request.ts
JaredSnider-Bitwarden: I think just the existing hash should be sufficient for ensuring the user has access to the current password b/c if they didn't have the right salt or KDF settings, it wouldn't match when we verify it on the server. However, @quexten or @Thomas-Avery , could you speak to whether we should send up the whole current `MasterPasswordAuthenticationData` for a change password request?

Thomas-Avery: The new `MasterPasswordUnlockData` and the new `MasterPasswordAuthenticationData` we would want to send. 

For the current is it being used for user authorization? Basically does the `currentMasterPasswordAuthenticationHash` match what is saved in the server? I can't think of any reason why to use the whole current `MasterPasswordAuthenticationData`. Maybe some debugging logs comparing the salt and KDF sent vs what's in the DB when that comparison fails?

libs/angular/src/auth/password-management/change-password/change-password.component.ts
```diff
@@ -156,7 +173,7 @@ export class ChangePasswordComponent implements OnInit {
 
         await this.syncService.fullSync(true);
```
rr-bw, line 175: In the new flagged code this sync call is moved to the `changePasswordAndRotateUserKey()` method on the ~`DefaultChangePasswordService`~.

Update: now on the `WebChangePasswordService`.

apps/web/src/app/auth/core/services/change-password/web-change-password.service.spec.ts
rr-bw: Some changes to this test file are now in fact related to other changes in the PR.

libs/angular/src/auth/password-management/change-password/default-change-password.service.ts
```diff
@@ -1,34 +1,92 @@
...
   constructor(
     protected keyService: KeyService,
     protected masterPasswordApiService: MasterPasswordApiService,
     protected masterPasswordService: InternalMasterPasswordServiceAbstraction,
+    protected masterPasswordUnlockService: MasterPasswordUnlockService,
+    protected syncService: SyncService,
   ) {}
 
-  async rotateUserKeyMasterPasswordAndEncryptedData(
+  async changePasswordAndRotateUserKey(
```
rr-bw, line 35: @JaredSnider-Bitwarden 

For the change password + key rotation scenario, here is my approach to having current password verification run before `rotateUserKeyMasterPasswordAndEncryptedData()`:

- New Path (flagged):
  - The `ChangePasswordComponent` calls this `changePasswordAndRotateUserKey()` method, which:
    - (a) performs current password validation + sync
    - (b) calls the new protected method: `this.rotateUserKeyMasterPasswordAndEncryptedData` (does not exist on the abstraction)
      - the `protected` modifier limits/reduces possibility of calling the method without first performing current password validation
- Old Path:
  - Current password verification happens on the `InputPasswordComponent`
  - The `ChangePasswordComponent` calls the old (public) `this.rotateUserKeyMasterPasswordAndEncryptedDataOld()` method directly (has "Old" suffix now)

Your work in https://github.com/bitwarden/clients/pull/19289 might mean we can remove `rotateUserKeyMasterPasswordAndEncryptedData` from the Web service and just call `userKeyRotationService.rotateUserKeyMasterPasswordAndEncryptedData` directly on the default service instead of throwing "rotateUserKeyMasterPasswordAndEncryptedData() is only implemented in Web". However, do note that the `UserKeyRotationService` is currently in the web/ folder.

rr-bw:
@JaredSnider-Bitwarden Here are some more recordings since my recent changes. I can do more recordings before passing off to QA.

### Web - Change Password

https://github.com/user-attachments/assets/26aa970d-0aba-4966-afc5-ae4a4ed335c6

### Web - Account Recovery

https://github.com/user-attachments/assets/c1ab431e-4f7d-474f-a979-7705d4453dcb

### Extension (account switching - next user locked)

https://github.com/user-attachments/assets/a62638ad-05c1-40e6-8c88-427d33cd15ad

### Desktop (account switching - next user unlocked)

https://github.com/user-attachments/assets/cfb13427-08c5-45fb-a4e1-2ff83447fe65

libs/angular/src/auth/password-management/change-password/change-password.component.ts
```diff
@@ -181,19 +203,33 @@ export class ChangePasswordComponent implements OnInit {
           message: this.i18nService.t("masterPasswordChanged"),
         });
 
-        // TODO: PM-23515 eventually use the logout service instead of messaging service once it is available without circular dependencies
-        this.messagingService.send("logout");
+        if (passwordInputResult.newApisWithInputPasswordFlagEnabled) {
+          // TODO: investigate refactoring logout and follow-up routing in https://bitwarden.atlassian.net/browse/PM-32660
+          await this.logoutService.logout(this.userId);
+          // navigate to root so redirect guard can properly route next active user or null user to correct page
+          await this.router.navigate(["/"]);
```
rr-bw: @JaredSnider-Bitwarden 

This navigation (placed here) will actually break a `changePassword()` flow on Web:

When an existing user with a MP gets invited to an org that has a MP policy enforced, when that user clicks the email invite they get prompted to change their password via `/change-password`. But upon changing their password this line routes to `/`, which triggers the redirect guard, which routes to `/vault`, which clears the deep link redirect URL due to the `deepLinkGuard` still recognizing the user as unlocked. This means that upon subsequent login they don't get routed back to `/accept-organization`, and the invite is never accepted. Video below.

Do you think I should?:
1. Move this logic back up to run only for `changePasswordForAccountRecovery()` (and not `changePassword()` also)?
2. Remove this block of code entirely (lines 206-211) and just use `messagingService.send()` like we were already doing - and then solve these logout/redirect issues separately as part of https://bitwarden.atlassian.net/browse/PM-32660?
3. Some other suggestion?

https://github.com/user-attachments/assets/9721e9ac-5e92-48fe-b38a-8ea5e22e4b4d

libs/angular/src/auth/password-management/change-password/default-change-password.service.ts
```diff
@@ -109,4 +252,57 @@ export class DefaultChangePasswordService implements ChangePasswordService {
...
+  private async verifyCurrentPasswordAndGetUserKey(currentPassword: string, userId: UserId) {
+    const currentPasswordVerified = await this.masterPasswordUnlockService.proofOfDecryption(
+      currentPassword,
+      userId,
+    );
+    if (!currentPasswordVerified) {
+      throw new InvalidCurrentPasswordError();
+    }
+
+    const userKey = await firstValueFrom(this.keyService.userKey$(userId));
```
rr-bw: @quexten @JaredSnider-Bitwarden 

Is this an acceptable way for me to access the user key? That is, I want to verify that the current password is correct before getting the user key, so I created a private method called `verifyCurrentPasswordAndGetUserKey()` which runs `proofOfDecryption`, and if `true`, then simply gets the user key from state and returns it.

Or is it expected that I should get the user key via master password service `masterPasswordUnlockData$` then `unwrapUserKeyFromMasterPasswordUnlockData`? Doing so would essentially mean re-creating the `proofOfDecryption` method, but returning the user key instead of a boolean.

Either way, I wonder if this would be a use-case for another helper method on the `MasterPasswordUnlockService` that proves decryption _and_ returns the decrypted user key?

quexten: Getting the user-key in this way is acceptable, yes. I don't know why we would need a helper function in the MPUnlockService *or* this file though. Why not just:

```ts
if (!verifyCurrentPassword(currentPassword, userId)) {
  throw ...
}

const userKey = await firstValueFrom(this.keyService.userKey$(userId));
if (userKey == null) {
  throw ...
}
```

or even
```ts
if (!verifyCurrentPassword(currentPassword, userId)) {
  throw ...
}

const userKey = await firstValueFromOrThrow(this.keyService.userKey$(userId));
```

I don't really understand for which use-case these two actions would be logically combined / why we would build an abstraction that does both.

libs/angular/src/auth/password-management/change-password/change-password.component.ts
JaredSnider-Bitwarden: Per sync discussion, we will leverage the change password service layer to add a new `shouldNavigateToRoot` method which will be `false` in the default implementation and `true` in the extension implementation.

libs/angular/src/auth/password-management/change-password/default-change-password.service.spec.ts
```diff
@@ -162,6 +178,272 @@ describe("DefaultChangePasswordService", () => {
...
+
+      it("should call the postPassword() API method with the correct PasswordRequest", async () => {
+        // Act
+        await sut.changePassword(passwordInputResult, userId);
+
+        // Assert
+        expect(masterPasswordApiService.postPassword).toHaveBeenCalledTimes(1);
+        expect(masterPasswordApiService.postPassword).toHaveBeenCalledWith(request);
+      });
+    });
```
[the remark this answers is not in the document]
rr-bw: Tests added here https://github.com/bitwarden/clients/pull/18507/changes/63d5aea03e07991c61cb321943ad6d0e94d1376c

libs/angular/src/auth/password-management/change-password/default-change-password.service.ts
JaredSnider-Bitwarden: Per sync discussion, we should move `changePasswordAndRotateUserKey` to the web change password service as it doesn't make sense to be on the default. The default should just throw as change password + key rotation is not supported on non-web clients just yet.

libs/angular/src/auth/password-management/change-password/change-password.component.ts
rr-bw: Addressed here: https://github.com/bitwarden/clients/pull/18507/changes/c9178cab5761a317bc6675970ff8ed64543060b2

libs/angular/src/auth/password-management/change-password/default-change-password.service.ts
rr-bw: Addressed here: https://github.com/bitwarden/clients/pull/18507/changes/1b83cf2264bafb20451c9f981991b05a11f62d8e

libs/angular/src/auth/password-management/change-password/default-change-password.service.ts
rr-bw: Thanks @quexten!

@JaredSnider-Bitwarden: addressed here: https://github.com/bitwarden/clients/pull/18507/changes/3aaf68197b0800def2912e8e6d191f13d7976514

libs/auth/src/angular/input-password/input-password.component.ts
rr-bw: Updated across a few commits - so just see current PR code for DefaultChangePasswordService.

libs/angular/src/auth/password-management/change-password/default-change-password.service.ts
rr-bw: Updated across a few commits - so just see current PR code for DefaultChangePasswordService.

rr-bw: With flag on, `proofOfDecryption()` now happens on `ChangePasswordService`, and does NOT happen on `InputPasswordComponent`.

rr-bw:
@JaredSnider-Bitwarden

I ended up removing the clear `forceSetPasswordReason` call here: https://github.com/bitwarden/clients/pull/18507/changes/91ae438d2ec0c8e20c855fa770cee4884c659d0d, as I don't think it's necessary.

Reasoning:

On the component, by the time we call `await this.router.navigate(["/"])` (which is only called on Extension), the current active user will already have switched to be the next up user, in which case any routing/authGuard/forceSetPasswordReason logic applies to that new user - not to the user who just completed account recovery. In other words, I don't think clearing the account-recovery user's forceSetPasswordReason (which is user scoped state) actually matters because we've already moved on to a different user by the time we route to `/`.

In the following video, I placed console logs to log out the active user in two places: (1) in the `ChangePasswordComponent` immediately before we call `await this.router.navigate(["/"])`, and (2) in the `AuthGuard`. Both logs show that after `user10` changes their password, by the time we route to `/` we have already switched to `user1` as the active user.


https://github.com/user-attachments/assets/8886732c-f07e-47ab-b354-05d28e1f5abf

libs/common/src/auth/models/request/password.request.ts
rr-bw: Thanks @Thomas-Avery,

Yes the current hash is being used for authentication by comparing it with the hash on the server (see [here](https://github.com/bitwarden/server/blob/main/src/Api/Auth/Controllers/AccountsController.cs#L200) then [here](https://github.com/bitwarden/server/blob/main/src/Core/Services/Implementations/UserService.cs#L479)). So it seems that just sending the current hash is sufficient.

However, I did notice that `CheckPasswordAsync` has the possibility of hitting the `PasswordVerificationResult.SuccessRehashNeeded` path [here](https://github.com/bitwarden/server/blob/main/src/Core/Services/Implementations/UserService.cs#L915).

The doc comment on `PasswordVerificationResult.SuccessRehashNeeded` says "Indicates password verification was successful however the password was encoded using a deprecated algorithm and should be rehashed and updated."

I'm not sure if this "deprecated algorithm" is referring to the moment when the password was hashed on the client, or if it refers to the moment when the hash was itself hashed on the server (per our [whitepaper](https://bitwarden.com/help/bitwarden-security-white-paper/#hashing-key-derivation-and-encryption): "Once reaching the server, the Master Password Hash is hashed again using PBKDF2-SHA256 with a random salt and 600,000 iterations".

Either way, do you see any reason here to send up the whole current `MasterPasswordAuthenticationData` as opposed to just the current hash?

apps/browser/src/auth/popup/change-password/extension-change-password.service.ts
```diff
@@ -17,13 +18,25 @@ export class ExtensionChangePasswordService
...
+
   closeBrowserExtensionPopout(): void {
     if (BrowserPopupUtils.inPopout(this.win)) {
       BrowserApi.closePopup(this.win);
     }
   }
+
+  /**
+   * In the extension, the user can popout the extension view and if we don't have a root navigation call, then
+   * the user is left on the change-password page after changing their password.
```
JaredSnider-Bitwarden: ❓ : I thought they were left on a blank new tab after logout?

libs/angular/src/auth/password-management/change-password/default-change-password.service.ts
```diff
@@ -73,7 +95,60 @@ export class DefaultChangePasswordService implements ChangePasswordService {
...
+        currentAuthenticationData.masterPasswordAuthenticationHash,
+        newAuthenticationData,
+        newUnlockData,
+        passwordInputResult.newPasswordHint,
+      );
+
+      try {
+        await this.masterPasswordApiService.postPassword(request);
+      } catch {
+        throw new Error("Error during change password attempt. Could not change password.");
```
JaredSnider-Bitwarden: 🎨 : we should have a `LogService` log of the original error here since we swallow it and throw a new error.

libs/angular/src/auth/password-management/change-password/default-change-password.service.ts
```diff
@@ -92,6 +167,53 @@ export class DefaultChangePasswordService implements ChangePasswordService {
...
+        passwordInputResult.newPasswordHint,
+      );
+
+      try {
+        // TODO: PM-23047 will look to consolidate this into the change password endpoint.
+        await this.masterPasswordApiService.putUpdateTempPassword(request);
+      } catch {
+        throw new Error(
+          "Error during change password attempt. Could not change password for account recovery.",
+        );
```
JaredSnider-Bitwarden: 🎨 : we should have a `LogService` log of the original error here since we swallow it and throw a new error.

libs/angular/src/auth/password-management/change-password/change-password.component.ts
```diff
@@ -181,19 +203,37 @@ export class ChangePasswordComponent implements OnInit {
...
+          // TODO: investigate refactoring logout and follow-up routing in https://bitwarden.atlassian.net/browse/PM-32660
+          await this.logoutService.logout(this.userId);
+
+          const shouldNavigateToRoot = this.changePasswordService.shouldNavigateToRoot();
+          if (shouldNavigateToRoot) {
+            // navigate to root so redirect guard can properly route next active user (account switching) or null user to correct page
+            await this.router.navigate(["/"]);
+          }
+        } else {
+          // TODO: PM-23515 eventually use the logout service instead of messaging service once it is available without circular dependencies
```
JaredSnider-Bitwarden: Nit: This comment isn't really necessary since the new flagged code has swapped to use the `logoutService`

libs/angular/src/auth/password-management/change-password/change-password.component.ts
```diff
@@ -142,11 +151,20 @@ export class ChangePasswordComponent implements OnInit {
     this.submitting = true;
 
     try {
+      // Handle change password with user key rotation via PasswordSettingComponent (settings page)
```
JaredSnider-Bitwarden: Nit: it's probably worth stripping off `via PasswordSettingComponent (settings page)` as the eventual goal is to be able to use this component for user key rotation on non-web clients which would make this no longer accurate. I.e., a component that's meant to be re-used in multiple places generally shouldn't make statements like this if a different re-use can make it invalid.

apps/web/src/app/auth/core/services/change-password/web-change-password.service.ts
JaredSnider-Bitwarden, line 1: Very glad to see this service API updated to properly represent that we only support user key rotation on web currently. Thank you!

libs/common/src/auth/models/request/password.request.ts
JaredSnider-Bitwarden: Per discussion with Thomas, just going to keep sending `currentMasterPasswordAuthenticationHash` for authorization for now. We can change request models in the future to send the full `currentMasterPasswordAuthenticationData` if there is a benefit.

libs/angular/src/auth/password-management/change-password/change-password.component.ts
rr-bw: https://github.com/bitwarden/clients/pull/18507/changes/98cb16faa10c5ae596db8385f7ef81c4d3974ba6

libs/angular/src/auth/password-management/change-password/change-password.component.ts
rr-bw: https://github.com/bitwarden/clients/pull/18507/changes/36a5ba3cd26a1e0a94858e083eccd660d00cf37a

libs/angular/src/auth/password-management/change-password/default-change-password.service.ts
rr-bw: https://github.com/bitwarden/clients/pull/18507/changes/ff0de316ef4a77fdd012b2966ae5c9b92f5d9568

libs/angular/src/auth/password-management/change-password/default-change-password.service.ts
rr-bw: https://github.com/bitwarden/clients/pull/18507/changes/ff0de316ef4a77fdd012b2966ae5c9b92f5d9568

apps/browser/src/auth/popup/change-password/extension-change-password.service.ts
rr-bw: I tested behavior and updated the comment for accuracy here: https://github.com/bitwarden/clients/pull/18507/changes/b8da552610bab3210e21467bfb17c56f287dfcfb

That blank open popout that we discussed [here](https://github.com/bitwarden/clients/pull/18204#discussion_r2849486840) is handled on the `ChangePasswordComponent` by the already-existing `closeBrowserExtensionPopout()` method.

JaredSnider-Bitwarden:
@rr-bw , double checked with KM, no KM files changed so we don't have to await a review from them.

PR #18509, opened 2026-01-22
Feature/pm 31157 no new send text

iivins-livefront:
## Objective

Make sure the Error states have the correct copy / button

## Key Changes

Instead of adding copy for newSends I just used the existing <tools-new-send-dropdown-v2/> component - this resolved the text not coming through

## Screenshots
Screenshot 2026-01-22 at 4 03 49 PM
Screenshot 2026-01-22 at 4 04 01 PM

### DIFF
--- a/.github/workflows/repository-management.yml
+++ b/.github/workflows/repository-management.yml
@@ -72,7 +72,6 @@ jobs:
     permissions:
       id-token: write
       contents: write
-      pull-requests: write
 
     steps:
       - name: Validate version input format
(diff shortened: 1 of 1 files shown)


Hinton:
We don't want to use the dropdown for this milestone since there is no difference between text and file.

Alternative implementation https://github.com/bitwarden/clients/pull/18523

iivins-livefront:
Closing this PR in lieu of  Alternative implementation https://github.com/bitwarden/clients/pull/18523

PR #18510, opened 2026-01-22
[PM-26342] members page does not default all button

JaredScar:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-26342

## 📔 Objective

Fixes the members page initial load not having the all bit-toggle selected


JaredScar:
> We generally add some testing artifacts to the PR. Can you provide a simple video recording? Otherwise, approved.

Probably abandoning this since it was deemed unnecessary due to the new members page coming in but thanks haha!

PR #18512, opened 2026-01-22
[PM-31192] - [Defect] Deleting item -Scroll does not persist when returning to Vault tab

jaasen-livefront:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-31192
## 📔 Objective

This PR overrides an undesired restoring of scroll position after a navigation change in firefox.

## 📸 Screenshots

https://github.com/user-attachments/assets/2d9a3f4d-f279-4b27-a5ac-866ca7f9ff2a

### DIFF
--- a/apps/browser/src/vault/popup/components/vault-v2/view-v2/view-v2.component.spec.ts
+++ b/apps/browser/src/vault/popup/components/vault-v2/view-v2/view-v2.component.spec.ts
@@ -531,10 +531,10 @@ describe("ViewV2Component", () => {
       expect(openSimpleDialog).toHaveBeenCalledTimes(1);
     });
 
-    it("navigates back", async () => {
+    it("navigates to vault", async () => {
       await component.delete();
 
-      expect(back).toHaveBeenCalledTimes(1);
+      expect(mockNavigate).toHaveBeenCalledWith(["/tabs/vault"]);
     });
 
     it("stops scroll position service", async () => {


apps/browser/src/vault/popup/services/vault-popup-scroll-position.service.ts
nick-livefront, line 55: The original intention of the `reset` parameter was to force the scroll position back to the top, I came across [scrollRestoration](https://developer.mozilla.org/en-US/docs/Web/API/History/scrollRestoration) which can disable this browser behavior. My initial testing shows this works if it is set to `manual` when `reset` is passed.

It's possible we may need to revert the `manual` though.

apps/browser/src/vault/popup/components/vault-v2/view-v2/view-v2.component.ts
```diff
@@ -265,8 +265,9 @@ export class ViewV2Component {
       return false;
     }
 
+    // Navigate directly to vault instead of back() to avoid Firefox bfcache scroll restoration
     this.popupScrollPositionService.stop(true);
-    await this.popupRouterCacheService.back();
+    await this.router.navigate(["/tabs/vault"]);
```
nick-livefront, line 270: ❌ Unfortunately this won't work for user scenarios that navigate to the `/view` from other screens than the vault. Like the trash or archive screens.

jaasen-livefront: @nick-livefront Maybe we only do this if the previous route was the vault. Hacky though.

PR #18517, opened 2026-01-23
[PM-30808] Migrate Phishing Detection storage to PhishingIndexedDbService

Banrion:
## 🎟️ Tracking

[PM-30808](https://bitwarden.atlassian.net/browse/PM-30808)

Note: Branch name is incorrect for the pm number

## 📔 Objective

This pull request migrates phishing URL storage to use the new **PhishingIndexedDbService** with **IndexedDB** instead of previous storage mechanisms.

The migration enables the phishing detection feature to handle much larger datasets efficiently while maintaining fast lookup performance through IndexedDB's indexed storage.

Key Changes:
- Integrated **PhishingIndexedDbService** into **PhishingDataService** to store phishing URLs in IndexedDB instead of compressed state storage
- Added **addUrls()** method to **PhishingIndexedDbService** to support daily incremental updates that append new phishing URLs without clearing existing data
- Removed obsolete compression code that was previously used to work around storage size limitations

Additional Changes:
- Updated test suites across both services to validate the new IndexedDB storage layer and the distinction between appending and replacing URL data
- Refactored the background update mechanism to use RxJS observables with proper subject triggers, improving the reactive data flow and allowing for better control of update timing.

Future Improvement Idea:
- Use [fake-indexeddb](https://www.npmjs.com/package/fake-indexeddb) within test suites

Additional Notes:
- Loading the full list into IndexedDb takes about 10 minutes but performance showed no signs of degradation while loading
- Daily list is appended when the full list is loaded or via the 24 hour scheduler

## 📸 Screenshots

Indexed DB with Data
image

Blocked DANGEROUS website as per Phishing.Database stored in IndexedDB
image

Blocked `phishing.testcategory.com` website
https://github.com/user-attachments/assets/7ce01690-df98-4e12-8bae-c11e114bdd65

[PM-30808]: https://bitwarden.atlassian.net/browse/PM-30808?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/browser/src/dirt/phishing-detection/services/phishing-data.service.ts
```diff
@@ -148,105 +153,65 @@ export class PhishingDataService {
...
 
-    // Custom matcher per resource
-    if (resource && resource?.match) {
-      for (const entry of set) {
-        if (resource.match(url, entry)) {
-          return true;
-        }
+    try {
+      // Quick lookup: check direct presence of hostname in IndexedDB
+      const hasUrl = await this.indexedDbService.hasUrl(url.hostname);
```
AlexRubik: We want to check the entire url/href.
```suggestion
      const hasUrl = await this.indexedDbService.hasUrl(url.href);
```
Here's the url obj:
image

Claude suggests normalizing the full url before the check like this but thats up to you, not necessarily a blocker . Olena and I have seen edge cases where the browser will automatically convert a http to https but our phish url list only has the http string and therefore doesn't block the site.

```suggestion

  // Strip protocol and trailing slash for consistent matching
  const normalizedUrl = url.href.toLowerCase().replace(/^https?:\/\//, "").replace(/\/$/, "");
  return await this.indexedDbService.hasUrl(normalizedUrl);
```

apps/browser/src/dirt/phishing-detection/services/phishing-data.service.ts
```diff
@@ -265,164 +230,147 @@ export class PhishingDataService {
...
+      }),
+      catchError((err: unknown) => {
+        this.logService.error(
+          `[PhishingDataService] Full dataset update failed using primary source ${err}`,
+        );
+        this.logService.warning(
+          `[PhishingDataService] Falling back to: ${resource.fallbackUrl} (Note: Fallback data may be less up-to-date)`,
+        );
+        // Try fallback URL
+        return from(this.apiService.nativeFetch(new Request(resource.fallbackUrl))).pipe(
```
AlexRubik, line 310: Process and save urls to db for fallback the same way we do it for primary url.

```suggestion
        // Try fallback URL
        return from(this.apiService.nativeFetch(new Request(resource.fallbackUrl))).pipe(
          switchMap((fallbackResponse) => {
            if (!fallbackResponse.ok || !fallbackResponse.body) {
              return throwError(
                () => new Error(`Fallback fetch failed: ${fallbackResponse.status}`),
              );
            }
            return from(this.indexedDbService.saveUrlsFromStream(fallbackResponse.body));
          }),
```

apps/browser/src/dirt/phishing-detection/services/phishing-data.service.ts
Banrion: Good catch on the href/hostname issue. I've addressed it. 

In regards to normalizing, the hostname is not case senstive, but most if not all of the rest is case sensitive. I will be following up on the matchers we are using for links but I think it best that we keep the one to one match for quick lookup. Most of the time, phishing links will be access by clicking on a link. These links will be submitted to Phishing.Database. They have their own checks and use things such as wildcards. We can iterate to match more closely if needed. 

Also, the Phishing.Database source file does contain both **http** and **https** links.

apps/browser/src/dirt/phishing-detection/services/phishing-data.service.ts
Banrion: Thanks, I don't know how I missed that!

apps/browser/src/dirt/phishing-detection/services/phishing-data.service.ts
```diff
@@ -258,171 +223,184 @@ export class PhishingDataService {
...
+            return timer(5 * 60 * 1000); // Wait 5 mins before next attempt
+          },
+        }),
+        catchError((err: unknown) => {
+          const elapsed = Date.now() - startTime;
+          this.logService.error(
+            `[PhishingDataService] Retries unsuccessful after ${elapsed}ms.`,
+            err,
+          );
+          return throwError(() => err);
```
AlexRubik: Return EMPTY so the _backgroundUpdateTrigger$ subscription doesn't die after exhausting retries. Without this, an unhandled error would terminate the subscription and future 24-hour scheduled updates would never run.

```suggestion
// Return EMPTY to complete this update attempt but keep the subscription alive for future scheduled updates

          return EMPTY;
```

image

Banrion: 👍  Returned previous meta for state consistency instead of a silent fail.

PR #18518, opened 2026-01-23
[PM-31173] [PM-1620] Add Chrome-like SPA navigation detection for credential saving

### LINKED ISSUE #1620: Bitwarden does not prompt to save or update credentials
## Describe the Bug

We are aware of many sites' login forms where the Bitwarden browser extension either on a single platform/browser or multiple will not prompt the user to save a new password or update an existing password.

## Expected Result

When you enter new credentials into a login form, or register for a new user account on a site, you should be automatically prompted to save those credentials upon submission by the Bitwarden browser extension. When you update your password on a site and you have a matching login in your vault, you should be automatically prompted to update those credentials upon form submission.

## Actual Result

This works sometimes, but not always and it can depend on _many, many_ factors. Some sites simply don't work, some don't work consistently and others who knows. This is what we're aiming to fix!

## Call to Action

Below is a link to a Google Form that we have created for capturing information that will help us track down all of these login pages and registration forms where Bitwarden has failed to prompt you to save those credentials. You may submit as many as you like, just please help us help you and the community by providing meaningful information and only those sites/pages where it legitimately doesn't work.

[Report failure of prompt to save credentials](https://forms.gle/tJXZSr4WwDqjq6vv5)

If you've been sent to this issue because another issue you had open was closed as a duplicate, thank you for taking the time to submit the issue to us, however we truly need to track this under a single item in aggregate fashion so we can manage it holistically and ensure we're taking broad measures that will improve this behavior with the maximum benefit to all. Please consider using the linked form above to submit the prior sites you had reported in the issue closed.

kovan:
## Summary

Implements Chrome's approach to detecting form submissions in Single Page Applications, addressing issue #1620 where Bitwarden fails to prompt users to save credentials on many sites.

Fixes #1620

## Changes

- Intercept `history.pushState`/`replaceState` to detect SPA navigation
- Add MutationObserver to detect when password fields are removed from DOM
- Trigger credential save prompt on visibility change when credentials are filled
- Add `beforeunload` handler to capture credentials before page unload

## How Chrome Does It

Chrome monitors for History API changes (pushState/replaceState) and checks if the password form was removed from the DOM. This is critical for SPAs that don't trigger traditional page navigation.

## Test plan

- [ ] Test on SPA sites (React, Angular, Vue apps) that previously failed to trigger save prompts
- [ ] Verify credential save prompt appears after successful login on SPAs
- [ ] Verify prompt appears when switching tabs after filling credentials
- [ ] Verify no duplicate prompts are triggered
- [ ] Run existing autofill tests (all passing)

### DIFF
--- a/apps/browser/src/autofill/services/autofill-overlay-content.service.spec.ts
+++ b/apps/browser/src/autofill/services/autofill-overlay-content.service.spec.ts
@@ -2737,5 +2737,181 @@ describe("AutofillOverlayContentService", () => {
 
       expect(autofillOverlayContentService["userFilledFields"]).toEqual(null);
     });
+
+    it("restores History API methods and removes event listeners", () => {
+      autofillOverlayContentService.init();
+
+      autofillOverlayContentService.destroy();
+
+      expect(globalThis.removeEventListener).toHaveBeenCalledWith(
+        "beforeunload",
+        autofillOverlayContentService["handleBeforeUnloadEvent"],
+      );
+      expect(globalThis.removeEventListener).toHaveBeenCalledWith(
+        "popstate",
+        autofillOverlayContentService["handlePopStateEvent"],
+      );
+    });
+
+    it("disconnects the password field removal observer", () => {
+      autofillOverlayContentService.init();
+      const mockObserver = { disconnect: jest.fn() };
+      autofillOverlayContentService["passwordFieldRemovalObserver"] =
+        mockObserver as unknown as MutationObserver;
+
+      autofillOverlayContentService.destroy();
+
+      expect(mockObserver.disconnect).toHaveBeenCalled();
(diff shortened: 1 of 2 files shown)


blackwood:
Thanks for taking the time to put this together and for surfacing the issue—improvements to SPA form submission detection is something we've had on our radar, and community PRs like this help validate that it's worth prioritizing.

After review, we've decided to move forward with an internal implementation that takes a somewhat different approach (leaning on native browser APIs where available rather than history monkey-patching), so we won't be merging this PR as-is. That said, this submission genuinely helped inform our thinking on the problem space, and we appreciate the effort.

We'll be tracking the work on our end—thanks again for the contribution!

kovan:
Ok, glad to be of help.

PR #18523, opened 2026-01-23
[PM-31157] Use correct key for new send

Hinton:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-31157
## 📔 Objective

Use the correct i18n key
## 📸 Screenshots

image


iivins-livefront:
Makes sense to me 👍

PR #18526, opened 2026-01-23
[PM-31158] Change default send to file for old ui

Hinton:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-31158
## 📔 Objective

The old UI defaulted to files, this also resolves the issue of the mode thinking it's in text when it's in file.


djsmith85:
@Hinton I'm, not sure that is correct, as files are a premium feature and text is always available.

Hinton:
@djsmith85 I thought the same thing, but premium start as files and non premium fallback to text. This does the same thing from my testing.

PR #18529, opened 2026-01-23
Feature/pm 30737 migrate delete account

iivins-livefront:
##  Tracking

https://bitwarden.atlassian.net/browse/PM-30737

##  Objective

The goal of this ticket is to have the Desktop application stop using a deprecated 'UserVerificationComponent' and begin using the 'UserVerificationFormInputComponent' instead

##  Key Changes
- Mostly a clean swap but it's worth noting the Error handling / Toast Alerting was touched
- Previous implementation had errors from accountApiService.deleteAccount() bubbled up unhandled, this is now managed within a try/catch and directly handled
- Added a locale from web for failed auth
- Added test coverage for sanity

## Screenshots
Screenshot 2026-01-23 at 11 43 32 AM
Screenshot 2026-01-23 at 11 44 02 AM


apps/desktop/src/auth/delete-account.component.ts
```diff
@@ -35,19 +38,36 @@ import { UserVerificationComponent } from "../app/components/user-verification.c
...
     private accountApiService: AccountApiService,
     private toastService: ToastService,
+    private configService: ConfigService,
   ) {}
 
+  async ngOnInit() {
+    this.migrationMilestone4 = await this.configService.getFeatureFlag(
+      FeatureFlag.DesktopUiMigrationMilestone4,
+    );
+  }
```
Hinton: **suggestion:** Can we just use the observable and `toSignal`?

apps/desktop/src/auth/delete-account.component.ts
```diff
@@ -57,12 +77,26 @@ export class DeleteAccountComponent {
...
+      });
+    } catch {
+      if (this.migrationMilestone4) {
+        this.invalidSecret = true;
+        this.toastService.showToast({
+          variant: "error",
+          title: this.i18nService.t("errorOccurred"),
+          message: this.i18nService.t("userVerificationFailed"),
+        });
+      }
```
Hinton: Async actions have built in support for errors, https://github.com/bitwarden/clients/blob/main/libs/common/src/platform/services/validation.service.ts#L12-L35. I think if we catch it here we also need to show an error for non milestone 4 users. But I suspect we might not need to.

apps/desktop/src/auth/delete-account.component.ts
```diff
@@ -35,19 +38,36 @@ import { UserVerificationComponent } from "../app/components/user-verification.c
...
+export class DeleteAccountComponent implements OnInit {
   deleteForm = this.formBuilder.group({
     verification: undefined as VerificationWithSecret | undefined,
   });
 
+  /**
+   * Tracks whether the verification failed due to invalid credentials.
+   * Used to show inline error messages in the verification component.
+   */
+  protected invalidSecret = false;
```
Hinton: **suggestion:** Signal?

apps/desktop/src/auth/delete-account.component.ts
rr-bw: @Hinton I ran a local test to check this and it does look like non milestone 4 users would no longer see a toast now with the `try/catch` as it currently is.

But since we need the `try/catch` here to handle the inline error state (`invalidSecret = true`), I think that means we'll need to keep the `showToast()` in this catch as well.

@iivins-livefront Maybe just place the `showToast()` outside the flag check?

apps/desktop/src/auth/delete-account.component.html
```diff
@@ -6,14 +6,22 @@
...
+        <div class="tw-mb-2 tw-rounded tw-border tw-border-solid tw-border-info-600/25">
+          <app-user-verification
+            ngDefaultControl
+            formControlName="verification"
+            name="verification"
+          ></app-user-verification>
+        </div>
+      }
       <div id="confirmIdentityHelp">
         <p>{{ "confirmIdentity" | i18n }}</p>
```
rr-bw: This "Confirm your identity" becomes redundant with the flag on.

Image

apps/desktop/src/auth/delete-account.component.html
```diff
@@ -6,14 +6,22 @@
       <bit-callout type="warning" title="{{ 'warning' | i18n }}">
```
rr-bw, line 6: @mblabaum-designer I noticed [the design](https://www.figma.com/design/9cGXB72BSoSjQpsrVnjdrF/Desktop-2025?node-id=529-59205&m=dev) for this uses "danger" instead of "warning" for the `bit-callout`, and the design also has different copy for the callout, as well as NO text above the copy.

_However_, I also noticed that [Web uses the same pattern](https://github.com/bitwarden/clients/blob/main/apps/web/src/app/auth/settings/account/delete-account-dialog.component.html#L4-L5) as Desktop here ("warning" and same copy). Would you like any changes to me made to Desktop here in this ticket? Or make a separate ticket to make the updates on both Web/Desktop at the same time so that the changes ship together (i.e. so we don't have different designs per client)?

(Below is a screenshot of desktop)

Image

apps/desktop/src/auth/delete-account.component.ts
Hinton: Either that or we can omit the catch completely and let the validator handle it. (Assuming the error it produces makes sense)

iivins-livefront:
Quick recordings to show the Error state UI for:
- main as of writing this comment
- this branch with the Feature Flag enabled
- this branch with the Feature Flag disabled

Open Delete Account modal -> type incorrect password -> click CTA to submit -> see Error state

https://github.com/user-attachments/assets/a9ddb8dc-6615-4e0f-b87c-f0dc5321003b

https://github.com/user-attachments/assets/35518035-87f4-488b-87a1-21e49be53270

https://github.com/user-attachments/assets/a9a4a9bf-7668-4fe3-9246-fd9716a43340

apps/desktop/src/auth/delete-account.component.html
iivins-livefront: @rr-bw After surfacing this in Slack

> Decision point: Let's align with the web flow for now

Looks like we want to re-ticket it and stick with parity between Desktop and Web for now

apps/desktop/src/auth/delete-account.component.ts
```diff
@@ -57,12 +72,32 @@ export class DeleteAccountComponent {
...
-      message: this.i18nService.t("accountDeletedDesc"),
-    });
+    if (!this.migrationMilestone4()) {
+      const verification = this.deleteForm.get("verification").value;
+      await this.accountApiService.deleteAccount(verification);
+      this.toastService.showToast({
+        variant: "success",
+        title: this.i18nService.t("accountDeleted"),
+        message: this.i18nService.t("accountDeletedDesc"),
+      });
```
JaredSnider-Bitwarden, line 85: We don't close this dialog in success scenarios when we should. 
Please inject `private dialogRef = inject(DialogRef);` and add a `this.dialogRef.close()` after this success toast.

iivins-livefront: Just to confirm - this block has the previous / old logic in place, do we need to add `this.dialogRef.close()` to the block with our Feature Flag enabled?

JaredSnider-Bitwarden: I'm sorry, yes, only in the new flagged code. I misread the conditional (missed the !).

rr-bw (dismissed):

Hinton:
Superseded by https://github.com/bitwarden/clients/pull/19308

PR #18530, opened 2026-01-23
[PM-31188] Desktop Trash Items Context Menu Updates

Jingo88:
## 🎟️ Tracking

[PM-31188](https://bitwarden.atlassian.net/browse/PM-31188)

## 📔 Objective

An item that is in desktop trash should only not apply options such as `copy password` and `copy username`. Added a `isDeleted` check to the other options in the menu (matching the previously applied checks to Archive and Unarchive).

## 📸 Screenshots

Screenshot 2026-01-23 at 12 23 09 PM

Screenshot 2026-01-23 at 12 22 59 PM

Screenshot 2026-01-23 at 12 22 48 PM

[PM-31188]: https://bitwarden.atlassian.net/browse/PM-31188?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/desktop/src/vault/app/vault/vault-v2.component.ts
+++ b/apps/desktop/src/vault/app/vault/vault-v2.component.ts
@@ -642,83 +642,80 @@ export class VaultV2Component<C extends CipherViewLike>
       });
     }
 
-    switch (cipher.type) {
-      case CipherType.Login:
-        if (
-          cipher.login.canLaunch ||
-          cipher.login.username != null ||
-          cipher.login.password != null
-        ) {
-          menu.push({ type: "separator" });
-        }
-        if (cipher.isDeleted) {
+    if (!cipher.isDeleted) {
+      switch (cipher.type) {
+        case CipherType.Login:
+          if (
+            cipher.login.canLaunch ||
+            cipher.login.username != null ||
+            cipher.login.password != null
+          ) {
+            menu.push({ type: "separator" });
+          }
+          if (cipher.login.canLaunch) {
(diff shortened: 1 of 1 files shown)


apps/desktop/src/vault/app/vault/vault-v2.component.ts
```diff
@@ -651,6 +651,9 @@ export class VaultV2Component<C extends CipherViewLike>
         ) {
           menu.push({ type: "separator" });
         }
+        if (cipher.isDeleted) {
```
shane-melton: ❓ This is only hiding the copy options for `CipherType.Login`, what about the identity and card types? Those will still show copy actions, no?

Jingo88: applied delete check to card as well. Identity types do not have options in context menu.

apps/desktop/src/vault/app/vault/vault-v2.component.ts
```diff
@@ -651,6 +651,9 @@ export class VaultV2Component<C extends CipherViewLike>
         ) {
           menu.push({ type: "separator" });
```
shane-melton: ⛏️ This is going to put a "separator" in the menu without any subsequent items when the cipher is deleted and has a username/password. 

Maybe we can just skip the entire `switch` statement when `cipher.isDeleted`?

PR #18536, opened 2026-01-23
[PM-30638] Cipher Add/Edit dialog focus

nick-livefront:
## 🎟️ Tracking

[PM-30638](https://bitwarden.atlassian.net/browse/PM-30638)

### Issue 

When switching between the view and form views of the add/edit dialog focus is lost because the button that was pressed is removed from the DOM. The focus returns to the body element and outside of the dialog. 
- The focus when the dialog is handled internally but updating the dialog content is not accounted for.

### Changes

- Expose `focusOnHeader` from the `DialogComponent` to allow child components to refocus on the heading element.
  - @bitwarden/team-ui-foundation I wanted to keep this as close to the initial focus behavior but I had to expose the underlying `DialogComponent` to use with `viewChild`. I could see why we _wouldn't_ want to do this so if there are any ideas here I'm open!
- Call `focusOnHeader` when ever the mode changes in the add edit component. 
- Resolve FIXMEs for `viewChild`

## 📸 Screenshots

<video src="https://github.com/user-attachments/assets/d3a77274-527c-421c-8573-43842732ed50" />

[PM-30638]: https://bitwarden.atlassian.net/browse/PM-30638?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/web/src/app/vault/components/vault-item-dialog/vault-item-dialog.component.ts
+++ b/apps/web/src/app/vault/components/vault-item-dialog/vault-item-dialog.component.ts
@@ -9,7 +9,6 @@ import {
   OnDestroy,
   OnInit,
   viewChild,
-  ViewChild,
 } from "@angular/core";
 import { takeUntilDestroyed } from "@angular/core/rxjs-interop";
 import { Router } from "@angular/router";
@@ -165,14 +164,9 @@ export class VaultItemDialogComponent implements OnInit, OnDestroy {
    * Reference to the dialog content element. Used to scroll to the top of the dialog when switching modes.
    * @protected
    */
-  // FIXME(https://bitwarden.atlassian.net/browse/CL-903): Migrate to Signals
-  // eslint-disable-next-line @angular-eslint/prefer-signals
-  @ViewChild("dialogContent")
-  protected dialogContent: ElementRef<HTMLElement>;
+  protected readonly dialogContent = viewChild.required<ElementRef<HTMLElement>>("dialogContent");
 
-  // FIXME(https://bitwarden.atlassian.net/browse/CL-903): Migrate to Signals
-  // eslint-disable-next-line @angular-eslint/prefer-signals
-  @ViewChild(CipherFormComponent) cipherFormComponent!: CipherFormComponent;
+  private readonly cipherFormComponent = viewChild.required(CipherFormComponent);
 
   private readonly dialogComponent = viewChild(DialogComponent);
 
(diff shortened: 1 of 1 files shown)


nick-livefront:
> I wonder if this maybe belongs in the service instead but, fine for now!

@BryanCunningham Which service did you have in mind? I can explore that integration here rather than creating another TODO

BryanCunningham:
> > I wonder if this maybe belongs in the service instead but, fine for now!
> 
> @BryanCunningham Which service did you have in mind? I can explore that integration here rather than creating another TODO

Nah, it's fine here currently. There is a `dialog.service`. But, this functionality was already in the component so, can leave it here for now. Was mostly just wondering out loud with that comment :)

apps/web/src/app/vault/components/vault-item-dialog/vault-item-dialog.component.ts
```diff
@@ -163,14 +164,11 @@ export class VaultItemDialogComponent implements OnInit, OnDestroy {
    * Reference to the dialog content element. Used to scroll to the top of the dialog when switching modes.
    * @protected
    */
-  // FIXME(https://bitwarden.atlassian.net/browse/CL-903): Migrate to Signals
-  // eslint-disable-next-line @angular-eslint/prefer-signals
-  @ViewChild("dialogContent")
-  protected dialogContent: ElementRef<HTMLElement>;
+  protected readonly dialogContent = viewChild.required<ElementRef<HTMLElement>>("dialogContent");
```
shane-melton, line 167: 👍 Thanks for fixing these!

libs/components/src/dialog/dialog/dialog.component.ts
```diff
@@ -145,6 +145,26 @@ export class DialogComponent implements AfterViewInit {
...
+  onAnimationEnd() {
+    this.animationCompleted.set(true);
+  }
+
+  /**
+   * Moves focus to the dialog header element.
+   * This is done automatically when the dialog is opened but can be called manually
+   * when the contents of the dialog change and focus should be reset.
+   */
+  focusOnHeader(): void {
```
vleague2, line 167: 👋 Hi hi! I am splitting out this function in #18657 to fix a regression and will make it public for use here. Do you mind waiting until that goes in, and then you won't need to make any changes to this file!

vleague2: Actually, I think your ticket has already passed QA, so I can just handle the conflicts on my PR once yours is merged!

PR #18541, opened 2026-01-23
Added devcontainer setup (devcontainer.json, docker-compose.yml, postCreateCommand.sh)

connerbw:
## 📔 Objective

Adds devcontainer configuration for the clients repository to enable consistent development environments in VS Code Dev Containers.

  - Adds devcontainer.json with VS Code extensions (Angular, Jest, Prettier, ESLint, Nx Console, Tailwind)
  - Adds docker-compose.yml using the TypeScript/Node 22 dev container image
  - Adds postCreateCommand.sh that installs mkcert for SSL certificates (needed for WebAuthn), generates localhost certs, and runs npm ci
  - Configures port forwarding for Web Vault (8080), Web Vault Self-hosted (8081), and Storybook (6006)


AmyLGalles:
@cturnbull-bitwarden have you reviewed the chexmarks failures on this?

connerbw:
> @cturnbull-bitwarden have you reviewed the chexmarks failures on this?

 @AmyLGalles thanks for calling that out, just reviewed them. I pushed https://github.com/bitwarden/clients/pull/18541/commits/9aa03c7d40e1a64e8828c906e0177a99b029b6b6 to exclude the `.devcontainer` folder from the scan since dev containers are localhost development tooling and aren't used in production. I found that the existing dev containers on server are likewise excluded here: https://github.com/bitwarden/server/blob/main/.checkmarx/config.yml#L13. Let me know if you've got any concerns there, though

withinfocus:
Why'd the Checkmarx config get reverted? I thought it was a good idea actually and did the same at https://github.com/bitwarden/server/pull/6912.

connerbw:
> Why'd the Checkmarx config get reverted? I thought it was a good idea actually and did the same at [bitwarden/server#6912](https://github.com/bitwarden/server/pull/6912).

@withinfocus I was still in the process of doing discovery on where these findings were coming from, and thought that `sca` and `containers` could have something to do with it, but I wasn't entirely sure. At the time, only the `kics` filter was present in the server repo, so I removed them again, simply due to my own uncertainty and not wanting to commit to something to main I was unsure of 😄. I went through the Checkmarx findings and marked them all as 'won't fix', which resolved that action's check. I'm happy to bring those filters back in, though.

.devcontainer/common/post-create-command.sh
```diff
@@ -0,0 +1,79 @@
...
+    echo "Installing mkcert..."
+    curl -JLO "https://dl.filippo.io/mkcert/latest?for=linux/amd64"
+    chmod +x mkcert-v*-linux-amd64
+    sudo mv mkcert-v*-linux-amd64 /usr/local/bin/mkcert
+
+    # Generate SSL certificates for localhost
+    echo "Generating SSL certificates..."
+    mkcert -install
+    cd /workspace/apps/web
+    mkcert -cert-file dev-server.local.pem -key-file dev-server.local.pem localhost bitwarden.test
```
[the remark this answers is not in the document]
connerbw: This is a false positive. When `-cert-file` and `-key-file` are the same, mkcert [explicitly handles this case](https://github.com/FiloSottile/mkcert/blob/master/cert.go#L115-L117) by concatenating both into a single write:

```go
if certFile == keyFile {
    err = ioutil.WriteFile(keyFile, append(certPEM, privPEM...), 0600)
}
```

Verified locally - the command produces a valid combined PEM file containing both the certificate and private key.

connerbw:
@withinfocus @AmyLGalles this PR adding dev container support to clients is ready for review if you both have a sec

.devcontainer/common/post-create-command.sh
```diff
@@ -0,0 +1,79 @@
...
+# Configure git safe directory
+git config --global --add safe.directory /workspace
+
+echo "Installing system dependencies..."
+# Packages needed:
+#   libnss3-tools: mkcert certificate installation
+#   build-essential, pkg-config: native module compilation
+#   libsecret-1-dev, libglib2.0-dev: desktop native module dependencies
+#   Remaining packages: Electron runtime dependencies
+sudo apt-get update && sudo apt-get install -y \
```
withinfocus, line 21: ⚠️ This is pretty lengthy and therefore a large attack surface to protect.

withinfocus (approved):
A somewhat-opinionated setup but I hear it's useful for some that aren't standing up the whole platform regularly.

.devcontainer/common/post-create-command.sh
connerbw: @withinfocus Good callout. I'll send you a followup PR aiming at addressing this

PR #18547, opened 2026-01-24
Dirt/indexeddb improvements

AlexRubik:
## Problem

Opening non-HTTP URLs like `chrome://newtab/` triggers the phishing check fallback path, which loads all URLs into memory (~10+ seconds), iterates through each one with the custom matcher, and then garbage collects after function returns. **This blocks all subsequent phishing checks due to sequential `concatMap` processing.**

## Solution

1. **Protocol filter**: Early return for non-http(s) URLs before any IndexedDB operations so that long process doesn't happen when a new chrome tab is opened
2. **Parallel processing**: Replace `concatMap` with `mergeMap(fn, 5)` so tab checks run concurrently instead of sequentially

This ensures new tabs are instant and one slow check doesn't block others.


apps/browser/src/dirt/phishing-detection/services/phishing-detection.service.ts
```diff
@@ -88,7 +88,9 @@ export class PhishingDetectionService {
           prev.ignored === curr.ignored,
       ),
       tap((event) => logService.debug(`[PhishingDetectionService] processing event:`, event)),
-      concatMap(async ({ tabId, url, ignored }) => {
+      // Use mergeMap for parallel processing - each tab check runs independently
+      // Concurrency limit of 5 prevents overwhelming IndexedDB
+      mergeMap(async ({ tabId, url, ignored }) => {
```
Banrion, line 93: ⚠️ I'm not sure this is the correct way to solve this problem. As previously mentioned, the cursor was implemented in a way that still loads all urls into memory and then checks for the phishing url. With this change using the mergeMap, this will cause an even greater memory load.

Please see pull request [18561](https://github.com/bitwarden/clients/pull/18561) for updating the search to use the cursor method.

PR #18548, opened 2026-01-24
[PM-28190] Migrate clients to use SDK for cipher sharing operations.

nikwithak:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-28190

## 📔 Objective

We want to replace the existing operations in cipher.service.ts with these new SDK operations, allowing the SDK to fully manage cipher sharing instead of calling the APIs directly from the client. This introduces a new feature flag and migrates the logic to use the SDK instead of the TS files when it is enabled.


libs/common/src/vault/services/cipher-sdk.service.ts
```diff
@@ -269,6 +269,81 @@ export class DefaultCipherSdkService implements CipherSdkService {
     );
   }
 
+  async shareWithServer(
+    cipherView: CipherView,
+    organizationId: OrganizationId,
+    collectionIds: CollectionId[],
+    userId: UserId,
+    _originalCipherView?: CipherView,
```
gbubemismith: ⛏️  for For optional parameters, the standard `paramName?: Type` syntax is sufficient  
```suggestion
    originalCipherView?: CipherView,
```

gbubemismith (changes requested):

libs/common/src/vault/services/cipher-sdk.service.ts
nikwithak: I updated it - it wasn't being passed through to the SDK, and I don't remember why that was in the first place.

I ended up changing the SDK interface here: https://github.com/bitwarden/sdk-internal/pull/822 - now it deals only with `CipherView` instead of a mix of `Cipher` and `CipherView`. Along with renaming it, I (a) updated to use the new changes in that PR, and (b) passed the `originalCipherView` through to the SDK instead of swallowing it.

Builds will fail until the SDK changes are in, but LMK if you have any concerns with this approach

gbubemismith:
@nikwithak looks like there are some dependency issues

nikwithak:
> looks like there are some dependency issues

@gbubemismith This is waiting on https://github.com/bitwarden/sdk-internal/pull/822 - I'll move this to draft until that one merges in and I can consume the changes.

gbubemismith (approved):

PR #18555, opened 2026-01-26
[PM-30404] Exclude .wasm.js from babel

coroiu:
## 📔 Objective

Debug SDK builds create > 30MB `.js` files which causes babel to consume huge amounts of memory. This PR excludes these files completely, causing them to be used as-is. Release builds of the SDK are much smaller so the gain is limited on that side.

According to Claude this fix is safe because:

  1. wasm-bindgen targets ES2015 - Same baseline as WebAssembly itself
  2. All WASM browsers support ES2015 - Chrome 57+, Firefox 52+, Safari 11+, Edge 16+
  3. No Babel transformations present - Verified by the test script
  4. Industry standard - Most projects don't transform WASM glue code

### Effects on CI

Seems to speed up docker build by ~10% (roughly 1 minute)

**Web build on a `main` commit**
https://actions-profiler.1b.nu/analyze?share=73f47ec0-cf65-4f4b-950f-e01e7eb01695

image

**Web build on this PR**
https://actions-profiler.1b.nu/analyze?share=4f83f2bd-83b8-4149-969c-12b234fcbced

image

### DIFF
--- a/apps/browser/webpack.base.js
+++ b/apps/browser/webpack.base.js
@@ -109,6 +109,7 @@ module.exports.buildConfig = function buildConfig(params) {
     },
     {
       test: /\.[cm]?js$/,
+      exclude: /\.wasm\.js$/,
       use: [
         {
           loader: "babel-loader",
--- a/apps/desktop/webpack.base.js
+++ b/apps/desktop/webpack.base.js
@@ -215,6 +215,7 @@ module.exports.buildConfig = function buildConfig(params) {
       rules: [
         {
           test: /\.[cm]?js$/,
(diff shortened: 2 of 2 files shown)


abergs:
That is a huge JS bundle. Nice. What's the impact on the local dev builds?

coroiu:
@abergs It longer crashes with an Out Of Memory exception lol

It's hard to say exactly since the times seem to fluctuate wildly, but here's one example:

```
With Babel
> Total   67.0621s

Excluded from Babel
> Total   44.4887s
```

PR #18561, opened 2026-01-26
[PM-31203] Change Phishing Url Check to use a Cursor Based Search

Banrion:
## 🎟️ Tracking

[PM-31203](https://bitwarden.atlassian.net/browse/PM-31203)

May fix [PM-31196)](https://bitwarden.atlassian.net/browse/PM-31196)

## 📔 Objective

This pull request adds a method to use a cursor based search rather than loading all urls into memory. It also updates usages and adds test cases.

Note: A follow up pull request will be made to remove obsolete methods.

[PM-31203]: https://bitwarden.atlassian.net/browse/PM-31203?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -142,7 +142,7 @@ jobs:
         run: cargo +nightly udeps --workspace --all-features --all-targets
 
       - name: Install cargo-deny
-        uses: taiki-e/install-action@2e9d707ef49c9b094d45955b60c7e5c0dfedeb14 # v2.66.5
+        uses: taiki-e/install-action@542cebaaed782771e619bd5609d97659d109c492 # v2.66.7
         with:
           tool: cargo-deny@0.18.6
 
--- a/apps/browser/src/_locales/en/messages.json
+++ b/apps/browser/src/_locales/en/messages.json
@@ -5001,6 +5001,9 @@
       }
     }
   },
+  "downloadAttachmentLabel": {
+    "message": "Download Attachment"
+  },
   "downloadBitwarden": {
     "message": "Download Bitwarden"
   },
--- a/apps/browser/src/autofill/background/notification.background.spec.ts
+++ b/apps/browser/src/autofill/background/notification.background.spec.ts
@@ -767,7 +767,6 @@ describe("NotificationBackground", () => {
         let createWithServerSpy: jest.SpyInstance;
         let updateWithServerSpy: jest.SpyInstance;
         let folderExistsSpy: jest.SpyInstance;
-        let cipherEncryptSpy: jest.SpyInstance;
 
         beforeEach(() => {
           activeAccountStatusMock$.next(AuthenticationStatus.Unlocked);
@@ -791,7 +790,6 @@ describe("NotificationBackground", () => {
           createWithServerSpy = jest.spyOn(cipherService, "createWithServer");
           updateWithServerSpy = jest.spyOn(cipherService, "updateWithServer");
           folderExistsSpy = jest.spyOn(notificationBackground as any, "folderExists");
-          cipherEncryptSpy = jest.spyOn(cipherService, "encrypt");
 
           accountService.activeAccount$ = activeAccountSubject;
         });
@@ -1190,13 +1188,7 @@ describe("NotificationBackground", () => {
           folderExistsSpy.mockResolvedValueOnce(false);
           convertAddLoginQueueMessageToCipherViewSpy.mockReturnValueOnce(cipherView);
           editItemSpy.mockResolvedValueOnce(undefined);
-          cipherEncryptSpy.mockResolvedValueOnce({
-            cipher: {
-              ...cipherView,
-              id: "testId",
-            },
-            encryptedFor: userId,
-          });
+          createWithServerSpy.mockResolvedValueOnce(cipherView);
 
           sendMockExtensionMessage(message, sender);
           await flushPromises();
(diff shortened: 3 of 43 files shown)


apps/browser/src/dirt/phishing-detection/services/phishing-data.service.spec.ts
```diff
@@ -120,70 +121,65 @@ describe("PhishingDataService", () => {
       expect(result).toBe(true);
       expect(mockIndexedDbService.hasUrl).toHaveBeenCalledWith("http://phish.com/testing-param");
       // Should not fall back to custom matcher when hasUrl returns true
-      expect(mockIndexedDbService.loadAllUrls).not.toHaveBeenCalled();
+      expect(mockIndexedDbService.findMatchingUrl).not.toHaveBeenCalled();
     });
 
-    it("should fall back to custom matcher when hasUrl returns false", async () => {
+    it("should return false when hasUrl returns false (custom matcher disabled)", async () => {
```
Banrion, line 127: ⚠️ Should add an alternative test for disabled custom matcher and leave the fallback test

apps/browser/src/dirt/phishing-detection/services/phishing-data.service.ts
```diff
@@ -153,37 +153,92 @@ export class PhishingDataService {
...
+      this.logService.info("[PhishingDataService] Found test web address: " + url.href);
       return true;
     }
 
     const resource = getPhishingResources(this.resourceType);
 
     try {
       // Quick lookup: check direct presence of href in IndexedDB
-      const hasUrl = await this.indexedDbService.hasUrl(url.href);
+      // Also check without trailing slash since browsers add it but DB entries may not have it
```
Banrion, line 175: ⚠️ We should confirm Phishing.Database's standards on the slash. Normalizing using new URL does add the slash to the base domain. I think Phishing.Database does this (and the check without the slash is unnecessary) but we can leave this and verify in a follow up.

apps/browser/src/dirt/phishing-detection/services/phishing-data.service.ts
```diff
@@ -153,37 +153,92 @@ export class PhishingDataService {
...
       }
     } catch (err) {
       this.logService.error("[PhishingDataService] IndexedDB lookup via hasUrl failed", err);
     }
 
-    // If a custom matcher is provided, iterate stored entries and apply the matcher.
-    if (resource && resource.match) {
+    // If a custom matcher is provided and enabled, use cursor-based search.
+    // This avoids loading all URLs into memory and allows early exit on first match.
+    // Can be disabled via useCustomMatcher: false for performance reasons.
```
Banrion, line 203: ⚠️ A better place for this is in the service rather than the resources. The resources could simply not have a matcher if it doesn't need one. The service should determine when to use custom matching regardless of the resource.

apps/browser/src/dirt/phishing-detection/services/phishing-data.service.ts
```diff
@@ -153,37 +153,92 @@ export class PhishingDataService {
...
+        );
+
+        if (found) {
+          this.logService.info(
+            "[PhishingDataService] Found phishing web address through custom matcher: " + url.href,
+          );
+        } else {
+          this.logService.debug(
+            "[PhishingDataService] No match found, returning false for: " + url.href,
+          );
```
Banrion, line 228: ⚠️ This function has a lot of redundant logs for debugging.

apps/browser/src/dirt/phishing-detection/services/phishing-detection.service.ts
```diff
@@ -63,7 +64,7 @@ export class PhishingDetectionService {
       tap((message) =>
         logService.debug(`[PhishingDetectionService] user selected continue for ${message.url}`),
       ),
-      concatMap(async (message) => {
+      mergeMap(async (message) => {
```
Banrion, line 67: ❌ Merge map will allow the same tab to run checks faster but introduces a race condition. If needing to switch to the latest check do to navigating off of a potential phishing site, we should use `switchMap` instead

apps/browser/src/dirt/phishing-detection/services/phishing-detection.service.ts
```diff
@@ -88,23 +89,40 @@ export class PhishingDetectionService {
...
-        if (!isPhishing) {
-          return;
-        }
-
-        const phishingWarningPage = new URL(
-          BrowserApi.getRuntimeURL("popup/index.html#/security/phishing-warning") +
-            `?phishingUrl=${url.toString()}`,
+      // Use mergeMap for parallel processing - each tab check runs independently
+      // Concurrency limit of 5 prevents overwhelming IndexedDB
+      mergeMap(async ({ tabId, url, ignored }) => {
```
Banrion, line 94: ❌  Revert active search counts. Use `switchMap`

apps/browser/src/dirt/phishing-detection/phishing-resources.ts
```diff
@@ -7,6 +7,8 @@ export type PhishingResource = {
   todayUrl: string;
   /** Matcher used to decide whether a given URL matches an entry from this resource */
   match: (url: URL, entry: string) => boolean;
+  /** Whether to use the custom matcher. If false, only exact hasUrl lookups are used. Default: true */
+  useCustomMatcher?: boolean;
```
Banrion, line 11: ❌ Change to variable in the service rather than adding to the resource type

PR #18562, opened 2026-01-26
Dirt/pm 20112 short term fix for member access report

prograhamming:
## 🎟️ Tracking

[PM-20112](https://bitwarden.atlassian.net/jira/software/c/projects/BW/boards/222?selectedIssue=PM-20112)

## 📔 Objective

Member Access Report backend query is too slow for large organizations. This approach will be front end only and use existing endpoints to pull to gather data and make associations on the front end.

## 📸 Screenshots

[PM-20112]: https://bitwarden.atlassian.net/browse/PM-20112?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


bitwarden_license/bit-web/src/app/dirt/reports/member-access-report/view/member-access-report.view.ts
Banrion, line 3: 💭 It seems like this is a view model. It should probably be in a view folder in the model folder. I saw "view" and thought it would be components since there was already a model folder

bitwarden_license/bit-web/src/app/dirt/reports/member-access-report/member-access-loading.component.ts
```diff
@@ -0,0 +1,107 @@
+import { CommonModule } from "@angular/common";
```
Banrion: ⛏️ The components should probably be added to a components folder

bitwarden_license/bit-web/src/app/dirt/reports/member-access-report/member-access-loading.component.ts
```diff
@@ -0,0 +1,107 @@
...
+            </span>
+          } @else {
+            <span class="tw-text-muted tw-text-sm tw-font-normal tw-leading-4">
+              {{ "thisMightTakeFewMinutes" | i18n }}
+            </span>
+          }
+        </div>
+      </div>
+    </div>
+  `,
```
Banrion: ❌  Prefer template in html file as per company standards

bitwarden_license/bit-web/src/app/dirt/reports/member-access-report/member-access-loading.component.ts
```diff
@@ -0,0 +1,107 @@
...
+  /**
+   * Progress state input from parent component.
+   * Recommended: delay emissions to this input to ensure each step displays for a minimum time.
+   */
+  readonly progressState = input<MemberAccessProgressState>({
+    step: MemberAccessProgress.FetchingMembers,
+    processedMembers: 0,
+    totalMembers: 0,
+    message: "",
+  });
```
Banrion: ❌ Separate bundled input input multiple signal inputs for change detection, removing the need for derivations (the computed functions below can be removed), reducing object creations, etc.

bitwarden_license/bit-web/src/app/dirt/reports/member-access-report/services/member-access-report.service.ts
```diff
@@ -1,154 +1,499 @@
...
+
+        if (groupCollections.length > 0) {
+          // Group has collection access - create a row for each collection
+          for (const access of groupCollections) {
+            exportItems.push({
+              email: member.email,
+              name: member.name || "(No Name)",
+              twoStepLogin: member.twoFactorEnabled
+                ? this.i18nService.t("memberAccessReportTwoFactorEnabledTrue")
+                : this.i18nService.t("memberAccessReportTwoFactorEnabledFalse"),
```
Banrion: ⛏️ There are a lot of duplicate if statements. It would be easier to read if if these were declared once and used as needed since these aren't dependent on the group access or even direct access. Having them declared in a base object above also implies the lack of dependency of those values

bitwarden_license/bit-web/src/app/dirt/reports/member-access-report/services/member-access-report.service.spec.ts
```diff
@@ -1,168 +1,599 @@
...
     );
-    reportApiService.getMemberAccessData.mockImplementation(() =>
-      Promise.resolve(memberAccessReportsMock),
-    );
-    memberAccessReportService = new MemberAccessReportService(
-      reportApiService,
-      i18nMock,
+
+    // Mock account service
+    mockAccountService.activeAccount$ = of({ id: userId } as any);
```
Banrion: 🤔  See [mockAccountServiceWith](https://github.com/bitwarden/clients/blob/ac907b90447dbf5b4278388233685f0020218a6b/libs/common/spec/fake-account-service.ts#L23) and related mock account and service features. I don't think you should need to mock the `activeAccount$`.

bitwarden_license/bit-web/src/app/dirt/reports/member-access-report/services/member-access-report.service.ts
```diff
@@ -1,154 +1,499 @@
...
+  /** Group ID if access is via group, null for direct access */
+  viaGroupId: string | null;
+  /** Group name if access is via group, null for direct access */
+  viaGroupName: string | null;
+}
+
+/**
+ * Internal interface for member data from the API
+ */
+interface MemberData {
```
Banrion: ⛏️ Should be in a model/data folder

bitwarden_license/bit-web/src/app/dirt/reports/member-access-report/services/member-access-report.service.ts
```diff
@@ -1,154 +1,499 @@
...
-
-    // group member access data by userGuid
-    const userMap = new Map<Guid, MemberAccessResponse[]>();
-    memberAccessData.forEach((userData) => {
-      const userGuid = userData.userGuid;
-      if (!userMap.has(userGuid)) {
-        userMap.set(userGuid, []);
+    // Clear cached data on new report generation
+    this.cachedLookupMaps = null;
+    this.cachedMembers = null;
```
Banrion, line 75: ❓ If you clear the cache before the results are done, there is no fallback if any of the following computations fail. If this is desired, this is fine. Do we instead want to take out the clearing of the cache to leave and only write over when the report is complete?

bitwarden_license/bit-web/src/app/dirt/reports/member-access-report/services/member-access-report.service.ts
```diff
@@ -1,154 +1,499 @@
...
    *
-   * @param {UserData} userData - The user data to aggregate.
-   * @param {ReportCollection[]} collections - An array of collections, each with an ID and a total number of items.
-   * @returns {MemberAccessReportView} The aggregated report view.
+   * @param organizationId - The organization to generate the report for
+   * @returns Array of aggregated member access views for display
    */
   async generateMemberAccessReportView(
     organizationId: OrganizationId,
   ): Promise<MemberAccessReportView[]> {
```
Banrion, line 67: 📝 I'm noting that this service previously used async/await and promises, but we will be aiming to convert these to rxjs implementations in future iterations.

bitwarden_license/bit-web/src/app/dirt/reports/member-access-report/services/member-access-report.service.spec.ts
```diff
@@ -1,168 +1,599 @@
...
+            OrganizationUserUserDetailsResponse,
+          ),
+        );
+
+        mockCollectionAdminService.collectionAdminViews$.mockReturnValue(of([]));
+
+        const group = new GroupDetailsView();
+        Object.assign(group, {
+          id: "group1",
+          organizationId: mockOrganizationId as any,
```
Banrion: ⛏️ Avoid `as any` 

Might need to import OrganizationId
```suggestion
          organizationId: mockOrganizationId as OrganizationId,
```

prograhamming:
@Banrion Implemented a new solution to solve issues for Member Access Report and Access Intelligence.

Issue #18140, opened 2025-12-29
Browser extension search box loses focus


snagle:
### Steps To Reproduce

1. Unlock the browser extension.
2. Set the browser extension timeout interval to “On browser restart”.
3. Return to the Vault view, and ensure that the Search field has focus.
4. Close the browser extension window.
5. Wait for at least 2 hours without closing the browser or opening the browser extension.
6. Open the browser extension window using the keyboard shortcut Cmd/Ctrl+Shift+Y.
7. Do not type or click anything until after Step 8 has been completed.
8. Wait for at least 5 seconds while observing the browser extension window.

### Expected Result

When the browser extension window opens the cursor should be in the search box.

### Actual Result

The browser extension window opens, the cursor is briefly in the search box for a moment. Then the browser extension window has a brief pause and appears to be 'refreshed' and the cursor disappears from the search box and the search box no longer has focus.

### Screenshots or Videos

<img width="321" height="335" alt="Image" src="https://github.com/user-attachments/assets/9597dcac-6a8c-4b6e-bad3-6e03b19ba1e1" />
<br>

Pressed cmd-shift-y to open browser extension

<br>

<img width="321" height="269" alt="Image" src="https://github.com/user-attachments/assets/025ac593-ee3a-463f-b54d-78b3d50de5bc" />

<br>

search box, blue border with cursor….

<br>

<img width="320" height="345" alt="Image" src="https://github.com/user-attachments/assets/c61ee9ad-1a48-4a25-a12a-a4f89245ad53" />

<br>

this is the ‘brief pause/refresh’ described above

<br>

<img width="317" height="132" alt="Image" src="https://github.com/user-attachments/assets/bc77c938-f559-4de6-83c3-75e621783678" />
<br>

finally, grey border, no cursor. my vault items are displayed below so I cropped those out

### Additional Context

Discussion on the bitwarden community site:
https://community.bitwarden.com/t/browser-search-box-loses-focus/92195

### Operating System

macOS

### Operating System Version

12.6

### Web Browser

Chrome

### Browser Version

Version 143.0.7499.170 (Official Build) (arm64)

### Environment Versions

Version: 2025.12.0

SDK: ‘main (0107af7)’

Server version: 2025.12.1

### Issue Tracking Info

- [x] I understand that work is tracked outside of Github. A PR will be linked to this issue should one be opened to address it, but Bitwarden doesn't use fields like "assigned", "milestone", or "project" to track progress.

abastardi:
[...]

PR #18581, opened 2026-01-26
[PM-30289] Fix for loss of focus on extension search input 

### LINKED ISSUE #20870: Search vault field is no longer focused by default when opening Bitwarden Desktop
### Steps To Reproduce

1. Launch the Bitwarden Desktop app
2. Using your keyboard, attempt to type any character

### Expected Result

The "Search vault" field should be focused by default when opening the application

### Actual Result

The "Search vault" field is not focused, so no characters are typed, and no search takes place

### Screenshots or Videos

_No response_

### Additional Context

This behaviour began in 2026.4.0 of the Desktop App. Previous versions of the desktop app prior to the UI redesign worked properly.

### Operating System

Windows

### Operating System Version

Microsoft Windows 11 Pro 10.0.26200 Build 26200

### Installation method

### Build Version

2026.4.0

### Issue Tracking Info

shane-melton:
## 🎟️ Tracking

[PM-30289](https://bitwarden.atlassian.net/browse/PM-30289)

## 📔 Objective

The VaultPopupItemService had a few inefficiencies that could lead to delayed and duplicate emissions. This has the potential to flash list of items in the Vault and/or reset the scroll position while in use. It also may have been the source of the search input losing focus (see #18140), though I was unable to successfully reproduce that specific issue to confirm if this fixes it entirely.

## 📸 Screenshots

| Before | After |
|--------|--------|
| Notice the reset of scroll position and flash of favicons <br /> <br /> <video src="https://github.com/user-attachments/assets/296e439a-ff8f-4e0b-8ad7-5d023eee4fcc" /> | <video src="https://github.com/user-attachments/assets/c6f691cd-1777-42fb-b4bf-a2e611884ada" /> |

[PM-30289]: https://bitwarden.atlassian.net/browse/PM-30289?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/browser/src/vault/popup/services/vault-popup-items.service.ts
```diff
@@ -99,6 +100,7 @@ export class VaultPopupItemsService {
         ...(showIdentities ? [CipherType.Identity] : []),
       ];
     }),
+    distinctUntilChanged((a, b) => a.length === b.length && a.every((v, i) => v === b[i])),
```
shane-melton, line 100: ℹ️ `vaultPopupAutofillService.nonLoginCipherTypesOnPage$` is a bit noisy on some pages and can cause `_otherAutoFillTypes$` to emit a few times back to back. Adding the `distinctUntilChanged` silences those duplicate emissions which were causing wasted effort down stream.

apps/browser/src/vault/popup/services/vault-popup-items.service.ts
```diff
@@ -111,6 +113,7 @@ export class VaultPopupItemsService {
       filter((userId): userId is UserId => userId != null),
       switchMap((userId) =>
         merge(this.cipherService.ciphers$(userId), this.cipherService.localData$(userId)).pipe(
+          debounceTime(0),
```
shane-melton, line 113: ℹ️ `ciphers$()` and `localData$()` both emit immediately upon first subscription and can cause a duplicate emission on first load. `debouncTime(0)` combines both the first emission and avoids duplicate work down the pipe.

apps/browser/src/vault/popup/services/vault-popup-items.service.ts
```diff
@@ -164,24 +167,35 @@ export class VaultPopupItemsService {
         }),
       ),
     ),
+    shareReplay({ refCount: true, bufferSize: 1 }),
```
shane-melton, line 158: ℹ️ `_activeCipherList$` is used in quite a few other observables in this service, and the entire pipeline was being re-executed (the collection mapping, org mapping etc.) for every subscriber; when it can be easily shared via `shareReplay()`

apps/browser/src/vault/popup/services/vault-popup-list-filters.service.ts
```diff
@@ -100,6 +100,9 @@ export class VaultPopupListFiltersService {
    */
   filters$ = this.filterForm.valueChanges.pipe(
     startWith(this.filterForm.value),
+    distinctUntilChanged(
+      (previous, current) => JSON.stringify(previous) === JSON.stringify(current),
```
shane-melton, line 112: ℹ️ During initial load, the form has an initial value set and then a subsequent value set from cache, when these values are the same it would cause the down stream observers to refresh even though nothing changed; so `distinctUntilChanged` is to prevent that duplicate downstream work.

apps/browser/src/vault/popup/services/vault-popup-items.service.ts
```diff
@@ -164,24 +167,35 @@ export class VaultPopupItemsService {
...
+  /**
+   * Observable that indicates whether there is search text present that is searchable.
+   * @private
+   */
+  private _hasSearchText = this._effectiveSearchText$.pipe(map((text) => text !== ""));
+
   private _filteredCipherList$: Observable<PopupCipherViewLike[]> = combineLatest([
     this._activeCipherList$,
-    this.searchText$,
+    this._effectiveSearchText$,
```
shane-melton, line 183: ℹ️ Only watching for changes in the "effective" search terms means the list doesn't quickly flash when a user types a single character for a query (which isn't > the minimum length) which has no impact on filtering.

Jingo88 (dismissed):

apps/browser/src/vault/popup/services/vault-popup-items.service.ts
jaasen-livefront: TIL!

jaasen-livefront (dismissed):
Nice! The `debounceTime(0)` should also help with https://bitwarden.atlassian.net/browse/PM-28428. I'll check it out.

apps/desktop/src/vault/app/vault-v3/vault.component.ts
```diff
@@ -147,6 +153,7 @@ type EmptyStateMap = Record<EmptyStateType, EmptyStateItem>;
     FormsModule,
     VaultBatchActionComponent,
     VaultOrganizationUserNotificationsComponent,
+    AutofocusDirective,
```
shane-melton, line 156: ℹ️ Missing import for desktop that was found when debugging the extension

libs/tools/send/send-ui/src/send-search/send-search.component.ts
```diff
@@ -19,7 +19,7 @@ const SearchTextDebounceInterval = 200;
 @Component({
   selector: "tools-send-search",
   templateUrl: "send-search.component.html",
-  imports: [FormsModule, I18nPipe, SearchModule],
+  imports: [FormsModule, I18nPipe, SearchModule, AutofocusDirective],
```
shane-melton, line 22: ℹ️ Missing import of the `AutofocusDirective` that prevented the search input being focused on the sends page

harr1424 (dismissed):
Tools owned file looks good!

apps/desktop/src/vault/app/vault-v3/vault.component.ts
djsmith85: ❓ Could this possibly address https://github.com/bitwarden/clients/issues/20870? Or are https://github.com/bitwarden/clients/pull/21962 or https://github.com/bitwarden/clients/pull/22116 better fixes?

apps/browser/src/platform/popup/components/popup-focus-wrap.directive.ts
```diff
@@ -10,12 +12,19 @@ import { AfterViewInit, Directive, ElementRef, OnDestroy, Renderer2 } from "@ang
  * This directive adds invisible sentinel elements at the start and end of the host.
  * When a sentinel receives focus (via Tab / Shift+Tab), focus is redirected to the
  * opposite end of the focusable elements within the host.
+ *
+ * Firefox-only: every other browser cycles focus correctly, and the sentinels are
+ * actively harmful elsewhere. Safari hands keyboard focus to the popover web view
+ * after load, which lands on the start sentinel and redirects to the last focusable
+ * element — the Settings tab button — instead of the page's autofocus target.
```
djsmith85: ⚠️ This comment is a little confusing, It applies only to Firefox but mentions other browsers handling this fine, but the sentinels are harmful "elsewhere".. where?

❓ @bitwarden/team-ui-foundation: Thoughts on taking ownership over this file? Setting focus, tab-order and ensuring accessibility, seems better handled by your team

djsmith85 (changes requested):

apps/browser/src/platform/popup/components/popup-focus-wrap.directive.ts
vleague2: We have an outstanding epic to introduce router-level focus handling to the extension, so I imagine we would take over this as part of that

apps/desktop/src/vault/app/vault-v3/vault.component.ts
shane-melton: Yes, this should address #20870. The search input is focused as expected upon unlocking the desktop app.

apps/browser/src/platform/popup/components/popup-focus-wrap.directive.ts
shane-melton: Removed the confusing comment in [263b5a5](https://github.com/bitwarden/clients/pull/18581/commits/263b5a5d00dcc4e87167fb1af872463376f5be66) (it was overly verbose) But in summary:

- Firefox requires the sentinels for tab wrapping
- Chromium doesn't need them, and the extension still behaves correctly with or without the sentinels.
- Safari doesn't need them, and they break the search auto focus behavior

Also happy to move ownership now or I can leave it for the UIF router-level focus handling Epic.

harr1424 (approved):
`libs/tools/send/send-ui/src/send-search/send-search.component.ts` still looks good 👍

djsmith85 (approved):

jengstrom-bw (approved):

vleague2 (approved):

PR #18583, opened 2026-01-26
SSH Agent v2: Add ssh key primitive types

neuronull:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-31255

## 📔 Objective

Introduces ssh key primitive types for ssh agent v2. These will be used in upcoming PRs for introducing the encrypted in memory key store. Note, we will expand the functionality of the primitive types as it is needed in subsequent PRs.


neuronull:
FYI @addisonbeck  , note the CODEOWNERS change in this PR ; ostensibly, you can consider @quexten  and @coltonhurst  as the required reviewers of the source code introduced in this PR. 🫡

apps/desktop/desktop_native/ssh_agent/src/crypto/mod.rs
```diff
@@ -0,0 +1,193 @@
...
+                    .to_owned(),
+            )),
+            _ => Err(anyhow!("Unsupported key type")),
+        }
+    }
+}
+
+/// Represents an SSH public key.
+///
+/// Contains the algorithm identifier (e.g., "ssh-ed25519", "ssh-rsa")
```
quexten, line 114: Thought: For parsing host-keys of hosts we connect to, we need ecdsa here even if we don't support it for signing. Does not need to be done here, just noting that it may be needed.

neuronull: ❓  By this, do you mean that we need to add it to the `PrivateKey` enum on line 96? (because I had started down that road, but ran into issues with ssh_key crate's random() when specifying any curve for ecdsa , in unit tests , and that led me to put the comment on line 17 of this file). And if that's the case, maybe I need to understand PM-29894 better.

quexten: So a few things, I recall that ssh-key / random had some incompatibilities. It may be easier to use the 0.7.0 pre-release. This release also add support for RSA signing with sha-512 / 256 as I recall (apologies I don't have all details present).

It's okay to use that crate since it's limited to the new ssh_agent crate / can be different from the deps used in core.

However, I only meant *public key* support, not private key support. I'm not saying we should add signing with ECDSA, that would be a feature increase (which I'm not opposed to, but I want to make that clear).

I meant correctly implementing the session-bind extension, which we use to correctly establish that a connection is forwarded via agent forwarding / *which* server it is forwarded to (verified by the hostkey and hostkey signature)
https://www.openssh.org/agent-restrict.html, section "building the forwarding path", https://github.com/bitwarden/clients/blob/419320e0742b8def57157f8994c490acb86867f2/apps/desktop/desktop_native/ssh_agent/src/protocol/requests.rs#L264

neuronull: > I recall that ssh-key / random had some incompatibilities. It may be easier to use the 0.7.0 pre-release. This release also add support for RSA signing with sha-512 / 256 as I recall (apologies I don't have all details present).

That's good to know. I did also try the 0.7.0 pre rc though , and still ran into issues with random() , but it could be that I was doing something wrong. It sounds like we can proceed for now though without adding it in the PrivateKey enum, due to that being a feature add. I'll make a note in PM-29894.

> However, I only meant public key support, not private key support. I'm not saying we should add signing with ECDSA, that would be a feature increase (which I'm not opposed to, but I want to make that clear).

Ah, got it. Thank you for clarifying that. I understand the signature bits a bit better now. Ok, it sounds like having ECDSA supported in public key for the session-bind is a requirement for feature parity of the v2 with the v1. We'll add that in a future PR (since this PR is meant to be fairly minimal to get the design laid out).

All that said, I'm fine with updating the code comment for the PublicKey struct in this PR to add `ecdsa`.

apps/desktop/desktop_native/ssh_agent/src/crypto/mod.rs
```diff
@@ -0,0 +1,193 @@
...
+///
+/// Contains the private and public components of an SSH key,
+/// along with a human-readable name and the cipher ID from the vault.
+#[derive(Clone)]
+pub(crate) struct KeyData {
+    private_key: PrivateKey,
+    public_key: PublicKey,
+    name: String,
+    cipher_id: String,
+}
```
coltonhurst: Question: The comment here says this represents an SSH key, but then the struct is named `KeyData`. (The comment at the top of the file also says `KeyData` is a "Complete SSH key".) It might be good to either update the comment or the name of the struct so it is a bit more clear.

(I don't have the full context yet but, for example, if this is actually an SSH key then we could name it `SSHKey`. If it's the core data for an ssh key, then we could call it `SSHKeyData`. However, if the name is the best possible name, that's fine, we could update the comment slightly to make it a bit more clear for new readers.)

apps/desktop/desktop_native/ssh_agent/src/crypto/mod.rs
```diff
@@ -0,0 +1,193 @@
...
+        let key_string = create_valid_ed25519_key_string();
+        let ssh_key = ssh_key::PrivateKey::from_openssh(&key_string).unwrap();
+
+        let private_key = PrivateKey::try_from(ssh_key).unwrap();
+        assert!(matches!(private_key, PrivateKey::Ed25519(_)));
+    }
+
+    #[test]
+    fn test_privatekey_from_rsa() {
+        let rsa_keypair = RsaKeypair::random(&mut OsRng, 2048).unwrap();
```
coltonhurst: Maybe worth pulling `2048` to be a constant, just to avoid magic #'s?

apps/desktop/desktop_native/ssh_agent/src/crypto/mod.rs
neuronull: Changed to `SSHKeyData` , and this prompted me to trim/update some of the rest of the doc strings. This current implementation is taking a flat / simple approach but if we add more metadata in the future we might consider creating a separate metadata struct that is nested here. But I feel little benefit to that at this point.

addisonbeck (approved):

neuronull:
resolved a merge conflict in CODEOWNERS

coltonhurst (approved):

PR #18584, opened 2026-01-26
[CL-82] rename `bit-icon` to `bit-svg`; create new `bit-icon` component for font icons

willmartian:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/CL-82

## 📔 Objective

                                                                                                                                                                                     
  Separated SVG illustrations and font icons into distinct components with clearer APIs.                                                                                                                                                                                                                                                                                                                                                                                                                                                             
                                                                                                                                                                                         
  - Renamed `bit-icon` → `bit-svg` for SVG illustrations (e.g., `<bit-svg [content]="MyIcon"></bit-svg>`)                                                                                      
    - Renamed icon input to content for clarity
    - Renamed associated constructs `Icon` --> `BitSvg`, `svgIcon` --> `svg`                                                                                                                                          
    - Updated ~25 usage sites across codebase                                                                                                                                            
  - Created new `bit-icon` component for font icons (bwi)                                                                                                                                  
    - Simple API: `<bit-icon name="bwi-lock"></bit-icon>`                                                                                                                                  
    - Type-safe with `BitwardenIcon` type                                                                                                                                                  
    - Auto-decorative (aria-hidden) unless `ariaLabel` input provided                                                                                                                            
  - Added ESLint rule `@bitwarden/components/no-bwi-class-usage`                                                                                                                           
    - Warns when using bwi classes directly instead of `<bit-icon>`                                                                                                                        
    - Encourages consistent icon usage patterns                                                                                                                                          
                                                                                                                                                                                         
### Migration                                                                                                                                                                              
  
```html                                                                                                                                                                                       
  <!-- OLD: SVG illustrations -->                                                                                                                                                        
  <bit-icon [icon]="MyIcon"></bit-icon>                                                                                                                                                  
                                                                                                                                                                                         
  <!-- NEW: SVG illustrations -->                                                                                                                                                        
  <bit-svg [content]="MyIcon"></bit-svg>                                                                                                                                                 
                                                                          
  <!-- OLD: Font icons -->                                                                                                                                                               
  <i class="bwi bwi-lock" aria-hidden="true"></i> 
                                                                                                               
  <!-- NEW: Font icons -->                                                                                                                                                               
  <bit-icon name="bwi-lock"></bit-icon> 
```

### Non-goals

- Updating `bitIconButton` to use the `BitwardenIcon` type
  - If is currently valid to leave this empty when using the `bitPasswordInputToggle` directive, which manipulates the icon property via TS. I don't like how this works, and plan to change it in the future--the type will be updated then. 
- Remove `bwi-` from the icon names. This is redundant and I want to remove it, but it adds like another 600 diffs, and we are already planning other work that maps existing icons to new alternatives--we will wait until then. 

### Acceptance criteria

- No visual diffs in Storybook (aside from some stories being renamed/moved)
- Builds successfully

### DIFF
--- a/libs/components/src/icon/icon.stories.ts
+++ b/libs/components/src/icon/icon.stories.ts
@@ -53,3 +53,9 @@ export const WithAriaLabel: Story = {
     ariaLabel: "Secure lock icon",
   },
 };
+
+export const CompareWithLegacy: Story = {
+  render: () => ({
+    template: `<bit-icon name="bwi-lock"></bit-icon> <i class="bwi bwi-lock"></i>`,
+  }),
+};


libs/components/src/icon/icon.component.ts
```diff
@@ -1,35 +1,30 @@
...
+  standalone: true,
   host: {
-    "[attr.aria-hidden]": "!ariaLabel()",
+    "[class]": "classList()",
+    "[attr.aria-hidden]": "ariaLabel() ? null : true",
     "[attr.aria-label]": "ariaLabel()",
-    "[innerHtml]": "innerHtml",
-    class: "tw-max-h-full tw-flex tw-justify-center",
   },
   template: ``,
```
BryanCunningham, line 13: @willmartian should this template render the icon as a `span` or `i`?

libs/components/src/svg/svg.component.ts
```diff
@@ -0,0 +1,31 @@
...
+  readonly content = input<BitSvg>();
+  readonly ariaLabel = input<string>();
+
+  protected readonly innerHtml = computed<SafeHtml | null>(() => {
+    const content = this.content();
+    if (!isBitSvg(content)) {
+      return null;
+    }
+    const svg = content.svg;
+    return this.domSanitizer.bypassSecurityTrustHtml(svg);
```
BryanCunningham, line 29: @willmartian Could we avoid this by just rendering the SVG or `innerHtml` in the template?

libs/components/src/icon/icon.component.ts
willmartian: `i` is just a convention for "icon" but is unnecessary. `span` isn't needed here because host elements (HTML unknown elements) are already inline by default.

libs/components/src/svg/svg.component.ts
willmartian: This is actually the same functionality from the prior icon component: https://github.com/bitwarden/clients/pull/18584/files#diff-4fda2a2cae2c1123e839fd2b483196170bdbff2ce6f158a91b497526bf6cfde8L25-L34

We _are_ passing this to `innerHtml`, in the `host` config. Passing to `innerHtml` can be dangerous if you pass untrusted/dynamic content to it, but the type we enforce only allows static HTML at compile time.

Now, we could get rid of the whole `bit-svg` component and just create angular components that use SVG directly, without `innerHtml`, like:

```ts
@Component({
  template: `<svg>...</svg>`
})
class FooComponent {}
```

But this is a larger change that was outside the goal of my PR, and I would have to think more about pros/cons and why it was made the way it is today.

BryanCunningham: Yeah, I knew this was how it worked previously. I guess I was just wondering if we could avoid doing something 'dangerous' and just render the computed `innerHtml`(assuming it returned the actual svg) signal like... 
```js
@Component({
  template: `{{ innerHtml() }}`
})
class FooComponent {}
```
Was a curiosity question more than anything

willmartian: Ah gotcha--no, that would render as a string instead of SVG/HTML

PR #18589, opened 2026-01-27
[PM-31264] Broken vault filters in milestone-1

Hinton:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-31264
## 📔 Objective

https://github.com/bitwarden/clients/commit/3228e986af79c84ba793d162d7df3c66443f6e3e changed how vault filters behave which broke the milestone-1 desktop filters since it relies more on the web filters than desktop filters.

To resolve this. we now use the same `createFilterFunction` as web rather than the custom proxy like approach.


gbubemismith:
Looks like a test is failing

PR #18591, opened 2026-01-27
"Testing" GH workflow oom kill fix

mzieniukbw:
## 📔 Objective

The "Testing" GH workflow seems to be OOM killed, due to 3 node processes eating 4.5 GB or more heap.
Since we spawn 3 node processes during testing, it can easily get close to maximum allowed memory for github runner (16 GB, excluding OS reserved memory).
Setting the limit to 3GB heap for node process should address the issue.

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,7 +18,7 @@
     "lint:fix": "eslint . --cache --cache-strategy content --fix",
     "lint:clear": "rimraf .eslintcache",
     "prettier": "prettier --cache --write .",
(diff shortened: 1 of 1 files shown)


mzieniukbw:
This doesn't work, getting `JavaScript heap out of memory` error. The 3GB of heap does not seem to be enough for the tests. Can't increase it to 4GB either, since that's too close to max available memory in the gh runner

PR #18596, opened 2026-01-27
SSH Agent v2: In-memory encrypted key store

neuronull:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-30752

## 📔 Objective

Building on https://github.com/bitwarden/clients/pull/18583 and https://github.com/bitwarden/clients/pull/18484 , this PR introduces the `KeyStore` trait, and an in-memory, encrypted implementation of it. The trait can be used to either mock the keystore in the upstream logic that will use it for the agent implementation, and/or to later add a persistent implementation in the future.

The `InMemoryEncryptedKeystore` is intended to be a simple interface. Since the `KeyData` has everything, callers can access it's getters, rather than adding functions for that to `InMemoryEncryptedKeystore`.

The `EncryptedMemoryStore` is used, with the `PulbicKey` as the map key. This is because lookup operations for the agent usage will be using public key.

`rkyv` was chosen over `bincode` due to it being unmaintained.

The serialization module is where the "magic" happens to provide a byte array representation of `KeyData` to the `EncryptedMemoryStore`. An intermediate struct was used, in order to handle converting the `ssh_key::Private` key, to a Serializable type. I found this method easier to read/rationalize about than attempting to use some of `rkyv`'s macros to wrap external structs.

Additionally, the trait `KeyData` was added, to allow mocking in forthcoming PRs that will use the KeyStore, to allow not having to create actual cryptographic keys.


neuronull:
⚠️ FYI , this PR is currently based on two upstream PRs, and CI will fail until the first is merged.

apps/desktop/desktop_native/ssh_agent/src/crypto/mod.rs
```diff
@@ -13,11 +13,37 @@
...
 use anyhow::anyhow;
+use rkyv::{Archive, Deserialize, Serialize};
 use ssh_key::private::{Ed25519Keypair, RsaKeypair};
 
+pub mod keystore;
+mod serialization;
+
+/// Represents SSH key that is queryable.
+///
+/// Allows bstracting  over different key data implementations,
```
quexten:
```suggestion
/// Allows abstracting over different key data implementations,
```
typo

apps/desktop/desktop_native/ssh_agent/src/crypto/keystore.rs
```diff
@@ -0,0 +1,296 @@
...
+/// implementations or mocks.
+#[cfg_attr(test, mockall::automock(type KeyData = MockQueryableKeyData;))]
+pub trait KeyStore: Send + Sync {
+    /// The type of key data stored by this keystore.
+    type KeyData: QueryableKeyData;
+
+    /// Stores or updates an SSH key in the keystore.
+    /// If a key with the same public key already exists, it will be overwritten.
+    ///
+    /// # Arguments
```
quexten: https://contributing.bitwarden.com/contributing/code-style/rust#arguments

The current contributing docs guidelines say that we don't want to document arguments and return values. I'm not going to block over this, as it's now autofill code, but something to note.

I do also feel the return behavior here is valuable, so I'm in favor of keeping that documented.

(Also cc @coltonhurst )

neuronull: We had a DM about this out of band last week (maybe you missed my last reply there) but I tend to disagree with the blanket rule that the contrib docs currently has. And I suspect it's just a mis-worded or incomplete guide, intended for helper functions. I'm of the opinion that `pub` should be documented thoroughly, because an external is consuming it. And, rustdocs are great, not sure what the guide is saying about lacking good conventions.
Anyway, it is still on my radar to open a PR to the Rust style guide for this. I actually created this task for it earlier this week so that I wouldn't forget.
https://bitwarden.atlassian.net/browse/PM-31660

quexten: Feel free to ignore in that case. Thank you for clarifying.

neuronull: 👍  thanks for the reminder! I think that even though Autofill Desktop has increasing surface area in the desktop native, that it's good to have a cohesive style across the repo, so this was a good call out.

neuronull:
The Desktop CI failure will resolve after merging in fix for https://github.com/bitwarden/clients/pull/18814

apps/desktop/desktop_native/ssh_agent/src/crypto/keystore.rs
neuronull: 👋  FYI , I did end up making a couple clean up changes to the doc strings here: https://github.com/bitwarden/clients/pull/18596/commits/528f4c1aa57789a739c1dd569dd672f0f790fc1e

. I still intend to open up a PR (or a discussion) about the Rust style guide, I think there is some room to add clarity to it.

I particularly find doc strings helpful even in non-library code, because IDEs can render the documentation as you would see it in official rust docs, which is extremely helpful.

There are nuances.

Anyway, your intuition was sound here, thank you for raising the concern 👍

apps/desktop/desktop_native/ssh_agent/src/crypto/mod.rs
```diff
@@ -13,11 +13,37 @@
...
+    /// # Returns
+    ///
+    /// A reference to the cipher ID that links this key to a vault entry.
+    fn cipher_id(&self) -> &String;
+}
+
 /// Represents an SSH key and its associated metadata.
 #[derive(Clone)]
-pub(crate) struct SSHKeyData {
+pub struct SSHKeyData {
```
coltonhurst: I know we discussed this before on a previous PR. Sorry to ask about this name again, this is definitely not a "critical" comment!

If this represents an SSH key, why the name `SSHKeyData`? Why not just `SSHKey`? (What is the `Data` part of the name communicating here?)

apps/desktop/desktop_native/ssh_agent/src/crypto/mod.rs
```diff
@@ -13,11 +13,37 @@
...
 
+pub mod keystore;
+mod serialization;
+
+/// Represents SSH key that is queryable.
+///
+/// Allows bstracting  over different key data implementations,
+/// for mocking in tests without requiring actual cryptographic keys.
+#[cfg_attr(test, mockall::automock)]
+pub trait QueryableKeyData: Send + Sync {
```
coltonhurst: If this represents an ssh key that is queryable, based on my question [here](https://github.com/bitwarden/clients/pull/18596/changes#r2848637645) (your answer to that may invalidate this current question which is fine!), is it worth considering a name change to `QueryableSSHKey` or something here?

apps/desktop/desktop_native/ssh_agent/src/crypto/mod.rs
```diff
@@ -51,38 +77,31 @@ impl SSHKeyData {
...
-    pub(crate) fn cipher_id(&self) -> &String {
+    fn cipher_id(&self) -> &String {
         &self.cipher_id
     }
 }
 
 /// Represents an SSH private key.
 #[derive(Clone, PartialEq, Debug)]
-pub(crate) enum PrivateKey {
+pub enum PrivateKey {
```
coltonhurst, line 23: Another "naming" comment... Based on the comment on line 102, if this is specifically meant to represent an SSH private key, would it be worth naming this `SSHPrivateKey`?

apps/desktop/desktop_native/ssh_agent/src/crypto/mod.rs
```diff
@@ -113,18 +132,18 @@ impl TryFrom<ssh_key::private::PrivateKey> for PrivateKey {
 ///
 /// Contains the algorithm identifier (e.g., "ssh-ed25519", "ssh-rsa")
 /// and the binary blob of the public key data.
-#[derive(Clone, Ord, Eq, PartialOrd, PartialEq)]
-pub(crate) struct PublicKey {
+#[derive(Clone, Ord, Eq, PartialOrd, PartialEq, Archive, Serialize, Deserialize)]
+pub struct PublicKey {
```
coltonhurst, line 55: Another "naming" comment... Based on the comment on line 131, if this is specifically meant to represent an SSH public key, would it be worth naming this `SSHPublicKey`?

apps/desktop/desktop_native/ssh_agent/src/storage/serialization.rs
```diff
@@ -0,0 +1,296 @@
...
+}
+
+fn parse_key(pem: &str) -> Result<ssh_key::PrivateKey, Error> {
+    match ssh_key::private::PrivateKey::from_openssh(pem) {
+        Ok(key) => match key.public_key().to_bytes() {
+            Ok(_) => Ok(key),
+            Err(e) => Err(anyhow!("Failed to parse public key: {e}")),
+        },
+        Err(e) => Err(anyhow!("Failed to parse key: {e}")),
+    }
```
coltonhurst, line 56: Looks like we are using `from_openssh(pem)` to turn the pem into an ssh private key. But then we verify that the to_bytes() function runs properly. I'm mostly just curious - when would the `from_openssh(pem)` fn succeed but the `to_bytes()` function fail?

Did you add this just as a sanity check so we know it's serializable? (Which is smart, I think!)

apps/desktop/desktop_native/ssh_agent/src/crypto/mod.rs
neuronull:
```
/// Represents an SSH key and its associated metadata.
```

, the "metadata" is the `name` (human readable name) and `cipher_id`.

I believe I mentioned in the precursor PR comments, that we can consider creating a separate metadata struct if we end up adding more stuff, or there is fundamental separation of needs, but I felt like at this time, just having the flat structure with 4 pieces of data is ok.

apps/desktop/desktop_native/ssh_agent/src/crypto/mod.rs
neuronull: I don't know if there is a right or wrong answer to that question, it's a bit of opinion. `SSHKeyData` , contains the `PrivateKey` and `PublicKey`. On the one hand you could argue adding the ssh prefix for consistency. On the other hand, is it redundant. On another hand, the whole crypto module is within the `ssh_agent` crate and isn't used outside of it.

If I had to justify the current approach I'd say that 

- `SSHKeyData` is more helpful to describe with the SSH prefix because it contains a set of things that are all related to SSH. While the private and public key structs are essentially each one thing, which we are implicitly using within the context of ssh.
- I left it out because I felt it didn't add a whole lot of value for the 3x characters it adds spread across where it would be used in this crate.
- It's used solely within the context of the ssh_agent crate.

But, I'm not going to die on that hill if you feel strongly about it.

apps/desktop/desktop_native/ssh_agent/src/storage/serialization.rs
neuronull: I can look into this deeper if it's something we're really concerned about but what I see in the ssh_key API is that the two operations are doing the reciprocal thing: we're first decoding a String, and then we are encoding into bytes. The public key to_bytes seems to be only fallible (AFAICT) at a conversion from a usize into a u32.

It's unclear from your comment if you're posing the question if we could use `expect()`? But if you are, I think it's sensible to handle both error cases.

apps/desktop/desktop_native/ssh_agent/src/crypto/mod.rs
neuronull: Following up here, as we had a sync chat on this subject.

Here is my thought, we can do one or more of the following:
1. keep everything as is, because we can debate about the quality of names all day
2. extract SSHKeyData (or whatever final name is), and QueryableKeyData into a separate file, because it represents the bridge between the private+public key (in this file) and the keystore module. 
3. try to rename SSHKeyData. Now that I'm looking at this more broadly, the "KeyData" suffix is used in some form in basically every module introduced in this PR. So whatever we came up with we would have to consider if it is sensible in all of those contexts. The most "verbatim" representation of it would be "SSHKeyPairAndMetadata", but that is quite a mouthful, and regarding the previous point, would imply things like "SSHKeyPairAndMetadataSerializable"

My 2c is we do # 1 and # 2, and make that module's documentation make it clear the relationship between SSHKeyData , and the keystore. I think that would help with this question that you first had about "if this is just an sshkey". And is a comprimise to not having unwieldy names.

WDYT?

apps/desktop/desktop_native/ssh_agent/src/storage/serialization.rs
neuronull: We talked out of band and noted there is no concern here purely curiosity. Marking as done.

apps/desktop/desktop_native/ssh_agent/src/storage/keystore.rs
```diff
@@ -0,0 +1,277 @@
...
+    ///
+    /// * `Ok(Some(KeyData))` if the key was found
+    /// * `Ok(None)` if no key with the given public key exists
+    /// * `Err(_)` if an error occurred during retrieval
+    fn get(&self, public_key: &PublicKey) -> Result<Option<Self::KeyData>>;
+
+    /// # Returns
+    ///
+    /// A vector of tuples containing each key's public key and human-readable name.
+    fn get_all_public_keys_and_names(&mut self) -> Result<Vec<(PublicKey, String)>>;
```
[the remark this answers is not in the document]
neuronull: this is addressed in the follow-up PR : #18880

apps/desktop/desktop_native/ssh_agent/src/crypto/mod.rs
coltonhurst: Your proposal sounds good 👍

neuronull: This prompted some further organizational thoughts. Here is what I have currently. I think it better organizes things. Glad we discussed it.

When you work on the signing task, we'd be adding a signature.rs to `crypto`.

```
└── src
    ├── crypto
    │   └── mod.rs
    ├── lib.rs
    └── storage
        ├── keydata.rs
        ├── keystore.rs
        ├── mod.rs
        └── serialization.rs
```

apps/desktop/desktop_native/ssh_agent/src/crypto/mod.rs
neuronull: I'll resolve this since it's looped into the other comment.

apps/desktop/desktop_native/ssh_agent/src/crypto/mod.rs
neuronull: same answer as https://github.com/bitwarden/clients/pull/18596#discussion_r2849917955

apps/desktop/desktop_native/ssh_agent/src/crypto/mod.rs
coltonhurst: Awesome! The structure looks good, thanks for sharing it. Whenever you are ready just request the re-review and I'll be happy to re-review / approve!

coltonhurst (approved):
I assume since the last review (I did check the commit history) the only thing that's really changed is the structure we discussed.

Approved

apps/desktop/desktop_native/ssh_agent/src/crypto/mod.rs
neuronull: huh, the re-request was already there! thanks for iterating on the organization.

neuronull:
> I assume since the last review (I did check the commit history) the only thing that's really changed is the structure we discussed.
> 

I also added an implementation of a time machine that only goes in reverse and to days of peace and love.

coltonhurst:
> > I assume since the last review (I did check the commit history) the only thing that's really changed is the structure we discussed.
> 
> I also added an implementation of a time machine that only goes in reverse and to days of peace and love.

Can you please provide docs so I know which functions to call? 😂

PR #18597, opened 2026-01-27
PM-31294: Unlock Passkey using getWebVaultUrl over getHostname 

abergs:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-31294

## 📔 Objective

This bug only surfaced in "cloud" environments. 

The getHostname method returned different urls between local+selfhost and cloud. On cloud it did not use the vault prefix which caused passkeys to not be listed as available.

This PR fetches the full url and then massages it to the hostname instead. 

We still need the seemingly double url conversion because util gets the host (even though the method name is getHostname) which means that ports etc are still included.

## 📸 Screenshots

This was the API that caused different urls to be returned. I haven't touched it, just used the same method that the selfhost api called.

image


MarsCandyBars:
An initial test of unlock with passkey works in the feature branch for web and browser

PR #18601, opened 2026-01-27
[PM-29892] - fix bulk share in vault

jaasen-livefront:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-29892

## 📔 Objective

This PR fixes an issue where if a user attempted to assign a personal cipher and an org cipher to a collection, the collection dropdown was empty. This was due to the collection filter filtering out `organizationId` of `null` instead of `undefined`. The change to `undefined` was from a recent cipher refactor.

As a bonus I decided to uncomment `ts-strict-ignore` and resolved all the type issues in the vault component.

## 📸 Screenshots

https://github.com/user-attachments/assets/cd805bf3-5532-47ef-bfce-d8ff99ec4764

### DIFF
--- a/apps/web/src/app/admin-console/organizations/collections/vault.component.ts
+++ b/apps/web/src/app/admin-console/organizations/collections/vault.component.ts
@@ -962,7 +962,7 @@ export class VaultComponent implements OnInit, OnDestroy {
     await this.editCipher(cipher, true);
   }
 
-  restore = async (c: CipherViewLike) => {
+  restore = async (c: CipherViewLike): Promise<void> => {
     const organization = await firstValueFrom(this.organization$);
     if (!CipherViewLikeUtils.isDeleted(c)) {
       return;
--- a/apps/web/src/app/vault/individual-vault/vault.component.ts
+++ b/apps/web/src/app/vault/individual-vault/vault.component.ts
@@ -467,7 +467,7 @@ export class VaultComponent<C extends CipherViewLike> implements OnInit, OnDestr
             collections,
             filter.collectionId,
           );
-          searchableCollectionNodes = selectedCollection?.children;
+          searchableCollectionNodes = selectedCollection?.children ?? [];
         }
 
         if (await this.searchService.isSearchable(activeUserId, searchText)) {
@@ -611,7 +611,7 @@ export class VaultComponent<C extends CipherViewLike> implements OnInit, OnDestr
           );
 
           this.showBulkMove = filter.type !== "trash";
-          this.isEmpty = collections.length === 0 && ciphers.length === 0;
+          this.isEmpty = collections?.length === 0 && ciphers?.length === 0;
           this.performingInitialLoad = false;
           this.refreshing = false;
 


apps/web/src/app/vault/individual-vault/vault.component.ts
```diff
@@ -1296,13 +1302,14 @@ export class VaultComponent<C extends CipherViewLike> implements OnInit, OnDestr
       await this.cipherService.restoreWithServer(uuidAsString(c.id), activeUserId);
       this.toastService.showToast({
         variant: "success",
-        title: null,
```
shane-melton, line 1284: ⛏️ These empty titles can actually be full removed instead of being set to an empty string.

apps/web/src/app/vault/individual-vault/vault.component.ts
```diff
@@ -207,20 +201,20 @@ export class VaultComponent<C extends CipherViewLike> implements OnInit, OnDestr
...
+  protected showBulkMove: boolean | null = null;
+  protected canAccessPremium: boolean | null = null;
+  protected allCollections: CollectionView[] | null = null;
   protected allOrganizations: Organization[] = [];
-  protected ciphers: C[];
-  protected collections: CollectionView[];
-  protected isEmpty: boolean;
+  protected ciphers: C[] | null = null;
+  protected collections: CollectionView[] | null = null;
+  protected isEmpty: boolean = false;
```
shane-melton: 🎨 We [want to move away from](https://contributing.bitwarden.com/architecture/adr/typescript-strict#avoid-null-prefer-undefined) `| null` and either have a sensible default `false` / `[]` or use `| undefined`.

I think it would be best to have `false` for these boolean values initially and empty arrays `[]` would mean less `?.` operators everywhere.

apps/web/src/app/vault/individual-vault/vault.component.ts
```diff
@@ -207,20 +201,20 @@ export class VaultComponent<C extends CipherViewLike> implements OnInit, OnDestr
...
   protected selectedCollection: TreeNode<CollectionView> | undefined;
   protected canCreateCollections = false;
   protected currentSearchText$: Observable<string> = this.route.queryParams.pipe(
     map((queryParams) => queryParams.search),
   );
   private searchText$ = new Subject<string>();
-  private refresh$ = new BehaviorSubject<void>(null);
+  private refresh$ = new BehaviorSubject(undefined);
```
shane-melton: ♻️ (non-blocking): This should probably just be a `Subject<void>()` instead. Though, we'd have to tweak the `firstSetup$` subscription that `switchMap`s to this for the first load (probably a `startWith` in the pipe). If its easy enough to tweak and you don't notice any issues, I'd say give it a go.

apps/web/src/app/vault/individual-vault/vault.component.ts
```diff
@@ -1270,19 +1276,19 @@ export class VaultComponent<C extends CipherViewLike> implements OnInit, OnDestr
     await this.editCipher(cipher, true);
   }
 
-  restore = async (c: C): Promise<boolean> => {
+  restore = async (c: CipherViewLike) => {
```
shane-melton, line 1256: ❓ Do we even need a return value for this function? I didn't see it being used anywhere.

apps/web/src/app/vault/individual-vault/vault.component.ts
shane-melton, line 1: 👍 Thanks for cleaning up this component! 🚀

shane-melton (changes requested):
Looks pretty good, just a few small suggestions.

apps/web/src/app/vault/individual-vault/vault.component.ts
jaasen-livefront: It's not used here but it's passed into the [vault item dialog](https://github.com/bitwarden/clients/blob/main/apps/web/src/app/vault/components/vault-item-dialog/vault-item-dialog.component.ts#L103) which expects a boolean return value, though it's not used. Should we just axe it everywhere?

apps/web/src/app/vault/individual-vault/vault.component.ts
jaasen-livefront: 🤦

apps/web/src/app/vault/individual-vault/vault.component.ts
jaasen-livefront: fixed https://github.com/bitwarden/clients/pull/18601/commits/5adf0cc9e18bdf9931186d4fd3d9a0a831d97bcb

apps/web/src/app/vault/individual-vault/vault.component.ts
jaasen-livefront: fixed https://github.com/bitwarden/clients/pull/18601/commits/5adf0cc9e18bdf9931186d4fd3d9a0a831d97bcb

apps/web/src/app/vault/individual-vault/vault.component.ts
jaasen-livefront: fixed https://github.com/bitwarden/clients/pull/18601/commits/5adf0cc9e18bdf9931186d4fd3d9a0a831d97bcb

apps/web/src/app/vault/individual-vault/vault.component.ts
```diff
@@ -479,7 +479,7 @@ export class VaultComponent<C extends CipherViewLike> implements OnInit, OnDestr
             collections,
             filter.collectionId,
           );
-          searchableCollectionNodes = selectedCollection?.children ?? [];
+          searchableCollectionNodes = selectedCollection.children;
```
shane-melton: ⛏️ `ServiceUtils.getTreeNodeObjectFromList()` lies about its return type and can technically return `null`. We'll still want the `?.` here.

apps/web/src/app/vault/individual-vault/vault.component.ts
```diff
@@ -618,12 +618,12 @@ export class VaultComponent<C extends CipherViewLike> implements OnInit, OnDestr
           this.collections = collections;
           this.selectedCollection = selectedCollection;
 
-          this.canCreateCollections = allOrganizations?.some(
+          this.canCreateCollections = allOrganizations.some(
```
shane-melton: 🎨  This one is probably okay, though if you dig down, there is potential for a `| undefined` in `this.organizations$`. Might be safest to leave it and default to `false` if `allOrganizations === undefined`.

apps/web/src/app/vault/individual-vault/vault.component.ts
```diff
@@ -891,7 +891,7 @@ export class VaultComponent<C extends CipherViewLike> implements OnInit, OnDestr
    * @returns
    */
   async editCipherAttachments(cipher: C) {
-    if (cipher?.reprompt !== 0 && !(await this.passwordRepromptService.showPasswordPrompt())) {
+    if (cipher.reprompt !== 0 && !(await this.passwordRepromptService.showPasswordPrompt())) {
```
shane-melton: ❓ Are we sure this is safe? I know the types _say_ it should always be present, but since we're making this strict for the first time and other dependent systems are still not migrated, we might not be able to safely make that assumption. In general, we should be extra careful removing these `?.` operators as they are technically, and subtly, changing behavior.

apps/web/src/app/vault/individual-vault/vault.component.ts
```diff
@@ -1598,13 +1599,13 @@ export class VaultComponent<C extends CipherViewLike> implements OnInit, OnDestr
    */
   private async getPasswordFromCipherViewLike(cipher: C): Promise<string | undefined> {
     if (!CipherViewLikeUtils.isCipherListView(cipher)) {
-      return Promise.resolve(cipher.login?.password);
+      return Promise.resolve(cipher.login.password);
```
shane-melton: 💭 I don't think we can make the guarantee that `cipher.login` is always present. Granted, you should never call this method with a non-login cipher; its not guaranteed.

apps/web/src/app/vault/individual-vault/vault.component.ts
```diff
@@ -1642,7 +1643,7 @@ export class VaultComponent<C extends CipherViewLike> implements OnInit, OnDestr
       organization$,
     ]).pipe(
       map(
-        ([policy, organization]) => (policy && policy.organizationId === organization?.id) ?? false,
+        ([policy, organization]) => (policy && policy.organizationId === organization.id) ?? false,
```
shane-melton: ⛏️ There's no strongly typed guarantee that an organization exists. It's heavily implied, I think we should keep the `?.` or absolutely ensure a value is present before removing it.

apps/web/src/app/vault/individual-vault/vault.component.ts
shane-melton: I'm good with axing it everywhere, its not a pattern we normally use.

apps/web/src/app/vault/individual-vault/vault.component.ts
jaasen-livefront: fixed https://github.com/bitwarden/clients/pull/18601/commits/6635e42377c59d6341df86c2507d9fe53dd8860d

apps/web/src/app/vault/individual-vault/vault.component.ts
jaasen-livefront: agreed. fixed https://github.com/bitwarden/clients/pull/18601/commits/6635e42377c59d6341df86c2507d9fe53dd8860d

apps/web/src/app/vault/individual-vault/vault.component.ts
jaasen-livefront: agreed. fixed https://github.com/bitwarden/clients/pull/18601/commits/6635e42377c59d6341df86c2507d9fe53dd8860d

apps/web/src/app/vault/individual-vault/vault.component.ts
jaasen-livefront: I think this should be typesafe as cipher initalizes a `LoginView` which initalizes a `password` field. I'm fine with reverting for now though.

apps/web/src/app/vault/individual-vault/vault.component.ts
jaasen-livefront: This is now removed so we're good. :)

jaasen-livefront:
@shane-melton Got a little ahead of myself. Should be gtg now.

apps/web/src/app/admin-console/organizations/collections/vault.component.ts
```diff
@@ -962,10 +962,10 @@ export class VaultComponent implements OnInit, OnDestroy {
     await this.editCipher(cipher, true);
   }
 
-  restore = async (c: CipherViewLike): Promise<boolean> => {
+  restore = async (c: CipherViewLike) => {
```
JimmyVo16: This should be `async (c: CipherViewLike): Promise<void>`

apps/web/src/app/vault/individual-vault/vault.component.ts
```diff
@@ -480,7 +467,7 @@ export class VaultComponent<C extends CipherViewLike> implements OnInit, OnDestr
             collections,
             filter.collectionId,
           );
-          searchableCollectionNodes = selectedCollection?.children ?? [];
+          searchableCollectionNodes = selectedCollection?.children;
```
shane-melton: ⛏️ We should keep the `?? []`, otherwise this can now be undefined, but its used without a conditional check below.

apps/web/src/app/vault/individual-vault/vault.component.ts
```diff
@@ -624,7 +611,7 @@ export class VaultComponent<C extends CipherViewLike> implements OnInit, OnDestr
           );
 
           this.showBulkMove = filter.type !== "trash";
-          this.isEmpty = collections?.length === 0 && ciphers?.length === 0;
+          this.isEmpty = collections.length === 0 && ciphers.length === 0;
```
shane-melton: ⛏️ Should we revert these too?

shane-melton (changes requested):

apps/web/src/app/vault/individual-vault/vault.component.ts
jaasen-livefront: Nice catch. fixed https://github.com/bitwarden/clients/pull/18601/commits/f4d7f5660c2cc1c7ed77c0d7df32148441048ecb

apps/web/src/app/vault/individual-vault/vault.component.ts
jaasen-livefront: fixed https://github.com/bitwarden/clients/pull/18601/commits/f4d7f5660c2cc1c7ed77c0d7df32148441048ecb

apps/web/src/app/admin-console/organizations/collections/vault.component.ts
jaasen-livefront: fixed https://github.com/bitwarden/clients/pull/18601/commits/f4d7f5660c2cc1c7ed77c0d7df32148441048ecb

Issue #14263, opened 2025-04-13
`Unlock with system authentication` setting is available on flatpak, but does not work


M4RC3L05:
### Steps To Reproduce

1. Enable `Unlock with system authentication`
2. Lock vault
3. Try unlock vault with `Unlock with system authentication`
4. Nothing happens

### Expected Result

Expect to be prompted for my os password to unlock bitwarden

### Actual Result

Nothing happens

### Additional Context

This is the logs with the error when i try to unlock with `Unlock with system authentication`

```
21:31:00.993 › [Process Reload Service] Ephemeral pin active, preventing process reload
polkit biometric error: MethodError(OwnedErrorName("org.freedesktop.PolicyKit1.Error.Failed"), Some("Unix process subject does not have uid set"), Msg { type: Error, serial: 3460, sender: UniqueName(":1.14"), reply-serial: 1747, body: Signature("s"), fds: [] })
21:31:59.263 › Error: Biometric authentication failed
    at OsBiometricsServiceLinux.<anonymous> (/app/Bitwarden/resources/app.asar/main.js:71961:23)
    at Generator.next (<anonymous>)
    at fulfilled (/app/Bitwarden/resources/app.asar/main.js:71910:58)
```

### Operating System

Linux

### Installation method

Other

### Build Version

2025.3.0

### Issue Tracking Info

- [x] I understand that work is tracked outside of GitHub. A PR will be linked to this issue should one be opened to address it, but Bitwarden doesn't use fields like "assigned", "milestone", or "project" to track progress.

S-Kakar:
Thank you for reporting this issue! We've added this to our internal tracking system.
ID: PM-20221

NovaSilentium:
Hi there, 

This  is due to the fact that the flatpak version does not support browser integration for biometric authentication.  You could get the .deb from [Download the Bitwarden Password Manager App for iPhone, Android, Chrome, Safari, and More | Bitwarden](https://bitwarden.com/download/)

We use GitHub issues as a place to track bugs and other development related issues. The Bitwarden Community Forums has a Feature Requests section for submitting, voting for, and discussing requests like this one: https://community.bitwarden.com/c/feature-requests/  

Please sign up on our forums (https://community.bitwarden.com/signup) and search to see if this request already exists. If so, you can vote for it and contribute to any discussions about it. If not, you can re-create the request there so that it can be properly tracked.

This issue will now be closed. 

Thanks!

M4RC3L05:
[...]

Issue #13216, opened 2025-02-02
"Unlock with system authentication" broken


Sharparam:
### Steps To Reproduce

1. Uninstall and clean system of any Bitwarden files
2. Install Bitwarden desktop (`pacman -S bitwarden`) (also tested with Flatpak which doesn't work either)
3. Install polkit policy manually
4. Launch it
5. Log in
6. Open settings and enable "Unlock with system authentication"
7. Lock
8. Try to unlock with system authentication
9. KDE shows authentication pop up
10. Bitwarden does nothing after it succeeds

Manual polkit policy saved at `/usr/share/polkit-1/actions/com.bitwarden.Bitwarden.policy`:

```xml
<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE policyconfig PUBLIC
 "-//freedesktop//DTD PolicyKit Policy Configuration 1.0//EN"
 "http://www.freedesktop.org/standards/PolicyKit/1.0/policyconfig.dtd">

<policyconfig>
    <action id="com.bitwarden.Bitwarden.unlock">
      <description>Unlock Bitwarden</description>
      <message>Authenticate to unlock Bitwarden</message>
      <defaults>
        <allow_any>no</allow_any>
        <allow_inactive>no</allow_inactive>
        <allow_active>auth_self</allow_active>
      </defaults>
    </action>
</policyconfig>
```

(Taken from https://github.com/bitwarden/clients/blob/main/apps/desktop/src/key-management/biometrics/os-biometrics-linux.service.ts#L14C23-L31)

### Expected Result

Bitwarden desktop can be unlocked with system authentication.

### Actual Result

Nothing happens after the system authentication dialog from KDE.

### Additional Context

![Image](https://github.com/user-attachments/assets/4a432555-8256-4c1b-ab36-acfaae33aefb)

**Downgrading Bitwarden desktop to version 2024.9.0 makes it work again.**

![Image](https://github.com/user-attachments/assets/25bcc8e3-38e7-4ece-8a9e-80e70466a1a9)

So something happened between 2024.9.0 and 2025.1.3.

### Operating System

Linux

### Operating System Version

Arch Linux melina 6.12.10-zen1-1-zen #1 ZEN SMP PREEMPT_DYNAMIC Sat, 18 Jan 2025 02:26:52 +0000 x86_64 GNU/Linux

### Installation method

Other

### Build Version

2025.1.3

### Issue Tracking Info

[...]

PR #18625, opened 2026-01-28
[PM-31358] Fix biometric authentication in sandboxed environments (Flatpak, Snap, etc.)

unlimitedsola:
## 🎟️ Tracking

https://github.com/bitwarden/clients/issues/14263

## 📔 Objective

Biometric authentication was failing in Flatpak with the error "Unix process subject does not have uid set". This occurred because polkit could not validate the sandboxed PID against the host PID namespace (`std::process::id()` will return the sandboxed PID).

Use polkit's system-bus-name subject type instead of unix-process. This allows polkit to query D-Bus for the connection owner's host PID and credentials, bypassing the PID namespace issue. Includes fallback to unix-process for edge cases where D-Bus unique name is unavailable.

## 📸 Screenshots

Tested in Flatpak with biometric-v1.

image

### DIFF
--- a/apps/desktop/desktop_native/core/src/biometric/unix.rs
+++ b/apps/desktop/desktop_native/core/src/biometric/unix.rs
@@ -21,7 +21,32 @@ impl super::BiometricTrait for Biometric {
     async fn prompt(_hwnd: Vec<u8>, _message: String) -> Result<bool> {
         let connection = Connection::system().await?;
         let proxy = AuthorityProxy::new(&connection).await?;
-        let subject = Subject::new_for_owner(std::process::id(), None, None)?;
+
+        // Use system-bus-name instead of unix-process to avoid PID namespace issues in
+        // sandboxed environments (e.g., Flatpak). When using unix-process with a PID from
+        // inside the sandbox, polkit cannot validate it against the host PID namespace.
+        //
+        // By using system-bus-name, polkit queries D-Bus for the connection's credentials,
+        // which includes the correct host PID and UID, avoiding namespace mismatches.
+        //
+        // If D-Bus unique name is not available, fall back to the traditional unix-process
+        // approach for compatibility with non-sandboxed environments.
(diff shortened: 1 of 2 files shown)


unlimitedsola:
@quexten Thanks for the quick review!

With this merged, do you think we should document how to setup biometrics manually?
Currently, when enabling "Unlock with system authentication" in the flatpak version, it prompts that ["manual setup is required"](https://github.com/bitwarden/clients/blob/c5bd811dfd4267665fd84f9a6cbdd926517dcc17/apps/desktop/src/locales/en/messages.json#L2334). However, the [link](https://github.com/bitwarden/clients/blob/5dc49f21d2e5aa145fe9611222b4c0c5659db1c5/apps/desktop/src/app/accounts/settings.component.ts#L670) it opens "https://bitwarden.com/help/biometrics/" doesn't have instructions on how the polkit policy file should be setup, like what is described [here](https://github.com/bitwarden/clients/blob/5dc49f21d2e5aa145fe9611222b4c0c5659db1c5/apps/desktop/src/key-management/biometrics/os-biometrics-linux.service.ts#L147-L150).

I personally was confused until I learned this from reading the issue https://github.com/bitwarden/clients/issues/13216 and the code it referenced.

quexten:
Changes look good and work locally for me (on Fedora), and this just needs a QA pass. We'll have to wait until a CI issue around flatpak is resloved. I'll work with the docs team to get the instructions updated to set up Flatpak unlock.

quexten:
@unlimitedsola Would you mind pulling in the latest changes from main? It seems I cannot force the PR to do that.

quexten:
I'll just merge this, so that it ends up in the march release, since I had tested this and it is super low risk. I'll note that there is a chance that this will get yanked again, should QA find a regression, but I don't consider this likely.

Thank you again for your contribution @unlimitedsola

PR #18634, opened 2026-01-28
[PM-27876] Update the review application flow to pre-select existing critical applications

Banrion:
## 🎟️ Tracking

[PM-27876](https://bitwarden.atlassian.net/browse/PM-27876)

## 📔 Objective

When reviewing new applications, the dialog now automatically pre-selects applications that are already marked as critical.

Primary Changes:

- Added `criticalApplicationIds` parameter to **NewApplicationsDialogData** interface
- Modified dialog constructor to pre-populate selected applications with existing critical apps
- Updated **AllActivityComponent** to subscribe to critical report results and pass critical app IDs to the dialog

Other changes:

- Modified `criticalReportResults$` setup to filter `applicationData` along with `reportData` to fix `criticalReportResults$` returning all applications
- Added a `TestHarness` for setting and access private variables for the service

## 📸 Screenshots

https://github.com/user-attachments/assets/9b82f778-2ec3-47e5-ada6-95ad0937e601

[PM-27876]: https://bitwarden.atlassian.net/browse/PM-27876?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/bitwarden_license/bit-common/src/dirt/reports/risk-insights/services/domain/risk-insights-orchestrator.service.spec.ts
+++ b/bitwarden_license/bit-common/src/dirt/reports/risk-insights/services/domain/risk-insights-orchestrator.service.spec.ts
@@ -225,9 +225,12 @@ describe("RiskInsightsOrchestratorService", () => {
       it("should complete destroy$ subject and unsubscribe reportStateSubscription", () => {
         const harness = new TestHarness(service);
 
-        // Spy on the methods you expect to be called.
-        const destroyCompleteSpy = jest.spyOn(harness.getDestroySubject(), "complete");
-        const unsubscribeSpy = jest.spyOn(harness.getReportStateSubscription(), "unsubscribe");
+        const destroySubject = harness.getDestroySubject();
+        const destroyCompleteSpy = jest.spyOn(destroySubject, "complete");
+
+        const reportStateSubscription = harness.getReportStateSubscription();
+        expect(reportStateSubscription).not.toBeNull();
+        const unsubscribeSpy = jest.spyOn(reportStateSubscription!, "unsubscribe");
 
         // Execute the destroy method.
         service.destroy();
@@ -242,7 +245,6 @@ describe("RiskInsightsOrchestratorService", () => {
   describe("criticalReportResults$", () => {
     it("should filter reportData and applicationData to only include critical applications", (done) => {
       // Arrange: Create test data with both critical and non-critical applications
-      // Note: Using plain object instead of mock() because arrays need to work with .filter()
       const testEnrichedReportData: RiskInsightsEnrichedData = {
         reportData: [
           { ...mockEnrichedReportData[0], isMarkedAsCritical: true }, // Critical app


bitwarden_license/bit-web/src/app/dirt/access-intelligence/activity/all-activity.component.ts
```diff
@@ -84,6 +86,19 @@ export class AllActivityComponent implements OnInit {
         this.updateShowNeedsReviewState();
       });
 
+    // Subscribe to critical report results to get critical application IDs
+    this.dataService.criticalReportResults$
+      .pipe(takeUntilDestroyed(this.destroyRef))
+      .subscribe((criticalResults) => {
+        const criticalAppIds = new Set<string>();
+        if (criticalResults?.applicationData) {
+          for (const app of criticalResults.applicationData) {
```
AlexRubik, line 95: All applications appear preselected as critical when opening the review modal. This happens because `applicationData` (all apps) is used instead of `reportData` (critical apps only). The full list gets passed to the dialog as `criticalApplicationIds`, which preselects everything.

Reproduce by having at least one non critical app and clicking review apps. The modal shows all apps preselected.

```suggestion
        if (criticalResults?.reportData) {
          for (const app of criticalResults.reportData) {
```

AlexRubik (changes requested):

bitwarden_license/bit-web/src/app/dirt/access-intelligence/activity/all-activity.component.ts
Banrion: `criticalResults` should return ONLY the critical applications. It is the filtered list of all applications, intended to be used where only critical applications need to be used to avoid filtering in multiple places. While this setup is intended to change once the model updates are started, this variable is to expose the filtered list in a reactive and reusable way. An alternative would have been to have a filter function to add to the pipeline. 

See the setup of critical reports for filtering logic in `_setupCriticalApplicationReport`.


My concern is that, while testing, the bug is accurate and returns the whole list as mentioned in this comment. I will investigate.

Banrion: I've found that the criticalReportResults$ was not returning the critical applications only as it was supposed to. I've updated the setup and test cases to fix this major bug. It would explain why some work arounds were needed in some other places for critical applications.

bitwarden_license/bit-common/src/dirt/reports/risk-insights/services/domain/risk-insights-orchestrator.service.spec.ts
```diff
@@ -208,4 +213,112 @@ describe("RiskInsightsOrchestratorService", () => {
       });
     });
   });
+
+  describe("criticalReportResults$", () => {
+    it("should filter reportData and applicationData to only include critical applications", (done) => {
+      // Arrange: Create test data with both critical and non-critical applications
+      const testEnrichedReportData = mock<RiskInsightsEnrichedData>({
```
AlexRubik: Mock object doesn't handle .filter() on arrays correctly.
  
```suggestion
      const testEnrichedReportData: RiskInsightsEnrichedData = {...};
```

bitwarden_license/bit-common/src/dirt/reports/risk-insights/services/domain/risk-insights-orchestrator.service.spec.ts
```diff
@@ -208,4 +213,112 @@ describe("RiskInsightsOrchestratorService", () => {
...
+          );
+
+          done();
+        }
+      });
+    });
+
+    it("should return empty arrays when no critical applications exist", (done) => {
+      // Arrange: Create test data with only non-critical applications
+      const testEnrichedReportData = mock<RiskInsightsEnrichedData>({
```
AlexRubik: Mock object doesn't handle .filter() on arrays correctly.
  
```suggestion
      const testEnrichedReportData: RiskInsightsEnrichedData = {...};
```

AlexRubik (changes requested):

bitwarden_license/bit-common/src/dirt/reports/risk-insights/services/domain/risk-insights-orchestrator.service.spec.ts
Banrion: Fixed as well as added a `TestHarness` since I do not like access private variables in services through bracket notation. [a1f90ce](https://github.com/bitwarden/clients/pull/18634/commits/a1f90ce81864c66506848f5a41e5594da6f8cdf4)

AlexRubik (approved):

bitwarden_license/bit-common/src/dirt/reports/risk-insights/services/domain/risk-insights-orchestrator.service.spec.ts
```diff
@@ -27,6 +32,42 @@ import { RiskInsightsEncryptionService } from "./risk-insights-encryption.servic
...
+    return this.setOrganizationContext(organizationId, orgName).setUserId(userId);
+  }
+
+  getDestroySubject() {
+    return this.service["_destroy$"];
+  }
+
+  getReportStateSubscription() {
+    return this.service["_reportStateSubscription"];
+  }
```
AlexRubik, line 68: Add non null assertion to fix strict type error
```suggestion
  getReportStateSubscription() {
    return this.service["_reportStateSubscription"]!;
  }
```

or run a check + throw descriptive error

```suggestion
  getReportStateSubscription() {
    const subscription = this.service["_reportStateSubscription"];
    if (!subscription) {
      throw new Error("Expected _reportStateSubscription to be initialized");
    }
    return subscription;
  }
```

AlexRubik (changes requested):
fix type errors `npm run test:types`

bitwarden_license/bit-common/src/dirt/reports/risk-insights/services/domain/risk-insights-orchestrator.service.spec.ts
Banrion: I'm not done with these changes due to the training session. I'll request a review when the I'm done and the build is passing!

Banrion:
Closing due to a change in desired functionality. Rather than pre-selecting an application that has been marked as critical, we will automatically mark the application as reviewed at the time of marking it as critical. The application will no longer appear as needing review.

PR #18649, opened 2026-01-29
[PM-23108] CLI Add Email Verification to Send Receive

harr1424:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-23108

## 📔 Objective

Add email OTP authentication support to the CLI's `bw send receive` command, matching the implementation already present in the web client. This will enable CLI users to access Sends that are protected with email verification. Existing functionality (password-based auth or no-auth) will be preserved, and new email OTP functionality will be gated behind a feature flag.

### DIFF
--- a/apps/cli/src/tools/send/commands/receive.command.ts
+++ b/apps/cli/src/tools/send/commands/receive.command.ts
@@ -333,7 +333,18 @@ export class SendReceiveCommand extends DownloadCommand {
           const error = otpResponse.error;
 
           if (otpInvalid(error)) {
-            return Response.badRequest("Invalid verification code");
+            return Response.badRequest("Invalid email or verification code");
+          }
+
+          /*
+            If the following evaluates to true, it means that the email address provided was not
+            configured to be used for email OTP for this Send.
+
+            To avoid leaking information that would allow email enumeration, instead return an
+            error indicating that some component of the email OTP challenge was invalid.
+           */
+          if (emailAndOtpRequired(error)) {
+            return Response.badRequest("Invalid email or verification code");
           }
         }
         return this.handleError(otpResponse);


apps/cli/src/tools/send/commands/receive.command.ts
```diff
@@ -117,33 +178,247 @@ export class SendReceiveCommand extends DownloadCommand {
...
     }
+
+    if (emailResponse.kind === "expected_server") {
+      const error = emailResponse.error;
+
+      if (emailAndOtpRequiredEmailSent(error)) {
+        return await this.promptForOtp(sendId, email);
+      } else if (emailInvalid(error)) {
+        process.stderr.write("Email address not authorized for this Send.\n");
+        return email;
```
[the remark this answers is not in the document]
harr1424: This is intentional and mirrors the flow implemented in the web client: 
- an invalid email should not behave differently than a valid email 
- this prevents an attacker from determining which emails may or may not be configured for the send

apps/cli/src/tools/send/commands/receive.command.ts
```diff
@@ -62,6 +82,51 @@ export class SendReceiveCommand extends DownloadCommand {
...
+      return url.origin;
+    } else if (this.platformUtilsService.isDev() && url.origin === urls.webVault) {
+      return urls.api;
+    } else {
+      return url.origin + "/api";
+    }
+  }
+
+  private async getUnlockedPassword(password: string, keyArray: Uint8Array) {
+    const passwordHash = await this.cryptoFunctionService.pbkdf2(
```
mcamirault, line 114: ❓ This function is marked as deprecated and mentions using the SDK, is that something we could do instead? I took a quick look and didn't see any functions that looked like viable replacements but might have missed something

apps/cli/src/tools/send/commands/receive.command.ts
```diff
@@ -117,33 +180,294 @@ export class SendReceiveCommand extends DownloadCommand {
...
+
+      if (response instanceof SendAccessToken) {
+        return response;
+      }
+
+      if (response.kind === "expired") {
+        expiredAttempts++;
+        continue;
+      }
+
```
mcamirault: 🎨 Since the `unexpected_server` and `unknown` errors aren't used in the happy path and are handled the same way in a couple places, could we consolidate the handling for them here?

apps/cli/src/tools/send/commands/receive.command.ts
```diff
@@ -117,33 +180,294 @@ export class SendReceiveCommand extends DownloadCommand {
...
+    if (emailResponse instanceof SendAccessToken) {
+      return emailResponse;
+    }
+
+    if (emailResponse.kind === "expected_server") {
+      const error = emailResponse.error;
+
+      if (emailAndOtpRequiredEmailSent(error)) {
+        return await this.promptForOtp(sendId, email);
+      } else if (emailInvalid(error)) {
```
mcamirault: ⚠️ With [this PR](https://github.com/bitwarden/clients/pull/18620) the error types have changed a little, so these two cases should be merged into a single `if (emailAndOtpRequired(error))` instead

apps/cli/src/tools/send/commands/receive.command.ts
```diff
@@ -117,33 +180,294 @@ export class SendReceiveCommand extends DownloadCommand {
...
+      return Response.error("Unknown error: " + currentResponse.error);
+    }
+
+    // The auth layer will immediately return a token for Sends with AuthType.None
+    // If this code is reached, something has gone wrong
+    if (authType === AuthType.None) {
+      return Response.error("Could not determine authentication requirements");
+    }
+
+    while (true) {
```
mcamirault: 🎨 I think we can simplify the flow here by removing the `while` loop, putting the `AuthType.None` check into another `if` branch after the `AuthType.Password` check below, and move the `accessSendWithToken` call into the `promptForOtp` function so that the `AuthType.Email` branch functions the same way the `AuthType.Password` branch does (`handleEmailOtpAuth` should have the same parameters and return type as `handlePasswordAuth`, IMO)

apps/cli/src/tools/send/commands/receive.command.spec.ts
```diff
@@ -0,0 +1,560 @@
...
+      );
+
+      sendApiService.postSendAccess.mockRejectedValue(error);
+
+      const response = await command.run(testUrl, { password: "wrong-password" });
+
+      expect(response.success).toBe(false);
+      expect(response.message).toContain("Incorrect or missing password");
+
+      delete process.env.BW_NOINTERACTION;
```
mcamirault, line 147: 🎨 We should set this back to what it was before, if anything, instead of deleting it. This is true for a couple of other tests but I'll only leave one comment about it

apps/cli/src/tools/send/commands/receive.command.ts
harr1424: I've reached out to Key Management (KM) team, and they confirm that for now we are OK to continue using this method. They suggested as a future goal to collaborate with KM team whereby KM implements low-level crypto functionality called by higher level Send APIs in the `bitwarden-send` crate. Eventually, the goal is to move all of our Send related APIs into the SDK.

apps/cli/src/tools/send/commands/receive.command.spec.ts
harr1424: Respectfully, I disagree. There is no global test setup involved here, and so the deletion of environment variables ensures the next test will execute under a default state unless some environment variable is explicitly set. 

I think that introducing book-keeping for env var state would either be: 
- unnecessarily verbose since the existing state would need to be detected, saved in a variable, and re-assigned
- made very brittle if instead of checking variable value we make assumptions 

Were you thinking something like this? 
```typescript
  it("should return error for incorrect password in non-interactive mode", async () => {                                                                                                                                       
    const originalBwNoInteraction = process.env.BW_NOINTERACTION;                                                                                                                                                 
    process.env.BW_NOINTERACTION = "true";                                                                                                                                                                                     
                                                                                                                                                                                                                               
    const error = new ErrorResponse(                                                                                                                                                                                           
      {                                                                                                                                                                                                                        
        statusCode: 401,                                                                                                                                                                                                       
        message: "Unauthorized",                                                                                                                                                                                               
      },                                                                                                                                                                                                                       
      401,                                                                                                                                                                                                                     
    );                                                                                                                                                                                                                         
                                                                                                                                                                                                                               
    sendApiService.postSendAccess.mockRejectedValue(error);                                                                                                                                                                    
                                                                                                                                                                                                                               
    const response = await command.run(testUrl, { password: "wrong-password" });                                                                                                                                               
                                                                                                                                                                                                                               
    expect(response.success).toBe(false);                                                                                                                                                                                      
    expect(response.message).toContain("Incorrect or missing password");                                                                                                                                                       
                                                                                                                                                                                                                               
    if (originalBwNoInteraction !== undefined) {                                                                                                                                                
      process.env.BW_NOINTERACTION = originalBwNoInteraction;                                                                                                                                                                  
    } else {                                                                                                                                                                                                                   
      delete process.env.BW_NOINTERACTION;                                                                                                                                                                                     
    }                                                                                                                                                                                                                          
  });    
```

apps/cli/src/tools/send/commands/receive.command.ts
mcamirault: Sounds good

apps/cli/src/tools/send/commands/receive.command.spec.ts
mcamirault: That was the kind of thing I was thinking of, I was worried there might be other tests that depend on default values of the environment variables but looking through the ones that even mention `process.env` (there's fewer than I originally expected) the way you've done this is consistent with them.

apps/cli/src/tools/send/commands/receive.command.ts
```diff
@@ -117,33 +179,280 @@ export class SendReceiveCommand extends DownloadCommand {
...
+    options: OptionValues,
+  ): Promise<Response> {
+    const email = await this.promptForEmail();
+
+    const emailResponse = await this.getTokenWithRetry(sendId, {
+      kind: "email",
+      email: email,
+    });
+
+    if (emailResponse instanceof SendAccessToken) {
```
mcamirault, line 305: ❓I realize that this matches the current web client behavior but I'm just now considering, we don't want to allow this case right? The user has only entered an email address so the server shouldn't give back a token; if it does then something has gone wrong and we should ignore it so we don't expose the Send to anybody who enters any email address.

apps/cli/src/tools/send/commands/receive.command.ts
```diff
@@ -117,33 +179,280 @@ export class SendReceiveCommand extends DownloadCommand {
...
+    } else if (emailResponse.kind === "unexpected_server") {
+      return Response.error("Server error: " + JSON.stringify(emailResponse.error));
+    } else if (emailResponse.kind === "unknown") {
+      return Response.error("Error: " + emailResponse.error);
+    }
+
+    return Response.error("Failed to verify email");
+  }
+
+  private async promptForOtp(sendId: string, email: string): Promise<SendAccessToken | Response> {
```
mcamirault: ⛏️ Could we make this function just do the OTP prompt and return the OTP string the way `promptForEmail` does? The rest of the logic can go in the `handleEmailOtpAuth` function (which might be a bit too nested after the change, but I think my above comment about consolidating `unexpected_server` and `unknown` error handling would help there)

apps/cli/src/tools/send/commands/receive.command.ts
mcamirault: Following our later discussion with Alex we can put this back to what it was before since it's not the client's bug to fix. Apologies for the back and forth here!

harr1424: I wanted to double check, are you sure we gain anything by reverting the change here? The type narrowing seems essential, the comment adds valuable context to avoid future confusion, and we are returning an error from what is expected to be an impossible situation. 

```typescript
  private async handleEmailOtpAuth(
    sendId: string,
    keyArray: Uint8Array,
    apiUrl: string,
    options: OptionValues,
  ): Promise<Response> {
    const email = await this.promptForEmail();

    const emailResponse = await this.getTokenWithRetry(sendId, {
      kind: "email",
      email: email,
    });

    if (emailResponse instanceof SendAccessToken) {
      /*
        At this point emailResponse should only be expected to be a GetSendAccessTokenError type,
        but TS must have a logical branch in case it is a SendAccessToken type. If a valid token is
        returned by the method above, something has gone wrong.
       */

      return Response.error("Unexpected server response");
    }
```

mcamirault: Good points all...the only nitpick I'd have with it is that it'll leave the CLI and web client behaviors for this case out of sync for a bit, but we can take a ticket to rectify that

PR #18657, opened 2026-01-29
[PM-31384] Prevent dialog header from stealing focus from autofocus inputs

vleague2:
## 🎟️ Tracking

[PM-31384](https://bitwarden.atlassian.net/browse/PM-31384)

## 📔 Objective

We recently changed the dialog component's autofocus behavior to wait a tick (via `setTimeout`) to allow the menu component, which can be a dialog trigger, to complete its own focus management before the dialog autofocus occurs. This had the unintended side effect of running the dialog focus after the `AutofocusDirective` behavior occurs, which caused any dialog with an autofocused input field to briefly autofocus the input and then blur the input to focus on the dialog header. That blur caused the inline validation to appear. The end result was that the user saw errors upon dialog open, without ever interacting with an input field.

This PR changes the dialog autofocus behavior to only occur if there is not a descendant element with the `AutofocusDirective` applied. This way, we can retain the precedence of all `AutofocusDirective` elements while still supporting the menu use case.

## 📸 Screenshots

Before:

https://github.com/user-attachments/assets/9397af52-3553-4e7b-b654-0a24bbe0a72c

After:

https://github.com/user-attachments/assets/0b2e8794-bc9a-4f53-9e93-6f87cb9eca5c

[PM-31384]: https://bitwarden.atlassian.net/browse/PM-31384?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/web/src/app/vault/components/vault-item-dialog/vault-item-dialog.component.spec.ts
+++ b/apps/web/src/app/vault/components/vault-item-dialog/vault-item-dialog.component.spec.ts
@@ -363,7 +363,7 @@ describe("VaultItemDialogComponent", () => {
     });
 
     it("refocuses the dialog header", async () => {
-      const focusOnHeaderSpy = jest.spyOn(component["dialogComponent"](), "focusOnHeader");
+      const focusOnHeaderSpy = jest.spyOn(component["dialogComponent"](), "handleAutofocus");
 
       await component["changeMode"]("view");
 


libs/components/src/dialog/dialog/dialog.component.ts
```diff
@@ -146,18 +148,56 @@ export class DialogComponent implements AfterViewInit {
...
      */
-    const headerFocusTimeout = setTimeout(() => {
-      this.dialogHeader().nativeElement.focus();
-    }, 0);
+    if (this.ngZone.isStable) {
+      this.handleAutofocus();
+    } else {
+      this.ngZone.onStable
+        .pipe(take(1), takeUntilDestroyed(this.destroyRef))
+        .subscribe(this.handleAutofocus.bind(this));
```
willmartian: nit pick, less complex:

```ts
await firstValueFrom(this.ngZone.onStable);

this.handleAutofocus();
```

libs/components/src/dialog/dialog/dialog.component.ts
```diff
@@ -146,18 +148,56 @@ export class DialogComponent implements AfterViewInit {
...
+   * choose the dialog header as the default fallback for dialog focus because it is always present,
+   * unlike possible interactive elements.
+   */
+  private handleAutofocus() {
+    /**
+     * Angular's contentChildren query cannot see into the internal templates of child components.
+     * We need to use a regular DOM query instead to see if there are descendants using the
+     * AutofocusDirective.
+     */
+    const scrollableContentEl = this.scrollableBody().getElementRef().nativeElement;
```
willmartian: 1. An element in the header/footer could also have the autofocus directive.
2. Just targeting the component's own element ref feels more appropriate to me; `scrollableBody` is an implementation detail of a separate subfeature

libs/components/src/dialog/dialog/dialog.component.ts
```diff
@@ -146,18 +148,56 @@ export class DialogComponent implements AfterViewInit {
...
+    // Must match selectors of AutofocusDirective
+    const autofocusDescendants = scrollableContentEl.querySelectorAll(
+      "[appAutofocus], [bitAutofocus]",
+    );
```
willmartian: If for some reason we changed these selectors, this would break. Could you add a comment to the directive?

libs/components/src/dialog/dialog/dialog.component.ts
vleague2: Oooh ty, I will also use this in the other PR that looks for `ngZone.onStable`

libs/components/src/dialog/dialog/dialog.component.ts
vleague2: [b9d02ac](https://github.com/bitwarden/clients/pull/18657/commits/b9d02ac462f709b67be188f63ca322a643f58887)

libs/components/src/dialog/dialog/dialog.component.ts
vleague2: [b9d02ac](https://github.com/bitwarden/clients/pull/18657/commits/b9d02ac462f709b67be188f63ca322a643f58887)

libs/components/src/dialog/dialog/dialog.component.ts
vleague2: [b9d02ac](https://github.com/bitwarden/clients/pull/18657/commits/b9d02ac462f709b67be188f63ca322a643f58887)

nikwithak (approved):
Short and sweet - vault changes look good!

PR #18658, opened 2026-01-29
[PM-30544] Added Critical app badge to Access Intelligence

voommen-livefront:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-30544

## 📔 Objective

1. Added the critical app badge to rows that are marked as critical
2. display the checkboxes for all records
3. truncate the application name if it is too long
4. show the application name as a tooltip (just in case it gets truncated)

## 📸 Screenshots

image

### DIFF
--- a/bitwarden_license/bit-web/src/app/dirt/access-intelligence/shared/app-table-row-scrollable-m11.component.html
+++ b/bitwarden_license/bit-web/src/app/dirt/access-intelligence/shared/app-table-row-scrollable-m11.component.html
@@ -50,7 +50,7 @@
         [attr.aria-label]="'viewItem' | i18n"
       >
         <div class="tw-flex tw-items-center tw-gap-2">
-          <div class="tw-max-w-32 tw-truncate" [title]="row.applicationName">
+          <div class="tw-max-w-md tw-truncate" [title]="row.applicationName">
             {{ row.applicationName }}
           </div>
           @if (row.isMarkedAsCritical) {


apps/web/src/locales/en/messages.json
```diff
@@ -32,6 +32,9 @@
       }
     }
   },
+  "criticalBadge":{
+    "message": "Critical"
```
lastbestdev, line 36: Just wondering: how are the rest of the language translations created? Do they need to be included in this PR?

bitwarden_license/bit-web/src/app/dirt/access-intelligence/shared/app-table-row-scrollable-m11.component.html
```diff
@@ -12,32 +12,18 @@
       <th bitSortable="memberCount" bitCell tabindex="0">{{ "totalMembers" | i18n }}</th>
     </ng-container>
     <ng-template bitRowDef let-row>
-      @if (showRowCheckBox) {
```
lastbestdev, line 15: If `showRowCheckBox` is no longer used, can we remove it from the component inputs?

apps/web/src/locales/en/messages.json
voommen-livefront: so the other translations are generated outside this PR. I don't know if they have translators, I suspect they do.

bitwarden_license/bit-web/src/app/dirt/access-intelligence/shared/app-table-row-scrollable-m11.component.html
voommen-livefront: yes, I can do that.

lastbestdev (approved):

PR #18659, opened 2026-01-29
[PM-22119] Update icon for password protected Sends on Desktop

mcamirault:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-22119

## 📔 Objective

This PR adjusts the icon for password-protected Sends in the Desktop app to match the changes being made in web and browser. It does not change the "New Send" button as was written in the ticket because the ongoing Desktop UI refresh will be replacing this entire component with the one used by web, and that one contains the button change. It does fix one visual bug in the old component that currently stops the icons from being viewed on Sends with very long names, adds some spacing between the icons, and adds some missing i18n keys that will be necessary when the Desktop UI refresh goes live.

## 📸 Screenshots

_Before_

Screenshot 2026-01-29 at 15 35 58

_After_

Screenshot 2026-01-29 at 15 21 31


harr1424 (approved):

apps/desktop/src/app/tools/send/send.component.html
```diff
@@ -73,11 +73,14 @@ <h2>{{ "types" | i18n }}</h2>
...
-            <i class="bwi bwi-fw bwi-lg" [ngClass]="s.type == 0 ? 'bwi-file-text' : 'bwi-file'"></i>
+            <i
+              class="bwi bwi-fw bwi-lg tw-text-muted"
+              [ngClass]="s.type == 0 ? 'bwi-file-text' : 'bwi-file'"
+            ></i>
           </span>
           <span class="item-content">
             <span class="item-title">
-              {{ s.name }}
+              <span class="title-text">{{ s.name }}</span>
```
vleague2, line 83: Any reason to not use tailwind classes directly here, for newly added css?

mcamirault: This component is going to be retired soon (the replacement uses Tailwind classes) and it was quicker for me to figure out how to make it display correctly with the old CSS style. This is more of a stopgap change to make it look better until the replacement happens.

vleague2: Awesome, thanks for the context!

harr1424 (approved):

PR #18660, opened 2026-01-29
[PM-29951] add archive flag check to vault-v3 desktop

Jingo88:
## 🎟️ Tracking

[PM-29951](https://bitwarden.atlassian.net/browse/PM-29951)

## 📔 Objective

Fix was previously added to vault-v2, now adding it to vault-v3. `Archive` and `Unarchive` will not show in desktop context menu unless flag is on.

## 📸 Screenshots

Screenshot 2026-01-29 at 3 47 05 PM

[PM-29951]: https://bitwarden.atlassian.net/browse/PM-29951?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/desktop/src/vault/app/vault-v3/vault.component.ts
+++ b/apps/desktop/src/vault/app/vault-v3/vault.component.ts
@@ -579,7 +579,7 @@ export class VaultComponent implements OnInit, OnDestroy, CopyClickListener {
       });
     }
 
-    if (userCanArchive && cipher.isArchived) {
+    if (cipher.isArchived && !cipher.isDeleted) {
       menu.push({
         label: this.i18nService.t("unArchive"),
         click: async () => {


apps/desktop/src/vault/app/vault-v3/vault.component.ts
```diff
@@ -579,7 +579,7 @@ export class VaultComponent implements OnInit, OnDestroy, CopyClickListener {
       });
     }
 
-    if (cipher.isArchived) {
+    if (userCanArchive && cipher.isArchived) {
```
shane-melton: ℹ️ Note this is different from what we do in vault-v2 for desktop. We weren't hiding the "Unarchive" action behind the feature flag or premium status. It was always visible if a user managed to archive an item and we for some reason rolled back the flag.

shane-melton: ⛏️ We added a `!cipher.isDeleted` check here too in vault-v2 recently as well.

shane-melton (commented):
Just two small comments before approval

jaasen-livefront (approved):

PR #18674, opened 2026-01-30
[PM-26195] Refactor selection logic in VaultItemsComponent to ensure bulk select visual bug fix

JaredScar:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-26195

## 📔 Objective

Fix visual bug where selecting first 500 items is not shown in the UI and it actually selects the 500 below

### DIFF
--- a/apps/web/src/app/vault/components/vault-items/vault-items.component.ts
+++ b/apps/web/src/app/vault/components/vault-items/vault-items.component.ts
@@ -392,8 +392,8 @@ export class VaultItemsComponent<C extends CipherViewLike> {
     const currentSort = this.dataSource.sort;
     const items = [...this.editableItems];
 
-    // If no sort column is set, return items in their original order (as displayed in table)
-    if (!currentSort || !currentSort.column || !currentSort.fn) {
+    // If no sort function is set, return items in their original order (as displayed in table)
+    if (!currentSort || !currentSort.fn) {
       return items;
     }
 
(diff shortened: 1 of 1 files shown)


apps/web/src/app/vault/components/vault-items/vault-items.component.ts
```diff
@@ -376,9 +376,30 @@ export class VaultItemsComponent<C extends CipherViewLike> {
...
+  /**
+   * Returns editableItems sorted according to the current table sort configuration.
+   * This ensures bulk selection matches the visual order displayed to the user.
+   */
+  private getSortedEditableItems(): VaultItem<C>[] {
+    const currentSort = this.dataSource.sort;
+    const items = [...this.editableItems];
+
+    // If no sort column is set, return items in their original order (as displayed in table)
+    if (!currentSort || !currentSort.column || !currentSort.fn) {
```
nikwithak: Is `| |` the right op here? i.e. should there always be *both* a `currentSort.column` and `currentSort.fn`? Or does it even matter if `column` is missing, since you aren't using it anywhere? (it looks like 

And optional nit: you could simplify this slightly with the `?` operator:
```suggestion
    if (!currentSort?.column || !currentSort?.fn) {
```

Or if you decide you don't need column after all:
```suggestion
    if (!currentSort?.fn) {
```

apps/web/src/app/vault/components/vault-items/vault-items.component.ts
```diff
@@ -376,9 +376,30 @@ export class VaultItemsComponent<C extends CipherViewLike> {
...
+      return items;
+    }
+
+    // Apply sort function with direction modifier (matches TableDataSource.sortData behavior)
+    const directionModifier = currentSort.direction === "asc" ? 1 : -1;
+    return items.sort((a, b) => currentSort.fn(a, b, currentSort.direction) * directionModifier);
```
nikwithak, line 402: This is a little counterintuitive - does `currentSort.fn(...)` not already handle the direction, since it takes `currentSort.direction` as a parameter? Wouldn't that mean that `desc` sorts are reversed to be  `asc` every time?

apps/web/src/app/vault/components/vault-items/vault-items.component.ts
JaredScar: Great catch, thanks!

apps/web/src/app/vault/components/vault-items/vault-items.component.ts
nikwithak: Looks like this is still in - is the intent here to always have it sorting in `asc` order? If so, would it be more intuitive to ignore `currentSort.direction` instead? I'm just confused why we're bothering to reverse the order twice rather than ignore it?

PR #18676, opened 2026-01-30
[PM-26706] Update search results header for extension

jengstrom-bw:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-26706

## 📔 Objective

Update header copy for search results from “All items” to “Search results”. Also updates item containers when searching/filtering

## 📸 Screenshots

https://github.com/user-attachments/assets/a399eb8b-d924-41df-b9dd-a11c13364605

### DIFF
--- a/apps/browser/src/vault/popup/components/vault-v2/vault-v2.component.spec.ts
+++ b/apps/browser/src/vault/popup/components/vault-v2/vault-v2.component.spec.ts
@@ -615,22 +615,18 @@ describe("VaultV2Component", () => {
     expect(spotlights.length).toBe(0);
   }));
 
-  it("does not render app-autofill-vault-list-items or favorites item container when hasSearchText$ is true", fakeAsync(() => {
+  it("does not render app-autofill-vault-list-items or favorites item container when hasSearchText$ is true", () => {
     itemsSvc.hasSearchText$.next(true);
 
     const fixture = TestBed.createComponent(VaultV2Component);
     component = fixture.componentInstance;
 
-    const readySubject$ = component["readySubject"] as unknown as BehaviorSubject<boolean>;
+    const readySubject$ = component["readySubject"];
     const allFilters$ = filtersSvc.allFilters$ as unknown as Subject<any>;
-
-    fixture.detectChanges();
-    tick();
 
     // Unblock loading
     readySubject$.next(true);
     allFilters$.next({});
-    tick();
     fixture.detectChanges();
 
     const autofillElement = fixture.debugElement.query(By.css("app-autofill-vault-list-items"));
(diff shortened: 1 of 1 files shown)


apps/browser/src/vault/popup/components/vault-v2/vault-v2.component.spec.ts
```diff
@@ -607,6 +615,147 @@ describe("VaultV2Component", () => {
     expect(spotlights.length).toBe(0);
   }));
 
+  it("does not render app-autofill-vault-list-items or favorites item container when hasSearchText$ is true", fakeAsync(() => {
+    itemsSvc.hasSearchText$.next(true);
+
+    const fixture = TestBed.createComponent(VaultV2Component);
+    component = fixture.componentInstance;
+
+    const readySubject$ = component["readySubject"] as unknown as BehaviorSubject<boolean>;
```
nick-livefront: nit (non-blocking): You don't need the casting here, the type is inferred from the component

```suggestion
    const readySubject$ = component["readySubject"];
```

apps/browser/src/vault/popup/components/vault-v2/vault-v2.component.html
```diff
@@ -107,20 +107,42 @@
...
+              ></app-vault-list-items-container>
+            } @else {
+              <app-vault-list-items-container
+                [title]="'items' | i18n"
+                [ciphers]="(remainingCiphers$ | async) || []"
+                id="allItems"
+                disableSectionMargin
+                collapsibleKey="allItems"
+              ></app-vault-list-items-container>
+            }
```
nick-livefront: 🎨 Using two instances of `app-vault-list-items-container` will cause a new instance of the application to be created. With only needing the `title` to update you can put a ternary right in the markup:

```suggestion
            <app-vault-list-items-container
              [title]="((numberOfAppliedFilters$ | async) === 0 ? 'allItems' : 'items') | i18n"
              [ciphers]="(remainingCiphers$ | async) || []"
              id="allItems"
              disableSectionMargin
              collapsibleKey="allItems"
            ></app-vault-list-items-container>
```

apps/browser/src/vault/popup/components/vault-v2/vault-v2.component.spec.ts
```diff
@@ -607,6 +615,147 @@ describe("VaultV2Component", () => {
...
+    const allFilters$ = filtersSvc.allFilters$ as unknown as Subject<any>;
+
+    fixture.detectChanges();
+    tick();
+
+    // Unblock loading
+    readySubject$.next(true);
+    allFilters$.next({});
+    tick();
+    fixture.detectChanges();
```
nick-livefront: nit (non-blocking): I see that this is common pattern in the test file, I probably would have done the same. Testing locally I don't see that you need the `fakeAsync`, `tick` across these tests. You can also get rid of the first `fixutre.detectChanges` as nothing is tested until after the second one. 

```suggestion
    // Unblock loading
    readySubject$.next(true);
    allFilters$.next({});
    fixture.detectChanges();
```

nick-livefront: This would be applicable to all of the added tests here.

jengstrom-bw:
updated tests/component based on Nick's feedback

nick-livefront (approved):
👏 Nice work and thank you for fixing those tests!

PR #18679, opened 2026-01-30
[PM-25365] Enhance layout responsiveness in vault component header

JaredScar:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-25365

## 📔 Objective

Gets rid of overlap of header with empty page state in table

## 📸 Screenshots

image
image

### DIFF
--- a/apps/web/src/app/vault/individual-vault/vault.component.html
+++ b/apps/web/src/app/vault/individual-vault/vault.component.html
@@ -83,7 +83,7 @@
       <span class="tw-sr-only">{{ "loading" | i18n }}</span>
     </div>
     <div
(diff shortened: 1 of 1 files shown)


apps/web/src/app/vault/individual-vault/vault.component.html
jengstrom-bw: I think you need to change here to get the spacing between the header and the bit-no-items
```suggestion
      class="tw-mt-6 tw-flex tw-h-full tw-flex-col tw-items-center tw-justify-start tw-flex-1"
```

Here is a screenshot with this change and all other changes on this file removed:

Image

sven-bitwarden (approved):

PR #18682, opened 2026-01-30
[PM-30543] Add select all checkbox to Access Intelligance Applications table

lastbestdev:
## 🎟️ Tracking
https://bitwarden.atlassian.net/browse/PM-30543

## 📔 Objective
Adds a "Select all" checkbox to the table in the Access Intelligence applications tab. This allows users to quickly select or deselect all applications currently showing in the table for marking as critical apps.

Changes do not include updates to the Mark as critical button, those will be implemented in https://bitwarden.atlassian.net/browse/PM-30540

## 📸 Screenshots
https://github.com/user-attachments/assets/0047d7e2-dec2-4f4c-8aff-f5f9ca17dd26


libs/common/src/enums/feature-flag.enum.ts
```diff
@@ -127,7 +127,7 @@ export const DefaultFeatureFlagValue = {
   [FeatureFlag.EventManagementForDataDogAndCrowdStrike]: FALSE,
   [FeatureFlag.EventManagementForHuntress]: FALSE,
   [FeatureFlag.PhishingDetection]: FALSE,
-  [FeatureFlag.Milestone11AppPageImprovements]: FALSE,
+  [FeatureFlag.Milestone11AppPageImprovements]: true,
```
voommen-livefront:
```suggestion
  [FeatureFlag.Milestone11AppPageImprovements]: FALSE,
```
by default feature flags must be `false` - there are unit tests that check for this and will fail if it is `true`

bitwarden_license/bit-web/src/app/dirt/access-intelligence/shared/app-table-row-scrollable-m11.component.ts
```diff
@@ -11,34 +11,53 @@ import { ApplicationTableDataSource } from "./app-table-row-scrollable.component
...
-  // eslint-disable-next-line @angular-eslint/prefer-signals
-  @Input() showAppAtRiskMembers!: (applicationName: string) => void;
-  // FIXME(https://bitwarden.atlassian.net/browse/CL-903): Migrate to Signals
-  // eslint-disable-next-line @angular-eslint/prefer-signals
-  @Input() unmarkAsCritical!: (applicationName: string) => void;
-  // FIXME(https://bitwarden.atlassian.net/browse/CL-903): Migrate to Signals
-  // eslint-disable-next-line @angular-eslint/prefer-signals
-  @Input() checkboxChange!: (applicationName: string, $event: Event) => void;
+  readonly dataSource = input<TableDataSource<ApplicationTableDataSource>>();
+  readonly showRowMenuForCriticalApps = input<boolean>(false);
```
voommen-livefront: I don't think we need the `showRowMenuForCriticalApps` property (after we deleted the code block that depended on this) - but if we delete it from here, we need to delete it from `applications.component.html` as well.

libs/common/src/enums/feature-flag.enum.ts
lastbestdev: Good catch! I flipped this to true for local development so my Applications tab would show properly and forgot to change back 👍🏻

lastbestdev:
> Also: there was a requirement to change the button text from "Mark app as critical" to "Mark # as critical" - I added a comment to do this.

Thanks for the assist here, but I have the updates to that button included in the follow up #18730. Including switching to "Mark # as critical". I'm going to revert that commit out, to try and avoid merge conflicts (the branch for that PR is based off of this one)

bitwarden_license/bit-web/src/app/dirt/access-intelligence/shared/app-table-row-scrollable-m11.component.ts
lastbestdev: Good catch! Done in latest

libs/common/src/enums/feature-flag.enum.ts
lastbestdev: Done in latest

voommen-livefront (approved):
Remember to update the story, indicating that the following AC is moved to PM-30540

> If user has checked Select all and then filters or searches to reduce the items in the list, the “Mark # as critical” button would reflect the number selected that is in the filtered down list currently.

PR #18684, opened 2026-01-30
[PM-29771] Make invitation non-plural if only 1 member selected

JaredScar:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-29771

## 📔 Objective

Change text from "invitations" to "invitation" if only 1 member selected to invite

## 📸 Screenshots

image
image

### DIFF
--- a/bitwarden_license/bit-web/src/app/admin-console/providers/manage/deprecated_members.component.ts
+++ b/bitwarden_license/bit-web/src/app/admin-console/providers/manage/deprecated_members.component.ts
@@ -339,4 +339,8 @@ export class MembersComponent extends BaseMembersComponent<ProviderUser> {
       .getCheckedUsers()
       .filter((member) => member.status === this.userStatusType.Invited).length;
   }
+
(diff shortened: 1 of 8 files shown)


bitwarden_license/bit-web/src/app/admin-console/providers/manage/deprecated_members.component.ts
```diff
@@ -333,4 +333,10 @@ export class MembersComponent extends BaseMembersComponent<ProviderUser> {
       return { success: false, error: error.message };
     }
   };
+
+  get selectedInvitedCount(): number {
```
JimmyVo16, line 337: Non-blocker: Can we modify `selectedInvitedCount` so that we move the `selectedInvitedCount === 1` check into a single method? Maybe `isSingleInvite`. 

This would improve DRY, since we’re doing this check in three places, and it doesn’t look like `selectedInvitedCount` has any other purpose.

PR #18700, opened 2026-01-31
[PM-32677] Add new workflow to create PR for sdk-internal update

trmartin4:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-32677

## 📔 Objective

Creates a workflow that will be triggered via `workflow_dispatch` to create a pull request in `clients` to merge in the latest SDK version, which is installed via `npm install` and replaces the manual process described [here](https://github.com/bitwarden/sdk-internal/tree/main?tab=readme-ov-file#web-clients-1).

This makes an automated process that matches the [workflow](https://github.com/bitwarden/android/blob/main/.github/workflows/sdlc-sdk-update.yml) for Android.

Corresponding `sdk-internal` PR to make this trigger on every merge of `sdk-internal` to `main`: https://github.com/bitwarden/sdk-internal/pull/729.  We don't have to make this automatic, but we have the option to.


.github/workflows/sdlc-sdk-update.yml
```diff
@@ -0,0 +1,498 @@
...
+        run: |
+          echo "👀 Committing SDK version update..."
+          SDK_VERSION_SHORT=$(echo "$SDK_VERSION" | grep -oE '[a-f0-9]{7}$' || echo "$SDK_VERSION")
+
+          git config user.name "$_BOT_NAME"
+          git config user.email "$_BOT_EMAIL"
+
+          git add package.json package-lock.json
+          git commit -m "Update SDK to $SDK_VERSION_SHORT ($SDK_VERSION)"
+          git push origin "$BRANCH_NAME"
```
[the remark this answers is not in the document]
trmartin4: Addressed with [`88990e8` (this PR)](https://github.com/bitwarden/clients/pull/18700/commits/88990e81139400674a5a9ca0ccb7eed764e310e6)

trmartin4:
@vvolkgang tagging you for review because I leaned heavily on your design for Android for this.

.github/workflows/sdk-update.yml
```diff
@@ -0,0 +1,519 @@
...
+
+env:
+  _BOT_NAME: "github-actions[bot]"
+  _BOT_EMAIL: "github-actions[bot]@users.noreply.github.com"
+
+defaults:
+  run:
+    shell: bash
+
+jobs:
```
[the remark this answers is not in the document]
trmartin4: Not sure if this is necessary, given that it's not in place for Android and hasn't caused any issues there.

.github/workflows/sdk-update.yml
```diff
@@ -0,0 +1,519 @@
+name: SDK Update
+run-name: "SDK ${{inputs.run-mode == 'Update' && format('Update - {0}', inputs.sdk-version) || format('Test - {0}', inputs.sdk-version)}}"
```
vvolkgang: ⛏️ If there's no opportunity to test something before it's merged into `main`, consider if you even need Test mode. If it's useful then this can be simplified to something like:

```suggestion
run-name: "SDK ${{ inputs.run-mode }} - ${{ inputs.sdk-version }} @ ${{ inputs.base-branch }}"
```

.github/workflows/sdk-update.yml
```diff
@@ -0,0 +1,519 @@
...
+        type: string
+        default: "main"
+
+env:
+  _BOT_NAME: "github-actions[bot]"
+  _BOT_EMAIL: "github-actions[bot]@users.noreply.github.com"
+
+defaults:
+  run:
+    shell: bash
```
vvolkgang, line 32: Just curious, did you bump into any issues by not having this set or were you just following a pattern in this repo, given the multiple different runners used?

.github/workflows/sdk-update.yml
```diff
@@ -0,0 +1,519 @@
...
+        type: choice
+        options:
+          - Update  # opens a PR in this repo updating the SDK
+          - Test    # validates SDK update without creating PR
+        default: Update
+      sdk-version:
+        description: "SDK Version"
+        type: string
+        required: true
+      base-branch:
```
vvolkgang, line 18: I'm not sure why I've been referring to this as "target" and at this point I'm too afraid to ask.

🤔 I'm thinking this would be used to opening PRs targeting release branches? That'll open up the opportunity for "unrelated branch histories" where a release branch has a changeset that never went through `main`, Not a problem on it's own if intentional and called out when we do it, as it may add unnecessary confusion while troubleshooting release issues.

.github/workflows/sdk-update.yml
```diff
@@ -0,0 +1,519 @@
...
+            CHANGELOG=$(gh api "repos/bitwarden/sdk-internal/compare/$OLD_REF...$NEW_REF" \
+              --jq '.commits | map("- " + .commit.message | split("\n")[0]) | join("\n")' 2>/dev/null || echo "")
+
+            if [ -n "$CHANGELOG" ]; then
+              echo "changelog<<__CHANGELOG_END__" >> "$GITHUB_OUTPUT"
+              echo "$CHANGELOG" >> "$GITHUB_OUTPUT"
+              echo "__CHANGELOG_END__" >> "$GITHUB_OUTPUT"
+              echo "✅ Generated changelog with $(echo "$CHANGELOG" | wc -l) commits"
+            else
+              echo "⚠️ Could not generate changelog"
```
vvolkgang: You may want to highlight these warnings in the github run summary:

```suggestion
              echo "::warning::Could not generate changelog"
```

.github/workflows/sdk-update.yml
```diff
@@ -0,0 +1,519 @@
...
+          fi
+
+      - name: Switch to branch
+        id: switch-branch
+        run: |
+          BRANCH_NAME="sdlc/sdk-update"
+          echo "branch_name=$BRANCH_NAME" >> "$GITHUB_OUTPUT"
+
+          if git switch "$BRANCH_NAME" 2>/dev/null; then
+            echo "✅ Switched to existing branch: $BRANCH_NAME"
```
vvolkgang, line 137: Some improvements on our list from a recent retro that we've not tackled yet, these aren't a blocker and can be dealt with at a later stage, just sharing for your awareness:

1. Rebase from `main` when we're updating an existing branch
2. Adding an early step to prevent updating an existing Update PR if a label is present (e.g. `stop-updates`, as we it may be useful for other workflows). This label would be set by the dev when they start fixing breaking changes. Team raised that on a busy days, new updates introduced additional breaking changes while they were in the process of fixing others, which isn't ideal when we're under time pressure to merge a fix and recent updates can wait.

.github/workflows/sdk-update.yml
```diff
@@ -0,0 +1,519 @@
...
+          npm install --save-exact @bitwarden/commercial-sdk-internal@$SDK_VERSION
+          echo "✅ SDK packages updated in package.json"
+
+      - name: Install dependencies
+        run: npm ci
+
+      - name: Run linter
+        id: lint
+        run: npm run lint
+        continue-on-error: false
```
vvolkgang: ⛏️ double check if you need this given that it's set to `false`.

.github/workflows/sdk-update.yml
```diff
@@ -0,0 +1,519 @@
...
+          elif [ "$TEST_STATUS" = "skipped" ]; then
+            TEST_RESULT="⏭️ Skipped"
+          else
+            TEST_RESULT="⚠️ Failed"
+          fi
+
+          # Build PR body
+          if [ "$DOWNGRADING" = "true" ]; then
+            PR_TITLE="⚠️ Downgrade SDK to $SDK_VERSION_SHORT ($SDK_VERSION)"
+            cat > /tmp/pr-body.md <<EOF
```
vvolkgang: 🤔 Bit hard to follow while reviewing without enhanced syntax highlighting, perhaps breaking this down to reduce repetiton or a different approach with template files?

.github/workflows/sdk-update.yml
```diff
@@ -0,0 +1,519 @@
...
+          ### Validation Results
+          - Lint: $LINT_RESULT
+          - Tests: $TEST_RESULT
+
+          ### Links
+          - [Workflow Run]($GITHUB_SERVER_URL/$GITHUB_REPOSITORY/actions/runs/$GITHUB_RUN_ID)
+          - [SDK Releases](https://github.com/bitwarden/sdk-internal/releases)
+
+          ---
+          *Auto-generated by [sdk-internal publish workflow](https://github.com/bitwarden/sdk-internal/blob/main/.github/workflows/publish-wasm-internal.yml)*
```
vvolkgang: 💭 Great idea! If you send the `run-id` as an input you could link back to the actual run that triggered this.

.github/workflows/sdk-update.yml
trmartin4: No issues I ran into, I was just following the pattern for the other jobs in this repo.

.github/workflows/sdk-update.yml
trmartin4: I was thinking that this could also be used in a case where an engineer wanted to update a feature branch with the latest SDK version instead of `main`.  For example, if they have a breaking SDK change that they want to code against in `clients`, they wouldn't want to merge it into `main` (and break everyone) until they're ready to do so, but may want to code against it in a feature branch.

Can you see any issues with that being a valid approach?

.github/workflows/sdk-update.yml
trmartin4: I'm tempted also to remove downgrading altogether.  Is that something that you've actually had to do?

In [`c031cb5` (this PR)](https://github.com/bitwarden/clients/pull/18700/commits/c031cb5ad8af0c1609fdb804153de3116e8961eb) I removed the concept of "downgrade" being called out, and treated them both the same.  Have you seen downgrades be used in a way that would necessitate that callout?

.github/workflows/sdk-update.yml
trmartin4: Good idea!  I've added that in [`51e14be` (this PR)](https://github.com/bitwarden/clients/pull/18700/commits/51e14be157690f5436cfbc3e2ad4d189d6ce288f).

.github/workflows/sdk-update.yml
trmartin4: Removed this in [`839989e` (this PR)](https://github.com/bitwarden/clients/pull/18700/commits/839989ec2c82af31b4378d3deb86bf37ece319d9)

.github/workflows/sdk-update.yml
trmartin4: I've added both of those steps, using a `stop-updates` label, in [`839989e` (this PR)](https://github.com/bitwarden/clients/pull/18700/commits/839989ec2c82af31b4378d3deb86bf37ece319d9).  I figured I might as well tackle them now and not have to follow up later.  Thank you for suggesting them 💪 .

.github/workflows/sdk-update.yml
trmartin4: Addressed in [`839989e` (this PR)](https://github.com/bitwarden/clients/pull/18700/commits/839989ec2c82af31b4378d3deb86bf37ece319d9), along with other places where I should be using `::warning::`.

.github/workflows/sdk-update.yml
trmartin4: Addressed this in [`839989e` (this PR)](https://github.com/bitwarden/clients/pull/18700/commits/839989ec2c82af31b4378d3deb86bf37ece319d9).

I opted to keep it with a condensed syntax, in case we do ever need to test it for some reason.

.github/workflows/sdk-update.yml
vvolkgang: Nope, it's a valid approach. Although devs will probably just update it manually, much faster.

.github/workflows/sdk-update.yml
```diff
@@ -185,6 +185,42 @@ jobs:
...
+
+      - name: Rebase the existing branch onto base branch to clean up diff
+        if: ${{ steps.switch-branch.outputs.updating_existing_branch == 'true' }}
+        env:
+          BASE_BRANCH: ${{ inputs.base-branch }}
+        run: |
+          echo "🔄 Rebasing onto origin/$BASE_BRANCH..."
+          if ! git rebase "origin/$BASE_BRANCH"; then
+            git rebase --abort
+            echo "::error::Rebase onto $BASE_BRANCH failed due to conflicts. Resolve conflicts manually and re-run."
```
vvolkgang, line 207: 🤔 Reckon this either happens because (1) sdk was updated in `main` and you can safely pick yours and overwrite; or (2) branch has other changes which should have been caught by the `$LATEST_COMMIT_AUTHOR` check. We could try to handle (1) here?

.github/workflows/sdk-update.yml
vvolkgang: Awesome, I'll take a note to copypaste it! 😂

.github/workflows/sdk-update.yml
trmartin4: That seems to be what Claude is suggesting here? https://github.com/bitwarden/clients/pull/18700#discussion_r2886209910

.github/workflows/sdk-update.yml
vvolkgang: Downgrade check was introduced for two reasons:

1. Changelog script diff borked if direction was flipped :finnadie: 
2. iOS updates are git hash changes, calling it out helped reviewers ensure a downgrade was intentional. This wasn't an issue on Android because package names have the (incremental and unique) workflow run number.

.github/workflows/sdk-update.yml
```diff
@@ -0,0 +1,524 @@
...
+          NEW_VERSION: ${{ inputs.sdk-version }}
+        run: |
+          # Extract git refs from version strings (format: x.y.z-build-gitref)
+          OLD_REF=$(echo "$OLD_VERSION" | grep -oE '[a-f0-9]{7,40}$' || echo "")
+          NEW_REF=$(echo "$NEW_VERSION" | grep -oE '[a-f0-9]{7,40}$' || echo "")
+
+          if [ -n "$OLD_REF" ] && [ -n "$NEW_REF" ]; then
+            echo "Fetching changelog for sdk-internal: $OLD_REF...$NEW_REF"
+            CHANGELOG=$(gh api "repos/bitwarden/sdk-internal/compare/$OLD_REF...$NEW_REF" \
+              --jq '.commits | map("- " + .commit.message | split("\n")[0]) | join("\n")' 2>/dev/null || echo "")
```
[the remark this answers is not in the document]
trmartin4: Not an issue in our case.

.github/workflows/sdk-update.yml
```diff
@@ -0,0 +1,524 @@
...
+          PR_TITLE="Update SDK to $SDK_VERSION_SHORT ($SDK_VERSION)"
+          cat > /tmp/pr-body.md <<EOF
+          ## 🤖 Automated SDK Update
+
+          This PR updates the SDK packages to version \`$SDK_VERSION\`.
+
+          ### Changes
+          - \`@bitwarden/sdk-internal\`: \`$OLD_SDK_VERSION\` → \`$SDK_VERSION\`
+          - \`@bitwarden/commercial-sdk-internal\`: \`$OLD_COMMERCIAL_VERSION\` → \`$SDK_VERSION\`
+          EOF
```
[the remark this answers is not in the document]
trmartin4: Addressed this in [`8a8944f`](https://github.com/bitwarden/clients/pull/18700/commits/8a8944fb201bfb3be7a350b5d89a2898b0bf21cc).

.github/workflows/sdk-update.yml
```diff
@@ -0,0 +1,524 @@
...
+          git config user.name "$_BOT_NAME"
+          git config user.email "$_BOT_EMAIL"
+
+          git add package.json package-lock.json
+
+          if git diff --cached --quiet; then
+            echo "::warning::No changes to commit — branch already contains SDK $SDK_VERSION"
+          else
+            git commit -m "Update SDK to $SDK_VERSION_SHORT ($SDK_VERSION)"
+            git push origin "$BRANCH_NAME"
```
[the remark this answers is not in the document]
trmartin4: Addressed this in [8a8944f](https://github.com/bitwarden/clients/pull/18700/commits/8a8944fb201bfb3be7a350b5d89a2898b0bf21cc).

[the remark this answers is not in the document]
trmartin4: Addressed this in [8a8944f](https://github.com/bitwarden/clients/pull/18700/commits/8a8944fb201bfb3be7a350b5d89a2898b0bf21cc).

.github/workflows/sdk-update.yml
```diff
@@ -0,0 +1,524 @@
...
+            exit 1
+          fi
+
+      - name: Switch to branch
+        id: switch-branch
+        run: |
+          BRANCH_NAME="sdlc/sdk-update"
+          echo "branch_name=$BRANCH_NAME" >> "$GITHUB_OUTPUT"
+
+          if git switch "$BRANCH_NAME" 2>/dev/null; then
```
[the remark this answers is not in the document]
trmartin4: Not sure if this is necessary, but will change if reviewers prefer it.

PR #18707, opened 2026-02-01
PM-18607 removed errorOnUnknownProperties property

bmbitwarden:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-18607

## 📔 Objective

Remove errorOnUnknownProperties, and update the test to pass by. importing the appropriate files.

## 📸 Screenshots

image


libs/tools/send/send-ui/src/send-list-filters/send-list-filters.component.html
```diff
@@ -1,4 +1,4 @@
-<div *ngIf="canAccessPremium$ | async" role="toolbar" [ariaLabel]="'filters' | i18n">
+<div *ngIf="canAccessPremium$ | async" role="toolbar" [attr.aria-label]="'filters' | i18n">
```
bmbitwarden, line 1: Image

bmbitwarden: Image

PR #18727, opened 2026-02-02
[CL-1025] icon font refresh

BryanCunningham:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/CL-1025

## 📔 Objective

- Implements new icon set. 
- Migrates legacy icon names to new names


scripts/material-icons/migration-report.json
willmartian, line 1: Do we need this / do we need to commit this to git?

scripts/material-icons/migrate-icon-names.ts
willmartian, line 1: Is the thought on committing this to git so we can run it again if other teams merge branches with stale icons? And then we can delete this in the near future?

BryanCunningham: Yes, we'll need to run it periodically after resolving merge conflicts

scripts/material-icons/migration-report.json
BryanCunningham: No. Deleted and moved to gitignore

scripts/material-icons/build-with-bwi-names.ts
```diff
@@ -0,0 +1,156 @@
+/**
```
iivins-livefront, line 1: Question: Are the `console.logs` intentionally left in here? I see it's a script run during development, so it could make sense to leave in

BryanCunningham: @iivins-livefront I went back and forth on ripping them out. They seem _sort of_ helpful. It was one of those things that I thought about after I had opened the PR and started getting approvals. Didn't want to dismiss any approvals I had already got so I just left them for now. We could def remove them as a followup though if we feel like they aren't necessary.

PR #18735, opened 2026-02-03
Run pre-commit hook on only changed files, excluding merges

withinfocus:
## 🎟️ Tracking

https://bitwarden.slack.com/archives/GRN5AAZ6Y/p1770127107341269?thread_ts=1769165727.198789&cid=GRN5AAZ6Y

## 📔 Objective

The Husky pre-commit hook that runs staged linting should operate on changed files and avoid merges.

### DIFF
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,1 +1,11 @@
+#!/bin/sh
+
+# Skip hook if this is a merge commit
+if [ -f .git/MERGE_HEAD ]; then
+        exit 0
+fi
+
+FILES=$(git diff --cached --name-only --diff-filter=ACMR | sed 's| |\\ |g')
+[ -z "$FILES" ] && exit 0
+
 npx lint-staged


Hinton:
@claude does this run on merge conflicts?

Hinton:
@claude so this means any code added in a merge conflict won't be formatted and you would need to manually fix it later?

withinfocus:
Some good background and rationale for why files in the merge should not be linted: https://github.com/lint-staged/lint-staged/issues/1039. This change will save enormous amounts of time compared to the risk of someone introducing a lint issue on conflict resolution. We should not be optimizing for the edge cases.

Hinton:
It's my opinion that not linting modified files during a merge will result in a context switch when the developer has to go back 30 min or days later later once they notice the CI failure potentially even requiring another round of reviews than the 30s it takes to run prettier and 2min to run linting (worst case).

Just a single instance of this would be worse than the speed up gained from 10+ merges.

The comments from the linked discussion seems to be about not running linting on non modified files which I'm not against but not linting modified files in a merge commit is counter productive to the goal of saving time.

PR #18740, opened 2026-02-03
[PM-31431] Refactor vault prompts

nick-livefront:
## 🎟️ Tracking

[PM-31431](https://bitwarden.atlassian.net/browse/PM-31431)

## 📔 Objective

There are a couple of prompts that can occur when a user lands on the vault. With the upcoming welcome work starting, we wanted a way to encapsulate all vault prompts in a single place. That way we can avoid the possibility of a user seeing multiple prompts at a single time.

This introduces a `WebVaultPromptService` that will hold all of the logic for prompts. (We already had a [onboarding service](https://github.com/bitwarden/clients/blob/9afacb12c6076de281225002a624f92821d3b649/apps/web/src/app/vault/individual-vault/vault-onboarding/services/vault-onboarding.service.ts#L29), so the naming isn't what I originally expected).
- Currently no changed logic, only moving to a service for future work

## 📸 Screenshots

Quick regression of moved features:
|Premium Dialog|Transfer Items|Auto Confirm|
|-|-|-|
|Screenshot 2026-02-03 at 3 37 37 PM|Screenshot 2026-02-03 at 3 46 52 PM|Screenshot 2026-02-03 at 4 36 28 PM|

[PM-31431]: https://bitwarden.atlassian.net/browse/PM-31431?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/web/src/app/vault/services/web-vault-prompt.service.ts
+++ b/apps/web/src/app/vault/services/web-vault-prompt.service.ts
@@ -50,7 +50,7 @@ export class WebVaultPromptService {
 
     void this.vaultItemTransferService.enforceOrganizationDataOwnership(userId);
 
-    this.setupAutoConfirm();
+    this.checkForAutoConfirm();
   }
 
   private async openAutoConfirmFeatureDialog(organization: Organization) {
@@ -63,7 +63,7 @@ export class WebVaultPromptService {
     });
   }
 
-  private setupAutoConfirm() {
+  private checkForAutoConfirm() {
     // if the policy is enabled, then the user may only belong to one organization at most.
     const organization$ = this.organizations$.pipe(map((organizations) => organizations[0]));
 


apps/web/src/app/vault/services/web-vault-prompt.service.ts
```diff
@@ -0,0 +1,123 @@
...
+
+    void this.unifiedUpgradePromptService.displayUpgradePromptConditionally();
+
+    void this.vaultItemTransferService.enforceOrganizationDataOwnership(userId);
+
+    this.setupAutoConfirm();
+  }
+
+  private async openAutoConfirmFeatureDialog(organization: Organization) {
+    if (this.autoConfirmDialogRef) {
```
shane-melton: ❓ Do we still need a persisted reference to the `autoConfirmDialogRef`?

From what I can tell the `setupAutoConfirm()` method will only ever be called at most once due to the `first()` pipe.

apps/web/src/app/vault/services/web-vault-prompt.service.ts
```diff
@@ -0,0 +1,123 @@
...
+        organizationId: organization.id,
+        firstTimeDialog: true,
+      },
+    });
+
+    await lastValueFrom(this.autoConfirmDialogRef.closed);
+    this.autoConfirmDialogRef = undefined;
+  }
+
+  private setupAutoConfirm() {
```
shane-melton: ⛏️ Thoughts on renaming this to something like `checkForAutoConfirm()`? 

"setup" leads me to believe this is setting up a long lived subscription; but we're only doing a one time check.

apps/web/src/app/vault/services/web-vault-prompt.service.ts
```diff
@@ -0,0 +1,123 @@
...
+        map((policies) => policies.find((p) => p.type === PolicyType.AutoConfirm && p.enabled)),
+      ),
+      organization$,
+    ]).pipe(
+      map(
+        ([policy, organization]) => (policy && policy.organizationId === organization?.id) ?? false,
+      ),
+    );
+
+    zip([organization$, featureFlag$, autoConfirmState$, policyEnabled$, this.userId$])
```
shane-melton, line 88: ♻️  (non-blocking): This is likely out of scope and just copies what was existing, but the way these observables are organized and consumed with a mostly empty subscribe is an odd pattern.

Reorganizing this and doing a `firstValueFrom` wrapped in a try/catch seems preferable imo

apps/web/src/app/vault/services/web-vault-prompt.service.ts
nick-livefront: Agreed, it doesn't look like it's used. It would prevent the dialog from being opened twice but that won't happen has this is triggered in `ngOnInit`

[d6cc66c](https://github.com/bitwarden/clients/pull/18740/commits/d6cc66c819f7fae0a22f1dc5ce956ee9e546854c)

apps/web/src/app/vault/services/web-vault-prompt.service.ts
nick-livefront: [69fccda](https://github.com/bitwarden/clients/pull/18740/commits/69fccda291f786251648c1d373ac7acd8b06d583)

apps/web/src/app/vault/services/web-vault-prompt.service.ts
nick-livefront: I agree, though I'd like to keep this functionally as close to the working implementation as possible to reduce any introduction of changed behavior. I'm going to leave as is for now.

PR #18744, opened 2026-02-04
[PM-31685] Removing email hashes

itsadrago:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-31685

## 📔 Objective

Based on a discussion with the KM team and AppSec, we will no longer store an encrypted list of email address hashes for Send Access functionality. Instead, we will only encrypt a list of email addresses associated with a send instance. The hashing as implemented was adding no marked benefit and complicating the design

### DIFF
--- a/libs/common/src/tools/send/models/domain/send.spec.ts
+++ b/libs/common/src/tools/send/models/domain/send.spec.ts
@@ -69,7 +69,7 @@ describe("Send", () => {
       expirationDate: null,
       deletionDate: null,
       password: undefined,
-      emails: null,
+      emails: undefined,
       disabled: undefined,
       hideEmail: undefined,
     });
@@ -95,7 +95,7 @@ describe("Send", () => {
       expirationDate: new Date("2022-01-31T12:00:00.000Z"),
       deletionDate: new Date("2022-01-31T12:00:00.000Z"),
       password: "password",
-      emails: null,
+      emails: "",
       disabled: false,
       hideEmail: true,
       authType: AuthType.None,
--- a/libs/common/src/tools/send/models/domain/send.ts
+++ b/libs/common/src/tools/send/models/domain/send.ts
@@ -66,6 +66,7 @@ export class Send extends Domain {
     this.expirationDate = obj.expirationDate != null ? new Date(obj.expirationDate) : null;
     this.hideEmail = obj.hideEmail;
     this.authType = obj.authType;
+    this.emails = obj.emails;
 
     switch (this.type) {
       case SendType.Text:
--- a/libs/common/src/tools/send/services/test-data/send-tests.data.ts
+++ b/libs/common/src/tools/send/services/test-data/send-tests.data.ts
@@ -80,6 +80,6 @@ export function testSend(id: string, name: string) {
   data.deletionDate = null;
   data.notes = new EncString("Notes!!");
   data.key = null;
-  data.emails = null;
+  data.emails = "";
   return data;
 }


libs/common/src/tools/send/models/domain/send.spec.ts
```diff
@@ -173,185 +170,6 @@ describe("Send", () => {
     });
   });
 
-  describe("Email decryption", () => {
```
harr1424, line 176: Some of these tests probably could have been salvaged to test email parsing and associated logic (handling whitespace etc.)

libs/common/src/tools/send/services/send.service.spec.ts
```diff
@@ -612,112 +609,6 @@ describe("SendService", () => {
     describe("when SendEmailOTP feature flag is ON", () => {
       beforeEach(() => {
         configService.getFeatureFlag.mockResolvedValue(true);
-        cryptoFunctionService.hash.mockClear();
-      });
-
-      describe("email encryption", () => {
```
harr1424, line 618: Some of these tests may still be useful.

libs/common/src/tools/send/services/send.service.ts
```diff
@@ -92,12 +90,10 @@ export class SendService implements InternalSendServiceAbstraction {
 
     if (sendEmailOTPEnabled && hasEmails) {
       const plaintextEmails = model.emails.join(",");
-      send.emails = await this.encryptService.encryptString(plaintextEmails, model.cryptoKey);
-      send.emailHashes = await this.hashEmails(plaintextEmails);
+      send.emails = plaintextEmails;
```
harr1424: I don't think the intermediate variable `plaintextEmails ` is necessary now: 

```typescript
send.emails = model.emails.join(",");
```

harr1424 (approved):
Changes look good overall, but please consider the review comments as time allows.

libs/common/src/tools/send/services/send.service.ts
```diff
@@ -91,13 +89,13 @@ export class SendService implements InternalSendServiceAbstraction {
     const hasEmails = (model.emails?.length ?? 0) > 0;
 
     if (sendEmailOTPEnabled && hasEmails) {
-      const plaintextEmails = model.emails.join(",");
-      send.emails = await this.encryptService.encryptString(plaintextEmails, model.cryptoKey);
-      send.emailHashes = await this.hashEmails(plaintextEmails);
+      send.emails = model.emails
+        .map((e) => e.trim())
+        .join(",")
+        .toLocaleLowerCase();
```
harr1424, line 95: I believe this call to `.toLocaleLowerCase()` might be problematic, what happens if the recipient enters a mixed-case email during the authentication flow? https://github.com/bitwarden/clients/blob/4a45414f4aaab20fccc97c29afb35e7c7ee76b31/apps/web/src/app/tools/send/send-access/send-auth.component.ts#L126

I don't think the server ever lowercases emails either: 
https://github.com/bitwarden/server/blob/9499e6f2efb8e72a3f9c414c7e55c2adc4c6c3fa/src/Api/Tools/Models/Request/SendRequestModel.cs#L254

https://github.com/bitwarden/server/blob/9499e6f2efb8e72a3f9c414c7e55c2adc4c6c3fa/src/Core/Tools/SendFeatures/Queries/SendAuthenticationQuery.cs#L44

https://github.com/bitwarden/server/blob/9499e6f2efb8e72a3f9c414c7e55c2adc4c6c3fa/src/Identity/IdentityServer/RequestValidators/SendAccess/SendEmailOtpRequestValidator.cs#L31

https://github.com/bitwarden/server/blob/9499e6f2efb8e72a3f9c414c7e55c2adc4c6c3fa/src/Sql/dbo/Tools/Stored%20Procedures/Send_Create.sql#L65

itsadrago: With the addition of https://github.com/bitwarden/server/pull/6945/changes/73a7155b2f9a9a00a1a15ea774578a50ecaa4ffa do you still think there could still be an issue?

harr1424: Looks good!

libs/common/src/tools/send/models/domain/send.ts
```diff
@@ -133,7 +129,7 @@ export class Send extends Domain {
       key: EncString.fromJSON(obj.key),
       name: EncString.fromJSON(obj.name),
       notes: EncString.fromJSON(obj.notes),
-      emails: EncString.fromJSON(obj.emails),
+      emails: obj.emails,
```
harr1424, line 133: Now that emails aren't being encrypted or decrypted, I don't think this should be here.

libs/common/src/tools/send/models/domain/send.ts
harr1424, line 70:
```suggestion
this.emails = obj.emails;
```
It looks like all `Send` properties are being assigned from the `SendData` either in `buildDomainModel` or elsewhere in the constructor, with the exception of emails. I think this is an error.

harr1424 (changes requested):

libs/common/src/tools/send/models/domain/send.ts
harr1424, line 140: ⛏️  Sorry my last comment wasn't as clear as it could have been. Since the CLI allows for creating Sends from JSON, emails should be assigned here (already correct) but as a nitpick I think the strucutre could be improved:

```typescript
  static fromJSON(obj: Jsonify<Send>) {
    if (obj == null) {
      return null;
    }

    const revisionDate = obj.revisionDate == null ? null : new Date(obj.revisionDate);
    const expirationDate = obj.expirationDate == null ? null : new Date(obj.expirationDate);
    const deletionDate = obj.deletionDate == null ? null : new Date(obj.deletionDate);
    const emails = obj.emails;  // added

    return Object.assign(new Send(), obj, {
      key: EncString.fromJSON(obj.key),
      name: EncString.fromJSON(obj.name),
      notes: EncString.fromJSON(obj.notes),
      text: SendText.fromJSON(obj.text),
      file: SendFile.fromJSON(obj.file),
      revisionDate,
      expirationDate,
      deletionDate,
      emails,  // added
    });
  }
```

itsadrago: My code shows...
```
static fromJSON(obj: Jsonify<Send>) {
    if (obj == null) {
      return null;
    }

    const revisionDate = obj.revisionDate == null ? null : new Date(obj.revisionDate);
    const expirationDate = obj.expirationDate == null ? null : new Date(obj.expirationDate);
    const deletionDate = obj.deletionDate == null ? null : new Date(obj.deletionDate);

    return Object.assign(new Send(), obj, {
      key: EncString.fromJSON(obj.key),
      name: EncString.fromJSON(obj.name),
      notes: EncString.fromJSON(obj.notes),
      emails: obj.emails,
      text: SendText.fromJSON(obj.text),
      file: SendFile.fromJSON(obj.file),
      revisionDate,
      expirationDate,
      deletionDate,
    });
  }
```
...where the `emails` field has already been added

See https://github.com/bitwarden/clients/blob/tools/pm-31685/remove-email-hashing/libs/common/src/tools/send/models/domain/send.ts

PR #18751, opened 2026-02-04
[PM-31668] Race condition in cipher cache clearing causes stale failed decryption state after leaving organization

gbubemismith:
## 🎟️ Tracking
https://bitwarden.atlassian.net/browse/PM-31668

## 📔 Objective
This PR addresses two issues:
- Remove getAllDecrypted() dependency from cipherListViews$: The cipherListViews$ observable was calling getAllDecrypted() internally for search indexing, which defeated the purpose of using the lightweight CipherListView from the SDK. This also introduced unnecessary decryption overhead.
- When a user leaves an organization and a full sync occurs, cipherListViews$ could emit stale cipher data

## 📸 Screenshots

https://github.com/user-attachments/assets/aece2805-688d-4bae-b5e4-60d610b8b6e3


libs/common/src/vault/utils/cipher-view-like-utils.ts
```diff
@@ -290,6 +291,71 @@ export class CipherViewLikeUtils {
...
+   * @param cipher - The cipher to extract fields from (either `CipherView` or `CipherListView`)
+   * @returns Array of field objects with `name` and `value` properties, `undefined` if not set
+   */
+  static getFields = (
+    cipher: CipherViewLike,
+  ): { name?: string | null; value?: string | undefined }[] | undefined => {
+    if (this.isCipherListView(cipher)) {
+      return cipher.fields;
+    }
+    return cipher.fields;
```
nick-livefront, line 317: nit (non-blocking): the `this.isCipherListView(cipher)` doesn't appear to be needed?

```suggestion
    return cipher.fields;
```

PR #18755, opened 2026-02-04
[CL-1028] create popover anchor directive and spotlight mode

willmartian:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/CL-1028

## 📔 Objective

- Creates a directive `bitPopoverAnchorFor`, which is meant to be used instead of `bitPopoverTriggerFor` if the popover is programmatically controlled and not associated with a trigger button.
- Refactor `bitPopoverTriggerFor` to use `bitPopoverAnchorFo` as a host directive
- Add a `spotlight` input/capability to `bitPopoverAnchorFor` that shows a spotlight scrim
- Refactored kitchen sink stories to enable better testing of spotlights 
  - in Storybook, see Kitchen Sink --> Guided Tours 

## 📸 Screenshots

See Storybook:

https://github.com/user-attachments/assets/0797ee7a-2ad5-4138-821c-d15e976ce395


willmartian:
@vleague2 Should I update the name to be `bitPopoverAnchorFor`? That is more consistent with the trigger directive but it always felt clunky to me.

willmartian:
@claude Can you review this?

BryanCunningham (approved):

libs/components/src/popover/spotlight.service.ts
```diff
@@ -0,0 +1,192 @@
...
+  private currentTarget: HTMLElement | null = null;
+  private currentPadding = 0;
+  private resizeObserver: ResizeObserver | null = null;
+  private scrollListener: (() => void) | null = null;
+  private hideTimeout: number | null = null;
+  private activePopover: PopoverAnchorDirective | null = null;
+
+  constructor() {
+    // Create backdrop element (initially hidden)
+    this.backdropElement = document.createElement("div");
```
vleague2, line 26: There's not an Angular cdk utility that does backdrop creation or anything? :')

libs/components/src/popover/popover-anchor.directive.ts
```diff
@@ -0,0 +1,237 @@
...
+
+/**
+ * Directive that anchors a popover to any element for programmatic control.
+ * Ideal for guided tours, tooltips, and contextual help.
+ * Use `[(popoverOpen)]` for two-way binding to control visibility.
+ *
+ * @example
+ * Basic usage:
+ * ```html
+ * <div [bitPopoverAnchor]="tourStep" [(popoverOpen)]="showTour">
```
vleague2: I feel like I don't want the `For` until I look at the usage example. It just feels more immediately understandable what the relationship is when it's `[bitPopoverAnchorFor]="tourStep"`. Like "this is an anchor for the tour step" rather than "the anchor is tour step"

jaasen-livefront:
@willmartian How are we looking on this? I've run it against my coachmarks work and it's working beautifully. ;)

libs/components/src/popover/spotlight.service.ts
willmartian: Updated to better utilize CDK. We can just attach the backdrop as a CDK Overlay itself, the same way we do the popover. Doesn't entirely avoid low level DOM, but is better! Thank you. [5f3292f](https://github.com/bitwarden/clients/pull/18755/commits/5f3292f4f0d54e9f79940e5ea1e30d9b579e0317)

libs/components/src/popover/popover-anchor.directive.ts
willmartian: Agreed. Fixed here [e45bfc2](https://github.com/bitwarden/clients/pull/18755/commits/e45bfc2f3c25ed64b0c5dccb71c7784bbdb56d58)

libs/components/src/popover/popover-anchor-for.directive.ts
```diff
@@ -0,0 +1,236 @@
...
+      return [preferredPosition, ...defaultPositions];
+    }
+
+    return defaultPositions;
+  }
+
+  get defaultPopoverConfig(): OverlayConfig {
+    return {
+      hasBackdrop: !this.spotlight(), // Spotlight manages its own backdrop
+      backdropClass: "cdk-overlay-transparent-backdrop",
```
jaasen-livefront, line 92: I think we may want to use the same [backdrop classes as the dialog](https://github.com/bitwarden/clients/blob/main/libs/components/src/dialog/dialog.service.ts#L214) to ensure consistent presentation.

willmartian:
> @willmartian How are we looking on this? I've run it against my coachmarks work and it's working beautifully. ;)

Got it open for review now. Thanks for your patience!

libs/components/src/popover/popover-anchor-for.directive.ts
willmartian: Great point, added here [9230ea2](https://github.com/bitwarden/clients/pull/18755/commits/9230ea2b1c0f840ebf1b330895dea8cf93495dcd)

libs/components/src/popover/popover.component.css
```diff
@@ -47,3 +47,8 @@
 .bit-popover-above-end .bit-popover-arrow {
   @apply tw-right-6 tw-translate-x-1/2;
 }
+
+/* Spotlight backdrop for guided tours */
+.bit-popover-spotlight-backdrop {
+  @apply tw-bg-transparent tw-pointer-events-auto;
+}
```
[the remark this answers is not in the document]
willmartian: Removed here [83b5149](https://github.com/bitwarden/clients/pull/18755/commits/83b5149e96247cdeb9262e5de902e1119fce730d)

libs/components/src/popover/popover-anchor-for.directive.ts
```diff
@@ -0,0 +1,236 @@
...
+      this.rafId1 = null;
+    }
+    if (this.rafId2 !== null) {
+      cancelAnimationFrame(this.rafId2);
+      this.rafId2 = null;
+    }
+
+    if (this.spotlight()) {
+      this.spotlightService.hideSpotlight();
+    }
```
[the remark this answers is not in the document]
willmartian: Fixed here [3ea84f8](https://github.com/bitwarden/clients/pull/18755/commits/3ea84f8eff7831d853aa540ad1dbfff8706429aa)

libs/components/src/dialog/dialog.service.ts
```diff
@@ -211,7 +211,7 @@ export class DialogService {
   private router = inject(Router);
   private authService = inject(AuthService, { optional: true });
 
-  private backDropClasses = ["tw-fixed", "tw-bg-black", "tw-bg-opacity-30", "tw-inset-0"];
+  private backDropClasses = ["tw-fixed", "tw-bg-bg-overlay", "tw-inset-0"];
```
jaasen-livefront, line 214:
```suggestion
  private backDropClasses = ["tw-fixed", "tw-bg-overlay", "tw-inset-0"];
```

vleague2: It actually is correct as the duplicated `bg-bg-overlay`, see new color docs [here](https://components.bitwarden.com/?path=/docs/documentation-colors--docs#semantic-color-tokens)

libs/components/src/popover/popover-anchor-for.directive.spec.ts
```diff
@@ -0,0 +1,342 @@
...
+      expect(overlayRef.attach).toHaveBeenCalled();
+
+      fixture.destroy();
+
+      expect(overlayRef.dispose).toHaveBeenCalled();
+
+      flush();
+    }));
+
+    it("should unsubscribe from closed events on destroy", fakeAsync(() => {
```
vleague2: This kind of feels like testing Angular

libs/components/src/popover/popover-anchor-for.directive.ts
```diff
@@ -0,0 +1,204 @@
...
+  exportAs: "popoverAnchor",
+})
+export class PopoverAnchorForDirective implements OnDestroy {
+  /** Controls popover visibility. Supports two-way binding with `[(popoverOpen)]` */
+  readonly popoverOpen = model(false);
+
+  /** The popover component to display */
+  readonly popover = input.required<PopoverComponent>({ alias: "bitPopoverAnchorFor" });
+
+  /** Preferred popover position (e.g., "right-start", "below-center") */
```
vleague2, line 60: Should we only allow a string union of accepted positions? (Or keys/ids of `defaultPositions`, whatever)

libs/components/src/popover/popover-anchor-for.directive.ts
```diff
@@ -0,0 +1,204 @@
...
+
+  private getClosedEvents(): Observable<any> {
+    if (!this.overlayRef) {
+      throw new Error("Overlay reference is not available");
+    }
+
+    const detachments = this.overlayRef.detachments();
+    const escKey = this.overlayRef
+      .keydownEvents()
+      .pipe(filter((event: KeyboardEvent) => event.key === "Escape"));
```
vleague2, line 174: I know we had to do some extra things in the menu to make sure that the escape key didnt cause the extension to close too, so maybe we should check that interaction with this component

libs/components/src/popover/popover-anchor-for.directive.ts
```diff
@@ -0,0 +1,204 @@
...
+  }
+
+  constructor(
+    private elementRef: ElementRef<HTMLElement>,
+    private viewContainerRef: ViewContainerRef,
+    private overlay: Overlay,
+  ) {
+    // Wait for the first render to complete so layout is stable before opening.
+    // Sets a signal so the effect below re-evaluates once the layout is ready.
+    afterNextRender(() => this.hasInitialized.set(true));
```
vleague2, line 115: Is this a substitute for the previous double RAF approach?

willmartian: Yes, as far as I can tell everything is working as intended with it. I don't think this API was available at the time when we added the double RAF

libs/components/src/popover/popover-anchor-for.directive.ts
willmartian: This problem with the menu and extension still exists. This is probably something we want to think about and fix outside of this PR (this PR just moves the existing bug to a new file). I will create an issue if one doesn't already exist.

willmartian:
@claude

1. delete the spec files for the popover anchor and trigger
2. address this comment https://github.com/bitwarden/clients/pull/18755#discussion_r2874811739 (improve typing of position input)

libs/components/src/popover/popover-anchor-for.directive.spec.ts
willmartian: Discussed in Slack, deleted tests that weren't doing much since we have Storybook coverage  [a28dac9](https://github.com/bitwarden/clients/pull/18755/commits/a28dac90eea993b000f323f30b199675f9ff5352)

libs/components/src/popover/popover-anchor-for.directive.ts
willmartian: 100%. Resolved here [a28dac9](https://github.com/bitwarden/clients/pull/18755/commits/a28dac90eea993b000f323f30b199675f9ff5352)

libs/components/src/popover/popover-anchor-for.directive.ts
vleague2: We added some code to the `menuTriggerFor` to fix it, is it not working anymore??

libs/components/src/popover/popover-anchor-for.directive.ts
vleague2: Iirc the main bug was that if you initialize the popover as `open` on page load, it was positioned incorrectly and would shift to the correct spot on a close/reopen. So if that's working, then I think we are good? (Also we could try turning the storybook stories back on and see if they are not flaking out still)

libs/components/src/popover/popover-anchor-for.directive.ts
willmartian: Not on firefox prod :(

vleague2: Interesting 🤔 Definitely works on Chrome prod. Maybe worth putting the code in here so that it's at least partially working once this releases? (In addition to ticket to figure out why Firefox is not working)

willmartian: Bug ticket here: https://bitwarden.atlassian.net/browse/CL-1087
Chrome fix applied here: [6c540e1](https://github.com/bitwarden/clients/pull/18755/commits/6c540e1c2361569c321846f7a9dc64e73e833b10)

libs/components/src/popover/popover-anchor-for.directive.ts
willmartian: I think? we are good. I re-enabled the Chromatic snapshots here [8367aee](https://github.com/bitwarden/clients/pull/18755/commits/8367aee0659de1f0611bf11fa8e5e9bb26fc1820)[8367aee](https://github.com/bitwarden/clients/pull/18755/commits/8367aee0659de1f0611bf11fa8e5e9bb26fc1820). We can turn it back off if it still flaky.

PR #18759, opened 2026-02-04
Add KM contact info to HAZMAT

Thomas-Avery:
## 📔 Objective

Add info to contact the Key Management team for new uses of HAZMAT methods in deprecated comment.


quexten (approved):
I wonder if we should add claude instructions around this...

PR #18762, opened 2026-02-04
PM-31733: Sends Drawer Persisting On Side Nav Change

iivins-livefront:
## Objective

The goal of this PR is to address QA Bug raised with the Sends Drawer when navigating using the Side Nav

Before this PR:
- The Send drawer stays open with the Vault page opened along side it.

With this PR:
- Navigating away from Send should close/dismiss the Send drawer so only the vault UI is shown.

No negative regressions found with other methods of closure

## Recording

https://github.com/user-attachments/assets/fb65a5dc-605d-49e0-b03f-933124568697

### DIFF
--- a/apps/desktop/src/app/tools/send-v2/send-v2.component.ts
+++ b/apps/desktop/src/app/tools/send-v2/send-v2.component.ts
@@ -1,14 +1,6 @@
 // FIXME: Update this file to be type safe and remove this and next line
 // @ts-strict-ignore
-import {
-  ChangeDetectorRef,
-  Component,
-  computed,
-  DestroyRef,
-  inject,
-  signal,
-  viewChild,
-} from "@angular/core";
+import { Component, computed, DestroyRef, inject, signal, viewChild } from "@angular/core";
 import { toSignal } from "@angular/core/rxjs-interop";
 import { combineLatest, map, switchMap, lastValueFrom } from "rxjs";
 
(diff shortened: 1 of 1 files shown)


apps/desktop/src/app/tools/send-v2/send-v2.component.ts
```diff
@@ -84,6 +92,10 @@ export class SendV2Component {
   private dialogService = inject(DialogService);
   private toastService = inject(ToastService);
   private logService = inject(LogService);
+  private cdr = inject(ChangeDetectorRef);
```
Hinton: cdr is unused.

apps/desktop/src/app/tools/send-v2/send-v2.component.ts
```diff
@@ -84,6 +92,10 @@ export class SendV2Component {
...
+  private destroyRef = inject(DestroyRef);
+
+  private activeDrawerRef: any = null;
```
Hinton: I think this type is more appropriate?
```suggestion
  private activeDrawerRef?: DialogRef<SendItemDialogResult, SendItemDialogParams>;
```

iivins-livefront: Is it okay to use `private activeDrawerRef?: DialogRef<SendItemDialogResult>;` since `DialogRef` takes `<ResultType, ComponentType>` as its type parameters instead of `<ResultType, DataType>`

`DialogRef` defaults to unknown for the component type which would match how it's used here if I'm understanding it right

Hinton: Yea that should be fine.

PR #18764, opened 2026-02-04
[PM-31675] remove archive from web edit

Jingo88:
## 🎟️ Tracking

[PM-31675](https://bitwarden.atlassian.net/browse/PM-31675)

## 📔 Objective

After discussions we have decided to remove the archive/unarchive buttons from the footer while the user is in the edit form of an item

- refactor how default-cipher-archive-service updates ciphers
- replace the `patchCipher` logic with the existed `onCipherSaved` method
- remove archive/unarchive options from vault-item-dialog-component
- further cleanup to add archive flag checks to archive badges

## 📸 Screen Recording

https://github.com/user-attachments/assets/3751a5eb-a26a-4337-b925-841c3fbf5d00

[PM-31675]: https://bitwarden.atlassian.net/browse/PM-31675?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/common/src/vault/services/default-cipher-archive.service.ts
```diff
@@ -91,21 +91,11 @@ export class DefaultCipherArchiveService implements CipherArchiveService {
...
-
-    for (const cipher of response.data) {
-      const localCipher = localCiphers[cipher.id as CipherId];
-
-      if (localCipher == null) {
-        continue;
-      }
-
-      localCipher.archivedDate = cipher.archivedDate;
-      localCipher.revisionDate = cipher.revisionDate;
```
nick-livefront, line 105: ❓ Wouldn't `revisionDate` still need to be set here?

Jingo88: We're all good here. the data coming back in the response will have everything updated

PR #18775, opened 2026-02-05
[PM-31753] Prevent SDK from disposing withit debounce period

quexten:
## 🎟️ Tracking

https://bitwarden.slack.com/archives/C0321H9M0KS/p1770295996439339
https://bitwarden.atlassian.net/browse/PM-31753

## 📔 Objective

The SDK disposing instantly means many call patterns have a severely negative performance impact that is alleviated by simply adding a debounce / disposal period.

Note: Tested locally against data diagnostics / recovery tool.


coroiu (approved):
Nice to see tests for it too! :)

PR #18777, opened 2026-02-05
[PM-31640] Fix SDK tracing / cipher decryption performance issues

quexten:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-31640

## 📔 Objective

https://github.com/bitwarden/sdk-internal/pull/738
Disables the browser integration for tracing since it was causing performance issues during bulk decryption.


quexten:
Looks like my NPM version was too new, so I had to do a few updates to make this work. Should be fixed @mzieniukbw!

PR #18778, opened 2026-02-05
[PM-31754] Add preinstall script to package.json

0xTeles:


0xTeles:
just testing a potential bug

PR #18787, opened 2026-02-05
[PM-23485] Share types between autofill_provider and napi crates

iinuwa:
## 🎟️ Tracking

[PM-23485](https://bitwarden.atlassian.net/browse/PM-23485)

## 📔 Objective

The `desktop_napi::autofill` module needs to communicate with types also used in `autofill_provider` crate. Right now these are all copied and pasted, with NAPI macros annotating in the `desktop_napi` crate, and `uniffi` annotations in the autofill_provider crate, with `serde` definitions duplicated in both. 

This makes it harder to add new types, as the same definitions need to be done in both files, which gets difficult to keep in sync.

This can also help us improve the de-/serialization. (Currently, we must manually disambiguate types with marker fields rather than using an enum and letting serde do that for us automatically.) I will follow up with another PR that does just that.

This PR:
- moves all the autofill types from `desktop_napi` to `autofill_provider` and adds a dependency on `autofill_provider` to `napi`
- removes the cdylib build of `autofill_provider` crate, which is required to allow napi and uniffi annotations to coexist. (We were already linking staticlib in the autofill extension code, but we were using the cdylib to generate the Swift code. The staticlib works fine for that.)
- bumps the version of `napi` and `napi-build` to versions that use dynamic loading instead of static linking, which allows sharing the types between crates without breaking the workspace build
- changes macOS-specific fields to optional, so that we're not generating different NAPI types for different platforms
- consolidates some duplicated fields on the passkey request types. (See below for context)

For reviewers, I recommend a commit-by-commit review for easier context.

---

We had [previously discussed](https://github.com/bitwarden/clients/pull/18075#discussion_r2676348806) consolidating the common parameters in the three `Passkey(Assertion|AssertionWithoutUserInterface|Registration)Request` types. I considered it further, and with the 4 different representations (Rust, NAPI, UniFFI, Serde), I think that repeating the fields is the least awkward way to do this. It also reads a little better to me to repeat the fields because that's closer to how it's laid out in the spec, but that is subjective. 

I re-ordered the fields on all three request objects to put the common ones together so they're easier to recognize. I also did consolidate the window-related fields into a single type and shared that, since that felt natural.

Finally, this PR only touches native autofill code behind feature/compile-time flags, so this only needs general regression testing.

[PM-23485]: https://bitwarden.atlassian.net/browse/PM-23485?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


neuronull:
FYI the validate linux CI job failure will resolve once merging main in.

neuronull (commented):
I'd like to get the @bitwarden/team-platform 's thoughts on this, as it has architectural implications.

While I empathize with the code-duplication and maintenance/out of sync-ness concern (it's also something I noticed in other contexts and have been thinking about), the approach here is to diverge from our current model of the napi layer being an isolated bridge, and instead introduces a dependency/coupling on one of the desktop_native's sub-crates to napi. There are benefits to that as described in the PR description, but it's fundamentally diverging from our existing architecture.

IMO the feature flag only hides the change in philosophy, it doesn't remediate it. A potential way to remediate it would be to have an additional, separate layer ,that doesn't have the napi or uniffi, and autofill_provider adds the uniffi and napi adds the napi. These are tradeoffs.

iinuwa:
> IMO the feature flag only hides the change in philosophy, it doesn't remediate it. A potential way to remediate it would be to have an additional, separate layer ,that doesn't have the napi or uniffi, and autofill_provider adds the uniffi and napi adds the napi. These are tradeoffs.

Could you describe a little more about how this would work? Do you mean something like a base struct that just has serde derives, and then each of the other crates also provide separate structs with a `impl From<BaseStruct> for MyStruct` or something?

coltonhurst (dismissed):
@neuronull thanks for your comments [above](https://github.com/bitwarden/clients/pull/18787#pullrequestreview-3842672628). I would agree; I think keeping ownership within an isolated bridge is important.

@iinuwa you have a valid point, and I appreciate you bringing it up with a solution! There are other routes we could take that maybe alleviate the duplication, etc. you are mentioning, but maybe we should align on the above first.

Any thoughts @coroiu @dani-garcia @Hinton ?

coltonhurst:
> > IMO the feature flag only hides the change in philosophy, it doesn't remediate it. A potential way to remediate it would be to have an additional, separate layer ,that doesn't have the napi or uniffi, and autofill_provider adds the uniffi and napi adds the napi. These are tradeoffs.
> 
> Could you describe a little more about how this would work? Do you mean something like a base struct that just has serde derives, and then each of the other crates also provide separate structs with a `impl From<BaseStruct> for MyStruct` or something?

I would imagine this might be a shared "core" of types, or something. If we do go this route we have some other brewing initiatives in the desktop native space that I think we should align on first as they may impact this.

iinuwa:
Summary: If moving these types to `desktop_core` with serde annotations and creating duplicated types in the `autofill_provider` and `napi` crates is acceptable, since this pattern already exists, then I'll close this and move forward with that solution instead.

> the approach here is to diverge from our current model of the napi layer being an isolated bridge, and instead introduces a dependency/coupling on one of the desktop_native's sub-crates to napi. There are benefits to that as described in the PR description, but it's fundamentally diverging from our existing architecture.

Another perspective on this: we already do have dependencies on the subcrates for function calls, and some of the crates also depend on types; albeit with `From` implementations between the NAPI types (like the `biometrics` module).

I agree that this does change how we interact with the NAPI crate. Currently, the `napi` crate has a few functions:

- It is a central place to run the TypeScript generation process.
- It "curates" the list of types that need to be shared over Electron FFI.
- (It could also provide extra transformations to make TypeScript types more ergonomic, but we don't currently do that.)

I think a difference between this situation is that it wasn't possible to have the NAPI annotations in the crates where they were defined due to NAPI limitations. Now that this is possible, we have the opportunity to reconsider.

It colocates the changes to the types, and for types that are merely a transparent shim from Node to Rust, it reduces the cognitive load for developers making changes here. It does have the disadvantage that a change in another crate would change the `napi/index.d.ts`, though I think that's pretty clear by the evidence of the `#[napi]` annotation.

> A potential way to remediate it would be to have an additional, separate layer ,that doesn't have the napi or uniffi, and autofill_provider adds the uniffi and napi adds the napi. These are tradeoffs.

Yeah. I think this is the least optimal from a code duplication perspective, but it provides the most flexibility of type representation in each of the APIs. At this time, I don't have a need for that much flexibility, but I could see that being useful.

> I would imagine this might be a shared "core" of types, or something. If we do go this route we have some other brewing initiatives in the desktop native space that I think we should align on first as they may impact this.

This decision is fairly reversible, so if there are other considerations I'm not aware of, I'm fine to continue existing patterns and re-align later. So if moving the base types into `desktop_core` with serde implementations, and then duplicating the types in `autofill_provider` and `napi` with `#[uniffi]` and `#[napi]` annotations, then that's OK with me and would unblock other work that needs to be done here.

neuronull:
Thanks for this further discussion and details!

My 2c would be to work within existing paradigms where possible, while implementing something like a POC/mvp for passkeys, which already introduces enough novel code on it's own. And to re-visit paradigm shifts in a more focused effort. This makes it easier to reason about the novel feature and reduces surface area of added risk.  There can be exceptions to that approach of course, sometimes the novel feature requires doing something that doesn't fit at all with existing paradigms, though that doesn't seem to be the case here.

If there is a strong conviction in this new approach though, I'd assert that we'd get discussion on it with folks from the Platform team and others with more history/scope (speaking for myself) with the project.

Hinton:
I believe there were previously a limitation in Napi that prevented re-exporting symbols from a dependency. I.e. if you added napi to a struct in another crate it wouldn't get exposed. I believe that was fixed in Napi 3. I'm generally in favor of not duplicating the structs which is how it works in the Rust SDK.

Using feature flags here to conditionally apply it seems appropriate but we should make an attempt at cleaning up the other places too.

coroiu:
My understanding is that this moves code out of `napi` and into the feature creates. In that case it mirrors the SDK pattern where the FFI layer is kept as thin as possible and types are defined where they live, not in the bridge. I think related code should live as close together as possible, so I fully endorse this approach.

neuronull:
Thanks for weighing in both of you!

> Using feature flags here to conditionally apply it seems appropriate but we should make an attempt at cleaning up the other places too.

I wonder if this deserves an epic, or a multiple user stories? Currently the linked ticket in the PR is a Jira Task from an Epic. It seems like it'd be worthwhile to at least consider making it one or more user stories, especially if aligning the other places that would change to use this model. User stories are also more appropriate for tracking work that needs to be QA'd (IME).

Also since it'd presumably be touching more than just Autofill in that case, wondering what team would make sense to own the parent item. (Currently the linked ticket is in the Autofill team).

melihemreguler:
👋 Following the macOS native AutoFill work here — the type consolidation in this PR looks like it would also make adding a *password* credential type (`ASPasswordCredentialRequest` / `ProvidesPasswords`) cleaner, since today the extension only provides passkeys.

There's community demand for password AutoFill on macOS (https://community.bitwarden.com/t/autofill-with-macos-26-tahoe/89661), and I'd be happy to contribute the password-provisioning path (a `prepare_password_assertion` flow mirroring the passkey assertion one, plus `ProvidesPasswords` + Swift handling), behind the existing `macos-native-credential-sync` flag.

Is that something you'd welcome as an external contribution, and would you prefer I wait for this refactor to land first so I build on the consolidated types rather than the current duplicated ones? Opened a discussion to track: https://github.com/orgs/bitwarden/discussions/21537

iinuwa:
Hi @melihemreguler! Thanks for the offer! I've actually done an experiment internally to get password autofill up and running, but the rest of the code isn't ready to ship yet. I'm excited to get there though.

We've got a few more refactorings to do after this one before we can get there. Since we're tackling the Windows and macOS autofill implementations concurrently, I'll say you should hold off until things settle a bit later to avoid having to rework the PR. I'll keep in touch in the discussion you opened.

abergs (approved):

addisonbeck (approved):

PR #18788, opened 2026-02-05
Remove unused styles for datetime input wrapper in tailwind.css

JaredScar:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-31792

## 📔 Objective

Fix CSS issue which affects all bit-form-fields

### DIFF
--- a/apps/web/src/scss/tailwind.css
+++ b/apps/web/src/scss/tailwind.css
@@ -93,13 +93,6 @@
     -webkit-appearance: none;
     appearance: none;
   }
-
-  /* Wrapper for datetime input to ensure proper positioning */
-  bit-form-field .tw-relative {
-    display: flex;
-    align-items: stretch;
-  }
-
   /* Ensure datetime inputs have proper padding and alignment */
   .datetime-input {
     padding-right: 2.5rem !important;


vleague2:
I know the styles were added in the previous PR to solve the bug from the original ticket, but we are ideally not adding new css to this file (as per the comment at the top of the `tailwind.css` file). What was the reason the new styles needed to be added? Do tailwind classes not work?

JaredScar:
> I know the styles were added in the previous PR to solve the bug from the original ticket, but we are ideally not adding new css to this file (as per the comment at the top of the `tailwind.css` file). What was the reason the new styles needed to be added? Do tailwind classes not work?

Posting here for visibility as well. The usual date input icons actually are OS specific and don't actually come from Bitwarden. The PR from before for the original ticket is to remove the OS specific icon and implement our own which Bitwarden can control. Where the CSS lives for this is probably okay to be moved though outside of the tailwind.css file if preferred?

JaredScar:
Reverting original PR, so closing this one for now. Seems to be an issue throughout the web app and not only on the event logs page of admin console. This issue will be revisited at a later date

PR #18798, opened 2026-02-05
[CL-973] chip component

BryanCunningham:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/CL-973

### Adds new chip components
- chip-filter: Replace chip-select
- chip-action: Used for interactive chips
- chip: Static chips that are dismissible. Primary use is in multi-select

Note to external reviewing teams:
- Diffs will be a rename of "Chip Select" --> "Chip Filter"

### Chip
image

### Chip Action
image

### Chip Filter
image

### DIFF
--- a/libs/components/src/chips/shared/shared-story-arg-types.ts
+++ b/libs/components/src/chips/shared/shared-story-arg-types.ts
@@ -5,15 +5,6 @@ import { BITWARDEN_ICONS } from "../../shared/icon";
 import { CHIP_SIZES, CHIP_VARIANTS } from "./base-chip.directive";
 
 const sharedArgTypes = {
-  size: {
-    options: CHIP_SIZES,
-    control: { type: "select" },
-    description: "Sets the size of the chip.",
-    table: {
-      type: { summary: CHIP_SIZES.join(" | ") },
-      defaultValue: { summary: "large" },
-    },
-  },
   startIcon: {
     control: "select",
     options: BITWARDEN_ICONS,
@@ -22,6 +13,18 @@ const sharedArgTypes = {
   disabled: {
     control: "boolean",
     description: "Disables the chip",
+  },
+} satisfies Partial<ArgTypes>;
+
+const sizeArgType = {
+  size: {
+    options: CHIP_SIZES,
+    control: { type: "select" },
+    description: "Sets the size of the chip.",
+    table: {
+      type: { summary: CHIP_SIZES.join(" | ") },
+      defaultValue: { summary: "large" },
+    },
   },
 } satisfies Partial<ArgTypes>;
 
@@ -52,4 +55,4 @@ const variantArgType = {
   },
 } satisfies Partial<ArgTypes>;
 
-export { sharedArgTypes, variantArgType, endIconArgType, fullWidthArgType };
+export { sizeArgType, sharedArgTypes, variantArgType, endIconArgType, fullWidthArgType };
--- a/libs/components/src/chips/chip-action/chip-action.stories.ts
+++ b/libs/components/src/chips/chip-action/chip-action.stories.ts
@@ -1,6 +1,6 @@
 import { Meta, StoryObj, moduleMetadata } from "@storybook/angular";
 
-import { sharedArgTypes, variantArgType } from "../shared/shared-story-arg-types";
+import { sharedArgTypes, variantArgType, sizeArgType } from "../shared/shared-story-arg-types";
 
 import { ChipActionComponent } from "./chip-action.component";
 
@@ -19,6 +19,7 @@ export default {
   argTypes: {
     ...sharedArgTypes,
     ...variantArgType,
+    ...sizeArgType,
   },
 } as Meta;
 
--- a/libs/components/src/chips/chip/chip.stories.ts
+++ b/libs/components/src/chips/chip/chip.stories.ts
@@ -3,7 +3,7 @@ import { Meta, StoryObj, moduleMetadata } from "@storybook/angular";
 import { I18nService } from "@bitwarden/common/platform/abstractions/i18n.service";
 
 import { I18nMockService } from "../../utils";
-import { sharedArgTypes } from "../shared/shared-story-arg-types";
+import { sharedArgTypes, sizeArgType } from "../shared/shared-story-arg-types";
 
 import { ChipComponent } from "./chip.component";
 
@@ -31,6 +31,7 @@ export default {
   },
   argTypes: {
     ...sharedArgTypes,
+    ...sizeArgType,
   },
 } as Meta;
 


libs/components/src/chip/chip-action/chip-action.component.ts
```diff
@@ -0,0 +1,36 @@
...
+  type ChipVariant,
+  type ChipSize,
+  ChipVariants,
+  ChipSizes,
+} from "../base-chip.directive";
+import { ChipContentComponent } from "../chip-content.component";
+
+@Component({
+  selector: "a[bitChipAction], button[bitChipAction]",
+  standalone: true,
```
vleague2: Isn't this the default now?

libs/components/src/chip/chip-action/chip-action.component.ts
```diff
@@ -0,0 +1,36 @@
...
+      inputs: ["variant", "size", "disabled", "fullWidth"],
+    },
+  ],
+})
+export class ChipActionComponent {
+  readonly variant = input<ChipVariant>(ChipVariants.Primary);
+  readonly size = input<ChipSize>(ChipSizes.Large);
+
+  readonly disabled = input<boolean, unknown>(false, { transform: booleanAttribute });
+  readonly fullWidth = input<boolean, unknown>(false, { transform: booleanAttribute });
```
vleague2: Not a blocker -- I think it was like this before your PR, but I don't know that we need to specify the generic type here. I think Angular inference works pretty well for these

libs/components/src/chip/chip-action/chip-action.mdx
```diff
@@ -0,0 +1,193 @@
...
+
+Action chips should be **secondary to the primary content or actions** on the screen.
+
+## Chip Types
+
+There are three types of chips in the component library:
+
+### Action chip
+
+**This component is the action chip variant.**
```
vleague2: Suggestion for docs organization -- the top-level `chip.mdx` page could be the  overarching documentation about chips in general and how we have 3 variants. Then all 3 individual variants can have nested docs with much more focused (and shorter) docs for themselves that don't need to include info about chips overall or the other variants.

libs/components/src/chip/chip-filter/chip-filter.component.html
```diff
@@ -0,0 +1,99 @@
...
+    >
+      @if (getParent(renderedOptions); as parent) {
+        <button
+          type="button"
+          bitMenuItem
+          (click)="viewOption(parent, $event)"
+          class="tw-text-[length:inherit]"
+          [title]="'backTo' | i18n: parent.label ?? placeholderText()"
+        >
+          <i slot="start" class="bwi bwi-angle-left" aria-hidden="true"></i>
```
vleague2: Should we swap to `bit-icon` for these while we're in here?

libs/components/src/chip/base-chip.directive.ts
```diff
@@ -0,0 +1,155 @@
...
+  "tw-border",
+  "tw-font-medium",
+  "tw-transition-colors",
+
+  // Button-specific resets (when applied to button elements)
+  "[&:is(button)]:tw-appearance-none",
+  "[&:is(button)]:tw-outline-none",
+  ...focusRing,
+  ...inactiveStyles,
+  // "[&:is(button)]:tw-bg-transparent",
```
vleague2: Is this supposed to be commented out?

libs/components/src/chip/chip-dismiss-button.component.ts
```diff
@@ -0,0 +1,27 @@
+import { ChangeDetectionStrategy, Component, ElementRef, inject, input } from "@angular/core";
+
+import { IconComponent } from "../icon/icon.component";
+import { ariaDisableElement } from "../utils/aria-disable-element";
+import { AriaDisableDirective } from "../a11y/aria-disable.directive";
```
vleague2: Lint error here

libs/components/src/chips/shared/shared-story-arg-types.ts
```diff
@@ -0,0 +1,52 @@
...
+
+import { BITWARDEN_ICONS } from "../shared/icon";
+
+import { ChipSizes, ChipVariants } from "./base-chip.directive";
+
+const sharedArgTypes = {
+  size: {
+    options: Object.values(ChipSizes),
+    control: { type: "select" },
+    description: "Sets the size of the chip.",
```
vleague2: Question -- if you put these descriptions as jsdoc comments above the actual inputs on the component file, do they show up in the storybook docs? I think having it in the code (and available to intellisense) would be more valuable than just specifying in the storybook docs, assuming it would propagate to the docs

libs/components/src/chip/chip-action/chip-action.component.ts
BryanCunningham: Yes, I believe so. I'll remove

libs/components/src/chip/chip-filter/chip-filter.component.html
BryanCunningham: Yep! Good call

libs/components/src/chip/chip-action/chip-action.mdx
BryanCunningham: @vleague2 I updated the docs. Does it look more like what you were thinking?

libs/components/src/chips/shared/shared-story-arg-types.ts
BryanCunningham: Unfortunately, they won't be included in the Storybook docs due to how host directives work with Storybook. We'll just have to dupe them in JSDoc comments

libs/components/src/chip/chip-action/chip-action.component.ts
```diff
@@ -0,0 +1,25 @@
+import { Component, ChangeDetectionStrategy, input, inject } from "@angular/core";
+
+import { BitwardenIcon } from "../../shared/icon";
+import { BaseChipDirective } from "../base-chip.directive";
+import { ChipContentComponent } from "../chip-content.component";
+
+@Component({
+  selector: "a[bitChipAction], button[bitChipAction]",
```
willmartian: Components should use kebab-case, camelCase is reserved for directives. The distinction is helpful, because an element can only support a single component, but multiple directives. 

(There are some places were we don't do this because the component was once a directive and we never migrated the usage, but those should be fixed at some point, and new components should follow it).

libs/components/src/chip/chip-action/chip-action.mdx
```diff
@@ -0,0 +1,157 @@
...
+  <button bitChipAction label="Export" startIcon="bwi-download"></button>
+</div>
+```
+
+## Choosing Between Button and Link
+
+Choose the appropriate HTML element based on the action:
+
+- **Use `<button>`** when the action affects the current page (filtering, sorting, opening a dialog)
+- **Use `<a>`** when the action navigates to a new page or external resource
```
willmartian: This is redundant with an above section

libs/components/src/chip/chip-action/chip-action.mdx
```diff
@@ -0,0 +1,157 @@
...
+
+## Accessibility
+
+- The component includes proper `aria-disabled` attributes when disabled
+- Use descriptive labels that clearly communicate the action
+- Ensure sufficient color contrast when using chips on different backgrounds
+- For icon-only chips (if the label is empty), ensure the icon conveys clear meaning or add an
+  `aria-label`
+- When using as a link, ensure the href attribute is properly set for keyboard navigation
```
willmartian: Angular router links do not use href, maybe just leave this out so it doesn't consume a dev or LLM

libs/components/src/chips/chip-filter/chip-filter.component.html
willmartian, line 1: What changed about this file beyond the renames? Just the usage of `bitBaseChip`?

libs/components/src/chips/chip-filter/chip-filter.component.ts
```diff
@@ -150,8 +166,8 @@ export class ChipSelectComponent<T = unknown> implements ControlValueAccessor {
   }
 
   /** The icon to show in the chip button */
-  protected get icon(): string | undefined {
-    return this.selectedOption?.icon || this.placeholderIcon();
+  protected get icon(): BitwardenIcon | undefined {
+    return (this.selectedOption?.icon || this.placeholderIcon()) as BitwardenIcon | undefined;
```
willmartian: Why is this typecasting necessary?

libs/components/src/chips/shared/chip-content.component.ts
```diff
@@ -0,0 +1,48 @@
...
+
+import { IconComponent } from "../icon";
+import { BitwardenIcon } from "../shared/icon";
+
+import { ChipSize, ChipSizes } from "./base-chip.directive";
+
+/**
+ * `<bit-chip-content>` is a content wrapper component that provides consistent chip styling
+ * and layout with support for icons, custom content slots, and dismiss functionality.
+ */
```
willmartian, line 13: Can you add a brief note like "@internal - only to be used within lib/components" to this and the other helper directive

libs/components/src/chip/base-chip.directive.ts
```diff
@@ -0,0 +1,153 @@
...
+} as const;
+
+export type ChipVariant = (typeof ChipVariants)[keyof typeof ChipVariants];
+
+export const ChipSizes = {
+  Small: "small",
+  Large: "large",
+} as const;
+
+export type ChipSize = (typeof ChipSizes)[keyof typeof ChipSizes];
```
willmartian: Could these be updated to use unions instead of enum-likes? 

`type ChipVariant = "primary" | "subtle" ...`

It is less boiler plate, and makes it easier to consume in downstream code. 

(Let me know if you disagree!)

libs/components/src/chips/shared/chip-dismiss-button.component.ts
willmartian, line 1: Do you think there is a good reason for this to be its own component vs inlined in `chip.component.html`?

willmartian (changes requested):
After resolving all other code changes, could you reorg the folder a bit?

- chips/
  - chip-action/
  - chip-filter/
  - chip/
  - shared/
  - chips.mdx
  - index.ts

libs/components/src/chips/shared/chip-dismiss-button.component.ts
BryanCunningham: @willmartian It's used in both the chip component and the chip-filter. Figured it made sense for it to be it's own thing

libs/components/src/chip/base-chip.directive.ts
BryanCunningham: @willmartian I personally like the ability to use these as enum like `ChipSize.Small`. But, I don't have super strong feelings about it. Just a habit of my personal preference

libs/components/src/chips/chip-filter/chip-filter.component.ts
BryanCunningham: @willmartian There is a bunch of upstream types that would have to be changed(string => BitwardenIcon) in order to make this typesafe. Wasn't sure how deep down the rabbit hole we wanted to go here. Or, if we wanted to leave this a tech debt thing and clean it up later

libs/components/src/chips/shared/chip-dismiss-button.component.ts
willmartian: Made sense, missed that. In that case, I would also mark it as `@internal` and move to the shared folder I suggested.

libs/components/src/chips/chip-filter/chip-filter.component.html
BryanCunningham: @willmartian Yeah, added the internal directive usages. And used the now shared dismiss button. And there were some minor html structure changes and some minor style changes

libs/components/src/chip/chip-action/chip-action.component.ts
BryanCunningham: @willmartian Just so I'm clear, you are expecting `a[bit-chip-action]... ` here?

libs/components/src/chip/base-chip.directive.ts
willmartian: Hmm, I think the main reason I don't like it is because now there are two ways to reference a size, either `ChipSizes.Small` or `"small"`.  So one might be inclined to find all references of `ChipSizes` when doing a refactor, and then miss instances which are just using `"small"`. 

That could be remedied by using a proper enum, but there are other downsides to that (can't be stripped by Typescript, can't be used in templates, etc). 

I think trimming multiple ways of doing something into one way is usually the best path--unless the benefits are really clear--so I would recommend removing in favor of type unions

libs/components/src/chip/chip-action/chip-action.component.ts
willmartian: Yes, similar to `a[bit-item-content]...`

This is mentioned in the Angular style guide here and here: 
- https://v19.angular.dev/style-guide#style-05-02
- https://v19.angular.dev/style-guide#style-02-06

**Note:** it looks like it was removed in the style guides past v19. I am not sure if this was intentional, or due to the remake of the docs site that occurred around then. I am thinking it is the latter, because Angular Material still follows it:

```html
<table mat-table [dataSource]="myDataArray">
  ...
</table>
```

BryanCunningham: Can do

libs/components/src/chips/chip-action/chip-action.mdx
```diff
@@ -0,0 +1,141 @@
...
+
+### End Icon
+
+```html
+<button bit-chip-action label="Next" endIcon="bwi-arrow-right"></button>
+```
+
+<Canvas of={stories.WithEndIcon} />
+
+## Disabled State
```
vleague2: Inactive?

libs/components/src/chips/shared/base-chip.directive.ts
```diff
@@ -109,37 +108,44 @@ export class BaseChipDirective {
...
+
+  /** Disabled state of the chip */
+  readonly disabled = input(false, { transform: booleanAttribute });
 
   /**
    * Computed class list based on variant, size, and state
    */
   protected readonly classList = computed(() => {
-    // Add size classes
+    const maxWidth = this.maxWidthClass() || "tw-max-w-52";
```
vleague2: If we're already using a default in `maxWidthClass()`, do we need one here? Also they are not the same default

libs/components/src/chips/shared/index.ts
```diff
@@ -0,0 +1,4 @@
+export * from "./base-chip.directive";
+export * from "./chip-content.component";
+export * from "./chip-dismiss-button.component";
+// Note: shared-story-arg-types.ts not exported (internal to stories)
```
vleague2: I don't know that we need this comment, I can't imagine anyone will be looking for it

libs/components/src/chips/shared/base-chip.directive.ts
```diff
@@ -1,33 +1,30 @@
...
-
-export type ChipSize = (typeof ChipSizes)[keyof typeof ChipSizes];
+// Helper constants for Storybook and default values
+export const CHIP_VARIANTS: ChipVariant[] = [
+  "primary",
+  "subtle",
+  "accent-primary",
+  "accent-secondary",
+];
+export const CHIP_SIZES: ChipSize[] = ["small", "large"];
```
vleague2: Are these two constants only used for Storybook? You could do like `ChipSize` as `(typeof CHIP_SIZES)[number]` so that you don't have to repeat the options in the type and can still use the constants for Storybook

libs/components/src/chips/chips.mdx
```diff
@@ -115,15 +143,15 @@ Chips are available in two sizes: small and large (default).
 <bit-chip label="Large" size="large"></bit-chip>
 ```
 
-## Disabled state
+### Disabled State
```
vleague2: Same question about inactive

libs/components/src/chips/shared/base-chip.directive.ts
BryanCunningham: Good call. Fixed

libs/components/src/chips/shared/index.ts
BryanCunningham: Removed :)

libs/components/src/chips/chips.mdx
BryanCunningham: Yeah, should be inactive. Will replace

libs/components/src/chips/chip-filter/chip-filter.component.ts
willmartian: Ah gotcha, fine to leave as tech debt.

libs/components/src/chips/shared/chip-content.component.html
```diff
@@ -0,0 +1,22 @@
+<div
+  [class]="
+    'tw-inline-flex tw-min-w-0 tw-w-full tw-max-w-full tw-items-center tw-justify-between ' +
+    gapClass()
+  "
+>
+  <ng-content select="[chipStart]"></ng-content>
```
vleague2: We usually do slots for these, like `select="[slot=chipStart]"` so that the selectors don't get confused with component or directive selectors

(This and `chipEnd`)

libs/components/src/chips/chip-filter/chip-filter.mdx
```diff
@@ -1,27 +1,29 @@
...
 
 Options are passed to the select via an `options` input.
 
 ```ts
 @Component({
-  selector: `
-    <bit-chip-select options=[options]></bit-chip-select>
-  `,
+  selector: "your-component",
+  template: ` <bit-chip-filter options="[options]"></bit-chip-filter> `,
```
vleague2: Ooooh looking at this example, I just realized that switching over to `bit-icon` means that the code below wouldn't work anymore (specifically the one with `icon: "bwi-exclamation-triangle tw-text-danger",`) because it's passing the color class but now the component only accepts the icon name. Do we know if we have any instances where chip select options are passing in anything except the icon name? If we're good, we can just remove that color class from the example below

libs/components/src/chips/chips.mdx
```diff
@@ -0,0 +1,169 @@
...
+<bit-chip label="Tag name" startIcon="bwi-tag" (chipDismissed)="handleRemove()"> </bit-chip>
+```
+
+```ts
+@Component({
+  selector: "app-my-component",
+  template: `
+    <div class="tw-flex tw-flex-wrap tw-gap-2">
+      <bit-chip
+        *ngFor="let tag of tags"
```
vleague2: Probably should use modern for loop syntax here

vleague2 (commented):
Assorted thoughts

libs/components/src/chips/chip-filter/chip-filter.mdx
BryanCunningham: @vleague2 I didn't change anything about the menu. So these option icons render as they did previously

vleague2: 🤦 You're right, disregard

vleague2:
There are build errors related to the icon name being passed to `bit-icon` from chip select

BryanCunningham:
> There are build errors related to the icon name being passed to `bit-icon` from chip select

@vleague2 Yeah, `option.icon` is a string and it needs to be `BitwardenIcon` I don't have time to get that fix in right now unfortunately. 😢

libs/components/src/chips/chip-action/chip-action.mdx
```diff
@@ -0,0 +1,133 @@
...
+
+The accent primary variant uses accent brand colors.
+
+```html
+<button bit-chip-action label="Info" variant="accent-primary" startIcon="bwi-info-circle"></button>
+```
+
+### Accent Secondary
+
+The accent secondary variant uses secondary accent colors.
```
willmartian: All of these variant doc sections feel kind of redundant to me, maybe just inline the all variant story?

libs/components/src/chips/chip-action/chip-action.mdx
```diff
@@ -0,0 +1,133 @@
...
+```html
+<button bit-chip-action label="Inactive" [disabled]="true" startIcon="bwi-filter"></button>
+```
+
+<Canvas of={stories.Inactive} />
+
+## Groups of Action Chips
+
+When displaying multiple action chips together, use consistent spacing:
+
```
willmartian: This feels superfluous unless we are giving a clear guide "always use X size" or "always use Y grouping component"

libs/components/src/chips/chip-filter/chip-filter.component.html
```diff
@@ -0,0 +1,106 @@
+<!-- removing padding here and adding it to the button for proper menu alignment -->
+<div
+  class="!tw-ps-0 !tw-pb-0"
+  bitBaseChip
```
willmartian: > removing padding here and adding it to the button for proper menu alignment

Do you think this warrants moving these styles off of the shared directive?

Actually, maybe my question is instead, why is `bitBaseChip` defined here instead of as a host directive? It is weird that we have a host directive _and_ input drilling

libs/components/src/chips/chip-filter/chip-filter.component.ts
```diff
@@ -18,54 +16,75 @@ import { ControlValueAccessor, NG_VALUE_ACCESSOR } from "@angular/forms";
...
+import { TypographyModule } from "../../typography";
+import { BaseChipDirective, ChipSize } from "../shared/base-chip.directive";
+import { ChipContentComponent } from "../shared/chip-content.component";
+import { ChipDismissButtonComponent } from "../shared/chip-dismiss-button.component";
+
+/** An option that will be showed in the overlay menu of `ChipFilterComponent` */
+export type ChipFilterOption<T> = Omit<Option<T>, "icon"> & {
   /** The options that will be nested under this option */
-  children?: ChipSelectOption<T>[];
+  children?: ChipFilterOption<T>[];
```
willmartian, line 35: Issue: we omit `icon` and then add it back.

Question: is `iconClass` being added so we can just give a color to the icon?

libs/components/src/chip-select/chip-select.component.ts
```diff
@@ -77,28 +96,15 @@ export class ChipSelectComponent<T = unknown> implements ControlValueAccessor {
...
-   */
-  protected readonly focusVisibleWithin = signal(false);
-  @HostListener("focusin", ["$event.target"])
-  onFocusIn(target: HTMLElement) {
-    this.focusVisibleWithin.set(target.matches("[data-fvw-target]:focus-visible"));
-  }
-  @HostListener("focusout")
-  onFocusOut() {
-    this.focusVisibleWithin.set(false);
-  }
```
willmartian, line 94: Is this just not needed anymore?

libs/components/src/chips/chip-filter/chip-filter.mdx
```diff
@@ -44,7 +46,7 @@ class MyComponent {
 }
 ```
 
-### Option Trees
+### Option Trees (Nested Options)
```
willmartian: Probably can just go with 1 term or the other

libs/components/src/chips/shared/chip-content.component.ts
```diff
@@ -0,0 +1,54 @@
...
+import { ChipSize } from "./base-chip.directive";
+
+/**
+ * `<bit-chip-content>` is a content wrapper component that provides consistent chip styling
+ * and layout with support for icons, custom content slots, and dismiss functionality.
+ *
+ * @internal only to be used within lib/components
+ */
+@Component({
+  selector: "bit-chip-content, [bitChipContent]",
```
willmartian: `[bit-chip-content]` is the correct style for components

libs/components/src/chips/shared/chip-content.component.ts
```diff
@@ -0,0 +1,54 @@
...
+  standalone: true,
```
willmartian: Standalone is already true by default

libs/components/src/chips/chips.mdx
```diff
@@ -0,0 +1,166 @@
...
+
+## Content Guidelines
+
+Chip copy should be **1-2 words** for optimal scannability.
+
+If necessary, chip text should always be truncated with ellipses, not wrapped to a second line.
+
+---
+
+## Base Chip Component (`bit-chip`)
```
willmartian: Can we remove references to "base" here? That is the terminology used with our internal directive.

libs/components/src/chips/chips.mdx
```diff
@@ -0,0 +1,166 @@
...
+}
+```
+
+### Dismissing Chips
+
+The chip component includes a built-in dismiss button (X icon) that emits a `chipDismissed` event
+when clicked. Handle this event to remove the chip from your collection.
+
+```html
+<bit-chip label="Removable tag" (chipDismissed)="handleDismiss()"> </bit-chip>
```
willmartian: `(chipDismissed)` is redundant, can this be simple `(dismissed)`?

libs/components/src/chips/chips.mdx
willmartian: I think this whole section needs to be moved to `/chips/chip/chip.mdx`

libs/components/src/chips/chips.mdx
```diff
@@ -0,0 +1,166 @@
...
+```html
+<bit-chip label="User" startIcon="bwi-user"> </bit-chip>
+
+<bit-chip label="Folder" startIcon="bwi-folder"> </bit-chip>
+```
+
+### Accessibility
+
+- The chip includes proper `aria-disabled` attributes when disabled
+- The dismiss button includes an internationalized aria-label for screen readers
```
willmartian: Should we provide info that doesn't inspire action?

libs/components/src/chips/chip-filter/chip-filter.component.ts
vleague2: I'm adding it back as a different type. I don't want to make the typing change to the `Option` because it's used in the select component, which feels out of scope.

And yes, there is one existing use case where we're using color on the icon. Based on the upcoming designs for menu items, we're going to have to support some kind of color specification (in the new designs it looks like it's the whole item, not just the icon). My thought was we can support the existing use case for now and revisit once we get to the larger menu item changes, but we could also start to address it here.

libs/components/src/chips/chips.mdx
vleague2: Removed in [ffaba9b](https://github.com/bitwarden/clients/pull/18798/commits/ffaba9b50b361d4b8ae898d44e79c03d4184571b)

libs/components/src/chips/chips.mdx
vleague2: Moved in [ffaba9b](https://github.com/bitwarden/clients/pull/18798/commits/ffaba9b50b361d4b8ae898d44e79c03d4184571b)

vleague2: Removed in [ffaba9b](https://github.com/bitwarden/clients/pull/18798/commits/ffaba9b50b361d4b8ae898d44e79c03d4184571b)

libs/components/src/chips/chip-filter/chip-filter.mdx
vleague2: Picked one in [ffaba9b](https://github.com/bitwarden/clients/pull/18798/commits/ffaba9b50b361d4b8ae898d44e79c03d4184571b)

libs/components/src/chips/chip-action/chip-action.mdx
vleague2: Removed in [ffaba9b](https://github.com/bitwarden/clients/pull/18798/commits/ffaba9b50b361d4b8ae898d44e79c03d4184571b)

libs/components/src/chips/chip-action/chip-action.mdx
vleague2: Updated in [ffaba9b](https://github.com/bitwarden/clients/pull/18798/commits/ffaba9b50b361d4b8ae898d44e79c03d4184571b)

libs/components/src/chips/chip-filter/chip-filter.mdx
```diff
@@ -1,26 +1,37 @@
...
 @Component({
-  selector: `
-    <bit-chip-select options=[options]></bit-chip-select>
+  selector: "your-component",
+  template: `
+    <bit-chip-filter
+      options="[options]"
+      placeholderText="Folder"
+      placeholderIcon="bwi-folder"
+    ></bit-chip-filter>
```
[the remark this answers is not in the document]
vleague2: Fixed in [ccd954b](https://github.com/bitwarden/clients/pull/18798/commits/ccd954b5f28f17756708215c53a97378c447e0bb)

libs/components/src/chips/shared/chip-content.component.ts
vleague2: Fixed in [90e03e4](https://github.com/bitwarden/clients/pull/18798/commits/90e03e44a1e99ad8072101bcb91f5d178f175c23)

libs/components/src/chips/shared/chip-content.component.ts
vleague2: Fixed in [6ee7ed5](https://github.com/bitwarden/clients/pull/18798/commits/6ee7ed5fd94ccde1e3f5fe40579a34116e3dcd02) for all components in /chip

libs/components/src/chips/chips.mdx
vleague2: Fixed in [c9e22cb](https://github.com/bitwarden/clients/pull/18798/commits/c9e22cb1d894c7c445375195d96d7345efbbfd4c)

libs/components/src/chip-select/chip-select.component.ts
vleague2: It looks to me like it is being handled with a different tailwind selector checking for child elements with focus-visible (currently lives in the base chip directive), like `[&:has(:focus-visible:not(button[bit-chip-dismiss-button]))]:tw-ring-2`

libs/components/src/chips/chip-filter/chip-filter.component.html
BryanCunningham: Let me look into the host directive approach. I _think_ this is a remnant of an initial implementation that had a different HTML structure. 

I'll explore moving these padding declarations to the content component instead.

BryanCunningham: @willmartian I moved the base chip to a host directive. I did have to keep the special styling in here though. Because the menu aligns to it's trigger, the menu doesn't align to the far left as it should. I could add the padding to the 'content' container but that creates a new problem because we have nested content containers for spacing consistency.

libs/components/src/chips/shared/base-chip.directive.ts
```diff
@@ -103,6 +111,12 @@ export class BaseChipDirective {
    */
   readonly variant = input<ChipVariant>("primary");
 
+  /** Internal variant state (programmatic control) - writable signal */
+  readonly variantState = signal<ChipVariant | null>(null);
+
+  /** Combined variant state from both input and programmatic control */
+  readonly resolvedVariant = computed(() => this.variantState() ?? this.variant());
```
vleague2: Curious about this approach over using a `model` input that can be written to?

BryanCunningham: @vleague2 I can change to a model if we want. Was just follwoing the convention I already started with the disabled input/model

willmartian: I would prefer updating this to be a model. I think the disabled input is a special case because of needing to support the boolean attribute transform for backwards compatibility.

libs/components/src/chips/shared/base-chip.directive.ts
```diff
@@ -0,0 +1,201 @@
...
+  readonly isSelected = computed(() => this.selected() || this.selectedState());
+
+  /** Chip will stretch to full width of its container */
+  readonly fullWidth = input(false, { transform: booleanAttribute });
+
+  /**
+   * Tailwind max-width class to apply when truncating is enabled.
+   * Must be a valid Tailwind max-width utility class (e.g., "tw-max-w-40", "tw-max-w-xs").
+   */
+  readonly maxWidthClass = input<`tw-max-w-${string}`>("tw-max-w-52");
```
willmartian, line 157: Hmm, I don't like this, but I get that it is preexisting. So, maybe we remove this in a future ticket. I would expect someone to just apply a !tw-max-w class on the host element as needed, instead of using this input.

libs/components/src/chips/shared/base-chip.directive.ts
```diff
@@ -0,0 +1,201 @@
...
+  readonly disabled = computed(() => this.disabledInput() || this.disabledState());
+
+  /** Set to true by consuming components that render a trailing icon or dismiss button */
+  readonly hasTrailingIcon = signal(false);
+
+  /**
+   * Computed class list based on variant, size, and state
+   */
+  protected readonly classList = computed(() => {
+    const size = this.size() || "large";
```
willmartian, line 178: `tailwind-variants` would make this so much better!

libs/components/src/chips/shared/chip-content.component.html
```diff
@@ -0,0 +1,15 @@
+<ng-content select="[slot=chipStart]"></ng-content>
```
willmartian: do you think the "chip" prefix for the slot name is helpful?

libs/components/src/chips/shared/chip-content.component.html
```diff
@@ -0,0 +1,15 @@
...
+  <bit-icon [name]="startIcon()" class="tw-text-base/5" />
+}
+
+<span class="tw-text-left tw-truncate tw-leading-5 tw-min-w-0 tw-flex-1">
+  <ng-content></ng-content>
+</span>
+
+@if (endIcon() && !dismissible()) {
+  <bit-icon [name]="endIcon()" class="tw-text-base/5" />
+}
```
willmartian: is `dismissible` needed here? Can't we always render this? Instead, we will just avoid passing an endIcon in the chip component. It feels weird to have "dimissible" as part of the API if it isn't actually rendering a dismiss button when true.

libs/components/src/chips/shared/chip-content.component.html
```diff
@@ -0,0 +1,15 @@
+<ng-content select="[slot=chipStart]"></ng-content>
+
+@if (startIcon()) {
+  <bit-icon [name]="startIcon()" class="tw-text-base/5" />
+}
+
+<span class="tw-text-left tw-truncate tw-leading-5 tw-min-w-0 tw-flex-1">
```
willmartian: Should this be `tw-text-start`?

libs/components/src/chips/shared/base-chip.directive.ts
BryanCunningham: I don't love it either. I guess it's handy if we want to enforce not allowing custom classes on CL components in the future.

libs/components/src/chips/shared/chip-content.component.html
BryanCunningham: Don't have strong feelings about it personally. Can change it to just `start` if we want. Not a problem

libs/components/src/chips/shared/base-chip.directive.ts
willmartian: Very true

libs/components/src/chips/shared/chip-content.component.html
willmartian: I think so, just for symmetry with all the other start slots we define

libs/components/src/chips/shared/chip-content.component.html
BryanCunningham: @willmartian This use case is from the chip-filter. When there is no selection, we render the `endIcon` which is the arrow. When a selection is made, it becomes dismissible and the chip-filter then renders the dismiss button in the 'end' slot. Def open to a different approach though

libs/components/src/chips/shared/chip-content.component.html
BryanCunningham: Sure should be. Good call

libs/components/src/chips/shared/chip-content.component.html
BryanCunningham: Basically, we never want to render both an end icon AND the end slot

willmartian: Just to make sure I am following: isn't that already handle by how the `endIcon` is set in chip-filter?

```
[endIcon]="
      !selectedOption ? (menuTrigger.isOpen ? 'bwi-angle-up' : 'bwi-angle-down') : undefined
    "
```

It is set to `undefined`. `selectedOption` being true is also what renders the dismiss button into the slot.

BryanCunningham: Yeah, I can remove this actually. :) Thanks for the feedback

willmartian: of course! thanks for talking it through

apps/browser/src/vault/popup/services/vault-popup-list-filters.service.ts
nick-livefront, line 1: Thank you for keeping the documentation up to date here! 🙇

PR #18799, opened 2026-02-05
[CL-1022] Update Berry Styles

lxiong-livefront:
## 🎟️ Tracking

[JIRA](https://bitwarden.atlassian.net/browse/CL-1022?atlOrigin=eyJpIjoiNTM0NzI2MTIxOWRhNDkxODg3NWFmNmYwYjJiMmE2YTUiLCJwIjoiaiJ9)

## 📔 Objective

- Created new bit-berry component and stories
- Updated `apps/browser/src/platform/popup/layout/popup-tab-navigation.component.ts` to use bit-berry internally

## 📸 Screenshots

Screenshot 2026-02-06 at 4 51 24 PM

### DIFF
--- a/libs/components/src/berry/berry.mdx
+++ b/libs/components/src/berry/berry.mdx
@@ -35,8 +35,8 @@ import { BerryComponent } from "@bitwarden/components";
 
 ## Count Behavior
 
-- Counts of **1-9**: Display in a compact circular shape
-- Counts of **10-999**: Display in a pill shape with padding
+- Counts of **1-99**: Display in a compact circular shape
+- Counts of **100-999**: Display in a pill shape with padding
 - Counts **over 999**: Display as "999+" to prevent overflow
 
 ## Accessibility


libs/components/src/berry/berry.component.ts
```diff
@@ -0,0 +1,84 @@
...
+  selector: "bit-berry",
+  imports: [NgClass],
+  templateUrl: "berry.component.html",
+  changeDetection: ChangeDetectionStrategy.OnPush,
+})
+export class BerryComponent {
+  protected readonly variant = input<BerryVariant>("primary");
+  protected readonly count = input<number>();
+
+  protected readonly content = computed(() => {
```
BryanCunningham, line 30: I think this can be simplified
```js
protected readonly content = computed(() => {
  const count = this.count();
  if (!count || count <= 0) { 
    return undefined
  };
  return count > 999 ? "999+" : `${count}`;
});

File: libs/components/src/berry/berry.component.ts [hunk 2]
```diff
@@ -0,0 +1,84 @@
...
+    const count = this.count();
+    if (count === undefined || count === null || count <= 0) {
+      return undefined;
+    }
+    return count > 999 ? "999+" : `${count}`;
+  });
+
+  protected readonly effectiveSize = computed<"small" | "large">(() => {
+    const count = this.count();
+    return count !== undefined && count !== null && count > 0 ? "large" : "small";
```
BryanCunningham: Same comment about simplification as above.

libs/components/src/berry/berry.component.ts
BryanCunningham: The only issue with this is we can't show the number `0` in this code. I don't think... would have to try it out. Depends on if we want to do that or not

BryanCunningham: If that IS the case, we could do something like this to simplify a little bit
```js
protected readonly content = computed(() => {
  const count = this.count();
  if (count == null || count <= 0) {
      return undefined;
    }
  return count > 999 ? "999+" : `${count}`;
});
```

libs/components/src/berry/berry.component.ts
```diff
@@ -0,0 +1,84 @@
...
+
+  protected readonly content = computed(() => {
+    const count = this.count();
+    if (count === undefined || count === null || count <= 0) {
+      return undefined;
+    }
+    return count > 999 ? "999+" : `${count}`;
+  });
+
+  protected readonly effectiveSize = computed<"small" | "large">(() => {
```
BryanCunningham: ❓ Does it make more sense for this to be something like... `computedShape` rather than 'size'? We aren't really determining a size here per se. More like determining if it's round or not. Any thoughts?

libs/components/src/berry/berry.component.ts
```diff
@@ -0,0 +1,84 @@
...
+      large: ["tw-h-4", "tw-min-w-4", this.padding()],
+    };
+
+    const variantClass = {
+      primary: "tw-bg-bg-brand",
+      subtle: "tw-bg-bg-contrast",
+      success: "tw-bg-bg-success",
+      warning: "tw-bg-bg-warning",
+      danger: "tw-bg-bg-danger",
+      accentPrimary: "tw-bg-bg-accent-primary",
```
BryanCunningham: I don't think this color is quite correct based on the designs. This color produces contrast ratio issues with the white text. Seems like it should be `...primary-strong`
```suggestion
      accentPrimary: "tw-bg-bg-accent-primary-strong",
```

libs/components/src/berry/berry.component.ts
```diff
@@ -0,0 +1,84 @@
...
+export class BerryComponent {
+  protected readonly variant = input<BerryVariant>("primary");
+  protected readonly count = input<number>();
+
+  protected readonly content = computed(() => {
+    const count = this.count();
+    if (count === undefined || count === null || count <= 0) {
+      return undefined;
+    }
+    return count > 999 ? "999+" : `${count}`;
```
BryanCunningham: ❓ Is `999` always the max? Or do we want this to be configurable? @willmartian @vleague2 did we discuss this? Or, do we need to reach out to design?

apps/browser/src/platform/popup/layout/popup-tab-navigation.component.html
```diff
@@ -28,7 +28,7 @@
             </span>
           </button>
           <div *ngIf="button.showBerry" class="tw-absolute tw-top-1.5 tw-left-[calc(50%+5px)]">
-            <div class="tw-bg-notification-600 tw-size-2.5 tw-rounded-full"></div>
```
BryanCunningham, line 31: ❓ This berry will change from pink to blue with this change. Are we ok with that? Or do we need to provide a variant here?

libs/components/src/berry/berry.component.ts
lxiong-livefront: Yes, we won't ever need to display 0 or negative numbers

libs/components/src/berry/berry.component.ts
lxiong-livefront: I think 'size' could make sense here, per the design notes as well as the element dimensions (small is a much smaller circle while large could be a larger circle rather than pill shape, if the count is a single digit)
https://www.figma.com/design/Zt3YSeb6E6lebAffrNLa0h/branch/rKUVGKb7Kw3d6YGoQl6Ho7/Tailwind-Component-Library?node-id=38367-199759&m=dev

libs/components/src/berry/berry.component.ts
BryanCunningham: Cool, then I think the first example is good then!

libs/components/src/berry/berry.component.ts
vleague2: The figma says the max is 999 in the left info panel, so I think this matches the spec

libs/components/src/berry/berry.component.ts
BryanCunningham: Got it. So if there is no content/count, it's always the small size then? No use case for it to be 'large' and not have a count?

libs/components/src/berry/berry.component.ts
lxiong-livefront: Simplified in [dcca32f](https://github.com/bitwarden/clients/pull/18799/commits/dcca32f490e149b26721b987920864a816ea9249)

libs/components/src/berry/berry.component.ts
lxiong-livefront: Simplified in [dcca32f](https://github.com/bitwarden/clients/pull/18799/commits/dcca32f490e149b26721b987920864a816ea9249)

libs/components/src/berry/berry.component.ts
lxiong-livefront: Thanks for catching! Fixed this color in [73fc482](https://github.com/bitwarden/clients/pull/18799/commits/73fc482bdabcdf649f0e711b593ea329dd83bc7f)

libs/components/src/berry/berry.component.ts
```diff
@@ -30,15 +30,15 @@ export class BerryComponent {
 
   protected readonly content = computed(() => {
     const count = this.count();
-    if (count === undefined || count === null || count <= 0) {
+    if (count == null || count <= 0) {
```
BryanCunningham: I think `0` could be passed in in this scenario. If we want to ensure 0 is never rendered we need to check if count is truthy or not. Or, explicitly check for `count === 0`. But, that feels heavy handed. What do you think?

libs/components/src/berry/berry.component.ts
BryanCunningham: Sounds good. I've got a feeling that we'll add the ability to set a max number in here at some point but, we'll cross that bridge when we get there :)

libs/components/src/berry/berry.component.ts
BryanCunningham: What I mean is: 
if count = 0
```
count == null || count < 0
```
is false

```
!count || count < 0
```
Is true

libs/components/src/berry/berry.component.html
```diff
@@ -0,0 +1 @@
+<span [ngClass]="berryClasses()">{{ content() }}</span>
```
vleague2: We are moving away from `[ngClass]` to using `[class]` ([source](https://contributing.bitwarden.com/contributing/code-style/web/angular-migration-guide#class-and-style-bindings))

libs/components/src/berry/berry.component.ts
```diff
@@ -0,0 +1,84 @@
...
+    const count = this.count();
+    if (count == null || count <= 0) {
+      return undefined;
+    }
+    return count > 999 ? "999+" : `${count}`;
+  });
+
+  protected readonly computedSize = computed<"small" | "large">(() => {
+    const count = this.count();
+    return count !== undefined && count > 0 ? "large" : "small";
```
BryanCunningham:
```suggestion
    return count && count > 0 ? "large" : "small";
```

apps/browser/src/platform/popup/layout/popup-tab-navigation.component.html
vleague2: The story for this component is showing the berry in a weird spot. I think we will need to make a positioning change here as well.

libs/components/src/berry/berry.component.ts
lxiong-livefront: Updated in [544f0c6](https://github.com/bitwarden/clients/pull/18799/commits/544f0c640e362003fc2fb653d41b341e12d9f267)

This should render the berry if `count` is undefined, but won't render if `count` is 0 or below

libs/components/src/berry/berry.component.ts
```diff
@@ -0,0 +1,84 @@
...
+  });
+
+  protected readonly textColor = computed(() => {
+    return this.variant() === "contrast" ? "tw-text-fg-dark" : "tw-text-fg-white";
+  });
+
+  protected readonly padding = computed(() => {
+    return this.count() > 9 ? "tw-px-1.5 tw-py-0.5" : "";
```
BryanCunningham: Not sure it's worth it but, makes it a little more clear what the intention is here
```suggestion
    return this.count() && this.count().toString().length > 2 ? "tw-px-1.5 tw-py-0.5" : "";
```

libs/components/src/berry/berry.component.ts
lxiong-livefront: Simplified in [f587366](https://github.com/bitwarden/clients/pull/18799/commits/f587366dd5dc15f68343a9d36baa37152de362c7)

libs/components/src/berry/berry.component.ts
lxiong-livefront: I agree this makes the intention more clear!
Updated in [3eef349](https://github.com/bitwarden/clients/pull/18799/commits/3eef349ed8c4d63a2a5497fd50111dd08b39bf9c)

libs/components/src/berry/berry.component.ts
BryanCunningham: Yeah, that will work too but, this logic would still be wrong. We'll need to change this and then just update the template to what it was before

libs/components/src/berry/berry.component.ts
lxiong-livefront: Yes, there wouldn't be any use case for a 'large' berry without count/content.

libs/components/src/berry/berry.component.html
lxiong-livefront: Good to know, thanks! I originally modeled after 'badge' which uses `ngClass`

Updated in [a777763](https://github.com/bitwarden/clients/pull/18799/commits/a777763f701d5f7761db91f018757ed2a88befb4)

apps/browser/src/platform/popup/layout/popup-tab-navigation.component.html
lxiong-livefront: I checked in with Emily on design, and she said we want to update the variant here to use 'danger'. Made this update in [ef98bae](https://github.com/bitwarden/clients/pull/18799/commits/ef98bae55b64a5a2e5a88877e527c62a43ac7f6f)

Fixed the positioning in [439be0e](https://github.com/bitwarden/clients/pull/18799/commits/439be0e05cb33c27273efd37a74ddcc7b52a92a0)

libs/components/src/berry/berry.component.ts
lxiong-livefront: We still want to display the small berry (no content) if the no count is passed in, so `@if (content)` wouldn't render the berry in this case. I updated the logic for content in [b492f2d](https://github.com/bitwarden/clients/pull/18799/commits/b492f2d56080dd5bf2fdbebf8180ec85ee8e5ac0)

Let me know your thoughts

BryanCunningham: @lxiong-livefront Yeah, you are right. I edited my comment after I sent it. We'd just remove the conditional from the template

lxiong-livefront: No worries! Use cases for this are a little complex. From my understanding, 
- display small berry with no content if no count is passed in
- display large berry with count content if a count is passed in and is > 0
- hide the berry altogether if a count is passed in with a value 0 or below

I think with the updated content logic and conditional rendering, all these cases will pass, but let me know if you'd prefer another/simpler approach! (We could avoid checking for count <= 0 in the logic, but feels a little weird to set the content for it even if it won't be rendered)

BryanCunningham: I think we can just remove the conditional logic from the template entirely. `content` will be undefined/not rendered  if the count is 0. And a small berry will be the result. But, let me know if there is some scenario I am overlooking

lxiong-livefront: If the count is 0 or below, we don't want to render any berry (not even a small), but we do still render a small berry if the content is undefined and no count was passed in, so I think we would still need the conditional logic for the template

lxiong-livefront: I moved the conditional check into `showBerry` in [06f4d24](https://github.com/bitwarden/clients/pull/18799/commits/06f4d24555b0be8d20098aafb4faf4991cfe15f2)

Might be cleaner/more readable than including that in the template

BryanCunningham: Actually... now that I think about this more. This component is doing too much probably :) The conditional logic to decide when/when not to render this should belong to the consumer. The component shouldn't have to care about this. It should only care about if there is content or not. If there is no content, render small berry. If there is, render the large one.

libs/components/src/berry/berry.component.html
vleague2: Yeah, unfortunately we are still midway through modernizing the CL Angular code, so it is difficult to know what examples to pull 😭

libs/components/src/berry/berry.component.ts
BryanCunningham: For posterity: After chatting with design we've decided to have this component take a `variant` input. This clears up internal rendering confusion. ie. What happens when `count` is 0, is it a small berry or do we render nothing?, How does a consumer indicate they want to render a small berry if the count is 0?... that sort of thing. Makes for a more intuitive API and an overall better DX(IMHO).

lxiong-livefront: Made some updates to align with what we discussed in [944acf9](https://github.com/bitwarden/clients/pull/18799/commits/944acf9f57be1e6454ac81020a2e51b10b8c98ed)

Added a type input that defaults to "count". Still need the conditional rendering in template to allow for "status" types with no content, but overall I think this direction makes the use cases more clear!

libs/components/src/berry/berry.component.ts
```diff
@@ -0,0 +1,80 @@
...
+    }
+    return value > 999 ? "999+" : `${value}`;
+  });
+
+  protected readonly textColor = computed(() => {
+    return this.variant() === "contrast" ? "tw-text-fg-dark" : "tw-text-fg-white";
+  });
+
+  protected readonly padding = computed(() => {
+    return this.value()?.toString().length > 2 ? "tw-px-1.5 tw-py-0.5" : "";
```
vleague2: It looks like there is a type error on this line?

apps/browser/src/platform/popup/layout/popup-tab-navigation.component.html
vleague2: Thanks! I think this needs the `status` type now that we have type variants

libs/components/src/berry/berry.component.ts
lxiong-livefront: fixed in [3bd7951](https://github.com/bitwarden/clients/pull/18799/commits/3bd79516fa6c9b4d784cd55b4789f2399169a344)

apps/browser/src/platform/popup/layout/popup-tab-navigation.component.html
lxiong-livefront: Absolutely! Thanks for catching, added the `status` type in [26f6aa9](https://github.com/bitwarden/clients/pull/18799/commits/26f6aa9d81b34bb947ab6d8de2663283f83a59ad)

PR #18800, opened 2026-02-05
[PM-31825] Add fido2 support to SDK create/update methods

nikwithak:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-31825

## 📔 Objective

Passkeys are currently not being passed between the SDK and TS clients, due to discrepancies with how they're stored. This PR fixes the bug where they were getting wiped entirely when editing a cipher, and consolidates the FIdo2 logic into the conversion methods between TS and SDK types.


libs/common/src/vault/services/cipher-sdk.service.ts
shane-melton, line 1: ❓ What do you think about re-using the same logic that is present in the `DefaultCipherEncryption` service so we don't have to maintain this Fido logic in multiple places? We might need to re-organize some of the code there to make it publicly accessible, but I think it would be beneficial to share this pattern.

nikwithak: I've refactored to use shared logic, adding it to the `CipherView` functions.  `CipherView.toSdkCipherView()` now takes an optional sdk paramater with the `CiphersClient`, and will use that to handle the passkeys if present.

shane-melton: Looks great, thanks!

PR #18801, opened 2026-02-05
CL-980 Update Icon Tile to new design spec

iivins-livefront:
## Overview

The goal of the ticket is to updated the Icon Tile component styling and variants to align with the new Flowbite design system.

Add testing coverage, add markdown, extended Storybook, and modernized angular aspects inline with the migration guide.

## Key Changes

- Added 3 new color variants (subtle, dark, contrast) and 2 new sizes (xs: 16px, xl: 64px), increasing from 15 to 35 total variant combinations
- Replaced legacy color tokens w/ semantic Flowbite tokens
- Implemented OnPush change detection, removed NgClass anti-pattern
- Added testing suite, mdx, and additional Storybook stories

## Screenshots

Screenshot 2026-02-05 at 5 45 51 PM

### DIFF
--- a/bitwarden_license/bit-web/src/app/dirt/access-intelligence/activity/application-review-dialog/assign-tasks-view.component.html
+++ b/bitwarden_license/bit-web/src/app/dirt/access-intelligence/activity/application-review-dialog/assign-tasks-view.component.html
@@ -17,7 +17,7 @@
         <bit-icon-tile
           icon="bwi-users"
           variant="primary"
-          size="large"
+          size="lg"
           shape="circle"
           aria-hidden="true"
         ></bit-icon-tile>
@@ -36,7 +36,7 @@
         <bit-icon-tile
           icon="bwi-desktop"
           variant="warning"
-          size="large"
+          size="lg"
           shape="circle"
           aria-hidden="true"
         ></bit-icon-tile>


libs/components/src/icon-tile/icon-tile.component.spec.ts
```diff
@@ -0,0 +1,293 @@
+import { ComponentFixture, TestBed } from "@angular/core/testing";
```
BryanCunningham, line 1: Other than the simple test to ensure the component renders and the a11y tests, I don't think the rest are necessary. Testing style/css class application is brittle and these things are covered by visual regression testing. 

❓ Should these even be visible to screen readers? My guess is no. They're seemingly presentation only. In that case, the a11y tests can go as well. Any thoughts @willmartian @vleague2

vleague2:
I think you will need to rebase onto the feature branch since you have a bunch of commits from `main` in here. (It will be easier to merge latest from the feature branch into these stacked PRs rather than merging latest from `main`. We can keep the feature branch updated with `main`.)

iivins-livefront:
@vleague2  Ah! I pulled `main` into `uif/CL-1026/ds-refresh-2` locally on accident - removing those changes now

Out of curiosity - are we attempting to keep the feature branch up-to-date with main before we want to release it? Trying to understand if and when to address potential conflicts

vleague2:
@iivins-livefront Yes we usually keep the feature branch up-to-date with main (I merged in `main` and pushed that up for the feature branch like two hours ago, but it'll be easier once we have a PR for the feature branch). I imagine we'll have quite a lot of conflicts on these particular feature branches due to the kinds of widespread change we're expecting to make, so I think we'll just have to address those on the feature branch level as they arise.

libs/components/src/icon-tile/icon-tile.component.spec.ts
iivins-livefront: Didn't realize we had a form of visual regression testing - is that normally done through Storybook during a release cycle?

I'm game to simplify this and keep it minimalistic ( i.e just a component render check and potentially the a11y tests ) - just let me know

BryanCunningham: Yep! We run chromatic and do screenshot testing on every story. _Should_ on every commit... but I don't see the job run in this branch for some reason. Maybe because the PR isn't pointed at `main`. I'm not for sure. @vleague2 any context as to why chromatic isn't running on these PRs to the feature branch?

vleague2: I would agree. Unit tests should be covering interaction behavior and we should only worry about specific implementation tests if we're needing to cover some bug edge case or something. The existing icon tile is hidden to screen readers, and we can confirm that with design, but I think it makes sense to continue to hide them.

Re: visual regression testing, the visual regression tests are run on every PR commit that targets `main` and any diffs must be reviewed and approved before merging into `main`. This is one thing that's annoying with feature branches because I don't think it'll run automatically on a PR into a feature branch 🤔 We might be able to manually run them for a branch though so that we can at least review before we approve a PR

vleague2: Lmao I didn't see Bryan's reply before I wrote mine, thanks Github

libs/components/src/icon-tile/icon-tile.component.html
```diff
@@ -1,7 +1,7 @@
 <div
-  [ngClass]="containerClasses()"
+  [class]="containerClasses().join(' ')"
```
vleague2: Huh I'm surprised you have to do this, I thought it accepted an array of classes already 🤔

libs/components/src/icon-tile/icon-tile.component.ts
vleague2: I just remembered that the figma no longer has the circle shape. I think we are just doing square. If so we can remove this input and any associated classes

libs/components/src/icon-tile/icon-tile.component.ts
```diff
@@ -1,47 +1,67 @@
...
+    icon: ["tw-text-xs", "tw-leading-[0]"],
   },
-  default: {
-    container: ["tw-w-8", "tw-h-8"],
-    icon: ["tw-text-base"],
+  sm: {
+    container: ["tw-w-6", "tw-h-6"],
+    icon: ["tw-text-sm", "tw-leading-[0]"],
   },
-  large: {
```
vleague2, line 29: It looks like there are some build errors related to client code passing in a size that no longer exists, so those will need to be updated to pass in the new correct size options

libs/components/src/icon-tile/icon-tile.component.ts
iivins-livefront: I can remove `"square"` and the associated classes - but do we want to remove `shape` as a param all together now since we're down to one?

I.e remove `shapeStyles`, `IconTileShape`, and the docs/tests around it

vleague2: I'm double-checking with Emily in the Figma file!

vleague2: Okay she confirmed that we don't need to support other shapes aside from square!

libs/components/src/icon-tile/icon-tile.component.html
iivins-livefront: I in fact, did not have to do this 😬 

> The expression can be one of:

- A space-delimited string of class names.
- An object with class names as the keys and truthy or falsy expressions as the values.
- An array of class names.

https://v17.angular.io/guide/class-binding#binding-to-multiple-css-classes

Removing here and in the `iconClasses()` instance below

iivins-livefront:
Addressing PR Feedback:
- Updated testing suite, kept render and accessibility checks
- Fixed build error ( single instance of "large" -> "lg" )
- Removed "shape" from docs, no more Circle, changed to `borderRadius` since it made more sense
- Removed redundant `.join(' ')` class bindings

Screenshot 2026-02-10 at 12 21 45 PM

libs/common/src/auth/send-access/types/invalid-request-errors.type.ts
```diff
@@ -44,7 +44,8 @@ export type EmailAndOtpRequired = InvalidRequest & {
 };
 
 export function emailAndOtpRequired(e: SendAccessTokenApiErrorResponse): e is EmailAndOtpRequired {
-  return e.error === "invalid_request" && e.send_access_error_type === "email_and_otp_required";
+  // return e.error === "invalid_request" && e.send_access_error_type === "email_and_otp_required";
+  return e.error === "invalid_request";
```
vleague2: Is this intentional?

libs/components/src/icon-tile/icon-tile.mdx
```diff
@@ -0,0 +1,113 @@
...
+are used for visual indicators, status representations, or decorative elements throughout the
+application. Unlike icon buttons, icon tiles do not respond to user interaction and serve purely
+presentational purposes.
+
+Use icon tiles to:
+
+- Display status or category indicators
+- Represent different types of content
+- Create visual hierarchy in lists or cards
+- Show app or service icons in a consistent format
```
vleague2: I believe this text should already appear on this page due to the docs comment on the component in `icon-tile.component.ts`, so we shouldn't need to duplicate it manually here.

libs/components/src/icon-tile/icon-tile.mdx
```diff
@@ -0,0 +1,113 @@
...
+
+## Variants
+
+Icon tiles support 7 color variants that align with the Flowbite design system, each using semantic
```
vleague2: I don't think we need to mention Flowbite here since it is being used mostly as design inspiration, not actually a perfect alignment to our implementation.

libs/components/src/icon-tile/icon-tile.component.ts
vleague2: Still seeing build errors in the github actions workflows

libs/components/src/icon-tile/icon-tile.mdx
```diff
@@ -0,0 +1,113 @@
...
+design tokens for consistent theming.
+
+### Primary
+
+Used for brand-colored indicators and primary visual elements.
+
+<Canvas of={stories.AllVariants} />
```
vleague2: I would move this canvas up above the Primary subheading, since it illustrates all variants.

libs/common/src/auth/send-access/types/invalid-request-errors.type.ts
iivins-livefront: Not intentionally committed - more than one wrong decision here. At some point this branch got ahead of the feature branch and I hadn't reinstalled everything, strange type error was being thrown when launching Storybook

libs/components/src/icon-tile/icon-tile.mdx
iivins-livefront: I was following `icon-button.mdx` that had a Usage section with the paths outlined - looks cleaner without the redundancy so I'll remove this section

libs/components/src/icon-tile/icon-tile.mdx
iivins-livefront: > Still seeing build errors in the github actions workflows

OOC - what's the _best way_ to be running these workflows locally during development?

I haven't configured Docker yet, got away with avoiding it for the Desktop effort - I'd like to avoid crossing my fingers and relying on the type checking, linter, unit tests, and code search results 😬

vleague2: I imagine you'd catch them just running the builds for each app. Like `npm run build:bit` on the `clients/apps/web` directory. You might also catch some with the local storybook script, I think it also fails when there are build errors (`npm run storybook` on `clients` root)

vleague2:
I'm going to run chromatic on this branch from the command line and see if it gives us a diff to review!
Update -- somehow it ran on the wrong branch, so.... failure lol

vleague2 (approved):

PR #18811, opened 2026-02-06
[BEEEP] Add deprecation notices to encstring encrypt API

quexten:
## 📔 Objective

Adds deprecation notices to the encrypt-side API's so that new use-cases are evaluated by the teams consuming these APIs. EncString should not be used, but instead the safe primitive "DataEnvelope" should be used.

encryptBytes should just not be used. Key encapsulation should also probably not be used, and usage indicates that the team is building an unsafe key-sharing protocol that lacks sender authentication. This should be done via IdentityKeyEnvelope instead.


Thomas-Avery (approved):
LGTM

PR #18816, opened 2026-02-06
Auth/PM-31804 - Migrate auth font icons to bit-icon

JaredSnider-Bitwarden:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-31804
Relates to https://github.com/bitwarden/clients/pull/18584

## 📔 Objective

To migrate all auth owned font icons to the new `bw-icon` component to clean up many warnings. 

Per #18584, the migration pattern is:
```
  <!-- OLD: Font icons -->                                                                                                                                                               
  <i class="bwi bwi-lock" aria-hidden="true"></i> 
                                                                                                               
  <!-- NEW: Font icons -->                                                                                                                                                               
  <bit-icon name="bwi-lock"></bit-icon> 
```

The addition of the aria label for all icons allowed for the deletion of some screen reader only spans while maintaining accessibility. 

## 📸 Screenshots

I independently verified every icon change on each component, but there's too many to post.


apps/browser/src/auth/popup/settings/account-security.component.html
```diff
@@ -123,7 +123,7 @@ <h2 bitTypography="h6">{{ "manageDevices" | i18n }}</h2>
       <bit-item>
         <button bit-item-content type="button" appStopClick routerLink="/device-management">
           {{ "devices" | i18n }}
-          <i slot="end" class="bwi bwi-chevron-right" aria-hidden="true"></i>
+          <bit-icon slot="end" name="bwi-angle-right" class="tw-text-lg"></bit-icon>
```
JaredSnider-Bitwarden, line 126: This icon change is due to the fact that chevron isn't an icon we have, and nothing was rendering previously in main.

apps/web/src/app/auth/emergency-access/accept/accept-emergency.component.html
```diff
@@ -1,11 +1,11 @@
...
     <p class="tw-text-center">
-      <i
-        class="bwi bwi-spinner bwi-spin bwi-2x tw-text-muted"
-        title="{{ 'loading' | i18n }}"
-        aria-hidden="true"
-      ></i>
+      <bit-icon
+        name="bwi-spinner"
+        class="bwi-spin bwi-2x tw-text-muted"
+        [ariaLabel]="'loading' | i18n"
```
enmande, line 7: 🤔 Regarding instances of spinner in particular: 
It looks [here](https://github.com/bitwarden/clients/blob/323f30c8e941d3bfb72acb812e6a90e949771a5d/libs/components/src/icon/icon.component.ts#L10) like setting `aria-label` will prevent the `aria-hidden` attribute from having a value on the icon, which is a change in behavior. We also have the `span` sibling marked for `tw-sr-only` which could result in multiple instances of "loading" being reported for the same `<p>` to a screen reader I think and be a WCAG consideration?

apps/web/src/app/auth/settings/emergency-access/takeover/emergency-access-takeover-dialog.component.html
```diff
@@ -7,12 +7,11 @@
...
-          class="bwi bwi-spinner bwi-spin bwi-2x tw-text-muted"
-          title="{{ 'loading' | i18n }}"
-          aria-hidden="true"
-        ></i>
-        <span class="tw-sr-only">{{ "loading" | i18n }}</span>
+        <bit-icon
+          name="bwi-spinner"
+          class="bwi-spin bwi-2x tw-text-muted"
+          [ariaLabel]="'loading' | i18n"
+        ></bit-icon>
```
enmande, line 14: 🤔 Similar to [above](https://github.com/bitwarden/clients/pull/18816/changes#r2805921600) question, but here we've removed the `tw-sr-only`-class `span` sibling, and I would anticipate a single "loading" evaluation for a screen reader.

I'm not an a11y expert, but I _think_ if there's no text to display, we can consistently do this option and have the `bit-icon` report the label?

apps/web/src/app/auth/emergency-access/accept/accept-emergency.component.html
JaredSnider-Bitwarden: Per sync discussion, ty for finding the missed sr-only spans. I meant to remove them all and replace them wiht aria label. I've done so now!

apps/web/src/app/auth/settings/emergency-access/takeover/emergency-access-takeover-dialog.component.html
JaredSnider-Bitwarden: See https://github.com/bitwarden/clients/pull/18816#discussion_r2806034773

djsmith85 (approved):

PR #18822, opened 2026-02-06
[PM-30927] Fix lock component initialization bug

Thomas-Avery:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-30927

## 📔 Objective

The purpose of this PR is to fix an initialization bug in the lock component. If `listenForUnlockOptionsChanges` would beat a longer running `listenForActiveAccountChanges` the active unlock option would be set twice. This could cause forms, such as master password and PIN, already in use to be reinitialized and any previous text written to be wiped out.

This PR makes polling for `listenForUnlockOptionsChanges` wait until the component is done loading which is controlled by `listenForActiveAccountChanges` & component intitialization.


Thomas-Avery:
Looks like the Linux desktop build is failing on main as well and is unrelated here.

libs/key-management-ui/src/lock/components/lock.component.ts
```diff
@@ -202,15 +202,15 @@ export class LockComponent implements OnInit, OnDestroy {
     timer(0, 1000)
       .pipe(
         mergeMap(async () => {
-          if (this.activeAccount?.id != null) {
+          // Only perform polling after the component has loaded. This prevents multiple sources setting the default active unlock option on initialization.
```
quexten, line 205: @Thomas-Avery The reason this is expensive is because we have backwards compatibility to the old biometrics protocol still active which we removed in 2024.12. We have a comment here:

https://github.com/bitwarden/clients/blob/f7a5ad712fad27e4fbc27d395d03625e037f4bbb/apps/browser/src/key-management/lock/services/extension-lock-component.service.ts#L77

And essentially, we wait for a 3 second timeout.

We should probably make a clean-up task here and up-prioritize it because this 3 second timeout is bound to cause other issues than this bug. (separate from this bug-fix)

quexten: It also most likely means that biometrics users will still wait 3 seconds for their lock component to init if the app is closed.

hboetes:
Can somebody be so kind to guestimate how long it will take before the updated add-on arrives at firefox?

pamperer562580892423:
> Can somebody be so kind to guestimate how long it will take before the updated add-on arrives at firefox?

You mean when the release, that contains this fix, will be approved by Mozilla's extension store? - From my experience of the recent years: anything between a few days and a few weeks. (I think quite recently it was only about a day... but last year, I think it was also about six weeks one time...)

PR #18824, opened 2026-02-06
[BEEEP] Desktop Native simpler `napi` errors

neuronull:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-31827

## 📔 Objective

I noticed we do a lot of 

```Rust
.map_err(|e| napi::Error::from_reason(e.to_string()))
```

, and that's not needed when using `anyhow` since `napi` has a `From` impl for it.

Additionally, I spotted a few places where our error strings are a little inconsistent or giving redundant output. Since the errors passed back are not directly user facing, there is no need to separate a debug (?) , formatting of the error, and in this case it's redundant info.

I think there is still room for improvement in passing back errors through the napi layer. There is one spot where we are doing a String compare on the error. A holistic look is pending.

I did confirm that none of the areas where I adjusted the error strings are being compared upstream in TS.


neuronull:
The Desktop CI failure will resolve after merging in fix for https://github.com/bitwarden/clients/pull/18814

neuronull:
FYI this is blocked on https://github.com/bitwarden/clients/pull/19113

iinuwa:
@neuronull, #19113 is merged.

neuronull:
Validation by QA is complete.

PR #18838, opened 2026-02-09
[PM-31880 | BEEEP] Move random out of keyservice to PureCrypto

quexten:
## 🎟️ Tracking

https://github.com/bitwarden/sdk-internal/pull/750
https://bitwarden.atlassian.net/browse/PM-31880

## 📔 Objective

We have a hand-rolled rejection-sampling based implementation in key-service. While this has been subject to audits, it is still better to replace it with an external implementation that is widely adopted. This pair of PR's moves randomness to PureCrypto, with the only consumer being tools' password generator code. Further, this moves randomness out of keyservice (where it does not belong from a domain perspective, the key service is just meant as a state service to handle holding keys, nothing else).


harr1424 (dismissed):

libs/tools/generator/core/src/engine/key-service-randomizer.ts
```diff
@@ -28,7 +32,8 @@ export class KeyServiceRandomizer implements Randomizer {
     }
 
     if (options?.number ?? false) {
-      const num = await this.keyService.randomNumber(1, 9);
+      await SdkLoadService.Ready;
+      const num = PureCrypto.random_number(1, 9);
```
Thomas-Avery: Did we consider using the `this.sdkService.client$` without user context and have this be something exposed as a randomizer client utility instead of pure crypto?

Thomas-Avery (changes requested):
Looks like there are build failures.

libs/tools/generator/core/src/engine/key-service-randomizer.ts
quexten: Build errors should subside after merging in the latest SDK version.

As to the client:
My mental heuristic for splitting into PureCrypto vs a client is the following:

PureCrypto is meant for:
- Short term things that have a clear path to removal
- Low level functions that we don't really want to offer as high level functionality
- ^ Equivalent to cryptofunctionservice/encryptservice

Clients are meant for:
- Long term stable work
- High level functions that we do want to offer, with appropriate mental models

All of these lean towards PureCrypto, since a generic random number selector is something that should not be offered to clients. They should use high level functionality, and the tools consumer already has a rust implementation with a high level interface. As soon as tools moves over clients to it (in addition to the current mobile use) we will drop it.

There is no benefit we gain from adding it onto a client here, but in the contrary, it may suggest to consumers to depend on it, whereas Purecrypto is clearly meant for short-term work, which this falls under. We would have to add a new client "low_level_crypto_do_not_use_client()" or "unstable_crypto_client()" to set expectations correctly.

Given the above considerations, I do believe PureCrypto is the appropriate place for this work to live.

quexten: https://github.com/bitwarden/clients/pull/19056

Thomas-Avery (dismissed):

libs/tools/generator/core/src/engine/purecrypto-randomizer.ts
```diff
@@ -28,7 +32,8 @@ export class KeyServiceRandomizer implements Randomizer {
     }
 
     if (options?.number ?? false) {
-      const num = await this.keyService.randomNumber(1, 9);
+      await SdkLoadService.Ready;
+      const num = PureCrypto.random_number(0, 9);
```
quexten, line 36: Please note the range here changed from 1-9 to 0-9 (inclusive). This is in-line with what the SDK implementation does.

quexten:
@harr1424 / @Thomas-Avery I had to adjust things to make the test suite pass. It seems the test suite needs textdecoder at import time because of the SDK dependency, and I did not find a cleaner way to do it other than putting the imports at the top of most spec files.

harr1424 (approved):

Thomas-Avery (approved):
KM changes LGTM

PR #18849, opened 2026-02-09
[PM-31433] Welcome Dialog with Extension Prompt

nick-livefront:
## 🎟️ Tracking

[PM-31433](https://bitwarden.atlassian.net/browse/PM-31433)

## 📔 Objective

Add welcome dialog to prompting new users to install the extension.
Shows when:
- The user's account is under 30 days old
- They have not interacted the dialog yet
- They do not have the extension installed
This shows after the premium upgrade prompt.

## 📸 Screenshots

|New user (skips all)|Open Chrome Store| Open Safari Store|
|-|-|-|
|<video src="https://github.com/user-attachments/assets/6d7fc7cc-88fb-430f-890c-bc70f0d0e094" />|<video src="https://github.com/user-attachments/assets/e0c2c7c4-bfeb-47f9-a093-829d76ffac2e" />|<video src="https://github.com/user-attachments/assets/00e4e3c4-8b37-4225-b3e5-049b4e217056" />|

[PM-31433]: https://bitwarden.atlassian.net/browse/PM-31433?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/web/src/app/vault/components/web-vault-extension-prompt/web-vault-extension-prompt-dialog.component.html
```diff
@@ -0,0 +1,34 @@
+<div
+  class="tw-max-w-3xl tw-p-9 tw-border tw-border-solid tw-border-border-base tw-rounded-xl tw-bg-background tw-text-center"
+>
+  <img
+    src="/images/vault/extension-mock-login.png"
+    alt="Extension Mock Login"
```
shane-melton: ⚠️ Missing i18n string for `alt` tag

apps/web/src/locales/en/messages.json
```diff
@@ -12852,6 +12852,21 @@
   },
   "storageFullDescription": {
     "message": "You have used all $GB$ GB of your encrypted storage. To continue storing files, add more storage."
+  },
+    "extensionPromptHeading": {
```
shane-melton: ⛏️ Looks like there are some extra spaces here.

apps/web/src/app/vault/components/web-vault-extension-prompt/web-vault-extension-prompt-dialog.component.ts
```diff
@@ -0,0 +1,59 @@
...
+  ButtonModule,
+  DialogModule,
+  DialogRef,
+  DialogService,
+  IconComponent,
+} from "@bitwarden/components";
+import { StateProvider, UserKeyDefinition, WELCOME_EXTENSION_DIALOG_DISK } from "@bitwarden/state";
+import { I18nPipe } from "@bitwarden/ui-common";
+
+export const WELCOME_EXTENSION_DIALOG_DISMISSED = new UserKeyDefinition<boolean>(
```
shane-melton: ❓ 🎨  Thoughts on moving this state definition into the `web-vault-extension-prompt` service instead of in the component? The service could also provide a method of retrieving the value so the component wouldn't need to interact with state provider at all.

It also would keep the imports a bit cleaner on the service which is currently importing from the component/dialog; which seems flipped imo.

apps/web/src/app/vault/services/web-vault-extension-prompt.service.ts
```diff
@@ -0,0 +1,91 @@
...
+  }
+
+  /**
+   * Returns true if the user's profile is within the defined thresholds for showing the extension prompt, false otherwise.
+   * Thresholds are defined as account age between a configurable number of days and 30 days.
+   */
+  private async profileIsWithinThresholds() {
+    const creationDate = await firstValueFrom(
+      this.accountService.activeAccount$.pipe(
+        map((account) => account?.creationDate ?? new Date()),
```
shane-melton: ❓I believe we default to `new Date()` for similar "welcome" features; just wanted to double check that is still want we want for this welcome dialog. If so, perhaps we can document that this is as an explicit decision?

shane-melton (changes requested):
Nice work! One minor issue, otherwise a few suggestions/questions.

apps/web/src/locales/en/messages.json
nick-livefront: 🦅 👁️ [865ed89](https://github.com/bitwarden/clients/pull/18849/commits/865ed890469b2263bba5b27c4a4de69f8f071b96)

apps/web/src/app/vault/components/web-vault-extension-prompt/web-vault-extension-prompt-dialog.component.html
nick-livefront: 🤦 And I had one defined already: [d9168f1](https://github.com/bitwarden/clients/pull/18849/commits/d9168f16866f623ade344c6de245728f9fc10194)

apps/web/src/app/vault/services/web-vault-extension-prompt.service.ts
nick-livefront: Good call, I actually removed the `new Date` and let the null/undefined be present to make this case explicitly defined: [9be0cb8](https://github.com/bitwarden/clients/pull/18849/commits/9be0cb8eb6b8a516442cb5c7363fb66a9c0fcd9c)

apps/web/src/app/vault/components/web-vault-extension-prompt/web-vault-extension-prompt-dialog.component.ts
nick-livefront: I love it: [ec6f8bf](https://github.com/bitwarden/clients/pull/18849/commits/ec6f8bf695f5b084196b67678092f33e3498af09)

shane-melton (dismissed):
🚀

djsmith85 (dismissed):
Approving for platform

nick-livefront:
@shane-melton @djsmith85 I didn't realize all of the failures I had were from a type change introduced by a numeric feature flag. I had to update some types: [11f976c](https://github.com/bitwarden/clients/pull/18849/commits/11f976c1720932ef55dbcfe3527c9cf678f7a060)

libs/common/src/enums/feature-flag.enum.ts
```diff
@@ -136,6 +138,8 @@ export const DefaultFeatureFlagValue = {
   [FeatureFlag.BrowserPremiumSpotlight]: FALSE,
   [FeatureFlag.PM27632_SdkCipherCrudOperations]: FALSE,
   [FeatureFlag.MigrateMyVaultToMyItems]: FALSE,
+  [FeatureFlag.PM29438_WelcomeDialogWithExtensionPrompt]: FALSE,
+  [FeatureFlag.PM29438_DialogWithExtensionPromptAccountAge]: 5,
```
vleague2, line 142: Is this kind of functioning as a const and not a feature flag? I haven't done much work with feature flags, but I'm not really sure how a number would actually work

apps/desktop/src/autofill/services/desktop-autofill.service.ts
```diff
@@ -51,7 +51,7 @@ import type { NativeWindowObject } from "./desktop-fido2-user-interface.service"
 export class DesktopAutofillService implements OnDestroy {
   private destroy$ = new Subject<void>();
   private registrationRequest: autofill.PasskeyRegistrationRequest;
-  private featureFlag?: FeatureFlag;
+  private featureFlag?: FeatureFlag.MacOsNativeCredentialSync;
```
coltonhurst: (Looking at this only from a team-autofill-desktop-dev standpoint)

I haven't looked deeply into this - so maybe there is an obvious answer - but why are we changing this to set the feature flag directly here? Below on line 67 there is logic to only set it to `FeatureFlag.MacOsNativeCredentialSync` if the device type is `DeviceType.MacOsDesktop`.

libs/common/src/enums/feature-flag.enum.ts
nick-livefront: The `5` here is the default value and essentially a `const` if the configuration isn't available. Using a number let's anyone in LaunchDarkly set the number of days that this prompt would show after without having to push a release.

apps/desktop/src/autofill/services/desktop-autofill.service.ts
nick-livefront: Previously the `featureFlag` variable here was inferred to have a type `boolean` because all feature flags were of type `boolean`. In this PR I'm introducing a numeric feature flag which changed the inferred type  to `boolean | number`.  

This caused a type error when [setting isEnabled here](https://github.com/bitwarden/clients/blob/115ac32675489b642ec719a31d2c1850915a8a45/apps/desktop/src/autofill/services/desktop-autofill.service.ts#L73). Without this change, when `this.featureFlag` is falsey it would attempt to set `isEnabled` to `boolean | number` when it is defined as a `boolean`, thus causing an error. 

I added the `typeof` keyword to hopefully cut down any confusion as this is creating a more strict type. [115ac32](https://github.com/bitwarden/clients/pull/18849/commits/115ac32675489b642ec719a31d2c1850915a8a45)

libs/common/src/enums/feature-flag.enum.ts
vleague2: Interesting, I feel like the semantics and temporary nature of "feature flag" don't really match the usage, but that is outside of my scope of concerns haha

vleague2 (approved):
UIF files are good!

PR #18850, opened 2026-02-09
[PM-29529]Remove @ts-strict-ignore in background-notification.background.ts

dan-livefront:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-29529
## 📔 Objective

The changes focus on handling null/undefined cases properly and tightening up types, without changing behavior.

The main updates include adding small helpers for safe tab access, using early returns when required data is missing, and cleaning up type annotations where inference wasn’t reliable. A few related types and tests were adjusted to match the new, safer behavior.

Also, after unlocking, the retry logic could re invoke handlers with a sender that had no tab. Some handlers assume sender.tab always exists, which could lead to errors or incorrect behavior during retries.

Updated the test to supply a retry sender that includes a valid tab, matching what production expects. This ensures retries run with the correct sender shape instead of relying on undefined behavior.


apps/browser/src/autofill/background/notification.background.ts
```diff
@@ -1115,9 +1151,16 @@ export default class NotificationBackground {
         inputScenario === inputScenarios.usernamePasswordNewPassword &&
         newLoginNotificationIsEnabled
       ) {
+        if (data?.username == null || data.username === "") {
```
blackwood: mixed comparison throughout; use strict comparison everywhere.

apps/browser/src/autofill/background/notification.background.ts
```diff
@@ -1213,34 +1262,37 @@ export default class NotificationBackground {
...
     sender: chrome.runtime.MessageSender,
   ) {
-    if ((await this.getAuthStatus()) < AuthenticationStatus.Unlocked) {
-      await BrowserApi.tabSendMessageData(sender.tab, "addToLockedVaultPendingNotifications", {
-        commandToRetry: {
-          message: {
-            command: message.command,
-            edit: message.edit,
-            folder: message.folder,
+    await this.withSenderTab(sender, async (tab) => {
```
blackwood: This function `handleSaveCipherMessage` is already wrapped with the HOF `withSenderTab` in the `extensionMessageHandlers` above, is that intentional?

blackwood: With the HOF used in the body of the method, the inner function's `return` no longer exits the method, just the lambda. May not impact current behavior, but could unintentionally misdirect a future reader if other top-level procedures end up in this method.

apps/browser/src/autofill/background/notification.background.ts
```diff
@@ -177,6 +178,26 @@ export default class NotificationBackground {
...
+    sender: chrome.runtime.MessageSender,
+    action: (tab: chrome.tabs.Tab) => void | Promise<void>,
+  ): Promise<void> {
+    if (!sender.tab) {
+      return;
+    }
+    await action(sender.tab);
+  }
+
+  private async withMessageTab(
```
blackwood: I feel uncertain about this second helper method—it only has one call site.

apps/browser/src/autofill/background/notification.background.ts
```diff
@@ -177,6 +178,26 @@ export default class NotificationBackground {
     this.cleanupNotificationQueue();
   }
 
+  private async withSenderTab(
+    sender: chrome.runtime.MessageSender,
+    action: (tab: chrome.tabs.Tab) => void | Promise<void>,
+  ): Promise<void> {
+    if (!sender.tab) {
+      return;
```
blackwood, line 190: 💭 Thinking about whether we want a `console.warn` here before the `return`, since it seems like an exception worth investigating. Thoughts?

dan-livefront: I agree went ahead and added, good callout!

apps/browser/src/autofill/background/notification.background.ts
dan-livefront: went ahead and inlined

apps/browser/src/autofill/background/notification.background.ts
dan-livefront: Good callout, the tab check was redundant, removed and also removed withSenderTab to avoid the callback not exiting the method 🧹

PR #18853, opened 2026-02-09
[PM-26704] Vault List Item Ordering for Extension

jengstrom-bw:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-26704

## 📔 Objective

Current functionality is that any item in Autofill suggestions or Favorites is not displayed in the all items list. Update this behavior so that any item displayed in these priority sections are repeated in the “All items” section as well.


apps/browser/src/vault/popup/components/vault/vault.component.html
```diff
@@ -127,7 +127,7 @@
             <!--Change the title header when a filter is applied-->
             <app-vault-list-items-container
               [title]="((numberOfAppliedFilters$ | async) === 0 ? 'allItems' : 'items') | i18n"
-              [ciphers]="(remainingCiphers$ | async) || []"
+              [ciphers]="(filteredCiphers$ | async) || []"
```
shane-melton, line 130: 🎨 I believe this was the only use of `remainingCiphers$` and we can probably clean that up in the `vault-v2.component.ts` and the `vault-popup-items.service.ts`.

jengstrom-bw: will do!

apps/browser/src/vault/popup/services/vault-popup-items.service.ts
```diff
@@ -243,31 +242,12 @@ export class VaultPopupItemsService {
...
-    shareReplay({ refCount: false, bufferSize: 1 }),
-  );
-
   /**
    * Observable that indicates whether the service is currently loading ciphers.
    */
   loading$: Observable<boolean> = merge(
     this._ciphersLoading$.pipe(map(() => true)),
-    this.remainingCiphers$.pipe(map(() => false)),
+    this.filteredCiphers$.pipe(map(() => false)),
```
shane-melton: ⛏️ `filteredCiphers$` will emit before `favoriteCiphers$` or `autoFillCiphers$` will emit. Meaning we could potentially set loading to false before those other 2 observables are ready and cause the UI to jump around while they're processing.

I think we'll want it to be `favoriteCiphers$` here so that we know everything is "done" before we set `loading = false`.

shane-melton (changes requested):
One small nit

apps/browser/src/vault/popup/services/vault-popup-items.service.ts
jengstrom-bw: Makes sense, want to know when the farthest down the "chain" is done, which is now `favoriteCiphers$` instead of `remainingCiphers`

Thanks for explaining!

shane-melton (approved):
Looks good, thanks!

PR #18860, opened 2026-02-09
[PM-31958] BUGFIX: Blank Screen showing after deleting item

nikwithak:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-31958

## 📔 Objective

Calling clearCache after the delete operation causes the cache not to refresh until it's updated. Calling clearCache before the delete resolves this issue by allowing the cache to rebuild after `deleteWithServer` is called. APpears to be a race condition, affecting primarily Chrome. Similar fix to https://github.com/bitwarden/clients/pull/18790.


libs/common/src/vault/services/cipher.service.ts
```diff
@@ -1692,8 +1692,8 @@ export class CipherService implements CipherServiceAbstraction {
   async restoreManyWithServer(ids: string[], userId: UserId, orgId?: string): Promise<void> {
     const useSdk = await firstValueFrom(this.sdkCipherCrudEnabled$);
     if (useSdk) {
-      await this.cipherSdkService.restoreManyWithServer(ids, userId, orgId);
       await this.clearCache(userId);
+      await this.cipherSdkService.restoreManyWithServer(ids, userId, orgId);
```
gbubemismith, line 1696: does this also happen when restoring an item?

nikwithak: Yes, it was kicked back from QA for failing at restore too. I scanned the remaining SDK integrations so far and this seems to be the last of them.

PR #18861, opened 2026-02-10
Uif/cl 1028/popover anchor border radius

jaasen-livefront:
## 📔 Objective

This PR adds border-radius option for the spotlight service so it can be used with rounded elements.


jaasen-livefront:
@willmartian You're absolutely right! Closing.

PR #18869, opened 2026-02-10
[PM-30993] More Descriptive Message on Expired Tokens

sven-bitwarden:
## 🎟️ Tracking
https://bitwarden.atlassian.net/browse/PM-30993

## 📔 Objective
The bulk of the work is done in: https://github.com/bitwarden/server/pull/6971

The current error message to users is nondescript and merely says "Invalid token.", but in the case of an expired token, we should instruct them to ask their org admin for another one. This simply keys off the new error message from the server, showing the `failedMessage` for expired tokens.

### Token Expired
image

### Any Other Error
Screenshot 2026-02-10 at 2 25 21 PM

### DIFF
--- a/apps/browser/src/vault/popup/components/vault/intro-carousel/intro-carousel.component.ts
+++ b/apps/browser/src/vault/popup/components/vault/intro-carousel/intro-carousel.component.ts
@@ -24,14 +24,14 @@ import { IntroCarouselService } from "../../../services/intro-carousel.service";
   ],
 })
 export class IntroCarouselComponent {
-  protected itemTypes = ItemTypes;
-  protected loginCards = LoginCards;
-  protected noCredentials = NoCredentialsIcon;
-  protected secureDevices = DevicesIcon;
+  protected readonly itemTypes = ItemTypes;
+  protected readonly loginCards = LoginCards;
+  protected readonly noCredentials = NoCredentialsIcon;
+  protected readonly secureDevices = DevicesIcon;
 
   constructor(
-    private router: Router,
-    private introCarouselService: IntroCarouselService,
+    private readonly router: Router,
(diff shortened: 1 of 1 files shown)


apps/web/src/app/common/base.accept.component.ts
```diff
@@ -62,7 +62,7 @@ export abstract class BaseAcceptComponent implements OnInit {
 
           if (error) {
             const message =
-              errorMessage != null
+              errorMessage != null && errorMessage !== "Expired token."
```
JaredSnider-Bitwarden: I think this slightly violates separation of concerns as it relates to the different types of accept flows we have. We've only modified the organization invite acceptance flow to add the possibility of getting the "Expired token." message, but we've modified the base component's error handling (thus, affecting all invite flows). 

We could use the template method pattern here to make this a bit more targeted. Claude generated (untested) implementation:
```
// In base.accept.component.ts
protected getErrorMessage(errorMessage: string | null): string {
  return errorMessage != null 
    ? this.i18nService.t(this.failedShortMessage, errorMessage)
    : this.i18nService.t(this.failedMessage);
}

// Then in ngOnInit (line 63-67):
if (error) {
  const message = this.getErrorMessage(errorMessage);
  this.platformUtilService.showToast("error", null, message, { timeout: 10000 });
  // ...
}

// In accept-organization.component.ts
protected override getErrorMessage(errorMessage: string | null): string {
  // Handle expired token specifically for org invites
  if (errorMessage === "Expired token.") {
    return this.i18nService.t(this.failedMessage);
  }
  return super.getErrorMessage(errorMessage);
}
```

sven-bitwarden: Great catch, will follow suit to ensure the change is only for accepting org invites.

PR #18870, opened 2026-02-10
[PM-31763] Add unlock service & module

quexten:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-31763

## 📔 Objective

Adds a new module - the unlock module - which is responsible for unlocking a user's client. It does this via the SDK. It needs https://github.com/bitwarden/sdk-internal/pull/702 to actually be effective at unlocking.

The unlock methods implemented here are:
- Pin
- Master Password
- Biometrics


.github/CODEOWNERS
```diff
@@ -249,3 +249,4 @@ apps/desktop/native-messaging-test-runner/package-lock.json @bitwarden/team-plat
 .github/workflows/respond.yml @bitwarden/team-ai-sme
 .github/workflows/review-code.yml @bitwarden/team-ai-sme
 libs/subscription @bitwarden/team-billing-dev
+libs/unlock @bitwarden/team-key-management-dev
```
djsmith85: ⛏️ This is probably caused by the create lib cli command. Could you please move this to the KeyManagement area above and while you are at it, please also move libs/subscription to Billing.

quexten: Sure, moved. And yes, it's auto-created like this by the `nx` command.

quexten:
Moving back to draft for a minute; I'm changing how this PR set is structured.

libs/unlock/src/default-unlock.service.ts
```diff
@@ -0,0 +1,227 @@
...
+    userId: UserId,
+  ): Promise<void> {
+    assertNonNullish(password, "password");
+    assertNonNullish(masterPasswordUnlockData, "masterPasswordUnlockData");
+    assertNonNullish(userId, "userId");
+    this.logService.info("[DefaultUnlockService] Setting legacy master key from unlock data");
+
+    // NOTE: This entire section is deprecated and will be removed as soon as
+    // the masterkey is dropped from state. It is very temporary.
+    await SdkLoadService.Ready;
```
quexten, line 205: Note: I chose note to pull in any existing services here but copy in the implementation deliberately, so that those sections of code can be dropped quicker. Further, this section can also be dropped as soon as the master-key does not need to be set anymore.

libs/unlock/src/default-unlock.service.ts
```diff
@@ -0,0 +1,227 @@
...
+  async unlockWithPin(userId: UserId, pin: string): Promise<void> {
+    const startTime = performance.now();
+    await firstValueFrom(
+      this.registerSdkService.registerClient$(userId).pipe(
+        map(async (sdk) => {
+          if (!sdk) {
+            throw new Error("SDK not available");
+          }
+          using ref = sdk.take();
+          return ref.value.crypto().initialize_user_crypto({
```
coroiu, line 56: Interesting way of doing it :) Using the registerClient to set crypto which then creates a new regular SDK client

quexten: Yeah... This is very much a transitionary period. When we move more state to the SDK we can consolidate again...

PR #18874, opened 2026-02-10
[PM-31938] refactor archive btn logic in web view modal

Jingo88:
## 🎟️ Tracking

[PM-31938](https://bitwarden.atlassian.net/browse/PM-31938)

## 📔 Objective

* The `unarchive` button in the view modal will always show no matter what the users premium status is
* move `userCanArchive` logic check from `showArchiveOptions` to only `showArchiveBtn`. 
* Add flag check to `showArchiveOptions` 

## 📸 Screen Recording

https://github.com/user-attachments/assets/3e9ba532-d9c8-4ee1-a4ea-237a2b985f73

[PM-31938]: https://bitwarden.atlassian.net/browse/PM-31938?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/web/src/app/vault/components/vault-item-dialog/vault-item-dialog.component.ts
+++ b/apps/web/src/app/vault/components/vault-item-dialog/vault-item-dialog.component.ts
@@ -10,7 +10,7 @@ import {
   OnInit,
   viewChild,
 } from "@angular/core";
-import { takeUntilDestroyed } from "@angular/core/rxjs-interop";
+import { takeUntilDestroyed, toSignal } from "@angular/core/rxjs-interop";
 import { Router } from "@angular/router";
 import { firstValueFrom, Observable, Subject, switchMap } from "rxjs";
 import { map } from "rxjs/operators";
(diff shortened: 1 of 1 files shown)


apps/web/src/app/vault/components/vault-item-dialog/vault-item-dialog.component.ts
```diff
@@ -357,6 +360,9 @@ export class VaultItemDialogComponent implements OnInit, OnDestroy {
       .subscribe();
 
     this.userCanArchive$.pipe(takeUntilDestroyed()).subscribe((v) => (this._userCanArchive = v));
+    this.archiveFlagEnabled$
```
shane-melton: 🎨 These (`userCanArchive$` and `archivedFlagEnabled$`) should probably be signals so we can avoid this cumbersome subscribe and `takeUntilDestroyed` pattern

PR #18880, opened 2026-02-10
SSH Agent v2: define napi-agent-server interface

neuronull:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-30753

## 📔 Objective

Building upon the prior SSH agent v2 commits, this PR introduces the high level interfaces between the napi layer, the agent, and the underlying server. As with the other commits so far, the code is not live yet; no electron service is consuming it.

Included below are (rough) diagrams of the architecture and a sequence diagram of the signing of a request. The diagrams aren't perfect representations but do a decent job conveying the relationships.

I did end up implementing some of the logic to connect these pieces. That became necessary for rationalizing the design. If any reviewer feels review fatigue from this PR, let me know and we can discuss separating out the AuthPolicy implementation if that would help.

Notable design details:
- Two traits are defined to act as abstraction layers:
    - AuthPolicy allows the underlying server to externally request authorization to complete the various ssh operation requests it receives. The implementation of this in the BitwardenSshAgent is effectively the agent's business logic, and has unit tests accordingly.
    - ApprovalRequester allows the agent to externally request approval for those authorization requests.
    - These layers allow better unit test-ability and integration test-ability, and additionally are (at least imo) as context-independent as possible from eachother; meaning the ssh agent server doesn't know about vault items, or that there is an electron app at all. Similarly, the agent knows it needs to request approval, but it doesn't have to be from an Electron app.
    
Divergences from existing designs:
- The napi layer now has two callbacks, in order to better separate concerns and allow type safety.
- The async/callback handling diverged significantly, I believe this approach provides the functionality we need with the least amount of messaging infrastructure.
    - Yes, these will need to be handled in the electron services accordingly but I think it's a valuable and not overly intrusive improvement.

## 📸 Screenshots

ssh-agent-v2-architecture

ssh-agent-v2-sequence-sign-request

### DIFF
--- a/apps/desktop/desktop_native/napi/src/sshagent_v2.rs
+++ b/apps/desktop/desktop_native/napi/src/sshagent_v2.rs
@@ -7,11 +7,10 @@
 pub mod sshagent_v2 {
     use std::{sync::Arc, time::Duration};
 
-    use anyhow::Context;
     use async_trait::async_trait;
     use napi::threadsafe_function::ThreadsafeFunction;
     use ssh_agent::{
-        ApprovalRequester, BitwardenSSHAgent, InMemoryEncryptedKeyStore,
+        ApprovalError, ApprovalRequester, BitwardenSSHAgent, InMemoryEncryptedKeyStore,
         SIGNamespace as SSHSIGNamespace, SignApprovalRequest as SSHSignApprovalRequest,
     };
     use tokio::time::timeout;
@@ -113,17 +112,17 @@ pub mod sshagent_v2 {
 
     #[async_trait]
     impl ApprovalRequester for ElectronApprovalRequester {
-        async fn request_unlock(&self) -> anyhow::Result<bool> {
+        async fn request_unlock(&self) -> Result<bool, ApprovalError> {
             debug!("Sending unlock request to Electron.");
 
             let is_approved = timeout(APPROVAL_CALLBACK_TIMEOUT, async {
                 self.unlock_callback
                     .call_async(Ok(()))
                     .await
-                    .context("Electron unlock callback failed")
+                    .map_err(|e| ApprovalError::HandlerFailed(e.into()))
             })
             .await
-            .context("Electron unlock callback timed out")
+            .map_err(|_| ApprovalError::Timeout)
             .flatten()?;
 
             debug!(%is_approved, "Unlock response from Electron.");
@@ -133,7 +132,7 @@ pub mod sshagent_v2 {
         async fn request_sign_approval(
             &self,
             request: SSHSignApprovalRequest,
-        ) -> anyhow::Result<bool> {
+        ) -> Result<bool, ApprovalError> {
             let request = SignRequestData::from(request);
 
             debug!(?request, "Sending sign approval request to Electron.");
@@ -142,10 +141,10 @@ pub mod sshagent_v2 {
                 self.sign_callback
                     .call_async(Ok(request))
                     .await
-                    .context("Electron sign callback failed")
+                    .map_err(|e| ApprovalError::HandlerFailed(e.into()))
             })
             .await
-            .context("Electron sign callback timed out")
+            .map_err(|_| ApprovalError::Timeout)
             .flatten()?;
 
             debug!(%is_approved, "Sign approval response from Electron.");
--- a/apps/desktop/desktop_native/ssh_agent/src/authorization.rs
+++ b/apps/desktop/desktop_native/ssh_agent/src/authorization.rs
@@ -6,10 +6,10 @@ use std::sync::{
 };
 
 use thiserror::Error;
-use tracing::{debug, error};
+use tracing::debug;
 
 use crate::{
-    approval::{ApprovalRequester, SignApprovalRequest},
+    approval::{ApprovalError, ApprovalRequester, SignApprovalRequest},
     crypto::QueryableKeyData,
     server::{AuthPolicy, AuthRequest},
     storage::keystore::KeyStore,
@@ -18,9 +18,9 @@ use crate::{
 /// Errors that can occur during authorization of SSH agent operations.
 #[derive(Debug, Error)]
 pub enum AuthError {
-    /// The approval handler failed to process the request.
-    #[error("Approval handler failed: {0}")]
-    HandlerFailed(#[source] anyhow::Error),
+    /// The approval handler did not receive an approved/denied response.
+    #[error(transparent)]
+    ApprovalUnresolved(#[from] ApprovalError),
 
     /// The requested public key was not found in the keystore.
     #[error("Public key not found in keystore")]
@@ -95,14 +95,7 @@ where
 
                 // otherwise request unlock from approval handler
                 debug!("Locked, requesting unlock approval.");
-                let unlock_approved =
-                    self.approval_handler
-                        .request_unlock()
-                        .await
-                        .map_err(|error| {
-                            error!(%error, "Unlock request failed.");
-                            AuthError::HandlerFailed(error)
-                        })?;
+                let unlock_approved = self.approval_handler.request_unlock().await?;
 
                 debug!(unlock_approved, "Unlock response received.");
 
@@ -130,10 +123,7 @@ where
                         cipher_id,
                     })
                     .await
-                    .map_err(|error| {
-                        error!(%error, "Approval handler failed.");
-                        AuthError::HandlerFailed(error)
-                    })
+                    .map_err(Into::into)
             }
         }
     }
@@ -146,7 +136,7 @@ mod tests {
 
     use super::*;
     use crate::{
-        approval::MockApprovalRequester,
+        approval::{ApprovalError, MockApprovalRequester},
         server::SIGNamespace,
         storage::{keydata::MockQueryableKeyData, keystore::MockKeyStore},
     };
@@ -280,7 +270,7 @@ mod tests {
         approval_handler
             .expect_request_unlock()
             .times(1)
-            .returning(|| Err(anyhow!("Handler failed")));
+            .returning(|| Err(ApprovalError::HandlerFailed(anyhow!("Handler failed"))));
 
         let policy = BitwardenAuthPolicy::new(keystore, approval_handler);
 
@@ -288,8 +278,13 @@ mod tests {
         let result = policy.authorize(&request).await;
 
         assert!(
-            matches!(result, Err(AuthError::HandlerFailed(_))),
-            "Should return HandlerFailed error when unlock handler fails"
+            matches!(
+                result,
+                Err(AuthError::ApprovalUnresolved(ApprovalError::HandlerFailed(
+                    _
+                )))
+            ),
+            "Should return ApprovalUnresolved error when unlock handler fails"
         );
     }
 
@@ -404,7 +399,7 @@ mod tests {
         approval_handler
             .expect_request_sign_approval()
             .times(1)
-            .returning(|_| Err(anyhow!("Handler failed")));
+            .returning(|_| Err(ApprovalError::HandlerFailed(anyhow!("Handler failed"))));
 
         let policy = BitwardenAuthPolicy::new(Arc::new(keystore), approval_handler);
 
@@ -412,8 +407,13 @@ mod tests {
         let result = policy.authorize(&request).await;
 
         assert!(
-            matches!(result, Err(AuthError::HandlerFailed(_))),
-            "Should return HandlerFailed error"
+            matches!(
+                result,
+                Err(AuthError::ApprovalUnresolved(ApprovalError::HandlerFailed(
+                    _
+                )))
+            ),
+            "Should return ApprovalUnresolved error"
         );
     }
 
--- a/apps/desktop/desktop_native/ssh_agent/src/lib.rs
+++ b/apps/desktop/desktop_native/ssh_agent/src/lib.rs
@@ -31,7 +31,7 @@ mod storage;
 
 // external exports for napi
 pub use agent::BitwardenSSHAgent;
-pub use approval::{ApprovalRequester, SignApprovalRequest};
+pub use approval::{ApprovalError, ApprovalRequester, SignApprovalRequest};
 pub use crypto::PublicKey;
 pub use server::{AuthRequest, SIGNamespace, SignRequest};
 pub use storage::keystore::InMemoryEncryptedKeyStore;
--- a/apps/desktop/desktop_native/ssh_agent/src/approval.rs
+++ b/apps/desktop/desktop_native/ssh_agent/src/approval.rs
@@ -2,7 +2,21 @@
 //! to be able to externally request approval for ssh
 //! authorization requests.
 
+use thiserror::Error;
+
 use crate::server::SignRequest;
+
+/// Errors that can occur when requesting approval from an external handler.
+#[derive(Debug, Error)]
+pub enum ApprovalError {
+    /// The handler did not respond within the allowed time.
+    #[error("Approval request timed out")]
+    Timeout,
+
+    /// The handler was invoked but encountered a failure during processing.
+    #[error("Approval handler failed: {0}")]
+    HandlerFailed(#[source] anyhow::Error),
+}
 
 /// Bundles a sign request with the vault cipher context needed to approve it.
 #[derive(Debug, Clone)]
@@ -23,11 +37,7 @@ pub trait ApprovalRequester: Send + Sync {
     ///
     /// * `Ok(true)` - Unlock was approved
     /// * `Ok(false)` - Unlock was denied
-    ///
-    /// # Errors
-    ///
-    /// If the handler failed to process the request
-    async fn request_unlock(&self) -> anyhow::Result<bool>;
+    async fn request_unlock(&self) -> Result<bool, ApprovalError>;
 
     /// Requests approval for a signing operation.
     ///
@@ -37,11 +47,10 @@ pub trait ApprovalRequester: Send + Sync {
     ///
     /// # Returns
     ///
-    /// * `Ok(true)` - Request was approved
-    /// * `Ok(false)` - Request was denied
-    ///
-    /// # Errors
-    ///
-    /// If the handler failed to process the request
-    async fn request_sign_approval(&self, request: SignApprovalRequest) -> anyhow::Result<bool>;
+    /// * `Ok(true)` - Sign was approved
+    /// * `Ok(false)` - Sign was denied
+    async fn request_sign_approval(
+        &self,
+        request: SignApprovalRequest,
+    ) -> Result<bool, ApprovalError>;
 }


neuronull:
The Desktop CI failure for validate flatpak will resolve after merging in fix for https://github.com/bitwarden/clients/pull/18814

apps/desktop/desktop_native/ssh_agent/src/server/auth_policy.rs
```diff
@@ -0,0 +1,47 @@
...
+
+use crate::{authorization::AuthError, crypto::PublicKey};
+
+/// Request to sign data using an SSH key.
+#[derive(Debug, Clone)]
+pub struct SignRequest {
+    /// The public key identifying which key to use for signing
+    pub public_key: PublicKey,
+    /// Name of the process making the request
+    pub process_name: String,
```
quexten: Should this be Option<String>? In Flatpak, Snap, Mac App Store, the process cannot be looked up and the process name would be empty.

neuronull: Good call out 👍  I suspect we'd have discovered this when implementing downstream bits. I will update it to be Option and update the doc comment.

neuronull: addressed in e720af9b583890ccafefea1c2a87d5b84ea5514f

apps/desktop/desktop_native/ssh_agent/src/authorization.rs
```diff
@@ -0,0 +1,454 @@
...
+                // otherwise request unlock from approval handler
+                debug!("Locked, requesting unlock approval.");
+                let unlock_approved =
+                    self.approval_handler
+                        .request_unlock()
+                        .await
+                        .map_err(|error| {
+                            error!(%error, "Unlock request failed.");
+                            AuthError::HandlerFailed(error)
+                        })?;
```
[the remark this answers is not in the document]
neuronull: added in https://github.com/bitwarden/clients/pull/18880/commits/a8fb904454a66941b962819e783c039317f66266

apps/desktop/desktop_native/ssh_agent/src/agent.rs
```diff
@@ -0,0 +1,99 @@
...
+
+    /// Updates the keystore with the new keys and signals the auth policy to an unlocked state.
+    pub fn set_keys(&self, _keys: Vec<K::KeyData>) -> Result<()> {
+        debug!("Received new key data.");
+
+        self.auth_policy.set_lock_state(LockState::Unlocked);
+
+        debug!("New key data is set.");
+        todo!()
+    }
```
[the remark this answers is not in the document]
neuronull: right but then you have to return `Ok()` after the debug, to satisfy the Result return type yo . I intentionally did it this way. But I see it confused claude, so I'll not assume it won't be confusing for anyone else, and do that I guess.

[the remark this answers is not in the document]
neuronull: also like, the todo!() statement panics, so this won't even compile yo

apps/desktop/desktop_native/ssh_agent/src/agent.rs
```diff
@@ -0,0 +1,102 @@
...
+    pub fn start_server(&mut self) -> Result<()> {
+        debug!("Starting the server.");
+        self.server.start()
+    }
+
+    /// Stops the ssh agent server
+    pub fn stop_server(&mut self) {
+        debug!("Stopping the server.");
+        self.server.stop()
+    }
```
neuronull, line 60: I left these as debug level, because I think it's better to have the server have info level messages on when it is starting and stopping. so having these be info as well would be redundant, but debug will be useful for debugging.

neuronull:
FYI this is impacted by https://github.com/bitwarden/clients/pull/19113 , will not want to merge it until that has merged.

apps/desktop/desktop_native/ssh_agent/src/lib.rs
```diff
@@ -1,7 +1,19 @@
 //! Bitwarden SSH Agent implementation
 //!
+//! TODO: add details
+//!
```
coltonhurst: Did we want to leave this in?

apps/desktop/desktop_native/ssh_agent/src/agent.rs
```diff
@@ -0,0 +1,102 @@
...
+    /// `true` if the server is running, `false` if it is not.
+    pub fn is_running(&self) -> bool {
+        self.server.is_running()
+    }
+
+    /// Updates the keystore with the new keys and signals the auth policy to an unlocked state.
+    pub fn set_keys(&self, _keys: Vec<K::KeyData>) -> Result<()> {
+        debug!("Received new key data.");
+
+        // TODO: set keys
```
coltonhurst: Can we add the ticket here that corresponds to implementing this function?

apps/desktop/desktop_native/ssh_agent/src/server/mod.rs
```diff
@@ -0,0 +1,58 @@
+//! SSH Agent Server implementation
+//!
+//! Adheres to the protocol defined in:
+//! <https://www.ietf.org/archive/id/draft-miller-ssh-agent-11.html>
```
coltonhurst: Not a big concern: It might be nice to link to something like [this](https://datatracker.ietf.org/doc/draft-ietf-sshm-ssh-agent) that has the same text viewable in multiple formats plus history info. (I gave the general link, not a version specific one, but the core idea of my thought should hopefully be clear.)

apps/desktop/desktop_native/ssh_agent/src/server/mod.rs
```diff
@@ -0,0 +1,58 @@
...
+    pub fn new(keystore: Arc<K>, auth_policy: Arc<A>) -> Self {
+        Self {
+            keystore,
+            auth_policy,
+            cancellation_token: None,
+        }
+    }
+
+    pub fn start(&mut self) -> Result<()> {
+        todo!();
```
coltonhurst, line 48: Same for the `todo!()`'s in this file, can we link ticket(s)?

apps/desktop/desktop_native/napi/src/sshagent_v2.rs
```diff
@@ -0,0 +1,174 @@
...
+    };
+    use tracing::{debug, error};
+
+    /// SSH key data, sent from Electron.
+    #[napi(object)]
+    pub struct SshKeyData {
+        pub private_key: String,
+        pub name: String,
+        pub cipher_id: String,
+    }
```
coltonhurst, line 29: Will this eventually be the `SSHKeyData` type from `apps/desktop/desktop_native/ssh_agent/src/crypto/mod.rs`, being used through `set_keys`?

If so, do we need the public key here?

apps/desktop/desktop_native/napi/src/sshagent_v2.rs
```diff
@@ -0,0 +1,174 @@
...
+    use napi::threadsafe_function::ThreadsafeFunction;
+    use ssh_agent::{
+        ApprovalRequester, BitwardenSshAgent, InMemoryEncryptedKeyStore,
+        SignRequest as SshSignRequest,
+    };
+    use tracing::{debug, error};
+
+    /// SSH key data, sent from Electron.
+    #[napi(object)]
+    pub struct SshKeyData {
```
coltonhurst: Just out of curiosity, in `apps/desktop/desktop_native/ssh_agent/src/crypto/mod.rs` we define `SSHKeyData` but here `SshKeyData` (and `Ssh...` in other places, etc). Do we want to be consistent on casing? Maybe there is a reason I don't know about.

apps/desktop/desktop_native/ssh_agent/src/authorization.rs
```diff
@@ -0,0 +1,480 @@
...
+    use super::*;
+    use crate::{
+        approval::MockApprovalRequester,
+        crypto::{keystore::MockKeyStore, MockQueryableKeyData},
+    };
+
+    fn create_test_public_key() -> crate::crypto::PublicKey {
+        crate::crypto::PublicKey {
+            alg: "ssh-ed25519".to_string(),
+            blob: vec![1, 2, 3],
```
coltonhurst, line 147: Based on the name `create_test_public_key` I would assume the blob would be a real key? Maybe that's not needed though.

apps/desktop/desktop_native/ssh_agent/src/lib.rs
neuronull: thanks for flagging, I wanted to fill it out after design is solidified. I think that's achievable now.

neuronull:
> Besides some small comments I have, I think it would be helpful (at least for me, so maybe also for others?) for updates to be made to the [ssh agent deep dive section](https://contributing.bitwarden.com/architecture/deep-dives/ssh) of the contributing docs.

To be totally honest, if I knew about that documentation at one time, I haven't looked at it in ages, probably since before Autotype MVP. I certainly didn't look at it when I was working on this v2 project yet. Thanks for calling it out. I added a ticket to track updating that. I think it makes sense to do after v2 has shipped.

apps/desktop/desktop_native/napi/src/sshagent_v2.rs
neuronull: The public key is derived from the private 👍

apps/desktop/desktop_native/ssh_agent/src/server/mod.rs
neuronull: If you feel strongly about it, I can add them, but all of them are addressed in https://github.com/bitwarden/clients/pull/19119 , which is in a review feedback loop already.

apps/desktop/desktop_native/napi/src/sshagent_v2.rs
neuronull: (adding a comment to this effect)

apps/desktop/desktop_native/ssh_agent/src/authorization.rs
neuronull: A real key isn't necessary to do the unit testing in this module. I changed the name to `create_stub_public_key` instead.

When we add integration tests as part of the future tickets, the behavior of a real key will be exercised there.

apps/desktop/desktop_native/ssh_agent/src/server/mod.rs
coltonhurst: Sounds good!

apps/desktop/desktop_native/ssh_agent/src/server/auth_policy.rs
```diff
@@ -0,0 +1,48 @@
...
+pub struct SignRequest {
+    /// The public key identifying which key to use for signing
+    pub public_key: PublicKey,
+    /// Name of the process making the request. If the agent is running in sandboxed environments,
+    /// it may not have access to the process name.
+    pub process_name: Option<String>,
+    /// Whether this is an agent forwarding request
+    pub is_forwarding: bool,
+    /// Optional namespace for SSH signature requests
+    pub namespace: Option<String>,
```
quexten: A thought: Should we make namespace an enum or, `Option<SshSigNamespace>`
```
/// The SSHSig namespace
enum SshSigNamespace {
  Git
  File
  Unknown(String)
}
```

I don't think much else uses the SSHSig namespace? If we then ever want to build policies based on allowing Git signing for a key, but not anything else for example, this would become simpler.

iinuwa:
@neuronull, #19113 is merged.

apps/desktop/desktop_native/ssh_agent/src/server/auth_policy.rs
neuronull: Generally sounds good to me.

Looking at the original implementation and it's usage, correct me if I'm wrong: we currently only utilize the cases `None` or `git` ? In which case, I'd suggest we not add `File` yet, but add it when we need it and the enum makes that easy to do. (but maybe I missed where that is used).

I'd also suggest we call it `Unsupported` over `Unknown`. To convey that whatever the resulting namespace that was successfully parsed, we don't have explicit handling to support it.

WDYT?

```rust
enum SignRequestNamespace {
  Git,
  Unsupported(String),
}
```

edit: I'll add that and it's easy enough to iterate on if needed

neuronull: Ah, since the napi doesn't support associated data in automatic conversions of enum variants, I remove the `(String)` from unsupported. Since the electron won't use the actual data there. When parsing, we can use debug level to log the namespace in the event it is present but not Git.

quexten: Re: File

If you sign files, the file namespace will be used. This being available is unclear from "Unsupported", and the enum being there makes discoverability better. I would add it just for discoverability / logging.

neuronull: thanks for clarifying. Since we don't explicitly have logic to parse the case for file, or handle it in the current implementation, I didn't know this. 👍 I will add it.

coltonhurst (approved):

coltonhurst:
Approved, thanks @neuronull 

@quexten is not a required reviewer but b/c he has more domain knowledge on the SSH Agent it might be good to have an approval from him too, prior to merging. Just an idea.

neuronull:
> Approved, thanks @neuronull
> 
> @quexten is not a required reviewer but b/c he has more domain knowledge on the SSH Agent it might be good to have an approval from him too, prior to merging. Just an idea.

Absolutely. That was the plan that was communicated and acknowledged in slack and no intentions here of diverging from that.

apps/desktop/desktop_native/napi/src/sshagent_v2.rs
djsmith85, line 1: ❌  Please add an entry in CODEOWNERS for this file.

`apps/desktop/desktop_native/napi/src/sshagent_v2.rs @bitwarden/team-autofill-desktop-dev`

djsmith85 (changes requested):

apps/desktop/desktop_native/napi/src/sshagent_v2.rs
neuronull: good catch. added in https://github.com/bitwarden/clients/pull/18880/commits/28b37e8031c9bf32f57ae4b108b5ab635d85acf8

djsmith85 (dismissed):

coltonhurst (approved):

apps/desktop/desktop_native/napi/src/sshagent_v2.rs
```diff
@@ -0,0 +1,209 @@
...
+
+    /// SSH public key data
+    #[napi(object)]
+    #[derive(Debug, Clone)]
+    pub struct PublicKey {
+        pub alg: String,
+        pub blob: Vec<u8>,
+    }
+
+    /// Namespace of a sign request.
```
quexten: SSHSig request. Regular sign requests for authentication to a server are separate:
https://github.com/openssh/openssh-portable/blob/master/PROTOCOL.sshsig

apps/desktop/desktop_native/napi/src/sshagent_v2.rs
```diff
@@ -0,0 +1,209 @@
...
+    }
+
+    /// SSH sign request fields.
+    #[napi(object)]
+    #[derive(Debug)]
+    pub struct SignRequest {
+        pub public_key: PublicKey,
+        pub process_name: Option<String>,
+        pub is_forwarding: bool,
+        pub namespace: Option<SignRequestNamespace>,
```
quexten:
```suggestion
        // Namespace of the SSHSig request. This value is none, for regular authentication to a server
        pub namespace: Option<SignRequestNamespace>,
```

apps/desktop/desktop_native/ssh_agent/src/server/auth_policy.rs
```diff
@@ -0,0 +1,56 @@
...
+#[derive(Debug, Clone)]
+pub struct SignRequest {
+    /// The public key identifying which key to use for signing
+    pub public_key: PublicKey,
+    /// Name of the process making the request. If the agent is running in sandboxed environments,
+    /// it may not have access to the process name.
+    pub process_name: Option<String>,
+    /// Whether this is an agent forwarding request
+    pub is_forwarding: bool,
+    /// If available, the parsed representation of the sign request's namespace.
```
quexten: SSHSig request

apps/desktop/desktop_native/ssh_agent/src/authorization.rs
```diff
@@ -0,0 +1,485 @@
...
+                if !self.is_locked.load(Ordering::Relaxed) {
+                    debug!("Already unlocked, allowing list request.");
+                    return Ok(true);
+                }
+
+                // otherwise request unlock from approval handler
+                debug!("Locked, requesting unlock approval.");
+                let unlock_approved =
+                    self.approval_handler
+                        .request_unlock()
```
quexten: Question: Do we assume timeouts are handled by the approvalhandler? Or should this handle a timeout?

apps/desktop/desktop_native/napi/src/sshagent_v2.rs
neuronull: Ahh. I see. What you've pointed out to me is something important. In this v2 design we need to be explicit about where we coalesce the raw server/protocol definitions, into the definitions for our agent/the approver. For example, in our case of the v1 implementation, the approver doesn't seem to distinguish between "sign" requests vs "sig" requests (thus the protocol knowledge doesn't really need to be upheld in the napi layer).

Right now, the types defined in auth_policy are in the format for the agent/approver. I think as we implement the parsing and handling of the server requests, we might do a mild shifting around of where the coalescence happens.

For now, I'll re-name it as you described but keep it where it is until we use it in the handling of server requests.

apps/desktop/desktop_native/napi/src/sshagent_v2.rs
neuronull: I am hoping that after some conclusions on other in-flight work, we won't have to deal with this explicit re-duplication of the type.

For now, I was hoping to not re-duplicate the documentation of the types (since they are 1-1). So I'll make sure the comment at the source type is correct.

apps/desktop/desktop_native/ssh_agent/src/authorization.rs
neuronull: Great callout! IMO they should be handled by the ApprovalHandler. This way, different kinds of approvers can decide if they need timeouts or not (in our existing case, we can utilize them).

I'd meant to circle back and add this but forgot to so thanks for calling it out. I'll add that.

apps/desktop/desktop_native/napi/src/sshagent_v2.rs
quexten: The approver does distinguish in the sense that the namespace is present vs not present, even in v1. But I agree with your framing. We need to be clear about where we translate the low level sign operation into the high level sshsig / authenticate requests / where the enrichment happens. Maybe these should even be two separate messages (later on) entirely with one having the namespace always present, and the other not even having it in the type?

apps/desktop/desktop_native/napi/src/sshagent_v2.rs
```diff
@@ -0,0 +1,222 @@
...
+
+            let is_approved = timeout(APPROVAL_CALLBACK_TIMEOUT, async {
+                self.unlock_callback
+                    .call_async(Ok(()))
+                    .await
+                    .context("Electron unlock callback failed")
+            })
+            .await
+            .context("Electron unlock callback timed out")
+            .flatten()?;
```
quexten, line 126: Maybe I'm missing something, but this seems to return error instead of setting is_approved to false when timing out? Is this intended? I'd expect it to just return false / map the error to false.

apps/desktop/desktop_native/napi/src/sshagent_v2.rs
neuronull: I definitely prefer having different messages to having one message with bools/options that correlate to different behaviors. I think this might imply it would alter the receiving (electron) end, so if you're cool with it I'd suggest we take a look at that in upcoming tickets where we have things wired up end to end.

apps/desktop/desktop_native/napi/src/sshagent_v2.rs
neuronull: I'd intended to keep the true/false to denote explicitly the response being approved/denied at this layer. Currently I think the agent will (correct me if I'm wrong) consider the timeout and failure and denied as the same. However, having the distinguishing of them enables us to do things like retries or map these error states to agent protocol heuristics. Essentially, if the user explicitly denied the request, it's a different scenario than if the request timed out or there was a transport error.

Thus, this comment made me realize that we need to better distinguish the modes. there are essentially 4 states: approved, denied, failed, and timeout.

I'm adding a separate Error type for the approval requester. That is also cascading into some minor improvements in the error propagation so that's also a plus.

quexten: Thought: Why not just return an enum that has the 4 states?

```rust
pub enum ActionResponse {
   Approved,
   Denied,
   Timeout,
   Error(...)
}
```

? If you don't think it's worth it happy to approve either way, but I thought I'd propose it.

neuronull: All ideas are welcome! I appreciate you hashing on it with me.

My opinion is that the current model is ideal because:

- even though there are 4 states, the types represent the semantics clearly (approved/denied is a good useage of `bool` (to put it another way- we generally wouldn't want to define an enum to represent Approved,Denied alone, because that's perfectly represented by the true/false nature of a bool already). and Err as part of `Result` is a clean separation of the error conditions from the cases where the approve/deny answer was not available). Combining them into one enum obscures the distinction.

- Ok/Err leverage Result and folding it into an enum removes the ability to use that, which means that in practice we can no longer utilize the `?` operator  (using the enum to combine these removes us from benefiting from the language's error handling mechanics), which cascades into having to handle those with match statements explicitly.

quexten: Okay, closing in this case. Thank you for the thorough response!

RichardFevrier:
Hello everyone 👋
Micro question before considering opening an issue.

Does the app `name` that was requesting a confirmation request to the agent was populated on the v1 on your side?
Because I can only get `Unknown application` and I don't know if it's because of the sandbox (flatpak) or the v1 implementation.

Like in:
```
[NAPI] [INFO] desktop_core::ssh_agent: Confirming request from application: Unknown application is_forwarding=false namespace=None host_key=xxx
```
Thank you for reading me.

neuronull:
> Does the app `name` that was requesting a confirmation request to the agent was populated on the v1 on your side? Because I can only get `Unknown application` and I don't know if it's because of the sandbox (flatpak) or the v1 implementation.

Hello! 

@quexten can correct me if I'm wrong, but I suspect that is because of the sandbox. 

There isn't thus far a plan to make specific changes as part of the v2 effort, around how we obtain the peer info.

apps/desktop/desktop_native/napi/src/sshagent_v2.rs
neuronull: I added a link to this GH comment thread in the two upcoming tickets that will get more into the details of end to end.

RichardFevrier:
Thanks for the answer @neuronull, I'll wait for quexten to answer too.
Forgot to specify that I also seen people in the issues having the same problem on macOS (.dmg or from the AppStore I don't know).
That's why I did prefer to ask more details in the first place.

quexten:
Yes, the application name is gathered by:
- Checking the unix socket for the PID of the process making the request
- Queriyng the PID information

While the first step works in flatpak, the second fails, because flatpak runs in a separate  PID namespace. The only way to circumvent this would be for us to give the desktop app permissions to spawn processes outside of the flatpak sandbox using e.g. https://man7.org/linux/man-pages/man1/flatpak-spawn.1.html, which would effectively make the app un-sandboxed.

MacOS app store vs dmg is slightly different, but functionally the same. The PID inforation lookup cannot be done since the app is sandboxed and does not have permission to get this information.

RichardFevrier:
Alright thank you too @quexten for the answer.

PR #18885, opened 2026-02-10
[PM-31700] Desktop Vault V3 Unarchive and Save Button

Jingo88:
## 🎟️ Tracking

[PM-31700](https://bitwarden.atlassian.net/browse/PM-31700)

## 📔 Objective

The vault v3 has been updated to include the latest archive changes
The core changes that address the ticket for the `unarchive and save` button take place in the `submitButtonText`
Other changes implemented to switch `this.cipher` to be a signal.

## 📸 Screen Recording

https://github.com/user-attachments/assets/5a4e9a8f-f02c-4a8c-ae62-3953b6554412

[PM-31700]: https://bitwarden.atlassian.net/browse/PM-31700?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/desktop/src/vault/app/vault-v3/vault.component.ts
+++ b/apps/desktop/src/vault/app/vault-v3/vault.component.ts
@@ -11,6 +11,7 @@ import {
   signal,
   ViewChild,
 } from "@angular/core";
+import { toSignal } from "@angular/core/rxjs-interop";
 import { ActivatedRoute, Router } from "@angular/router";
 import {
   combineLatest,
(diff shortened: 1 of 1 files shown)


apps/desktop/src/vault/app/vault-v3/vault.component.ts
```diff
@@ -247,6 +266,7 @@ export class VaultComponent implements OnInit, OnDestroy, CopyClickListener {
       )
       .subscribe((canAccessPremium: boolean) => {
         this.userHasPremiumAccess = canAccessPremium;
+        this.userHasPremium.set(canAccessPremium);
```
shane-melton: 🎨 Can we consolidate both of these properties into one signal instead of having both? It looks like the existing property is only used in a small number of spots and can be updated to the signal.

apps/desktop/src/vault/app/vault-v3/vault.component.ts
shane-melton: 🎨 Similarly, we can use `toSignal()` instead of subscribing:

```ts
private userHasPremium = toSignal(
    filter((account): account is Account => !!account),
    switchMap((account) =>
      this.billingAccountProfileStateService.hasPremiumFromAnySource$(account.id)
    ),
  ));

PR #18886, opened 2026-02-10
[PM-26901] Add notification handler for auto confirm

BTreston:
## 🎟️ Tracking
https://bitwarden.atlassian.net/browse/PM-26379

## 📔 Objective
Adds server push notification and command for auto confirm. Calls the command when the feature flag is enabled to dispatch the push notification to the client.

## 📸 Screenshots
DESIGN REVIEW:
Given:
- the organization ability is enabled
- the policy is enabled
- the client setting in the extension is enabled
- an organization admin is actively logged in 

Users that accept their organization invite will be automatically confirmed

https://github.com/user-attachments/assets/cb333334-e793-4c7f-bcc6-b3ead0a435cf

### DIFF
--- a/libs/auto-confirm/src/services/default-auto-confirm.service.spec.ts
+++ b/libs/auto-confirm/src/services/default-auto-confirm.service.spec.ts
@@ -397,7 +397,7 @@ describe("DefaultAutomaticUserConfirmationService", () => {
       } as UserKeyResponse);
       jest.spyOn(Utils, "fromB64ToArray").mockReturnValue(mockPublicKeyArray);
       organizationUserService.buildConfirmRequest.mockReturnValue(of(mockConfirmRequest));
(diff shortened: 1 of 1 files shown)


libs/auto-confirm/src/index.ts
BTreston, line 1: Since this is a shared nx lib, we cannot import angular code into the other clients (browser specifically for auto-confirm). Therefore, I've split this into an angular specific sub-module and a client agnostic sub-module. @addisonbeck Curious of your thoughts on this. I think we have done this before from what I can tell.

nick-livefront (approved):
Only import changes for vault ✅

libs/auto-confirm/src/index.ts
addisonbeck: I figured we'd just have two libs (auto-confirm & auto-confirm-angular) where one depends on the other, but this is ultimately up to teams. I don't write much UI and wouldn't give the wisest advise :)

PR #18887, opened 2026-02-10
[PM-29521]Remove @ts-strict-ignore in fido2.background.ts

dan-livefront:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-29521
## 📔 Objective

Removed @ts-strict-ignore from fido2.background.ts and made the file typesafe.

Added explicit defaults for class properties, tightened up null/undefined checks such as tab.id, windowId, requestId, and centralized validation in handleExtensionMessage so non null assertions are safely guarded.

Also fixed Set iteration and improved the URL check to explicitly allow https:// and `http://localhost/` since we matched on it.

### DIFF
--- a/apps/browser/src/autofill/background/notification.background.spec.ts
+++ b/apps/browser/src/autofill/background/notification.background.spec.ts
@@ -253,10 +253,16 @@ describe("NotificationBackground", () => {
       });
 
       it("triggers a retryHandler if the message target is `notification.background` and a handler exists", async () => {
+        const retrySender = mock<chrome.runtime.MessageSender>({
+          tab: { id: 1 } as chrome.tabs.Tab,
+        });
         const message: NotificationBackgroundExtensionMessage = {
           command: "unlockCompleted",
           data: {
-            commandToRetry: { message: { command: "bgSaveCipher" } },
+            commandToRetry: {
+              message: { command: "bgSaveCipher" },
+              sender: retrySender,
+            },
             target: "notification.background",
           } as LockedVaultPendingNotificationsData,
         };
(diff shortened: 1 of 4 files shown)


apps/browser/src/autofill/fido2/background/fido2.background.ts
```diff
@@ -74,7 +73,11 @@ export class Fido2Background implements Fido2BackgroundInterface {
       .subscribe(([previous, current]) => this.handleEnablePasskeysUpdate(previous, current));
     this.currentAuthStatus$ = this.authService.activeAccountStatus$
       .pipe(startWith(undefined), pairwise())
-      .subscribe(([_previous, current]) => this.handleAuthStatusUpdate(current));
+      .subscribe(([_previous, current]) => {
+        if (current !== undefined) {
```
jprusik: !

apps/browser/src/autofill/fido2/background/fido2.background.ts
```diff
@@ -102,8 +105,8 @@ export class Fido2Background implements Fido2BackgroundInterface {
 
     for (let index = 0; index < tabs.length; index++) {
       const tab = tabs[index];
-
-      if (tab.url?.startsWith("https")) {
+      const url = tab.url ?? "";
+      if (url.startsWith("https://") || url.startsWith("http://localhost/")) {
```
jprusik: Should we drop the trailing slash to include cases where the port is specified (or there is no path)?

```suggestion
      if (url.startsWith("https://") || url.startsWith("http://localhost")) {
```

dan-livefront: Great suggestion!

PR #18890, opened 2026-02-10
[PM-32088] Switch phishing data source to GitHub

AlexRubik:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-32088 - most recent
https://bitwarden.atlassian.net/browse/PM-32026

## 📔 Objective

Swap phishing urls to now use the bw owned url and remove all fallback logic.


Banrion (approved):
📝 No blocking comments. We could remove the domains resource completely as well as the phishing resources type overall since the client will not be owning the location on where to get the file. We can iterate to clean up this file in a separate pull request to determine if custom matchers are still need or if the current one for links is sufficient for both links and domain style lists.

PR #18905, opened 2026-02-11
[No ticket] Disable process reload on desktop during dev builds

quexten:
## 🎟️ Tracking

No ticket

## 📔 Objective

Disable process reload on dev builds. On Linux, process reload current causes dev builds to freeze making it unusable. Further, dev builds don't need process reload, and this will help when developing since the dev tools don't get cleared.


quexten:
@mzieniukbw I don't believe there are any side effects / there should not be any, unless you are testing process-reload specific functionality. For those cases we can either enable it locally or use regular builds.

PR #18906, opened 2026-02-11
[PM-31763] Migrate userkey state to be client-managed SDK-available state

quexten:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-31763
https://github.com/bitwarden/sdk-internal/pull/702

## 📔 Objective

Migrates the user-key state to be client-managed-sdk-available state. The state has to be in the form of a repository. Thus, we use a single key here that is constant.


libs/vault/src/components/add-edit-folder-dialog/add-edit-folder-dialog.component.ts
```diff
@@ -127,7 +127,7 @@ export class AddEditFolderDialogComponent implements AfterViewInit, OnInit {
 
     try {
       const activeUserId = await firstValueFrom(this.activeUserId$);
-      const userKey = await this.keyService.getUserKey(activeUserId!);
+      const userKey = (await this.keyService.getUserKey(activeUserId!))!;
```
coroiu, line 130: thought: the non-null `!` assertions feels... concerning

PR #18907, opened 2026-02-11
[PM-31049] Enable unlock via SDK

quexten:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-31049
https://github.com/bitwarden/sdk-internal/pull/788

## 📔 Objective

Enables SDK-based unlock, gated behind a feature-flag.

Please note, side-effects from setUserKey (store additional keys) are not yet performed with this PR.

### DIFF
--- a/libs/key-management-ui/src/lock/components/master-password-lock/master-password-lock.component.ts
+++ b/libs/key-management-ui/src/lock/components/master-password-lock/master-password-lock.component.ts
@@ -142,8 +142,9 @@ export class MasterPasswordLockComponent implements OnInit, OnDestroy {
           this.logService.error(
             "[MasterPasswordLockComponent] Failed to retrieve user key after master password unlock",
           );
+          throw Error("Failed to retrieve user key");
         }
-        this.successfulUnlock.emit({ userKey: userKey!, masterPassword });
+        this.successfulUnlock.emit({ userKey: userKey, masterPassword });
       } catch (error) {
         this.logService.error(
           "[MasterPasswordLockComponent] Failed to unlock via master password",


libs/common/src/platform/services/sdk/default-sdk.service.ts
```diff
@@ -250,22 +251,48 @@ export class DefaultSdkService implements SdkService {
...
-      kdfParams:
-        kdfParams.kdfType === KdfType.PBKDF2_SHA256
-          ? { pBKDF2: { iterations: kdfParams.iterations } }
-          : {
-              argon2id: {
-                iterations: kdfParams.iterations,
-                memory: kdfParams.memory,
-                parallelism: kdfParams.parallelism,
+    // Initialize the SDK managed database and the client managed repositories.
+    await initializeState(userId, client.platform().state(), this.stateProvider);
```
quexten, line 256: State has to be initialized first, otherwise clientmanagedState doesn't work as an unlock method.

libs/key-management-ui/src/lock/components/lock.component.ts
```diff
@@ -491,38 +501,63 @@ export class LockComponent implements OnInit, OnDestroy {
 
     const MAX_INVALID_PIN_ENTRY_ATTEMPTS = 5;
 
-    try {
-      const userKey = await this.pinService.decryptUserKeyWithPin(pin, this.activeAccount.id);
+    if (await this.configService.getFeatureFlag(FeatureFlag.UnlockViaSDK)) {
+      try {
+        await this.unlockService.unlockWithPin(this.activeAccount.id, pin);
```
quexten, line 506: Note, all unlock methods currently don't handle side effects. Captured here: https://bitwarden.atlassian.net/browse/PM-31119

libs/common/src/platform/services/sdk/default-sdk.service.ts
```diff
@@ -250,22 +251,48 @@ export class DefaultSdkService implements SdkService {
...
+
+    if (await this.configService.getFeatureFlag(FeatureFlag.UnlockViaSDK)) {
```
dani-garcia, line 259: I think rather than duplicating the call and all the parameters to `initialize_user_crypto`, it might be simpler to extract method to a field:
```ts
    let method: InitUserCryptoMethod;
    if (await this.configService.getFeatureFlag(FeatureFlag.UnlockViaSDK)) {
      method = { clientManagedState: {} };
    } else {
      method = { decryptedKey: { decrypted_user_key: userKey.toBase64() } };
    }
    await client.crypto().initialize_user_crypto({ method, ... });
```

libs/common/src/platform/services/sdk/default-sdk.service.ts
dani-garcia: Sounds reasonable, can you move `loadFeatureFlags` so it goes right after state too? There might be a case in the future were we want to use these flags in the SDK initialization code

libs/common/src/platform/services/sdk/default-sdk.service.ts
quexten: I don't think I agree with this, since now removing feature flag support requires logic change instead of just dropping one of the branches, as dropping one of the branches would result in non-ideal code:
```ts
    let method: InitUserCryptoMethod;
    method = { clientManagedState: {} };
    await client.crypto().initialize_user_crypto({ method, ... });
```

With the above approach all we need to do is drop a branch, which seems safer (even though in this case, the risk is low either way, i suppose).

We will remove this as soon as we rolled out the feature anyways, which should not be a significant amount of time...

quexten: Or - let me rephrase - @dani-garcia what is the end state after feature flag removal you'd prefer here?

libs/common/src/platform/services/sdk/default-sdk.service.ts
quexten: Nice callout; I moved feature flag loading now.

libs/common/src/platform/services/sdk/default-sdk.service.ts
dani-garcia: Hmm good point, I agree that for short lived flags it might be easier this way. 

I guess my concern is more around how big the parameters list is, and so duplicating the call makes it worse. But maybe the best solution is just to extract out the `kdfParams` to a function somewhere, which would simplify the code.

quexten: Switched the param to use the SDK config, and moved the mapping to use the function that's implemented on the class. Should be cleaner now, but I'm looking forward to moving state to the SDK.

libs/key-management-ui/src/lock/components/master-password-lock/master-password-lock.component.ts
```diff
@@ -128,22 +134,45 @@ export class MasterPasswordLockComponent implements OnInit, OnDestroy {
...
+    if (await this.configService.getFeatureFlag(FeatureFlag.UnlockViaSDK)) {
+      try {
+        await this.unlockService.unlockWithMasterPassword(activeUserId, masterPassword);
+        const userKey = await firstValueFrom(this.keyService.userKey$(activeUserId));
+        if (!userKey) {
+          this.logService.error(
+            "[MasterPasswordLockComponent] Failed to retrieve user key after master password unlock",
+          );
+        }
+        this.successfulUnlock.emit({ userKey: userKey!, masterPassword });
```
mzieniukbw: You need to throw error or return early (with error toast) on error. We can't just override TS null check, considering we are expecting a user key present in this callback.
```suggestion
        if (!userKey) {
          this.logService.error(
            "[MasterPasswordLockComponent] Failed to retrieve user key after master password unlock",
          );
          throw Error("Failed to retrieve user key");
        }
        this.successfulUnlock.emit({ userKey: userKey, masterPassword });
```

mzieniukbw:
@quexten shouldn't this be https://bitwarden.atlassian.net/browse/PM-31049 ?

quexten:
@mzieniukbw Yeah you are right, it seems the tickets were split out during refinement. I'll change it to the correct story.

PR #18909, opened 2026-02-11
fix: remove broken flatpak install command in validation job

coroiu:
## 📔 Objective

The `flatpak install -y --user flathub` command was ambiguous and never specified a valid ref, causing an interactive prompt that fails in CI. The line is unnecessary since runtime dependencies are auto-resolved from the flathub remote when the Bitwarden bundle is installed.


coroiu:
Fixed by another PR

Issue #17837, opened 2025-12-05
Browser extension completely breaks passkeys/webauthn in browser process (Chromium/Linux)


pdf:
### Steps To Reproduce

1. Go to https://webauthn.io
2. Click on `Register`
3. Scroll down to error message
4. `Error: The operation either timed out or was not allowed.`

This happens for all sites implementing webauthn, both for registration and auth.

Bizarrely, if I open devtools (as a separate window only, not as a docked pane), the browser extension works as expected and intercepts the call, popping the extension UI, and the native UI if the extension flow is cancelled.

If I disable the browser extension, the native webauthn UI is displayed correctly on register/auth.

### Expected Result

Browser extension pops to handle passkeys.

### Actual Result

Extension causes browser to return an error and completely breaks webauthn for the entire browser process.

### Operating System

Linux

### Operating System Version

`Arch Linux hostname 6.17.9-zen1-1-zen #1 ZEN SMP PREEMPT_DYNAMIC Mon, 24 Nov 2025 15:21:16 +0000 x86_64 GNU/Linux`

### Web Browser

Chrome

### Browser Version

142.0.7444.162

### Environment Versions

Version: 2025.11.1

SDK: 'main (8ef7951)'

Server version: 2025.11.1

### Issue Tracking Info

- [x] I understand that work is tracked outside of Github. A PR will be linked to this issue should one be opened to address it, but Bitwarden doesn't use fields like "assigned", "milestone", or "project" to track progress.

abergs:
Hmm. The error should still be around in the console if you open it after the crash. Could you look?

pdf:
@abergs there are no errors in the browser console, this error is returned from the webauthn API I believe.

If I enable developer mode and collect errors on the extension I can get some output, though not certain how enlightening it is, see attached.

[console-1764954266005.log](https://github.com/user-attachments/files/23966216/console-1764954266005.log)

abergs:
That log was enlightening. Do you have any other extensions running? Or multiple instances of bitwarden?

pdf:
Created a fresh browser profile with only Bitwarden for that test.

pdf:
Another odd datapoint - having devtools open only causes the extension to start functioning correctly if devtools is in a separate window, does not have any effect if devtools is docked.

Attached is a log with devtools for the calling page popped to a separate window, where the extension triggers as expected while devtools is open.

EDIT: Replaced log with a slightly cleaner version from a new session, after logging into the extension.

[...]

PR #15825, opened 2025-07-29
[PM-24289] fix popout windows on wayland (browser)

krjan02:
## 🎟️ Tracking

https://github.com/bitwarden/clients/issues/9045
https://github.com/bitwarden/clients/issues/18743

## 📔 Objective

Bitwarden's popup windows in the browser extension currently rely on precise `left` and `top` coordinates derived from the main browser window's position. While effective on X11, these coordinates are consistently reported as `0` by Chrome on Wayland, due to Wayland's design philosophy of abstracting global window positions from client applications.

This meant our popups were always appearing in the top-left corner of the screen on Wayland, instead of their intended relative position.

This commit introduces a check: if `senderWindow.left` and `senderWindow.top` are both `0` (indicating a Wayland-like environment), we explicitly set the new popup's `left` and `top` properties to 0. This allows the Wayland compositor to use its default placement logic for the window, ensuring it opens reliably and in a sensible location, rather than attempting to apply invalid coordinates.

For X11 and other environments where coordinates are valid, the original, precise positioning logic remains unchanged. This ensures consistent and functional popup behavior across different Linux display servers.

## 📸 Screenshots

image

## ⏰ Reminders before review

- Contributor guidelines followed ✅
- All formatters and local linters executed and passed ✅
- Written new unit and / or integration tests where applicable - Not applicable
- Protected functional changes with optionality (feature flags)  - Not applicable
- Used internationalization (i18n) for all UI strings  - Not applicable
- CI builds passed ✅
- Communicated to DevOps any deployment requirements 
- Updated any necessary documentation (Confluence, contributing docs) or informed the documentation team

## 🦮 Reviewer guidelines

- 👍 (`:+1:`) or similar for great changes
- 📝 (`:memo:`) or ℹ️ (`:information_source:`) for notes or general info
- ❓ (`:question:`) for questions
- 🤔 (`:thinking:`) or 💭 (`:thought_balloon:`) for more open inquiry that's not quite a confirmed issue and could potentially benefit from discussion
- 🎨 (`:art:`) for suggestions / improvements
- ❌ (`:x:`) or ⚠️ (`:warning:`) for more significant problems or concerns needing attention
- 🌱 (`:seedling:`) or ♻️ (`:recycle:`) for future improvements or indications of technical debt
- ⛏ (`:pick:`) for minor or nitpick changes

PR #18928, opened 2026-02-11
[PM-32118] Fix webauthn failure on wayland non 0 origin environments

lucemans:
## 🎟️ Tracking

This is an attempt at fixing https://github.com/bitwarden/clients/issues/17837#issuecomment-3875760202

TLDR;
If linux user, and multi-monitor, and (x: 0, y: 0) is not on a display, all webauthn flows break.

## 📔 Objective

On Linux with the Wayland compositor, the X: 0, Y: 0 coordinate could be "off screen" (due to the entire desktop environment essentially being one big "screen"), thus users with a monitor layout where the most top-left pixel is not occupied by a physical screen (non-horizontal layouts for ex), chrome will fail to open the popup window.

This seems to have been easily patched by adding a retry that tries opening the window but without the specific positioning arguments.

Tested this locally on a chromium, ungoogled chromium, and brave build, and can confirm that this is successfully "fixed".
Ideally this is fixed more upstream via chromium or wayland but for now this patch should recover the webauthn functionality for a lot of users.

## 📸 Screenshots

See above linked github issue

## 🤫 Sidenote

I asked claude to fix this, and so please note that refactoring / improvement is still required.
But for now it fixes the issue, so sharing it here.

### DIFF
--- a/apps/browser/src/platform/browser/browser-api.ts
+++ b/apps/browser/src/platform/browser/browser-api.ts
@@ -119,8 +119,28 @@ export class BrowserApi {
   }
 
   static async createWindow(options: chrome.windows.CreateData): Promise<chrome.windows.Window> {
+    const newWindow = await BrowserApi.attemptCreateWindow(options);
+
+    // If window creation failed and we had explicit positioning, retry without it.
+    // This handles cases where the calculated position is off-screen or invalid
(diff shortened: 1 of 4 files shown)


lucemans:
Closing this in favor of #15825

PR #18950, opened 2026-02-12
[PM-29199] fix not clear only 20 emails input

JaredScar:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-29199

## 📔 Objective

Change language for emails input when inviting members based on how many invites the org has left.

## 📸 Screenshots

image


apps/web/src/app/admin-console/organizations/members/components/member-dialog/member-dialog.component.html
```diff
@@ -25,19 +25,24 @@
         <bit-tab [label]="'role' | i18n">
           <ng-container *ngIf="!editMode">
             <p bitTypography="body1">{{ "inviteUserDesc" | i18n }}</p>
-            <bit-form-field *ngIf="{ seats: remainingSeats$ | async } as remaining">
+            <bit-form-field
+              *ngIf="{
+                seats: remainingSeats$ | async,
```
eliykat: I suggest maintaining the name for clarity, as `seats` might refer to the total seats (not the remaining seats).

```suggestion
                remainingSeats: remainingSeats$ | async,
```

apps/web/src/app/admin-console/organizations/members/components/member-dialog/member-dialog.component.ts
```diff
@@ -124,6 +124,7 @@ export class MemberDialogComponent implements OnDestroy {
   showNoMasterPasswordWarning = false;
   isOnSecretsManagerStandalone: boolean;
   remainingSeats$: Observable<number>;
+  emailBatchLimit$: Observable<number>;
```
eliykat, line 135: Can you add some jsdoc here please - in particular, it's unclear what `emailBatchLimit` is. (Is it an arbitrary limit, or specific to the org, etc)

apps/web/src/app/admin-console/organizations/members/components/member-dialog/member-dialog.component.ts
```diff
@@ -293,15 +294,34 @@ export class MemberDialogComponent implements OnDestroy {
       ),
     );
 
+    this.remainingSeats$ = this.organization$.pipe(
+      map((organization) => {
+        if (!this.isEditDialogParams(this.params)) {
+          return organization.seats - this.params.occupiedSeatCount;
+        }
+
+        return organization.seats;
```
eliykat: I know that this is existing code, but why does the remaining seat count change if we're inviting vs. editing? How is this used in view mode?

apps/web/src/locales/en/messages.json
```diff
@@ -3886,6 +3890,15 @@
   "inviteZeroEmailDesc": {
     "message": "You have 0 invites remaining."
   },
+  "inviteEmailsBatchLimitDesc": {
```
eliykat: Suggested i18n string names, just to make the distinction a bit clearer:
- `inviteMultipleEmailsWithSeatLimit`
- `inviteMultipleEmailsNoSeatLimit`

apps/web/src/app/admin-console/organizations/members/components/member-dialog/member-dialog.component.ts
```diff
@@ -293,15 +294,34 @@ export class MemberDialogComponent implements OnDestroy {
...
+      }),
+    );
+
+    this.emailBatchLimit$ = combineLatest([this.organization$, this.remainingSeats$]).pipe(
+      map(([organization, remainingSeats]) => {
+        const standardLimit =
+          organization.productTierType === ProductTierType.TeamsStarter ? 10 : 20;
+        return Math.min(standardLimit, remainingSeats);
+      }),
```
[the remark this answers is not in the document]
eliykat: I agree with Claude's comments - the interaction between plan type, remaining seats and batch limit needs to be rethought. I think it would be helpful to write out the business rules being applied and potentially include them in a jsdoc.

Additionally:
- you've extracted logic out of the validator, but that means the component is more complex, and this is already quite a complex component. I think it's better to have more validation logic in the validator
- there may be opportunities to reuse some logic here between the validators, e.g. repeating the `standardLimit` logic in both is not ideal. We want a single source of truth if possible, especially for these magic numbers

eliykat (changes requested):
Follow-up review from our discussions earlier in the week.

apps/web/src/app/admin-console/organizations/members/components/member-dialog/member-dialog.component.ts
JaredScar: Valid point... I don't think we need to have that change... Let me know what you think of the new implementation

apps/web/src/app/admin-console/organizations/members/components/member-dialog/member-dialog.component.html
```diff
@@ -25,19 +25,32 @@
...
-              <bit-hint *ngIf="remaining.seats === 1">{{
+              <bit-hint *ngIf="!inviteInfo.isDynamicSeatPlan && inviteInfo.remainingSeats === 1">{{
                 "inviteSingleEmailDesc" | i18n
               }}</bit-hint>
 
-              <bit-hint *ngIf="remaining.seats === 0">{{ "inviteZeroEmailDesc" | i18n }}</bit-hint>
+              <bit-hint *ngIf="!inviteInfo.isDynamicSeatPlan && inviteInfo.remainingSeats <= 0">{{
+                "inviteZeroEmailDesc" | i18n
+              }}</bit-hint>
             </bit-form-field>
```
eliykat: Please use the new `@if` templating syntax rather than `*ngIf`.

It also might be worth nesting all of the non-dynamic seat plan options together, e.g.

```
@if (inviteInfo.isDynamicSeatPlan) {
...
@else {
  @if (inviteInfo.remainingSeats > 1) {
  @if (inviteInfo.remainingSeats == 1) {
  @if (inviteInfo.remainingSeats <= 0) {
```

But I would also understand if that causes excessive nesting. Your call. I was also able to understand it as written.

apps/web/src/app/admin-console/organizations/members/components/member-dialog/validators/org-seat-limit-reached.validator.ts
```diff
@@ -3,6 +3,31 @@ import { AbstractControl, ValidationErrors, ValidatorFn } from "@angular/forms";
 import { Organization } from "@bitwarden/common/admin-console/models/domain/organization";
 import { ProductTierType } from "@bitwarden/common/billing/enums";
 
+/**
+ * Returns the maximum number of unique emails an admin may submit in a single invite operation.
+ *
+ * Business rules:
```
eliykat: This is really helpful, thank you! I recommend using the jsdoc `@remarks` decorator here, as the business rules section is like "more info if you want it".


```suggestion
 * @remarks Business rules:
```

apps/web/src/app/admin-console/organizations/members/components/member-dialog/validators/org-seat-limit-reached.validator.ts
```diff
@@ -3,6 +3,31 @@ import { AbstractControl, ValidationErrors, ValidatorFn } from "@angular/forms";
...
+ * - Dynamic-seat plans (Teams, Enterprise, etc.) can auto-purchase seats on demand, so remaining
+ *   seat count is irrelevant. These plans always allow up to 20 emails per batch.
+ * - Fixed-seat plans (Free, Families, TeamsStarter) have a hard seat cap. The limit is clamped to
+ *   how many seats are still available so the admin cannot queue more invites than the plan allows:
+ *   - TeamsStarter caps at 10 (matching its overall 10-seat plan maximum).
+ *   - Free and Families cap at 20.
```
eliykat: Not sure if this is right - a free org has a seat limit of 2 and a families org has a seat limit of 5.

eliykat: This is incorrect - Free orgs are capped at 2 users and Families are capped at 6. However, I'm not sure we need to reference specific plans at all anymore - see next comment.

apps/web/src/app/admin-console/organizations/members/components/member-dialog/validators/org-seat-limit-reached.validator.ts
```diff
@@ -3,6 +3,31 @@ import { AbstractControl, ValidationErrors, ValidatorFn } from "@angular/forms";
...
+ * - For any fixed-seat plan the limit floors at 0 when the org is already oversubscribed
+ *   (occupiedSeatCount > seats). In that state orgSeatLimitReachedValidator will also reject any
+ *   submission, so 0 is the correct signal to surface to the user.
+ */
+export function getEmailBatchLimit(organization: Organization, occupiedSeatCount: number): number {
+  const standardLimit = organization.productTierType === ProductTierType.TeamsStarter ? 10 : 20;
```
eliykat: I think this can be fixed at 20. Double check me, but:
- if it's a dynamic plan, it'll always be 20 (TeamsStarter is not a dynamic plan)
- if it's a fixed plan, it'll be limited to the number of available seats or 20, whichever is lower; and the number of available seats will reflect the limit of fixed plans. e.g. a TeamsStarter organization will have `organization.seats = 10`, so the `remainingSeats` calculation will handle this for you without having to change `standardLimit`.

So:


```suggestion
  const standardLimit = 20;
```

Also: please add a comment explaining why 20 is the limit.

apps/web/src/app/admin-console/organizations/members/components/member-dialog/validators/org-seat-limit-reached.validator.ts
eliykat, line 1: Do we need this validator anymore? As far as I can tell, after the changes in this PR, `input-email-limit.validator.ts` enforces both the 20 batch limit and the available seat limit. I don't think there's anything left for the `org-seat-limit-reached.validator.ts` to do. But please double check me.

apps/web/src/app/admin-console/organizations/members/components/member-dialog/validators/org-seat-limit-reached.validator.ts
```diff
@@ -3,6 +3,31 @@ import { AbstractControl, ValidationErrors, ValidatorFn } from "@angular/forms";
...
+ *   seat count is irrelevant. These plans always allow up to 20 emails per batch.
+ * - Fixed-seat plans (Free, Families, TeamsStarter) have a hard seat cap. The limit is clamped to
+ *   how many seats are still available so the admin cannot queue more invites than the plan allows:
+ *   - TeamsStarter caps at 10 (matching its overall 10-seat plan maximum).
+ *   - Free and Families cap at 20.
+ * - For any fixed-seat plan the limit floors at 0 when the org is already oversubscribed
+ *   (occupiedSeatCount > seats). In that state orgSeatLimitReachedValidator will also reject any
+ *   submission, so 0 is the correct signal to surface to the user.
+ */
+export function getEmailBatchLimit(organization: Organization, occupiedSeatCount: number): number {
```
eliykat: This belongs in the `input-email-limit.validator.ts` file that uses it. It is unreferenced in this file.

eliykat (changes requested):
Good work @JaredScar

amorask-bitwarden (dismissed):

apps/web/src/app/admin-console/organizations/members/components/member-dialog/validators/input-email-limit.validator.ts
```diff
@@ -3,35 +3,83 @@ import { AbstractControl, ValidationErrors, ValidatorFn } from "@angular/forms";
...
+ *   No per-plan overrides are needed here — the remaining-seats calculation handles every fixed
+ *   plan automatically (e.g. a TeamsStarter org already has `organization.seats = 10`).
+ * - For any fixed-seat plan the limit floors at 0 when the org is already oversubscribed
+ *   (occupiedSeatCount > seats). `inputEmailLimitValidator` uses this value directly; passing
+ *   `existingEmails` ensures that re-inviting an already-accepted member is still permitted
+ *   (no new seat is consumed) even when the org is at full capacity.
+ */
+export function getEmailBatchLimit(organization: Organization, occupiedSeatCount: number): number {
+  // 20 is the maximum number of email addresses the invite input accepts in a single submission.
+  const batchLimit = 20;
```
eliykat: This is stating "what" but not "why", which is more helpful for a comment.

To be honest: there is no "why". 20 emails is just an arbitrary limit! But it's still worth stating that so that future developers aren't trying to find a reason.

apps/web/src/app/admin-console/organizations/members/components/member-dialog/validators/input-email-limit.validator.ts
```diff
@@ -3,35 +3,83 @@ import { AbstractControl, ValidationErrors, ValidatorFn } from "@angular/forms";
...
     .split(",")
     .filter((email: string) => email && email.trim() !== "");
   const uniqueEmails: string[] = Array.from(new Set(emails));
 
-  return uniqueEmails;
+  if (existingEmails.length === 0) {
+    return uniqueEmails;
+  }
+
+  return uniqueEmails.filter((email) => !existingEmails.includes(email.trim()));
```
eliykat: Suggestions: 
- you're trimming the input emails twice - do it once upfront only
- change `existingEmails` to a `Set` for n(1) lookup.

apps/web/src/app/admin-console/organizations/members/components/member-dialog/validators/input-email-limit.validator.spec.ts
```diff
@@ -32,21 +37,14 @@ describe("inputEmailLimitValidator", () => {
...
-  describe("TeamsStarter limit validation", () => {
-    let teamsStarterOrganization: Organization;
-
-    beforeEach(() => {
-      teamsStarterOrganization = orgFactory({
-        productTierType: ProductTierType.TeamsStarter,
-        seats: 10,
-      });
-    });
+  describe("10 email limit validation", () => {
```
eliykat: No longer a valid test case because we removed the fixed 10 email limit logic.

apps/web/src/app/admin-console/organizations/members/components/member-dialog/validators/input-email-limit.validator.spec.ts
```diff
@@ -189,3 +218,92 @@ describe("inputEmailLimitValidator", () => {
...
+        const organization = orgFactory({ productTierType: plan, seats: 100 });
+
+        expect(getEmailBatchLimit(organization, 0)).toBe(20);
+        expect(getEmailBatchLimit(organization, 99)).toBe(20);
+        expect(getEmailBatchLimit(organization, 150)).toBe(20);
+      },
+    );
+  });
+
+  describe("TeamsStarter (fixed, 10-seat cap)", () => {
```
eliykat: Also still referring to the old 10-seat cap logic.

I suggest re-thinking how these test cases are structured - what are the essential behaviours of this sut that we're trying to verify?

Suggestion:
- dynamic seat plan -> returns 20 regardless of seat count
- fixed seat plan ->
  - returns 20 if available seats are greater than 20
  - returns available seats if they are less than 20
  - correctly calculates available seats
  - returns 0 if oversubscribed

I think your tests touch on most or all of this functionality, but they're still organized around specific plan types which causes some duplication and makes it difficult to understand the functionality being tested (and you already test the fixed/dynamic plan functions separately).

eliykat (changes requested):
Some final minor comments! Thanks for iterating on this. I know it's a relatively minor PR, but we've returned to this code a few times with different regressions, so I'm trying to nail down the exact logic and structure of these validators. I think it's come out well though.

eliykat (dismissed):

eliykat:
Thanks! Please note this will need QA before merge.

eliykat (dismissed):

amorask-bitwarden (dismissed):

apps/web/src/app/admin-console/organizations/members/components/member-dialog/validators/org-seat-limit-reached.validator.ts
eliykat, line 1: Why has this been reintroduced?

nick-livefront:
There are no longer vault changes here, removing us as reviewers

JaredScar:
Further clarifying why QA brought this back originally.

It seemed to always be showing 20 when QA was testing it, rather than the number of seats the organization had left. After discussing with Yashoda and Sukhleen they had explained they wanted it to only show the number 20 if they have more than 20 seats remaining. If they had less than 20 seats remaining, they wanted it to show the number of seats remaining rather than the number 20. Further discussion is happening with @eliykat via slack regarding dynamic seat plans and how we want to target this PR to handle that case as well

eliykat:
As discussed, this needs further discussion with QA so I'm removing my review request. Please request my review again when it's ready to look at.

djsmith85:
There are no longer platform changes here, removing us as reviewers

eliykat (approved):

eliykat:
Effectively reverted to the previous approved commit [d2905e3](https://github.com/bitwarden/clients/pull/18950/commits/d2905e30a9c772e3bef83f7db7c8a7feb6ae9d0c) with merge conflicts resolved 👍

amorask-bitwarden (approved):

PR #18954, opened 2026-02-12
[PM-32028] Remove Save and Cancel Buttons

sbrown-livefront:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-32028

## 📔 Objective

Corrects the behavior of displaying the `Save` and `Cancel` buttons when changing the payment method for upgrading.

## 📸 Screenshots

https://github.com/user-attachments/assets/a27c6c4b-8f3b-46dd-a70c-dea3679d7238


apps/web/src/app/billing/individual/upgrade/premium-org-upgrade-payment/premium-org-upgrade-payment.component.ts
```diff
@@ -312,11 +294,21 @@ export class PremiumOrgUpgradePaymentComponent implements OnInit, AfterViewInit
...
-
     if (!billingAddress.country || !billingAddress.postalCode) {
       throw new Error("Billing address is incomplete");
     }
 
+    const paymentMethodComponent = this.paymentMethodComponent();
+    // If the user is changing their payment method, process that first
+    if (paymentMethodComponent && paymentMethodComponent.isChangingPayment()) {
+      const newPaymentMethod = await paymentMethodComponent.getTokenizedPaymentMethod();
+      await this.subscriberBillingClient.updatePaymentMethod(
```
cyprain-okeke, line 304: The updatePaymentMethod call could fail, but there's no specific error handling. The error would propagate up and show a
  generic "upgradeErrorMessage", which might be confusing since the upgrade itself didn't fail—the payment method update did.

apps/web/src/app/billing/individual/upgrade/premium-org-upgrade-payment/premium-org-upgrade-payment.component.ts
```diff
@@ -440,11 +432,28 @@ export class PremiumOrgUpgradePaymentComponent implements OnInit, AfterViewInit
...
+  /**
+   * Checks if the form is valid.
+   */
+  protected isFormValid(): boolean {
+    const isParentFormValid =
+      this.formGroup.controls.organizationName.valid &&
+      this.formGroup.controls.billingAddress.valid;
+
+    const paymentMethodComponent = this.paymentMethodComponent();
+    const isChangingPayment = paymentMethodComponent?.isChangingPayment();
```
cyprain-okeke, line 446: When the user has NO payment method on file, they must add one to upgrade. However, this validation only checks the payment method form when isChangingPayment() is true. What if a user has never added a payment method? They would need to add one, but the validation might not enforce this.

apps/web/src/app/billing/payment/components/display-payment-method-inline.component.ts
```diff
@@ -120,51 +123,94 @@ import { EnterPaymentMethodComponent } from "./enter-payment-method.component";
 export class DisplayPaymentMethodInlineComponent {
   readonly subscriber = input.required<BitwardenSubscriber>();
   readonly paymentMethod = input.required<MaskedPaymentMethod | null>();
+  readonly updatePaymentInParent = input<boolean>(false);
```
cyprain-okeke: Unused input property

apps/web/src/app/billing/individual/upgrade/premium-org-upgrade-payment/premium-org-upgrade-payment.component.ts
sbrown-livefront: The error will be thrown from `updatePaymentMethod` and logged above if that happens in addition to the generic message toarst. Since it is part of the upgrade workflow, what the user would see is a failure, but the logs would reveal the root cause.

apps/web/src/app/billing/individual/upgrade/premium-org-upgrade-payment/premium-org-upgrade-payment.component.ts
sbrown-livefront: `isChangingPayment()` is a property for showing the fields for updating payment in the display-payment component. If a user submits with no payment method, the upgrade will fail, but I'll add a check on submit to make sure the `paymentmethod` signal we set in this component is not null. Great callout.

apps/web/src/app/billing/payment/components/display-payment-method-inline.component.ts
sbrown-livefront: addressed in: https://github.com/bitwarden/clients/pull/18954/commits/d0a73acebc2dbb9bad0589429806059c1a2b9e84

apps/web/src/app/billing/individual/upgrade/premium-org-upgrade-payment/premium-org-upgrade-payment.component.ts
sbrown-livefront: Addressed in https://github.com/bitwarden/clients/pull/18954/commits/123997388a59b7072d7737ffeae0542117475e4d

PR #18957, opened 2026-02-12
[PM-30825] Bugfix: Upload multiple attachments in a row to org cipher

nikwithak:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-30825

## 📔 Objective

After uploading an attachment to a cipher belonging to an organization, the cipher is not properly upserted into the cache, causing future edits to fail due to a `revisionDate` mismatch. The reason for not upserting is that we call the Admin API, which does not return a full cipher object.

This PR fixes the bug by changing the following:
1. Does not call the `admin` version of the API for all org items - instead, only pass the admin flag when using the Admin Console to edit.
2. When the `admin` flag is set, retrieve the cipher from the API instead of from state (bypassing the `upsert` failure)
3. Manually override the `cipher.edit` flag when editing a cipher's attachments from the Admin Console, since admins should be able to edit the ciphers.

## 📸 Screenshots

https://github.com/user-attachments/assets/b3b23c11-ce96-4263-9cca-507c50e71266


apps/browser/src/vault/popup/services/vault-popup-items.service.ts
```diff
@@ -119,7 +119,7 @@ export class VaultPopupItemsService {
               this.cipherService
                 .cipherListViews$(userId)
                 .pipe(filter((ciphers) => ciphers != null)),
-              this.cipherService.failedToDecryptCiphers$(userId),
+              this.cipherService.failedToDecryptCiphers$(userId).pipe(startWith([])),
```
shane-melton, line 121: ❓ This seems unrelated, is it fixing something else you noticed?

nikwithak: Whoops - this was a commit leftover when I switched branches from a different bugfix. This commit was part of #18895  - which has already merged.

nikwithak: GH was able to resolve this with a ff merge, so it's gone from the diff now!

PR #18958, opened 2026-02-12
[PM-29833] The "Change at-risk password" link within the web app doesn't always work.

jengstrom-bw:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-29833

## 📔 Objective

Clicking the link should open the website for the cipher. If there is no http(s) URIs, don't make it a link and remove the link icon.

## 📸 Screenshots

Example with no http(s) link:
Screenshot 2026-02-18 at 3 06 19 PM


libs/vault/src/cipher-view/cipher-view.component.ts
```diff
@@ -267,6 +267,12 @@ export class CipherViewComponent {
     return this.hasLoginUri() && (this.hadPendingChangePasswordTask() || this.passwordIsAtRisk());
   });
 
+  protected readonly changePasswordUrl = resource({
+    params: () => ({ cipher: this.cipher() }),
+    loader: async ({ params }) =>
+      await this.changeLoginPasswordService.getChangePasswordUrl(params.cipher),
```
nick-livefront: ❓ This is going to increase the number of calls to the backend by quite a bit:
- As implemented, this will call the backend each time a cipher is viewed regardless of it it is at-risk or not. Can this be dependent on `showChangePasswordLink`?
- By nature of the change, it will still increase the amount of calls to the backend by calling when the cipher is viewed rather than when the link is clicked. I don't know what the current usage but it could be worth noting? 

cc: @shane-melton

nick-livefront: > I don't know what the current usage but it could be worth noting?

I wasn't very specific here, maybe just looping in @Jingo88 is a good starting point

jengstrom-bw: I think we need to get the URL before the link is clicked to fix the issue, since the cause is people right/middle clicking and opening in a new tab when the link just shows `.../#`

nick-livefront: I agree with the solution. `changeLoginPasswordService.getChangePasswordUrl()` is called for every cipher even if the `At risk` or `change password` links are not shown, sending extra network requests to the back end when they're not needed.

jengstrom-bw: Updated when the resource fire the backend call to rely on when a password is `at-risk`

PR #18959, opened 2026-02-12
[PM-32161]  Remove all emails when email list field is cleared and send is saved

harr1424:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-32161

## 📔 Objective

We currently have no way of removing all emails from a send. The desired flow to remove emails from a send is the following:
1. User opens a send that uses email verification
2. User clears the Emails field while editing the send
3. User is forced to change the “Who can view” drop-down to anyone with link or anyone with password options
4. User saves the send

### English translation value for new validation error message
Screenshot 2026-02-12 at 16 28 19

### Validation error placement
Screenshot 2026-02-12 at 16 28 10


harr1424:
@claude please review

PR #18965, opened 2026-02-13
[CL-987] Update Nav Group and Nav Item components

lxiong-livefront:
## 🎟️ Tracking

[JIRA](https://bitwarden.atlassian.net/browse/CL-987?atlOrigin=eyJpIjoiN2Q1Mzg5ZmNiY2RkNGU5ZmE0MDQzYmU0MTU1OWIxYWUiLCJwIjoiaiJ9)

## 📔 Objective

- updated 'Nav Group' and 'Nav Item' components to align with new design
- removed `default` or `tree` variant and set `tree` as the only variant used for all nav items
- added 'Trailing Elements' story to 'Nav Group' for interactive and non-interactive elements

## 📸 Screenshots

Screenshot 2026-02-18 at 6 22 31 PM
Screenshot 2026-02-18 at 6 22 42 PM

## Testing

- [ ] selected (active), hover, and focus states
- [ ] indentation of standalone 'nav-item' and nested 'nav-group' elements
- [ ] trailing element spacing (interactive and non-interactive should align)
- [ ] right arrow for closed 'nav-group' elements, down arrow for open 'nav-group' elements

### DIFF
--- a/libs/components/src/navigation/side-nav.component.html
+++ b/libs/components/src/navigation/side-nav.component.html
@@ -8,7 +8,7 @@
     [style.width.rem]="open ? (sideNavService.width$ | async) : undefined"
     [style]="
       variant() === 'secondary'
-        ? '--color-sidenav-text: var(--color-admin-sidenav-text); --color-sidenav-background: var(--color-admin-sidenav-background); --color-sidenav-active-item: var(--color-admin-sidenav-active-item); --color-sidenav-item-hover: var(--color-admin-sidenav-item-hover); --color-sidenav-active-item-hover: var(--color-admin-sidenav-active-item-hover); --color-sidenav-arrow-hover: var(--color-admin-sidenav-arrow-hover); --color-sidenav-active-arrow-hover: var(--color-admin-sidenav-active-arrow-hover); --color-sidenav-focus: var(--color-admin-sidenav-focus);'
+        ? '--color-sidenav-text: var(--color-sidenav-secondary-text); --color-sidenav-background: var(--color-sidenav-secondary-background); --color-sidenav-active-item: var(--color-sidenav-secondary-active-item); --color-sidenav-item-hover: var(--color-sidenav-secondary-item-hover); --color-sidenav-active-item-hover: var(--color-sidenav-secondary-active-item-hover); --color-sidenav-arrow-hover: var(--color-sidenav-secondary-arrow-hover); --color-sidenav-active-arrow-hover: var(--color-sidenav-secondary-active-arrow-hover); --color-sidenav-focus: var(--color-sidenav-secondary-focus);'
         : ''
     "
     [cdkTrapFocus]="isOverlay"
--- a/libs/components/src/tw-theme.css
+++ b/libs/components/src/tw-theme.css
@@ -320,8 +320,8 @@
   --color-bg-hover: rgba(var(--color-gray-950-rgb), 0.05);
   --color-bg-sidenav-hover: rgba(var(--color-black-rgb), 0.2);
   --color-bg-sidenav-hover-strong: rgba(var(--color-black-rgb), 0.45);
-  --color-bg-admin-sidenav-hover: rgba(var(--color-brand-950-rgb), 0.1);
-  --color-bg-admin-sidenav-hover-strong: rgba(var(--color-brand-950-rgb), 0.25);
+  --color-bg-sidenav-secondary-hover: rgba(var(--color-brand-950-rgb), 0.1);
+  --color-bg-sidenav-secondary-hover-strong: rgba(var(--color-brand-950-rgb), 0.25);
   --color-bg-hover-contrast: rgba(var(--color-white-rgb), 0.05);
   --color-bg-overlay: rgba(var(--color-gray-950-rgb), 0.3);
 
@@ -376,14 +376,14 @@
   --color-sidenav-active-arrow-hover: var(--color-bg-sidenav-hover-strong);
   --color-sidenav-focus: var(--color-white);
 
-  --color-admin-sidenav-background: var(--color-gray-100);
-  --color-admin-sidenav-text: var(--color-gray-900);
-  --color-admin-sidenav-active-item: var(--color-gray-300);
-  --color-admin-sidenav-item-hover: var(--color-bg-admin-sidenav-hover);
-  --color-admin-sidenav-active-item-hover: var(--color-bg-admin-sidenav-hover-strong);
-  --color-admin-sidenav-arrow-hover: var(--color-bg-admin-sidenav-hover);
-  --color-admin-sidenav-active-arrow-hover: var(--color-bg-admin-sidenav-hover-strong);
-  --color-admin-sidenav-focus: var(--color-black);
+  --color-sidenav-secondary-background: var(--color-gray-100);
+  --color-sidenav-secondary-text: var(--color-gray-900);
+  --color-sidenav-secondary-active-item: var(--color-gray-300);
+  --color-sidenav-secondary-item-hover: var(--color-bg-sidenav-secondary-hover);
+  --color-sidenav-secondary-active-item-hover: var(--color-bg-sidenav-secondary-hover-strong);
+  --color-sidenav-secondary-arrow-hover: var(--color-bg-sidenav-secondary-hover);
+  --color-sidenav-secondary-active-arrow-hover: var(--color-bg-sidenav-secondary-hover-strong);
+  --color-sidenav-secondary-focus: var(--color-black);
 }
 
 .theme_light {
@@ -591,14 +591,14 @@
   --color-sidenav-active-arrow-hover: var(--color-bg-sidenav-hover-strong);
   --color-sidenav-focus: var(--color-white);
 
-  --color-admin-sidenav-background: var(--color-gray-800);
-  --color-admin-sidenav-text: var(--color-white);
-  --color-admin-sidenav-active-item: var(--color-gray-600);
-  --color-admin-sidenav-item-hover: var(--color-bg-sidenav-hover);
-  --color-admin-sidenav-active-item-hover: var(--color-bg-sidenav-hover-strong);
-  --color-admin-sidenav-arrow-hover: var(--color-bg-sidenav-hover);
-  --color-admin-sidenav-active-arrow-hover: var(--color-bg-sidenav-hover-strong);
-  --color-admin-sidenav-focus: var(--color-white);
+  --color-sidenav-secondary-background: var(--color-gray-800);
+  --color-sidenav-secondary-text: var(--color-white);
+  --color-sidenav-secondary-active-item: var(--color-gray-600);
+  --color-sidenav-secondary-item-hover: var(--color-bg-sidenav-hover);
+  --color-sidenav-secondary-active-item-hover: var(--color-bg-sidenav-hover-strong);
+  --color-sidenav-secondary-arrow-hover: var(--color-bg-sidenav-hover);
+  --color-sidenav-secondary-active-arrow-hover: var(--color-bg-sidenav-hover-strong);
+  --color-sidenav-secondary-focus: var(--color-white);
 }
 
 @layer components {


libs/components/src/navigation/nav-group.component.html
```diff
@@ -1,27 +1,25 @@
...
     [forceActiveStyles]="forceActiveStyles()"
+    navGroupItem="true"
   >
     <ng-template #button>
       <button
         type="button"
-        class="tw-ms-auto tw-text-fg-sidenav-text"
-        [class]="variantValue === 'tree' && !open() ? 'tw-transform tw-rotate-[90deg]' : ''"
-        [bitIconButton]="toggleButtonIcon()"
+        class="tw-ms-auto tw-text-fg-sidenav-text !tw-p-0 !tw-w-6 !tw-h-6 focus-visible:tw-border-2 focus-visible:tw-border-solid focus-visible:tw-border-border-focus-contrast"
```
BryanCunningham: @lxiong-livefront why do we need the custom width/height/padding styles here? Is the small button size not correct?

libs/components/src/navigation/nav-item.component.html
```diff
@@ -1,28 +1,40 @@
...
       <div class="tw-relative tw-flex tw-items-center tw-h-full">
         @if (open) {
-          <div
-            class="tw-absolute tw-left-[0px] tw-transform tw--translate-x-[calc(100%_+_4px)] [&>*:focus-visible::before]:!tw-ring-border-focus [&>*:hover]:!tw-border-text-alt2 [&>*]:tw-text-alt2"
-          >
-            <ng-content select="[slot=start]"></ng-content>
-          </div>
+          @if (navGroupItem()) {
+            <div
+              class="tw-absolute tw-left-[0px] tw-transform tw--translate-x-[calc(100%_+_4px)] [&>*:focus-visible::before]:!tw-ring-border-focus [&>*:hover]:!tw-border-text-alt2 [&>*]:tw-text-alt2"
```
BryanCunningham: Can we update `[&>*]:tw-text-alt2` to use the new sidenav-text color. Should be `fg-sidenav-text`

libs/components/src/navigation/nav-item.component.html
```diff
@@ -1,28 +1,40 @@
...
+            >
+              <ng-content select="[slot=start]"></ng-content>
+            </div>
+          } @else {
+            <!-- Renders invisible placeholder arrow button for non-group nav items to maintain visual alignment -->
+            <div class="tw-transform tw--translate-x-[calc(100%_+_4px)] tw-invisible">
+              <button
```
BryanCunningham: Do we actually need a button rendered here? Could probably just be a div with the same width. Or, we could adjust the padding on the parent element. Feels a little more lightweight to me personally. What do you think?

libs/components/src/navigation/nav-item.component.html
```diff
@@ -50,7 +61,8 @@
 
         @if (open) {
           <div
-            class="tw-flex tw-items-center tw-pe-1 tw-gap-1 [&>*:focus-visible::before]:!tw-ring-border-focus [&>*:hover]:!tw-border-border-focus [&>*]:tw-text-fg-sidenav-text empty:tw-hidden"
+            class="tw-flex tw-items-center tw-gap-1 tw-pe-3 [&:has(button[biticonbutton])]:tw-pe-1 [&>*:focus-visible::before]:!tw-ring-border-focus [&>*:hover]:!tw-border-border-focus [&>*]:tw-text-fg-sidenav-text empty:tw-hidden"
```
BryanCunningham: We probably don't want the border applied in the hover state do we? Also, should this be the `border-focus-contrast` color?

libs/components/src/navigation/nav-group.component.html
```diff
@@ -1,27 +1,25 @@
...
-@let variantValue = variant();
-
 @if (!hideIfEmpty() || nestedNavComponents().length > 0) {
   <bit-nav-item
     [text]="text()"
     [icon]="icon()"
     [route]="route()"
-    [variant]="variantValue"
-    [treeDepth]="treeDepth()"
+    [treeDepth]="treeDepth() + 1"
```
BryanCunningham: Why do we need the `+ 1` here? Should we just default the model to `1`? Or, probably a better option than changing the default, adjust the logic to work with `0`? Arbitrarily adding 1 here feels strange/magic to me.

libs/components/src/navigation/nav-group.component.html
lxiong-livefront: @BryanCunningham I added these to better align with the padding in the new design. Otherwise, the hover state kind of covers that padding, but that also still exists for any trailing buttons 🤔 I could revert these if you suggest that as best course of action for now, and wait until there is a smaller button variant?

BryanCunningham: I guess what I'm meaning to say is, we shouldn't need to have a one-off styled button for these. We may need to chat with design about why the 'small' button does not work here. Rather than having bespoke styling for this one button

libs/components/src/navigation/nav-item.component.html
lxiong-livefront: I agree, we can make this more lightweight with a div! I did have a thought that if the button itself changes at any point, we'll need to make a note to update this.. what if we have a separate button component specific to nav-items that this can pull in?

libs/components/src/navigation/nav-group.component.html
BryanCunningham: Or, we may need an new button size like `extra-small`. Just so we only have one source of truth for button styles. Happy to chat with you and design about the issue and see what we can come up with

libs/components/src/navigation/nav-item.component.html
BryanCunningham: I alluded to this in another comment but, I don't love the idea of having a bespoke button in here that is inconsistent in design with the others. But, maybe that is what we need. We'll have to chat with design and see why our normal small icon button does not work here. And yes, you raise a good point about if the button changes sizes we'll not know to update this. So, maybe using a button is the best path here.

libs/components/src/navigation/nav-item.component.html
lxiong-livefront: Hmm, the `[&>*:hover]:!tw-border-border-focus` one is pre-existing - weird, I haven't noticed a border on hover states. I can look into this and can probably remove it

libs/components/src/navigation/nav-group.component.html
lxiong-livefront: Initially, setting the model to 1 or adjusting logic was a bit tricky since it shifted the indentation for non-group nav items, but that might work now that we intentionally want them to be indented. I'll look into the logic again

libs/components/src/navigation/nav-item.component.html
BryanCunningham: Oh yeah, didn't realize that was already here... that _seems_ wrong but, just looking at the code it's hard to say if it can be removed or not.

libs/components/src/navigation/nav-group.component.html
BryanCunningham: Awesome. Ideally the logic works with the model defaulted to `0`. Would feel strange for it to be `1` IMHO.

libs/components/src/navigation/nav-group.component.html
lxiong-livefront: Yes, for sure! I'll revert the custom styling for this button. I think having an 'xs' button variant in this case would be great :)

BryanCunningham: It's also possible this design is using a button style that does not exist yet. We're in the process of updating our buttons right now. If that is the case we can just use the small variant and it should work fine when the new button styles merge. We'll have to confirm with design though

libs/components/src/navigation/nav-group.component.html
lxiong-livefront: Was able to resolve this by updating `navItemIndentationPadding` logic, which also works without needing the placeholder arrow :) [a2ede4d](https://github.com/bitwarden/clients/pull/18965/commits/a2ede4d92ae0dc3732aafc140397e8193cbf956b)

libs/components/src/navigation/nav-item.component.html
lxiong-livefront: Removed placeholder altogether (see comment [below](https://github.com/bitwarden/clients/pull/18965#discussion_r2805399545))

libs/components/src/navigation/nav-item.component.html
lxiong-livefront: I removed these text styles here since they're already inheriting from their parent, which is using `fg-sidenav-text` [b926fc3](https://github.com/bitwarden/clients/pull/18965/commits/b926fc3698228187e4dab06b350e6140b00993f3)

libs/components/src/navigation/nav-item.component.html
lxiong-livefront: I think this was meant to be resolved, but feel free to un-resolve if that's the case!

libs/components/src/navigation/nav-group.component.html
lxiong-livefront: Added 'xsmall' size variant in [9e0ae0e](https://github.com/bitwarden/clients/pull/18965/commits/9e0ae0ecd6c1050a68fe9f488b3171ac21fbe7bc)
and updated `nav-group` and `nav-item` to use this size in [b2e97fd](https://github.com/bitwarden/clients/pull/18965/commits/b2e97fd807ade3195a3676ea54916458a976b19c)

libs/components/src/navigation/nav-group.component.html
```diff
@@ -19,27 +16,22 @@
...
       ></button>
     </ng-template>
-    @if (variantValue === "tree") {
-      <ng-container slot="start">
-        <ng-container *ngTemplateOutlet="button"></ng-container>
-      </ng-container>
-    }
+    <ng-container slot="start">
+      <ng-container *ngTemplateOutlet="button"></ng-container>
+    </ng-container>
```
willmartian: Since this now always goes in the same slot, we can remove the usage of `ngTemplateOutlet` and just place the button in the slot directly:

```html
<button slot="start">
```

libs/components/src/navigation/nav-item.component.html
```diff
@@ -21,8 +22,7 @@
           <!-- The `data-fvw` attribute passes focus to `this.focusVisibleWithin` -->
           <!-- The following `class` field should match the button class field below -->
           <a
-            class="tw-size-full tw-px-4 tw-block tw-truncate tw-border-none tw-bg-transparent tw-text-start !tw-text-fg-sidenav-text hover:tw-text-fg-sidenav-text hover:tw-no-underline focus:tw-outline-none [&_i]:tw-leading-[1.5rem]"
-            [class.!tw-ps-0]="variant() === 'tree' || treeDepth() > 0"
+            [class]="containerClasses"
```
willmartian, line 28: What do you think about declaring this in a `@let` variable instead, since it is local to the template?

libs/components/src/tw-theme.css
```diff
@@ -362,12 +363,14 @@
   --color-sidenav-background: var(--color-brand-800);
   --color-sidenav-text: var(--color-white);
   --color-sidenav-active-item: var(--color-brand-900);
-  --color-sidenav-item-hover: var(--color-brand-900);
+  --color-sidenav-item-hover: var(--color-bg-sidenav-hover);
+  --color-sidenav-item-active-hover: var(--color-brand-950);
```
willmartian: Inconsistent ordering:
`--color-sidenav-active-item`
`--color-sidenav-item-active-hover`

libs/components/src/shared/base-button.directive.ts
```diff
@@ -279,7 +280,7 @@ export const getButtonColorStyles = (buttonType: ButtonType): string[] => {
 export class BaseButtonDirective {
   readonly buttonType = input<ButtonType>("secondary");
 
-  readonly size = model<ButtonSize>("default");
+  readonly size = model<ButtonSize | IconButtonSize>("default");
```
willmartian: Hmm, this is not great. A dev might apply a size to `bitButton` that only works with `bitIconButton` or vice versa. Part of the recent button work was unifying the button and icon button styles so that we didn't need bifurcated options. 

1. We will want the variant to work with `bitButton` if we are adding it for icons
2. We need to get this documented in Figma; could you connect with Emily on that?

libs/components/src/navigation/nav-group.component.html
lxiong-livefront: updated in [76a1dab](https://github.com/bitwarden/clients/pull/18965/commits/76a1dab505798bd393b517c45794e91ce06c072e)

libs/components/src/navigation/nav-item.component.html
lxiong-livefront: I like it! Declared classes in template in [bf0bda1](https://github.com/bitwarden/clients/pull/18965/commits/bf0bda1b77e4feda0834a0929ea333e2bc40973b)

libs/components/src/tw-theme.css
lxiong-livefront: I'm assuming you meant about the naming order, but let me know if not! Updated the naming order here in [758cab9](https://github.com/bitwarden/clients/pull/18965/commits/758cab9c52f9a9610ad25c138f9e8033a4b30bca) and [d5c18e2](https://github.com/bitwarden/clients/pull/18965/commits/d5c18e2132bf0e06e44e369066727f2bc0f2b8e8)

libs/components/src/shared/base-button.directive.ts
lxiong-livefront: Updated to let each 'Button' and 'Icon Button' component handle sizes instead of the Button Directive. Since the 'xsmall' size variant of 'Icon Button' has a different `border-radius`, I moved that into each component as well. [97943df](https://github.com/bitwarden/clients/pull/18965/commits/97943dffdcadef25eda16c40110b932ef21612e7)

libs/components/src/navigation/nav-group.component.html
```diff
@@ -16,30 +13,21 @@
...
-    </ng-template>
-    @if (variantValue === "tree") {
-      <ng-container slot="start">
-        <ng-container *ngTemplateOutlet="button"></ng-container>
-      </ng-container>
-    }
+    <button
+      slot="start"
+      type="button"
+      class="tw-ms-auto tw-text-fg-sidenav-text !tw-text-2xl"
```
willmartian: The `!tw-text-2xl` seems out of place. Shouldn't text sizes be handled by the new icon button size?

libs/components/src/navigation/nav-item.component.html
```diff
@@ -53,7 +55,8 @@
 
         @if (open) {
           <div
-            class="tw-flex tw-items-center tw-pe-1 tw-gap-1 [&>*]:tw-text-fg-sidenav-text empty:tw-hidden"
+            class="tw-flex tw-items-center tw-gap-1 tw-pe-3 [&:has(button[biticonbutton])]:tw-pe-1 [&>*]:tw-text-fg-sidenav-text empty:tw-hidden"
+            (click)="$event.stopPropagation()"
```
willmartian, line 59: Why is this stop propagation needed?

libs/components/src/shared/base-button.directive.ts
```diff
@@ -12,42 +12,6 @@ export const focusRing = [
   "focus-visible:tw-z-10",
 ];
 
-export const getIconButtonSizeStyles = (size: ButtonSize): string[] => {
```
willmartian, line 15: Since this is duplicated between both buttons, and should stay in sync, let's leave this here. We can type it as `size: ButtonSize | IconButtonSize`

libs/components/src/navigation/nav-item.component.html
lxiong-livefront: I noticed in desktop, where there is an 'edit folder' button next to folders, clicking on that button also selects the parent 'nav-item' and changes the vault list to that filter, which seemed a little disruptive to me so I added `stopPropagation` here to prevent that. I can remove this though, if you think that's best!

libs/components/src/navigation/nav-group.component.html
lxiong-livefront: removed this in [622bb0d](https://github.com/bitwarden/clients/pull/18965/commits/622bb0dcff1f4cbd30f3c918991d5fdfed7f67a8)

libs/components/src/shared/base-button.directive.ts
willmartian: Talked about this sync. It is not duplicated, and should stay where it is.

libs/components/src/navigation/nav-item.component.ts
```diff
@@ -117,10 +115,17 @@ export class NavItemComponent extends NavBaseComponent {
...
+      ? "tw-z-10 tw-rounded tw-outline-none tw-ring tw-ring-inset tw-ring-border-sidenav-focus"
       : "",
   );
 
+  /* Determine start slot hover style depending on if item is active/selected */
+  protected readonly startSlotHoverClass = computed(() =>
+    this.showActiveStyles
+      ? "[&>*]:hover:!tw-bg-bg-sidenav-active-arrow-hover"
+      : "[&>*]:hover:!tw-bg-bg-sidenav-arrow-hover",
+  );
```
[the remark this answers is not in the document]
lxiong-livefront: [353671c](https://github.com/bitwarden/clients/pull/18965/commits/353671c41d8a4785485cade4a0ac4f692706eb7e)

libs/components/src/navigation/nav-base.component.ts
```diff
@@ -25,12 +25,7 @@ export abstract class NavBaseComponent {
...
   /**
-   * If this item is used within a tree, set `variant` to `"tree"`
-   */
-  readonly variant = input<"default" | "tree">("default");
-
-  /**
-   * Depth level when nested inside of a `'tree'` variant
+   * Depth level
    */
   readonly treeDepth = model(0);
```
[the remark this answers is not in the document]
lxiong-livefront: [a866250](https://github.com/bitwarden/clients/pull/18965/commits/a86625015d89e2252f115cc9054b46a2ce275770)

libs/components/src/navigation/nav-item.component.html
willmartian: Do we need a keyboard event as well? (See Sonar comment)

libs/components/src/navigation/side-nav.component.html
```diff
@@ -8,7 +8,7 @@
     [style.width.rem]="open ? (sideNavService.width$ | async) : undefined"
     [style]="
       variant() === 'secondary'
```
willmartian, line 10: If this is the "secondary" variant of the side nav, should the colors also use that instead of "admin console"? Does Design think we should still call this variant "secondary"?

libs/components/src/tw-theme.css
```diff
@@ -100,13 +100,15 @@
   --color-brand-800: #0d43af;
   --color-brand-900: #0c3276;
   --color-brand-950: #162455;
+  --color-brand-950-rgb: 22, 36, 85;
```
willmartian, line 103: It is a little sad we have to duplicate these

libs/components/src/navigation/side-nav.component.html
lxiong-livefront: Renamed in [d6e168c](https://github.com/bitwarden/clients/pull/18965/commits/d6e168c41172260c4b12df834ecb481007b147e4)

libs/components/src/tw-theme.css
lxiong-livefront: 😢

PR #18967, opened 2026-02-13
[PM-32204] Add per-SSH-key "Use with SSH agent" toggle

discapes:
## 🎟️ Tracking

https://community.bitwarden.com/t/ssh-agent-allow-specifying-limiting-private-key-offers/79954/24

## 📔 Objective

I, like many others, have over 5 ssh keys. This happens when following best practices like rotating ssh keys and using different keys for different targets. However, using them with Bitwarden's ssh agent causes big inconveniences when the key I need to use happens to not be in the first 5. This PR fixes that issue by allowing the user to toggle whether each key is used with the ssh agent, allowing the user to disable rarely used keys and enable often used keys. The default is to use the key with the ssh agent, so users are not inconvenienced by this change.

Note that the code changes here, while being pretty minimal and straightforward, were mostly written by a language model. I'm not rapid firing AI slop merge requests across open source repositories, I made this feature specifically for my own convenience, but I would be happy if I see it benefit others. Even though most of the changes are very small, feel free to close this if it seems low quality or doesn't align with your policies. Regardless, I would appreciate it if this feature were implemented one way or another, and this will at least serve as a proof-of-concept that it can be done without major architectural changes.

This is a client-side change, and it should affect the desktop application only, which I have tested to work. I have not tested the mobile applications or other platforms (though i have asked the language model to critically review changes in cross-platform code for any possibility of things breaking). State is stored locally.

If anyone who is familiar with Bitwarden's build instructions wants to test this fork, it can be built with
```sh
npm ci && npx nx build desktop && cd apps/desktop && npm run pack:dir
```

## 📸 Screenshots

image

image

image

image

image

### DIFF
--- a/apps/desktop/src/autofill/services/desktop-ssh-agent-key-settings.service.ts
+++ b/apps/desktop/src/autofill/services/desktop-ssh-agent-key-settings.service.ts
@@ -14,7 +14,7 @@ export class DesktopSshAgentKeySettingsService extends SshAgentKeySettings {
 
   isKeyEnabledForAgent$(cipherId: string): Observable<boolean> {
     return this.desktopSettingsService.sshAgentEnabledKeys$.pipe(
-      map((keys) => keys[cipherId] === true),
+      map((keys) => keys[cipherId] !== false),
     );
   }
 
--- a/apps/desktop/src/autofill/services/ssh-agent.service.ts
+++ b/apps/desktop/src/autofill/services/ssh-agent.service.ts
@@ -161,7 +161,7 @@ export class SshAgentService implements OnDestroy {
               (cipher) =>
                 cipher.type === CipherType.SshKey &&
                 !cipher.isDeleted &&
-                enabledKeys[cipher.id] === true,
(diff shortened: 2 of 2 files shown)


eduardo-bolognini:
cool

dwbit:
Thanks for your work on this! We're currently working on V2 of the agent implementation at the moment, which means we'll have to put a pause on submissions until that is rolled out. In the meantime, feel free to open a Github Discussions post and we can let you know when that work has been completed so we can continue the collaboration.

neuronull:
👋 @discapes , to expand on the above message, want to also disclose that we do generally like the UX approach taken here and think this is a valuable feature for users.

Thank you for your patience! The V2 agent implementation will have some changes to the electron bits, but not massively. You changes will likely still be applicable with some small conflicts to resolve.

I created a task to track following up on this with you when the V2 is complete. Looking forward to working with you on it 🚀

neuronull:
Circling back to link to note that we have a community forums post up now that addresses the v2 and contributions 🙏 

PR #18975, opened 2026-02-13
[CL-958] Update avatar component to new styles

vleague2:
## 🎟️ Tracking

[CL-958](https://bitwarden.atlassian.net/browse/CL-958)

## 📔 Objective

This PR updates the avatar component to follow the new design system spec, and updates the avatar color picker on the web app. 

Notable changes:
- We are introducing a new default palette of colors for avatars, while still keeping support for custom hexes that are in use today. Avatars using an auto-generated color will now choose from one of the new defaults.
- We are adding support for button avatars. Existing custom button usages are not being migrated in this PR.
  - `<button bit-avatar></button`

## 📸 Screenshots

**Avatar colors**

Screenshot 2026-02-18 at 10 19 44 AM

**Avatar picker**

Before

Screenshot 2026-02-18 at 10 24 55 AM

After
 
Screenshot 2026-02-18 at 10 22 14 AM

[CL-958]: https://bitwarden.atlassian.net/browse/CL-958?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/components/src/avatar/avatar.component.ts
```diff
@@ -1,73 +1,99 @@
...
+import { AriaDisableDirective } from "../a11y";
+import { ariaDisableElement } from "../utils";
 
-const SizeClasses: Record<SizeTypes, string[]> = {
-  xlarge: ["tw-h-24", "tw-w-24", "tw-min-w-24"],
-  large: ["tw-h-16", "tw-w-16", "tw-min-w-16"],
-  default: ["tw-h-10", "tw-w-10", "tw-min-w-10"],
-  small: ["tw-h-7", "tw-w-7", "tw-min-w-7"],
-  xsmall: ["tw-h-6", "tw-w-6", "tw-min-w-6"],
+export type AvatarSizes = "2xlarge" | "xlarge" | "large" | "base" | "small";
```
BryanCunningham: Can we make the types singular?
```suggestion
export type AvatarSize = "2xlarge" | "xlarge" | "large" | "base" | "small";
```

libs/components/src/avatar/avatar.component.ts
```diff
@@ -1,73 +1,99 @@
...
+
+export const AvatarDefaultColors = ["teal", "coral", "brand", "green", "purple"];
+export type AvatarColors = (typeof AvatarDefaultColors)[number];
```
BryanCunningham: Same here
```suggestion
export type AvatarColor = (typeof AvatarDefaultColors)[number];
```

libs/components/src/avatar/avatar.component.ts
```diff
@@ -1,73 +1,99 @@
...
+
+const SizeClasses: Record<AvatarSizes, string[]> = {
```
BryanCunningham: These feel like they should be camel case right? 
```suggestion
const sizeClasses: Record<AvatarSizes, string[]> = {
```

libs/components/src/avatar/avatar.component.ts
```diff
@@ -1,73 +1,99 @@
...
+  small: ["tw-h-6", "tw-w-6", "tw-min-w-6"],
+};
+
+/**
+ * Palette avatar color options. Prefer using these over custom colors. These are chosen for
+ * cohesion with the rest of our Bitwarden color palette and for accessibility color contrast.
+ * We reference color variables defined in tw-theme.css to ensure the avatar color handles light and
+ * dark mode.
+ */
+export const DefaultAvatarColors: Record<AvatarColors, string> = {
```
BryanCunningham: Same comment about camel case here

libs/components/src/avatar/avatar.component.ts
```diff
@@ -1,73 +1,99 @@
...
+  green: "tw-bg-bg-avatar-green",
+  purple: "tw-bg-bg-avatar-purple",
+};
+
+/**
+ * Hover colors for each default avatar color, for use when the avatar is interactive. We reference
+ * color variables defined in tw-theme.css to ensure the avatar color handles light and
+ * dark mode.
+ */
+const DefaultAvatarHoverColors: Record<AvatarColors, string> = {
```
BryanCunningham: Also, the camel case thing 😄

libs/components/src/avatar/avatar.component.ts
```diff
@@ -79,36 +105,135 @@ export class AvatarComponent {
...
+   */
+  protected readonly avatarBackgroundColor = computed(() => {
+    // If using custom color instead of default avatar color, early exit
+    if (this.usingCustomColor()) {
+      return "";
+    }
+
+    if (this.showHoverColor()) {
+      return DefaultAvatarHoverColors[
+        (this.color() as AvatarColors) ?? this.avatarDefaultColorKey()
```
BryanCunningham: This is probably fine but, is there a way we could check that `this.color()` is of type `AvatarColor` so that we don't have to coerce it?

libs/components/src/avatar/avatar.component.ts
```diff
@@ -144,4 +269,31 @@ export class AvatarComponent {
...
+    if (!Utils.isNullOrWhitespace(id)) {
+      magicString = id!.toString();
+    } else {
+      magicString = this.text()?.toUpperCase() ?? "";
+    }
+
+    const colorKeys = Object.keys(DefaultAvatarColors) as AvatarColors[];
+
+    let hash = 0;
+    for (let i = 0; i < magicString.length; i++) {
```
BryanCunningham: Do we like a `for..of` loop better? Just feels nicer to me personally but, I don't have super strong feelings about it
```js
for (const char of magicString) {
  hash = char.charCodeAt(0) + ((hash << 5) - hash);
}
```
Could also probably just do 
```js
magicString.forEach((char) => ....)
```

vleague2: This was pulled directly from the utils file that was being used previously, but I'll play around and see if I can improve it!

libs/components/src/avatar/avatar.component.ts
vleague2: I can do a real typeguard for better safety

BryanCunningham (dismissed):

libs/components/src/avatar/avatar.component.ts
```diff
@@ -1,73 +1,107 @@
...
+
+export const AvatarDefaultColors = ["teal", "coral", "brand", "green", "purple"] as const;
+export type AvatarColor = (typeof AvatarDefaultColors)[number];
+
+const sizeClasses: Record<AvatarSize, string[]> = {
+  "2xlarge": ["tw-h-16", "tw-w-16", "tw-min-w-16"],
+  xlarge: ["tw-h-14", "tw-w-14", "tw-min-w-14"],
+  large: ["tw-h-11", "tw-w-11", "tw-min-w-11"],
+  base: ["tw-h-8", "tw-w-8", "tw-min-w-8"],
+  small: ["tw-h-6", "tw-w-6", "tw-min-w-6"],
```
willmartian: Do we want to move to shortened size names?

`2xl, xl, l, base, sm`

(A general question. If so, we would also do elsewhere.)

apps/web/src/app/auth/settings/account/selectable-avatar.component.ts
willmartian, line 1: Is this component still necessary? Can we recommend deleting it? If not, should bit-avatar have additional functionality that allows us to delete it?

libs/components/src/avatar/avatar.component.ts
vleague2: I think that would probably look better. I don't think this is something we do consistently across the CL (buttons use the expanded names for example), though that doesn't need to be a blocker here

apps/web/src/app/auth/settings/account/selectable-avatar.component.ts
vleague2: It is probably not necessary anymore. I was aiming to keep the amount of changes in this PR as small as possible for areas owned by other teams, but I can make a ticket for auth to delete it

libs/components/src/avatar/avatar.component.ts
vleague2: @BryanCunningham do you have an opinion on this?

BryanCunningham: Design uses `2xl` and `xl`. But, also uses the the abbreviated `lg` and `sm`... so, my vote would be we go with 
`2xl, xl, large, base, small` and have design update on their side. They have been using the longer form in other components so they should follow that convention here probably

BryanCunningham: I don't love that `2xl` and `xl` are the shortened version but, it's way easier that way

vleague2: So you prefer the unshortened overall?

BryanCunningham: Not necessarily. I was just trying to match design, and our other components. I do prefer shortened for the xl and 2xl for sure. I'd be fine with moving to the shortened version everywhere

vleague2: Yeah I kinda want them to be either all shortened or all expanded. 2xl and xl look much better shortened, so I am also good with moving to shortened (obviously we don't have to go do this for everything right away). (Maybe except for `base`, idk what I'd shorten that to and it should never be directly passed anyway if it's the default)

libs/components/src/avatar/avatar.stories.ts
```diff
@@ -32,48 +32,156 @@ export const Default: Story = {
...
+export const Sizes: Story = {
+  render: (args) => {
+    return {
+      props: args,
+      template: `
+        <span class="tw-font-bold">Static</span>
+        <div class="tw-flex tw-gap-4 tw-mb-10">
+          <div class="tw-flex tw-flex-col tw-gap-2 tw-items-center">
+            <span> small </span>
+            <bit-avatar [color]="'brand'" [text]="'Walt Walterson'" [size]="'small'"></bit-avatar>
```
lxiong-livefront: Walt Walterson is an excellent name 👍

willmartian (commented):
(Repeat of what we talked about in Slack)

using CSS variables gets around the Tailwind compile-time limitation:

`"tw-bg-[--bit-avatar-bg]"`

`"hover:[&:is(button,a)]:tw-bg-[hsl(from_var(--bit-avatar-bg)_h_calc(s-10)_calc(l-10))]"`

color mix can also be used instead of hsl, if that is simpler, and you can reference Tailwind config variables with theme(...)

`<div class="bg-[color-mix(in_srgb,theme(colors.blue.500)_60%,white)]">Custom mixed background</div>`

(above code is probably slightly off correct syntax)

libs/components/src/avatar/avatar.component.html
```diff
@@ -0,0 +1,25 @@
...
+    [class]="svgClass()"
+    attr.viewBox="0 0 {{ svgSize }} {{ svgSize }}"
+  >
+    <text
+      text-anchor="middle"
+      y="50%"
+      x="50%"
+      dy="0.35em"
+      pointer-events="auto"
+      class="[&:is(button:is([aria-disabled='true'])_*)]:tw-fill-fg-disabled"
```
willmartian: Unless I am missing something, this will never apply because `&` will never match `button`, only `text`

libs/components/src/avatar/avatar.component.ts
```diff
@@ -79,36 +113,79 @@ export class AvatarComponent {
...
+   *
+   * All return values must be strings that can be parsed as css variables.
+   */
+  protected readonly avatarBgColors = computed(() => {
+    if (this.usingCustomColor()) {
+      return {
+        avatarBgColor: this.color()!,
+        // Drop the custom color's saturation and lightness by 10% when hovering
+        avatarBgColorHover: `hsl(from ${this.color()} h calc(s - 10) calc(l - 10))`,
+      };
```
willmartian: Since the two computed signals rely on much of the same data, perhaps this file would be a little easier to follow if we combined bg colors and text colors:

```ts
protected readonly colors = computed<{
  bg: string;
  bgHover: string;
  text: string;
}>(...);
```

libs/components/src/avatar/avatar.component.ts
```diff
@@ -79,36 +113,79 @@ export class AvatarComponent {
...
-      return this.color()!;
+    if (Utils.isNullOrWhitespace(color)) {
+      return false;
     }
 
-    if (!Utils.isNullOrWhitespace(id)) {
-      return Utils.stringToColor(id!.toString());
-    }
+    return !isAvatarColor(color);
+  });
```
willmartian: It is a little confusing to have and use both `usingCustomColor` and `isAvatarColor`, when they are semantically inverse of each other. I am guessing it is because one is a type guard, and you can't use a signal as a typeguard?

libs/components/src/avatar/avatar.stories.ts
willmartian, line 1: A story that shows the pseudo state styles would be nice

willmartian (dismissed):
Looks good! Non blocking comments

libs/components/src/avatar/avatar.component.html
vleague2: The full `&:is` check is basically "is a child of any button that is aria-disabled" (the `_*` is the sneaky part)

libs/components/src/avatar/avatar.component.ts
vleague2: Yeah :'( I'll see if I can improve the readability

libs/components/src/avatar/avatar.stories.ts
vleague2: Added in [080bb76](https://github.com/bitwarden/clients/pull/18975/commits/080bb765c888efe2cc610426c02872f4205a80b2)

libs/components/src/avatar/avatar.component.ts
vleague2: Ok I think I have simplified this in [080bb76](https://github.com/bitwarden/clients/pull/18975/commits/080bb765c888efe2cc610426c02872f4205a80b2)

libs/components/src/avatar/avatar.component.ts
vleague2: Combined in [080bb76](https://github.com/bitwarden/clients/pull/18975/commits/080bb765c888efe2cc610426c02872f4205a80b2)

willmartian (changes requested):
one BIG problem:

Image

willmartian (approved):

willmartian (dismissed):

Banrion (dismissed):

jrmccannon (dismissed):

rr-bw (dismissed):

emily-livefront:
@vleague2 Reviewed the design and this looks great to me! ✅

vleague2:
This has now passed QA, the only changes made since the last round of approvals is adding hover background colors and a lint fix!

libs/components/src/avatar/avatar.component.ts
```diff
@@ -1,73 +1,108 @@
...
-      </svg>
-    </span>
-  `,
-  imports: [NgClass],
+  selector: "bit-avatar, button[bit-avatar]",
+  templateUrl: "avatar.component.html",
   changeDetection: ChangeDetectionStrategy.OnPush,
+  host: {
+    class:
+      "tw-inline-block tw-leading-[0px] focus-visible:tw-outline-none tw-rounded-full focus-visible:tw-ring-2 focus-visible:tw-ring-offset-1 focus-visible:tw-ring-border-focus !focus-visible:tw-border-[transparent] focus-visible:tw-z-10 tw-group/avatar aria-disabled:tw-cursor-not-allowed [&.tw-test-hover_svg]:tw-bg-[--avatar-bg-hover] [&.tw-test-focus-visible_svg]:tw-bg-[--avatar-bg-hover]",
```
BryanCunningham, line 80: @vleague2 Can we just add this `--avatar-bg-hover` to the tailwind config rather than using the variable directly? Something like `bg-avatar-hover` or whatever

BryanCunningham (dismissed):
Looks good. One minor comment about adding avatar colors to TW config

libs/components/src/avatar/avatar.component.ts
BryanCunningham: Oh, never mind. I see why we need to do this :)

jrmccannon (dismissed):

bitwarden_license/bit-web/src/app/dirt/reports/member-access-report/member-access-report.component.html
```diff
@@ -43,10 +43,8 @@ <h2 bitTypography="h1">{{ "loading" | i18n }}</h2>
       <td bitCell>
         <div class="tw-flex tw-items-center">
           <bit-avatar
-            size="small"
             [text]="row.name ? row.name : row.email"
             [id]="row.userIdFromOrgUser"
-            [color]="row.avatarColor"
```
Banrion, line 49: ❓ Was removing this line intentional? This pulled the avatar color from the user details

vleague2: Hmm no I do not think this was intentional, it looks like it may have been from a merge conflict that I did not resolve correctly. Thank you for catching!!

rr-bw (approved):

Banrion (approved):

jrmccannon (approved):

BryanCunningham (approved):

PR #18983, opened 2026-02-13
Desktop Native: Add missing docs lint for _some_ modules in core

neuronull:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-32232

## 📔 Objective

We state in the contributing docs style guide that we enforce missing_docs everywhere. We don't actually enforce it in desktop_native though. Arguably, it may not make sense for each crate in desktop_native, since isolated crates aren't used externally.
While I personally believe that `pub` should be documented in general, for developer use especially, it's arguable that it isn't always necessary in application code. This is debatable. We might need to consider adjusting the style guide's verbage regardless. The approach I took in this PR was to only add missing_docs for the subset of `core` that seemed appropriate.
If folks would rather see missing_docs apply to the entire desktop_native and add exceptions where needed, that's fine too.

For the present PR, this change adds the missing_docs lint for select crates and adds the respective documentation for where it seems appropriate.

This is a "best effort" documentation strings; I'm happy to adjust the documentation strings themselves to more accurately describe the code 🙏 

### Side quest meanderings

`core` is arguably a mixture of too many things.

In the future we perhaps could consider splitting core 3 modules (or 2, I could see an argument to combine the first two) into something that represents:

- infrastructure to support desktop_native ↔︎ electron

- functionality for the Desktop app itself to operate, which is not feature-specific (either not tied to a feature, or that multiple features can leverage)

- library code that crates within desktop_native leverage (isn't feature specific, and which doesn't otherwise belong in the internal SDK)

, and removing things currently in core/ that don’t fit the above categories. 

IMO the following could be pulled out of core:
  - autofill
  - autostart
  - clipboard 
  - powermonitor
  - sshagent (already a v2 effort in progress which will move this out of core)

With a proper crate organization, there will be less of the "allow" macro being used / the setting of the lint can be set at the appropriate lib.rs


apps/desktop/desktop_native/core/src/biometric/mod.rs
```diff
@@ -13,28 +15,43 @@ use sha2::{Digest, Sha256};
 
 use crate::crypto::{self, CipherString};
 
+/// Cryptographic key material for biometric encryption.
```
quexten: TBH, this entire module is scheduled for deletion, so it's not necessary to document it, but I'm also not going to require changes on the already added documentation.

apps/desktop/desktop_native/core/src/biometric_v2/linux.rs
```diff
@@ -21,13 +21,15 @@ use zbus_polkit::policykit1::{AuthorityProxy, CheckAuthorizationFlags, Subject};
 
 use crate::secure_memory::{encrypted_memory_store::EncryptedMemoryStore, SecureMemoryStore as _};
 
+/// Biometric lock system using Polkit for authentication on Linux.
```
quexten:
```suggestion
/// Biometric lock system using Polkit for authentication and secure memory to hold the key on Linux.
```

apps/desktop/desktop_native/core/src/biometric_v2/mod.rs
```diff
@@ -1,3 +1,5 @@
+//! Biometric authentication with key management (v2 implementation).
```
quexten:
```suggestion
//! Biometric unlock module
//! 
//! This modules can protect a key, either in-memory or in persisted operating system APIs
//! and release it only after authenticating via biometrics.
```

apps/desktop/desktop_native/core/src/biometric_v2/mod.rs
```diff
@@ -11,6 +13,7 @@ pub mod windows_focus;
 
 pub use biometric_v2::BiometricLockSystem;
 
+/// Platform-specific biometric authentication with key management.
```
quexten:
```suggestion
/// Platform-specific biometric-protected key storage
```

apps/desktop/desktop_native/core/src/biometric_v2/windows_focus.rs
```diff
@@ -1,3 +1,5 @@
+//! Windows-specific window focus management for biometric prompts.
```
quexten, line 1:
```suggestion
//! Windows-specific window focus management for biometric prompts.
//!
//! A large part of this is hacks to get around limitations with the Windows-Hello API used,
//! since it does not bring the authentication prompt into focus automatically.
```

apps/desktop/desktop_native/core/src/biometric/mod.rs
neuronull: For the agent v1, I set the ignore_missing docs flag in the lib.rs , and the same can be done for biometric v1. I can change that.

neuronull: done in 

https://github.com/bitwarden/clients/pull/18983/commits/11d007c1cb8313d3d418fab962f0659aec16eda0

and 

https://github.com/bitwarden/clients/pull/18983/commits/3815bcc2c73a637276dcaca89530f7203330b806

neuronull:
It looks like @eligrubb  was tagged automatically but @quexten reviewed for KM already. I'll drop request for @eligrubb  (but feel free to still review if desired)

quexten:
Approving on behalf of KM.

coroiu:
@djsmith85 What should I bring up?

djsmith85:
> @djsmith85 What should I bring up?

@coroiu 
Quote from the PR description
> While I personally believe that pub should be documented in general, for developer use especially, it's arguable that it isn't always necessary in application code. This is debatable. We might need to consider adjusting the style guide's verbage regardless. The approach I took in this PR was to only add missing_docs for the subset of core that seemed appropriate.
> If folks would rather see missing_docs apply to the entire desktop_native and add exceptions where needed, that's fine too.

The PR adds `#[allow(missing_docs)]` to most [pub crates](https://github.com/bitwarden/clients/pull/18983/changes#diff-a4de81fd5a2389d7b512dd37989a42a452fe36cbc1f32d16d5832880355d5669R10) and you could start a conversation with the tech-leads to remove those and start adding documentation to all public crates/methods. Like suggested by @neuronull .

neuronull:
> The PR adds `#[allow(missing_docs)]` to most [pub crates](https://github.com/bitwarden/clients/pull/18983/changes#diff-a4de81fd5a2389d7b512dd37989a42a452fe36cbc1f32d16d5832880355d5669R10) and you could start a conversation with the tech-leads to remove those and start adding documentation to all public crates/methods. Like suggested by @neuronull .

We can discuss but if we want to distribute the ownership of updating the docs to those areas owned by other teams, I can adjust the PR to add `// TODO` statements with links to any tickets we create.

neuronull:
> > The PR adds `#[allow(missing_docs)]` to most [pub crates](https://github.com/bitwarden/clients/pull/18983/changes#diff-a4de81fd5a2389d7b512dd37989a42a452fe36cbc1f32d16d5832880355d5669R10) and you could start a conversation with the tech-leads to remove those and start adding documentation to all public crates/methods. Like suggested by @neuronull .
> 
> We can discuss but if we want to distribute the ownership of updating the docs to those areas owned by other teams, I can adjust the PR to add `// TODO` statements with links to any tickets we create.

We elected not to do add todo or fixme comments

PR #18985, opened 2026-02-13
[PM-31322] Automatically focus accept button in delete Send confirmation dialog

mcamirault:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-31322

## 📔 Objective

This PR makes it so that the "Accept" button of a simple dialog is automatically focused when the dialog opens. 

## 📸 Screenshots

https://github.com/user-attachments/assets/c53bc594-b4dd-41f3-b314-57907ad03dd0

### DIFF
--- a/libs/components/src/dialog/simple-dialog/simple-configurable-dialog/simple-configurable-dialog.component.html
+++ b/libs/components/src/dialog/simple-dialog/simple-configurable-dialog/simple-configurable-dialog.component.html
@@ -6,7 +6,7 @@
   <div bitDialogContent>{{ content }}</div>
 
   <ng-container bitDialogFooter>
-    <button type="submit" bitButton bitFormButton buttonType="primary">
+    <button type="submit" bitButton bitFormButton buttonType="primary" appAutoFocus>
       {{ acceptButtonText }}
     </button>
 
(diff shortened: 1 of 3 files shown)


mcamirault:
@bitwarden/team-ui-foundation This one puzzled me for a bit, but I eventually found [this line in the menu directive](https://github.com/bitwarden/clients/blob/main/libs/components/src/menu/menu-trigger-for.directive.ts#L219) that explained why the dialog button kept losing focus, even after I added the autofocus directive. Because of this I don't think that this change will actually affect any clients that don't currently stop the click event from propagating and the menu from closing when they open a dialog, they'll just lose focus the way they always have. Even so, stopping the event feels like a bit of a workaround so you may want to make a ticket to investigate improving the interaction between these two components at some point.

vleague2:
@mcamirault Ah, this is a miss on my part. I have been working on dialog focus tickets but neglected to make the same changes to the simple dialog. We have some timing-related constraints for "autofocus on open" behavior due to the menu, as you found, and we have an approach in place in the dialog that we should apply to the simple dialog. Would you be ok with me opening up a PR for this in lieu of this one?

mcamirault:
@vleague2 Sounds good, I'll close this PR and link the Jira ticket to yours once it's up. I did see that autofocus logic in the main dialog component - one of the many rabbit holes I went down trying to figure this out 😁 - but wasn't sure if the simple dialog was being kept intentionally simple in that regard.

PR #18987, opened 2026-02-13
[PM-31630] Fix native messaging state management

Thomas-Avery:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-31630

## 📔 Objective

The `NativeMessagingBackground` service had a bug in its internal state management around it's `connecting` state and what happens when disconnecting. This was causing biometric unlock settings and the lock component not to detect when the desktop became available.  

The early guard would return when `connecting` was still true during a failed initial connection. So a new port was never created and connected to when the desktop did become available.
```
if (this.connected || this.connecting) {
      return;
    }
```


apps/browser/src/background/nativeMessaging.background.ts
```diff
@@ -362,6 +355,7 @@ export class NativeMessagingBackground {
 
       this.secureChannel = undefined;
```
quexten: Nit: Do we want to re-use disconnect()?

Thomas-Avery: Yes I think we do. After setting these values to false we have to reconnect so it makes sense to disconnect from the port here as well.

quexten (approved):

PR #18990, opened 2026-02-13
[PM-32127] Access Intelligence: persist selected applications through filter updates

lastbestdev:
## 🎟️ Tracking
https://bitwarden.atlassian.net/browse/PM-32127 (also fixes https://bitwarden.atlassian.net/browse/PM-32415)

## 📔 Objective
This PR includes changes to the Access Intelligence table view, which keep Applications selected in the table as the user makes changes to filters (search bar, critical applications filter). This required updating logic to ensure only visible rows in the table are considered for updates to critical status with the "Mark # as critical" button, while still maintaining the full list of selected applications in the component's `selectedUrls`. 

The Applications table component is also refactored to use Angular `output` for checkbox state, emitting events on checkbox changes for individual table rows and "select all". The parent component handles these events by updating the set of selected Applications (`selectedUrls`) accordingly.

Test cases are updated/added to cover the updated checkbox functionality.

## 📸 Screenshots
https://github.com/user-attachments/assets/fbd6a0e6-15ea-4fe9-860b-1ceb3d8aeae3

### DIFF
--- a/apps/web/src/app/tools/send/send-access/send-access-text.component.html
+++ b/apps/web/src/app/tools/send/send-access/send-access-text.component.html
@@ -10,7 +10,14 @@
   }
 </div>
 <div class="tw-mb-3">
-  <button bitButton type="button" buttonType="primary" [block]="true" (click)="copyText()">
-    <bit-icon name="bwi-clone"></bit-icon>{{ "copyValue" | i18n }}
+  <button
+    bitButton
+    type="button"
+    buttonType="primary"
+    [block]="true"
+    (click)="copyText()"
+    startIcon="bwi-clone"
+  >
+    {{ "copyValue" | i18n }}
   </button>
 </div>


bitwarden_license/bit-web/src/app/dirt/access-intelligence/all-applications/applications.component.ts
```diff
@@ -234,10 +250,11 @@ export class ApplicationsComponent implements OnInit {
 
   markAppsAsCritical = async () => {
```
ttalty: ❓ Should this clear `selectedUrls` after saving the critical applications? If someone were to select 2 applications to mark as critical: amazon, google. Filter to show only google. Mark google as critical. The amazon selection would disappear.

lastbestdev: Good callout! I reviewed the Figma files for this flow and you are correct, the selection should clear after marking critical/non-critical 

Figma link https://www.figma.com/design/BIa5GU9UGxPH6kc9Nj1b64/Access-Intelligence---Admin-Console?node-id=5851-60519&t=51poO0XLstItrmKw-1

lastbestdev: Turns out the component was already resetting the selected applications after submit. See other comments

bitwarden_license/bit-web/src/app/dirt/access-intelligence/all-applications/applications.component.ts
lastbestdev, line 252: @ttalty here

lastbestdev, line 284: Also here @ttalty. Good callout still! 👍🏻

bitwarden_license/bit-web/src/app/dirt/access-intelligence/all-applications/applications.component.ts
```diff
@@ -335,6 +357,21 @@ export class ApplicationsComponent implements OnInit {
     });
   };
 
+  onSelectAllChange = (checked: boolean) => {
```
Banrion: ⛏️ It looks like there are a lot of arrow functions in this file already, but we should prefer regular methods for consistency. We should add this to our standards to discuss arrow functions vs regular methods and binding

lastbestdev: Changed these in latest

PR #18991, opened 2026-02-13
[PM-32242] Error message is incorrectly formatted for password protected Send

harr1424:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-32242

## 📔 Objective

During the Send recipient flow, when a password is entered incorrectly, we display an error toast. This should instead be an inline error message to keep the feedback contextual to the form and provide persistent feedback. 

## 📸 Screenshots

Screenshot 2026-02-13 at 16 19 26

### DIFF
--- a/.github/workflows/release-web.yml
+++ b/.github/workflows/release-web.yml
@@ -97,4 +97,3 @@ jobs:
           artifacts: "apps/web/artifacts/web-${{ needs.setup.outputs.release_version }}-selfhosted-COMMERCIAL.zip,
             apps/web/artifacts/web-${{ needs.setup.outputs.release_version }}-selfhosted-open-source.zip"
           token: ${{ secrets.GITHUB_TOKEN }}
-          draft: true
--- a/apps/browser/package.json
+++ b/apps/browser/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@bitwarden/browser",
-  "version": "2026.1.0",
+  "version": "2026.1.1",
   "scripts": {
     "build": "npm run build:chrome",
     "build:bit": "npm run build:bit:chrome",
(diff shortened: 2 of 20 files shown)


apps/web/src/app/tools/send/send-access/send-auth.component.ts
```diff
@@ -106,6 +106,10 @@ export class SendAuthComponent implements OnInit {
       if (e instanceof ErrorResponse) {
         if (e.statusCode === 401) {
           this.sendAuthType.set(AuthType.Password);
+          this.sendAccessForm.controls.password?.setErrors({
```
mcamirault: ~~The first time you get a 401 back in the V1 flow it shouldn't be an error; that's how the client knows the Send requires a password. These changes will make the error show up before the user has had a chance to enter anything, which could be confusing. We should move this error setting before line 108 and put it in an `if` that checks whether `sendAuthType` is already set to `AuthType.Password`~~ Nope, I was wrong

apps/web/src/app/tools/send/send-access/send-auth.component.ts
```diff
@@ -104,7 +104,27 @@ export class SendAuthComponent implements OnInit {
     } catch (e) {
       if (e instanceof ErrorResponse) {
         if (e.statusCode === 401) {
+          if (this.sendAuthType() === AuthType.Password) {
```
mcamirault, line 107: 🎨 Ah, I believe I had a mistaken understanding of how the V1 Send access flow worked (I thought that sending no password and sending a wrong password would both result in a 401). Given that, I don't think you need to check the `sendAuthType` here and on 119, you can just rely on the statusCodes

harr1424: While not absolutely necessary, this check may avoid future confusion during PR reviews. If you have a strong preference for a comment instead, I think that would be reasonable, but the check is not completely without justification.

mcamirault: No strong preference here, I'm fine with leaving it

PR #18465, opened 2026-01-20
[PM-30470] Revert to using X11 on Linux desktop

dereknance:
## 🎟️ Tracking

[PM-30470](https://bitwarden.atlassian.net/browse/PM-30470)

## 📔 Objective

Electron 39 contains a bug on Linux desktop for Wayland when hardware acceleration is in use that results in a crash during startup. This PR changes the Linux desktop app's default usage of Wayland to X11 to mitigate this crash until the upstream bug is fixed.

## 📸 Screenshots

[PM-30470]: https://bitwarden.atlassian.net/browse/PM-30470?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #19006, opened 2026-02-16
[PM-32342] Fix high CPU usage on flatpak / wayland

quexten:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-32342

## 📔 Objective

While wayland is active, with electron 39 regressions, CPU usage goes to 100-400% even while the window is closed. Forcing X11 fixes this. This still leaves an opt-out via the USE_X11 environment variable.


quexten:
@dani-garcia I did not check though I suspect it's related to https://github.com/bitwarden/clients/pull/18465

PR #19014, opened 2026-02-16
[PM-25820] Update client to handle new DeleteAttachmentResponseModel from server      

gbubemismith:
## 🎟️ Tracking
https://bitwarden.atlassian.net/browse/PM-25820
[server change](https://github.com/bitwarden/server/pull/7013)

## 📔 Objective
Updates the client to handle the new `DeleteAttachmentResponseModel` returned by the delete attachment endpoints


libs/common/src/abstractions/api.service.ts
```diff
@@ -243,8 +244,14 @@ export abstract class ApiService {
     id: string,
     request: AttachmentRequest,
   ): Promise<AttachmentUploadDataResponse>;
-  abstract deleteCipherAttachment(id: string, attachmentId: string): Promise<any>;
-  abstract deleteCipherAttachmentAdmin(id: string, attachmentId: string): Promise<any>;
```
nick-livefront, line 247: 👋 👋 Bye `any`!

PR #19017, opened 2026-02-16
[CL-965] update spinner component

iivins-livefront:
## Overview

The goal of this PR was to modernize the Spinner component to use Angular signals, OnPush change detection, and semantic color variants, replacing the legacy implementation. Added a new SpinnerLockup component that combines a spinner with contextual title and body text using content projection slots.

## Key Changes

- Migrated to signal inputs, OnPush change detection, and `inject()` function following Angular best practices
- Replaced `noColor` boolean with 6 variants
- Changed from `fill/small/large` to explicit `sm/md/base/lg` w/ tailwind classes
- Created new SpinnerLockupComponent supporting `horizontal`/`vertical` layouts with flexible `title` and `body` content slots
- Created unit tests for both components covering render, accessibility, and content projection

## Screenshots

Screenshot 2026-02-16 at 5 30 02 PM
Screenshot 2026-02-16 at 5 30 24 PM
Screenshot 2026-02-16 at 5 31 22 PM
Screenshot 2026-02-16 at 5 31 24 PM
Screenshot 2026-02-16 at 5 31 11 PM
Screenshot 2026-02-16 at 5 31 55 PM


libs/components/src/spinner/spinner.component.ts
```diff
@@ -1,61 +1,57 @@
...
-  imports: [CommonModule],
+  changeDetection: ChangeDetectionStrategy.OnPush,
 })
 export class SpinnerComponent {
-  /**
-   * The size of the spinner. Defaults to `large`.
-   */
-  // FIXME(https://bitwarden.atlassian.net/browse/CL-903): Migrate to Signals
-  // eslint-disable-next-line @angular-eslint/prefer-signals
-  @Input() size: "fill" | "small" | "large" = "large";
```
BryanCunningham, line 19: Did we check to see if any of these variants are in use currently? I'm assuming we'd see it broken in the build

libs/components/src/spinner/spinner.component.ts
```diff
@@ -1,61 +1,57 @@
...
-
-  /**
-   * Disable the default color of the spinner, inherits the text color.
-   */
-  // FIXME(https://bitwarden.atlassian.net/browse/CL-903): Migrate to Signals
-  // eslint-disable-next-line @angular-eslint/prefer-signals
-  @Input({ transform: booleanAttribute }) noColor = false;
```
BryanCunningham, line 26: Same here, any usage of this in the consuming code currently?

iivins-livefront: There were 2 instances of `noColor`:
- libs/components/src/button/button.component.html
- libs/components/src/icon-button/icon-button.component.html

Example:
`<bit-spinner size="fill" noColor></bit-spinner>`

^I updated those to just `<bit-spinner />`

I couldn't find any use cases of `"small" | "large"`

libs/components/src/spinner/spinner.component.ts
iivins-livefront: Just the 2 uses cases - one in Button and one in Icon Button

iivins-livefront:
Updated "contrast" variable colors

Screenshot 2026-02-17 at 5 11 04 PM
Screenshot 2026-02-17 at 5 14 33 PM

willmartian (changes requested):
Spinners used in button loading states are broken. Maybe this is related to the removal of the `fill` size?

https://62a88a6de5b807fa98886113-louzbjjnff.chromatic.com/?path=/docs/component-library-async-actions-standalone--docs

https://github.com/user-attachments/assets/b1e43c8a-f3a9-44a8-970b-ea534fd71061

libs/components/src/spinner-lockup/spinner-lockup.component.spec.ts
willmartian, line 1: These tests are redundant with Storybook + Chromatic and can be removed. My rule of thumb is "avoid testing that Angular is working as expected"

libs/components/src/spinner-lockup/spinner-lockup.component.html
```diff
@@ -0,0 +1,11 @@
+<div [class]="layoutClasses().container">
+  <bit-spinner [size]="size()" [variant]="variant()"></bit-spinner>
+  <div [class]="layoutClasses().textContainer">
+    <span [class]="textClasses().title">
+      <ng-content select="[title]"></ng-content>
```
willmartian: For content projection, we have been recently been trying to follow a pattern of using the "slot" attribute:

definition:
`<ng-content select="[slot=title]"></ng-content>`

usage:
`<span slot="title">`

The benefit is that it is immediately clear to the consumer what this attribute is doing, and that it isn't a directive.

iivins-livefront:
@willmartian removed the duplicate test file - added size adjustments for the loading states - and updated the slots instance

https://github.com/user-attachments/assets/fe1ae338-1bc9-4c4a-8a2b-6085e2e49163

libs/components/src/button/button.component.ts
```diff
@@ -171,6 +171,10 @@ export class ButtonComponent implements ButtonLikeAbstraction {
    * This pattern of converting a signal to an observable and back to a signal is not
    * recommended. TODO -- find better way to use debounce with signals (CL-596)
    */
+  protected readonly spinnerSize = computed(
+    (): SpinnerSize => (this.size() === "small" ? "sm" : "md"),
+  );
```
willmartian: Moving the type to the generic position is more idiomatic: `computed<SpinnerSize>(...)`

libs/components/src/spinner-lockup/spinner-lockup.component.ts
```diff
@@ -0,0 +1,64 @@
...
+  },
+};
+
+@Component({
+  selector: "bit-spinner-lockup",
+  templateUrl: "spinner-lockup.component.html",
+  changeDetection: ChangeDetectionStrategy.OnPush,
+  imports: [SpinnerComponent],
+})
+export class SpinnerLockupComponent {
```
willmartian, line 42: Could you add brief jsdoc that explains what this component does?

libs/components/src/spinner-lockup/spinner-lockup.component.ts
```diff
@@ -0,0 +1,64 @@
...
+  readonly size = input<SpinnerSize>("base");
+  readonly variant = input<SpinnerVariant>("primary");
+  readonly layout = input<SpinnerLockupLayout>("horizontal");
+
+  readonly sizeClasses = computed(() => spinnerSizeStyles[this.size()]);
```
willmartian: The only public fields in components should be inputs/outputs

libs/components/src/spinner-lockup/spinner-lockup.component.ts
```diff
@@ -0,0 +1,64 @@
...
+
+@Component({
+  selector: "bit-spinner-lockup",
+  templateUrl: "spinner-lockup.component.html",
+  changeDetection: ChangeDetectionStrategy.OnPush,
+  imports: [SpinnerComponent],
+})
+export class SpinnerLockupComponent {
+  readonly size = input<SpinnerSize>("base");
+  readonly variant = input<SpinnerVariant>("primary");
```
willmartian, line 47: Could you add JSDoc to all public fields?

libs/components/src/spinner-lockup/spinner-lockup.component.ts
```diff
@@ -0,0 +1,64 @@
...
+  readonly layout = input<SpinnerLockupLayout>("horizontal");
```
willmartian: `layout` is a little abstract, perhaps `orientation`?

libs/components/src/spinner/spinner.component.spec.ts
willmartian, line 1: I don't think these tests are super helpful, could we remove them? I think Storybook covers the "did it render" and the a11y plugin catches other errors.

willmartian (changes requested):

willmartian (changes requested):
FYI, there are merge conflicts from the M1 merge.

willmartian (changes requested):
The spinner in the select view is misaligned. What did we change to cause this? Should we be worried about spinning positioning for consumers elsewhere in the app?

Image

libs/components/src/spinner-lockup/spinner-lockup.component.html
```diff
@@ -0,0 +1,11 @@
+<div [class]="orientationClasses().container">
+  <bit-spinner [size]="size()" [variant]="variant()"></bit-spinner>
+  <div [class]="orientationClasses().textContainer">
+    <span [class]="textClasses().title">
+      <ng-content select="[slot=title]"></ng-content>
+    </span>
+    <span [class]="textClasses().body">
+      <ng-content select="[slot=body]"></ng-content>
```
willmartian: `no-items`, `spinner-lockup`, `progress-lockup`  all have a concept like secondary text, but we are using different names:
- no-items: description
- spinner-lockup: body
- progress-lockup: subtitle

I think it would make sense to unify these. `no-items` already exists in the codebase and is used by many consumers, so I lean towards "description".

willmartian (changes requested):

libs/components/src/spinner-lockup/spinner-lockup.component.html
iivins-livefront: Updated - since lockup isn't used right now there's no consumers that need addressing

iivins-livefront:
> The spinner in the select view is misaligned. What did we change to cause this? Should we be worried about spinning positioning for consumers elsewhere in the app?
> 
> Image

Added `host: { class: "tw-inline-block" }` to fix consumer alignment regression caused by the removal of `@HostBinding`

libs/components/src/icon-button/icon-button.component.ts
```diff
@@ -91,6 +91,10 @@ export class BitIconButtonComponent implements ButtonLikeAbstraction, FocusableE
     return classes.join(" ");
   });
 
+  protected readonly spinnerSize = computed<SpinnerSize>(
+    (): SpinnerSize => (this.size() === "small" ? "sm" : "md"),
```
BryanCunningham: Nit: This return type can be removed. It's handled by the generic in the `computed` signal. And, we can implicitly return without the parens AFAIK. But maybe `computed` doesn't like that
```suggestion
    () => this.size() === "small" ? "sm" : "md",
```

BryanCunningham (approved):

libs/components/src/button/button.component.html
```diff
@@ -12,7 +12,7 @@
   </span>
   @if (showLoadingStyle()) {
     <span class="tw-absolute tw-inset-0 tw-flex tw-items-center tw-justify-center">
-      <bit-spinner size="fill" noColor></bit-spinner>
+      <bit-spinner [size]="spinnerSize()" />
```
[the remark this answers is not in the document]
BryanCunningham: This does not apply as whenever a button is 'loading' it is also inactive. So, inactive styles apply to the button

libs/components/src/spinner-lockup/spinner-lockup.mdx
```diff
@@ -0,0 +1,61 @@
...
+
+import * as stories from "./spinner-lockup.stories";
+
+<Meta of={stories} />
+
+```ts
+import { SpinnerLockupComponent } from "@bitwarden/components";
+```
+
+# Spinner Lockup
```
vleague2: Not a blocker, but we are using the Description block in other stories, so probably should do the same here

libs/components/src/spinner-lockup/spinner-lockup.stories.ts
```diff
@@ -0,0 +1,317 @@
...
+    `,
+  }),
+  args: {
+    variant: "primary",
+    size: "base",
+    orientation: "vertical",
+  },
+};
+
+export const VerticalWithRichContentDangerVariant: Story = {
```
vleague2: I don't see much difference between the default stories and these rich content stories

libs/components/src/spinner/spinner.component.ts
```diff
@@ -1,60 +1,58 @@
...
-        return ["tw-h-16"];
-      default:
-        return ["tw-h-full", "tw-w-full"];
-    }
-  }
+  private readonly i18nService = inject(I18nService);
+
+  readonly variant = input<SpinnerVariant>("primary");
+  readonly size = input<SpinnerSize>("base");
+  readonly title = input<string>(this.i18nService.t("loading")); // for accessibility, not visually rendered
```
vleague2: Can we convert to a jsdoc comment?

libs/components/src/spinner/spinner.component.ts
```diff
@@ -1,60 +1,58 @@
...
-// eslint-disable-next-line @angular-eslint/prefer-on-push-component-change-detection
+export type SpinnerVariant = "primary" | "subtle" | "success" | "warning" | "danger" | "contrast";
+
+export type SpinnerSize = "sm" | "md" | "base" | "lg";
+
+export const spinnerSizeStyles: Record<SpinnerSize, string[]> = {
+  sm: ["tw-w-4", "tw-h-4"],
+  md: ["tw-w-6", "tw-h-6"],
+  base: ["tw-w-14", "tw-h-14"],
+  lg: ["tw-w-20", "tw-h-20"],
```
vleague2: If these are all square, I think we can do the combined tailwind classes (like `size-20` iirc)

vleague2 (commented):
Small thoughts

vleague2 (approved):

willmartian (approved):

PR #19024, opened 2026-02-17
[PM-32237] Add back functionality to email OTP auth flow 

harr1424:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-32237

## 📔 Objective

To improve usability, we should add a Back affordance to the recipient verification page. This will allow users to quickly go back and update their email if entered incorrectly. When a user clicks Back, the expectation is that the input will continue to show the previously entered Email. 

On the verification page, we should also add subtitle below the H1 with the users email address. This provides visual context on whether the user to confirm that they entered the email correctly. 

## 📸 Screenshots

Screenshot 2026-02-18 at 11 06 41


harr1424:
@claude please review

apps/web/src/app/tools/send/send-access/send-access-email.component.ts
```diff
@@ -18,12 +26,31 @@ export class SendAccessEmailComponent implements OnInit, OnDestroy {
...
 
   readonly loading = input.required<boolean>();
+  readonly backToEmail = output<void>();
 
-  constructor() {}
+  constructor() {
+    // Update validators when enterOtp changes
+    effect(() => {
+      const isOtpMode = this.enterOtp();
+      if (this.email && this.otp) {
```
mcamirault: ⛏️ If you move this `effect` into the `ngOnInit` I don't think you'll have to do this check. I think it's better to consolidate all the initialization logic into one function as well

apps/web/src/app/tools/send/send-access/send-access-email.component.html
```diff
@@ -16,20 +16,31 @@
...
+    class="tw-mb-3"
+  >
+    <span>{{ "viewSend" | i18n }} </span>
+  </button>
+  <button
+    bitButton
+    type="button"
+    buttonType="secondary"
+    [block]="true"
+    (click)="backToEmail.emit(); $event.preventDefault(); $event.stopPropagation()"
```
mcamirault: 🎨 These actions should be moved into a function in `send-access-email.component.ts` called `onBackClick` or something similar.

Also just curious, what does preventing the event default and propagation do here?

apps/web/src/app/tools/send/send-access/send-auth.component.ts
```diff
@@ -89,6 +90,19 @@ export class SendAuthComponent implements OnInit {
...
+  onBackToEmail() {
+    this.enterOtp.set(false);
+    const otpControl = this.sendAccessForm.controls.otp;
+    if (otpControl) {
+      otpControl.clearValidators();
+      otpControl.setValue("");
+      otpControl.setErrors(null);
+      otpControl.markAsUntouched();
+      otpControl.markAsPristine();
+    }
```
mcamirault: 🎨 This should be done in the aforementioned `onBackClick` function in the `send-access-email.component.ts` since it's responsible for the form fields

PR #19025, opened 2026-02-17
[PM-26378] Auto confirm events

BTreston:
## 🎟️ Tracking
https://bitwarden.atlassian.net/browse/PM-26378

## 📔 Objective
Adds event logging handlers and logic to the auto confirm workflow + new event types for logging.

## 📸 Screenshots

https://github.com/user-attachments/assets/ec95eee7-710a-4a34-87b9-2946c1836ca3

https://github.com/user-attachments/assets/97ba43e0-9524-41f1-bdce-6f360de4adde


apps/web/src/app/core/event.service.ts
djsmith85, line 1: This should be moved under ownership of @bitwarden/team-admin-console-dev or @bitwarden/team-data-insights-and-reporting-dev ? Maybe `apps/web/src/app/admin-console/organizations/core/services/` for now? Until everything is handed over to DIRT?

BTreston: This will be handled in https://bitwarden.atlassian.net/browse/PM-28488, so I'd say out of scope for this ticket.

PR #19026, opened 2026-02-17
[CL-959] update badge component styles

BryanCunningham:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/CL-959

Note to consumer teams about changes you may see in your team's code:
1. bitBadge replaced with bit-chip-action
Anywhere you're using bitBadge as a button/link, migrate to the bit-chip-action component:

Before: 
```js
<button bitBadge variant="primary">Fill</button>
```
After: 
```js
<bit-chip-action variant="primary" label="Fill"></bit-chip-action>
```
Text content moves from slot to [label] property binding

2. Notification badges replaced with bit-berry
Before:
```js
<span bitBadge variant="notification">1</span>
```

After:
```js
<bit-berry variant="danger" [value]="1"></bit-berry>
```

3. Import changes
Remove BadgeModule from your imports array
Add ChipActionComponent for interactive chip/button use cases
Add BerryComponent for notification count indicators

4. Icon removed from bitBadge in SM secrets list
The warning icon (<i>) inside a badge in the Secrets Manager secrets list template was removed. As it's the default icon in the warning badge

## 📔 Objective

Update badge
- Badges are now static
- Update existing interactive badges to use new chip component

## 📸 Screenshots
image

### DIFF
--- a/libs/vault/src/components/organization-name-badge/organization-name-badge.component.ts
+++ b/libs/vault/src/components/organization-name-badge/organization-name-badge.component.ts
@@ -5,8 +5,6 @@ import { RouterModule } from "@angular/router";
 
 import { JslibModule } from "@bitwarden/angular/jslib.module";
 import { Unassigned } from "@bitwarden/common/admin-console/models/collections";
-import { AvatarService } from "@bitwarden/common/auth/abstractions/avatar.service";
-import { TokenService } from "@bitwarden/common/auth/abstractions/token.service";
 import { I18nService } from "@bitwarden/common/platform/abstractions/i18n.service";
 import { ChipActionComponent } from "@bitwarden/components";
 import { OrganizationId } from "@bitwarden/sdk-internal";
@@ -35,11 +33,7 @@ export class OrganizationNameBadgeComponent implements OnChanges {
   textColor: string;
   isMe: boolean;
 
-  constructor(
-    private i18nService: I18nService,
-    private avatarService: AvatarService,
-    private tokenService: TokenService,
-  ) {}
+  constructor(private i18nService: I18nService) {}
 
   // ngOnChanges is required since this component might be reused as part of
   // cdk virtual scrolling


libs/components/src/badge/badge.component.ts
```diff
@@ -8,78 +8,107 @@ import {
...
+  | LegacyInfoVariant
+  | LegacySecondaryVariant;
+
+export type BadgeSize = "small" | "large";
+
+const variantStyles: Record<BadgeVariant, string[]> = {
+  primary: ["tw-bg-bg-brand-softer", "tw-border-border-brand-soft", "tw-text-fg-brand-strong"],
+  info: ["tw-bg-bg-brand-softer", "tw-border-border-brand-soft", "tw-text-fg-brand-strong"],
+  subtle: ["tw-bg-bg-secondary", "tw-border-border-base", "tw-text-fg-body"],
+  secondary: ["tw-bg-bg-primary", "tw-border-border-base", "tw-text-fg-body"],
```
vleague2: Should these two be the same, if secondary is the legacy of subtle?

libs/components/src/badge/badge.component.ts
```diff
@@ -88,56 +117,57 @@ export class BadgeComponent implements FocusableElement {
...
 
   /**
-   * Tailwind max-width class to apply when truncating is enabled.
+   * Tailwind max-width class to apply to constrain badge width.
    * Must be a valid Tailwind max-width utility class (e.g., "tw-max-w-40", "tw-max-w-xs").
    */
   readonly maxWidthClass = input<`tw-max-w-${string}`>("tw-max-w-40");
 
+  readonly startIcon = input<BitwardenIcon>(undefined);
+  readonly hideStartIcon = input(false, { transform: booleanAttribute });
```
vleague2: Part of me wants to get rid of this input and have people pass `null` to `startIcon` if they want to avoid the default icon.... but there are pros and cons to both approaches. What do you think?

BryanCunningham: Personally, when there are default icons involved, I like the idea of having to explicitly say 'hide this thing'. Rather than having to try and figure out why passing nothing still renders an icon

BryanCunningham: @vleague2 After our chat in dev sync, I'll change this to accept `null` and remove the explicit input

libs/components/src/badge/badge.component.ts
BryanCunningham: They should be. I'll fix. Thanks!

apps/browser/src/vault/popup/components/at-risk-passwords/at-risk-passwords.component.html
```diff
@@ -65,27 +65,20 @@
         <bit-item-action slot="end">
           <button
             type="button"
-            bitBadge
-            variant="primary"
+            bitButton
```
vleague2, line 68: I don't know this component enough to know how this'll look, but should we run this change by design since the button is not really visually the same as the old badges?

apps/browser/src/vault/popup/components/at-risk-passwords/at-risk-passwords.component.html
```diff
@@ -65,27 +65,20 @@
...
+            buttonType="primary"
+            [loading]="launchingCipher() == cipher"
```
vleague2, line 70: Wonder if we should have a follow-up ticket for the team to use `bitAction` instead?

apps/web/src/app/dirt/reports/pages/inactive-two-factor-report.component.html
```diff
@@ -108,9 +108,13 @@
             }
             <td bitCell class="tw-text-right">
               @if (cipherDocs.has(row.id)) {
-                <a bitBadge href="{{ cipherDocs.get(row.id) }}" target="_blank" rel="noreferrer">
-                  {{ "instructions" | i18n }}</a
-                >
+                <a
+                  bit-chip-action
+                  [href]="cipherDocs.get(row.id)"
+                  [label]="'instructions' | i18n"
```
vleague2, line 114: I see SonarCloud is throwing a warning for the anchor not having accessible text, now I am wondering if we should've used a slot instead of an input

libs/components/src/badge/badge.component.html
```diff
@@ -1,3 +1,9 @@
-<span [ngClass]="{ 'tw-truncate tw-block': truncate() }">
-  <ng-content></ng-content>
-</span>
+<div class="tw-inline-flex tw-min-w-0 tw-w-full tw-max-w-full tw-items-center tw-gap-0.5">
+  @if (computedIcon()) {
+    <bit-icon [name]="computedIcon()" [class]="iconSizeStyles()" />
+  }
+
+  <span #textContainer class="tw-px-1 tw-text-left tw-min-w-0 tw-flex-1 tw-truncate">
```
vleague2: Logical property for `tw-text-left`?

libs/components/src/badge/badge.stories.ts
```diff
@@ -30,118 +27,95 @@ export const Default: Story = {
...
+          <span bitBadge variant="accent-primary">Accent Primary</span>
+        </div>
+      </div>
+    `,
+  }),
+  parameters: {
+    chromatic: {
+      modes: {
+        light: { theme: "light" },
+        dark: { theme: "dark" },
```
vleague2, line 91: I don't know if we want to start screenshotting in dark mode

libs/components/src/badge/badge.component.ts
```diff
@@ -8,78 +7,107 @@ import {
...
- * > `NOTE:` The Focus and Hover states only apply to badges used for interactive events.
  *
- * > `NOTE:` The `disabled` state only applies to buttons.
+ * The Badge directive can only be used on a `<span>` tag
  */
 @Component({
-  selector: "span[bitBadge], a[bitBadge], button[bitBadge]",
-  providers: [{ provide: FocusableElement, useExisting: BadgeComponent }],
-  imports: [CommonModule],
+  selector: "span[bitBadge]",
```
willmartian: Can we add `bit-badge`? That should be preferred over placing on a span. (Remove comment)

I created a ticket for us to migrate this later, targeting `main`: https://bitwarden.atlassian.net/browse/CL-1093

libs/components/src/badge/badge.component.ts
```diff
@@ -8,78 +7,107 @@ import {
...
+
+const defaultIconMap: Record<BadgeVariant, BitwardenIcon> = {
+  info: "bwi-info-circle",
+  subtle: "bwi-info-circle",
+  secondary: "bwi-info-circle",
+  primary: "bwi-info-circle",
+  success: "bwi-check-circle",
+  warning: "bwi-exclamation-triangle",
+  danger: "bwi-error",
+  "accent-primary": "bwi-clock",
```
willmartian: Clock is the desired default? Should this just not have a icon?

apps/browser/src/vault/popup/components/at-risk-passwords/at-risk-passwords.component.html
BryanCunningham: @vleague2 I've chatted with Emily about this. She says this should be a button instead. In the upcoming design updates for this page, these are buttons

apps/web/src/app/dirt/reports/pages/inactive-two-factor-report.component.html
BryanCunningham: @vleague2 Hmm... interesting problem. I don't really want a slot here as it would open the door for inconsistencies. Wondering if there is a way to change SonarCloud to exclude links with this `bit-chip-action` attr instead

libs/components/src/badge/badge.stories.ts
BryanCunningham: @vleague2 I don't have super strong feelings about this per se. I've been adding this to the 'all variants' versions of the updated components. Feels like it could be valuable to me in case someone changes one of the dark mode theme colors. Open to removing it though if we don't feel it's necessary. 

CC// @willmartian thoughts on this?

libs/components/src/badge/badge.component.ts
BryanCunningham: @vleague2 Good call. Should have thought about that :)

libs/components/src/badge/badge.component.ts
BryanCunningham: @willmartian According to design, this is the correct default icon. Easy enough to change later though if they decide they want a different default

libs/components/src/badge/badge.stories.ts
willmartian: I think we should keep it unless instructed to remove.

vleague2: If we want to enable dark mode tests, why not do that at the top level then?

libs/components/src/badge/badge.component.ts
vleague2: Credit to @willmartian 😂

libs/components/src/badge/badge.stories.ts
BryanCunningham: Just didn't think it was necessary for all the stories. Just the ones that show all the variants in one story. It's not full coverage by any means but, it adds a little safety net without taking ALL the screenshots in both themes

libs/components/src/badge/badge.component.ts
BryanCunningham: @vleague2 Oh, hahaha. My bad. Was just used to tagging you in the other comments :) Thanks @willmartian

libs/components/src/badge/badge.component.ts
willmartian: Just chatted with Emily and we are good to have no default on "accent-primary", since it is not semantic and devs will commonly need to change it to something other than the clock

libs/components/src/badge/badge.stories.ts
vleague2: Oooooh okay okay yes this seems reasonable then

libs/components/src/badge/badge.component.ts
```diff
@@ -88,59 +116,55 @@ export class BadgeComponent implements FocusableElement {
...
+  protected readonly titleContent = computed(() => {
+    // Use custom title if provided, otherwise use text content
+    const customTitle = this.title();
+    if (customTitle !== undefined) {
+      return customTitle;
     }
-    return this.truncate() ? this.el.nativeElement?.textContent?.trim() : null;
-  });
 
-  getFocusTarget() {
```
vleague2, line 143: Hmmm..... I wonder what this was used for..... 🤔

libs/components/src/multi-select/multi-select.component.html
```diff
@@ -23,22 +23,14 @@
     <bit-spinner></bit-spinner>
   </ng-template>
   <ng-template ng-label-tmp let-item="item" let-clear="clear">
-    <button
-      type="button"
-      bitBadge
-      variant="primary"
-      class="tw-me-1 disabled:tw-border-0 tw-flex tw-gap-1.5 tw-items-center"
+    <bit-chip
```
vleague2, line 26: The stories need to be updated for multiselect, looks like missing translations now

libs/components/src/badge/badge.component.ts
BryanCunningham: It was used by `FocusableElement` that was provided before.
```js
providers: [{ provide: FocusableElement, useExisting: BadgeComponent }]
```
As it's no longer a focusable element, it's not necessary

vleague2: Ty I knew there was something there in the back of my brain!!!

libs/components/src/multi-select/multi-select.component.html
vleague2: Nevermind, it's working now. The error looks unrelated and I think the page just wasn't done loading.

apps/web/src/app/dirt/reports/pages/inactive-two-factor-report.component.html
vleague2: To close this out, we are modifying the sonar cloud rule so that it won't see this as a problem

libs/components/src/badge/badge.component.ts
```diff
@@ -8,78 +7,107 @@ import {
...
- * Badges are primarily used as labels, counters, and small buttons.
- * Typically Badges are only used with text set to `text-xs`. If additional sizes are needed, the component configurations may be reviewed and adjusted.
- *
- * The Badge directive can be used on a `<span>` (non clickable events), or an `<a>` or `<button>` tag
- *
- * > `NOTE:` The Focus and Hover states only apply to badges used for interactive events.
+ * Badges are used as labels.
  *
- * > `NOTE:` The `disabled` state only applies to buttons.
+ * The Badge directive can only be used on a `<span>` tag
```
willmartian, line 92: This comment is no longer needed.

libs/components/src/badge/badge.component.ts
```diff
@@ -88,59 +116,55 @@ export class BadgeComponent implements FocusableElement {
...
+      ...variantStyles[this.variant()],
+    ].concat(this.truncate() ? this.maxWidthClass() : "");
   });
 
-  protected readonly titleAttr = computed(() => {
-    const title = this.title();
-    if (title !== undefined) {
-      return title;
+  /**
+   * Computed title content - only shows when content is overflowing
```
willmartian, line 159: Where/how do we achieve "only shows when content is overflowing"?

BryanCunningham: We don't really. This comment isn't really accurate. Will update

willmartian: Do we want to remove this `title` binding altogether? `title` has largely been replaced by tooltip--but is that needed here either? Curious what you think.

BryanCunningham: I was curious about that as well. I can try it out

BryanCunningham: The only problem is we don't know if the item is actually truncated. So, showing the title even when it's not truncated is slightly less intrusive

BryanCunningham: For posterity: We chatted about this and decided to keep the title attr for now until a bettor solution is found

libs/vault/src/components/organization-name-badge/organization-name-badge.component.ts
```diff
@@ -37,35 +33,16 @@ export class OrganizationNameBadgeComponent implements OnChanges {
   textColor: string;
   isMe: boolean;
```
jaasen-livefront, line 34: 🎨 Not blocking but we should probably remove these since they're no longer  used.

PR #19027, opened 2026-02-17
[PM-31060] Product Update Notification

nick-livefront:
## 🎟️ Tracking

[PM-31060](https://bitwarden.atlassian.net/browse/PM-31060)

## 📔 Objective

- Adds notification to users that the autofill action has changed.
  - The animation only shows 4 cycles and then stops
  - Once a user dismisses the icon they won't see it again. 

### Open Question for Design

- There is a Refresh icon that is present when the extension is opened in the sidebar in FireFox. Are there any changes in relation to the two icons being present?
- The AC of the ticket requires that this shows for "existing" users. I currently have this showing for any user that is older than 7 days. How does that feel?

## 📸 Screenshots

|Dark mode | Light mode | FireFox|
|-|-|-|
|<video src="https://github.com/user-attachments/assets/ef53bcb9-9e34-4fbc-ba38-8af48ab8af45" />|<video src="https://github.com/user-attachments/assets/135330de-4256-4260-8fa9-8635971462c5" />| <video src="https://github.com/user-attachments/assets/a573e975-1d34-46fb-9642-5e74b063d96f" />|

[PM-31060]: https://bitwarden.atlassian.net/browse/PM-31060?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


emily-livefront:
Thanks @nick-livefront !

This looks great — no design changes needed that I'm seeing. Regarding your two questions:

> There is a Refresh icon that is present when the extension is opened in the sidebar in FireFox. Are there any changes in relation to the two icons being present?

- No changes needed to this! The mismatch in size should be resolved with the upcoming icon button and iconography changes.

> The AC of the ticket requires that this shows for "existing" users. I currently have this showing for any user that is older than 7 days. How does that feel?

- Is it possible to have this show for any user that is older than the date the feature is released? I know that might be a tough place to get to, so totally open to other options, but ideally we want to get any user who has interacted with items in their vault before this change was released included in the group who sees the notification. Maybe even "any user older than 1 day" would suffice?

nick-livefront:
> * Is it possible to have this show for any user that is older than the date the feature is released? I know that might be a tough place to get to, so totally open to other options, but ideally we want to get any user who has interacted with items in their vault before this change was released included in the group who sees the notification. Maybe even "any user older than 1 day" would suffice?

After a discussion outside of GitHub, this feature will be beneficial to any user to give more detail on how the autofill feature works. [215a3b6](https://github.com/bitwarden/clients/pull/19027/commits/215a3b665653d0704f09c0fc0a3444a8f0b64158)

libs/assets/src/svg/svgs/info-filled.icon.ts
```diff
@@ -0,0 +1,7 @@
+import { svg } from "../svg";
+
+export const InfoFilledIcon = svg`
+  <svg width="24" height="24" viewBox="0 0 24 24" fill="none" xmlns="http://www.w3.org/2000/svg">
```
vleague2: Can we remove the hardcoded width and height? I know this is temporary until the icon font is updated, but juuust in case, I want to follow the responsiveness guidelines we have for svgs

nick-livefront: ✅ [ff6b880](https://github.com/bitwarden/clients/pull/19027/commits/ff6b880010dda287473695e181807d47767e99f2)

gbubemismith (approved):
This looks great. 💪🏼

PR #19032, opened 2026-02-17
[PM-29528]Remove @ts-strict-ignore in autofill.service.ts

dan-livefront:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-29528
## 📔 Objective

- Added null guards for tab, account, and form-related fields
- Fixed ReplaySubject typing from empty tuple to PageDetail[]
- Introduced discriminated union to safely support password only forms
- Hardened observable pipeline to avoid accumulating invalid PageDetail entries
- Added test coverage for password only forms with username: null

### DIFF
--- a/.github/workflows/publish-desktop.yml
+++ b/.github/workflows/publish-desktop.yml
@@ -32,6 +32,14 @@ on:
         required: true
         default: true
         type: boolean
+      snap_channel:
+        description: 'Snap Store channel to publish to'
+        required: false
+        default: 'stable'
+        type: choice
+        options:
+          - stable
+          - beta
       choco_publish:
         description: 'Publish to Chocolatey store'
         required: true
@@ -234,14 +242,17 @@ jobs:
 
       - name: Download artifacts
         working-directory: apps/desktop/dist
-        run: wget "https://github.com/bitwarden/clients/releases/download/${_RELEASE_TAG}/bitwarden_${_PKG_VERSION}_amd64.snap"
+        run: |
+          wget "https://github.com/bitwarden/clients/releases/download/${_RELEASE_TAG}/bitwarden_${_PKG_VERSION}_amd64.snap"
+          wget "https://github.com/bitwarden/clients/releases/download/${_RELEASE_TAG}/bitwarden_${_PKG_VERSION}_arm64.snap"
 
       - name: Deploy to Snap Store
         if: ${{ inputs.publish_type != 'Dry Run' }}
         env:
           SNAPCRAFT_STORE_CREDENTIALS: ${{ steps.retrieve-secrets.outputs.snapcraft-store-token }}
         run: |
-          snapcraft upload "bitwarden_${_PKG_VERSION}_amd64.snap" --release stable
+          snapcraft upload "bitwarden_${_PKG_VERSION}_amd64.snap" --release ${{ inputs.snap_channel }}
+          snapcraft upload "bitwarden_${_PKG_VERSION}_arm64.snap" --release ${{ inputs.snap_channel }}
           snapcraft logout
         working-directory: apps/desktop/dist
 
--- a/apps/browser/src/autofill/popup/settings/autofill.component.ts
+++ b/apps/browser/src/autofill/popup/settings/autofill.component.ts
@@ -1,5 +1,3 @@
-// FIXME: Update this file to be type safe and remove this and next line
-// @ts-strict-ignore
 import { CommonModule } from "@angular/common";
 import { Component, DestroyRef, OnInit } from "@angular/core";
 import { takeUntilDestroyed } from "@angular/core/rxjs-interop";
@@ -159,7 +157,7 @@ export class AutofillComponent implements OnInit {
   clearClipboard!: ClearClipboardDelaySetting;
   clearClipboardOptions: { name: string; value: ClearClipboardDelaySetting }[];
   defaultUriMatch: UriMatchStrategySetting = UriMatchStrategy.Domain;
-  uriMatchOptions: { name: string; value: UriMatchStrategySetting; disabled?: boolean }[];
+  uriMatchOptions: { name: string; value: UriMatchStrategySetting | null; disabled?: boolean }[];
   showCardsCurrentTab: boolean = true;
   showIdentitiesCurrentTab: boolean = true;
   /** Non-null asserted. */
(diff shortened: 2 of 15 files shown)


apps/browser/src/autofill/services/autofill.service.spec.ts
```diff
@@ -189,7 +189,7 @@ describe("AutofillService", () => {
           sender: AutofillMessageSender.collectPageDetailsFromTabObservable,
           tab,
         },
-        null,
+        undefined,
```
dan-livefront, line 192: Changed to match chrome api 

```

export function sendMessage<M = any, R = any>(
            tabId: number,
            message: M,
            options: MessageSendOptions | undefined,
            /** @since Chrome 99 */
            callback: (response: R) => void,
        ): void;


```

apps/browser/src/autofill/services/autofill.service.ts
```diff
@@ -1216,23 +1299,23 @@ export default class AutofillService implements AutofillServiceInterface {
...
           if (
             siOptions[0][0] != null &&
             siOptions[0][0] !== "" &&
             (siOptions[12][0] == null || siOptions[12][0] === "")
           ) {
-            index = parseInt(card.expMonth, null) - 1;
+            index = parseInt(card.expMonth, 10) - 1;
           } else {
-            index = parseInt(card.expMonth, null);
+            index = parseInt(card.expMonth, 10);
```
dan-livefront, line 1289: Always pass 10 to `parseInt` to force decimal parsing otherwise otherwise the radix may be inferred, which can cause inconsistent results across environments. Using 10 guarantees predictable behavior for normal numers

jprusik: yep - protect against possible implementation differences now and/or in the future. ✨

apps/browser/src/autofill/services/autofill.service.ts
```diff
@@ -740,22 +761,25 @@ export default class AutofillService implements AutofillServiceInterface {
...
         }
       });
 
       pageDetails.fields.forEach((field) => {
-        // eslint-disable-next-line
-        if (filledFields.hasOwnProperty(field.opid)) {
+        if (field.opid == null) {
+          return;
+        }
+        const fieldOpid = field.opid;
```
jprusik: Non-blocking nit
```suggestion
        const fieldOpid = field.opid;
        if (fieldOpid == null) {
          return;
        }
```

apps/browser/src/autofill/services/autofill.service.ts
```diff
@@ -740,22 +761,25 @@ export default class AutofillService implements AutofillServiceInterface {
       return null;
     }
 
-    let fillScript = new AutofillScript();
+    const fillScript = new AutofillScript();
```
jprusik, line 758: ~~Given that this is now unused elsewhere (beyond the final return), are we able to move it further down any more?~~

**edit:** nvm, it can't, since it's passed to `generateLoginFillScript` later from this scope level

apps/browser/src/autofill/services/autofill.service.ts
```diff
@@ -773,40 +797,40 @@ export default class AutofillService implements AutofillServiceInterface {
...
           }
 
-          filledFields[field.opid] = field;
+          filledFields[fieldOpid] = field;
           AutofillService.fillByOpid(fillScript, field, val);
         }
       });
     }
 
+    let result: AutofillScript | null = null;
```
jprusik, line 818: ~It took me a minute to see what you did will `fillScript` [over here](https://github.com/bitwarden/clients/pull/19032/changes#diff-ff919521412a590452c7e201f94d1c80997766b29cf4870faa84275768edc6a8L743-R764), but I do think I like this better (over reusing the variable from the sibling block).~

*edit:* poking at this more, does it need to be `new AutofillScript()`?

```suggestion
    let result: AutofillScript | null = new AutofillScript();
``` 

the `result` name feels a bit more ambiguous to me, but is a non-blocking nit imo

**edit edit:** nvm, I see now - separating the concerns of representing `AutofillScript` and storing the return value from `generateLoginFillScript` - good de-coupling 👍

apps/browser/src/autofill/services/autofill.service.ts
```diff
@@ -842,12 +866,15 @@ export default class AutofillService implements AutofillServiceInterface {
...
+    let pf: AutofillField | null = null;
+    let username: AutofillField | null = null;
+    let totp: AutofillField | null = null;
     const login = options.cipher.login;
     fillScript.savedUrls =
-      login?.uris?.filter((u) => u.match != UriMatchStrategy.Never).map((u) => u.uri) ?? [];
+      login?.uris
+        ?.filter((u) => u.match != UriMatchStrategy.Never)
+        ?.map((u) => u.uri)
+        ?.filter((uri): uri is string => uri != null) ?? [];
```
jprusik: This could be a single reducer to avoid multiple array iterations, but given we're talking about 1-3 entries at most (in most cases) and that you're aiming for logical parity/small change footprint, probably not a huge deal for this pass.

apps/browser/src/autofill/services/autofill.service.ts
```diff
@@ -874,9 +901,9 @@ export default class AutofillService implements AutofillServiceInterface {
     const prioritizedPasswordFields =
       loginPasswordFields.length > 0 ? loginPasswordFields : registrationPasswordFields;
 
-    const focusedField =
-      options.focusedFieldOpid &&
-      pageDetails.fields.find((f) => f.opid === options.focusedFieldOpid);
+    const focusedField = options.focusedFieldOpid
+      ? pageDetails.fields.find((f) => f.opid === options.focusedFieldOpid)
+      : undefined;
```
jprusik, line 902: ~~Does `find` not return non-matches as `undefined`? Come to think of it, the check in the original logic seems unnecessary too...? Or is this just ts finickiness around being explicit?~~

**edit:** Typescript seems fine with 

```suggestion
    const focusedField = pageDetails.fields.find((f) => f.opid === options.focusedFieldOpid);
```

dan-livefront: My reasoning here was, “no focused field” is getting conflated with “the focused field has no opid”, the ternary avoids this by not calling find when focusedFieldOpid is falsy, this was more of a semantic issue.

apps/browser/src/autofill/services/autofill.service.ts
dan-livefront: Yep, I was trying to keep the scope reduced 🥁 but can add this, I think it aligns with our goal to be more performant where possible, these changes can make a difference as they compound over time.

jprusik: Maybe something like?

```suggestion
    const loginURIs = login?.uris || [];

    if (loginURIs.length > 0) {
      fillScript.savedUrls = loginURIs
      .reduce<string[]>((acc, savedURI) => {
        if (savedURI.match != UriMatchStrategy.Never && savedURI.uri != null) {
          return [...acc, savedURI.uri];
        }

        return acc;
      }, []);
    }
```

apps/browser/src/autofill/services/autofill.service.ts
dan-livefront: Ope! Thanks for keeping me honest there

apps/browser/src/autofill/services/autofill.service.ts
dan-livefront: will implement something like 

```
const loginURIs = login?.uris ?? [];
    fillScript.savedUrls = loginURIs.reduce<string[]>((acc, savedURI) => {
      if (savedURI.match != UriMatchStrategy.Never && savedURI.uri != null) {
        acc.push(savedURI.uri);
      }
      return acc;
    }, []);

```

so that we avoid `savedUrls` never being left unset

apps/browser/src/autofill/services/autofill.service.ts
jprusik: ah right - and also; why pay the cost of a lookup when you already know the outcome? 👍

apps/browser/src/autofill/services/autofill.service.ts
```diff
@@ -411,20 +416,30 @@ export default class AutofillService implements AutofillServiceInterface {
...
+      !options.cipher ||
+      !options.pageDetails ||
+      !options.pageDetails.length
+    ) {
       throw new Error("Nothing to autofill.");
     }
 
     let totp: string | null = null;
 
+    const tabUrl = tab.url;
```
jprusik: Non-blocking nit

```suggestion
    const tabUrl = tab?.url;
    if (
      !tabUrl ||
      !options.cipher ||
      !options.pageDetails ||
      (options.pageDetails?.length || 0) > 0
    ) {
      throw new Error("Nothing to autofill.");
    }

    let totp: string | null = null;

```

jprusik (approved):

PR #19033, opened 2026-02-17
[PM-32013] Empty state incorrectly rendered

voommen-livefront:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-32013

## 📔 Objective

Update the Access intelligence module -> Applications tab to do the following:
- When there are no rows to render show the message
- When apps are marked as critical or non-critical, then update the chip-select to reflect the latest numbers

## 📸 Screenshots

https://github.com/user-attachments/assets/4809b1f2-2813-4c3f-be5d-7ede7ac625c7

### DIFF
--- a/libs/components/src/chip-select/chip-select.component.ts
+++ b/libs/components/src/chip-select/chip-select.component.ts
@@ -115,9 +115,7 @@ export class ChipSelectComponent<T = unknown> implements ControlValueAccessor {
 
       // when the options change, we need to change our selectedOption
       // to reflect the changed options.
-      if (
-        currentSelection?.value != null
-      ) {
+      if (currentSelection?.value != null) {
         this.selectedOption = this.findOption(this.rootTree, currentSelection.value);
       }
 


libs/components/src/chip-select/chip-select.component.ts
```diff
@@ -106,8 +106,23 @@ export class ChipSelectComponent<T = unknown> implements ControlValueAccessor {
...
       this.initializeRootTree(this.options());
 
+      // when the options change, we need to change our selectedOption
+      // to reflect the changed options.
+      if (currentSelection !== undefined && currentSelection !== null) {
+        const foundOption =
+          Array.from(this.childParentMap.entries()).find(
+            ([key, _]) => key.value === currentSelection.value,
+          )?.[0] ?? null;
+        this.selectedOption = foundOption;
```
vleague2: I didn't look too in-depth, but does using the existing `this.findOption` function work here?

libs/components/src/chip-select/chip-select.component.ts
```diff
@@ -106,8 +106,23 @@ export class ChipSelectComponent<T = unknown> implements ControlValueAccessor {
...
+
+      // when the options change, clear the childParentMap
+      this.childParentMap.clear();
+
       this.initializeRootTree(this.options());
 
+      // when the options change, we need to change our selectedOption
+      // to reflect the changed options.
+      if (currentSelection !== undefined && currentSelection !== null) {
+        const foundOption =
```
lastbestdev: Should this use `findOption` instead?


```suggestion
        const foundOption = this.findOption(this.rootTree, currentSelection.value)
```

lastbestdev: Resolving, same comment as @vleague2 above

vleague2: Haha 👯

libs/components/src/chip-select/chip-select.component.ts
voommen-livefront: @vleague2 you are correct - I could use `findOption` instead of searching the `childParentMap` myself.  Updated the PR. Thanks for catching this.

libs/components/src/chip-select/chip-select.component.ts
voommen-livefront: @lastbestdev Thanks for catching this - yes, I can use `findOption` instead of searching within the `childParentMap`.

libs/components/src/chip-select/chip-select.component.ts
```diff
@@ -106,8 +106,23 @@ export class ChipSelectComponent<T = unknown> implements ControlValueAccessor {
...
+      this.childParentMap.clear();
+
       this.initializeRootTree(this.options());
 
+      // when the options change, we need to change our selectedOption
+      // to reflect the changed options.
+      if (
+        currentSelection !== undefined &&
+        currentSelection !== null &&
+        currentSelection.value !== null
```
lastbestdev: This condition can be reduced to handle both `null` and `undefined` like so:


```suggestion
        currentSelection?.value != null
```

lastbestdev (approved):

lastbestdev (approved):

vleague2 (approved):

PR #19044, opened 2026-02-17
[PM-32421] CLI: prevent false-success SSO login when account cannot decrypt (#18992)

### LINKED ISSUE #18992: CLI SSO login succeeds while actually failing
### Steps To Reproduce

`bw login` appears to succeed, but it does not: `bw unlock` immediately after fails in a way it should not.  See output below.

We use SSO login, so there are no master keys.

If the `login` process cannot set a user key, nor prepare `unlock` to be used, then it shold not report (in green, implying success) `you are logged in`, rather it should report an error or at the very least a warning that tells the user that yes, perhaps you are technically logged in, but we failed at setting up the environment for yoy.

Then, `unlock` should not be prompting for a master key on an SSO org account that cannot have a master key, it should be producing some sort of (idealy, useful) error instead of blindly prompting for something non-existent.

We'll follow up the login/unlock problem with [paid!] customer support, but the utility not being able to set things up correctly and then saying "yay, we're all good", and then incorrectly prompting for the master password is (IMHO) a bug in the tool, not just a support case.

### Expected Result

`login` should be able to set a user key, or whatever it needs to do, and emit a warning or error, not success, if it can't.

`unlock` should not prompt for a non-existent master password.

### Actual Result

```
C:\Users\athompson\Downloads>bw login --sso
Unable to set user key due to missing device key.
Unable to set user key due to missing encrypted device private key.
Unable to set user key due to missing encrypted user key.
You are logged in!
```

then

```
C:\Users\athompson\Downloads>bw unlock
? Master password: [hidden]
Master password is required
C:\Users\athompson\Downloads>
```

### Screenshots or Videos

_No response_

### Additional Context

```
C:\Users\athompson\Downloads>bw --version
2025.12.1
```

### Operating System

Windows

### Operating System Version

Windows 11 25H2 build 26200.7840

### Shell

PowerShell

### Build Version

CMD.EXE Version 10.0.26200.7840

### Issue Tracking Info

HadiAyache:
## Summary
Fixes a CLI SSO edge case where login reports **"You are logged in!"** even though the account cannot decrypt/unlock in CLI.

This updates SSO validation to fail early only when all are true:
- user has no master password
- no key connector path is configured
- no user key is available in memory

This avoids over-blocking valid no-master-password flows (for example, key connector setup).

## Changes
- `apps/cli/src/auth/commands/login.command.ts`
  - refine `validateSsoUserInMpEncryptionOrgHasMp` gating logic
  - return clearer error message for non-decryptable SSO CLI state
- `apps/cli/src/auth/commands/login.command.spec.ts` (new)
  - regression test for #18992 case
  - allow-case tests for key connector, existing user key, and master-password users

## Tests
Ran locally:
- `npx jest apps/cli/src/auth/commands/login.command.spec.ts --runInBand`
- `npx jest libs/auth/src/common/login-strategies/sso-login.strategy.spec.ts --runInBand`
- `npx jest apps/cli/src/key-management/commands/unlock.command.spec.ts --runInBand`

All passed.

Closes #18992

### DIFF
--- a/apps/cli/src/auth/commands/login.command.spec.ts
+++ b/apps/cli/src/auth/commands/login.command.spec.ts
@@ -1,0 +1,138 @@
+import { mock } from "jest-mock-extended";
+import { of } from "rxjs";
+
+import {
+  LoginStrategyServiceAbstraction,
+  SsoUrlService,
+  UserDecryptionOptionsServiceAbstraction,
+} from "@bitwarden/auth/common";
+import { OrganizationService } from "@bitwarden/common/admin-console/abstractions/organization/organization.service.abstraction";
+import { PolicyApiServiceAbstraction } from "@bitwarden/common/admin-console/abstractions/policy/policy-api.service.abstraction";
+import { PolicyService } from "@bitwarden/common/admin-console/abstractions/policy/policy.service.abstraction";
+import { AccountService } from "@bitwarden/common/auth/abstractions/account.service";
+import { AuthService } from "@bitwarden/common/auth/abstractions/auth.service";
+import { MasterPasswordApiService } from "@bitwarden/common/auth/abstractions/master-password-api.service.abstraction";
+import { TwoFactorApiService, TwoFactorService } from "@bitwarden/common/auth/two-factor";
+import { CryptoFunctionService } from "@bitwarden/common/key-management/crypto/abstractions/crypto-function.service";
+import { EncryptedMigrator } from "@bitwarden/common/key-management/encrypted-migrator/encrypted-migrator.abstraction";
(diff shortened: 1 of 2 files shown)


HadiAyache:
Quick status: current failing checks appear to be permission-gated workflow checks (author permission level) rather than deterministic code/test failures in this branch. If maintainers can rerun/approve the gated workflow path, I can immediately address any concrete code failures that remain.

JaredSnider-Bitwarden:
@HadiAyache thank you for your submission. We've chosen to solve this with product approved copy internally and simply based on known supported user decryption options with https://github.com/bitwarden/clients/pull/22411.

PR #19045, opened 2026-02-18
[PM-32423] Add a less confusing argument and environment variable name to disable updates

marcquark:
## 🎟️ Tracking

It's good to have the ability to disable the auto-updater, for example when an installation is managed through a package manager such as chocolatey or something more enterprise-y like MS Intune. However the currently available environment variable name `ELECTRON_NO_UPDATER` sounds like it might impact other Electron-based apps installed on a machine, which may not be desirable.

## 📔 Objective

I'm proposing the addition of an additional environment variable and CLI argument, keeping the old ones around for backwards compatibility. The new name makes it more obvious that the environment variable is meant for Bitwarden only and nothing else.

Related Community thread: https://community.bitwarden.com/t/add-an-option-to-disable-update/8123


addisonbeck (approved):

addisonbeck:
Thanks for the contribution! We just cut release branches this morning, so it'll still be 4 weeks before this lands in production. You'll see it in `2026.7.0`!

PR #19047, opened 2026-02-18
[PM-32412] -  fix click on "Fill" text

jaasen-livefront:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-32412

## 📔 Objective

This PR fixes an issue where clicking directly on the "Fill" text didn't correctly autofill. It also fixes an issue where the cursor would change to arrow when within the padding of the `<bit-item>` and the background highlight would be lost.

## 📸 Screenshots

Before

https://github.com/user-attachments/assets/0727c852-713c-4998-ae17-28ffc35ac843

After

https://github.com/user-attachments/assets/87c236f3-d637-43a0-a742-d6ce28da9718


jaasen-livefront:
@shane-melton Added a missing screen recording if you wanna take a look!

PR #19058, opened 2026-02-18
[PM-32442] Enforce tsc-strict on desktop and CLI apps

coroiu:
## 📔 Objective

`apps/desktop` and `apps/cli` were missing from the root `tsconfig.json`, meaning `tsc-strict` was never checking them. This PR adds both apps to the root `tsconfig` and suppresses existing violations with `// @ts-strict-ignore` so the check passes immediately, leaving individual files as tracked debt to fix properly over time.

### DIFF
--- a/.storybook/tsconfig.json
+++ b/.storybook/tsconfig.json
@@ -1,7 +1,8 @@
 {
   "extends": "../tsconfig",
   "compilerOptions": {
-    "allowSyntheticDefaultImports": true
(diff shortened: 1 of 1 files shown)


coltonhurst:
Approved for autofill desktop concerns, thanks @coroiu. I created a ticket for our team to resolve these (for our files): https://bitwarden.atlassian.net/browse/PM-32443

coltonhurst (approved):

PR #19070, opened 2026-02-18
[SM-1756] Removing feature flag from client side

cd-bitwarden:
## 🎟️ Tracking

https://bitwarden.atlassian.net/jira/software/c/projects/SM/boards/74?assignee=625cb516fd06270069beaf5d&selectedIssue=SM-1756

## 📔 Objective

Removing feature flag from clients side for disableSMAdsForUsers


apps/web/src/app/layouts/product-switcher/shared/product-switcher.service.ts
```diff
@@ -159,16 +159,14 @@ export class ProductSwitcherService {
...
-    this.configService.getFeatureFlag$(FeatureFlag.SM1719_RemoveSecretsManagerAds),
   ]).pipe(
     map(
-      ([orgs, providers, userHasSingleOrgPolicy, paramMap, , removeSecretsManagerAdsFlag]: [
+      ([orgs, providers, userHasSingleOrgPolicy, paramMap]: [
         Organization[],
         Provider[],
         boolean,
         ParamMap,
         void,
```
vleague2, line 169: Should this get removed too now that `paramMap` is the last arg?

cd-bitwarden: I believe void is still required here because it requires 5 parameters to be passed into it, the last being void was a preexisting parameter before I had added and subsequently removed the "removeSecretsManagerAdsFlag"

PR #19071, opened 2026-02-18
[PM-32468] Keeper direct importer

detunized:
## 🎟️ Tracking

N/A

## 📔 Objective

Add a direct importer for Keeper password manager, similar to the existing LastPass direct importer. This allows users to import their vault data directly from Keeper by providing their credentials, without needing to manually export a file first.

Key components:
- **Access module** (`libs/importer/src/importers/keeper/access/`): handles Keeper API authentication (including device approval, 2FA/TOTP/SMS, region redirect, WebSocket push notifications) and vault sync/decryption
- **Importer**: converts decrypted Keeper vault records (logins, cards, identities, secure notes, SSH keys, etc.) into Bitwarden's `ImportResult` format with folder structure preserved
- **UI dialogs**: device approval method selection and code entry, 2FA selection and code entry

## 📸 Screenshots

image
image
image
image
image
image
image
image

### DIFF
--- a/libs/importer/src/importers/keeper/access/vault.ts
+++ b/libs/importer/src/importers/keeper/access/vault.ts
@@ -172,7 +172,7 @@ export class Vault {
         const key = await decryptKeeperKey(folder.sharedFolderKey, folder.keyType, masterKey);
         result.set(uid, key);
       } catch {
-        // TODO: Log this?
+        // TODO: Log decryption errors
       }
     }
     return result;


libs/importer/src/importers/keeper/access/services/crypto.ts
```diff
@@ -0,0 +1,343 @@
...
+  crypto.getRandomValues(arr);
+  return arr;
+}
+
+export function generateEncryptionKey(): Uint8Array {
+  return getRandomBytes(32);
+}
+
+export async function decryptAesV1(data: Uint8Array, key: Uint8Array): Promise<Uint8Array> {
+  const iv = data.subarray(0, AES_BLOCK_SIZE);
```
quexten: @detunized Question: Does keeper really use CBC without any integrity checks or are we skipping some other integrity check (like bitwarden's hmac equivalent) here?

libs/importer/src/importers/keeper/keeper-direct-importer.ts
```diff
@@ -0,0 +1,524 @@
...
+    };
+    if (!keyPair.privateKey) {
+      return false;
+    }
+
+    const passphrase = this.getFirstFieldValue(record, "password") ?? "";
+
+    let keyView: SshKeyView | null = null;
+    try {
+      keyView = import_ssh_key(keyPair.privateKey, passphrase);
```
quexten: @detunized  Does keeper suport ecdsa (p256/p384/p521) keys?

If so, imports containing these specific keys may break until https://github.com/bitwarden/sdk-internal/pull/762 is merged and the compile-time flag is removed.

apps/desktop/project.json
```diff
@@ -91,7 +91,7 @@
     },
     "serve": {
       "executor": "nx:run-commands",
-      "dependsOn": ["build-native"],
+      "dependsOn": ["build-native", "@bitwarden/importer:build"],
```
coroiu: should this be added to the build target too?

detunized: Most likely. I will check other targets now.

PR #19072, opened 2026-02-18
[CL-966] Updated Progress Component

lxiong-livefront:
## 🎟️ Tracking

[JIRA](https://bitwarden.atlassian.net/browse/CL-966?atlOrigin=eyJpIjoiYWZkNjcxNzkyNTcxNGE3ZTk1YTk5NWE3MTU5NGUyMTAiLCJwIjoiaiJ9)

## 📔 Objective

- Updated Progress component to align with new [design](https://www.figma.com/design/Zt3YSeb6E6lebAffrNLa0h/branch/rKUVGKb7Kw3d6YGoQl6Ho7/Tailwind-Component-Library?node-id=36059-11959&m=dev)

- Renamed to 'Progress Bar'

## 📸 Screenshots

Screenshot 2026-03-06 at 5 49 45 PM
Screenshot 2026-03-06 at 5 49 53 PM
Screenshot 2026-03-06 at 5 50 07 PM


libs/assets/src/svg/svgs/spot-report.ts
```diff
@@ -0,0 +1,30 @@
+import { svg } from "../svg";
+
+const SpotReport = svg`
+<svg xmlns="http://www.w3.org/2000/svg" width="120" height="120" viewBox="0 0 120 120" fill="none">
```
vleague2: We remove hardcoded widths and heights for svgs so that they can behave responsively. The container should handle the width/height.

libs/components/src/progress/progress.component.ts
```diff
@@ -1,47 +1,50 @@
...
-
-  get displayText() {
-    return this.showText() && this.size() !== "small";
-  }
+export class ProgressBarComponent {
+  readonly variant = input<BackgroundType>("primary");
+  readonly showLabel = input<boolean>(false);
+  readonly label = input<string>();
+  readonly progressAmount = input<number>(0);
+  readonly showLeftText = input<boolean>(false);
```
vleague2: Prefer to avoid "right" and "left" and use "start" and "end" instead, so that it is more friendly to RTL languages

libs/components/src/progress/progress.component.ts
```diff
@@ -1,47 +1,50 @@
...
+  readonly customLeftText = input<string>();
+  readonly showRightText = input<boolean>(false);
+  readonly rightText = input<string>();
+  readonly showIllustration = input<boolean>(false);
```
vleague2: In the ticket, we last talked about doing 2 components for this -- the progress bar itself with label and start/end text, and then a lockup component that displays the progress bar, graphical element, and title/subtitle text. This way, the progress bar itself can be composed into other layouts easily. If you already tried this let me know!

libs/components/src/progress/progress.component.ts
```diff
@@ -1,47 +1,50 @@
...
-  }
+export class ProgressBarComponent {
+  readonly variant = input<BackgroundType>("primary");
+  readonly showLabel = input<boolean>(false);
+  readonly label = input<string>();
+  readonly progressAmount = input<number>(0);
+  readonly showLeftText = input<boolean>(false);
+  readonly customLeftText = input<string>();
+  readonly showRightText = input<boolean>(false);
+  readonly rightText = input<string>();
```
vleague2: If the `rightText` is optional, do we need a separate property `showRightText`? Can we just check if `rightText` exists for conditional renders? (Same question for the other pairs of properties)

libs/components/src/progress/progress.component.html
```diff
@@ -1,18 +1,44 @@
...
+        @if (showSubtitle()) {
+          <span class="tw-text-fg-body">{{ subtitle() }}</span>
+        }
       </div>
     }
+    <div>
+      @if (showLabel()) {
+        <div class="tw-pe-1">{{ label() }}</div>
+      }
+      <div [ngClass]="outerBarStyles">
```
vleague2: While we're in here, we can migrate from `[ngClass]` to `[class]` per the [modern Angular guidelines](https://contributing.bitwarden.com/contributing/code-style/web/angular-migration-guide#class-and-style-bindings)

libs/components/src/progress/progress.component.html
```diff
@@ -1,18 +1,44 @@
...
+        <div
+          [ngClass]="innerBarStyles"
+          role="progressbar"
+          aria-valuemin="0"
+          aria-valuemax="100"
+          attr.aria-valuenow="{{ progressAmount() }}"
+          [ngStyle]="{ width: progressAmount() + '%' }"
```
vleague2: Same for migrating from `[ngStyle]` to `[style]`

libs/components/src/progress/progress.component.html
```diff
@@ -1,18 +1,44 @@
...
+        ></div>
+      </div>
+      @if (showLeftText() || showRightText()) {
+        <div class="tw-text-sm tw-text-fg-body">
+          @if (showLeftText()) {
+            <span>{{ leftText() }}</span>
+          }
+          @if (showRightText()) {
+            <span class="tw-float-right">{{ rightText() }}</span>
```
vleague2: Can we try using a more modern layout strategy like flexbox instead of floats?

vleague2 (changes requested):
Some quick thoughts on a first look! I also noticed there are build errors related to this component

libs/components/src/progress/progress.component.ts
lxiong-livefront: removed all the `show...` inputs in [75eb198](https://github.com/bitwarden/clients/pull/19072/commits/75eb198ca167024c8491c9b500d0d38d5025028a)

libs/components/src/progress/progress.component.ts
lxiong-livefront: renamed in [fbae875](https://github.com/bitwarden/clients/pull/19072/commits/fbae875a2e553853786db5b982e2c6aa1091eba6)

libs/components/src/progress/progress.component.html
lxiong-livefront: ^ also updated in [c33a952](https://github.com/bitwarden/clients/pull/19072/commits/c33a952d14dcb97cc645c4d9662ad0b6c22b09e7)

libs/components/src/progress/progress.component.html
lxiong-livefront: ^ also updated in [c33a952](https://github.com/bitwarden/clients/pull/19072/commits/c33a952d14dcb97cc645c4d9662ad0b6c22b09e7)

libs/assets/src/svg/svgs/spot-report.ts
lxiong-livefront: removed in [8a37d27](https://github.com/bitwarden/clients/pull/19072/commits/8a37d27a34bf1e026688cbe461d6235ea2ef9c0e)

libs/components/src/progress/progress.component.html
lxiong-livefront: updated in [c33a952](https://github.com/bitwarden/clients/pull/19072/commits/c33a952d14dcb97cc645c4d9662ad0b6c22b09e7)

libs/components/src/progress/progress.component.ts
lxiong-livefront: created the second 'progress bar lockup' component and updated the base component in [78db1b4](https://github.com/bitwarden/clients/pull/19072/commits/78db1b4ef65b6117d6c508043c4500d1b7d4f1b6)

libs/components/src/progress/progress-bar.component.html
```diff
@@ -0,0 +1,15 @@
...
+    role="progressbar"
+    aria-valuemin="0"
+    aria-valuemax="100"
+    attr.aria-valuenow="{{ barWidth() }}"
+    [style]="{ width: barWidth() + '%' }"
+  ></div>
+</div>
+<div class="tw-flex tw-text-sm tw-text-fg-body">
+  <span class="tw-mr-auto">{{ startTextContent() }}</span>
+  <span class="tw-ml-auto">{{ endText() }}</span>
```
vleague2: Can we use margin start and end instead of margin right and left?

libs/components/src/progress/progress-bar-lockup.component.html
```diff
@@ -0,0 +1,20 @@
+<div class="tw-flex tw-w-full tw-gap-8">
+  <div class="tw-flex tw-justify-center tw-items-center">
+    <ng-content select="[slot=illustration]"></ng-content>
+  </div>
+  <div class="tw-flex tw-flex-col tw-justify-center tw-flex-1 tw-gap-6">
+    <div class="tw-flex tw-flex-col tw-gap-1">
+      <span class="tw-text-2xl">{{ title() }}</span>
+      <span class="tw-text-fg-body">{{ subtitle() }}</span>
```
vleague2: Can we use `bitTypography` for these pieces of text?

libs/components/src/progress/progress-bar.component.html
vleague2: Also can we use `bitTypography` here too?

libs/components/src/progress/progress-bar.component.ts
```diff
@@ -0,0 +1,60 @@
...
+  imports: [CommonModule],
+  changeDetection: ChangeDetectionStrategy.OnPush,
+})
+export class ProgressBarComponent {
+  readonly variant = input<BackgroundType>("primary");
+  readonly label = input<string>();
+  readonly barWidth = input<number>(0);
+  readonly showStartText = input<boolean>(true);
+  readonly startText = input<string>();
+  readonly endText = input<string>();
```
vleague2: Could you add jsdoc comments here explaining what each input is for? I think this will be especially helpful given that start and end text are defaulted a little differently

libs/components/src/progress/progress.stories.ts
```diff
@@ -2,56 +2,74 @@ import { Meta, StoryObj } from "@storybook/angular";
...
 export const Base: Story = {
   render: (args) => ({
-    props: args,
+    props: {
+      ...args,
+    },
     template: `
-      <bit-progress ${formatArgsForCodeSnippet<ProgressComponent>(args)}></bit-progress>
+      <div class="tw-w-[552px]">
+        <bit-progress-bar ${formatArgsForCodeSnippet<ProgressBarComponent>(args)} />
```
vleague2: Can we make sure the stories have the accessibility attributes that we're asking engineers to include in the mdx docs? It will provide a better usage example. (I know they weren't there before, but this is a good chance to add them.)

libs/components/src/progress/progress-bar.stories.ts
vleague2, line 1: Should we rename to `progress-bar.stories.ts`?

libs/components/src/progress/progress.stories.ts
```diff
@@ -2,56 +2,74 @@ import { Meta, StoryObj } from "@storybook/angular";
 
 import { formatArgsForCodeSnippet } from "../../../../.storybook/format-args-for-code-snippet";
 
-import { ProgressComponent } from "./progress.component";
+import { ProgressBarComponent } from "./progress-bar.component";
 
 export default {
-  title: "Component Library/Progress",
-  component: ProgressComponent,
+  title: "Component Library/Progress Bar",
```
vleague2: Thoughts on making a group for the progress components? Kind of like how we do the skeleton and forms. It could be `/Progress/Progress Bar` and `/Progress/Progress Bar Lockup` ?

libs/components/src/progress/progress-bar.component.html
lxiong-livefront: addressed in [b2ccafd](https://github.com/bitwarden/clients/pull/19072/commits/b2ccafd15728f72b1e6736fd22b5167de3e2173d)

libs/components/src/progress/progress-bar-lockup.component.html
lxiong-livefront: ^

libs/components/src/progress/progress-bar.component.ts
lxiong-livefront: added jsdoc comments in [afe9d5d](https://github.com/bitwarden/clients/pull/19072/commits/afe9d5dd0a1574747ff4edb16ee848e09427e2e9)

let me know if any of them could be adjusted/made more clear :)

libs/components/src/progress/progress.stories.ts
lxiong-livefront: added accessibility attributes in [ab2356c](https://github.com/bitwarden/clients/pull/19072/commits/ab2356c026712f209158b2a0c2e138d2d4f81694)

libs/components/src/progress/progress-bar.stories.ts
lxiong-livefront: renamed in [924c60d](https://github.com/bitwarden/clients/pull/19072/commits/924c60d929ccac6751eed6bd26b51d420e1e87ef)

libs/components/src/progress/progress.stories.ts
lxiong-livefront: Yes! I agree that it's much better to group these two - [9d5c23b](https://github.com/bitwarden/clients/pull/19072/commits/9d5c23b6b5c3e2d21cc7b715cc6de34d1afe9bd0)

libs/components/src/progress/progress-bar.component.html
```diff
@@ -0,0 +1,15 @@
+<div class="!tw-mb-0" bitTypography="h5">{{ label() }}</div>
+<div [class]="outerBarStyles">
+  <div
+    [class]="innerBarStyles"
```
vleague2: I noticed the design file has the end of the progress fill rounded, can we make that style change?

libs/components/src/progress/progress-bar.component.ts
```diff
@@ -0,0 +1,68 @@
...
+  readonly startText = input<string>();
+  /* The ending helper text displayed below the progress bar */
+  readonly endText = input<string>();
+
+  protected readonly startTextContent = computed(() => {
+    if (!this.showStartText()) {
+      return undefined;
+    }
+
+    return this.startText() || `${this.barWidth()}% complete`;
```
vleague2: This text should get localized

libs/components/src/progress/progress-bar-lockup.component.html
```diff
@@ -0,0 +1,20 @@
...
+  <div class="tw-flex tw-justify-center tw-items-center">
+    <ng-content select="[slot=illustration]"></ng-content>
+  </div>
+  <div class="tw-flex tw-flex-col tw-justify-center tw-flex-1 tw-gap-6">
+    <div class="tw-flex tw-flex-col tw-gap-1">
+      <span bitTypography="h2">{{ title() }}</span>
+      <span class="tw-text-fg-body" bitTypography="body1">{{ subtitle() }}</span>
+    </div>
+    <div>
+      <bit-progress-bar
```
vleague2: I am realizing that we have to pass quite a lot of inputs through the lockup to the progress bar. What do you think about taking a different approach here by allowing the consumer to pass their own `bit-progress-bar` to the component and doing a `ng-content select="bit-progress-bar"` instead? 

Example of consumer code:
```
<bit-progress-bar-lockup ${formatArgsForCodeSnippet<ProgressBarLockupComponent>(args)}>
  <ng-container slot="illustration">
    <bit-svg [content]="illustration" aria-hidden="true" class="tw-w-[120px]" />
  </ng-container>
  <bit-progress-bar label="whatever" startText="whatever" etc></bit-progress-bar>
</bit-progress-bar-lockup>
```

You can check the switch component for an example of using ng-content to select a component instead of a slot.

Then the progress bar lockup only needs to accept inputs for itself, since the consumer would be using their own progress bar and passing it inputs.

libs/components/src/progress/progress-bar.component.ts
vleague2: Only note is that for `endText`, maybe we can add "if nothing is passed, no text is displayed in the end slot"

libs/components/src/progress/progress-bar.mdx
```diff
@@ -0,0 +1,80 @@
...
+<Canvas of={stories.StartText} />
+
+### Strength indicator styles
+
+For a strength indicator use the following styles for fill:
+
+- **Weak:** `danger-600`
+- **Weak2:** `warning-600`
+- **Good:** `primary-600`
+- **Strong:** `success-600`
```
vleague2: Consumers wouldn't be passing the color tokens directly, so this might confuse people. I think we can specify the variant here instead, like "for weak passwords use variant `danger`"

libs/components/src/progress/progress-bar-lockup.mdx
```diff
@@ -0,0 +1,36 @@
...
+- Leading illustration is optional.
+  - Illustration `svg` component can be supplied using the `illustration` slot.
+- Title and Subtitle are optional.
+  - If using a Title, the copy should be written verb first: “Loading” or “Loading [object]”,
+    “Uploading [object]”
+  - If using descriptive subtitle, should be as succinct as possible, telling the user what is
+    happening, how long it will take, why it is happening, or any actions they should consider in
+    the meantime.
+
+## Accessibility
```
vleague2: If we make the change to pass in the progress bar, we can delete this section from the progress-bar-lockup docs.

libs/components/src/progress/progress-bar.mdx
```diff
@@ -0,0 +1,80 @@
...
+## Content Guidelines
+
+- In upload scenarios, file size is placed at the end of the bar. You don’t need to spell out the
+  size labels; write as MB or GB.
+
+## Accessibility
+
+Be sure to include the proper `aria-valuemin`, `aria-valuemax`, and `aria-valuenow` attributes. An
+`aria-valuetext` should also be configurable to include the text a screen reader should read to the
+user.
```
willmartian: We set these within our template, and also ask devs to set them on the host element? That seems off?

Should we instead add required inputs for these?

cc @vleague2

libs/components/src/progress/progress-bar.component.ts
```diff
@@ -0,0 +1,68 @@
...
+  /* Determines the color of the progress bar */
+  readonly variant = input<BackgroundType>("primary");
+  /* The label displayed above the progress bar */
+  readonly label = input<string>();
+  /* The progress amount, represented as a percentage of the progress bar that is filled */
+  readonly barWidth = input<number>(0);
+  /* Whether to show the starting helper text below the progress bar. Defaults to true */
+  readonly showStartText = input<boolean>(true);
+  /* The starting helper text displayed below the progress bar. Defaults to "<barWidth>% complete" */
+  readonly startText = input<string>();
```
willmartian: Do you think it is clearer to have two inputs, or to allow passing `null` to `startText` to hide it? I lean towards the latter but I sometimes try to over condense

libs/components/src/progress/progress-bar-lockup.component.html
willmartian: +1 to this change request

libs/components/src/progress/progress-bar-lockup.component.html
```diff
@@ -0,0 +1,20 @@
...
+        [variant]="variant()"
+        [label]="label()"
+        [barWidth]="barWidth()"
```
willmartian: Since we are already diffing all usage sites, do we also want to change `barWidth` to `value`, to be more in line with the vanilla HTML progress element? Which do you find more clear?

libs/components/src/progress/progress-bar-lockup.component.html
```diff
@@ -0,0 +1,20 @@
+<div class="tw-flex tw-w-full tw-gap-8">
+  <div class="tw-flex tw-justify-center tw-items-center">
+    <ng-content select="[slot=illustration]"></ng-content>
+  </div>
+  <div class="tw-flex tw-flex-col tw-justify-center tw-flex-1 tw-gap-6">
+    <div class="tw-flex tw-flex-col tw-gap-1">
+      <span bitTypography="h2">{{ title() }}</span>
+      <span class="tw-text-fg-body" bitTypography="body1">{{ subtitle() }}</span>
```
willmartian: For the spinner lockup, we are using slots over inputs. https://github.com/bitwarden/clients/pull/19017#discussion_r2870756859

We probably want to do the same in both components for API symmetry. 

I lean towards slots because it allows us to pass arbitrary content in the future, if design ever wants that.

libs/components/src/progress/progress-bar.component.ts
lxiong-livefront: For sure! Added in [1cf809a](https://github.com/bitwarden/clients/pull/19072/commits/1cf809aa1f43782beb33f647cecd62163578dc70)

libs/components/src/progress/progress-bar.component.html
lxiong-livefront: Updated in [1c0d5df](https://github.com/bitwarden/clients/pull/19072/commits/1c0d5dfa607d580ca8554402ff5ac6732a7e39ca)

libs/components/src/progress/progress-bar.component.ts
lxiong-livefront: Localized in [4b6d885](https://github.com/bitwarden/clients/pull/19072/commits/4b6d885922eb36a73bb0410f03db1e66e836703e)

libs/components/src/progress/progress-bar-lockup.component.html
lxiong-livefront: I think this is definitely a better approach! Updated in [342b111](https://github.com/bitwarden/clients/pull/19072/commits/342b1113198acd0a95b21ac1b4ce8c04b108987c)

libs/components/src/progress/progress-bar.mdx
lxiong-livefront: Updated in [7caaaf2](https://github.com/bitwarden/clients/pull/19072/commits/7caaaf22d9edefb0d1b011da493b4bac7149095d)

libs/components/src/progress/progress-bar-lockup.mdx
lxiong-livefront: ^ removed in [342b111](https://github.com/bitwarden/clients/pull/19072/commits/342b1113198acd0a95b21ac1b4ce8c04b108987c)

libs/components/src/progress/progress-bar.mdx
vleague2: Oh yes, sorry, I misunderstood this in my last review. I think these accessibility guidelines came from a design document but are not needed for developer documentation since we are setting the properties in the component itself. We can remove these aria value guidelines from the mdx page and the properties from the stories, sorry @lxiong-livefront for the bad guidance from me!

libs/components/src/progress/progress-bar.component.ts
vleague2: I usually lean towards the latter as well so that you're only having to pass 1 thing to affect 1 element. I think where I'm struggling here is that the `endText` input doesn't behave the same way, and like half of the existing usages hide the start text and already are using the `showStartText` input to hide it. But I think I'm good either way

lxiong-livefront: I like not having a separate `showStartText` input! Updated in [16ff5c1](https://github.com/bitwarden/clients/pull/19072/commits/16ff5c1659479151e983d91d44851848b7ae922d) and in consumers in [7baebe3](https://github.com/bitwarden/clients/pull/19072/commits/7baebe321572013580d47807b01ba9dab6d98b1e)

libs/components/src/progress/progress-bar-lockup.component.html
lxiong-livefront: I think `value` is more clear for its intended usage! Updated in [7baebe3](https://github.com/bitwarden/clients/pull/19072/commits/7baebe321572013580d47807b01ba9dab6d98b1e)

libs/components/src/progress/progress-bar-lockup.component.html
lxiong-livefront: Updated to slots in [394fe1e](https://github.com/bitwarden/clients/pull/19072/commits/394fe1eacdd62759de4870f30a0facf29c0e0af6)

I kept the `bitTypography` styles, but let me know if we also want the consumer to handle that as well

libs/components/src/progress/progress-bar.mdx
lxiong-livefront: No worries! I can remove these from mdx and stories :) 

I noticed that per the accessibility guidelines, we don't currently have an `aria-valuetext` set in our template. Do we want to create an input for this or set ourselves? Also unsure if/how we'd want to handle an `aria-label` since the `label` input is optional, or if we want to set it to something like "progress bar" (I don't see that currently set in any of the locales)

vleague2: Hmm good point. I think adding an optional input for it makes sense, because it looks like it's meant to be an override of the default accessible text that gets read out from the existing aria properties. I think we probably will not be using it often since our progress bars are measuring numerical values most of the time, but I could see it being useful for something like the password strength indicator. Consumers could pass in whatever localized string they want, since they are also passing in the `value` and will know what text should be read out for each value.

amorask-bitwarden (dismissed):

libs/components/src/progress/progress-bar.mdx
lxiong-livefront: Added a `ariaValueText` input and ids for the progress bar as well as the 'label' for `aria-labelledby` to point to the progress bar. Progress bar `label` will be made required in another ticket to avoid missing `aria-label` error that currently exists for any `bit-progress-bar` without a label.  cc @willmartian 

[e0b23d8](https://github.com/bitwarden/clients/pull/19072/commits/e0b23d85e4383f77a5e76091b441c8fb874541da)

mcamirault (dismissed):

libs/components/src/progress/progress-bar.component.ts
```diff
@@ -39,9 +44,16 @@ export class ProgressBarComponent {
   readonly startText = input<string | null>();
   /* The ending helper text displayed below the progress bar. If nothing is passed, no text is displayed in the end slot */
   readonly endText = input<string>();
+  /* The ARIA value text for the progress bar. Overrides default accessible text. */
+  readonly ariaValueText = input<string>();
+  /* The ID of the progress bar element, used for attaching `aria-describedby` attributes. */
+  readonly progressBarId = input<string>(`bit-progress-bar-${this.id}`);
```
vleague2, line 47: Any particular reason to have this as an input instead of a readonly property? Also, I don't see the `aria-describedby` attr being used, do we need this?

Banrion (dismissed):
DIRT files look good

libs/components/src/progress/progress-bar.component.ts
lxiong-livefront: I added this as an optional input in case an `aria-describedby` needs to be attached to the password field and pointed to the associated progress bar, as documented in the pre-existing Accessibility section. I could be overlooking something though, since an `id` wasn't originally set for the progress bar 🤔

vleague2: Ooh okay I am following now haha. Interestingly, our current password strength component usage does not do any association with the input fields, but it probably should. Let's leave what you have so that we can improve the password strength component!

vleague2:
I'm seeing more build errors related to `showText` on `bit-progress` usages

lxiong-livefront:
> I'm seeing more build errors related to `showText` on `bit-progress` usages

@vleague2 Ah that might've snuck in after a merge - fixed/updated in [7c849b0](https://github.com/bitwarden/clients/pull/19072/commits/7c849b0beca62b6e2957692b8b3007ace4d37a32)

JimmyVo16 (dismissed):

vleague2:
I don't know why Chromatic has UI diffs for some components you didn't touch (probably a quirk of the branching we're doing), but there are some stories (like the kitchen sink) that have errors related to translations needed for the progress bar. Can you take a look through the UI tests and look for other stories that seem really off and fix any progress-related issues?

lxiong-livefront:
> I don't know why Chromatic has UI diffs for some components you didn't touch (probably a quirk of the branching we're doing), but there are some stories (like the kitchen sink) that have errors related to translations needed for the progress bar. Can you take a look through the UI tests and look for other stories that seem really off and fix any progress-related issues?

@vleague2 Hmm, I'm not seeing any errors in stories/Chromatic UI Tests and did another global search in '/libs` for `bit-progress`, but all seem to be consuming the new progress bar correctly. Can you show me where you are seeing any errors or where stories seem off? 

**Update** : Still not seeing any explicit errors, but I added a translation in [825ac17](https://github.com/bitwarden/clients/pull/19072/commits/825ac17244db8f3f05798e549051111ee5f3e25b) which seems to have fixed some dialogs not displaying correctly in the kitchen sink. After looking more closely at other stories and anywhere else 'bit-progress-bar' is consumed, I noticed that the original default progress component wouldn't display any `startText` if the progress value is `0`. This was causing the `startText` to appear briefly on load in the 'Input Password' stories. Updated to reflect this behavior in [c4cd28f](https://github.com/bitwarden/clients/pull/19072/commits/c4cd28f457d4cbc35fbdb2f69cbde6ee6d4a5d47). 

Hopefully this covers everything that needed fixing, but let me know if you notice anything else I may have missed :)

libs/components/src/progress/progress-bar.component.ts
```diff
@@ -55,7 +55,7 @@ export class ProgressBarComponent {
   protected readonly labelId = `bit-progress-bar-label-${this.id}`;
 
   protected readonly startTextContent = computed(() => {
-    if (this.startText() === null) {
+    if (this.startText() === null || this.value() === 0) {
```
vleague2: Let's double-check this with design to see if they want to keep this behavior of the start text not showing when the value is at 0%

apps/web/src/app/admin-console/organizations/members/components/bulk/bulk-progress-dialog.component.html
```diff
@@ -2,12 +2,12 @@
...
-          [showText]="false"
+        <bit-progress-bar
+          [startText]="null"
           size="default"
           bgColor="primary"
-          [barWidth]="progressPercentage()"
-        ></bit-progress>
+          [value]="progressPercentage()"
+        ></bit-progress-bar>
       </div>
```
willmartian: I think we should use slots over inputs to make this more flexible. The design of this component is very similar to the form fields / form controls, which commonly pass non-text/interactive content to the label:

```html
<bit-form-field>
  <bit-label>
    Label
    <a href="#" slot="end" bitLink [appA11yTitle]="'More info'">
      <i class="bwi bwi-question-circle" aria-hidden="true"></i>
    </a>
  </bit-label>
  <input bitInput placeholder="Placeholder" />
</bit-form-field>
```

I think it would be beneficial to adopt a similar API here:

```html
<bit-progress-bar>
    <bit-label>Foo</bit-label>
    <bit-hint slot="start">Bar</bit-hint>
    <bit-hint slot="end">Baz</bit-hint>
</bit-progress-bar>
```

This also sets us up to support complex hints used in the password strength indicator.

libs/components/src/progress/progress-bar.component.ts
```diff
@@ -0,0 +1,72 @@
+import { CommonModule } from "@angular/common";
+import { Component, input, ChangeDetectionStrategy, computed, inject } from "@angular/core";
+
+import { I18nService } from "@bitwarden/common/platform/abstractions/i18n.service";
+
+import { TypographyModule } from "../typography";
+
+export type BackgroundType = "primary" | "subtle" | "success" | "warning" | "danger";
```
willmartian: This type should be renamed to `ProgressBarVariant` or similar

libs/components/src/progress/progress-bar.component.ts
willmartian: Should `aria-describedby` reference the internal progress bar, or the entire component?

libs/components/src/progress/progress-bar.component.ts
```diff
@@ -0,0 +1,72 @@
...
+    return this.startText() || this.i18nService.t("percentageCompleted", this.value().toString());
+  });
+
+  get outerBarStyles() {
+    return ["tw-overflow-hidden", "tw-rounded", "tw-bg-secondary-100", "tw-h-2", "tw-my-1"];
+  }
+
+  get innerBarStyles() {
+    return ["tw-transition-all", "tw-h-2", "tw-rounded"].concat(BackgroundClasses[this.variant()]);
+  }
```
willmartian: Prefer computed signals over `get`

libs/components/src/progress/progress-bar.component.ts
```diff
@@ -0,0 +1,72 @@
...
+    if (this.startText() === null || this.value() === 0) {
+      return undefined;
+    }
+
+    return this.startText() || this.i18nService.t("percentageCompleted", this.value().toString());
+  });
+
+  get outerBarStyles() {
+    return ["tw-overflow-hidden", "tw-rounded", "tw-bg-secondary-100", "tw-h-2", "tw-my-1"];
+  }
```
willmartian: If this doesn't change, it should use a `readonly` property (instead of `get`), or be inlined in the template.

libs/components/src/progress/progress-bar.component.html
```diff
@@ -0,0 +1,20 @@
...
+    <div
+      [id]="progressBarId()"
+      [class]="innerBarStyles"
+      role="progressbar"
+      aria-valuemin="0"
+      aria-valuemax="100"
+      [attr.aria-valuenow]="value()"
+      [attr.aria-valuetext]="ariaValueText()"
+      [attr.aria-labelledby]="labelId"
+      [style]="{ width: value() + '%' }"
```
willmartian: `[style.width]` is a bit nicer here (nit pick 😄)

willmartian (dismissed):

apps/web/src/app/admin-console/organizations/members/components/bulk/bulk-progress-dialog.component.html
willmartian: I also notice we left a `bgColor` in the template here. Could you double check that all of those have been migrated to `variant`?

libs/components/src/progress/progress-bar.component.ts
lxiong-livefront: The 🤖 explanation:
```
The id should be set on the bar itself (the inner role="progressbar" div), which is the current implementation.

Here's the reasoning:

aria-describedby is meant to point to an element that describes the referenced element. When used externally (e.g., a button that says "Upload" with an aria-describedby pointing to a progress bar), the reference should target the role="progressbar" element — because that's what screen readers will announce as the progress indicator.
The outer wrapper <div> has no semantic role and provides no useful accessible information on its own.
The ARIA spec states that aria-describedby should reference elements whose content supplements the accessible name of the referencing element. The progressbar role element with its aria-valuenow, aria-valuetext, etc. is exactly that.
```

apps/web/src/app/admin-console/organizations/members/components/bulk/bulk-progress-dialog.component.html
lxiong-livefront: Updated to use slots  and changed/checked `bgColor` to be `variant` in [afc2cc4](https://github.com/bitwarden/clients/pull/19072/commits/afc2cc4acdea50f5e67e9af8d5b72f3edcc55e7e) and used `bit-hint` as the default `startHint` in [7c886ab](https://github.com/bitwarden/clients/pull/19072/commits/7c886ab30a9617586e498424f777124c792ddb97)

libs/components/src/progress/progress-bar.component.ts
lxiong-livefront: Good call! Checked in with Emily and she said to keep this displayed even if value is 0, since it's no longer displayed inside the bar itself

[24b02bf](https://github.com/bitwarden/clients/pull/19072/commits/24b02bffe2e4398100fbeebc788497ebb1e5b48e)

libs/components/src/progress/progress-bar-lockup.mdx
```diff
@@ -0,0 +1,27 @@
...
+
+<Title />
+<Description />
+
+<Primary />
+<Controls />
+
+## Content Guidelines
+
+- Leading illustration is optional.
```
vleague2, line 19: Maybe we should add a story for this without the illustration?

libs/angular/src/tools/password-strength/password-strength-v2.component.html
```diff
@@ -1,10 +1,6 @@
-<bit-progress
-  [size]="size"
-  [text]="text"
-  [bgColor]="color"
-  [showText]="showText"
-  [barWidth]="scoreWidth"
-></bit-progress>
+<bit-progress-bar [variant]="color" [hideStartHint]="!showText" [value]="scoreWidth">
+  <bit-hint slot="start">{{ text }}</bit-hint>
```
vleague2, line 2: Should we be putting the `bit-hint` wrapper in the component, or are we allowing for more flexibility for the start elements?

lxiong-livefront: I assumed we want to allow for more flexibility and have consumers pass in `bit-hint` as a guideline, aligning with the new API Will suggested [here](https://github.com/bitwarden/clients/pull/19072#discussion_r2919046877). @willmartian Any thoughts on this?

libs/components/src/progress/progress-bar-lockup.mdx
lxiong-livefront: added story in [5724a8e](https://github.com/bitwarden/clients/pull/19072/commits/5724a8ee6fdd310a858958a09b266422814e5215)

libs/components/src/progress/progress.module.ts
```diff
@@ -1,9 +1,9 @@
 import { NgModule } from "@angular/core";
 
-import { ProgressComponent } from "./progress.component";
+import { ProgressBarComponent } from "./progress-bar.component";
 
 @NgModule({
-  imports: [ProgressComponent],
-  exports: [ProgressComponent],
+  imports: [ProgressBarComponent],
+  exports: [ProgressBarComponent],
```
lxiong-livefront: @vleague2 @willmartian Do we want to include the 'Progress Bar Lockup' in this progress module as well?

vleague2: I would like to remove the module entirely since we don't really use them anymore (we prefer to import/export directly), but I think that would mean more consumer code changes. I think for now we can add it to the module to keep things simple so that at least it is usable from somewhere

lxiong-livefront: That was my original thought as well, but since we're already updating all the consumer files, I opted to just remove the progress module and have separate imports for `progress-bar` and `progress-bar-lockup`. (only added 5 more diffed files with this change 😅) [47e3219](https://github.com/bitwarden/clients/pull/19072/commits/47e3219a070c7db91151212155378a534552b639)

libs/components/src/progress/progress-bar-lockup.mdx
```diff
@@ -0,0 +1,31 @@
+import { Meta, Canvas, Primary, Controls, Title, Description } from "@storybook/addon-docs/blocks";
+
+import * as stories from "./progress-bar-lockup.stories";
+
+<Meta of={stories} />
+
+```ts
+import { ProgressBarLockupComponent } from "./progress-bar-lockup.component";
```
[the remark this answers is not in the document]
lxiong-livefront: [471e20b](https://github.com/bitwarden/clients/pull/19072/commits/471e20b28a924c1d5db6101b63f74bbf0f759b8e)

bitwarden_license/bit-web/src/app/dirt/access-intelligence/shared/report-loading.component.html
```diff
@@ -2,12 +2,7 @@
   <div class="tw-flex tw-flex-col tw-items-center tw-gap-4">
     <!-- Progress bar -->
     <div class="tw-w-64" role="progressbar" attr.aria-label="{{ 'loadingProgress' | i18n }}">
```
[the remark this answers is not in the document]
lxiong-livefront: [1201408](https://github.com/bitwarden/clients/pull/19072/commits/1201408852d5d07e7e21a3ecc3e5dd0507b47368)

libs/components/src/progress/progress-bar.component.ts
lxiong-livefront: Addressed this batch of comments in [920f0e5](https://github.com/bitwarden/clients/pull/19072/commits/920f0e5e23eab0d2949187b3ca4068a2614dd52a)

libs/components/src/progress/progress-bar.component.ts
lxiong-livefront: ^[920f0e5](https://github.com/bitwarden/clients/pull/19072/commits/920f0e5e23eab0d2949187b3ca4068a2614dd52a)

libs/components/src/progress/progress-bar.component.ts
lxiong-livefront: ^[920f0e5](https://github.com/bitwarden/clients/pull/19072/commits/920f0e5e23eab0d2949187b3ca4068a2614dd52a)

libs/components/src/progress/progress-bar.component.html
lxiong-livefront: ^[920f0e5](https://github.com/bitwarden/clients/pull/19072/commits/920f0e5e23eab0d2949187b3ca4068a2614dd52a)

libs/components/src/progress/progress-bar.mdx
```diff
@@ -0,0 +1,78 @@
...
+the indicator to those who may be unable to see part or all of the indicator. Helper hints may be
+passed in with the `start` and `end` slots.
+
+<Canvas of={stories.WithLabelAndHelperHint} />
+
+## Start Hint
+
+When measuring something other than progress, such as password strength, pass in a `bit-hint` to the
+`start` slot that fits the context of the implementation. If nothing is passed into the `start`
+slot, it will default to `[value]% complete` while `value` is greater than `0`. Set `hideStartHint`
```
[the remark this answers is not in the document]
lxiong-livefront: [5f477f9](https://github.com/bitwarden/clients/pull/19072/commits/5f477f951ce070e3797d103cf21381ac9ff8b174)

apps/web/src/app/billing/individual/user-subscription.component.html
```diff
@@ -134,7 +134,7 @@ <h3 bitTypography="h3" class="tw-mt-8">{{ "storage" | i18n }}</h3>
     <p bitTypography="body1">
       {{ "subscriptionStorage" | i18n: sub.maxStorageGb || 0 : sub.storageName || "0 MB" }}
     </p>
-    <bit-progress [barWidth]="storagePercentage" bgColor="success" size="default"></bit-progress>
+    <bit-progress-bar [value]="storagePercentage" variant="success" />
```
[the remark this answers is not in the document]
lxiong-livefront: [0eb6b37](https://github.com/bitwarden/clients/pull/19072/commits/0eb6b372b8b37269fc15e76012ee5ef93ee3a506)

apps/web/src/app/billing/organizations/organization-subscription-cloud.component.html
```diff
@@ -160,7 +160,7 @@ <h4 bitTypography="h4" class="tw-mt-9">{{ "storage" | i18n }}</h4>
         <p bitTypography="body1">
           {{ "subscriptionStorage" | i18n: sub.maxStorageGb || 0 : sub.storageName || "0 MB" }}
         </p>
-        <bit-progress [barWidth]="storagePercentage" bgColor="success"></bit-progress>
+        <bit-progress-bar [value]="storagePercentage" variant="success" />
```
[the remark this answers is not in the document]
lxiong-livefront: [0eb6b37](https://github.com/bitwarden/clients/pull/19072/commits/0eb6b372b8b37269fc15e76012ee5ef93ee3a506)

libs/components/src/progress/progress-bar.component.ts
```diff
@@ -0,0 +1,69 @@
...
+})
+export class ProgressBarComponent {
+  private readonly id = nextId++;
+
+  /** Hides the default start hint text. Defaults to false. */
+  readonly hideStartHint = input<boolean>(false);
+  /** Determines the color of the progress bar */
+  readonly variant = input<ProgressBarVariant>("primary");
+  /** The progress amount, represented as a percentage of the progress bar that is filled */
+  readonly value = input<number>(0);
```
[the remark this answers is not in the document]
lxiong-livefront: [032dd78](https://github.com/bitwarden/clients/pull/19072/commits/032dd7870bc8c13680b6f584ebff1b367bc456d0)

libs/subscription/src/components/storage-card/storage-card.component.ts
```diff
@@ -47,7 +54,7 @@ export class StorageCardComponent {
     if (storage.available === 0) {
       return 0;
     }
-    return Math.min((storage.used / storage.available) * 100, 100);
+    return (storage.used / storage.available) * 100;
```
vleague2, line 57: This is causing a test to fail. I think the team had this set up to only return 100 as the max on purpose

lxiong-livefront: I clamped the value between 0 and 100 in the progress bar in [032dd78](https://github.com/bitwarden/clients/pull/19072/commits/032dd7870bc8c13680b6f584ebff1b367bc456d0) which should achieve the same 🤔

lxiong-livefront: Removed the test since progress bar is clamping the values [a865eda](https://github.com/bitwarden/clients/pull/19072/commits/a865eda140b9a09e1571e50a5bc6481d11b93f39)

libs/angular/src/tools/password-strength/password-strength-v2.component.html
vleague2: For posterity, we decided to leave as is for the flexibility

libs/components/src/progress/progress-bar-lockup.component.ts
```diff
@@ -0,0 +1,30 @@
...
+  ngAfterViewInit() {
+    if (!this.progressBar()) {
+      // This is only here so Angular throws a compilation error if no progress bar is provided.
+      // the `this.progressBar()` value must try to be accessed for the required content child check to throw
+      // eslint-disable-next-line no-console
+      console.error(
+        "No progress bar component provided. <bit-progress-bar-lockup> must be used with a <bit-progress-bar>.",
+      );
+    }
+  }
```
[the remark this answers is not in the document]
vleague2: What do you think about this suggestion to simplify this check? (eta since Github is not helpful -- this is referencing a Claude comment)

libs/components/src/progress/progress-bar.component.ts
```diff
@@ -0,0 +1,71 @@
...
+  // Necessary for `aria-labelledby` to point to the label element
+  protected readonly labelId = `bit-progress-bar-label-${this.id}`;
+
+  protected readonly clampedValue = computed(() => Math.max(0, Math.min(100, this.value())));
+
+  protected readonly defaultStartText = computed(() => {
+    return this.i18nService.t("percentageCompleted", this.clampedValue().toString());
+  });
+
+  protected readonly outerBarStyles = [
```
vleague2: Nit: since these aren't dynamic, any reason to keep them here instead of moving them into the template? (To answer my own question -- if we are going to be moving to `tailwind-variants` soon, it'll end up in the ts file anyway, so maybe it doesn't matter at this stage)

lxiong-livefront: These were already here from the previous file, but that included a dynamic 'size' style. I'm ok with moving it into the template for now, even if it gets moved back at a later stage!

lxiong-livefront: moved in [da97d23](https://github.com/bitwarden/clients/pull/19072/commits/da97d2390c6d5c2c649dbc8c4b948b86027ae03c)

libs/components/src/progress/progress-bar-lockup.component.ts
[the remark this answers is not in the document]
lxiong-livefront: Updated to this suggestion in [7efceb5](https://github.com/bitwarden/clients/pull/19072/commits/7efceb54aa6225e2fb7de8618252c16cfb6649ec)

libs/angular/src/tools/password-strength/password-strength-v2.component.html
vleague2, line 2: I think this might be causing some extra whitespace in usages of this component? I haven't inspected deeply but it might be that `text` is empty even though `showText` is true?

Image

lxiong-livefront: Thanks for catching this! Fixed in [548a9a2](https://github.com/bitwarden/clients/pull/19072/commits/548a9a2f0a7d9fe6cf95c47c1faf854dfc0588e9)

vleague2 (dismissed):
Now that this is targeting `main` I added the QA tag! But things look good to me, thanks for your patience with all the iterations!

vleague2 (dismissed):

JimmyVo16 (dismissed):

mcamirault (dismissed):

amorask-bitwarden (dismissed):

dereknance (dismissed):

Banrion (dismissed):

shane-melton (dismissed):

vleague2 (approved):

amorask-bitwarden (approved):

dereknance (approved):

JimmyVo16 (approved):

Banrion (approved):

mcamirault (approved):

shane-melton (approved):

PR #19076, opened 2026-02-18
Extract urlOriginsMatch utility and refactor senderIsInternal

audreyality:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-3843

## 📔 Objective

This PR extracts `urlOriginsMatch` from `senderIsInternal` so that the logic can be reused to detect when the Bitwarden extension is opening its own context menu.

## 📓 Implementation Notes

Adds `urlOriginsMatch` to `@bitwarden/platform`, which compares two URLs by scheme, host, and port. Uses `protocol + "//" + host` rather than `URL.origin` because non-special schemes (e.g. chrome-extension://) return the opaque string `"null"` from `.origin`, making equality comparison unreliable. URLs without a host (file:, data:) are explicitly rejected to prevent hostless schemes from comparing equal.

Refactors `senderIsInternal` to delegate to `urlOriginsMatch` and to derive the extension URL via `BrowserApi.getRuntimeURL("")` rather than inline chrome/browser API detection. Adds full test coverage for `senderIsInternal`.

The previous string-based comparison used `startsWith` after stripping trailing slashes, which was safe in `senderIsInternal` where inputs are tightly constrained. As a general utility accepting arbitrary URLs, `startsWith` can produce false positives (e.g. "https://example.com" matching "https://example.com.evil.com"). Structural host comparison is the correct contract for unrestricted input.

The extracted method was added in the platform module so that it wouldn't clutter up `common`.


libs/platform/src/util.ts
```diff
@@ -0,0 +1,53 @@
...
+ * - Either argument cannot be parsed as a valid URL
+ * - Either URL has no host (e.g. `file:`, `data:` schemes), since URLs without
+ *   a meaningful host cannot be distinguished by origin
+ *
+ * @param canonical - The reference URL whose origin acts as the baseline.
+ * @param suspect - The URL being tested against the canonical origin.
+ * @returns `true` if both URLs share the same scheme, host, and port; `false` otherwise.
+ */
+export function urlOriginsMatch(canonical: string | URL, suspect: string | URL): boolean {
+  function toURL(input: string | URL): URL | null {
```
coroiu: suggestion(non-blocking): consider extracting these from the function so they don't have to get recreated on every call to this function?

coroiu (approved):
Looks good, just one small non-blocking comment (feel free to ignore)

libs/platform/src/util.ts
audreyality: TIL that JavaScript recreates nested functions even when they don't close over any variables.

Addressed in ac681a6549b65.

PR #19078, opened 2026-02-19
[PM-27772] Add TrendWidget component with Chart.js implementation

lastbestdev:
## 🎟️ Tracking
https://bitwarden.atlassian.net/browse/PM-27772

To view component: `npm run storybook`

## 📔 Objective

Add reusable TrendWidget component with Chart.js line chart for displaying risk trends over time in Access Intelligence.

### Components
- **LineChartComponent** (`dirt/shared/`) - Reusable Chart.js wrapper with reactive updates using Angular signals
- **TrendWidgetComponent** - Displays dual-line chart (at risk vs total) with loading/error states

### Features
- Configurable datetime/linear x-axis scales
- Theme-aware colors from design system (brand-700/400, gray-200/600)
- Full i18n support (10 new translation keys)
- Loading spinner and error state handling
- Proper Chart.js lifecycle management

### Testing
- 20 unit tests covering chart display, loading, error states, i18n labels, and theme colors
- Canvas context mocked for jsdom environment

## 📸 Screenshots

**With data**

image

**Loading**

image

**Error**

image

**Dark mode**

image

### DIFF
--- a/bitwarden_license/bit-web/src/app/dirt/access-intelligence/activity/trend-widget/trend-widget.stories.ts
+++ b/bitwarden_license/bit-web/src/app/dirt/access-intelligence/activity/trend-widget/trend-widget.stories.ts
@@ -186,6 +186,24 @@ export const SingleDataPoint: Story = {
 };
 
 /**
+ * Two data points.
+ */
+export const TwoDataPoints: Story = {
+  args: {
+    data: {
+      timeframe: "past-month",
+      dataView: "passwords",
+      dataPoints: [
+        { timestamp: "2026-02-18T00:00:00Z", atRisk: 45, total: 180 },
+        { timestamp: "2026-02-20T00:00:00Z", atRisk: 60, total: 180 },
+      ],
+    },
+    loading: false,
+    error: null,
+  },
+};
+
+/**
  * 24 data points showing a month of daily data with upward trend.
  */
 export const TwentyFourDataPoints: Story = {


Banrion:
Due to possible arch changes required for the date support of the chart, let's update the branch to merge into to a feature branch to unblock tickets that rely on the initial skeleton.

bitwarden_license/bit-web/src/app/dirt/shared/line-chart.component.ts
```diff
@@ -0,0 +1,182 @@
...
+              usePointStyle: true,
+            },
+          },
+          tooltip: {
+            enabled: true,
+          },
+        },
+        scales: {
+          x: {
+            type: configuration.xAxisType === "datetime" ? "time" : "linear",
```
Banrion, line 134: 🤔 With this implementation, the line chart component has to be updated every time we want to support passing a new configuration, for example, to support disabling the tooltip which is set to true no matter what someone may pass in to the configuration. This means that the configuration that is passed in is actually not applied until the component itself references the input configuration. This restricts the configuration, which may be a good way of enforcing consistency and standards with the component, but we may want to update the configuration to actually be a Partial defining the accepted configuration instead of allowing someone to pass in a full ChartConfiguration object. The alternative is to accept the full object and merge it with the default object. Thoughts?

package.json
```diff
@@ -175,6 +175,8 @@
     "buffer": "6.0.3",
     "bufferutil": "4.1.0",
     "chalk": "4.1.2",
+    "chart.js": "4.5.1",
+    "chartjs-adapter-date-fns": "3.0.0",
```
Banrion, line 179: 📝 Adding a note here to explain that we have added this library while developing, but this may change based on further discussions.

bitwarden_license/bit-web/src/app/dirt/shared/line-chart.component.ts
lastbestdev: The thought behind restricting the input configuration was mostly so that we wouldn't become dependent on the Chart.js configuration API, and our component would define its own API instead. This way, default configuration settings are baked in and not required to be passed in by consumers, and ideally the charting library used internally could change later on without the component API changing. 

If this is overkill, I don't mind the idea of doing a partial configuration object that matches the Chart.js configuration object. I do think passing the full configuration is too much overhead for using the component and also discourages consistency in how we use Chart.js

lastbestdev: Merging, with understanding we can improve/update the input configuration as needed later.

PR #19080, opened 2026-02-19
[PM-29823] New Organization Upgrade Path

sbrown-livefront:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-29823

## 📔 Objective

This PR adds the functionality to upgrade to a new organization from premium similar to https://bitwarden.atlassian.net/browse/PM-29593 but using the `New organization` button within the vault. 

The `organization-plans` files have been modernized to use signals and place some of the more complex logic into helper function. In addition, tests for this new functionality has been added to the existing test suite to retain base functionality while adding new functionality.

### Note:
I updated `apps/web/src/app/admin-console/organizations/sponsorships/families-for-enterprise-setup.component.html` to use the new signal inputs and outputs instead of interacting with the viewchild directly.

### Free User Upgrade to Enterprise

https://github.com/user-attachments/assets/4d9f5309-6209-4098-8af2-2cf50addfebd

#### Hides the ability to change storage and seat amount.
https://github.com/user-attachments/assets/a2551908-cfd3-4b86-a1d4-e62935e1a437

### Free Family Sponsorship

https://github.com/user-attachments/assets/edc82690-31b3-46dd-860c-85dbbc8eeeb3

### Create Organizations Pre-Selected Plan

https://github.com/user-attachments/assets/c0e5f67b-66fa-4a4c-ade3-942a9c7207ea

### Create Organizations Pre-Selected Product Tier

https://github.com/user-attachments/assets/51cf937c-76ac-4ee0-933c-a96bad8ae9ba


eliykat (dismissed):
Changes to AC files are minimal. Should Billing own the F4E component?

sbrown-livefront:
> Changes to AC files are minimal. Should Billing own the F4E component?

@eliykat great question. It seems like it. I'll ask the team.

eliykat (dismissed):

apps/web/src/app/billing/organizations/organization-plans.component.html
```diff
@@ -1,464 +1,436 @@
...
+                      }
+                    </ul>
+                  }
+                }
+              </bit-hint>
+            </bit-radio-button>
+
+            @if (tier != productTypes.Free) {
+              <span class="tw-pl-4">
+                @if (pm.basePrice && !acceptingSponsorship) {
```
amorask-bitwarden: ❌ `acceptingSponsorship()`

apps/web/src/app/billing/organizations/organization-plans.component.ts
```diff
@@ -86,120 +94,364 @@ const Allowed2020PlansForLegacyProviders = [
...
+  /**
+   * Initial plan for form initialization only.
+   * After initialization, the form control becomes the source of truth.
+   */
+  readonly initialPlan = input<PlanType>(PlanType.Free);
+
+  // Derived signals
+  readonly account = toSignal(this.accountService.activeAccount$);
+  readonly hasPremiumPersonally = toSignal(
+    this.billingAccountProfileStateService.hasPremiumPersonally$(this.account()!.id),
```
amorask-bitwarden: ❓ I assume this was tested, but I feel like invoking a signal at init-time within a `toSignal` could cause problems. It might be safer to `toSignal` on the full observable chain:

```typescript
readonly hasPremiumPersonally = toSignal(
    filter(Boolean),
    switchMap(account =>
      this.billingAccountProfileStateService.hasPremiumPersonally$(account.id)
    ),
  ),
  { initialValue: false },
);
```

apps/web/src/app/billing/organizations/organization-plans.component.html
```diff
@@ -1,464 +1,436 @@
...
+                        <li>{{ "limitedUsers" | i18n: pm.maxSeats }}</li>
+                      }
+                      @if (
+                        tier != productTypes.Free &&
+                        tier != productTypes.TeamsStarter &&
+                        pm.maxSeats
+                      ) {
+                        <li>{{ "addShareLimitedUsers" | i18n: pm.maxSeats }}</li>
+                      }
+                      @if (pm.maxSeats) {
```
amorask-bitwarden: ❌ This needs to be inverted. If the Password Manager plan has a max seats, we wouldn't want to show the user copy about adding unlimited users.

apps/web/src/app/billing/organizations/organization-plans.component.ts
```diff
@@ -721,14 +860,83 @@ export class OrganizationPlansComponent implements OnInit, OnDestroy {
...
+      throw new Error("Organization name is required for organization upgrade");
+    }
+
+    const billingAddress = getBillingAddressFromForm(this.billingFormGroup.controls.billingAddress);
+
+    if (!billingAddress?.country || !billingAddress?.postalCode) {
+      throw new Error("Billing address information is incomplete");
+    }
+
+    await this.accountBillingClient.upgradePremiumToOrganization(
```
amorask-bitwarden: We have a defect with this command in `server` that's either going to require holding this PR open or a follow-up PR to avoid merge conflicts. I noticed here that we're not passing the organization's encrypted default collection or the public/private key pair as part of this request, which means we won't be able to establish zero-knowledge encryption with the organization like we do when we sign up a new organization.

Issue raised: https://bitwarden.atlassian.net/browse/PM-32572

apps/web/src/app/billing/organizations/organization-plans.component.ts
```diff
@@ -721,14 +860,83 @@ export class OrganizationPlansComponent implements OnInit, OnDestroy {
...
+    );
+
+    await this.syncService.fullSync(true);
+
+    // Get the newly created organization
+    const organizations = await firstValueFrom(
+      this.organizationService.organizations$(this.account()!.id),
+    );
+
+    const newOrg = organizations?.find((org) => org.name === organizationName && org.isOwner);
```
amorask-bitwarden: ❓ Would it be easier if the endpoint returned the newly created organization ID?

apps/web/src/app/billing/organizations/organization-plans.component.html
```diff
@@ -1,464 +1,436 @@
...
+                      <li>{{ "xDayFreeTrial" | i18n: trialPeriodDays }}</li>
+                    }
+                  </ul>
+                } @else {
+                  @if (tier === productTypes.Teams && teamsStarterPlanIsAvailable()) {
+                    <ul class="tw-pl-0 tw-list-inside tw-mb-0">
+                      <li>{{ "includeAllTeamsStarterFeatures" | i18n }}</li>
+                      <li>{{ "chooseMonthlyOrAnnualBilling" | i18n }}</li>
+                      <li>{{ "abilityToAddMoreThanNMembers" | i18n: 10 }}</li>
+                      <li *ngIf="trialPeriodDays && createOrganization() && !upgradingFromPremium">
```
amorask-bitwarden: ⛏️ Can we clean up this last `ngIf`?

apps/web/src/app/billing/organizations/organization-plans.component.html
sbrown-livefront: Addressed in https://github.com/bitwarden/clients/pull/19080/commits/698a6d469196f92bd9bc1558f2d668ca5c12b4dd

apps/web/src/app/billing/organizations/organization-plans.component.html
sbrown-livefront: Thanks for this. I want through with a fine-tooth comb since you brought this up. Addressed in https://github.com/bitwarden/clients/pull/19080/commits/698a6d469196f92bd9bc1558f2d668ca5c12b4dd

apps/web/src/app/billing/organizations/organization-plans.component.html
sbrown-livefront: Addressed in https://github.com/bitwarden/clients/pull/19080/commits/698a6d469196f92bd9bc1558f2d668ca5c12b4dd

apps/web/src/app/billing/organizations/organization-plans.component.ts
sbrown-livefront: Addressed in https://github.com/bitwarden/clients/pull/19080/commits/c4cf71cda6b4d7f4992c0e200b106df91898f6ef. I'll wait for server changes to make sure these are correct

apps/web/src/app/billing/organizations/organization-plans.component.ts
sbrown-livefront: Addressed in https://github.com/bitwarden/clients/pull/19080/commits/c4cf71cda6b4d7f4992c0e200b106df91898f6ef again awaiting server changes

apps/web/src/app/billing/organizations/organization-plans.component.ts
sbrown-livefront: Addressed in https://github.com/bitwarden/clients/pull/19080/commits/0c49156e342518e9c9ab91ec9fafef4106a3426a

sbrown-livefront:
> Changes to AC files are minimal. Should Billing own the F4E component?

@eliykat we will adopt these files. I'll put those changes in another PR though. After changing the directory, there are some admin-console dependencies I'll have to work through for the move. I don't want to increase the scope of this PR too much more.

apps/web/src/app/billing/individual/upgrade/premium-org-upgrade-payment/services/premium-org-upgrade.service.ts
```diff
@@ -84,32 +90,22 @@ export class PremiumOrgUpgradeService {
...
     const tier: ProductTierType = this.ProductTierTypeFromSubscriptionTierId(planDetails.tier);
-    const [encryptedKey] = await this.keyService.makeOrgKey<OrgKey>(account.id);
+    const encryptionData = await this.generateOrganizationEncryptionData(account.id);
 
-    if (!encryptedKey.encryptedString) {
-      throw new Error("Failed to generate encrypted organization key");
-    }
-
-    await this.accountBillingClient.upgradePremiumToOrganization(
+    const orgId = await this.accountBillingClient.upgradePremiumToOrganization({
```
sbrown-livefront, line 95: 📝 Because the function signature changed, this service also needed to be updated

apps/web/src/app/billing/clients/account-billing.client.ts
```diff
@@ -66,22 +77,21 @@ export class AccountBillingClient {
...
+        key: request.organizationKey,
+        collectionName: request.collectionName,
+        publicKey: request.publicKey,
+        encryptedPrivateKey: request.encryptedPrivateKey,
+        targetProductTierType: request.planTier,
+        cadence: request.cadence as string,
+        billingAddress: request.billingAddress,
       },
       true,
       false,
```
amorask-bitwarden: ❌ This is still setting `hasResponse: false` which means this will result in `undefined`, not `string`.

apps/web/src/app/billing/organizations/organization-plans.component.ts
```diff
@@ -86,120 +94,368 @@ const Allowed2020PlansForLegacyProviders = [
...
+            Allowed2020PlansForLegacyProviders.includes(plan.type)))
+      );
+    });
+
+    result.sort((planA, planB) => planA.displaySortOrder - planB.displaySortOrder);
+
+    return result;
+  });
 
-  set plan(plan: PlanType) {
```
amorask-bitwarden, line 138: ❌ This and `productTier` used to be setters on this component and their reactivity to changes in the parent component triggered updates to the `formGroup`. That's important, because the `create-organization.component` renders the `organization-plans.component` and selects a plan based on a query parameter during `ngOnInit`.

`create-organization.component`:

- First Render: `Free`
- `ngOnInit`: Selection based on query parameter (now ignored by `organization-plans`)

To get around this, we can use `effect`, which will track the input signal dependency and run a side-effect on change for both `plan` and `productTier`:
```typescript
effect(() => {
    const plan = this.initialPlan(); // <-- tracked dependency
    if (plan !== PlanType.Free) {
      this.formGroup.controls.plan.setValue(plan);
    }
  });
```

apps/web/src/app/billing/organizations/organization-plans.component.ts
```diff
@@ -721,14 +858,75 @@ export class OrganizationPlansComponent implements OnInit, OnDestroy {
...
+      organizationKey: encryptionData.key,
+      collectionName: encryptionData.collectionCt,
+      publicKey: encryptionData.orgKeys[0],
+      encryptedPrivateKey: encryptionData.orgKeys[1].encryptedString as string,
+      planTier: this.formValues().productTier!,
+      cadence: SubscriptionCadenceIds.Annually,
+      billingAddress,
+    });
+
+    await this.syncService.fullSync(true);
```
amorask-bitwarden: ❌ `fullSync` is already run in the parent method on line 829.

apps/web/src/app/billing/organizations/organization-plans.component.ts
```diff
@@ -721,14 +858,75 @@ export class OrganizationPlansComponent implements OnInit, OnDestroy {
...
-      case ProductTierType.Families:
-        return false;
-      default:
-        return true;
+  /**
+   * Generates encryption data needed for creating a new organization.
+   * Uses the active user account signal to get the user ID.
+   * @returns Organization encryption data including keys and encrypted collection name
+   */
+  private async generateOrganizationEncryptionData(): Promise<{
```
amorask-bitwarden: ❓ Isn't this duplicated in the `premium-org-upgrade.service`?

apps/web/src/app/billing/organizations/organization-plans.component.ts
```diff
@@ -721,14 +858,75 @@ export class OrganizationPlansComponent implements OnInit, OnDestroy {
...
+    };
+  }
+
+  /**
+   * Upgrades a premium subscription to an organization plan.
+   * Uses the account signal to get the user ID.
+   * @param encryptionData Organization encryption data from generateOrganizationEncryptionData
+   * @returns The ID of the newly created organization
+   */
+  private async upgradePremiumToOrganization(encryptionData: {
```
amorask-bitwarden: ❓ I think this is also in the `premium-org-upgrade.service`?

apps/web/src/app/billing/individual/upgrade/premium-org-upgrade-payment/services/premium-org-upgrade.service.ts
```diff
@@ -49,6 +53,8 @@ export class PremiumOrgUpgradeService {
     private syncService: SyncService,
     private keyService: KeyService,
     private organizationService: OrganizationService,
```
amorask-bitwarden: ⛏️  Unused

apps/web/src/app/billing/organizations/organization-plans.component.ts
```diff
@@ -86,120 +94,368 @@ const Allowed2020PlansForLegacyProviders = [
...
+  readonly onTrialBillingSuccess = output<OnTrialBillingSuccessArgs>();
+
+  // State properties
+  protected loading = true;
+  protected selfHosted = false;
+  protected readonly productTypes = ProductTierType;
+  protected formPromise: Promise<string> | null = null;
+  protected singleOrgPolicyAppliesToActiveUser = false;
+  protected isInTrialFlow = false;
+  protected selectedFile: File | null = null;
```
amorask-bitwarden: ⛏️  Unused

apps/web/src/app/billing/organizations/organization-plans.component.ts
```diff
@@ -86,120 +94,368 @@ const Allowed2020PlansForLegacyProviders = [
...
-
-  // FIXME(https://bitwarden.atlassian.net/browse/CL-903): Migrate to Signals
-  // eslint-disable-next-line @angular-eslint/prefer-signals
-  @Input()
-  get productTier(): ProductTierType {
-    return this._productTier;
-  }
+  // ViewChildren
+  readonly enterPaymentMethodComponent = viewChild(EnterPaymentMethodComponent);
+  readonly cartSummaryComponent = viewChild(CartSummaryComponent);
```
amorask-bitwarden: ⛏️ Unused

apps/web/src/app/billing/individual/upgrade/premium-org-upgrade-payment/premium-org-upgrade-payment.component.spec.ts
```diff
@@ -197,9 +199,22 @@ describe("PremiumOrgUpgradePaymentComponent", () => {
...
+          useValue: {
+            encryptString: jest.fn().mockResolvedValue(new EncString("encrypted-collection")),
+          },
         },
-        { provide: SyncService, useValue: { fullSync: jest.fn().mockResolvedValue(undefined) } },
+        {
+          provide: SyncService,
+          useValue: { fullSync: jest.fn().mockResolvedValue(undefined) },
+        },
+        { provide: OrganizationService, useValue: mockOrganizationService },
```
amorask-bitwarden: ⛏️  Already provided on 199.

apps/web/src/app/billing/organizations/organization-plans.component.ts
sbrown-livefront: Good call. I'll import the service.

apps/web/src/app/billing/organizations/organization-plans.component.ts
sbrown-livefront: I weighed whether or not to use the effect, but thought better of it because it didn't seem like the appropriate use case. The docs said

```
Use cases for effects

- Logging signal values, either for analytics or as a debugging tool.
- Keeping data in sync with different kind of storages: window.localStorage, session storage, cookies etc.
- Adding custom DOM behavior that can't be expressed with template syntax.
- Performing custom rendering to a <canvas> element, charting library, or other third party UI library.

When not to use effects

Avoid using effects for propagation of state changes. This can result in ExpressionChangedAfterItHasBeenChecked errors, infinite circular updates, or unnecessary change detection cycles.

Instead, use [computed](https://angular.dev/api/core/computed) signals to model state that depends on other state.
```

Since this seemed to lean more to the state change, I opted to have the form group as the source of truth for setting values since that's what seemed to drive a lot of the logic. (I added a comment over the two signals).

The `initialProductTier`and `initialPlan` signals are set in the `formGroup` object on instantiation which is the sets the `formvalues` signal as the initial value. 

All of the setting logic for plan and productTier after that are made to the formgroup, see `preselectUpgradePlan` and `setupProviderConfiguration` which triggers `formvalues` for cartSummary and other subscribers. 

I think effect could work, but it didn't fit the bill for this. Should I update it to use effects instead?

apps/web/src/app/billing/individual/upgrade/premium-org-upgrade-payment/premium-org-upgrade-payment.component.spec.ts
sbrown-livefront: Addressed in 36f8dd939de4bde1c11497c315ac8fa647a55591

apps/web/src/app/billing/organizations/organization-plans.component.ts
sbrown-livefront: Addressed in ac3357c64cd5ed52d57c6bdf6a82e182dfe222a0

apps/web/src/app/billing/organizations/organization-plans.component.ts
sbrown-livefront: Addressed in ac3357c64cd5ed52d57c6bdf6a82e182dfe222a0

apps/web/src/app/billing/organizations/organization-plans.component.ts
sbrown-livefront: Addressed in 9476e71f817597c4eec8e2193775409efacd3640

apps/web/src/app/billing/organizations/organization-plans.component.ts
sbrown-livefront: Addressed in 8c667e3a7bc2286b9ce4246cd55871dbf136617a

apps/web/src/app/billing/individual/upgrade/premium-org-upgrade-payment/services/premium-org-upgrade.service.ts
sbrown-livefront: Addressed in ca82a9462a27c3ef0e921741ddd9f0b49d91dd8e

apps/web/src/app/billing/clients/account-billing.client.ts
sbrown-livefront: Addressed in https://github.com/bitwarden/clients/pull/19080/commits/8fb7b92c5b629c986a7891be258c68e32688f50f

apps/web/src/app/billing/organizations/organization-plans.component.ts
sbrown-livefront: Updated the OnInit to cover this case. I also added some tests to be sure. In addition, there are some screen shares for both.

Addressed in https://github.com/bitwarden/clients/pull/19080/commits/c7672eca4b2bdfec6e80fcbb7e92f8bf86b97f69

PR #19083, opened 2026-02-19
[PM-32481] Apply same custom scrollbar to nav

Hinton:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-32481
## 📔 Objective

Adds the same custom scrollbar to the navigation that is used in the rest of the application.

## 📸 Screenshots

image

### DIFF
--- a/apps/desktop/src/scss/base.scss
+++ b/apps/desktop/src/scss/base.scss
@@ -121,7 +121,11 @@ div:not(.modal)::-webkit-scrollbar-thumb,
 #bit-side-nav::-webkit-scrollbar-thumb {
   border-radius: 10px;
   margin-right: 1px;
+}
 
+div:not(.modal)::-webkit-scrollbar-thumb,
+.cdk-virtual-scroll-viewport::-webkit-scrollbar-thumb,
+.vault-filters::-webkit-scrollbar-thumb {
   @include themify($themes) {
     background-color: themed("scrollbarColor");
   }


pamperer562580892423:
@Hinton Sorry to hop in here, but as I just read (and emphasis my edit)...

> Adds the same custom scrollbar to the navigation that is used **in the rest of the application.**

... maybe you (BW) are already aware, but if not: that custom scrollbar is not used in the (whole) rest of the application. At least the "notes" field is missing it too:

In "View item mode":

2026-02-19--13-25-04-d31yBHwUrn

&nbsp;

And in "Edit item mode":

2026-02-19--13-25-11-QwmAjhIzZf

&nbsp;

(screenshots from the desktop app 2026.1.0, of course // EU cloud server 2026.2.0)

apps/desktop/src/scss/base.scss
```diff
@@ -102,20 +102,23 @@ textarea {
...
+.vault-filters::-webkit-scrollbar-track,
+#bit-side-nav::-webkit-scrollbar-track {
   background-color: transparent;
 }
 
 div:not(.modal)::-webkit-scrollbar-thumb,
 .cdk-virtual-scroll-viewport::-webkit-scrollbar-thumb,
-.vault-filters::-webkit-scrollbar-thumb {
+.vault-filters::-webkit-scrollbar-thumb,
+#bit-side-nav::-webkit-scrollbar-thumb {
```
vleague2, line 121: Maybe I'm missing something here, but why add the selector to this css if we're overriding the colors for the sidenav scrollbar thumb right below this?

vleague2:
Hi @pamperer562580892423, thanks for pointing those scrollbars out! I think we can take a look at overall scrollbar consistency in our apps to unify everything.

apps/desktop/src/scss/base.scss
Hinton: It applies border-radius and margin-right. I can split out the colors from. this generic selector if we want?

PR #19085, opened 2026-02-19
[PM-32472] [Defect] Generator page will not display on desktop

harr1424:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-32472

## 📔 Objective

The issue is related to a change made in [PR #18388](https://github.com/bitwarden/clients/pull/18388), where the the `ButtonLinkDirective` class was removed and replaced by the `LinkComponent` in terms of both using the `button[bitLink]` selector. Since the credential generator component also makes use of the bit-item-content selector, which resolves to a component, Angular complains that you cannot have 2 components associated with a single HTML element.

The proposed fix is to remove the usage of `bitLink`.

Note that the relevant link has the English `i18n` value "Generator history" in the screenshot below. 

## 📸 Screenshots

Screenshot 2026-02-19 at 15 58 21

### DIFF
--- a/apps/desktop/src/app/tools/generator/credential-generator.component.html
+++ b/apps/desktop/src/app/tools/generator/credential-generator.component.html
@@ -3,9 +3,19 @@
   <ng-container bitDialogContent>
     <tools-credential-generator />
     <bit-item>
-      <button type="button" bit-item-content aria-haspopup="dialog" (click)="openHistoryDialog()">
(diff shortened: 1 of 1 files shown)


itsadrago:
There are some missing css styles we need to account for by removing the `bitLink` selector. See the old link directive code at https://github.com/bitwarden/clients/blob/web-v2026.1.1/libs/components/src/link/link.directive.ts#L114 and the css that was added

PR #19087, opened 2026-02-19
[PM-32488] Disables autocomplete when explicitly requested

vvalentin-lucca:
Sometimes the plugin misinterprets the need to offer autocomplete.

This happens, for example, when the word ‘email’ is present in a field label and that field does not expect an email address.

The autocomplete attribute set to off is an expected way of indicating this, and should be taken into account.


jprusik:
Our current functionality intentionally ignores `autocomplete=off` cases as it only unambiguously signals the desire of the host page developer and not the user. Since the user expects autofill to operate when requested, this is not a filtering mechanism we'll be incorporating at this time.

(`data-bwignore` is a legacy concept for internal Bitwarden consumption. It should also not be relied upon by developers as a control for users' extension/client behaviour)

PR #19091, opened 2026-02-19
[PM-32341] clear cache if user cancels save flow for browser

Jingo88:
## 🎟️ Tracking

[PM-32341](https://bitwarden.atlassian.net/browse/PM-32341)

## 📔 Objective

Removing the previous `clearCache` solution as it did not resolve the flow of a user closing the new login save popup window with the window close event. 

We will now have an `initialCacheValue` inside the default cipher form service. This will prevent any overriding of the cache that may happen when the `cipher-form` is initialized and components such as `login details, autofill, identity, etc` call `patchCipher`

## 📸 Screen Recording

https://github.com/user-attachments/assets/38d7d5e3-906f-4249-b574-c290d7097180

[PM-32341]: https://bitwarden.atlassian.net/browse/PM-32341?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/browser/src/vault/popup/components/vault/add-edit/add-edit.component.spec.ts
+++ b/apps/browser/src/vault/popup/components/vault/add-edit/add-edit.component.spec.ts
@@ -373,11 +373,6 @@ describe("AddEditComponent", () => {
   });
 
   describe("handleBackButton", () => {
-    beforeEach(() => {
-      jest.spyOn(component, "cipherFormComponent").mockReturnValue({
-        clearCache: jest.fn(),
-      } as any);
-    });
     it("disables warning and aborts fido2 popout", async () => {
       // @ts-expect-error - `inFido2PopoutWindow` is a private getter, mock the response here
       // for the test rather than setting up the dependencies.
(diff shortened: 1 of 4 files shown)


apps/browser/src/vault/popup/components/vault/add-edit/add-edit.component.spec.ts
```diff
@@ -373,6 +373,11 @@ describe("AddEditComponent", () => {
   });
 
   describe("handleBackButton", () => {
+    beforeEach(() => {
+      jest.spyOn(component, "cipherFormComponent").mockReturnValue({
+        clearCache: jest.fn(),
+      } as any);
```
nick-livefront: 🤔 Do you need this `any`?

nick-livefront (approved):

jaasen-livefront (approved):

apps/browser/src/vault/popup/components/vault/add-edit/add-edit.component.spec.ts
Jingo88: Just added this back. It was passing locally but without it browser build failed.

PR #19092, opened 2026-02-19
[PM-32502] fixed icon / copy value spacing in button on send access page

itsadrago:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-32502

## 📔 Objective

While testing send access, we found that the distance between the clipboard icon and the “Copy value” text in the button on the view send page didn’t match the Figma diagram. There should be 8px of padding between the label and the icon.

## 📸 Screenshots

**Before:**

image

**After:**

image


harr1424 (approved):

apps/web/src/app/tools/send/send-access/send-access-text.component.html
```diff
@@ -11,6 +11,6 @@
 </div>
 <div class="tw-mb-3">
   <button bitButton type="button" buttonType="primary" [block]="true" (click)="copyText()">
-    <bit-icon name="bwi-clone"></bit-icon>{{ "copyValue" | i18n }}
+    <bit-icon name="bwi-clone" class="tw-mr-2"></bit-icon>{{ "copyValue" | i18n }}
```
Hinton: ~This should be `tw-me-2` (logistical css) for LTR/RTL support, https://components.bitwarden.com/?path=/docs/component-library-button--docs#with-icon.~

Actually it should probably use the `startIcon` input.

```html
<button bitButton type="button" buttonType="primary" [block]="true" (click)="copyText()" startIcon="bwi-clone">
  {{ "copyValue" | i18n }}
</button>
```

PR #19104, opened 2026-02-20
Add linting rule to detect when icons are used in buttons

Hinton:
## 📔 Objective

Adds a linting rule which warns when adding an icon to buttons informing them to use `iconStart` and `iconEnd`.

## 📸 Screenshots

image


vleague2:
I think these slots were also added for bit-link, should we add that to the lint rule too?

PR #19106, opened 2026-02-20
Fix type error on main

nick-livefront:
## 🎟️ Tracking

N/A

## 📔 Objective

https://github.com/bitwarden/clients/pull/18849 introduced a change of inferred types for the feature flags by using a numeric feature flag. This broke a test on `main` that wasn't in the feature branch. Adding a cast here to satisfy typescript. 

## 📸 Screenshots

N/A


willmartian (approved):
ty!

PR #19111, opened 2026-02-20
[PM-30127] [Defect] Automatic pop out is not necessary when editing a file send 

harr1424:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-30127

## 📔 Objective

https://github.com/bitwarden/clients/pull/17950 mitigated a situation where the extension would crash upon displaying a file picker on specific platforms and browsers. This PR required an extension pop-out too eagerly. This PR relaxes the pop-out behavior as follows: 
- editing a Send (including a file type Send) does not expose a file picker and should never require a pop-out 
- creating a text type Send does not expose a file picker and should never require a pop-out (but currently does on Firefox and Safari)


apps/browser/src/popup/app-routing.module.ts
```diff
@@ -356,7 +356,7 @@ const routes: Routes = [
   {
     path: "edit-send",
     component: SendAddEditV2Component,
-    canActivate: [authGuard, filePickerPopoutGuard()],
```
harr1424, line 359: Neither file or text type Sends expose a file picker

apps/browser/src/tools/popup/guards/file-picker-popout.guard.spec.ts
harr1424, line 1: Throughout this spec file pertaining to `filePickerPopoutGuard` test coverage related to the `edit-send` functionality has been removed because this is now enforced at the router level, and is not the responsibility of the guard.

harr1424:
@claude please review

PR #19119, opened 2026-02-20
SSH Agent v2: server framework

neuronull:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-30757

## 📔 Objective

Building on #18880 , this change adds the basic framework for handling new client connections in the server. There are stubs and todos, as the implementation of listener will be completed in the next PR.


apps/desktop/desktop_native/ssh_agent/src/server/connection.rs
```diff
@@ -0,0 +1,56 @@
...
+    auth_policy: Arc<A>,
+    stream: S,
+    token: CancellationToken,
+}
+
+impl<K, A, S> ConnectionHandler<K, A, S>
+where
+    K: KeyStore,
+    A: AuthPolicy,
+    S: AsyncRead + AsyncWrite + Unpin,
```
quexten, line 32: Thought: Could we introduce a newtype here, 

```
struct Connection<S> {
  stream: S,
  peerinfo: PeerInfo,
}
```

? This would make more sense from a mental model - the connection handler takes connections - and also leave the peerinfo in-tact such that if we somehow need it, for the auth policy or else, we have it.

This connection struct could later contain info such as:
- Which peer process is requesting access
- Is this coming over agent forwarding
  - If so, which host is the connection forwarded to?

apps/desktop/desktop_native/ssh_agent/src/server/mod.rs
```diff
@@ -41,18 +49,102 @@ where
...
     pub fn start(&mut self) -> Result<()> {
-        todo!();
+        if self.is_running() {
+            return Err(anyhow::anyhow!("Server is already running"));
+        }
+
+        let token = CancellationToken::new();
+
+        // TODO: Create socket/pipe listener
+        //
```
quexten: Thought: We should consider for V2 whether we can support a list of listeners.
Specifically, we want to be able to listen on multiple types of sockets, since on Windows, *both* unix sockets *and* named pipes are allowed.

The V1 agent only supports *one* of these.

neuronull: Sounds good. I think it makes sense to add that provision for multiple listeners here. IMO we should keep the existing single-listener when deploying the v2. But this is a good call out to lay the foundation for it now 👍

apps/desktop/desktop_native/ssh_agent/src/server/connection.rs
neuronull: Makes sense to me 👍  I'll do it.

(As an aside, the NewType pattern involves wrapping some struct in a new single-field tuple struct, so it isn't exactly what you're suggesting. You're suggestion replacing a tuple ( `(Stream, PeerInfo)` ) with a strict type, which is a good suggestion but not the NewType pattern)

apps/desktop/desktop_native/ssh_agent/src/server/mod.rs
neuronull: done in https://github.com/bitwarden/clients/pull/19119/commits/7c4b2e9f80344cd56a59e29a3d4ac61a251de08d

apps/desktop/desktop_native/ssh_agent/src/server/connection.rs
neuronull: done in https://github.com/bitwarden/clients/pull/19119/commits/3fc5eb27f91a9ba6fc0e6d38e3d9d924ec8bdb57

quexten: Ah, thanks for the clarification on newtype. Looks like what I should have written is "new type" then ;).

apps/desktop/desktop_native/ssh_agent/src/server/mod.rs
quexten: I agree that we should not do feature creep, but making bigger changes later is always hard, while now the code is malleable so where it makes sense, I do support them / am in favor of them.

apps/desktop/desktop_native/ssh_agent/src/server/connection.rs
```diff
@@ -0,0 +1,70 @@
...
+        loop {
+            tokio::select! {
+                () = self.token.cancelled() => {
+                    info!("Connection handler received cancellation signal");
+                    break;
+                }
+
+                // TODO: read SSH protocol message from self.connection.stream
+                // TODO: parse message type, use auth policy and keystore to satisfy requests
+                // TODO: build response and write back to self.connection.stream
```
coltonhurst: Can we add tickets for these?

(And the other TODO comments in the pr?)

apps/desktop/desktop_native/ssh_agent/src/server/mod.rs
```diff
@@ -28,31 +35,129 @@ pub struct SshAgentServer<K, A> {
     auth_policy: Arc<A>,
     /// Async task coordination to use when asked to stop. Is `None` when non running.
```
coltonhurst: While we're here... maybe we can update this to follow the `accept_handle` comment

```suggestion
    /// Async task coordination to use when asked to stop. Is `None` when not running.
```

apps/desktop/desktop_native/ssh_agent/src/server/mod.rs
```diff
@@ -28,31 +35,129 @@ pub struct SshAgentServer<K, A> {
...
+    async fn accept<L>(
+        listeners: Vec<L>,
+        keystore: Arc<K>,
+        auth_policy: Arc<A>,
+        cancel_token: CancellationToken,
+    ) where
+        L: Listener + 'static,
+        L::Stream: 'static,
+    {
+        let (tx, mut rx) = mpsc::channel::<Connection<L::Stream>>(32);
```
coltonhurst: Can we add a const to avoid magic numbers?

neuronull: 👍  meant to do that but forgot to circle back

neuronull:
@quexten  and @coltonhurst , all the changes since your last approvals on this PR were merging in from the upstream so if I resolved the merge conflicts correctly there should be very few new changes since your last approval.

PR #19125, opened 2026-02-22
[PM-32530] [Defect] Double / infinite spinner is displayed when accessing an email verified Send

harr1424:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-32530
https://bitwarden.atlassian.net/browse/PM-32529

## 📔 Objective

This PR addresses a situation where a user would see two progress spinners when accessing an email verified Send. Per an internal discussion, the spinner center top of the card has been removed, leaving only the spinner in the submit button. 

Additionally, if the OTP code form field was submitted with no value, the first submission would result in the inline error message shown below, while a subsequent submission would result in an infinite spinner. The logic backing the "View Send" button has been made more strict such that: 
- empty form field submissions are no longer allowed 
- if an unhandled exception does occur, the spinner will be set to false, allowing for subsequent attempts

#### Before
Screenshot 2026-02-21 at 18 59 19

#### After
Screenshot 2026-02-21 at 18 07 51

### Screenshot demonstrating the "View Send" button is disabled when the OTP form field has no value 
Screenshot 2026-02-21 at 18 56 03

### DIFF
--- a/apps/web/src/app/tools/send/send-access/send-access-email.component.html
+++ b/apps/web/src/app/tools/send/send-access/send-access-email.component.html
@@ -27,6 +27,7 @@
     buttonType="primary"
     [loading]="loading()"
     [block]="true"
+    [disabled]="!otp?.value"
     class="tw-mb-3"
   >
     <span>{{ "viewSend" | i18n }} </span>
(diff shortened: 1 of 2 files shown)


harr1424:
@claude please review

apps/web/src/app/tools/send/send-access/send-access-email.component.html
mcamirault, line 13: ❓ Do we want the email field to have the same disabled logic?

harr1424: Good catch, this is probably nice to have for consistency, but unlike the OTP code prior to this PR, an empty email cannot be submitted.

PR #19136, opened 2026-02-23
PM-31767 resend feature

bmbitwarden:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-31767

## 📔 Objective

[Hint text should be displayed ](https://www.figma.com/design/Wqxyfz2gnEJBzxeIl6MHrl/Email-verification-for-Send?node-id=8080-29710&t=VWb23bBOT7vE8lyv-1)on the verification code input field to guide the recipient (e.g., "Did not receive code? Contact the sender to confirm your email address or resend code").

## 📸 Screenshots

https://github.com/user-attachments/assets/5092b41b-bb32-47bc-8499-6628e4fc452a

https://github.com/user-attachments/assets/5092b41b-bb32-47bc-8499-6628e4fc452a


harr1424 (changes requested):
When testing this out I noticed a problem we should address. If a user leaves the OTP submission page (to switch to their email tab) or even interacts with the page (clicking outside of the card) the resend link and related text disappears: 

### On initial view render (looks good!) 

Image

### After any user event besides OTP submission 

Image

bmbitwarden:
> When testing this out I noticed a problem we should address. If a user leaves the OTP submission page (to switch to their email tab) or even interacts with the page (clicking outside of the card) the resend link and related text disappears:
> 
> ### On initial view render (looks good!)
> Image
> ### After any user event besides OTP submission
> Image

i reviewed this item you bring up with alex yesterday - - since the input is in focus when the page loads, when the user leaves it enters "touched" state

Alex said this is intented for the validation message to show at that moment

mcamirault (changes requested):
The only problem I noticed is that the "verificationCodeRequired" message is shown as a toast every time the OTP entry field blurs, which I believe is not desired. Behavior and code changes look good otherwise.

apps/web/src/app/tools/send/send-access/send-auth.component.ts
```diff
@@ -187,23 +182,37 @@ export class SendAuthComponent implements OnInit {
...
         }
-        this.otpSubmitted = true;
+        this.enterOtp.set(true);
         this.updatePageTitle();
+      } else if (otpInvalid(response.error)) {
+        this.toastService.showToast({
+          variant: "error",
+          title: undefined,
+          message: this.i18nService.t("invalidVerificationCode"),
+        });
```
mcamirault: ⚠️ This function and the error it checks for no longer exist, so this branch should be removed

mcamirault (changes requested):

harr1424 (commented):
Pending design review from @sukhleenb but this looks good to me! 

I see a failing test for strict typechecking: 
```
Error: [typescript-strict-plugin] apps/web/src/app/tools/send/send-access/send-auth.component.ts(253,11): error TS2322: Type 'null' is not assignable to type 'string | undefined'.
```

mcamirault (dismissed):

mcamirault (dismissed):

mcamirault (approved):

harr1424 (approved):

PR #19151, opened 2026-02-23
[PM-31426] add categories to policy page

JaredScar:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-31426

## 📔 Objective

Add categories to policies page via new feature

## 📸 Screenshots
image


willmartian (dismissed):
Hey @JaredScar, this should use our existing tab component: https://components.bitwarden.com/?path=/story/component-library-header--with-tabs

That component will be updated to look more like what you have here in the near future. I apologize for the confusion!

apps/web/src/app/admin-console/organizations/policies/pipes/policy-category.pipe.ts
eliykat, line 1: Policies use a data-driven pattern, where you define policy metadata by implementing the [BasePolicyEditDefinition](https://github.com/bitwarden/clients/blob/main/apps/web/src/app/admin-console/organizations/policies/base-policy-edit.component.ts#L32). See the [README](https://github.com/bitwarden/clients/blob/main/apps/web/src/app/admin-console/organizations/policies/README.md?plain=1#L45) for more information.

Instead of this centralized map, we should require each `PolicyEditDefinition` to define what category it's in. That also means you don't need a fallback, because each policy will be required to specify its category.

apps/web/src/app/admin-console/organizations/policies/policies.component.html
```diff
@@ -1,8 +1,65 @@
...
+      <button
+        type="button"
+        role="tab"
+        id="tab-data-controls"
+        aria-controls="data-controls"
+        [attr.aria-selected]="activeCategory() === PolicyCategory.DataControl"
+        [class]="
+          activeCategory() === PolicyCategory.DataControl
+            ? 'tw-px-4 tw-py-3 tw-font-semibold tw-text-sm tw-bg-transparent tw-border-0 tw-border-b-2 tw-border-solid tw-border-primary-600 tw-text-primary-600 tw-cursor-pointer focus-visible:tw-outline-none focus-visible:tw-ring-2 focus-visible:tw-ring-primary-600 focus-visible:tw-ring-inset'
+            : 'tw-px-4 tw-py-3 tw-font-semibold tw-text-sm tw-bg-transparent tw-border-0 tw-border-b-2 tw-border-solid tw-border-transparent tw-text-main tw-cursor-pointer hover:tw-underline focus-visible:tw-outline-none focus-visible:tw-ring-2 focus-visible:tw-ring-primary-600 focus-visible:tw-ring-inset'
```
eliykat: I think this will be resolved when you use the component library, but having to specify this many Tailwind classes in one of our components is usually a code smell. If the styling is this extensive, it suggests we should be using a pre-built component.

apps/web/src/app/admin-console/organizations/policies/policies.component.html
```diff
@@ -13,26 +70,163 @@
...
-                    {{ p.name | i18n }}
-                  </button>
-                  @if (policiesEnabledMap.get(p.type)) {
-                    <span bitBadge variant="success">{{ "on" | i18n }}</span>
-                  }
-                </div>
-                <small class="tw-text-muted tw-block">{{ p.description | i18n }}</small>
-              </td>
-            </tr>
+    <!-- Data controls -->
```
eliykat: I expect that each policy section has the same layout - is that right? In that case, I suggest writing this once and wrapping it in a for loop that iterates over each section and renders the title and policies for that section.

apps/web/src/app/admin-console/organizations/policies/policies.component.html
```diff
@@ -1,8 +1,65 @@
...
+        {{ "vaultManagement" | i18n }}
+      </button>
+    </nav>
+  </app-header>
+</div>
 
 @let organization = organization$ | async;
 @let policiesEnabledMap = policiesEnabledMap$ | async;
 @let organizationId = organizationId$ | async;
+@let orderedPolicies = policies$ | async | policyOrder;
```
eliykat: Does the `POLICY_ORDER_MAP` still work now that it's split into sections? e.g. `SendOptions` is 14 in the order map, but third on the page. I think the order map is no longer a clear way to organize this.

I suggest that the policy edit definition defines a priority number which is used for sorting. That is a bit awkward, because it only works in relation to the priority number of other definitions in the same section, but it keeps the policy definitions as the single source of truth for policy metadata.

apps/web/src/app/admin-console/organizations/policies/policies.component.html
```diff
@@ -13,26 +70,163 @@
...
-            </tr>
+    <!-- Data controls -->
+    <section id="data-controls" role="tabpanel" aria-labelledby="tab-data-controls" #dataSection>
+      <h2 bitTypography="h2" class="tw-py-3">
+        {{ "dataControls" | i18n }}
+      </h2>
+      <bit-table>
+        <ng-template body>
+          @for (p of orderedPolicies | policyCategory: PolicyCategory.DataControl; track $index) {
+            @if (p.display$(organization, configService) | async) {
```
eliykat: (non-blocking) There's a bit of transformation going on now:
- order policies by category
- order policies within each category
- omit policies that shouldn't be displayed

This is only a suggestion, but instead of multiple pipes and the templating syntax, would it be simpler to do this work in an observable/computed signal in the component? Then the template is kept simple and just iterates over what the component gives it.

Or better yet, the `PolicyListService` can handle this so that the whole component stays simple.

apps/web/src/app/admin-console/organizations/policies/policies.component.ts
eliykat, line 1: The changes to the component here are pretty low level and also not what I'd expect to see outside of the Component Library. I assume using the CL will clean this up.

libs/components/src/banner/banner.component.ts
eliykat, line 1: The components in this directory are widely shared and we should be very reluctant to change them, as we may cause regressions elsewhere in the app. Are these changes still needed?

eliykat (changes requested):

amorask-bitwarden (dismissed):
Billing file looks good.

JaredScar:
> Hey @JaredScar, this should use our existing tab component: https://components.bitwarden.com/?path=/story/component-library-header--with-tabs
> 
> That component will be updated to look more like what you have here in the near future. I apologize for the confusion!

This got moved to phase 2 of changes! Thanks for the update though on where the update is going haha 😄

Thomas-Avery (dismissed):
KM changes LGTM. Does look like CI has failures fyi.

amorask-bitwarden (dismissed):
Single billing file change is simple.

apps/web/src/app/admin-console/organizations/policies/pipes/policy-order.pipe.ts
eliykat, line 1: Per Claude's comment - this is now unused and can be deleted.

apps/web/src/app/admin-console/organizations/policies/policy-edit-definitions/auto-confirm-policy.component.ts
```diff
@@ -16,12 +16,15 @@ import { ConfigService } from "@bitwarden/common/platform/abstractions/config/co
...
 import { BasePolicyEditDefinition, BasePolicyEditComponent } from "../base-policy-edit.component";
+import { PolicyCategory } from "../pipes/policy-category";
 import { AutoConfirmPolicyDialogComponent } from "../policy-edit-dialogs/auto-confirm-edit-policy-dialog.component";
 
 export class AutoConfirmPolicy extends BasePolicyEditDefinition {
   name = "automaticUserConfirmation";
   description = "autoConfirmDescription";
   type = PolicyType.AutoConfirm;
+  category = PolicyCategory.VaultManagement;
+  priority = 9;
```
eliykat: I suggest making priority numbers multiples of 10 so we can easily insert new ones in the future without renumbering the rest. e.g. 10, 20, 30, 40 instead of 1, 2, 3, 4.

apps/web/src/app/admin-console/organizations/policies/policies.component.html
eliykat, line 1: The template looks great - the for loop really cleans it up. Good work!

apps/web/src/app/admin-console/organizations/policies/policies.component.html
```diff
@@ -1,36 +1,63 @@
...
-                <small class="tw-text-muted tw-block">{{ p.description | i18n }}</small>
-              </td>
-            </tr>
-          }
-        }
-      </ng-template>
-    </bit-table>
+    @for (section of policySections; track section.category) {
+      <section [class.tw-mt-6]="!$first">
+        <h2 bitTypography="h2" class="tw-py-3 tw-pl-3 !tw-mb-0">
```
eliykat: I'm still concerned about some of the custom classes here. What about the default typography isn't meeting our needs? This may require more discussion with the designer or UIF Team. Generally our Figmas should match our Component Library. If there's some good reason to deviate that's OK, I just want to make sure there is.

Same concerns with the custom TW classes applied inside the `bitCell` divs below.

apps/web/src/app/admin-console/organizations/policies/policies.component.html
```diff
@@ -1,36 +1,63 @@
...
+                      <small class="tw-text-muted tw-block tw-truncate">
+                        {{ p.description | i18n }}
+                      </small>
+                    </div>
+                    <div class="tw-shrink-0">
+                      @if (policiesEnabledMap.get(p.type)) {
+                        <span bitBadge variant="success" aria-hidden="true">
+                          {{ "on" | i18n }}
+                        </span>
+                        <span class="tw-sr-only">{{ "on" | i18n }}</span>
```
eliykat: Please double check this with UIF: I think our component library should handle accessibility, so we don't need to use `aria-hidden` on the badge and then have a separate `sr-only` span.

apps/web/src/app/admin-console/organizations/policies/policies.component.ts
```diff
@@ -76,14 +71,43 @@ export class PoliciesComponent {
...
-      return policiesEnabledMap;
+      const map = new Map<PolicyType, boolean>();
+      orgPolicies.forEach((op) => map.set(op.type, op.enabled));
+      return map;
     }),
   );
 
+  protected policies$: Observable<readonly BasePolicyEditDefinition[]> = defer(() =>
+    of(this.policyListService.getPolicies()),
+  );
```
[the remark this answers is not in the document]
eliykat: Does this Claude [comment](https://github.com/bitwarden/clients/pull/19151/changes#r2885872627) need actioning?

apps/web/src/app/admin-console/organizations/policies/policies.component.ts
```diff
@@ -76,14 +71,39 @@ export class PoliciesComponent {
...
-        policiesEnabledMap.set(op.type, op.enabled);
-      });
-      return policiesEnabledMap;
+      const map = new Map<PolicyType, boolean>();
+      orgPolicies.forEach((op) => map.set(op.type, op.enabled));
+      return map;
     }),
   );
 
+  protected policySections$: Observable<PolicySection[]> = this.organization$.pipe(
```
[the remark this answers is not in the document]
eliykat: Does this Claude [comment](https://github.com/bitwarden/clients/pull/19151/changes#r2879212870) need actioning?

apps/web/src/app/admin-console/organizations/policies/policies.component.ts
```diff
@@ -76,14 +71,43 @@ export class PoliciesComponent {
...
+            map((results) => ({
+              category: section.category,
+              labelKey: section.labelKey,
+              policies: results.filter((p): p is BasePolicyEditDefinition => p !== null),
+            })),
+          );
+        }),
+      ).pipe(map((sections) => sections.filter((s) => s.policies.length > 0))),
+    ),
+  );
```
eliykat, line 97: This is pretty difficult to understand, with lots of nesting, inner observables and ternary if statements. Can it be simplified?

apps/web/src/app/admin-console/organizations/policies/policy-list.service.ts
eliykat, line 1: I like how the service handles some of the basic sorting. Can you please add jsdoc to the public methods.

eliykat (changes requested):
This is looking really good, just some minor feedback.

apps/web/src/app/admin-console/organizations/policies/policies.component.html
JaredScar: I believe this happened as part of me trying to fix the design for Sukhleen via design review if I recall correctly

amorask-bitwarden (dismissed):

apps/web/src/app/admin-console/organizations/policies/policies.component.ts
```diff
@@ -77,37 +77,37 @@ export class PoliciesComponent {
     }),
   );
 
-  protected policies$: Observable<readonly BasePolicyEditDefinition[]> = defer(() =>
-    of(this.policyListService.getPolicies()),
-  );
-
   protected policySections$: Observable<PolicySection[]> = this.organization$.pipe(
```
eliykat: This is improved, but I think that "filter an array of items based on an observable that corresponds to each item" is just kinda hard to do well in rxjs. I think this is fine if you don't want to mess with it further, but maybe the best solution is the current one: consume the `sections` directly in the template and use an `if` statement to hide a policy that shouldn't be displayed.

eliykat (dismissed):
LGTM but up to you whether you action the below.

apps/web/src/app/admin-console/organizations/policies/policies.component.html
eliykat: @willmartian Can you please check me here - do our typography directives include spacing/padding that we should adhere to? Or is it reasonable to apply these additional TW classes?

apps/web/src/app/admin-console/organizations/policies/policies.component.html
```diff
@@ -1,36 +1,57 @@
...
+        <h2 bitTypography="h2" class="tw-py-3 tw-pl-3 !tw-mb-0">
+          {{ section.labelKey | i18n }}
+        </h2>
+        <bit-table>
+          <ng-template body>
+            @for (p of section.policies; track p.type) {
+              <tr bitRow>
+                <td bitCell>
+                  <div class="tw-flex tw-items-center tw-justify-between tw-gap-4">
+                    <div class="tw-min-w-0 tw-max-w-[740px]">
```
eliykat: Similar questions here - I think the outer div is OK (it arranges items within the cell) but setting a max pixel width seems wrong - what's the correct (responsive) way to handle this?

JaredScar: @willmartian

apps/web/src/app/admin-console/organizations/policies/policies.component.html
willmartian: Yes, these should use the [section](https://components.bitwarden.com/?path=/docs/component-library-section--docs) and typography components, without css overrides. CSS overrides on components you don't own are brittle--who knows how they will break as we modify components?

Could you share a screenshot with these applied and the overrides removed? cc @sukhleenb 

Depending on the scope of the changes, we could update those shared component styles here, or create an issue for UIF to sync with design and do it in parallel.

sukhleenb: I'd agree there's no need to override the current styling and am okay using the section styling to storybook! Attached a screenshot. 

Screenshot 2026-03-11 at 11 35 52 AM

JaredScar: image
Update with changes implemented @sukhleenb @willmartian

willmartian (commented):
I synced with @sukhleenb, and we agree that this can actually use the `bit-item` component. That should greatly reduce the bespoke UI code needed in this PR. 🥳  

I vibecoded a version here: https://github.com/bitwarden/clients/pull/19507/changes

Story demo: https://62a88a6de5b807fa98886113-uorydnkicx.chromatic.com/?path=/story/component-library-item--policies-list

Also, if y'all feel so inclined, I think this would be a great thing to include in Storybook as a simple UI component! `<ac-policies-list>` etc.

eliykat (commented):
Please use `bit-item` per @willmartian 's recommendation above - just clearing my review request until then.

You can implement the presentational component and story in a follow-up PR.

JaredScar:
image
@sukhleenb @eliykat @willmartian Here is the update using bit-item

apps/web/src/app/admin-console/organizations/policies/policies.component.html
sukhleenb: @willmartian I should have caught this earlier, but can we update the policy row left and right padding to 12px (from 16px) so the text aligns with the H2 above it?

eliykat (dismissed):
LGTM provided that @sukhleenb approves. Thanks @JaredScar !

Thomas-Avery (dismissed):
KM changes look good.

amorask-bitwarden (dismissed):

Thomas-Avery (approved):

amorask-bitwarden (approved):

eliykat (approved):

PR #19160, opened 2026-02-24
[PM-31192] - [Defect] Deleting item -Scroll does not persist when returning to Vault tab

jaasen-livefront:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-31192

## 📔 Objective

This PR fixes an issue where scrolling wasn't being set back to the top of the vault after deleting an item from the edit form in chrome and in the view and edit form in firefox.

## 📸 Screenshots

Firefox

https://github.com/user-attachments/assets/71fea188-b33f-42a2-8969-9181f4e3b6ff

Chrome

https://github.com/user-attachments/assets/cada2564-3a92-488a-9ca9-46316a7097a1


jaasen-livefront:
> Nice! I'm glad we got to this solution. There is a slight "jump" back from the bottom of the list to the top on FireFox that I noticed. Are you seeing the same?
> 
> Based on our chat yesterday I don't think there is anything we can do about that

@nick-livefront Indeed, if we reset the scroll position via `this.scrollPositionService.stop(true);` it triggers FF to autoscroll after navigation which breaks the intended behavior so I think this is the best we can expect.

PR #19189, opened 2026-02-24
[PM-31406] fix: TypeScript 5.9 type compatibility fixes for tools-owned code

quexten:

## Summary
Migrates to uint8 array for typescript 5.9 compatibility. See https://www.typescriptlang.org/docs/handbook/release-notes/typescript-5-9.html#notable-behavioral-changes

Please note, there is also a second PR in tools-owned code here: https://github.com/bitwarden/clients/pull/19221


quexten:
Noting, there is also https://github.com/bitwarden/clients/pull/19221 which is separate.

mcamirault (approved):

PR #19196, opened 2026-02-24
[PM-31406] Narrow types for platform utils service

quexten:
## 🎟️ Tracking

Narrow the return type to hint to the ts compiler that the returned uint8array has the generic type parameter (and thus underlying array type) arraybuffer, excluding the sharedarraybuffer, preventing consumers having to type-assert.


quexten:
Ooops, pushed to the wrong branch.

quexten:
@dani-garcia; I've undone the accidentally pushed changes here and reverted to the version you approved. I would appreciate a re-review.

PR #19200, opened 2026-02-24
[PM-32642] Hide delete button for users without deletion permissions

nick-livefront:
## 🎟️ Tracking

[PM-32642](https://bitwarden.atlassian.net/browse/PM-32642)

## 📔 Objective

When the archive buttons were introduced the logic to show the right aligned footer actions properly considered the `delete` permissions but the logic wasn't added to hide the individual delete button when the user can not delete a cipher. This is applicable when an organization has the  `Restrict item deletion to members with the Manage Collection permission`

## 📸 Screenshots

|Non-manage collections|Manage collections|
|-|-|
|<video src="https://github.com/user-attachments/assets/c6c789c8-5f32-47da-b11e-7faebb83a17e" />|<video src="https://github.com/user-attachments/assets/cd752f51-bca4-4522-82d8-2bd2f9107e66"/>

[PM-32642]: https://bitwarden.atlassian.net/browse/PM-32642?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/desktop/src/vault/app/vault/item-footer.component.spec.ts
+++ b/apps/desktop/src/vault/app/vault/item-footer.component.spec.ts
@@ -1,0 +1,172 @@
+import { ComponentFixture, TestBed } from "@angular/core/testing";
+import { By } from "@angular/platform-browser";
+import { mock } from "jest-mock-extended";
+import { of } from "rxjs";
+
+import { JslibModule } from "@bitwarden/angular/jslib.module";
+import { AccountService } from "@bitwarden/common/auth/abstractions/account.service";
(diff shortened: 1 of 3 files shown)


apps/desktop/src/vault/app/vault/item-footer.component.spec.ts
```diff
@@ -0,0 +1,172 @@
+import { ComponentFixture, TestBed } from "@angular/core/testing";
```
gbubemismith, line 1: Thanks for the tests

PR #19206, opened 2026-02-24
[PM-31406] Extract dangerous crypto functions from crypto function service

quexten:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-31406 requires migrating crypto function service to use Uint8Array correctly / in a compatible way to TS 5.9 / TS 7. However, instead of migrating it, I decided to extract out the logic and cleanly re-write it.

The re-implementation is aimed at the special purpose dangerous use-cases, is cleanly documented and tested, and compatible with TS5.9 / TS7.


quexten:
Closing to replace with smaller set of PR's.

PR #19208, opened 2026-02-24
[PM-32249] Electron custom file protocol

dani-garcia:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-32249

## 📔 Objective

The desktop application is currently loading files using the privileged `file://` scheme. If the application has an XSS vulnerability, the attacker could use this to read arbitrary files on the user's machine. To solve this, we have to define a custom scheme and a scheme handler for it, which gives us control over what files are and aren't accessible. The problem is that one of those privileges that we lose is the ability to ignore CORS, so now server has to specifically allow CORS from the custom scheme origin for the desktop app to be able to make requests.

This PR implements the client side custom protocol, disabled by default under a environment variable as we need to wait for self hosted users to be updated before we can enable this by default.

This new CORS value is implemented on the server here: https://github.com/bitwarden/server/pull/7080
Some more info on electron's page: https://www.electronjs.org/docs/latest/tutorial/security#18-avoid-usage-of-the-file-protocol-and-prefer-usage-of-custom-protocols


coltonhurst (approved):
The only change under `team-autofill-desktop-dev` ownership is in `apps/desktop/src/autofill/main/main-desktop-autotype.service.spec.ts` which looks fine; approving for this file!

addisonbeck (approved):

coltonhurst:
@dani-garcia are you still working on this one?

@kmancusi0105 I see your request for review, I approved previously (and I still can again) but there is a merge conflict that has to be resolved!

dani-garcia:
@coltonhurst Yeah I have to look into a bug found during testing, but haven't had the time yet. I'll fix the conflicts, thanks for the heads up!

addisonbeck:
@dani-garcia sorry for losing track of this.

addisonbeck (approved):

coltonhurst (approved):
DN owned `apps/desktop/src/autofill/main/main-desktop-autotype.service.spec.ts` looks good 👍

PR #19209, opened 2026-02-24
[PM-32612] Only show subscription menu option when premium user has subscription

amorask-bitwarden:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-32612

## 📔 Objective

Fixes a redirect loop on the Subscription page for premium users whose `AccountBillingClient.getSubscription()` returns `null` (no subscription record). Previously, `canViewSubscription$` on `BillingAccountProfileStateService` controlled whether the Subscription nav link was visible, using a combination of premium status flags and a billing-history API call. This approach showed the link even when no subscription existed, causing the subscription page to redirect to the premium upgrade page, which then redirected back — creating an infinite loop.

### Changes

**Moved subscription visibility logic out of `BillingAccountProfileStateService` and into the consuming components:**

- **`user-layout.component.ts`** — `showSubscription$` now combines `hasPremiumPersonally$`, `hasPremiumFromAnyOrganization$`, and a new `hasSubscription$` (via `AccountBillingClient.getSubscription()`). The Subscription nav link is hidden when the user has org-provided premium without personal premium, or when no subscription record exists.

- **`cloud-hosted-premium.component.ts`** — The redirect logic now requires both `hasPremiumPersonally` and `hasSubscription` before navigating to the subscription page. Users with personal premium but no subscription record will see the upgrade page instead of entering a redirect loop.

**Removed `canViewSubscription$` from `BillingAccountProfileStateService`:**

- Removed the abstract method declaration, implementation, and all associated tests
- Removed `PlatformUtilsService` and `ApiService` dependencies from the service (no longer needed)
- Updated DI configuration in `jslib-services.module.ts`

### DIFF
--- a/apps/browser/src/background/main.background.ts
+++ b/apps/browser/src/background/main.background.ts
@@ -948,8 +948,6 @@ export default class MainBackground {
 
     this.billingAccountProfileStateService = new DefaultBillingAccountProfileStateService(
       this.stateProvider,
-      this.platformUtilsService,
-      this.apiService,
     );
 
     this.restrictedItemTypesService = new RestrictedItemTypesService(
--- a/apps/cli/src/service-container/service-container.ts
+++ b/apps/cli/src/service-container/service-container.ts
@@ -737,8 +737,6 @@ export class ServiceContainer {
 
     this.billingAccountProfileStateService = new DefaultBillingAccountProfileStateService(
       this.stateProvider,
-      this.platformUtilsService,
-      this.apiService,
     );
 
     this.taskSchedulerService = new DefaultTaskSchedulerService(this.logService);


apps/web/src/app/layouts/user-layout.component.ts
```diff
@@ -75,10 +74,44 @@ export class UserLayoutComponent implements OnInit {
...
+      }),
+    );
   }
 
   async ngOnInit() {
     document.body.classList.remove("layout_frontend");
     await this.syncService.fullSync(false);
   }
+
+  private ifAccountExistsCheck(predicate$: (userId: UserId) => Observable<boolean>) {
```
sbrown-livefront, line 112: 🤌 I like this! Super helpful for some of our reused account calls helper function-like.

PR #19215, opened 2026-02-25
[PM-32697] Fix excessive phishing blocklist CDN downloads

AlexRubik:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-32697

## 📔 Objective

Fix excessive CDN downloads in phishing detection. The browser extension was re-downloading the full phishing blocklist on every MV3 service worker restart instead of once per day. At 5% rollout this was generating 20-25 TB/hour of CDN egress.

### Root cause

The primary root cause was the use of RxJS `iif` in `_backgroundUpdate`. The `iif` operator **eagerly evaluates both branch observables** at subscription time, even though only one branch is subscribed to based on the condition. Since `_updateFullDataSet()` constructs a `Request` and calls `nativeFetch` as part of its observable creation, the CDN fetch was being triggered on every update cycle regardless of whether `checksumChanged` or `appVersionChanged` was true. The same issue affected the `_updateDailyDataSet()` branch.

This is resolved in commit [`1bed3a3`](https://github.com/bitwarden/clients/commit/1bed3a393d) which replaces `iif` with `if/else` inside `concatMap`, ensuring the fetch observables are only created when the conditions are met.

### Additional issues addressed

Beyond the `iif` root cause, several other issues compounded the excessive download frequency:

1. **No rate limit** — `_backgroundUpdate` had no guard preventing repeated runs. Every MV3 service worker restart triggered a full update cycle.
2. **Metadata race condition** — `first()` could read `undefined` before the persisted disk value loaded via `chrome.storage.local`, causing the service to treat every restart as a first-run scenario.
3. **`share()` reset on re-subscription** — lock/unlock and account switch caused `phishingDetectionActive$` to toggle, resetting `share()` and firing `startWith(undefined)` again.
4. **Stale daily delta** — The `todayUrl` (`phishing-links-NEW-today.txt`) hasn't been updated in ~2 months, making `_updateDailyDataSet()` dead code that fetched a stale file on every cycle.
5. **Metadata not always saved** — The timestamp was only persisted when data was actually downloaded, leaving it stale in code paths where no download occurred.

### Fixes (by commit)

1. [`1d9c94d`](https://github.com/bitwarden/clients/commit/1d9c94db9c) — Add 24h rate-limit guard at top of `_backgroundUpdate`
2. [`b4bd6e1`](https://github.com/bitwarden/clients/commit/b4bd6e1305) — Wait for persisted state before reading metadata (`first(meta => meta !== undefined)`)
3. [`4116a29`](https://github.com/bitwarden/clients/commit/4116a2969f) — Replace `share()` with `shareReplay({ bufferSize: 1, refCount: false })`
4. [`f3c2989`](https://github.com/bitwarden/clients/commit/f3c2989eeb) — Remove stale daily delta mechanism from pipeline
5. [`1bed3a3`](https://github.com/bitwarden/clients/commit/1bed3a393d) — Replace `iif` with `if/else` inside `concatMap` (fixes the primary root cause); simplify pipeline; always save metadata; use checksum to skip CDN download on quiet days
6. [`05ef70d`](https://github.com/bitwarden/clients/commit/05ef70df9f) — Trim checksum response whitespace, fix `fetchPhishingChecksum` default parameter
7. [`49f1728`](https://github.com/bitwarden/clients/commit/49f1728fca) — Rewrite tests for new update logic

## 📸 Screenshots

Not applicable — no UI changes.

### DIFF
--- a/apps/browser/src/dirt/phishing-detection/services/phishing-data.service.spec.ts
+++ b/apps/browser/src/dirt/phishing-detection/services/phishing-data.service.spec.ts
@@ -296,131 +296,159 @@ describe("PhishingDataService", () => {
     });
   });
 
-  describe("phishing meta data updates", () => {
-    it("should not update metadata when no data updates occur", async () => {
-      // Set up existing metadata
-      const existingMeta = {
-        checksum: "existing-checksum",
-        timestamp: Date.now() - 1000, // 1 second ago (not expired)
-        applicationVersion: "1.0.0",
-      };
-      await fakeGlobalStateProvider.get(PHISHING_DOMAINS_META_KEY).update(() => existingMeta);
-
-      // Mock conditions where no update is needed
-      fetchChecksumSpy.mockResolvedValue("existing-checksum"); // Same checksum
-      platformUtilsService.getApplicationVersion.mockResolvedValue("1.0.0"); // Same version
-      const mockResponse = {
-        ok: true,
-        body: {} as ReadableStream,
-      } as Response;
-      apiService.nativeFetch.mockResolvedValue(mockResponse);
-
-      // Trigger background update
-      const result = await firstValueFrom(service["_backgroundUpdate"](existingMeta));
-
-      // Verify metadata was NOT updated (same reference returned)
-      expect(result).toEqual(existingMeta);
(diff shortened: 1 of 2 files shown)


AlexRubik:
duplicate

PR #19220, opened 2026-02-25
[PM-31406] Introduce dangerous-compat crypto implementations

quexten:
## 🎟️ Tracking

## 📔 Objective

Instead of migrating rather complex crypto code in cryptofunctionservice, this PR introduces a cleanly re-written version of the special, dangerous use-cases for external compatibility. ECB and CBC without hmac, and CBC with no padding should not be part of our public API. This keeps them in a separate module, with a DANGEROUS prefix to make it abundantly clear to callers that they should not be used.

The are also re-written to be more comprehensible than the cryptofunctionservice version.

A subsequent PR will drop the cryptofunction service functions, removing it as public API. And another set of PR's will migrate the callers over to this.

For context, these are the functions we are dropping: https://github.com/bitwarden/clients/pull/19222

### DIFF
--- a/apps/desktop/src/app/shared/shared.module.ts
+++ b/apps/desktop/src/app/shared/shared.module.ts
@@ -12,6 +12,12 @@ import { IconModule } from "@bitwarden/components";
 import { AvatarComponent } from "../components/avatar.component";
 import { ServicesModule } from "../services/services.module";
 
+/**
+ * @deprecated Please directly import the relevant directive/pipe/component.
+ *
+ * This module is overly large and adds many unrelated modules to your dependency tree.
+ * https://angular.dev/guide/ngmodules/overview recommends not using `NgModule`s for new code.
+ */
 @NgModule({
   imports: [
     CommonModule,
--- a/apps/web/src/app/shared/shared.module.ts
+++ b/apps/web/src/app/shared/shared.module.ts
@@ -35,11 +35,10 @@ import {
 } from "@bitwarden/components";
 
 /**
- * This NgModule should contain the most basic shared directives, pipes, and components. They
- * should be widely used by other modules to be considered for adding to this module. If in doubt
- * do not add to this module.
+ * @deprecated Please directly import the relevant directive/pipe/component.
  *
- * See: https://angular.io/guide/module-types#shared-ngmodules
+ * This module is overly large and adds many unrelated modules to your dependency tree.
+ * https://angular.dev/guide/ngmodules/overview recommends not using `NgModule`s for new code.
  */
 @NgModule({
   imports: [
--- a/libs/common/src/services/audit.service.spec.ts
+++ b/libs/common/src/services/audit.service.spec.ts
@@ -21,7 +21,7 @@ describe("AuditService", () => {
 
   beforeEach(() => {
     mockCrypto = {
-      hash: jest.fn().mockResolvedValue(Buffer.from("AABBCCDDEEFF", "hex")),
+      hash: jest.fn().mockResolvedValue(new Uint8Array([0xaa, 0xbb, 0xcc, 0xdd, 0xee, 0xff])),
     } as unknown as jest.Mocked<CryptoFunctionService>;
 
     mockApi = {


libs/common/src/key-management/crypto/dangerous/dangerous_duckduckgo_crypto.ts
```diff
@@ -0,0 +1,84 @@
+import * as forge from "node-forge";
```
quexten, line 1: Noting: Webcrypto does not support ECB mode. We could think about moving to rust if we want to get rid of node forge in the long term.

libs/common/src/key-management/crypto/dangerous/dangerous_duckduckgo_crypto.ts
```diff
@@ -0,0 +1,84 @@
...
+// ref: https://paragonie.com/blog/2015/11/preventing-timing-attacks-on-string-comparison-with-double-hmac-strategy
+function compareConstantTime(a: Uint8Array, b: Uint8Array): boolean {
+  const rand = self.crypto.getRandomValues(new Uint8Array(32));
+
+  const hmac = forge.hmac.create();
+  hmac.start("sha256", Utils.fromArrayToByteString(rand));
+  hmac.update(Utils.fromArrayToByteString(a));
+  const mac1 = hmac.digest().getBytes();
+
+  hmac.start("sha256", Utils.fromArrayToByteString(rand));
```
Thomas-Avery, line 71: Previous implementation looks to use null here for the second input? 

https://github.com/bitwarden/clients/blob/4c2aec162d0d79b3fb2d53dca7c64cf1ac0a6cad/libs/common/src/key-management/crypto/services/web-crypto-function.service.ts#L178

libs/common/src/key-management/crypto/dangerous/dangerous_duckduckgo_crypto.ts
```diff
@@ -0,0 +1,84 @@
...
+
+  // Calculate the MAC and compare it. If the mac does not match, throw an error.
+  // Note: The mac is over both IV and data
+  const hmac = forge.hmac.create();
+  hmac.start("sha256", Utils.fromArrayToByteString(key.authenticationKey)!);
+  hmac.update(Utils.fromArrayToByteString(concatUint8Arrays(iv, data)));
+  const expectedMac = Utils.fromByteStringToArray(hmac.digest().getBytes());
+  if (!compareConstantTime(mac, expectedMac)) {
+    throw new Error("MAC verification failed.");
+  }
```
Thomas-Avery, line 33: ⚠️  This is a change in behavior previously it would return null.

https://github.com/bitwarden/clients/blob/4c2aec162d0d79b3fb2d53dca7c64cf1ac0a6cad/apps/desktop/src/services/duckduckgo-message-handler.service.ts#L266-L269

libs/common/src/key-management/crypto/dangerous/dangerous_duckduckgo_crypto.ts
quexten: @Thomas-Avery Key == null re-uses the previous key. If you think about it, a different key being used for hmac would actually make the check fail. Both values need to be hmac'd under the same key to produce equal outputs.

I've explicitly chosen to pass in the key here, as it is equivalent, but way easier to understand for the reader. For the null value, you have to look up how forge crypto's API works first, and have to look if the key value of the hmac object was modified in the meantime, which is way more confusing.

libs/common/src/key-management/crypto/dangerous/dangerous_duckduckgo_crypto.ts
quexten: Good call-out. I think this is the appropriate place to make a deliberate breaking change, since this is a new function. The subsequent PR will make sure to handle the error value appropriately in the caller's code.

quexten: Actually I just verified, the duckduckgo implementation catches the errors returned by the function, so no changes are needed to the caller.

Thomas-Avery: I'm good as long as we are making sure the caller is maintaining the in place contract.

PR #19222, opened 2026-02-25
[PM-31406] Migrate cryptofunction service and remove old crypto

quexten:
## 🎟️ Tracking

https://github.com/bitwarden/clients/pull/19220
https://github.com/bitwarden/clients/pull/19223
https://github.com/bitwarden/clients/pull/19221

## 📔 Objective

Removes now unused crypto functions so that we do not have to migrate them.

## 📸 Screenshots

image


libs/node/src/services/node-crypto-function.service.ts
```diff
@@ -64,18 +54,18 @@ export class NodeCryptoFunctionService implements CryptoFunctionService {
...
       throw new Error("prk is too small.");
     }
-    const infoBuf = this.toUint8Buffer(info);
+    const infoBuf = typeof info === "string" ? Utils.fromUtf8ToArray(info) : info;
     const infoArr = new Uint8Array(infoBuf);
     let runningOkmLength = 0;
-    let previousT = new Uint8Array(0);
+    let previousT: Uint8Array<ArrayBuffer> = new Uint8Array(0);
     const n = Math.ceil(outputByteSize / hashLen);
     const okm = new Uint8Array(n * hashLen);
```
quexten, line 62: I'd really prefer us to drop this function / replace it with an external lib in rust... (Not in this PR).

libs/common/src/key-management/crypto/services/web-crypto-function.service.ts
```diff
@@ -328,7 +202,8 @@ export class WebCryptoFunctionService implements CryptoFunctionService {
     if (typeof value === "string") {
       bytes = forge.util.encodeUtf8(value);
     } else {
-      bytes = Utils.fromBufferToByteString(value);
+      // Null assertion is safe because this function takes a non-null value and is private.
+      bytes = Utils.fromArrayToByteString(this.toBuf(value))!;
```
Thomas-Avery:
```suggestion
      bytes = Utils.fromArrayToByteString(this.toBuf(value));
```

❓ Looking at these method signatures I don't see why we need the null assertion?

libs/node/src/services/node-crypto-function.service.ts
```diff
@@ -247,60 +124,19 @@ export class NodeCryptoFunctionService implements CryptoFunctionService {
...
-    const publicKey = forge.pki.publicKeyFromAsn1(asn1);
-    return forge.pki.publicKeyToPem(publicKey);
-  }
-
-  private toNodeCryptoAesMode(mode: "cbc" | "ecb"): string {
-    return mode === "cbc" ? "aes-256-cbc" : "aes-256-ecb";
+  private hmac(
+    value: Uint8Array,
+    key: Uint8Array,
+    algorithm: "sha1" | "sha256" | "sha512",
```
Thomas-Avery:
```suggestion
    algorithm: "sha256" | "sha512",
```

Looks like we can drop support for `sha1` like we did in web-crypto here too.

libs/node/src/services/node-crypto-function.service.ts
```diff
@@ -247,60 +124,19 @@ export class NodeCryptoFunctionService implements CryptoFunctionService {
...
+  ): Promise<Uint8Array> {
```
Thomas-Avery:
```suggestion
  ): : Promise<Uint8Array<ArrayBuffer>>  {
```
Optional 

💭  It looks like we can tighten the signature here and not have to do the casts above?

libs/common/src/key-management/crypto/services/web-crypto-function.service.ts
quexten: Left over from before merging https://github.com/bitwarden/clients/pull/19219

libs/node/src/services/node-crypto-function.service.ts
```diff
@@ -64,18 +54,18 @@ export class NodeCryptoFunctionService implements CryptoFunctionService {
...
+    let previousT: Uint8Array<ArrayBuffer> = new Uint8Array(0);
     const n = Math.ceil(outputByteSize / hashLen);
     const okm = new Uint8Array(n * hashLen);
     for (let i = 0; i < n; i++) {
       const t = new Uint8Array(previousT.length + infoArr.length + 1);
       t.set(previousT);
       t.set(infoArr, previousT.length);
       t.set([i + 1], t.length - 1);
-      previousT = await this.hmac(t, prk, algorithm);
+      previousT = (await this.hmac(t, prk, algorithm)) as Uint8Array<ArrayBuffer>;
```
Thomas-Avery:
```suggestion
      previousT = await this.hmac(t, prk, algorithm);
```

No need to cast anymore.

PR #19223, opened 2026-02-25
[PM-31406] Migrate duckduckgo to purpose-built dangerous-crypto-compat function

quexten:
## 🎟️ Tracking

https://github.com/bitwarden/clients/pull/19220

## 📔 Objective

Migrate to the special purpose functions introduced in https://github.com/bitwarden/clients/pull/19220

Note: This needs QA before merging.


quexten:
Had to bump the TS version since it was not compatible with Uint8array being generic.

coroiu (approved):

PR #19226, opened 2026-02-25
[PM-32623] Add support for new vault in encrypted migrator

quexten:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-32623

## 📔 Objective

The encrypted migrator triggers only on specific routes. Since the new desktop UI changed routes, the scheduled migrator no longer triggers. This fixes this by adding the new route to the migrator.

### DIFF
--- a/libs/common/src/key-management/user-key-mapper.ts
+++ b/libs/common/src/key-management/user-key-mapper.ts
@@ -1,0 +1,21 @@
+import { SdkRecordMapper } from "@bitwarden/common/platform/services/sdk/client-managed-state";
+import { UserKeyDefinition } from "@bitwarden/common/platform/state";
+import { UserKeyState } from "@bitwarden/sdk-internal";
+
+import { SymmetricCryptoKey } from "../platform/models/domain/symmetric-crypto-key";
+import { USER_KEY } from "../platform/services/key-state/user-key.state";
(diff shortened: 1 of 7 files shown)


libs/angular/src/key-management/encrypted-migration/encrypted-migrations-scheduler.service.ts
```diff
@@ -38,7 +38,7 @@ export const ENCRYPTED_MIGRATION_DISMISSED = new UserKeyDefinition<Date>(
   },
 );
 const DISMISS_TIME_HOURS = 24;
-const VAULT_ROUTES = ["/vault", "/tabs/vault", "/tabs/current"];
+const VAULT_ROUTES = ["/new-vault", "/vault", "/tabs/vault", "/tabs/current"];
```
quexten, line 41: I think we should track work to migrate away from this pattern as it's evidently not stable. (However, I also don't want to tie this into a bug-fix).

Thomas-Avery: Agreed.

quexten:
@Thomas-Avery failing tests were intermittent.

PR #19228, opened 2026-02-25
[PM-32717] Individual Vault - Edit Cipher permissions

nick-livefront:
## 🎟️ Tracking

[PM-32717](https://bitwarden.atlassian.net/browse/PM-32717)

## 📔 Objective

https://github.com/bitwarden/clients/pull/18375 removed the `disableForm` option which was set for admin console and reporting instances that use the vault item dialog. It _wasn't_ configured for the individual vault which always allowed users to access the edit screen of the cipher. Because this wasn't explicitly defined it was missed in the refactor.

~~I'm adding a `bypassCipherEditPermission` option that can be passed to the `cipherAuthorizationService` so it is explicitly defined and tested.~~ I ended up using the configuration mode in the form for `partial-edit` directly.

I want to add unit tests for the `vault.component` for this and others but that is a large implementation that I don't want to hold up this fix. I'll follow up.

## 📸 Screenshots

<video src="https://github.com/user-attachments/assets/ecc4796f-1ebf-4d1e-82e1-43eb3ef6e2ac"/>

[PM-32717]: https://bitwarden.atlassian.net/browse/PM-32717?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/common/src/vault/services/cipher-authorization.service.ts
+++ b/libs/common/src/vault/services/cipher-authorization.service.ts
@@ -59,7 +59,6 @@ export abstract class CipherAuthorizationService {
    *
    * @param {CipherLike} cipher - The cipher object to evaluate for edit permissions.
    * @param {boolean} isAdminConsoleAction - Optional. A flag indicating if the action is being performed from the admin console.
-   * @param {boolean} bypassCipherEditPermission - Optional. A flag indicating if edits should be allowed regardless of permissions, typically used to allow users with only view permissions to access the edit form for certain fields.
    *
    * @returns {Observable<boolean>} - An observable that emits a boolean value indicating if the user can edit the cipher.
    */


apps/web/src/app/vault/components/vault-item-dialog/vault-item-dialog.component.ts
```diff
@@ -97,6 +97,13 @@ export interface VaultItemDialogParams {
    */
   isAdminConsoleAction?: boolean;
 
+  /**
+   * When true, allows the user to access the edit form regardless of their permissions,
+   * on the cipher. A user with only `view` permissions still needs to be able to
+   * access the form to edit fields like folder or favorites.
+   */
+  bypassCipherEditPermission?: boolean;
```
shane-melton: 🎨 ❓ Instead of needing to pass in another dialog param. We should be able to make use of `this.formConfig.mode == 'partial-edit'` which should already be determined by the caller already.

nick-livefront: Great catch! [c1dfa6f](https://github.com/bitwarden/clients/pull/19228/commits/c1dfa6f52fd581ac5dd77c2aeaf8d794373dbca9)

apps/web/src/app/vault/components/vault-item-dialog/vault-item-dialog.component.ts
```diff
@@ -392,6 +392,7 @@ export class VaultItemDialogComponent implements OnInit, OnDestroy {
         this.cipherAuthorizationService.canEditCipher$(
           this.cipher,
           this.params.isAdminConsoleAction,
+          this.formConfig.mode == "partial-edit",
```
shane-melton: 🎨 Oh, I was thinking we could also remove it from the `cipherAuthorizationService` as well. We should only need to adjust the `disableEdit()` getter and the [section of init code that ](https://github.com/bitwarden/clients/blob/323f30c8e941d3bfb72acb812e6a90e949771a5d/apps/web/src/app/vault/components/vault-item-dialog/vault-item-dialog.component.ts#L398-L403) forces back to view mode that was added in the original PR.

shane-melton (approved):
Thanks for making those changes!

PR #19242, opened 2026-02-25
[PM-32765] [PM-32744] Default import destination to My Items when org data owner

vlad-trofimov:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-32744

## 📔 Objective

When the Enforce organization data ownership policy is active, the import tool now defaults the collection selector to the user's My Items collection, rather than leaving it blank. 

Background

The Enforce organization data ownership policy requires that all vault items be stored within a designated organization rather than a user's personal "My Vault". When this policy is in effect, the import tool correctly hides the "My Vault" destination option and auto-selects the user's first availableorganization. However, there is no default collection selected - the user must manually locate and select their My Items collection from the list.

My Items is a DefaultUserCollection (type = 1) provisioned per-user by the org when this policy is enabled. It is the purpose-built destination for this exact scenario, and is already accessible via CollectionService.defaultUserCollection$(userId, orgId) - which was introduced as part of the broader "Migrate My Vault to My Items" feature.

## 📸 Screenshots

https://github.com/user-attachments/assets/a8517c53-53ab-4b2f-9b51-37b65f5415ff


libs/common/src/enums/feature-flag.enum.ts
```diff
@@ -56,6 +56,7 @@ export enum FeatureFlag {
   UseSdkPasswordGenerators = "pm-19976-use-sdk-password-generators",
   SendUIRefresh = "pm-28175-send-ui-refresh",
   SendEmailOTP = "pm-19051-send-email-verification",
+  DefaultImportMyItems = "pm-32744-default-import-my-items",
```
harr1424: Per product, we won't be gating these changes behind any feature flag


```suggestion
```

libs/common/src/enums/feature-flag.enum.ts
```diff
@@ -123,6 +124,7 @@ export const DefaultFeatureFlagValue = {
   [FeatureFlag.UseSdkPasswordGenerators]: FALSE,
   [FeatureFlag.SendUIRefresh]: FALSE,
   [FeatureFlag.SendEmailOTP]: FALSE,
+  [FeatureFlag.DefaultImportMyItems]: FALSE,
```
harr1424:
```suggestion
```

libs/importer/src/components/import.component.ts
```diff
@@ -42,6 +42,8 @@ import { Organization } from "@bitwarden/common/admin-console/models/domain/orga
 import { AccountService } from "@bitwarden/common/auth/abstractions/account.service";
 import { getUserId } from "@bitwarden/common/auth/services/account.service";
 import { ClientType } from "@bitwarden/common/enums";
+import { FeatureFlag } from "@bitwarden/common/enums/feature-flag.enum";
```
harr1424:
```suggestion
```

libs/importer/src/components/import.component.ts
```diff
@@ -370,12 +373,16 @@ export class ImportComponent implements OnInit, OnDestroy, AfterViewInit {
...
+      this.organizations$,
+      this.configService.getFeatureFlag$(FeatureFlag.DefaultImportMyItems),
+    ])
       .pipe(takeUntil(this.destroy$))
-      .subscribe(([value, organizations]) => {
+      .subscribe(([value, organizations, flagEnabled]) => {
         this.organizationId = value !== "myVault" ? value : undefined;
 
-        if (!this._importBlockedByPolicy) {
+        if (!this._importBlockedByPolicy || flagEnabled) {
```
harr1424: These changes should be reverted since we won't be using a feature flag

libs/importer/src/components/import.component.ts
```diff
@@ -269,6 +271,7 @@ export class ImportComponent implements OnInit, OnDestroy, AfterViewInit {
     private restrictedItemTypesService: RestrictedItemTypesService,
     private destroyRef: DestroyRef,
     protected importMetadataService: ImportMetadataServiceAbstraction,
+    protected configService: ConfigService,
```
harr1424:
```suggestion
```

libs/importer/src/components/import.component.ts
```diff
@@ -395,27 +402,49 @@ export class ImportComponent implements OnInit, OnDestroy, AfterViewInit {
...
         ),
       ),
       this.organizations$,
-    ])
-      .pipe(takeUntil(this.destroy$))
-      .subscribe(([policyApplies, orgs]) => {
-        this._importBlockedByPolicy = policyApplies;
-        if (policyApplies && orgs.length == 0) {
-          this.formGroup.disable();
-        }
```
harr1424, line 412: ❓ Why is this being removed?

libs/importer/src/components/import.component.ts
```diff
@@ -395,27 +402,49 @@ export class ImportComponent implements OnInit, OnDestroy, AfterViewInit {
...
+          ([[policyApplies, orgs], flagEnabled]) => flagEnabled && policyApplies && orgs.length > 0,
+        ),
+        switchMap(([[, orgs]]) =>
+          this.collectionService.defaultUserCollection$(userId, orgs[0].id as OrganizationId),
+        ),
+        filter(Boolean),
+        takeUntil(this.destroy$),
+      )
+      .subscribe((defaultCollection) => {
+        this.formGroup.controls.targetSelector.setValue(defaultCollection);
```
harr1424: This looks good overall, but please revert the feature flag usage

harr1424 (changes requested):
Product has confirmed we will not be gating these changes behind a feature flag, please see comments regarding this, and one question regarding a removed section of code.

vlad-trofimov:
Removed feature gating.

The new feature requires observing both the policy state and organizations list (to auto-select the "My Items" collection). Rather than duplicating the `combineLatest` logic or mixing async/await into the subscribe callback, extracting `policyAndOrgs$` allows both subscriptions to share the same source observable

The `_importBlockedByPolicy` check was removed intentionally. The feature requires enabling the collection selector when the policy is active so users can see and modify the "My Items" default.

libs/importer/src/components/import.component.ts
harr1424: Per contributor: 

> The new feature requires observing both the policy state and organizations list (to auto-select the "My Items" collection). Rather than duplicating the combineLatest logic or mixing async/await into the subscribe callback, extracting policyAndOrgs$ allows both subscriptions to share the same source observable

>The _importBlockedByPolicy check was removed intentionally. The feature requires enabling the collection selector when the policy is active so users can see and modify the "My Items" default.

harr1424:
@vlad-trofimov thanks for providing the clarification and for removing the feature flag. 

✅  The code changes look good, I'd like to spend some time testing this locally before approving, but I don't think any additional changes will be necessary. I may reach out with more questions or request changes if anything comes up. 

⛏️ The PR description still describes this functionality as being gated behind a feature flag, I recommend updating this to avoid confusion. The code in this file is complex and poorly documented (pre-dating your changes), if you are open to adding comments that better explain what is happening here, I think it would go a long ways.

vlad-trofimov:
additional comments added / adjusted language on some existing comments. feel free to change as necessary

harr1424 (approved):

PR #19248, opened 2026-02-26
[PM-32770] Fix file picker crash on SSH key download in popup

### LINKED ISSUE #19137: Browser crashing when trying to download ssh key attachment
### Steps To Reproduce

1. Go to 'browser extension'
2. Click on 'stored ssh keys'
3. Scroll down to 'attachments stored for chosen key'
4. Click on 'desired attachment, e.g. stored private ssh key file'

### Expected Result

File dialog should open, so I can choose where to save attachment.

### Actual Result

On clicking download next to attachment, whole browser instantly crashes and closes.

### Screenshots or Videos

_No response_

### Additional Context

I understand that I am using 3rd party server, VaultWarden. I believe this bug is caused by browser extension due to (according to logs and my understanding of the issue) it has to do with Wayland directly. 

Uploading of attachments works without issues. Seems like only downloading due to issue with Wayland is the problem. 

To understand what is happening, I ran brave-browser in debug mode, and only thing it left were following lines:

`[8081:8081:0223/151822.618181:ERROR:ui/events/platform/wayland/wayland_event_watcher.cc:47] libwayland: zxdg_exporter_v2#21: error 0: exported surface had an invalid role

[0223/151822.643489:ERROR:third_party/crashpad/crashpad/snapshot/elf/elf_dynamic_array_reader.h:64] tag not found
[0223/151822.645783:ERROR:third_party/crashpad/crashpad/snapshot/elf/elf_dynamic_array_reader.h:64] tag not found
[0223/151822.668574:ERROR:third_party/crashpad/crashpad/util/file/directory_reader_posix.cc:43] opendir /home/nix/.config/BraveSoftware/Brave-Browser/Crash Reports/attachments/0db3be22-147d-4b3e-80c8-ae373895dc08: No such file or directory (2)
/usr/bin/brave-browser: line 40:  8081 Trace/breakpoint trap   (core dumped) "$HERE/brave" "$@"
`

### Operating System

Linux

### Operating System Version

Ubuntu 24.04

### Web Browser

Brave

### Browser Version

1.87.190

### Environment Versions

Version: 2026.1.1

SDK: 'main (8e5efd8)'

Server version (Third-party): 2025.12.0

Connected to third-party server implementation, Vaultwarden. Please verify bugs using the official server, or report them to the third-party server.

### Issue Tracking Info

Sebastianhayashi:
## 🎟️ Tracking

Resolves #19137

## 📔 Objective

Downloading file attachments from the `view-cipher` detail view in the browser extension popup crashes Chromium-based browsers on Linux (Wayland) and macOS, and Firefox on all platforms.

**Root cause:** Opening a native OS file dialog (open/save) from an extension popup can cause Chromium/Firefox to tear down the popup while the dialog is active or pending, crashing the renderer. Bitwarden already mitigates this at the route level with [filePickerPopoutGuard()](file:///Users/microseyuyu/repo_rank/clients/apps/browser/src/tools/popup/guards/file-picker-popout.guard.ts#9-82) for routes like `/attachments`, `/export`, and `/add-send`. However, the recent vault-v2 UI update embedded `<app-attachments-v2-view>` directly into the `/view-cipher` modal, which correctly lacks the route guard (to avoid forcing a popout on every item view). This leaves download actions from within `view-cipher` unprotected.

**Fix:** This PR adds a service-level guard inside `BrowserFileDownloadService.download()` that mirrors the logic of the route-level [filePickerPopoutGuard()](file:///Users/microseyuyu/repo_rank/clients/apps/browser/src/tools/popup/guards/file-picker-popout.guard.ts#9-82). Before executing `a.click()`, the service checks whether a popout is required and, if so, redirects the user to a safe Popout window via `BrowserPopupUtils.openCurrentPagePopout(window)` instead of triggering the native dialog directly.

**Covered browsers (matching [filePickerPopoutGuard](file:///Users/microseyuyu/repo_rank/clients/apps/browser/src/tools/popup/guards/file-picker-popout.guard.ts#9-82)):**
| Browser | Condition | Behavior |
|---|---|---|
| Firefox (all platforms) | `!inPopout && !inSidebar` | → Popout |
| Chromium (Linux / macOS) | `!inPopout && !inSidebar` | → Popout |
| Chromium (Windows) | — | No change |
| Safari | — | Not affected (uses separate [downloadSafari](file:///Users/microseyuyu/repo_rank/clients/apps/browser/src/platform/popup/services/browser-file-download.service.ts#71-97) path) |

**Design rationale:**
- While file *upload* dialogs are the typical catalyst for this bug, the issue reporter confirmed that *downloading* attachments on Linux/Wayland triggers the exact same renderer crash — the native OS save dialog falls into the same popup lifecycle problem.
- The guard is scoped strictly to affected browser/OS combinations and only when running inside an extension Popup (`!inPopout && !inSidebar`), preventing UX regressions for Windows users, non-Chromium browsers, and Sidebar/Popout contexts.

**Verification note:**
This PR was developed by an external contributor without access to a Premium account or Vaultwarden instance (required for attachment upload/download). I request the team to verify using the following path:
1. Build the extension and load it unpacked in a Chromium browser on **Linux (Wayland)** or **macOS**, or in **Firefox**.
2. Log into an account with Premium features enabled (or connect to a Vaultwarden instance).
3. Ensure the browser is configured to **"Ask where to save each file before downloading"** (to force the native save dialog).
4. In the extension **Popup**, open an item with file attachments via `view-cipher`.
5. Click the **Download** button on an attachment.
6. **Expected:** The Popup does not trigger the native save dialog directly. Instead, it opens a Popout window to the same `view-cipher` screen (the Popup closes as a safe side-effect). Clicking Download again in the Popout triggers the OS save dialog without crashing.

## 📸 Screenshots

N/A — No UI changes; behavioral change only (automatic popout redirection).


Sebastianhayashi:
Hi team!
Happy to verify this myself — just need a Premium account to do so!

micahblut:
Hi @Sebastianhayashi, after reviewing the fix here, we would not like to accept this as is. A better fix would be for the download button to trigger the pop-out depending on the specific platform detected, so that we don't have the user downloading "twice"

Sebastianhayashi:
> Hi @Sebastianhayashi, after reviewing the fix here, we would not like to accept this as is. A better fix would be for the download button to trigger the pop-out depending on the specific platform detected, so that we don't have the user downloading "twice"

Hi @micahblut, thanks for reviewing the PR!
Having the user click the download button a second time once the pop-out opens may not so convenient. 
As what you said " A better fix would be for the download button to trigger the pop-out depending on the specific platform detected"

Here is a approach:
1. We move the platform/environment detection from the service level closer to the UI (where the download action originates).
2. If the user clicks "Download" on an affected platform, we trigger the pop-out of the current view, but we also pass along some state or parameter (e.g., the specific attachmentId).
3. Once the new pop-out window renders, it detects this state/parameter and automatically triggers the actual download right away.

Then user clicks the UI button once then download immediately.

And since you know the codebase best, do you have a preferred way we should pass this state between the popup and the new pop-out? Would appending a simple query parameter (like ?autoDownloadAttachmentId=xxx) to the route be acceptable, or would you prefer using an existing state-management/messaging service for this?

ngarafol:
> Hi team! Happy to verify this myself — just need a Premium account to do so!

Hi, if you still need premium - try family trial.

PR #19250, opened 2026-02-26
[PM-32773] Fix/issue 18842

### LINKED ISSUE #18842: Keyboard driven navigation with tab after search goes to sidebar menu instead of vault item list first
### Steps To Reproduce

1. Go to search bar and search for item
2. Press tab 2-3 times (I also don't think the tab should select the account box first either, but whatever)
3. A new hover box appears that says "skip to content"; if enter is then pressed it will jump to the item in the list, whereas before it would go to the item list.
4. Continue pressing tab and the selector goes to the left navigation bar instead of the item list. Cycling through the whole bar then takes you to the item list.
5. This is exceptionally inefficient and time consuming to users with keyboard driven workflows.

Linux snapd version 2026.1.0 rev 154 latest/stable.

### Expected Result

Pressing tab after a search jumps to the first item in the search list and cycles through all matching entries.

### Actual Result

Other menus such as the account menu and the side bar are selected before the item list when tabbing.

### Screenshots or Videos

Image

Image

### Additional Context

_No response_

### Operating System

Linux

### Operating System Version

debian 13.3

### Installation method

Snap

### Build Version

2026.1.0 rev 154

### Issue Tracking Info

Sebastianhayashi:
## 🎟️ Tracking

Closes #18842

## 📔 Objective

This PR resolves a keyboard accessibility bug where focus dropped unexpectedly when a user performed a query in the Vault Search and then pressed [Tab](file:///Users/microseyuyu/repo_rank/clients/apps/browser/src/background/runtime.background.ts#480-488) to navigate to the results. Instead of focusing the filter or items list, focus leaped to the top-level "Skip to Content" link and the main side navigation.

**Root cause:** The [RouterFocusManagerService](file:///Users/microseyuyu/repo_rank/clients/libs/components/src/a11y/router-focus-manager.service.ts#11-82) uses Angular's `currentNavigation()` to evaluate the `focusAfterNav` state. However, on the `NavigationEnd` event hook, `currentNavigation()` evaporates to `null`, causing the search bar's intentional `focusAfterNav: false` parameter to be ignored. It defaulted to aggressively focusing the `<main>` container, pushing keyboard flow out of the inner vault component and forcing a wrap-around layout tab.

**Fix:** Modified `RouterFocusManagerService.start$` to actively intercept and cache `this.router.currentNavigation()?.extras?.state?.focusAfterNav` during the `NavigationStart` routing phase. This cached state is then safely consumed at `NavigationEnd`, guaranteeing that the original `focusAfterNav: false` parameter is accurately delivered and respected instead of evaluating to null and moving focus to `<main>`.

### ❌ Before (Bug: Focus leaps out of Search to top-left Side Navigation)

bug_focus_highlighted

<br/>

### ✅ After (Fix: Focus remains natively in Search and flows into the "Clear Search" and Vault List naturally)

fix_focus_highlighted


vleague2:
Hi @Sebastianhayashi, thanks for opening a PR! I am looking at the issue you linked -- the `router-focus-manager` service is not used on the desktop app, which is what the linked issue is about. Is there another issue on the web app that you are looking to solve with those code changes?

I also see some changes to local environment setup, which may have been committed by mistake, and some other commits looking like they are related to other issues -- can you speak to those changes more?

willmartian:
@Sebastianhayashi This PR will be closed in 14 days if the above is not addressed. Thank you!

PR #19251, opened 2026-02-26
[PM-32784] Increase search service search performance by 50x and defer indexing

### LINKED ISSUE #17704: Bitwarden chrome extension takes a while to authenticate me
### Steps To Reproduce

1. Open Chromium browser.
2. Click the Bitwarden icon.
3. Put your master password in.
4. Press enter and wait.

### Expected Result

Tell me if my password is wrong, display my vault otherwise.

### Actual Result

Took 26 seconds to see if my password was correct and then displayed my vault.

### Screenshots or Videos

https://github.com/user-attachments/assets/f08db3eb-679c-41c1-9638-5da4833b07c4

### Additional Context

I'm using Helium here, but I've seen this issue in other Chromium browsers too.
Not Firefox though.

### Operating System

Linux

### Operating System Version

Arch Linux KDE Plasma 6.5 Wayland

### Web Browser

Chrome

### Browser Version

Hellium Browser v0.6.7.1

### Environment Versions

Version: 2025.11.1

SDK: 'main (8ef7951)'

Server version: 2025.11.1

### Issue Tracking Info

quexten:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-32784

## 📔 Objective

Tl;dr; speed up unlock by a few seconds and make search faster by 50x.

---

Search service is incredibly slow. There are two parts to this slowness, index and search.

We use the lunr search engine to support special queries. These generally start with `>`. Building the index and searching in lunr is slow.

This PR fixes slowness for most users by deferring indexing to when you actually do a LUNR query, and replacing basic-query-via-lunr by just doing a basic query. The basic-query-via-lunr used wildcard suffix + prefix, and wildcards are already incredibily slow (we can dig into parser construction here, but I think it suffices to say that wildcards are slower than linear substring search). basic-query-via-lunr is now entirely replaced by basic search.

LUNR is only used for LUNR queries starting with `>`. This affords us to index only exactly when the user actually does a LUNR query, which most users do not do.

A typical large vault takes 1-3 seconds to index during login and during every unlock. This drops indexing to 0msec for login and unlock.

A lunr query now does indexing the *first time* when it does not have an index, and is thus slower, all subsequent lunr queries remain the same speed as before.

Most searches are faster by orders of magnitude. In my testing for a 30k vault item test vault, my searching dropped from 300msec to 5msec, but this may vary depending on length and amount of names, subtitles, login uri's.

## LUNR Service Extraction
Please note that due to an issue that came up in QA, I increased scope and moved out all LUNR logic into a separate class, to make the code easier to reason about. Further, this changes behavior in that filter functions for advanced searches are moved to the consumer, and indexed entity ID is completely removed. 

The indexed entity id made the caching code quite complex and required rebuilds of the index. This instead always builds a full index, and then requires the consumer to filter out to the results they want (i.e the vault items for a vault).


MGibson1 (commented):
I tend to agree this is probably a better level of priority to set on advanced searches.

I do have a question though: How does this impact
1. Re-use of cipher search service for org vaults (I suspect the same improvements)
2. the serialized lunr index. It seems like that should be removed, but without touching it I'm worried it'll be deserialized and used in some situations without having been updated in the same way, resulting in stale search results

quexten:
@MGibson1
1. Could you clarify what you mean by re-use? The search just searches on all ciphers, org and personal.
2. I did not want to change more than is required here, but I agree, i'm not sure what we are using the serialized index for.

MGibson1:
1. Re-use meaning we use the same search service for personal vault (which searches org and personal ciphers) and org vault views, which are organization specific and search only organization vaults. This is done as a hack through the `indexedEntityId` parameter on `indexCiphers`. This is a separate use case I want to make sure we think about. I _think_ it's fine and has the same considerations, but it's been a while since I worked in this section of the codebase


2. I understand and agree with the sentiment, but I do at least want an opinion from @bitwarden/team-vault on the impact.

gbubemismith:
1. I think the same improvements apply.
2. I think we can work on removing the serialized lunr index in a follow up PR, I think the hitting stale index might be lower after this PR. @shane-melton I remembered you refactored this service sometime ago, can you confirm if it's safe to remove here or should we just have a follow-up PR

libs/common/src/vault/services/cipher.service.ts
```diff
@@ -177,10 +177,7 @@ export class CipherService implements CipherServiceAbstraction {
             return await this.decryptCiphersWithSdk(ciphers, userId, false);
           }),
           tap(([decrypted, failures]) => {
-            void Promise.all([
-              this.setFailedDecryptedCiphers(failures, userId),
-              this.searchService.indexCiphers(userId, decrypted),
-            ]);
+            void Promise.all([this.setFailedDecryptedCiphers(failures, userId)]);
```
quexten, line 183: note: I did not clean up here, but this can probably be cleaned up. I wanted to keep the changes as minimal as possible.

JimmyVo16 (dismissed):

libs/common/src/vault/services/search.service.ts
```diff
@@ -266,56 +260,49 @@ export class SearchService implements SearchServiceAbstraction {
...
+        if (await this.getIsIndexing(userId)) {
+          await new Promise((r) => setTimeout(r, 500));
+        }
+      }
+      let index = await this.getIndexForSearch(userId);
+      // If there is no index, build an index
+      if (index == null) {
+        this.logService.info("Building lunr index for search...");
+        await this.indexCiphers(userId, ciphers);
+        index = await this.getIndexForSearch(userId);
```
shane-melton: ⚠️  `index` is still potentially `null` after this for slow index builds. Can we still log an error and continue to fallback to "basic" search if the index fails to build/load in time. We may need to strip the leading `>` character in such a scenario.

libs/common/src/vault/services/search.service.ts
```diff
@@ -266,56 +260,49 @@ export class SearchService implements SearchServiceAbstraction {
...
+        await new Promise((r) => setTimeout(r, 250));
+        if (await this.getIsIndexing(userId)) {
+          await new Promise((r) => setTimeout(r, 500));
+        }
+      }
+      let index = await this.getIndexForSearch(userId);
+      // If there is no index, build an index
+      if (index == null) {
+        this.logService.info("Building lunr index for search...");
+        await this.indexCiphers(userId, ciphers);
```
shane-melton: ⚠️ This is going to cause issues on the Admin Console. The AC provides its own list of `ciphers` that are are unique from the ciphers in the user's sync.

If a user were to first perform a lunr search in the AC, an index would be built and cached for those org ciphers. Then if they go to their personal vault, the search service will see the existing index and attempt to use it for searches against their personal vault which will not turn up any results.

We should accept an optional `indexedEntityId` in `searchCiphers()` and if it doesn't matched the saved `searchIndexEntityIdState()` we should clear the existing index and continue from there. Which will force a rebuild of the index for the users individual vault (or vice versa).

Only the AC would need to pass in a value for `indexedEntityId` when searching, otherwise we can assume if the value is missing / `undefined` then its for the users personal vault.

libs/common/src/vault/abstractions/search.service.ts
```diff
@@ -17,11 +17,6 @@ export abstract class SearchService {
    * For short Lunr.js queries (starts with '>'), a valid search index must exist for the user.
```
shane-melton: ⛏️ (non-blocking): This comment is now out of date, as we no longer check for an existing search index for Lunr queries in the implementation and can be removed.

shane-melton (changes requested):
I see two potential issues, mostly around how the AC works with search service. The other is just being a bit more defensive for slow index builds.

libs/common/src/vault/services/search.service.ts
quexten: Nice catch. Added a comment to make this more discoverable in the future and added a fix that falls back to basic search.

libs/common/src/vault/services/search.service.ts
quexten: Ah, oops. I was not aware of this. I don't really like the design of a shared index; Ideally we would not need an index because we are really just searching a few MB of text (I think rewriting it without LUNR, and possibly porting to the SDK would make it so fast that we don't need indexes); If we do need it one, one option could be to always index the entire vault, but then do a cheap `.filter` on the remaining items.

Either way, I went with your suggestion here and made it so the index is rebuilt if the indexedEntityId does not match, since anything else would be a much larger effort that requires more planning.

libs/common/src/vault/abstractions/search.service.ts
quexten: Applied

libs/common/src/vault/services/search.service.ts
shane-melton: Perfect, thank you!

libs/common/src/vault/services/search.service.spec.ts
```diff
@@ -70,4 +79,73 @@ describe("SearchService", () => {
       expect(mockIndex$).not.toHaveBeenCalled();
     });
   });
+
+  describe("searchCiphers with indexedEntityId", () => {
```
shane-melton: 👍 Thanks for the tests!

shane-melton (dismissed):
Looks good! Thanks for those fixes! I think the failing lint is unrelated to your changes and might just be from `main`.

quexten:
@shane-melton We had an issue coming up during QA around the index not re-building correctly. I found the search service logic quite tangled and hard to work on since it's doing both basic search and lunr search. I've moved out the lunr search entirely, and simplified the logic a bit, dropping the indexed entity id concept and moving the filtering by indexed entity id to the caller. Let me know if you have concerns with this.

libs/common/src/vault/services/search.service.ts
```diff
@@ -160,165 +60,34 @@ export class SearchService implements SearchServiceAbstraction {
...
-    indexedEntityId?: IndexedEntityId,
   ): Promise<C[]> {
+    this.logService.info("Starting cipher search", { query });
     this._isCipherSearching$.next(true);
-    const results: C[] = [];
     const searchStartTime = performance.now();
-    if (query != null) {
-      query = SearchService.normalizeSearchQuery(query.trim().toLowerCase());
-    }
+    query = normalizeSearchQuery(query.trim().toLowerCase());
```
shane-melton, line 71: ⛏️ We dropped the `null` check on `query` and are calling `.trim()` unconditionally. Can we add the guard back or a `.?` accessor.

libs/common/src/vault/services/search.service.ts
```diff
@@ -86,70 +27,29 @@ export class SearchService implements SearchServiceAbstraction {
...
-    await this.searchIndexEntityIdState(userId).update(() => null);
-    await this.searchIndexState(userId).update(() => null);
-    await this.searchIsIndexingState(userId).update(() => null);
+  async ciphersUpdated(userId: UserId): Promise<void> {
+    await this.lunrSearchService.ciphersUpdated(userId);
   }
 
-  async isSearchable(userId: UserId, query: string | null): Promise<boolean> {
-    query = SearchService.normalizeSearchQuery(query);
+  async isSearchable(query: string | null): Promise<boolean> {
```
shane-melton, line 43: ⚠️ This no longer matches the abstract `isSearchable()` signature. It still has `userId` while this only has `query`.

libs/common/src/vault/services/lunr-search.service.ts
shane-melton, line 1: 👍 Really appreciate you pulling this out into its own service!

libs/common/src/vault/services/lunr-search.service.ts
```diff
@@ -0,0 +1,337 @@
...
+  }
+
+  async searchCiphers<C extends CipherViewLike>(
+    userId: UserId,
+    query: string,
+    ciphers: C[],
+  ): Promise<C[]> {
+    const results: C[] = [];
+    const searchStartTime = performance.now();
+    const index = await this.updateIndexForUser(userId, ciphers);
```
shane-melton: ⚠️  This completely rebuilds the Lunr index for every call to `searchCiphers()` meaning every query/keystroke for queries that start with `> ` is going to cause an index rebuild. Or perhaps I'm overlooking something?

I see a check for if an index operation is _in progress_, but nothing checks if one exists. And even so, we wouldn't know which list of ciphers the cached index belongs too. Most often this is the users sync'd ciphers, but for Admin Console, we can have a different set of ciphers that are not included in that user's sync (this is what the indexed entity id was solving for).

libs/common/src/vault/services/lunr-search.service.ts
```diff
@@ -0,0 +1,337 @@
...
+    // Currently have to ensure this is only done a single time. Lunr allows you to register a function
+    // multiple times but they will add a warning message to the console. The way they do that breaks when ran on a service worker.
+    if (!LunrSearchService.registeredPipeline) {
+      LunrSearchService.registeredPipeline = true;
+      //register lunr pipeline function
+      lunr.Pipeline.registerFunction(normalizeAccentsPipelineFunction, "normalizeAccents");
+    }
+  }
+
+  async ciphersUpdated(userId: UserId): Promise<void> {
```
shane-melton: ⛏️ (non-blocking): Can we rename this to something like `clearCachedIndex(userId)` to more accurately reflect what its doing.

apps/web/src/app/admin-console/organizations/collections/vault.component.ts
```diff
@@ -418,15 +411,14 @@ export class VaultComponent implements OnInit, OnDestroy {
...
 
           if (await this.searchService.isSearchable(userId, searchText)) {
-            return await this.searchService.searchCiphers<CipherView>(
+            const searchFilteredCiphers = await this.searchService.searchCiphers<CipherView>(
               userId,
               searchText,
-              [filterFunction],
               ciphers,
-              // We are searching on the organization ciphers, which requires keeping a separate index. The index is
-              // specified by the IndexedEntityId
```
shane-melton: 💭 Here specifically; the `ciphers` here will often include `Unassigned` ciphers and those that the admin has elevated access to (via the Owners/Admins access all ciphers setting). So we need a means to tell the search service that it needs to re-index these ciphers specifically/separately.

shane-melton (changes requested):
I like the extraction to the separate `LunrSearchService`! However, we're still going to have issues with the Admin Console and it looks like we are effectively ignoring the cached values (except in the case of an in progress indexing).

apps/web/src/app/admin-console/organizations/collections/vault.component.ts
quexten: Thank you. I'll be honest, I should have dug deeper into what the exact purpose for the indexed entity id is; I had made assumptions here that were not accurate.

quexten:
@shane-melton Thanks for the previous set of reviews. I've further simplified the lunr indexing service. Mainly:

- Initiating re-indexing externally
- Having one shared index state that gets mutated and replaced depending on the view the user is one

both seemed like brittle points, and have been replaced. Would you mind having another look?

jprusik (dismissed):
Approved for Autofill team concerns

ike-kottlowski (dismissed):

JimmyVo16 (dismissed):

libs/angular/src/vault/components/vault-items.component.ts
```diff
@@ -163,22 +163,25 @@ export class VaultItemsComponent<C extends CipherViewLike> implements OnDestroy
...
+            const restrictedTypeFilter = (cipher: CipherViewLike) =>
+              !this.restrictedItemTypesService.isCipherRestricted(cipher, restricted);
+
+            const searchFilteredCiphers = await this.searchService.searchCiphers(
+              userId,
+              null,
+              searchText,
+              allCiphers,
+            );
+            return searchFilteredCiphers.filter(restrictedTypeFilter).filter(filter);
```
shane-melton: :pick: `filter` is initially null (though it is usually immediately set to a value) and we lost the null guard that was in the old `searchCiphers` method.

apps/browser/src/vault/popup/services/vault-popup-items.service.ts
```diff
@@ -158,12 +158,9 @@ export class VaultPopupItemsService {
...
    * @private
    */
-  private _hasSearchText = combineLatest([
-    this.searchText$,
-    getUserId(this.accountService.activeAccount$),
-  ]).pipe(
-    switchMap(([searchText, userId]) => {
-      return this.searchService.isSearchable(userId, searchText);
+  private _hasSearchText = this.searchText$.pipe(
+    switchMap(([searchText]) => {
```
shane-melton: :warning: We're destructuring a `string` now instead of the array it was before; meaning we're always passing in a single character to `isSearchable` which will always return `false`.   We should just be able to remove the brackets.

libs/angular/src/tools/send/send.component.ts
```diff
@@ -92,11 +90,9 @@ export class SendComponent implements OnInit, OnDestroy {
         this.disableSend = policyAppliesToUser;
       });
 
-    combineLatest([this._searchText$, this.accountService.activeAccount$.pipe(getUserId)])
+    this._searchText$
       .pipe(
-        switchMap(([searchText, userId]) =>
-          from(this.searchService.isSearchable(userId, searchText)),
-        ),
+        switchMap(([searchText]) => from(this.searchService.isSearchable(searchText))),
```
shane-melton: :warning: Same here, left over destructuring that grabs the first character of the search text.

shane-melton (changes requested):
The new approach looks good to me! Only few small refactoring bugs we need to address.

libs/angular/src/vault/components/vault-items.component.ts
quexten: Applied!

apps/browser/src/vault/popup/services/vault-popup-items.service.ts
quexten: Interesting that these did not cause issues during tests. Applied.

libs/angular/src/tools/send/send.component.ts
quexten: Interesting that these did not cause issues during tests. Applied.

jprusik (dismissed):

theMickster:
Hello @quexten et al. 

A local multi-agent Claude Code review produced the following report (I'm field testing a forthcoming addition to the plugin marketplace).  Kindly review at your convenience and lemme know if the findings are legitimate or not. If **not**, it would be beneficial to know briefly what context Claude may be missing or misunderstanding to make better assessments of our code. 

<details>
<summary>
Code Review: [PM-32784] Increase search service search performance by 50x and defer indexing (#19251)
</summary>
**Date:** 2026-04-16 | **Reviewed by:** Claude Code (Opus)

## Summary

| Severity      | Count |
| ------------- | ----- |
| 🛑 Blocker    | 1     |
| ⚠️ Important  | 1     |
| ♻️ Refactor   | 1     |
| 💡 Suggestion | 3     |

Strong performance improvement PR — deferring Lunr indexing and replacing basic-query-via-Lunr with substring search is a sound architectural decision. The extraction of `LunrSearchService` cleanly separates concerns. One blocker around Trash/Archive search regression and one important lock safety issue should be addressed before merge.

## Findings

### 🛑 Blockers

#### searchCiphersBasic silently filters out deleted/archived ciphers — breaks Trash and Archive search

`libs/common/src/vault/services/search.service.ts:90`

  The new `searchCiphers` routes all non-`>` queries through `this.searchCiphersBasic(ciphers, query)` without passing the `deleted` or `archived` parameters. `searchCiphersBasic` defaults these to `false`, which filters out all deleted and archived ciphers:

  ```typescript
  searchCiphersBasic<C extends CipherViewLike>(
      ciphers: C[], query: string, deleted = false, archived = false,
  ) {
      return ciphers.filter((c) => {
          if (deleted !== CipherViewLikeUtils.isDeleted(c)) { return false; }
          if (archived !== CipherViewLikeUtils.isArchived(c)) { return false; }
          // ...
      });
  }
  ```

  **Impact:** When a user searches in the Trash or Archive view, the consumer's `createFilterFunction` pre-selects only deleted/archived ciphers, then passes them to `searchCiphers`. But `searchCiphersBasic` with `deleted=false` immediately filters all of them back out. Result: **zero search results in Trash and Archive views** for all regular (non-`>`) queries on web and desktop.

  This bug exists today as a rare edge case (when the Lunr index is null), but the PR makes `searchCiphersBasic` the primary code path for all regular queries, turning it into a guaranteed regression.

  **Suggested fix:** Remove the `deleted`/`archived` filtering from `searchCiphersBasic` (consistent with the PR's design of moving all filtering to consumers), or thread those parameters through from `searchCiphers`. The cleanest approach aligns with the PR's philosophy: let consumers handle all filtering.

  **Confidence:** 82/100
  </details>

### ⚠️ Important

#### Lock never released if buildCipherIndex throws — permanent deadlock of lunr searches

`libs/common/src/vault/services/lunr-search.service.ts:84-91`

  In `getOrCreateIndex`, the lock acquire and release are not wrapped in `try/finally`:

  ```typescript
  await this.aquireIndexLock();
  const index = await buildCipherIndex(ciphers);  // if this throws...
  this.lunrIndices.set(indexId, { ... });
  await this.releaseIndexLock();                    // ...this never executes
  ```

  If `buildCipherIndex` throws (malformed cipher data, out-of-memory, lunr builder error), `this.isIndexing` stays `true` permanently. All subsequent calls to `aquireIndexLock()` spin forever in the `while(this.isIndexing)` loop, deadlocking all `>` searches until the app is restarted.

  Unlike the old state-provider-based approach (which had `clearOn: ["lock", "logout"]` for implicit recovery), the new in-memory boolean has no recovery mechanism.

  **Suggested fix:**

  ```typescript
  await this.aquireIndexLock();
  try {
    const index = await buildCipherIndex(ciphers);
    this.lunrIndices.set(indexId, {
      lunrIndex: index,
      numberOfCiphers: ciphers.length,
      revisionDate: new Date(),
    });
    return index;
  } finally {
    await this.releaseIndexLock();
  }
  ```

  **Confidence:** 90/100
  </details>

### ♻️ Refactor

#### TOCTOU race in getOrCreateIndex — staleness check happens before lock acquisition

`libs/common/src/vault/services/lunr-search.service.ts:79-84`

  `isIndexUpToDate` is checked before the lock is acquired. In JavaScript's async model, two concurrent `>` queries can both observe a stale index, both enter the if-block, both wait for the lock, then both build the index sequentially — the second build is entirely redundant.

  For large vaults (10k+ items), Lunr index building takes hundreds of milliseconds, so this wastes significant time.

  **Suggested fix:** Re-check `isIndexUpToDate` after acquiring the lock (double-checked locking pattern):

  ```typescript
  await this.aquireIndexLock();
  try {
    if (this.isIndexUpToDate(indexId, ciphers)) {
      return this.lunrIndices.get(indexId)!.lunrIndex;
    }
    const index = await buildCipherIndex(ciphers);
    // ... store and return
  } finally {
    await this.releaseIndexLock();
  }
  ```

  **Confidence:** 85/100
  </details>

### 💡 Suggestions

#### Missing integration test for lunr query routing through SearchService.searchCiphers

`libs/common/src/vault/services/search.service.spec.ts`

  The new `search.service.spec.ts` tests `isSearchable` and basic search, but does not verify that a `>` prefixed query is routed to `LunrSearchService`. While `lunr-search.service.spec.ts` tests Lunr queries directly, the integration (SearchService dispatching to LunrSearchService) is untested. At minimum, one test verifying `>` queries delegate to `lunrSearchService.searchCiphers` would catch regressions in the routing logic.

  **Confidence:** 82/100
  </details>

#### Typo: method name `aquireIndexLock` should be `acquireIndexLock`

`libs/common/src/vault/services/lunr-search.service.ts:131`

  The method name `aquireIndexLock` is misspelled. The correct spelling is "acquire." This will appear in stack traces and IDE completions. The rest of the codebase consistently uses the correct spelling.

  **Confidence:** 95/100
  </details>

#### Typos in file-level comment — "querys" and "preformance"

`libs/common/src/vault/services/lunr-search.service.ts:1`

  Line 1 reads: `// Lunr search is used for advanced querys which most users do not use. It is preformance heavy and should only be built when needed.`

  "querys" → "queries", "preformance" → "performance".

  **Confidence:** 90/100
  </details>

## Reviewed and Dismissed

   <details><summary>🔍 8 initial findings dismissed after validation</summary>

   #### isSearchable lost internal normalization — fragile for direct callers
   `libs/common/src/vault/services/search.service.ts:48`
   **Dismissed because:** The new `isSearchable` DOES call `normalizeSearchQuery(query)` at line 48 — normalization was not removed. The finding's premise is incorrect.

   #### Org vault stale Lunr index when switching between org and individual vault
   `libs/common/src/vault/services/lunr-search.service.ts:54`
   **Dismissed because:** The PR introduces `makeIndexId(userId, organizationId)` which creates separate index keys per (userId, orgId) pair. The org vault passes `organizationId` to `searchCiphers` while the individual vault passes `null`, so separate indices are maintained. The concern is fully addressed by the PR's design.

   #### Filter contract change — not all consumers updated
   `libs/common/src/vault/abstractions/search.service.ts:25`
   **Dismissed because:** All consumers in the diff were verified to apply filters after calling `searchCiphers`: desktop vault, web individual vault, web org vault, vault-items.component, fido2.component, and vault-popup-items.service all correctly apply filters post-search.

   #### Clock-skew sensitivity in isIndexUpToDate staleness check
   `libs/common/src/vault/services/lunr-search.service.ts:112`
   **Original confidence:** 78/100
   **Dismissed because:** Theoretical concern. The comparison between local build time (`new Date()`) and server-generated `revisionDate` is unlikely to cause issues in practice since client clocks are NTP-synchronized. Worst case is an unnecessary rebuild (harmless) or briefly stale results (not a security issue).

   #### Comment about callers passing null to query parameter
   `libs/common/src/vault/services/search.service.ts:64`
   **Dismissed because:** Defensive null check with a comment explaining why. This is a pragmatic approach to handle runtime contract violations. Not a real issue a senior engineer would flag.

   #### LunrSearchService naming does not follow Default* prefix convention
   `libs/common/src/vault/services/lunr-search.service.ts:26`
   **Original confidence:** 75/100
   **Dismissed because:** `LunrSearchService` is an internal implementation detail composed within `SearchService`, not injected via DI. The `Default*` convention applies to classes that implement a shared abstraction and are DI-registered. The descriptive name is appropriate here.

   #### Orphaned LUNR_SEARCH_INDEXED_ENTITY_ID state key definition
   `libs/common/src/vault/services/search.service.ts:54`
   **Dismissed because:** The PR removes all three state key definitions (`LUNR_SEARCH_INDEX`, `LUNR_SEARCH_INDEXED_ENTITY_ID`, `LUNR_SEARCH_INDEXING`) along with the `IndexedEntityId` type. No orphaned state.

   #### IndexedEntityId type in guid.ts becomes dead code
   `libs/common/src/types/guid.ts:16`
   **Dismissed because:** The PR removes the `IndexedEntityId` type from `guid.ts`. The cleanup is already done.

   </details>
</details>

[PM-32784]: https://bitwarden.atlassian.net/browse/PM-32784?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

theMickster:
Secondarily, perhaps testing using the various presets found in the Seeder Utility will provide additional validation ensuring that the search enhancements function as expected across the board for the variety of data densities we see in the wild 🤔 🤷🏼

[util/Seeder/Seeds/docs/presets.md](https://github.com/bitwarden/server/blob/main/util/Seeder/Seeds/docs/presets.md)

quexten:
@theMickster The requested changes are reasonable enough.

> If buildCipherIndex throws (malformed cipher data, out-of-memory, lunr builder error), this.isIndexing stays true permanently. All subsequent calls to aquireIndexLock() spin forever in the while(this.isIndexing) loop, deadlocking all > searches until the app is restarted.

It is unclear when this can happen.

> Unlike the old state-provider-based approach (which had clearOn: ["lock", "logout"] for implicit recovery), the new in-memory boolean has no recovery mechanism.

This commentary doesn't make much sense though. Lock and logout happen much less frequently than clearing the process memory. All cases where you would lock and logout, would also already implicitly break the lock introduced here.

Essentially it is saying "the previous behavior did not have this bug", but it did, assuming this is actually a bug / real behavior that can occur. I did not dig into the lunr dependency to validate whether building the index can actually throw.

These response comments of mine are nit's though.

apps/cli/src/commands/list.command.ts
```diff
@@ -144,12 +144,7 @@ export class ListCommand {
     }
 
     if (options.search != null && options.search.trim() !== "") {
-      ciphers = this.searchService.searchCiphersBasic(
-        ciphers,
-        options.search,
-        options.trash,
-        options.archived,
-      );
+      ciphers = this.searchService.searchCiphersBasic(ciphers, options.search);
```
shane-melton, line 148: ⚠️ We dropped the trash/archive filtering here when a search term is present. We need to run the `ciphers = ciphers.filter((c) => this.matchesStateOptions(c, options));` filter manually in this branch now. (it applies the trash/archive filter).

shane-melton (dismissed):
One small issue with the CLI list command, otherwise looks good.

apps/cli/src/commands/list.command.ts
quexten: Applied! Thank you!

jprusik (approved):

ike-kottlowski (approved):

harr1424 (approved):

shane-melton (approved):

JaredScar:
@claude Please review

JaredScar:
@claude Please review

JaredScar:
Testing this locally via Admin Console and not seeing performance improvements in Admin Console's collections page? I would think that the changes here affect speeding up the AC collections page as well? 🤔

quexten:
I'll respond to claude's review here; It brings up the same point that @theMickster's local review already brought up. (I'm not sure if it just copied the comments from the PR discussions that it likely read).

> ⚠️ Important — Lunr index is never cleared on lock/logout/account switch

The index in no longer stored in a state provider and thus cleared on logout and lock by process reload, and even by juts closing the renderer window. This is not a valid comment.

> ♻️ Refactor — TOCTOU in getOrCreateIndex causes redundant rebuilds

The timing for this is exceedingly rare, but also as far as I understand the JS runtime, it only yields async execution when the code hits an await, which means that actually it never happens, since there is no await between the isIndexing check and breaking out of the loop / setting isIndexing here.

JaredScar (approved):

PR #19252, opened 2026-02-26
[PM-32783] Implement cached json state store (speed up login / unlock / general app interactions)

quexten:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-32783

## 📔 Objective

Reduces reads and writes by two orders of magnitude during login and unlock. Further, for large vaults, this significantly increases responsiveness of all UI-based app interactions in desktop.

Electron-store fully reads / writes the data.json every time a state is read or wrote. This completely blocks the main process, blocking it even for in-memory reads and writes, locking up the UI.

Essentially, this replaces electron-store with a custom json store that caches the state in-memory.


quexten:
replaced by https://github.com/bitwarden/clients/pull/19590

PR #19258, opened 2026-02-26
[PM-32160] Web uses shared Send list component, move functionality in…

mcamirault:
…to base Send component and update web and desktop

## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-32160

## 📔 Objective

This PR began with just updating the web Send component to use the shared `SendListComponent` recently introduced and used by desktop. As part of this I had to add the SendType filter toggle group to the `SendSearchComponent` and add logic for conditionally hiding it (as desktop uses the nav menu to apply the filters instead). In the spirit of reducing code duplication as mentioned in the ticket I then decided to move as much functionality as possible into the base Send component that the web component extends and update the desktop Send component to extend the base class as well. I also modernized the base Send component to use the same services that the `SendListComponent` does and cleaned up all components throughout.

This whole process should result in minimal UI changes. When using search filters (either text of SendType) that result in no Sends being shown the web will now match the desktop and show a "No search results returned" message.

### DIFF
--- a/apps/desktop/src/app/tools/send-v2/send-v2.component.ts
+++ b/apps/desktop/src/app/tools/send-v2/send-v2.component.ts
@@ -1,31 +1,21 @@
 // FIXME: Update this file to be type safe and remove this and next line
 // @ts-strict-ignore
 import { Component, computed, DestroyRef, inject, signal, viewChild } from "@angular/core";
-import { toSignal } from "@angular/core/rxjs-interop";
-import { combineLatest, map, switchMap, lastValueFrom } from "rxjs";
+import { takeUntilDestroyed, toSignal } from "@angular/core/rxjs-interop";
+import { lastValueFrom } from "rxjs";
 
 import { JslibModule } from "@bitwarden/angular/jslib.module";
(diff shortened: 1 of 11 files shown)


apps/desktop/src/app/tools/send-v2/send-v2.component.ts
```diff
@@ -67,24 +57,14 @@ type Action = (typeof Action)[keyof typeof Action];
   ],
   templateUrl: "./send-v2.component.html",
 })
-export class SendV2Component {
+export class SendV2Component extends BaseSendComponent {
```
Hinton, line 60: Extending other components is not a recommended pattern. In 95% of the cases you should prefer composition over inheritance, https://contributing.bitwarden.com/contributing/code-style/web/angular#composition-over-inheritance-adr-0009.

Hinton:
Rather than adding filters to send what about providing slots and letting the application insert the filters?

mcamirault:
Rather than try to undo all the changes I didn't need on this branch it was easier to start from scratch on a new one. The new PR is located [here](https://github.com/bitwarden/clients/pull/19454). Closing this PR so discussion can continue on the new one.

PR #19259, opened 2026-02-26
[PM-32783] Add caching layer to electron to speed up unlock, and all app interaction

quexten:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-32783

## 📔 Objective

Makes login, unlock and general UI interaction **_several orders of magnitude faster_** on desktop. Electron store is blocking the main thread with json seralization and deserialization, which makes both persistent state but also in-memory state access slow. This compounds if there is a lot of reads or writes, which blocks UI.

The caching layer here reads the state once, and then never again. Writing happens on a timer, at a maximum of once per second, with an additional flush when the application quits.

### DIFF
--- a/apps/desktop/src/main.ts
+++ b/apps/desktop/src/main.ts
@@ -307,6 +307,7 @@ export class Main {
     );
 
     app.on("will-quit", () => {
+      this.storageService.dispose();
       this.mainDesktopAutotypeService.dispose();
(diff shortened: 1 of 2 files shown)


quexten:
replaced by https://github.com/bitwarden/clients/pull/19590

PR #19263, opened 2026-02-26
Share inline menu qualification service implementation of keyword matching with autofill service

blackwood:
## 🎟️ Tracking

[PM-30478](https://bitwarden.atlassian.net/browse/PM-30478)

## 📔 Objective

Creates `utils/qualification.ts` to share implementation of field attribute keyword matching functions between inline menu and autofill services. (favored inline menu, which was more exhaustive, including labels). Also includes `getSubmitButtonKeywordsSet`. Expanding qualification in autofill service to labels allows a field with an ambiguous ID (like "oid" on username input, but only "user ID" in label, on BoA) to be qualified identically in both services and properly autofilled. 

Removes "fuzzy" misnomer everywhere in autofill concerns (substring or exact tokenized keyword matching are the only current matching methods). (i.e. `fieldContainsKeyword`)

Begins consolidation of keywords/keyphrases in autofill-constants.

[PM-30478]: https://bitwarden.atlassian.net/browse/PM-30478?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/browser/src/autofill/services/autofill-constants.ts
+++ b/apps/browser/src/autofill/services/autofill-constants.ts
@@ -68,6 +68,37 @@ export class AutoFillConstants {
     "create",
   ];
 
+  static readonly AccountCreationFieldKeywords: string[] = [
+    // Field-level keywords indicating account creation / registration context.
+    // Broader than RegistrationKeywords (which is used at the form level).
+    "register",
+    "registration",
+    "create password",
+    "create a password",
+    "create an account",
+    "create account password",
+    "create user password",
+    "confirm password",
+    "confirm account password",
+    "confirm user password",
+    "new user",
+    "new email",
+    "new e-mail",
+    "new password",
+    "new-password",
+    "neuer benutzer",
+    "neues passwort",
+    "neue e-mail",
+    "pwdcheck",
+  ];
+
+  static readonly UpdatePasswordFieldKeywords: string[] = [
+    "update password",
+    "change password",
+    "current password",
+    "kennwort ändern",
+  ];
+
   static readonly NewsletterFormNames: string[] = ["newsletter"];
 
   static readonly FieldIgnoreList: string[] = ["captcha", "findanything", "forgot"];
--- a/apps/browser/src/autofill/services/autofill-constants.ts
+++ b/apps/browser/src/autofill/services/autofill-constants.ts
@@ -105,7 +105,17 @@ export class AutoFillConstants {
     "kennwort ändern",
   ];
 
-  static readonly NewsletterFormNames: string[] = ["newsletter"];
+  /**
+   * Form-level keywords indicating a non-login context such as newsletter signup or
+   * subscription forms. Used to exclude fields within these forms from login autofill.
+   */
+  static readonly NonLoginFormKeywords: string[] = [
+    "newsletter",
+    "subscribe",
+    "subscription",
+    "unsubscribe",
+    "mailing",
+  ];
 
   static readonly FieldIgnoreList: string[] = ["captcha", "findanything", "forgot"];
 
@@ -434,6 +444,7 @@ export class IdentityAutoFillConstants {
     "label-top",
     "data-recurly",
     "accountCreationFieldType",
+    "type",
   ];
 
   static readonly FullNameFieldNames: string[] = ["name", "full-name", "your-name"];
--- a/apps/browser/src/autofill/services/autofill-constants.ts
+++ b/apps/browser/src/autofill/services/autofill-constants.ts
@@ -68,9 +68,11 @@ export class AutoFillConstants {
     "create",
   ];
 
+  /**
+   * Field-level keywords indicating account creation or registration context.
+   * Broader than {@link RegistrationKeywords}, which is used at the form level.
+   */
   static readonly AccountCreationFieldKeywords: string[] = [
-    // Field-level keywords indicating account creation / registration context.
-    // Broader than RegistrationKeywords (which is used at the form level).
     "register",
     "registration",
     "create password",
@@ -92,6 +94,10 @@ export class AutoFillConstants {
     "pwdcheck",
   ];
 
+  /**
+   * Field-level keywords indicating a password update or change context, as distinguished
+   * from a new account creation or initial login context.
+   */
   static readonly UpdatePasswordFieldKeywords: string[] = [
     "update password",
     "change password",
--- a/apps/browser/src/autofill/services/autofill-constants.ts
+++ b/apps/browser/src/autofill/services/autofill-constants.ts
@@ -111,10 +111,12 @@ export class AutoFillConstants {
    */
   static readonly NonLoginFormKeywords: string[] = [
     "newsletter",
-    "subscribe",
-    "subscription",
-    "unsubscribe",
-    "mailing",
+    // @TODO expand list thoughtfully
+    // consider possible collisions with login forms
+    // consider using a "maybe" check
+    // "subscribe",
+    // "subscription",
+    // "unsubscribe",
   ];
 
   static readonly FieldIgnoreList: string[] = ["captcha", "findanything", "forgot"];
--- a/apps/browser/src/autofill/services/autofill.service.ts
+++ b/apps/browser/src/autofill/services/autofill.service.ts
@@ -55,7 +55,7 @@ import { AutofillPort } from "../enums/autofill-port.enum";
 import AutofillField from "../models/autofill-field";
 import AutofillPageDetails from "../models/autofill-page-details";
 import AutofillScript from "../models/autofill-script";
-import { fieldContainsKeyword } from "../utils/qualification";
+import { fieldContainsKeyword, isNonLoginFormContext } from "../utils/qualification";
 
 import {
   AutoFillOptions,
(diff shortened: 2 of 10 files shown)


apps/browser/src/autofill/utils/qualification.ts
```diff
@@ -0,0 +1,122 @@
...
+ * @param keywords - Keywords to search for
+ * @param exactMatch - If true, keyword must be an exact token in the field's normalized
+ *   attribute data. If false (default), keyword is searched as a substring across all tokens.
+ */
+export function fieldContainsKeyword(
+  field: AutofillField,
+  keywords: string[],
+  exactMatch = false,
+): boolean {
+  const parsedKeywords = keywords.map((k) => k.replace(/-/g, ""));
```
bensbits91, line 88: ~~Should we add `.toLowerCase()` before matching against `keywordsSet` to ensure upper case letter don't cause failures?~~

bensbits91: Hm never mind: Keywords passed to `fieldContainsKeyword` are static constants from `AutoFillConstants` arrays, which are all lowercase. Disregard.

bensbits91 (dismissed):
Looks good Miles!

apps/browser/src/autofill/utils/qualification.ts
```diff
@@ -0,0 +1,122 @@
...
+    field.placeholder,
+    field.autoCompleteType,
+    field.dataSetValues,
+    field["label-data"],
+    field["label-aria"],
+    field["label-left"],
+    field["label-right"],
+    field["label-tag"],
+    field["label-top"],
+  ];
```
jprusik, line 61: I wonder if this is a good point to implement some of @claude  's previous recommendations?

`aria-describedby`
https://github.com/bitwarden/clients/pull/17832#discussion_r2615102866

`aria-labelledby`
https://github.com/bitwarden/clients/pull/17832#discussion_r2615103112

jprusik: ^ non-blocking @blackwood ; address or defer as you wish

blackwood: Let's defer to keep this change more focused.

apps/browser/src/autofill/services/autofill-constants.ts
```diff
@@ -68,6 +68,37 @@ export class AutoFillConstants {
     "create",
   ];
 
+  static readonly AccountCreationFieldKeywords: string[] = [
+    // Field-level keywords indicating account creation / registration context.
+    // Broader than RegistrationKeywords (which is used at the form level).
```
jprusik: jsdoc convention for better tooling?

```suggestion
  /**
   * Field-level keywords indicating account creation / registration context.
   * Broader than RegistrationKeywords (which is used at the form level). 
   */
  static readonly AccountCreationFieldKeywords: string[] = [
```

Image

apps/browser/src/autofill/utils/qualification.ts
```diff
@@ -0,0 +1,132 @@
...
+ */
+function buildAutofillFieldKeywords(field: AutofillField) {
+  if (autofillFieldKeywordsCache.has(field)) {
+    return autofillFieldKeywordsCache.get(field)!;
+  }
+  const attributeValues = [
+    field.htmlID,
+    field.htmlName,
+    field.htmlClass,
+    field.type,
```
bensbits91, line 50: Might including `field.type` in `buildAutofillFieldKeywords` cause false matches on pages that have no password or username field, but do have an `<input type="email">` for something unrelated to login (e.g. a newsletter signup)? Since "email" is in `UsernameFieldNames` (via the spread `...EmailFieldNames`), any `type="email"` field would now unconditionally qualify as a username field (unless I'm missing something)— was this intentional?

bensbits91 (commented):
Looks great. I have added one question.

apps/browser/src/autofill/services/inline-menu-field-qualification.service.ts
```diff
@@ -1203,92 +1122,6 @@ export class InlineMenuFieldQualificationService implements InlineMenuFieldQuali
...
-  private getAutofillFieldDataKeywords(
-    autofillFieldData: AutofillField,
-    returnStringValue: boolean,
-  ) {
-    if (!this.autofillFieldKeywordsMap.has(autofillFieldData)) {
-      const keywords = [
-        autofillFieldData.htmlID,
-        autofillFieldData.htmlName,
-        autofillFieldData.htmlClass,
-        autofillFieldData.type,
```
blackwood, line 1246: Reference: previous inclusion of `type` in inline menu field qualification.

jprusik (dismissed):

apps/browser/src/autofill/utils/qualification.ts
bensbits91: Referencing a [slack convo](https://bitwarden.slack.com/archives/C07SW41CA6Q/p1772651524822049) for posterity:

> I think it makes sense to tighten the form-level qualification (isUsernameFieldForLoginForm, isFieldForIdentityForm) so the upstream form-level check prevents us from targeting unrelated forms. 


This seems better suited for followup work and shouldn't block this PR.

bensbits91 (dismissed):

apps/browser/src/autofill/services/autofill.service.ts
```diff
@@ -2420,53 +2421,63 @@ export default class AutofillService implements AutofillServiceInterface {
...
         }
+      } else {
+        bestCandidate = field;
       }
     }
 
-    // Prefer username field in same form, fall back to any username field
-    return usernameFieldInSameForm || usernameField;
+    // Prefer a same-form candidate, fall back to any matching field
+    return bestCandidate || sameFormCandidate;
```
bensbits91: Should this be `sameFormCandidate || bestCandidate` so we prefer / prioritize fields in the same form over keyword-matched fields? Or was reversing this to `bestCandidate || sameFormCandidate` intentional (or inconsequential)?

The old code returned `usernameFieldInSameForm || usernameField`, preferring same-form fields. 

If I understand correctly, `bestCandidate || sameFormCandidate` might cause us to prefer a keyword-matched field from a different form over a same-form field with no keyword match.

For example: On a page with both a login form (with a generically-named username field + a password field) and a newsletter signup form (with a clearly-named email field), `bestCandidate || sameFormCandidate` might cause the password field to be paired with the newsletter email input instead of the co-located login field.

blackwood: Yes, this was a typo, the comment shows that I meant to flip those, thanks for catching!

bensbits91 (approved):

apps/browser/src/autofill/services/autofill-constants.ts
```diff
@@ -68,7 +68,56 @@ export class AutoFillConstants {
...
+    "current password",
+    "kennwort ändern",
+  ];
+
+  /**
+   * Form-level keywords indicating a non-login context such as newsletter signup or
+   * subscription forms. Used to exclude fields within these forms from login autofill.
+   */
+  static readonly NonLoginFormKeywords: string[] = [
+    "newsletter",
```
jprusik, line 113: "mailing list"?

apps/browser/src/autofill/utils/qualification.ts
```diff
@@ -0,0 +1,173 @@
...
+  }
+
+  return keywordsSet;
+}
+
+/**
+ * Returns true if the field's parent form contains keywords indicating a non-login
+ * context (e.g. newsletter signup, subscription forms). Checks the form's {@link AutofillForm.htmlID},
+ * {@link AutofillForm.htmlName}, and {@link AutofillForm.htmlAction} attributes against
+ * {@link AutoFillConstants.NonLoginFormKeywords}. Returns false when the field has no parent form.
```
jprusik, line 140: > Returns false when the field has no parent form.

~Wouldn't a standalone newsletter signup be less likely to have an enclosing form? What does the absence of an enclosing form indicate?~ 

**edit:** Nevermind, I see this is specifically about the form, and not the input, so defensive checking for form presence makes sense.

apps/browser/src/autofill/utils/qualification.ts
```diff
@@ -0,0 +1,173 @@
...
+ *
+ * @param field - The AutofillField whose parent form is to be checked
+ * @param pageDetails - Page details containing the forms map
+ */
+export function isNonLoginFormContext(
```
jprusik, line 145: In isolation, this function is clear: "Does the field belong to a non-login form?". However, everywhere (?) we invoke it, we're using as a proxy to answer the question "is this a non-login username field"? However, it doesn't appear to actually check the field, and the surrounding logic of such callsites don't appear to do that check either (for example, see `isUsernameFieldForLoginForm` in `InlineMenuFieldQualificationService`).

So if the form has generic metadata but the input has relevant disqualifiers, they will be missed (this is probably the job of a separate function).

jprusik: A secondary concern:

I'm examining [this real case](https://www.tlc.com/giveaways):
Image

The form itself has no indicators of being a non-login form, but it has siblings and parents with relevant login case disqualifiers. We likely want to search adjacent items for relevant metadata, like we do for fields.

jprusik: (☝️ this is not a blocking concern for this PR, but we should come back to it)

PR #19268, opened 2026-02-26
[PM-32466] Icons on the send page are not announced by the screen reader

harr1424:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-32466

## 📔 Objective

This PR resolves a defect where icons on the Send list view are not announced by a screen reader.

PR #18940 introduced a regression involving icon announcements by replacing `aria-hidden` + `<span class="tw-sr-only">` with `aria-label` + `tabindex="0"` on `<i>` elements. Screen readers ignore `aria-label` on `<i>` elements. The `tabindex` also introduced unnecessary tab stops on non-interactive elements. This PR restores the original pattern.   

## 📸 Screenshots

Screenshot 2026-02-26 at 15 43 37

<br><br>

Screenshot 2026-02-26 at 15 40 28

<br><br>

Screenshot 2026-02-26 at 15 32 52

### DIFF
--- a/libs/tools/send/send-ui/src/send-table/send-table.component.html
+++ b/libs/tools/send/send-ui/src/send-table/send-table.component.html
@@ -23,6 +23,23 @@
             </span>
             <button type="button" bitLink>
               {{ s.name }}
+              @if (s.disabled) {
+                <span class="tw-sr-only">, {{ "disabled" | i18n }}</span>
+              }
+              @if (s.authType !== authType.None) {
+                @let titleKey =
+                  s.authType === authType.Email ? "emailProtected" : "passwordProtected";
+                <span class="tw-sr-only">, {{ titleKey | i18n }}</span>
+              }
+              @if (s.maxAccessCountReached) {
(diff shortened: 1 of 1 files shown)


libs/tools/send/send-ui/src/send-table/send-table.component.html
```diff
@@ -29,9 +29,9 @@
                 class="bwi bwi-exclamation-triangle"
                 appStopProp
                 title="{{ 'disabled' | i18n }}"
-                aria-label="{{ 'disabled' | i18n }}"
-                tabindex="0"
+                aria-hidden="true"
               ></i>
```
mcamirault, line 50: I'd recommend changing all these `<i>` and `<span class="tw-sr-only">` combos to using `<bit-icon ariaLabel=...>` instead

mcamirault: After internal discussion we decided not to do this yet

PR #19272, opened 2026-02-27
[PM-29841] Account Security page: fix flickering form values

lastbestdev:
## 🎟️ Tracking
https://bitwarden.atlassian.net/browse/PM-29841

## 📔 Objective
Originally to fix the phishing detection form control in the Account Security component from switching after load, but all fields in this component have the same issue. Adds a loading state to prevent the values from flickering during load.

## 📸 Screenshots
**Before**

https://github.com/user-attachments/assets/f50601a0-a293-4b83-a02f-3b629bfa0cdf

**After**

https://github.com/user-attachments/assets/99fc02c8-aa8b-49b6-a836-c94ed61bbc75

### DIFF
--- a/apps/browser/src/auth/popup/settings/account-security.component.html
+++ b/apps/browser/src/auth/popup/settings/account-security.component.html
@@ -12,7 +12,9 @@
     (onDismiss)="dismissAccountSecurityNudge()"
     class="tw-mb-6"
   ></bit-spotlight>
-  @if (!loading()) {
+  @if (loading()) {
+    <div class="tw-h-full tw-place-content-center"><bit-spinner></bit-spinner></div>
+  } @else {
     <div [formGroup]="form">
       <bit-section>
         <bit-section-header>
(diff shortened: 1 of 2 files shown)


apps/browser/src/auth/popup/settings/account-security.component.html
```diff
@@ -12,157 +12,163 @@
...
-          <bit-label for="pinEphemeral" class="tw-whitespace-normal">
-            {{ "lockWithMasterPassOnRestart1" | i18n }}
-          </bit-label>
-        </bit-form-control>
-      </bit-card>
-    </bit-section>
-
-    <bit-section>
-      @if (consolidatedSessionTimeoutComponent$ | async) {
+  @if (!loading()) {
```
lastbestdev: The diff here turned out a lot uglier than I'd like, due to formatting. All this PR changes to this template is wrapping the `formGroup` div with the loading check

lastbestdev:
> @lastbestdev , thank you for the improvement! However, we probably should add a usage of the `bit-spinner` ([docs](https://components.bitwarden.com/?path=/docs/component-library-spinner--docs)) and add a loading state to complete this (i.e., add an `@if(loading) { // show spinner }`). If you put the `this.loading.set(false);` in a `setTimeout` of some duration, I think the user will be left with a blank screen without a loading spinner.

Good call @JaredSnider-Bitwarden, updated with spinner in latest

PR #19278, opened 2026-02-27
[PM-33101] Remove master key from uv service

quexten:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-33101

## 📔 Objective

UV service does not need the master-key since it uses proof-of-decryption. We remove this dependency on the master-key so we can unblock dropping the master-key.

### DIFF
--- a/apps/cli/src/service-container/service-container.ts
+++ b/apps/cli/src/service-container/service-container.ts
@@ -691,7 +691,6 @@ export class ServiceContainer {
       this.kdfConfigService,
       this.accountService,
       this.masterPasswordService,
-      this.cryptoFunctionService,
       this.stateProvider,
       this.logService,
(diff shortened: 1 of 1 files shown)


libs/common/src/auth/services/user-verification/user-verification.service.ts
```diff
@@ -230,8 +221,7 @@ export class UserVerificationService implements UserVerificationServiceAbstracti
       }
     }
 
-    await this.masterPasswordService.setMasterKey(masterKey, userId);
```
Patrick-Pimentel-Bitwarden, line 233: I'm trying to think through if this has any consequences by not setting this into state anymore. Have you thought this change through? Why were we setting this to begin with?

PR #19281, opened 2026-02-27
Move Sync HTTP request out of the critical unlock path

quexten:
## 📔 Objective

Currently sync adds a few hundred milliseconds to a few seconds to every unlock, depending on latency to the server. There are one or two requests being made. If a sync is not required, a revision-date request is made. If a sync is required, additionally the sync request is made. The sum of these two requests is blocking unlock, making unlock slower by a second or more, which is undesired.

There are two options to solve this; Either sync in the background entirely, or start syncing during a long operation (i.e before master password unlock / pin unlock) and wait for the sync to finish before continuing. The second approach keeps the external behavior the same but is more complex. We go with the second behavior here since we did not analyze the effects of not finishing a sync before continuing with unlock.

This leads to a large increase in most unlocks.


libs/key-management-ui/src/lock/components/master-password-lock/master-password-lock.component.ts
```diff
@@ -121,6 +123,8 @@ export class MasterPasswordLockComponent implements OnInit, OnDestroy {
 
     const activeUserId = await firstValueFrom(getUserId(this.accountService.activeAccount$));
 
+    (async () => { this.syncService.fullSync(false) })();
```
Thomas-Avery, line 126: I think we can write the fire and forgets like this to make the linter happy.
```suggestion
    void this.syncService.fullSync(false);
```

PR #19297, opened 2026-02-27
[PM-32890] Fix self-hosted premium subscription page with FF on

amorask-bitwarden:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-32890

## 📔 Objective

When the `pm-29594-update-individual-subscription-page` feature flag is enabled on a self-hosted instance, the premium subscription page renders blank. This happens because the `AccountSubscriptionComponent` and supporting components call `AccountBillingClient.getSubscription()`, which hits a cloud-only API endpoint (`/api/account/billing/vnext/subscription`) that doesn't exist on self-hosted servers.

This PR fixes the regression by:

- **Guarding `getSubscription()` calls** in `SubscriptionComponent` and `UserLayoutComponent` so they are only made on cloud-hosted instances. On self-hosted, `hasPremiumPersonally$` is used instead.
- **Creating a dedicated `SelfHostedAccountSubscriptionComponent`** that renders a self-hosted-appropriate premium subscription page (license expiration, update license dialog, link to cloud subscription management) without calling cloud-only APIs.
- **Renaming `AccountSubscriptionComponent` to `CloudHostedAccountSubscriptionComponent`** to clarify its scope.
- **Updating the individual billing routing module** to use a three-route `canMatch` strategy for `/user-subscription`: self-hosted (flag on) → `SelfHostedAccountSubscriptionComponent`, cloud (flag on) → `CloudHostedAccountSubscriptionComponent`, default (flag off) → `UserSubscriptionComponent`.

**Follow-up commits addressing PR review feedback:**

- **Moved premium redirect logic from component to route guard** — Extracted the self-hosted premium check from `SelfHostedAccountSubscriptionComponent` into a `hasPremiumPersonallyGuard` `canActivate` guard, keeping navigation concerns in the routing layer.
- **Extracted subscription route logic into `PremiumSubscriptionRoutingService`** — Moved the sidebar nav link routing logic out of `UserLayoutComponent` into a dedicated service, reducing the component's complexity and responsibilities.
- **Addressed local and remote PR review feedback** — Simplified the guard to a plain `CanActivateFn` const, added `LogService` error logging to `catchError` blocks in the routing service and `SubscriptionComponent`, fixed self-hosted nav link visibility for users with both org-premium and personal premium, improved routing comments for accuracy, reused `isSelfHosted` helper for the existing `/premium` route, and changed `isActive` to default to `false` when expiration data is missing.

### DIFF
--- a/apps/web/src/app/billing/individual/subscription/self-hosted-account-subscription.component.ts
+++ b/apps/web/src/app/billing/individual/subscription/self-hosted-account-subscription.component.ts
@@ -47,14 +47,11 @@ export class SelfHostedAccountSubscriptionComponent {
 
   readonly isActive = computed<boolean>(() => {
     const expiration = this.expiration();
-    if (!expiration || expiration.trim() === "") {
-      return true;
+    if (!expiration) {
+      return false;
     }
     const expirationDate = new Date(expiration);
-    if (isNaN(expirationDate.getTime())) {
-      return true;
-    }
-    return expirationDate > new Date();
+    return !isNaN(expirationDate.getTime()) && expirationDate > new Date();
   });
 
   readonly cloudSubscriptionUrl = toSignal(
--- a/apps/web/src/app/billing/guards/has-premium-personally.guard.ts
+++ b/apps/web/src/app/billing/guards/has-premium-personally.guard.ts
@@ -10,17 +10,15 @@ import { BillingAccountProfileStateService } from "@bitwarden/common/billing/abs
  * CanActivate guard that checks if the user has a personal premium subscription.
  * Redirects to the premium upgrade page if not.
  */
-export function hasPremiumPersonallyGuard(): CanActivateFn {
-  return (): Observable<boolean | UrlTree> => {
-    const router = inject(Router);
-    const accountService = inject(AccountService);
-    const billingAccountProfileStateService = inject(BillingAccountProfileStateService);
+export const hasPremiumPersonallyGuard: CanActivateFn = (): Observable<boolean | UrlTree> => {
+  const router = inject(Router);
+  const accountService = inject(AccountService);
+  const billingAccountProfileStateService = inject(BillingAccountProfileStateService);
 
-    return accountService.activeAccount$.pipe(
-      switchMap((account) =>
-        account ? billingAccountProfileStateService.hasPremiumPersonally$(account.id) : of(false),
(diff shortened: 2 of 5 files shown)


amorask-bitwarden:
> There is a fair bit of logic churn in this component. Is there some service most of this could be moved to, and then just map it to the route strings here? What does the binary represent in terms of business logic?

@willmartian Sorry, what is this question in reference to? It's not pointed at any specific changes in the PR.

apps/web/src/app/billing/individual/subscription/self-hosted-account-subscription.component.ts
```diff
@@ -0,0 +1,94 @@
...
+  );
+
+  private readonly subscription = resource({
+    params: () => ({
+      account: this.account(),
+      hasPremiumPersonally: this.hasPremiumPersonally(),
+    }),
+    loader: async ({ params: { account, hasPremiumPersonally } }) => {
+      if (!account || !hasPremiumPersonally) {
+        await this.router.navigate(["/settings/subscription/premium"]);
```
willmartian: This sounds to me like a job for a router guard https://angular.dev/guide/routing/route-guards

Rerouting on component load is a bit of an anti-pattern, due to how late it occurs in the navigation lifecycle.

willmartian:
> > There is a fair bit of logic churn in this component. Is there some service most of this could be moved to, and then just map it to the route strings here? What does the binary represent in terms of business logic?
> 
> @willmartian Sorry, what is this question in reference to? It's not pointed at any specific changes in the PR.

The logic in the user-layout component, sorry!

apps/web/src/app/billing/individual/subscription/self-hosted-account-subscription.component.ts
amorask-bitwarden: Good call - moved this to a guard.

amorask-bitwarden:
> > > There is a fair bit of logic churn in this component. Is there some service most of this could be moved to, and then just map it to the route strings here? What does the binary represent in terms of business logic?
> > 
> > 
> > @willmartian Sorry, what is this question in reference to? It's not pointed at any specific changes in the PR.
> 
> The logic in the user-layout component, sorry!

All good! Moved to the new `premium-subscription-routing.service`.

PR #19301, opened 2026-02-27
[PM-32687] Create Claude skill to add more item types easily

gbubemismith:
## 🎟️ Tracking
https://bitwarden.atlassian.net/browse/PM-32687


.claude/skills/cipher-type-planner/SKILL.md
```diff
@@ -0,0 +1,325 @@
...
+user-invokable: true
+argument-hint: "<cipher-type-name> - e.g., BankAccount, Passport"
+---
+
+# Cipher Type Planner
+
+Plans the full implementation of a new or modified cipher type across all layers of the Bitwarden
+clients monorepo, the server repository, and the SDK repository.
+
+**Reference documentation:** `docs/cipher-types.md` in the repo root.
```
theMickster: ⚠️ I don't see this file in the repo nor the pull request.

.claude/skills/cipher-type-planner/SKILL.md
```diff
@@ -0,0 +1,325 @@
...
+
+## Workflow
+
+### Step 1: Gather Requirements
+
+Read the reference documentation at `docs/cipher-types.md` to refresh context on the five-layer
```
theMickster: ⚠️ I don't see this file in the repo nor the pull request.

.claude/skills/cipher-type-planner/SKILL.md
```diff
@@ -0,0 +1,325 @@
+---
+name: cipher-type-planner
+description: Plans the creation or modification of a cipher type (vault item type) across the Bitwarden clients monorepo. Use this skill when a user wants to add a new cipher type (e.g., BankAccount, Passport), modify an existing cipher type, or asks about what is needed to implement a cipher type. DO NOT invoke for general vault or cipher questions unrelated to adding or changing a cipher type.
```
theMickster: 🎨 Let's remove the "(e.g., BankAccount, Passport)" from the skill description. We want Claude focused intently on what the skill is for and not for in these sentences.

.claude/skills/cipher-type-planner/SKILL.md
```diff
@@ -0,0 +1,325 @@
...
+user-invokable: true
+argument-hint: "<cipher-type-name> - e.g., BankAccount, Passport"
+---
+
+# Cipher Type Planner
+
+Plans the full implementation of a new or modified cipher type across all layers of the Bitwarden
```
theMickster: ⚠️ The introductory sentence here tells me the best home for our new skill is our [ai-plugins marketplace](https://github.com/bitwarden/ai-plugins) because we speak of repos that are more than just clients.

.claude/skills/cipher-type-planner/SKILL.md
```diff
@@ -0,0 +1,325 @@
...
+
+Read the reference documentation at `docs/cipher-types.md` to refresh context on the five-layer
+model stack, container classes, UI components, and cross-repo dependencies.
+
+Then ask the user the following questions (use `AskUserQuestion`). Adapt questions based on what
+the user has already provided.
+
+**Required questions:**
+
+1. **Type name and value** - What is the cipher type name (e.g., `BankAccount`) and integer value
```
theMickster: ⚠️ Let's work on refactoring this questions by trimming down to the bare minimum, focused context that Claude must have to guide the user.  The engineer should not be using the skill to create a new cipher type if they do not have a very clear understanding of the name and numerical value. We should not be encouraging Claude to suggest anything in the workflow of such a critically important entity in our software.

```suggestion
1. **Type name and value** - What is the cipher type name and integer value?
```

.claude/skills/cipher-type-planner/SKILL.md
```diff
@@ -0,0 +1,325 @@
...
+   (next available after 5 for SshKey)? If the user hasn't specified a value, suggest the next
+   available integer.
+2. **Fields** - What fields/properties does this cipher type have? List each field with:
```
theMickster: ⚠️ Let's refactor

```suggestion
2. **Fields** - What are the cipher's properties? Each property must contain:
```

.claude/skills/cipher-type-planner/SKILL.md
```diff
@@ -0,0 +1,325 @@
...
+   - Field name (e.g., `routingNumber`)
```
theMickster: ⚠️ Let's simplify

```suggestion
   - Field name
```

.claude/skills/cipher-type-planner/SKILL.md
```diff
@@ -0,0 +1,325 @@
...
+   - Data type (string, number, boolean)
+   - Whether it is encrypted (most vault data fields should be)
```
theMickster: ⚠️ Let's simplify

```suggestion
   - Encryption required?
```

.claude/skills/cipher-type-planner/SKILL.md
```diff
@@ -0,0 +1,325 @@
...
+   - Whether it is required
```
theMickster: ⚠️ Let's simplify

```suggestion
   - Required? 
```

.claude/skills/cipher-type-planner/SKILL.md
```diff
@@ -0,0 +1,325 @@
...
+3. **Autofill** - Should this type participate in browser autofill? (Currently only Login, Card,
+   and Identity support autofill.)
+4. **Linked fields** - Should this type support linked custom fields? If yes, which properties
+   should be linkable?
+5. **Import/Export** - Should this type support import and export? Yes/No.
+6. **Feature flag** - What is the feature flag name? (e.g., `vault-bank-account`.)
+7. **Cross-repo status** - Has the server and/or SDK work already been completed, or does it need
+   to be planned as well?
+
+**Optional questions (ask if relevant):**
```
theMickster: ⚠️ Please either chose to have these here or not. Claude will guess here and that's not what we want to have occur. We also don't want engineers to guess either. We should rephrase these to be direct questions. If not applicable, then we need to direct Claude to allow for the engineer to state that as such.

"Ask if relevant" is ambiguous — Claude has to guess what's relevant, and the engineer doesn't know these questions exist if Claude skips them.

```suggestion
  **Additional questions:**

  Ask each of the following. If the engineer does not have an answer, accept "N/A" or "not yet decided" and note it as a gap in the plan.

  - **UI details** - Are there specific UI requirements for the form or view sections (e.g., dropdowns, masked fields, copy buttons)?
  - **Subtitle** - What value should the `subTitle` getter on the view model return? This appears in vault list items.
  - **Icon** - What icon represents this type in the vault? Bitwarden uses `bwi-` icon classes.
  - **Organization policy** - Does this type appear in the restricted item types policy UI?
```

.claude/skills/cipher-type-planner/SKILL.md
```diff
@@ -0,0 +1,325 @@
...
+- **Cipher type name:** (e.g., `BankAccount`)
+- **Integer value:** (e.g., `6`)
+- **Feature flag:** (e.g., `pm-32009-new-item-types`)
+- **Minimum client version:** (e.g., `2026.3.0`)
+- **Fields:** Table of all fields with name, type, encrypted (yes/no), required (yes/no)
+- **Supports autofill:** Yes/No
+- **Supports linked fields:** Yes/No
+- **Supports import/export:** Yes/No
+
+### 2. Cross-Repository Prerequisites
```
theMickster: ❓ Section 2. reads to me like these things have already been accomplished. Is that true? Do we have other Claude Skills in the works for these Prerequisites? IMO these sounds like an opportunity for additional skills for this process in the ai-plugins marketplace.

.claude/skills/cipher-type-planner/SKILL.md
```diff
@@ -0,0 +1,325 @@
...
+- `libs/common/src/vault/services/cipher-sdk.service.spec.ts`
+- `libs/common/src/vault/icon/build-cipher-icon.spec.ts`
+- `libs/common/src/models/export/cipher.export.spec.ts`
+- `libs/vault/src/cipher-form/components/cipher-form.component.spec.ts`
+- `libs/vault/src/cipher-view/cipher-view.component.spec.ts`
+- Form section component spec (new)
+
+### 7. Recommended Implementation Order
+
+Follow the order from `docs/cipher-types.md` Section 14, customized for this specific type:
```
theMickster: ⚠️ I don't see this file in the repo nor the pull request.

.claude/skills/cipher-type-planner/SKILL.md
```diff
@@ -0,0 +1,325 @@
...
+
+1. **Type name and value** - What is the cipher type name (e.g., `BankAccount`) and integer value
+   (next available after 5 for SshKey)? If the user hasn't specified a value, suggest the next
+   available integer.
+2. **Fields** - What fields/properties does this cipher type have? List each field with:
+   - Field name (e.g., `routingNumber`)
+   - Data type (string, number, boolean)
+   - Whether it is encrypted (most vault data fields should be)
+   - Whether it is required
+3. **Autofill** - Should this type participate in browser autofill? (Currently only Login, Card,
```
theMickster: ⚠️ For questions 3–7, the core issue is they're phrased inconsistently — some are yes/no, some are open-ended, some embed context Claude shouldn't be guessing about.


```suggestion
  3. **Autofill** - Does this type participate in browser autofill? If yes, which fields are auto-fillable?
  4. **Linked fields** - Does this type support linked custom fields? If yes, which properties are linkable?
  5. **Import/Export** - Does this type support import and export?
  6. **Feature flag** - What is the feature flag name?
  7. **Cross-repo status** - Has the server and/or SDK work been completed? If not, which pieces still need to be planned?
```

.claude/skills/cipher-type-planner/SKILL.md
```diff
@@ -0,0 +1,325 @@
...
+
+#### Critical Warnings
+
+- **CLI has no copy menu UI** — do not add copy-related i18n keys to the CLI locale.
+- **Only expose fields that should be copiable** — not every cipher field needs a copy action. Check
+  with product requirements for which fields get copy buttons.
+
+---
+
+## Key Rules
```
theMickster: ♻️ These should be the first thing that Claude reads.  Let's move them above the `## Workflow` section.

.claude/skills/cipher-type-planner/SKILL.md
```diff
@@ -0,0 +1,325 @@
...
+- **No code regions** - refactor for readability instead
+- **Tailwind `tw-` prefix** required for all CSS classes
+- **Never log or expose decrypted vault data**
+- Follow the SshKey implementation as the canonical template
+- Use `@linkedFieldOption` decorators if linked fields are supported
+- View models must extend `ItemView` and implement the `subTitle` getter
+
+## References
+
+- `docs/cipher-types.md` - Complete cipher type reference documentation
```
theMickster: ⚠️ I don't see this file in the repo nor the pull request.

.claude/skills/cipher-type-planner/SKILL.md
```diff
@@ -0,0 +1,325 @@
...
+- **No new encryption logic** in the clients repo (CLAUDE.md critical rule)
+- **No TypeScript enums** - use const objects with type aliases (ADR-0025)
+- **No code regions** - refactor for readability instead
+- **Tailwind `tw-` prefix** required for all CSS classes
+- **Never log or expose decrypted vault data**
+- Follow the SshKey implementation as the canonical template
+- Use `@linkedFieldOption` decorators if linked fields are supported
+- View models must extend `ItemView` and implement the `subTitle` getter
+
+## References
```
theMickster: ♻️ I think we need to reconsider these references (I know we have them in many places in our Claude things _hoping_ that Claude will read them). 

From Claude itself after a local pull request review

<blockquote>
No. Claude will never read those.

  - The three ADR links are external URLs. Claude Code doesn't browse to URLs unprompted during skill execution — it would need an explicit WebFetch instruction to do that, and the skill doesn't include one.
  - The SshKey commit hashes are useful to a human reading the skill, but Claude won't run git show b18fa68acc unless told to. The skill already tells Claude to inspect the SshKey files by path in Step 2, which is what actually works.
  - docs/cipher-types.md is already referenced in the workflow steps where Claude will read it (setting aside that it doesn't exist yet).

This section is written for a human audience, not for Claude. That's fine if that's the intent — but if it's meant to influence Claude's behavior, it needs to be actionable instructions, not a bibliography.

</blockquote>

theMickster (changes requested):
Great start to a new Skill! I know I left a lot of items for improvement, but to me, those are relatively minor wording / phrasing to be more direct with Claude. The heart and flow of the Skill feels strong and I am excited to see us put this into action! 

Bravo on your first skill @gbubemismith 👏🏼

.claude/skills/cipher-type-planner/SKILL.md
SaintPatrck: ⛏️ It's redundant to reiterate the purpose of the skill here. The front-matter `description` already does this, so Claude already knows what it's doing when it invokes the skill.

.claude/skills/cipher-type-planner/SKILL.md
```diff
@@ -0,0 +1,325 @@
...
+
+---
+
+## Key Rules
+
+- **No new encryption logic** in the clients repo (CLAUDE.md critical rule)
+- **No TypeScript enums** - use const objects with type aliases (ADR-0025)
+- **No code regions** - refactor for readability instead
+- **Tailwind `tw-` prefix** required for all CSS classes
+- **Never log or expose decrypted vault data**
```
SaintPatrck: These are almost verbatim copies of rules already present in `CLAUDE.md`. Let's remove them since they'll already be in context.

.claude/skills/cipher-type-planner/SKILL.md
```diff
@@ -0,0 +1,325 @@
...
+model stack, container classes, UI components, and cross-repo dependencies.
+
+Then ask the user the following questions (use `AskUserQuestion`). Adapt questions based on what
+the user has already provided.
+
+**Required questions:**
+
+1. **Type name and value** - What is the cipher type name (e.g., `BankAccount`) and integer value
+   (next available after 5 for SshKey)? If the user hasn't specified a value, suggest the next
+   available integer.
```
SaintPatrck: 🤔 I'm not so sure hardcoding `5` here is a good idea. If another CipherType is added will this need to be updated to something like "...after 6 for Passport" so Claude doesn't suggest 6 again?

The intent is still clear without the parenthetical content, so we can probably remove it.

```suggestion
1. **Type name and value** - What is the cipher type name and integer value? If the user hasn't specified a value, determine the next available integer by reading the `CipherType` enum definition.
```

.claude/skills/cipher-type-planner/SKILL.md
gbubemismith: 🤦 Claude kept referencing my local docs folder

.claude/skills/cipher-type-planner/SKILL.md
gbubemismith: Yes, Section 2 lists Prerequisites. There's an important step between the server and the SDK: generating the API bindings. The SDK's [bitwarden-api-api](https://github.com/bitwarden/sdk-internal/tree/main/crates/bitwarden-api-api) crate is auto generated form the server's OpenAPI specs. There are two approaches:
  1. Local generation: You can run scripts locally to generate the bindings from the server's OpenAPI specs. I used this approach to get everything working end-to-end during development, but it's not the recommended workflow.
  2. CI (recommended): Merge the server PR to main, then a job on the SDK repository [(update-api-bindings)](https://github.com/bitwarden/sdk-internal/blob/main/.github/workflows/update-api-bindings.yml) is
  triggered to automatically generate the bindings. This is the approach we'd want to use in practice.

Regarding additional skills, what will be the best approach for this considering there are other dependencies outside the clients?

.claude/skills/cipher-type-planner/SKILL.md
gbubemismith: Yeah I noticed kept using old angular syntax and code styles, so I made it update the skill with the new learnings and it added that reference. I'll  remove it

.claude/skills/cipher-type-planner/SKILL.md
gbubemismith: Fixed c035c19bf787f6e0d1a2f0492fb87f46b742f5e4

.claude/skills/cipher-type-planner/SKILL.md
gbubemismith: Fixed c035c19bf787f6e0d1a2f0492fb87f46b742f5e4

.claude/skills/cipher-type-planner/SKILL.md
gbubemismith: Fixed c035c19bf787f6e0d1a2f0492fb87f46b742f5e4

.claude/skills/cipher-type-planner/SKILL.md
gbubemismith: Fixed b78598925cf79d0462c9d10f8d7aa6f60ff4579d

.claude/skills/cipher-type-planner/SKILL.md
gbubemismith: Fixed b78598925cf79d0462c9d10f8d7aa6f60ff4579d

.claude/skills/cipher-type-planner/SKILL.md
gbubemismith: Fixed b78598925cf79d0462c9d10f8d7aa6f60ff4579d

.claude/skills/cipher-type-planner/SKILL.md
gbubemismith: Fixed b78598925cf79d0462c9d10f8d7aa6f60ff4579d

.claude/skills/cipher-type-planner/SKILL.md
gbubemismith: Fixed b78598925cf79d0462c9d10f8d7aa6f60ff4579d

.claude/skills/cipher-type-planner/SKILL.md
gbubemismith: Fixed b78598925cf79d0462c9d10f8d7aa6f60ff4579d

.claude/skills/cipher-type-planner/SKILL.md
gbubemismith: Fixed b78598925cf79d0462c9d10f8d7aa6f60ff4579d

.claude/skills/cipher-type-planner/SKILL.md
gbubemismith: Fixed b78598925cf79d0462c9d10f8d7aa6f60ff4579d

.claude/skills/cipher-type-planner/SKILL.md
gbubemismith: I think we can just remove the key rules

.claude/skills/cipher-type-planner/SKILL.md
gbubemismith: Fixed https://github.com/bitwarden/clients/commit/c035c19bf787f6e0d1a2f0492fb87f46b742f5e4

.claude/skills/cipher-type-planner/SKILL.md
gbubemismith: Fixed https://github.com/bitwarden/clients/commit/b78598925cf79d0462c9d10f8d7aa6f60ff4579d

.claude/skills/cipher-type-planner/SKILL.md
gbubemismith: Fixed https://github.com/bitwarden/clients/commit/b78598925cf79d0462c9d10f8d7aa6f60ff4579d

.claude/skills/cipher-type-planner/SKILL.md
theMickster: This SKILL.md should assume all prerequisites are already done by a human — it shouldn't include them. As written, I think they confuse Claude into trying to verify each prerequisite exists, when the human should have already completed those steps. 

IMO these should collapse into a single confirmation question: "Have all prerequisites been completed?"

What are your thoughts on that approach for the first iteration of this `SKILL.md` @SaintPatrck??

SaintPatrck: I agree, this sounds like something that needs to live in the marketplace. The primary cue to me is dependencies outside of this repo. 

As a starting point, I suggest reigning in the scope to focus solely on steps required in this repo, and assuming all pre-reqs are already met. Including a brief statement to remind the user, like @theMickster suggests, is a great idea. 

If there are steps in this skill that must wait for steps in other repos to complete, that may mean we need to decompose this into multiple Skills so they can be orchestrated efficiently from a higher level.

A somewhat simple design for composing complex workflows across repos would look something like below.

```
⏺ ai-plugins/                  
  ├── client-development-tools/          
  │   └── skill/               
  │       └── define-new-client-item-type                                                                                                                                                                                                                                                                                                                                                                                         
  │           # Client-specific implementation steps for defining a new item type.
  │           # Assumes all prerequisites are already satisfied.                                                                                                                                                                                                                                                                                                                                                                  
  │                                                                                                                                                                                                                                                                                                                                                                                                                             
  ├── sdk-development-tools/                                                                                                                                                                                                                                                                                                                                                                                                      
  │   └── skill/                                                                                                                                                                                                                                                                                                                                                                                                                
  │       └── define-new-sdk-item-type                                                                                                                                                                                                                                                                                                                                                                                          
  │           # SDK-specific implementation steps for defining a new item type.
  │           # Assumes all prerequisites are already satisfied.
  │
  └── orchestration-tools/
      └── skill/
          └── orchestrate-new-bitwarden-item-type-creation
              # Coordinates end-to-end creation of a new Bitwarden item type
              # across client and SDK. Verifies each step's prerequisites
              # are met before invoking the corresponding skill(s).
```
*_Disclaimer: I am not familiar with the every step required in every system to add a new item type, so this is an over-simplification to show the pattern._

I think what you're aiming for is a great idea, and worth a larger discussion with others so we all aligned on a design to prevent duplicate/conflicting work in other repositories. I could see this playing into some of the things we're doing in the mobile repo's too.

theMickster: @gbubemismith & @SaintPatrck 
To ensure we keep the work in scope and manageable, let's do the following

1. Remove Section 2 regarding prerequisites in favor of a prompt further up where Claude confirms with the user that all server and sdk prerequisites are completed. 
2. Create a follow-up Jira item that would encompass the following
     - Create a new plugin in our plugin marketplace for new vault item creation workflow
     - Create new Skills for the prerequisites listed above (and the whole Open API auto-generation flow)
     - Migrate this current Skill to our plugin marketplace
     - Create a new command to orchestrate the workflow

Does that seem reasonable enough to ensure we have a good Skill here and avoid scope creep?

SaintPatrck: Sounds good to me

gbubemismith: Thanks @theMickster @SaintPatrck

.claude/skills/cipher-type-planner/SKILL.md
```diff
@@ -0,0 +1,271 @@
...
+All vault filter files must be feature-flag-gated so the new type only appears when the flag is
+enabled. Use `ConfigService.getFeatureFlag$()` with `combineLatest` to filter the type out of
+arrays when the flag is off.
+
+- `libs/vault/src/services/vault-filter.service.ts` - **CRITICAL**: Add type to `buildCipherTypeTree()` `allTypeFilters` array. Without this, ciphers of the new type will not appear in the vault sidebar or list.
+- `libs/vault/src/models/filter-function.ts` - Add filter case for the new type
+- `apps/web/src/app/vault/individual-vault/vault-filter/components/vault-filter.component.ts` - Add to `allTypeFilters`, `searchPlaceholder`, and feature-flag-gate in `buildAllFilters()`
+- `apps/web/src/app/admin-console/organizations/collections/vault-filter/vault-filter.component.ts` - Feature-flag-gate in `buildAllFilters()`
+- `apps/desktop/src/vault/app/vault-v3/vault-filter/filters/type-filter.component.ts` - Add `ConfigService`, `combineLatest` with feature flag
+- `apps/desktop/src/vault/app/vault/vault-filter/filters/type-filter.component.ts` - Same pattern
```
theMickster: ⚠️ This file no longer exists and appears to be a stale reference to a legacy directory structure.  I am concerned that it may cause Claude to hallucinate. Line 149 appears to already cover the filter component. 

```suggestion
```

theMickster (changes requested):
Last change and I think we are good.

.claude/skills/cipher-type-planner/SKILL.md
```diff
@@ -0,0 +1,271 @@
+---
+name: cipher-type-planner
+description: Plans the creation or modification of a cipher type (vault item type) across the Bitwarden clients monorepo. Use this skill when a user wants to add a new cipher type, modify an existing cipher type, or asks about what is needed to implement a cipher type. DO NOT invoke for general vault or cipher questions unrelated to adding or changing a cipher type.
+user-invokable: true
+argument-hint: "<cipher-type-name>"
```
SaintPatrck: ⛏️ Not sure if it matters, but the [frontmatter reference](https://github.com/bitwarden/android/pull/6615) examples use `argument-hint: [cipher-type-name]`

🌱 I also notice this argument is not actually being referenced with `$ARGUMENTS` or `$ARGUMENTS[0]` anywhere. Might be worth adding to step one.

Not requesting changes, but something to consider. Especially if you think this skill could eventually take in more arguments.

.claude/skills/cipher-type-planner/SKILL.md
gbubemismith: Nice callout 👍🏼 , I need to sync this branch with main

.claude/skills/cipher-type-planner/SKILL.md
gbubemismith: I think we can just drop the argument since we still ask the question for the cipher name. What do you think?

gbubemismith: Removed 50072e122bd1eda32478f0273c7814a75339851a

theMickster (dismissed):

SaintPatrck (dismissed):

gbubemismith:
@theMickster as discussed, I pushed ba038586fd1a6c5829aad87d6bcafa0a3fe2bdc7 with the optional params

.claude/skills/cipher-type-planner/SKILL.md
```diff
@@ -0,0 +1,311 @@
...
+- `apps/web/src/app/admin-console/organizations/policies/policy-edit-definitions/restricted-item-types.component.ts` - Policy configuration UI
+- `apps/web/src/app/admin-console/organizations/policies/policy-edit-definitions/restricted-item-types.component.html`
+
+_Include if `$0` is `browser` or `all`:_
+
+- `apps/browser/src/vault/popup/components/vault/item-more-options/item-more-options.component.ts` - Restricted type checks
+
+_Include if `$0` is `cli` or `all`:_
+
+- `apps/cli/src/commands/create.command.ts` - Restricted type checks
```
theMickster: ⚠️ Looks like we have a nonexistent file path `apps/cli/src/commands/create.command.ts` The correct path appears to be `apps/cli/src/vault/create.command.ts`.

theMickster (changes requested):
Minor change to a file path and I think we are good

theMickster:
Did we have a merge into main with broken tests or did we have a flaky test run? 
Seems odd that tests would fail when they weren't touched in the PR.

nick-livefront:
> Did we have a merge into main with broken tests or did we have a flaky test run? Seems odd that tests would fail when they weren't touched in the PR.

@theMickster The failure was caused by a type checking test, the change has been reverted from `main` https://github.com/bitwarden/clients/pull/19979

.claude/skills/cipher-type-planner/SKILL.md
nick-livefront: Updated! [6140328](https://github.com/bitwarden/clients/pull/19301/commits/6140328c4a70fc2716d27767880032c98c6d841c)

theMickster (approved):

PR #19302, opened 2026-02-28
[PM-32686] Add bank account item type

gbubemismith:
## 🎟️ Tracking

[PM-32686](https://bitwarden.atlassian.net/browse/PM-32686)
SDK PR: https://github.com/bitwarden/sdk-internal/pull/949
Server PR: https://github.com/bitwarden/server/pull/7112

## 📔 Objective

Adds bank account item type across all four platforms. 
- This work was done as a part of the dev training, hence the one _giant_ PR for each platform. Following new item type work will be separated out by platform.
- Bank account copyable fields are present across all platforms
- CLI has added `get item.bankaccount` and `get template`
- Added `bwi-bank` icon
- @bitwarden/team-ui-foundation, when implementing the view component I observed that the copy/hide buttons on the form field were not present on the first render. I deduced this to be caused by the view component being the first to use the `OnPush` change detection. I switched how the form field used signals to react to changes after the lifecycle events.

## 📸 Screenshots

|Web|
|-|
|<video src="https://github.com/user-attachments/assets/919f0119-0ee6-450b-9617-a7e2c56b75d9" />|

|Extension|
|-|
|<video src="https://github.com/user-attachments/assets/47334657-ca28-4ca5-842b-e2ee3428f432"/>

|Desktop|
|-|
|<video src="https://github.com/user-attachments/assets/b8240039-0de4-4b28-85b8-3fa219055c12" />

|CLI - Template| CLI - Response|
|-|-|
|bank-account-cli-template|bank-account-cli|

[PM-32686]: https://bitwarden.atlassian.net/browse/PM-32686?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/common/src/vault/models/domain/bank-account.spec.ts
```diff
@@ -0,0 +1,175 @@
...
+
+    mockContainerService();
+  });
+
+  it("Convert", () => {
+    const bankAccount = new BankAccount(data);
+
+    expect(bankAccount).toEqual({
+      bankName: { encryptedString: "bankName", encryptionType: 0 },
+      nameOnAccount: { encryptedString: "nameOnAccount", encryptionType: 0 },
```
quexten: Can we please use actual test strings here? This is leaking implementation details of how EncString represents contents and not a stable representation that should be used for testing. Any refactor in KM's code would break this test and require changes here.

Also, encryption type 0 is not supported and should not be used even in tests.

libs/vault/src/services/password-reprompt.service.ts
```diff
@@ -23,7 +23,7 @@ export class PasswordRepromptService {
   enabled$ = Utils.asyncToObservable(() => this.userVerificationService.hasMasterPassword());
 
   protectedFields() {
-    return ["TOTP", "Password", "H_Field", "Card Number", "Security Code"];
+    return ["TOTP", "Password", "H_Field", "Card Number", "Security Code", "PIN", "Account Number"];
```
quexten: A question from the product perspective, how is account number a protected field but swift code and IBAN are not? Especially the IBAN in european bank accounts contains the account number, and fulfills the same function as an account number?

https://images.stripeassets.com/3sz5ney9ml0h/2bUyhTfOJOMB5c8Y2aZluh/1b619fe2b94bb274e34dd179dae4253b/21-IBAN-structure.png?w=1912&q=80&fm=webp

Image

libs/common/src/vault/models/domain/bank-account.spec.ts
nick-livefront: Yes, great call and appreciate the feedback! [50b92a3](https://github.com/bitwarden/clients/pull/19302/commits/50b92a303d0add92ff3ee716a217333fc97ba5b2)

libs/vault/src/services/password-reprompt.service.ts
nick-livefront: Yes! There was an internal discussion [here](https://livefront.slack.com/archives/C04N1TX1CGM/p1776412890270219) that surfaced the same issue with certain fields. I just didn't get around to updating this PR until now. 

[7191e08](https://github.com/bitwarden/clients/pull/19302/commits/7191e08197021725face1082e7a406744e3dc58e) + [e03ba0b](https://github.com/bitwarden/clients/pull/19302/commits/e03ba0b5446044d59b63ef5cac3b11aa0a5ee6dd)

libs/components/src/form-field/form-field.component.ts
willmartian, line 1: I don't know why were were doing it the way we were, but this looks much better and I see no Chromatic diffs. 💯

willmartian (dismissed):
UIF changes look great--thank you.

willmartian (commented):
It would be great to capture the new bank UIs in Storybook somewhere.

JaredScar (dismissed):
Admin Console changes look good, thanks!

Banrion (commented):
✅ DIRT files

quexten (dismissed):

ike-kottlowski (dismissed):

apps/cli/src/commands/get.command.ts
```diff
@@ -149,6 +157,8 @@ export class GetCommand extends DownloadCommand {
         return Response.error("Access to this item type is restricted by organizational policy.");
       }
 
+      decCipher = decCipher.filter((c) => isNewItemEnabled || c.type !== CipherType.BankAccount);
```
shane-melton: ❓ Should be filtering out bank account types if a user is explicitly trying to view/get a bank account item they've created? 

This should only happen if we enable the flag, user creates bank account cipher, we disable the flag, the user tries to view the bank account cipher. In that scenario, I would expect we still show the item otherwise it may be confusing if it appears the item just disappeared.

The list command also appears unmodified and will return bank account types, regardless of the flag, so we're a bit inconsistent between the commands currently.

libs/vault/src/services/copy-cipher-field.service.ts
```diff
@@ -78,6 +82,14 @@ const CopyActions: Record<CopyAction, CopyActionInfo> = {
...
   publicKey: { typeI18nKey: "sshPublicKey", protected: true },
   keyFingerprint: { typeI18nKey: "sshFingerprint", protected: true },
+  accountNumber: {
+    typeI18nKey: "accountNumber",
+    protected: true,
+    event: EventType.Cipher_ClientCopiedBankAccountNumber,
+  },
+  routingNumber: { typeI18nKey: "bankRoutingNumber", protected: false },
+  pin: { typeI18nKey: "pin", protected: true, event: EventType.Cipher_ClientCopiedBankAccountPin },
+  iban: { typeI18nKey: "iban", protected: true },
```
shane-melton, line 92: ❓ Should we have an EventType for IBANs since they're protected?

docs/cipher-types.md
shane-melton, line 1: ⛏️ This is missing the new bank account type. The documentation is great though!

shane-melton (commented):
This is looking really good! Just a few questions and a documentation nit.

docs/cipher-types.md
nick-livefront: Claude created this during out sessions, didn't consider the type it was implementing. Added [473ea57](https://github.com/bitwarden/clients/pull/19302/commits/473ea5764e505a75706de5c4639cf2b467c6a3d5)!

apps/cli/src/commands/get.command.ts
nick-livefront: That's a good point, unique scenario as well. I dug through the past commits and I couldn't find filtering for the SSHKey implementation. For that reason, I removed the current filtering [b9a7186](https://github.com/bitwarden/clients/pull/19302/commits/b9a718636576e7dd88bcbe3bf70c81e918a86053)

libs/vault/src/services/copy-cipher-field.service.ts
nick-livefront: Yes! This was a gap that was identified after we already implemented the server & sdk changes for bank account. I made a ticket to handle this at one shot so we can make the server + client change at on time: [PM-35656](https://bitwarden.atlassian.net/browse/PM-35656)

quexten (approved):

Banrion (approved):

harr1424 (approved):

mcamirault (approved):

shane-melton (approved):

willmartian (approved):

JaredScar (approved):

ike-kottlowski (approved):

JaredScar (approved):
AC changes look fine, thanks

PR #19308, opened 2026-02-28
[PM-32919] Migrate DeleteAccountDialog to shared code

djsmith85:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-32919
## 📔 Objective

In an effort to clean up the old `UserverificationComponent`, I looked to remove it from desktop. Instead of just replacing it by the new component in the `DeleteAccountDialog`, I noticed I could just re-use the dialog from web. So I decided to migrate that into shared code (`libs/auth/src/angular`) and then just update the calls on web and desktop.

After doing all the work, I thought there might some stories/tasks that were created to track the removal, which seems to be: https://bitwarden.atlassian.net/browse/PM-5243 

I then found a more recent PR: https://github.com/bitwarden/clients/pull/18529 which aims to update the existing `DeleteAccountComponent` to consume the new `UserVerificationFormInput`. The task has been marked as Abandoned in Jira. https://bitwarden.atlassian.net/browse/PM-30737  Instead as described above: I just used the existing `DeleteAccountDialogComponent` from web by moving it into shared code.

### Changes:

- **Move DeleteAccountDialogComponent from web to libs/auth/angular** - https://github.com/bitwarden/clients/commit/2d4cdda60fab8bd9510875369516132d8e68774d
  - Export via libs/auth/angular barrel file
  - Replace imports from web's SharedModule by direct imports
  - Update import of UserVerificationFormInputComponent
  - Add eslint ignore to imports from bitwarden/components
    - Not happy about this, but it follow the pattern of the other components. Also better to put it here, where the underlying UserVerification-components are.
- **Use new DeleteAccountDialogComponent on desktop** - https://github.com/bitwarden/clients/commit/f60fd0d4a5e873ee84e82bcb5c40e823bb66cab7
- **Delete desktop specific DeleteAccountComponent and UserVerificationComponent** - https://github.com/bitwarden/clients/commit/7893bb9a099af4b1c90e79093dc3ef33ea46e625

## 📸 Screenshots

- 2 users (account switching), delete and switch to other account

https://github.com/user-attachments/assets/9f4932fa-8996-40af-9247-a85a68da4685

- 1 user, delete and logout 

https://github.com/user-attachments/assets/5f746c3e-3c78-44f2-9932-0bf37e4cdc2a


libs/angular/src/auth/delete-account-dialog/delete-account-dialog.component.ts
JaredSnider-Bitwarden, line 1: Per sync discussion, we aren't putting any more code in `libs/auth` so can this be moved back to `libs/angular/src/auth`?

djsmith85: Addressed with f8e79c6

JaredSnider-Bitwarden:
For visibility, we discovered some issues with account deletion leaving the user in a bad state. Pending further investigation.

JaredSnider-Bitwarden (commented):
LGTM! Please ensure this is tested on the feature!

JaredSnider-Bitwarden (approved):

PR #19311, opened 2026-03-01
Remove masterkey

quexten:

### DIFF
--- a/libs/common/src/key-management/key-connector/services/key-connector.service.ts
+++ b/libs/common/src/key-management/key-connector/services/key-connector.service.ts
@@ -35,13 +35,12 @@ import { UserId } from "../../../types/guid";
 import { UserKey } from "../../../types/key";
 import { AccountCryptographicStateService } from "../../account-cryptography/account-cryptographic-state.service";
 import { KeyGenerationService } from "../../crypto";
-import { InternalMasterPasswordServiceAbstraction } from "../../master-password/abstractions/master-password.service.abstraction";
-import { SecurityStateService } from "../../security-state/abstractions/security-state.service";
 import { KeyConnectorService as KeyConnectorServiceAbstraction } from "../abstractions/key-connector.service";
 import { KeyConnectorDomainConfirmation } from "../models/key-connector-domain-confirmation";
 import { KeyConnectorUserKeyRequest } from "../models/key-connector-user-key.request";
 import { SetKeyConnectorKeyRequest } from "../models/set-key-connector-key.request";
-import { EncString } from "@bitwarden/sdk-internal";
+import { EncString, PureCrypto } from "@bitwarden/sdk-internal";
+import { CryptoFunctionService } from "../../crypto/abstractions/crypto-function.service";
 
 export const USES_KEY_CONNECTOR = new UserKeyDefinition<boolean | null>(
   KEY_CONNECTOR_DISK,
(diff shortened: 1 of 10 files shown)


libs/auth/src/common/login-strategies/login.strategy.ts
```diff
@@ -260,37 +255,32 @@ export abstract class LoginStrategy {
...
 
+    // Unlock the user by by setting the decrypted user-key depending on the login strategy.
+    await this.unlockUser(response, userId);
+    this.messagingService.send("loggedIn");
+
     return result;
   }
 
-  // The keys comes from different sources depending on the login strategy
-  protected abstract setMasterKey(response: IdentityTokenResponse, userId: UserId): Promise<void>;
```
quexten, line 280: Removed -- don't need this anymore for any flow.

libs/auth/src/common/login-strategies/login.strategy.ts
```diff
@@ -260,37 +255,32 @@ export abstract class LoginStrategy {
...
-
-  protected abstract setUserKey(response: IdentityTokenResponse, userId: UserId): Promise<void>;
-
-  protected abstract setAccountCryptographicState(
```
quexten, line 284: Doesn't warrant a flow-specific function and is moved to shared function

libs/auth/src/common/login-strategies/login.strategy.ts
```diff
@@ -260,37 +255,32 @@ export abstract class LoginStrategy {
...
+    await this.unlockUser(response, userId);
+    this.messagingService.send("loggedIn");
+
     return result;
   }
 
-  // The keys comes from different sources depending on the login strategy
-  protected abstract setMasterKey(response: IdentityTokenResponse, userId: UserId): Promise<void>;
-
-  protected abstract setUserKey(response: IdentityTokenResponse, userId: UserId): Promise<void>;
```
quexten, line 282: Renamed to unlockUser. Thinking on keys is the wrong abstraction layer here. We shoudld think in the action that happens, i.e "unlocking the user".

libs/auth/src/common/login-strategies/login.strategy.ts
```diff
@@ -260,37 +255,32 @@ export abstract class LoginStrategy {
...
-
-  protected abstract setAccountCryptographicState(
-    response: IdentityTokenResponse,
-    userId: UserId,
-  ): Promise<void>;
-
-  // Old accounts used master key for encryption. We are forcing migrations but only need to
-  // check on password logins
-  protected encryptionKeyMigrationRequired(response: IdentityTokenResponse): boolean {
```
quexten, line 291: No longer used

libs/common/src/key-management/key-connector/services/key-connector.service.ts
```diff
@@ -117,6 +115,10 @@ export class KeyConnectorService implements KeyConnectorServiceAbstraction {
     );
   }
 
+  setUserKeyFromUrl(keyConnectorUrl: string, keyConnectorkeyEncryptedUserKey: EncString, userId: UserId): Promise<void> {
+    throw new Error("Method not implemented.");
```
quexten, line 118: We should probably just move this to the unlock service. The unlock service can unlock via key connector directly.

libs/common/src/key-management/key-connector/services/key-connector.service.ts
```diff
@@ -128,13 +130,12 @@ export class KeyConnectorService implements KeyConnectorServiceAbstraction {
...
   async migrateUser(keyConnectorUrl: string, userId: UserId) {
-    const masterKey = await firstValueFrom(this.masterPasswordService.masterKey$(userId));
-    const keyConnectorRequest = new KeyConnectorUserKeyRequest(
-      Utils.fromBufferToB64(masterKey.inner().encryptionKey),
-    );
+    // TODO
+    //const masterKey = null;
 
     try {
-      await this.apiService.postUserKeyToKeyConnector(keyConnectorUrl, keyConnectorRequest);
```
quexten, line 137: TODO: Server should support writing master-key-wrapped user key. Then the procedure will be as follows:

- First sample a new key-connector-key
- Then wrap the user-key with KC key
- Then post kc key to KC
- Then post migrate with the KC-key-wrapped-user-key to server

PR #19317, opened 2026-03-02
[PM-32248] [PM-32233] [PM-32235] Block links from opening in electron window

coroiu:
## 📔 Objective

This PR is a combination of fixes for 3 different reported VULNs

### [PM-32248] [PM-32233] Block links from opening in electron window

Currently any anchor links `<a>` in the renderer will open a new electron window and navigate to that link there. This means opening potentially compromised websites inside of our electron process. This PR aims to prevent this by either:

- Redirecting "safe" URLs to the external browser by converting the call to a `shell.openExternal`
- Blocking "unsafe" URLs completely

### [PM-32235] Block unsafe open external call with arbitrary web vault url argument

To open a webpage in an external browser electron provides `shell.openExternal`. In some cases we've called this function with a variable that we didn't check if it was safe. One example of this is `Change master password` where we simply called `openExternal` with the environment `webVaultUrl`. However, if this value was tampered with then we could end up opening things like `file://` URLs. This PR adds a wrapper around `openExternal` which always checks the URL before opening it.

- Introduces new `SafeShell` class that wraps electrons `shell`
- Introduces safe `openExternal` that checks if URL is safe before calling it
- Replaces all usages of `shell.openExternal` with the new safe version
- Adds lint rule to prevent electrons function from being used directly

## 📸 Screenshots

[PM-32248]: https://bitwarden.atlassian.net/browse/PM-32248?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[PM-32233]: https://bitwarden.atlassian.net/browse/PM-32233?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[PM-32235]: https://bitwarden.atlassian.net/browse/PM-32235?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/desktop/src/main/window.main.ts
```diff
@@ -400,6 +401,20 @@ export class WindowMain {
       });
     });
 
+    this.win.webContents.setWindowOpenHandler(({ url }) => {
+      // For security reasons, we block all attempts to open new windows from the renderer process.
+      if (SafeUrls.canLaunch(url)) {
+        // If the url is considered safe, we open it with the default system handler (e.g. browser) instead of a new Electron window.
+        this.logService.debug(`Redirecting link to external browser: ${url}`);
+        void shell.openExternal(url);
```
coroiu: thought: maybe we should limit this to web urls instead of launchable urls

mandreko-bitwarden: Would that have any sort of complications with launching a custom scheme URI if a user saved it in their vault? For example, application `testapplication` registers a `testapplication://` URI handler. The user then adds a link to it in their vault. They click the "launch URL" button in Bitwarden and expect it to trigger their `testapplication` application. 

I'm unsure if it's applicable or not, or even a common occurrence. And I like limiting attack surface. However, I want to ensure it doesn't come at the cost of functionality either.

coroiu: I'm not modifying that logic here, that is triggered using IPC. I have a follow up PR where I am making changes to it though: https://github.com/bitwarden/clients/pull/19322/changes#diff-1784a6f85929cadf4eb5d3ed20ee506f4300491e44eed9433ea5754b682a73eeL70 but it seems like we've never allowed custom protocols (according to the code I linked)

libs/common/src/platform/misc/safe-urls.ts
```diff
@@ -1,29 +1,35 @@
...
+    "vnc://",
+    // https://docs.microsoft.com/en-us/windows-server/remote/remote-desktop-services/clients/remote-desktop-uri
+    "rdp://", // Legacy RDP URI scheme
+    "ms-rd:", // Preferred RDP URI scheme
+    "chrome://",
+    "iosapp://",
+    "androidapp://",
+  ],
+  /** URI schemes for web pages opened in an external browser */
+  WebUrl: ["https://", "http://"],
```
dani-garcia: This is only used for certain internal hardcoded links and the web vault, which requires HTTPS, so maybe we can even be a bit more strict and only allow HTTPS for `WebUrl`?

coroiu: Hmm, true

PR #19321, opened 2026-03-02
[PM-28926] Add bulk operations to repository

Hinton:
## 🎟️ Tracking

[bitwarden.atlassian.net/browse/PM-28926](https://bitwarden.atlassian.net/browse/PM-28926)
## 📔 Objective

Adds the new bulk operations to the client managed state. Also adds a test.

Companion PR to https://github.com/bitwarden/sdk-internal/pull/804


dani-garcia (dismissed):
New additions look good to me, but I just noticed that the old set/delete methods are using firstValueFrom()+update() rather than using the state value from update(). Should those be changed to match the new setBulk() & removeBulk()? Not sure what the difference is between both ways.

Hinton:
The difference is that something else could push state between the get. update would be atomic (i think)

PR #19339, opened 2026-03-03
[PM-33106]Harden WebAuthn message listener against sandboxed iframe

dan-livefront:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-33106
## 📔 Objective

Harden WebAuthn message listener with defense in layers, currently PM-33023 and PM-33019 are adding additional layers


apps/browser/src/autofill/fido2/content/messaging/messenger.ts
```diff
@@ -106,6 +106,9 @@ export class Messenger {
 
   private createMessageEventListener() {
     return async (event: MessageEvent<MessageWithMetadata>) => {
+      if (!window.origin || String(window.origin).toLowerCase() === "null") {
```
jprusik, line 111: It might be a good idea to include a comment explaining why we're checking for a string value of "null" (it's not an oversight, but certainly looks like one to the outside eye)

dan-livefront: Sounds good, I also wanted to point out here 

```
if (!window.origin || window.origin === "null") {
  return;
}
```

would have been sufficient but I wanted to have parity with `currentlyInSandboxedIframe`

PR #19340, opened 2026-03-03
[deps] Jest v30

cd-bitwarden:


.claude/settings.local.json
```diff
@@ -0,0 +1,10 @@
+{
+  "permissions": {
+    "allow": [
+      "mcp__ide__getDiagnostics",
+      "Bash(npm test:*)",
+      "Bash(xargs:*)",
+      "Bash(npm install:*)"
```
quexten, line 7: I don't think giving claude access to `xargs` or `npm install` is reasonable. F.e ` echo "" | xargs  sh -c 'mymaliciouscode` would just allow claude to run anything.

quexten (changes requested):
KM changes look good, but the claude changes here are concerning.

package.json
```diff
@@ -115,10 +115,11 @@
     "html-webpack-injector": "1.1.4",
     "html-webpack-plugin": "5.6.5",
     "husky": "9.1.7",
+    "jest": "30.2.0",
     "jest-diff": "30.2.0",
     "jest-junit": "16.0.0",
-    "jest-mock-extended": "3.0.7",
-    "jest-preset-angular": "14.6.2",
+    "jest-mock-extended": "4.0.0",
+    "jest-preset-angular": "15.0.3",
```
mzieniukbw, line 122: the lock file does not represent this change and contains many changes, including unexpected upgrades like `jest-preset-angular` to "16.0.0.".

mzieniukbw (changes requested):

.claude/settings.local.json
SaintPatrck, line 1: ⚠️ This file should not be tracked by git. It's meant to be a personal, local settings override.

PR #19341, opened 2026-03-03
[PM-28167] Desktop - migrate vault drawers UI to shared lib

iivins-livefront:
## Overview

The web and desktop vault applications previously each had their own implementation of the vault item dialog. This feature consolidates them so both apps consume a single shared component from the vault library, eliminating duplication and ensuring consistent behavior across platforms.

## Key Changes:

- Moved the authoritative vault item dialog implementation into `libs/vault` as the single source of truth, accessible to all vault consumers
- Updated the web application to import the dialog directly from the shared library, removing the app-level re-export shim that was the legacy of the original split
- Decoupled web service dependencies from the shared component. `RoutedVaultFilterService` is injected as optional and `PremiumUpgradePromptService` uses the common abstraction
- Updated the desktop application to open vault items using the shared dialog via `VaultItemDialogComponent.openDrawer()` through the existing drawer mechanism in the layout
- Implemented `upgradeConfirmed$` on `DesktopPremiumUpgradePromptService` to satisfy the shared component contract. Desktop uses EMPTY since it handles upgrade prompts inline without navigating away
- Removed redundant desktop behaviors that are now handled by the shared dialog, including event handling and the decryption failure dialog which is now managed within the shared component

## Summary

Both web and desktop now share a single vault item dialog implementation with zero web-specific dependencies in the shared library. Each platform provides its own service implementations where behavior differs, injected through the existing common abstractions rather than hardcoded platform logic.

## Testing Notes

Web vault:
- Open any vault item and confirm the dialog opens and all interactions work
- Verify report pages still open the dialog correctly when clicking a cipher
- Verify the admin console org vault dialog works as before

Desktop vault
- Enable the desktop-ui-migration-milestone-3 feature flag
- Click a vault item and confirm the drawer opens with full content rendered
- Test all actions within the drawer
- Click an item that requires premium for attachments and confirm the upgrade prompt opens and the drawer remains functional afterward
- Confirm decryption failure scenario opens the correct dialog and does not leave a blank drawer

## Screenshots

Screenshot 2026-03-03 at 11 28 10 AM
Screenshot 2026-03-03 at 11 28 24 AM
Screenshot 2026-03-03 at 11 28 46 AM


libs/vault/src/cipher-form/components/cipher-form.component.ts
Hinton, line 1: What is the purpose of these changes? The web dialog presumably does not use them?

apps/desktop/src/vault/app/vault-v3/vault-items/vault-item-drawer.component.ts
Hinton, line 1: This seems to be missing many of the archive logic that was recently added to web. Is it still mostly a copy from web, if so could we just share it? (I assume we would need to properly fix the `providers` usage but that seems pretty straightforward.

libs/vault/src/cipher-form/components/cipher-form.component.ts
iivins-livefront: Came up when testing this locally - it's an attempt to check for unsaved changes when closing / cancelling

The old vault approach queried the DOM with `document.querySelectorAll("vault-cipher-form .ng-dirty")` and I intentionally replaced that with a cleaner component API ( angular modernization skill picked it up )

Web also has its own `_cipherModified` tracking pattern but `CipherFormComponent` lives in `libs/vault`

apps/web/src/app/vault/components/vault-item-dialog/vault-item-dialog.component.spec.ts
Hinton, line 1: Did we intend to delete this file? The moved file seems to be a fraction of the old size.

apps/desktop/src/services/desktop-premium-upgrade-prompt.service.ts
```diff
@@ -15,6 +16,8 @@ export class DesktopPremiumUpgradePromptService implements PremiumUpgradePromptS
   private configService = inject(ConfigService);
   private dialogService = inject(DialogService);
 
+  readonly upgradeConfirmed$: Observable<boolean> = EMPTY;
```
Hinton, line 19: Technically this property is optional, and it seems like the `VaultItemDialogComponent` should check if it's null before using it.

apps/desktop/src/app/app.module.ts
```diff
@@ -59,6 +60,10 @@ import { SharedModule } from "./shared/shared.module";
       provide: PremiumUpgradePromptService,
       useClass: DesktopPremiumUpgradePromptService,
     },
+    {
+      provide: CipherFormGenerationService,
+      useClass: DesktopCredentialGenerationService,
+    },
```
Hinton: **issue:** This should be defined in the service module not in app module.

apps/web/src/app/vault/components/vault-item-dialog/vault-item-dialog.component.spec.ts
iivins-livefront: Intended to delete this file - did not intent to lose everything it covered, updating it now

libs/vault/src/vault-item-dialog/vault-item-dialog.component.ts
```diff
@@ -165,6 +149,14 @@ export class VaultItemDialogComponent implements OnInit, OnDestroy {
 
   private readonly dialogComponent = viewChild(DialogComponent);
 
+  // Optional — only provided by platforms that support it (e.g. desktop minimizeOnCopy)
+  private readonly copyClickListener = inject(COPY_CLICK_LISTENER, {
+    optional: true,
+  }) as CopyClickListener | null;
```
Hinton: **issue:** This seems unused?

libs/vault/src/vault-item-dialog/vault-item-dialog.component.ts
```diff
@@ -165,6 +149,14 @@ export class VaultItemDialogComponent implements OnInit, OnDestroy {
...
+
+  // Optional — only provided in web where routed vault filter exists
+  private readonly routedVaultFilterService = inject(RoutedVaultFilterService, { optional: true });
```
Hinton: **question:** Desktop should use `RoutedVaultFilterService`, is that not working? Or is it just not indicating that the trash filter is active?

libs/vault/src/vault-item-dialog/vault-item-dialog.component.ts
```diff
@@ -368,7 +359,6 @@ export class VaultItemDialogComponent implements OnInit, OnDestroy {
       if (this.cipher.decryptionFailure) {
         this.dialogService.open(DecryptionFailureDialogComponent, {
           data: { cipherIds: [this.cipher.id] },
-          positionStrategy: new CenterPositionStrategy(),
```
Hinton, line 366: **question:** Do we need to remove this? It seems benign to keep since it should just work.

libs/vault/src/vault-item-dialog/vault-item-dialog.component.ts
```diff
@@ -764,4 +754,16 @@ export class VaultItemDialogComponent implements OnInit, OnDestroy {
       },
     );
   }
+
+  /**
+   * Opens the VaultItemDialog as a drawer (desktop).
```
Hinton: **suggestion:** These comment will get stale quickly, can we just remove desktop & web from them?


```suggestion
   * Opens the VaultItemDialog as a drawer.
```

libs/vault/src/vault-item-dialog/vault-item-dialog.component.ts
```diff
@@ -1,15 +1,7 @@
-// FIXME: Update this file to be type safe and remove this and next line
+// FIXME(https://bitwarden.atlassian.net/browse/PM-XXXXX): Update this file to be type safe and remove this and next line
```
Hinton:
```suggestion
// FIXME: Update this file to be type safe and remove this and next line
```

Hinton (commented):
Review of the dialog changes

libs/vault/src/vault-item-dialog/vault-item-dialog.component.html
```diff
@@ -36,7 +36,7 @@
               {{ "attachments" | i18n }}
               <app-premium-badge></app-premium-badge>
             </div>
-            <i slot="end" class="bwi bwi-angle-right" aria-hidden="true"></i>
+            <bit-icon name="bwi-angle-right" slot="end" />
```
Hinton:
```
[Rend]   ERROR in ../../libs/vault/src/vault-item-dialog/vault-item-dialog.component.html:39:13 - error NG8001: 'bit-icon' is not a known element:
[Rend]   1. If 'bit-icon' is an Angular component, then verify that it is included in the '@Component.imports' of this component.
[Rend]   2. If 'bit-icon' is a Web Component then add 'CUSTOM_ELEMENTS_SCHEMA' to the '@Component.schemas' of this component to suppress this message.
[Rend]
[Rend]   39             <bit-icon name="bwi-angle-right" slot="end" />
[Rend]                  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
[Rend]
[Rend]     ../../libs/vault/src/vault-item-dialog/vault-item-dialog.component.ts:122:16
[Rend]       122   templateUrl: "vault-item-dialog.component.html",
[Rend]                          ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
[Rend]       Error occurs in the template of component VaultItemDialogComponent.
```

apps/desktop/src/vault/app/vault-v3/vault.component.ts
Hinton, line 1: Looks like we can remove some more imports.

```
[Rend]   WARNING in src/vault/app/vault-v3/vault.component.ts:140:5 - warning NG8113: I18nPipe is not used within the template of VaultComponent
[Rend]
[Rend]   140     I18nPipe,
[Rend]           ~~~~~~~~
[Rend]
[Rend]   WARNING in src/vault/app/vault-v3/vault.component.ts:144:5 - warning NG8113: PremiumBadgeComponent is not used within the template of VaultComponent
[Rend]
[Rend]   144     PremiumBadgeComponent,
[Rend]           ~~~~~~~~~~~~~~~~~~~~~
```

Hinton, line 1: There seems to be a horizontal scrollbar on the vault page now. 

Image

Hinton, line 1: `archiveFlagEnabled` and `hasArchivedCiphers$` seems unused?

apps/desktop/src/vault/app/vault-v3/vault.component.ts
```diff
@@ -1005,98 +869,45 @@ export class VaultComponent<C extends CipherViewLike>
...
-    if (this.action() === action && (!cipher || this.cipherId === cipher.id)) {
-      return false;
-    } else if (this.dirtyInput() && (await this.wantsToSaveChanges())) {
-      return false;
-    }
-    return true;
+    });
   }
 
+  /** Refresh the current cipher object */
```
Hinton: This comment doesn't make sense.

Banrion (dismissed):
✅ DIRT files

libs/vault/src/vault-item-dialog/vault-item-dialog.component.ts
iivins-livefront: Fixed this by adding `filter?: RoutedVaultFilterModel` to `VaultItemDialogParams`.

Dialog now prefers `params.filter` first and falls back to the service - Desktop's `openDialog` reads the current filter from its own `routedVaultFilterService` and passes it through.

apps/desktop/src/services/desktop-premium-upgrade-prompt.service.ts
iivins-livefront: Added `?.pipe()` so if `upgradeConfirmed$` is null or undefined, the whole chain is skipped rather than throwing

Hinton: Can we undo the changes to this service in that case?

libs/vault/src/vault-item-dialog/vault-item-dialog.component.ts
```diff
@@ -165,6 +155,10 @@ export class VaultItemDialogComponent implements OnInit, OnDestroy {
 
   private readonly dialogComponent = viewChild(DialogComponent);
 
+  // Optional — only provided by platforms that support it (e.g. desktop minimizeOnCopy)
```
Hinton: This comment is orphant?

apps/desktop/src/services/desktop-premium-upgrade-prompt.service.ts
Hinton: Not sure why we reverted decision here. But if it's needed outside this component we should update the abstract that `upgradeConfirmed$` is required in the service.

libs/vault/src/vault-item-dialog/vault-item-dialog.component.ts
```diff
@@ -164,6 +156,8 @@ export class VaultItemDialogComponent implements OnInit, OnDestroy {
 
   private readonly dialogComponent = viewChild(DialogComponent);
 
+  private readonly routedVaultFilterService = inject(RoutedVaultFilterService, { optional: true });
```
shane-melton: ⚠️ `RoutedVaultFilterService` is [provided at the component level](https://github.com/bitwarden/clients/blob/124c04503d81267f82d6dddd59b02b9c09d3c608/apps/web/src/app/vault/individual-vault/vault.component.ts#L162) in the Web Vault and won't be provided by the `DialogService` injector. ([Desktop provides it at the services module](https://github.com/bitwarden/clients/blob/46bb683d7be6f417258b590108471ae7a08d6bdc/apps/desktop/src/app/services/services.module.ts#L567-L571))

Instead of passing filter through params or injecting `RoutedVaultFilterService`, the dialog could just use `cipher.isDeleted` to drive trash-specific UI (Restore button, Close vs Cancel, hiding Edit).
  Deleted ciphers are already filtered out of non-trash views, so `cipher.isDeleted` is functionally equivalent to `isTrashFilter` whenever the dialog is open. This would eliminate the routing dependency
  entirely and simplify the dependencies.

apps/desktop/src/vault/app/vault-v3/vault.component.ts
```diff
@@ -147,46 +129,16 @@ type EmptyStateMap = Record<EmptyStateType, EmptyStateItem>;
...
-    },
-    {
-      provide: PremiumUpgradePromptService,
-      useClass: DesktopPremiumUpgradePromptService,
-    },
-    { provide: CipherFormGenerationService, useClass: DesktopCredentialGenerationService },
-    {
-      provide: COPY_CLICK_LISTENER,
-      useExisting: VaultComponent,
-    },
```
shane-melton, line 196: ⚠️ This `COPY_CLICK_LISTENER` provider is used to send a `minimizeOnCopy` message vai the electron ipc to minimize the desktop application whenever a cipher field is copied. 

Since the cipher view is now being opened in a dialog, we'll need to move the provider somewhere else. The provider could be moved to a root-level service (e.g., a `DesktopCopyListenerService` provided in the Desktop `AppModule`) so it's available to `CopyClickDirective` instances inside the drawer.

shane-melton (changes requested):
Looking really good, nice work! I just see these two issues with regards to the provider changes that will likely break functionality around the restore buttons in Web and minimize on copy for desktop.

libs/vault/src/vault-item-dialog/vault-item-dialog.component.ts
Hinton: In desktop `RoutedVaultFilterService` is provided by the root `ServiceModule`.

apps/desktop/src/vault/app/vault-v3/vault.component.ts
Hinton: We should register this in the root `ServiceModule` so it's available everywhere.

iivins-livefront:
> Looking really good, nice work! I just see these two issues with regards to the provider changes that will likely break functionality around the restore buttons in Web and minimize on copy for desktop.

I replaced the `isTrashFilter`, which relied on `RoutedVaultFilterService`, with `cipher.isDeleted` directly in the shared component. That should remove the routing dependency from the shared lib entirely from my understanding

Also moved that `COPY_CLICK_LISTENER` logic into a dedicated `DesktopCopyListenerService` and registered it in the root `ServiceModule` as suggested

libs/vault/src/vault-item-dialog/vault-item-dialog.component.ts
coroiu: I'm completely out of the loop here, but the reason it's provided at the component level is so that it can get access to `ActivatedRoute`

apps/desktop/src/app/services/services.module.ts
```diff
@@ -574,6 +587,12 @@ const safeProviders: SafeProvider[] = [
     useClass: RoutedVaultFilterBridgeService,
     deps: [Router, RoutedVaultFilterService, VaultFilterServiceAbstraction],
   }),
+  safeProvider({
+    provide: COPY_CLICK_LISTENER as SafeInjectionToken<CopyClickListener>,
+    useFactory: (messagingService: MessagingServiceAbstraction) =>
+      new DesktopCopyListenerService(messagingService),
+    deps: [MessagingServiceAbstraction],
```
Hinton: **issue:** Please use `useClass`.

```suggestion
    provide: COPY_CLICK_LISTENER,
    useClass: DesktopCopyListenerService,
    deps: [MessagingServiceAbstraction],
```

**suggestion:** Update `COPY_CLICK_LISTENER` to use `SafeInjectionToken`.

apps/web/src/app/vault/services/web-vault-extension-prompt.service.ts
Hinton, line 1: **question:** Did you indent do edit this file?

iivins-livefront: Kind of? I can pull it - but this got noticed when I was getting the imports and test files re-synced ( Claude probably saw the file "open" but I didn't end up editing the imports )

Hinton (dismissed):

Banrion (dismissed):

Hinton (dismissed):

apps/desktop/src/vault/app/vault-v3/vault.component.ts
```diff
@@ -1004,98 +830,45 @@ export class VaultComponent<C extends CipherViewLike>
...
+      }
+      this.activeDrawerRef.close();
     }
-    this.router
-      .navigate([], {
-        relativeTo: this.route,
-        queryParams: queryParams,
+    this.activeDrawerRef = VaultItemDialogComponent.openDrawer(this.dialogService, {
+      mode,
+      formConfig,
```
shane-melton, line 892: ⚠️ We're missing the `restore` action here. Without it, the restore button in the dialog footer will only close the dialog without restoring. We should be able to re-use the existing `handleRestoreEvent()` in this component.

shane-melton (changes requested):
Looking really good! Just one final issue around the restore functionality in Desktop.

jrmccannon (dismissed):

ike-kottlowski (dismissed):
Auth changes are good.

shane-melton (dismissed):
Changes look good! Thanks for lifting the `VaultItemDialog` to libs and all the work that came with it.

Hinton (dismissed):

dani-garcia (dismissed):

jrmccannon (dismissed):

BryanCunningham (dismissed):

Banrion (dismissed):

apps/web/src/app/vault/individual-vault/vault.component.ts
```diff
@@ -926,6 +926,7 @@ export class VaultComponent<C extends CipherViewLike> implements OnInit, OnDestr
       formConfig,
       activeCollectionId,
       restore: this.restore,
+      isAdminConsoleAction: true,
```
Hinton: @shane-melton I don't believe this is correct right? Individual vault should not pretend to be admin console.

shane-melton: Correct, this should not be set to `true` for the individual vault.

shane-melton (changes requested):

shane-melton (dismissed):

dani-garcia (dismissed):
Platform changes lgtm

Banrion (dismissed):

jrmccannon (dismissed):

shane-melton (approved):

jrmccannon (approved):

dani-garcia (approved):

Banrion (approved):

Hinton (approved):

theMickster:
Hello @iivins-livefront et. al,

A local multi-agent Claude Code review produced the following report (I'm field testing a forthcoming addition to the plugin marketplace). Kindly review at your convenience and lemme know if the findings are legitimate or not. If not, it would be beneficial to know briefly what context Claude may be missing or misunderstanding to make better assessments of our code.

<details>
<summary>Code Review: [PM-28167] Desktop - migrate vault drawers UI to shared lib (#19341)</summary>

**Date:** 2026-04-16 | **Reviewed by:** Claude Code (Sonnet)

## Summary

| Severity      | Count |
| ------------- | ----- |
| 🛑 Blocker    | 0     |
| ⚠️ Important  | 1     |
| ♻️ Refactor   | 0     |
| 💡 Suggestion | 0     |

This PR is a well-structured consolidation of the vault item dialog into `libs/vault`, cleanly decoupling platform-specific service dependencies. The `asAdmin` deletion gate change is a security improvement. One functional regression was found: the desktop `editCipher` method drops the `partial-edit` mode for users with limited edit permissions.

## Findings

### ⚠️ Important

#### `editCipher` always passes `"edit"` mode, dropping `partial-edit` for restricted users

`apps/desktop/src/vault/app/vault-v3/vault.component.ts:597`

  The new `editCipher` method always builds the form config with mode `"edit"`:

  ```typescript
  const formConfig = await this.cipherFormConfigService.buildConfig(
    "edit",                    // always "edit" — no cipher.edit check
    cipher.id as CipherId,
    cipher.type,
  );
  await this.openDialog("form", formConfig);
  ```

  The old code explicitly downgraded to `"partial-edit"` when the user lacked full edit permission:

  ```typescript
  await this.buildFormConfig("edit");
  if (!cipher.edit && this.config) {
    this.config.mode = "partial-edit";
  }
  ```

  `DefaultCipherFormConfigService.buildConfig` does **not** internally handle this downgrade — it passes the mode through verbatim. The shared `VaultItemDialogComponent.ngOnInit` has a guard that forces `view` mode when `!canEdit && formConfig.mode !== "partial-edit"`. So for ciphers with `cipher.edit === false`, the user is forced into read-only view instead of partial-edit, losing the ability to update fields they do have permission to modify.

  Notably, `viewCipher` in the same file correctly uses the ternary at line ~567:
  ```typescript
  cipher.edit ? "edit" : "partial-edit"
  ```

  **Impact:** Organization members with view-only collection access who click "Edit" on a shared cipher in desktop vault-v3 will see a read-only view dialog instead of the partial-edit form. Silent loss of functionality.

  **Suggested fix:**
  ```typescript
  const formConfig = await this.cipherFormConfigService.buildConfig(
    cipher.edit ? "edit" : "partial-edit",
    cipher.id as CipherId,
    cipher.type,
  );
  ```

  **Confidence:** 95/100

  </details>

## Reviewed and Dismissed

   <details><summary>🔍 12 initial findings dismissed after validation</summary>

   #### `takeUntilDestroyed` subscriptions accumulate on repeated drawer opens
   `apps/desktop/src/vault/app/vault-v3/vault.component.ts:~597`
   **Dismissed because:** `DrawerDialogRef._closed` is a Subject that calls `.complete()` after `.next()`, auto-unsubscribing all subscribers. `DialogService.openDrawer()` also calls `activeDrawer?.close()` before opening a new one. No subscription accumulation occurs.

   #### `firstValueFrom(dialogRef.closed)` without `.catch()` in `openAttachmentsDialog`
   `apps/desktop/src/vault/app/vault-v3/vault.component.ts:~579`
   **Dismissed because:** `DialogRef.closed` always emits exactly one value before completing. `firstValueFrom` cannot throw `EmptyError`. The old `.catch()` was unnecessary defensive code; ~60 other call sites in the codebase use this pattern without `.catch()`.

   #### `asAdmin` logic change is a behavioral regression for web admin console
   `libs/vault/src/vault-item-dialog/vault-item-dialog.component.ts:~728`
   **Dismissed because:** This is a security improvement, not a regression. The individual vault never passes `isAdminConsoleAction: true`, so the old code incorrectly elevated personal vault deletes to admin-scoped API calls. The admin console path (`isAdminConsoleAction: true`) is unaffected.

   #### Missing `Cipher_ClientViewed` event collection in desktop `viewCipher`
   `apps/desktop/src/vault/app/vault-v3/vault.component.ts:~571`
   **Original confidence:** 83/100
   **Dismissed because:** The shared `VaultItemDialogComponent.ngOnInit` collects the `Cipher_ClientViewed` event. The removal from the desktop caller is correct — event collection moved to the shared component.

   #### `openDialog` doesn't wait for old drawer close animation before opening new one
   `apps/desktop/src/vault/app/vault-v3/vault.component.ts:~597`
   **Dismissed because:** `DrawerDialogRef.close()` is fully synchronous — `DrawerService.close()` calls `this.portal.set(undefined)` which immediately removes the drawer from the DOM via an Angular signal. No async animation pathway exists.

   #### `restore` arrow function wrapper is unnecessary indirection
   `apps/desktop/src/vault/app/vault-v3/vault.component.ts:~609`
   **Dismissed because:** The wrapper serves two real purposes: adapting the `CipherViewLike` parameter type to `CipherView` (required by `handleRestoreEvent`) and adapting the `Promise<boolean>` return to `Promise<void>` (required by `VaultItemDialogParams.restore`).

   #### `upgradeConfirmed$: Observable<boolean> = EMPTY` duplicated across three implementations
   `apps/browser/src/vault/popup/services/browser-premium-upgrade-prompt.service.ts:~18`
   **Original confidence:** 78/100
   **Dismissed because:** Follows the codebase pattern of purely abstract service contracts with no concrete defaults. The codebase uses separate noop/unsupported concrete classes for this purpose. Three one-line assignments is not meaningful duplication.

   #### Split import paths for `VaultItemDialogResult` within the PR
   **Original confidence:** 92/100
   **Dismissed because:** The PR correctly updates all consumer imports to use `@bitwarden/vault`. The architecture agent read stale main branch files instead of applying the diff.

   #### `upgradeConfirmed$` is optional but consumed unconditionally
   `libs/common/src/vault/abstractions/premium-upgrade-prompt.service.ts:~9`
   **Original confidence:** 87/100
   **Dismissed because:** The PR changes the property from optional (`abstract upgradeConfirmed$?: Observable<boolean>`) to required (`abstract upgradeConfirmed$: Observable<boolean>`) and adds implementations in all three platform services plus the emergency view noop.

   #### `VaultItemDialogComponent` in `libs/vault` bakes in app-specific providers
   `libs/vault/src/vault-item-dialog/vault-item-dialog.component.ts`
   **Original confidence:** 83/100
   **Dismissed because:** The PR removes `WebVaultPremiumUpgradePromptService`, `WebCipherFormGenerationService`, and `RoutedVaultFilterService` from the component's providers. The decoupling is complete.

   #### `openDrawer` static method missing or incomplete
   `libs/vault/src/vault-item-dialog/vault-item-dialog.component.ts`
   **Original confidence:** 77/100
   **Dismissed because:** The PR adds both `static open()` (for web modal) and `static openDrawer()` (for desktop drawer) to the shared component.

   #### `PremiumUpgradePromptNoop` in emergency view dialog not updated
   `apps/web/src/app/auth/settings/emergency-access/view/emergency-view-dialog.component.ts`
   **Original confidence:** 78/100
   **Dismissed because:** The PR adds `readonly upgradeConfirmed$: Observable<boolean> = EMPTY` to the noop class, satisfying the updated interface contract.

   </details>

</details>

[PM-28167]: https://bitwarden.atlassian.net/browse/PM-28167?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

BryanCunningham (approved):

ike-kottlowski (approved):

PR #19342, opened 2026-03-03
[pm-32250] fix: default to "No Folder" in cipher save prompt

bensbits91:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-32250

## 📔 Objective

Update folder selection logic to default to "No Folder" when prompting user to save a new cipher. 

This addresses a reported issue that causes the first folder (alphabetically) to be selected by default. If the user doesn't notice, the cipher will be unexpectedly categorized into a user-created folder, leading to confusing, disruption and frustration.

## 📸 Screenshots

After this fix, "No Folder" is the default folder selected when prompting the user to save the cipher:

image


apps/browser/src/autofill/content/components/notification/button-row.ts
```diff
@@ -83,8 +83,8 @@ export function NotificationButtonRow({
           {
             icon: Folder,
             text: name,
-            value: id === null ? defaultNoneSelectValue : id,
-            default: id === null,
+            value: !id ? defaultNoneSelectValue : id,
+            default: !id,
```
jprusik, line 87: Ah! So what was this coming back as? I presume `undefined`?

bensbits91: It was coming back as an empty string `""`.

The "No Folder" FolderView is created in `folder.service.ts` via new FolderView(), and FolderView.id is typed as string with a default of "":

```ts
export class FolderView implements View, ITreeNodeObject {
  id: string = "";
  ...
}
```

"No Folder" has id of `""` so the original check `id === null` resulted in "No Folder" never being identified as the default option.

PR #19343, opened 2026-03-03
[BRE-1550] Added cherry-pick workflow

gitclonebrian:
## 🎟️ Tracking

[BRE-1550](https://bitwarden.atlassian.net/browse/BRE-1550)

## 📔 Objective
Adding the [cherry-pick.yml](https://github.com/bitwarden/gh-actions/blob/main/cherry-pick) workflow to the `clients` repo. The objective is to help remove some of the process friction that was introduced when a repository rule set was added that now requires a PR to merge into `rc` branches. The [`README`](https://github.com/bitwarden/gh-actions/blob/main/cherry-pick/README.md) for the workflow provides the details of the workflow, but the short description is that the workflow can be run via workflow dispatch to automatically cherry-pick any PR that's been merged into `main` into an `rc` branch.

[BRE-1550]: https://bitwarden.atlassian.net/browse/BRE-1550?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -145,6 +145,8 @@ apps/web/src/translation-constants.ts @bitwarden/team-platform-dev
 .github/workflows/stale-bot.yml @bitwarden/team-platform-dev
 .github/workflows/test.yml @bitwarden/team-platform-dev
 .github/workflows/version-auto-bump.yml @bitwarden/team-platform-dev
+.github/workflows/cherry-pick.yml @bitwarden/team-platform-dev
+
 # ESLint custom rules
 libs/eslint @bitwarden/team-platform-dev
 libs/eslint/components @bitwarden/team-ui-foundation


.github/workflows/cherry-pick.yml
```diff
@@ -0,0 +1,35 @@
...
+        description: 'Pull request to cherry pick from. (e.g., 456)'
+        required: true
+        type: string
+      dest_branch:
+        description: 'Destination branch to cherry pick to. (e.g., hotfix-rc)'
+        required: true
+        type: choice
+        options:
+          - rc
+          - hotfix-rc
```
trmartin4: This will need to be expanded to include the client hotfix branches:
- hotfix-rc-web
- hotfix-rc-desktop
- hotfix-rc-browser
- hotfix-rc-cli

The hotfix-rc branch can be removed from the list as well.

gitclonebrian: Thanks @trmartin4 , `hotfix-rc` has been removed, and the other branches added.

trmartin4:
@gitclonebrian what are your thoughts on ownership?  Would this be owned by BRE or Platform?  We'll want to update `CODEOWNERS` for any new workflows.  Since this is related to the development process, I'd think it would fall to Platform, but I'd be curious if you have any other thoughts.

gitclonebrian:
> @gitclonebrian what are your thoughts on ownership? Would this be owned by BRE or Platform? We'll want to update `CODEOWNERS` for any new workflows. Since this is related to the development process, I'd think it would fall to Platform, but I'd be curious if you have any other thoughts.

@trmartin4 , I agree, I think it makes the most sense for Platform to own it. I'll make the change to CODEOWNERS and push that to this branch.

PR #19347, opened 2026-03-03
[PM-33061] Tax Id Should Show for Teams and Enterprise Plans When Upgrading

sbrown-livefront:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-33061

## 📔 Objective

Adds validation and UI changes to the Premium to organization upgrade flow for Tax IDs. 

**Validation logic updates:**

* Added a check in `PremiumOrgUpgradeService` to require a tax ID for non-US billing addresses unless the selected plan is `Families`. If missing, an error is thrown.

**UI improvements:**

* Updated the billing address form so that the tax ID field is only shown for plans other than "Families." This is controlled by the new `showTaxIdField` computed property in the component.

**Type and data structure changes:**

* Modified the `UpgradePremiumToOrganizationRequest` type to include `taxId` in the billing address payload.

NOTE: Related to https://github.com/bitwarden/server/pull/7131 changes.
## 📸 Screenshots

https://github.com/user-attachments/assets/c7a0c02a-fc29-4a91-8f33-1130ac8fdc46

https://github.com/user-attachments/assets/7bd1d470-aee5-461f-9def-1fba97e955bd


apps/web/src/app/billing/individual/upgrade/premium-org-upgrade-payment/services/premium-org-upgrade.service.ts
```diff
@@ -89,6 +89,14 @@ export class PremiumOrgUpgradeService {
       throw new Error("Billing address information is incomplete");
     }
 
+    if (
+      billingAddress?.country != "US" &&
```
kdenney: ⛏️ It's not likely to matter but this should probably be using !== strict comparison operator.

apps/web/src/app/billing/individual/upgrade/premium-org-upgrade-payment/premium-org-upgrade-payment.component.ts
```diff
@@ -136,6 +136,10 @@ export class PremiumOrgUpgradePaymentComponent implements OnInit, AfterViewInit
     () => this.PLAN_MEMBERSHIP_MESSAGES[this.selectedPlanId()] ?? "",
   );
 
+  protected readonly showTaxIdField = computed<boolean>(() => {
+    return this.selectedPlanId() == PersonalSubscriptionPricingTierIds.Families ? false : true;
```
kdenney: ⛏️ Same here, should use strict comparison. Also, to make this more readable, I'd prefer if it were: 

```
return this.selectedPlanId() !== PersonalSubscriptionPricingTierIds.Families;
```

sbrown-livefront: Thanks for this. Addressed in https://github.com/bitwarden/clients/pull/19347/commits/46ebf310e2f95e085866b458abde6c4c54c56d0c

apps/web/src/app/billing/individual/upgrade/premium-org-upgrade-payment/services/premium-org-upgrade.service.ts
sbrown-livefront: Addressed in https://github.com/bitwarden/clients/pull/19347/commits/46ebf310e2f95e085866b458abde6c4c54c56d0c

PR #19354, opened 2026-03-03
[PM-31928] Cleanup backwards compatibility code in extension lock service

Thomas-Avery:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-31928

## 📔 Objective

Remove the backwards compatibility code for old biometrics ipc protocol. This is safe to remove now.

To test properly we need the fix from `km/pm-31630` so I based this off that.


apps/browser/src/key-management/lock/services/extension-lock-component.service.ts
```diff
@@ -66,23 +66,18 @@ export class ExtensionLockComponentService implements LockComponentService {
   }
 
   getAvailableUnlockOptions$(userId: UserId): Observable<UnlockOptions> {
+    // Note: defer is preferable b/c it delays the execution of the function until the observable is subscribed to
```
mzieniukbw: 🤔  i don't think this comment is accurate anymore and might be removed, since the the "function" mentioned is `getBiometricsStatusForUser`, which will only execute once the `getAvailableUnlockOptions$` observable is subscribed to, rather than immediately - rather expected behaviour.
(Plus we have a comment just below this one explaining this)

Thomas-Avery: I read this as more of describing why `defer` was used within the `combineLatest` for all of them because we also have:

```
  defer(() => this.pinService.isPinDecryptionAvailable(userId)),
      defer(async () => {
        const available = await this.webAuthnPrfUnlockService.isPrfUnlockAvailable(userId);
        return { available };
      }),
```


Is the comment even useful or should we just take it out?

mzieniukbw: I would take it out, it provides next to no value. The reason why we use still use defer for Pin and WebAuthn is simply because we don't have reactive equivalent.

mzieniukbw (approved):

Issue #19073, opened 2026-02-18
Please reduce the number of key strokes from 30 to 2 to get between the search box and search results


frastlin:
### Steps To Reproduce

1. Press enter on the search box, search for something, and press enter.
2. Tab until you reach the search results. This is 30 tabs.

### Expected Result

Instead of tabbing 30 times, I would like the search results to be the first thing that I land on when I press tab from the search box.

### Actual Result

Currently, it takes 30 tabs to get between the edit box and the first search result. It is possible to press enter on the "skip to main content" link, which is why I am still using Bitwardin, but the user experience is still tab 3 times, press enter, then tab through the results.

### Additional Context

I am on windows 11. I am a keyboard only user. 99.9% of the time when I am opening Bitwardin, I am searching first.

### Operating System

Windows

### Operating System Version

Windows 11

### Installation method

Direct Download (from bitwarden.com)

### Build Version

Version 2026.1.0 SDK 'main (8e5efd8)' Shell 39.2.6 Renderer 142.0.7444.226 Node 22.21.1 Architecture x64

### Issue Tracking Info

- [x] I understand that work is tracked outside of GitHub. A PR will be linked to this issue should one be opened to address it, but Bitwarden doesn't use fields like "assigned", "milestone", or "project" to track progress.

pamperer562580892423:
Probably a duplicate of https://github.com/bitwarden/clients/issues/18842

TroyBW:
Hi there!

Thank you for your report, it seems like it is a duplicate of this one: [https://github.com/bitwarden/clients/issues/18842](https://github.com/bitwarden/clients/issues/18842). 

If you wish to add any further information/screenshots/recordings etc., please feel free to do so at any time in there - our engineering team will be happy to review these.

This issue will now be closed.

Thanks!

ruant:
@TroyBW was you plan on closing this one in favor of the one you linked?

Also, any timeline on the other issue? This issue breaks the usage of Bitwarden pretty bad.

PR #19355, opened 2026-03-04
[PM-31704] - fix tab focus with filled search input

### LINKED ISSUE #18748: Firefox extension popout incorrectly traps focus when searching
### Steps To Reproduce

1. launch Firefox
2. tab to / click on the Bitwarden Password Manager icon (in the toolbar, under the Extensions menu, or in the sidebar)
3. log in to Bitwarden (the search field is auto-focused)
4. type one or more characters to filter the list
5. attempt to tab into an item under "Autofill suggestions" (this fails: focus will move somewhere--maybe the popout itself?--then the "New" button, then the "Pop out to a new window" button, then the "Bitwarden account ..." button,  then the "Search" field and repeat)
6. tab back to the "Search" field
7. backspace until there are no characters in the field
8. attempt to tab into an item under "Autofill suggestions" (this works)

---

- the same issue does not occur in the Chrome version of the extension
- Firefox 147.0.2
- GNOME Shell 49.2

### Expected Result

Having filtered the list of vault items using the Search field, I expect to be able to tab to items in the vault.

### Actual Result

Having filtered the list of vault items using the Search field, it is impossible to tab to any item in the vault.

### Screenshots or Videos

_No response_

### Additional Context

Discovered during a web page accessibility audit. The search becomes totally unusable when using only keyboard navigation.

### Operating System

Linux

### Operating System Version

Nixos 25.11.20260203.e576e3c (Xantusia)

### Web Browser

Firefox

### Browser Version

147.0.2

### Environment Versions

> Author Bitwarden Inc.
> Version 2025.12.1
> Last Updated January 14, 2026
> Homepage https://bitwarden.com/
> Rating 3,014 reviews

(There's no such menu as "Settings" in the extension)

### Issue Tracking Info

jaasen-livefront:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-31704

fixes #18748

### Problem

If the extension search bar has any input, pressing Tab produces incorrect focus order: it attempts to focus on the "X" clear button but it's removed from the DOM so focus is set on the document. The next tab then focuses on the "+ New" button, then the account profile button, and loops — skipping the vault filter toggle and item list entirely.

**Root cause:** The `bit-search` component's reset button was conditionally rendered with `*ngIf="searchText && showResetButton()"`. `showResetButton()` is a computed signal that returns `true` only when the input is focused or the form is hovered. When the user presses Tab:

1. The input fires `blur` → `isInputFocused` becomes `false`
2. `showResetButton()` becomes `false` → `*ngIf` removes the button from the DOM
3. The browser loses its tab target and resets focus to the document body
4. Subsequent Tabs cycle through elements from the top of the page

### Fix

**`search.component.ts`** — Added an `isResetButtonFocused` signal and included it in the `showResetButton` computed, so the button remains visible while it has keyboard focus.

**`search.component.html`** — Changed the `*ngIf` to depend only on `searchText` (keeping the button in the DOM whenever there is search text) and added `(focus)`/`(blur)` handlers on the button to drive the new signal.

The button's visual appearance is still controlled by the existing `tw-opacity-0` / `tw-bg-text-muted` `ngClass` toggle — no visual change when the button is not focused/hovered.

### Tabbing issue

Before

https://github.com/user-attachments/assets/94d67fcc-3516-46f0-be5a-5172de642bd6

After

https://github.com/user-attachments/assets/ea12da49-88e8-4852-9d27-c571aabb0139

### Input clear issue

https://github.com/user-attachments/assets/98647cbc-2f0e-4489-ae3b-a737d6484c79


vleague2 (approved):
Seems like a good improvement to me!

djsmith85:
@jaasen-livefront / @vleague2 Do you agree, that this likely also solves the following issues on desktop: https://github.com/bitwarden/clients/issues/19073 and https://github.com/bitwarden/clients/issues/18842 ?

vleague2:
@djsmith85 We can certainly test it in desktop and see if it fixes it. (I don't know if desktop is using our search component, off the top of my head, but probably in the new refreshed version it is)

vleague2:
@djsmith85 Okay tested desktop on this branch, and it appears that this solves both bug reports **on the refreshed version of the desktop UI**. There is one remaining bug where using the keyboard to clear the search bar causes focus to be lost and returned to the top of the page, but given that both reports are more specifically about entering text and then proceeding to search results without clearing, I think we are good to say that they will be resolved.

jaasen-livefront:
> @djsmith85 Okay tested desktop on this branch, and it appears that this solves both bug reports **on the refreshed version of the desktop UI**. There is one remaining bug where using the keyboard to clear the search bar causes focus to be lost and returned to the top of the page, but given that both reports are more specifically about entering text and then proceeding to search results without clearing, I think we are good to say that they will be resolved.

@vleague2 Nice! 3 birds, one stone. I'm holding off merging until RC is cut on Monday.

Is there a ticket for that search clearing bug? Even though it's highly unlikely a keyboard user would take that action I can jump on that if needed.

I can also easily add it to this ticket.

jaasen-livefront:
@vleague2 I went ahead and added the search clear fix since it was quite straightforward.

vleague2:
Nice, thank you! Can you pull in latest? I am seeing build failures that I think will be resolved with that

mcint:
Noticed Codecov flagged zero coverage on the changed lines — I drafted a .spec.ts that encodes the tab-focus regression. Happy to open a follow-up PR if that'd be useful

djsmith85:
@mcint Please do, I'm sure the team appreciates more coverage with added tests, especially if it can catch accessibility issues.

mmikeww:
apparently this fix is included in desktop app version 2026.3.1

however the functionality is unchanged and still broken

see #18842

jaasen-livefront:
@vleague2 @djsmith85 A community member brought up a good point in https://github.com/bitwarden/clients/issues/18842 regarding some sub-optimal UX when tabbing with search results in desktop that are still present after this fix. I think it's worth discussing a better tab flow in this context. Some suggestions were also provided.

PR #19358, opened 2026-03-04
[PM-32856] Make sync non-blocking on desktop, browser

quexten:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-32856

## 📔 Objective

Desktop and browser wait for a full sync or at least the revision date request, even if they are up-to-date. With a slow network or a high latency to the server this adds a few hundred msec to the unlock time. This PR eliminates that unlock time.


libs/key-management-ui/src/lock/components/lock.component.ts
```diff
@@ -612,11 +612,13 @@ export class LockComponent implements OnInit, OnDestroy {
...
-    const startSync = new Date().getTime();
-    // TODO: This should probably not be blocking
-    await this.syncService.fullSync(false);
-    this.logService.info(`[LockComponent] Sync took ${new Date().getTime() - startSync}ms`);
+    if (this.platformUtilsService.getClientType() == ClientType.Web) {
+      // Web does not cache vault data and would be in a unusable state when unlocked.
+      await this.syncService.fullSync(true);
+    } else {
+      // On non-web clients, we start a sync in the background, but to not block by it
+      void this.syncService.fullSync(false);
```
quexten, line 659: This is fine since essentially, triggering a background sync is also what happens on a push notification / on manual sync.

quexten:
@gbubemismith I added vault here as a co-reviewer. Is there any concerns with data not being up-to-date? (I'm asking since vault is the main first thing that a user sees after unlock).

I've not noticed any side effects, and I don't know why making the sync non-blocking would be an issue except on web since on web we don't cache ciphers.

libs/key-management-ui/src/lock/components/lock.component.ts
```diff
@@ -612,11 +612,13 @@ export class LockComponent implements OnInit, OnDestroy {
...
     }
 
-    // Vault can be de-synced since server notifications get ignored while locked. Need to check whether sync is required using the sync service.
-    const startSync = new Date().getTime();
-    // TODO: This should probably not be blocking
-    await this.syncService.fullSync(false);
-    this.logService.info(`[LockComponent] Sync took ${new Date().getTime() - startSync}ms`);
+    if (this.platformUtilsService.getClientType() == ClientType.Web) {
+      // Web does not cache vault data and would be in a unusable state when unlocked.
+      await this.syncService.fullSync(true);
```
Thomas-Avery, line 654:
```suggestion
    if (this.platformUtilsService.getClientType() == ClientType.Web) {
      const startSync = performance.now();
      
      // Web does not cache vault data and would be in a unusable state when unlocked.
      await this.syncService.fullSync(true);
      
      this.logService.measure(startSync, "KeyManagement", "LockComponent", "sync complete");
```

:pick: can we do the performance logging for the blocking method?

gbubemismith:
> @gbubemismith I added vault here as a co-reviewer. Is there any concerns with data not being up-to-date? (I'm asking since vault is the main first thing that a user sees after unlock).
> 
> I've not noticed any side effects, and I don't know why making the sync non-blocking would be an issue except on web since on web we don't cache ciphers.

@quexten what do you mean not being up-to-date? if I create an item on a different client and then login to the desktop client, will my desktop client be out of sync?

quexten:
@gbubemismith The assumption here is that:
- The app is locked
- Something gets changed on another app of the same user, f.e a cipher gets changed
- The app is unlocked

Currently, it only proceeds from being unlocked once the sync is done and the client is guaranteed up-to-date, at least assuming the device is online. This change would mean that you could unlock and decrypt your existing, local copy of your ciphers, and then the sync would kick of in the background and update the ciphers once done (f.e a few seconds later, after the client is already unlocked).

I guess I'm asking if vault is aware of any obvious side effects this would have.

gbubemismith:
> @gbubemismith The assumption here is that:
> 
> * The app is locked
> * Something gets changed on another app of the same user, f.e a cipher gets changed
> * The app is unlocked
> 
> Currently, it only proceeds from being unlocked once the sync is done and the client is guaranteed up-to-date, at least assuming the device is online. This change would mean that you could unlock and decrypt your existing, local copy of your ciphers, and then the sync would kick of in the background and update the ciphers once done (f.e a few seconds later, after the client is already unlocked).
> 
> I guess I'm asking if vault is aware of any obvious side effects this would have.

Oh okay, yeah this sounds reasonable to me

PR #19360, opened 2026-03-04
[PM-30584] Move key-connector migration to sdk

quexten:
## 🎟️ Tracking

https://github.com/bitwarden/server/pull/7136
https://github.com/bitwarden/sdk-internal/pull/809
https://github.com/bitwarden/clients/pull/19360

## 📔 Objective

We do not want to keep the same master-key that was used during password derivation, but use a separately sampled key - named "key-connector-key". This means that the conversion flow now requires:
- Posting the key to key-connector
- Setting the "key" field on the user to "key-connector-key-wrapped-user-key"

To keep backwards compatibility, we make the request body optional, but if present, the request body contains the key-connector-key-wrapped-user-key. This is subsequently set to the user object.

This will unblock setting the master-key to state during unlock and login, which will improve unlock time, since we can remove double-kdf-derivation.

### DIFF
--- a/libs/common/src/enums/feature-flag.enum.ts
+++ b/libs/common/src/enums/feature-flag.enum.ts
@@ -49,7 +49,7 @@ export enum FeatureFlag {
   PM27279_V2RegistrationTdeJit = "pm-27279-v2-registration-tde-jit",
   EnableAccountEncryptionV2KeyConnectorRegistration = "enable-account-encryption-v2-key-connector-registration",
   EnableAccountEncryptionV2JitPasswordRegistration = "enable-account-encryption-v2-jit-password-registration",
-  SdkKeyConnectorMigration = "sdk-key-connector-migration",
+  SdkKeyConnectorMigration = "use-sdk-for-key-connector-migration",
   UnlockViaSDK = "unlock-via-sdk",
 
   /* Tools */


libs/common/src/key-management/key-connector/services/key-connector.service.ts
```diff
@@ -131,18 +131,41 @@ export class KeyConnectorService implements KeyConnectorServiceAbstraction {
   }
 
   async migrateUser(keyConnectorUrl: string, userId: UserId) {
-    const masterKey = await firstValueFrom(this.masterPasswordService.masterKey$(userId));
-    const keyConnectorRequest = new KeyConnectorUserKeyRequest(
-      Utils.fromBufferToB64(masterKey.inner().encryptionKey),
+    const sdkKeyConnectorMigration = await firstValueFrom(
+      this.configService.getFeatureFlag$(FeatureFlag.SdkKeyConnectorMigration),
     );
+    if (sdkKeyConnectorMigration) {
```
mzieniukbw, line 137: 🤔 No test coverage

quexten: Hmm...

I'm not sure what useful test coverage looks like here? Just testing the implementation detail that the sdk service is called? It feels like that is not very useful / unit tests feel more useful if there is any kind of complexity in the logic / interesting boundary conditions (null values).

I guess, that's a general thing I've been wondering with code that is just a basic sdk wrapper where all it does is pass through a call to the SDK -- I wonder what kind of bugs we are anticipating to catch here.

(IMO, wrappers such as this one would be better covered by an e2e test of the entire flow, but we don't really have e2e tests readily accessible currently... I'll try to bring this up on the next dev sync).

PR #19367, opened 2026-03-04
[PM-30584] Add unlock for key connector with SDK

quexten:
## 🎟️ Tracking

## 📔 Objective

Adds unlock via the unlock service in the login strategies. The unlock service implements key-connector-unlock via the SDK.


ike-kottlowski:
https://github.com/bitwarden/clients/pull/19422 has been merged.

libs/auth/src/common/login-strategies/user-api-login.strategy.ts
```diff
@@ -63,11 +66,14 @@ export class UserApiLoginStrategy extends LoginStrategy {
     response: IdentityTokenResponse,
     userId: UserId,
   ): Promise<void> {
-    if (response.key) {
+    if (response.canUnlockWithKeyConnector() && await this.configService.getFeatureFlag(FeatureFlag.UnlockKeyConnectorWithSdk)) {
+      await this.masterPasswordService.setMasterKeyEncryptedUserKey(response.key, userId);
```
mzieniukbw: ❓ This is needed to retrieve the user key later on, but i thought the user key is going to be set into client's state by SDK, which makes this redundant ?

libs/unlock/src/default-unlock.service.ts
```diff
@@ -19,6 +19,7 @@ import { MasterKey } from "@bitwarden/common/types/key";
 import { BiometricsService, KdfConfig, KdfConfigService } from "@bitwarden/key-management";
 import { LogService } from "@bitwarden/logging";
 import {
+EncString,
```
mzieniukbw: 🎨  formatting
```suggestion
  EncString,
```

libs/unlock/src/default-unlock.service.ts
```diff
@@ -141,6 +153,34 @@ export class DefaultUnlockService implements UnlockService {
     this.logService.measure(startTime, "Unlock", "DefaultUnlockService", "unlockWithBiometrics");
   }
 
+  async unlockWithKeyConnector(userId: UserId, keyConnectorUnlockData: KeyConnectorUnlockData): Promise<void> {
+    // Now that we have the decrypted user key from the key connector, we can initialize the SDK with it to complete the unlock process.
```
mzieniukbw: 🤔 This sentence is somehow confusing. It speaks about present, but we do not have decrypted user key yet (only when `initialize_user_crypto` finishes). Also it suggests the "user key from the key connector", but that's not true either. I guess it depends on the context, whether we speak about key connector server or the unlock method, i would say too vague.

mzieniukbw (changes requested):

libs/unlock/src/default-unlock.service.ts
quexten: Seems to be outdated from when this used the old key-connector unlock method. Oops. Will replace it.

libs/auth/src/common/login-strategies/user-api-login.strategy.ts
quexten: Hmm, makes sense. I think we should probably track the work to remove masterkeyencrypteduserkey entirely, since KC does not need it in state and since master-password now uses the unlockdata.

mzieniukbw (dismissed):

mzieniukbw (dismissed):

mzieniukbw (dismissed):

libs/auth/src/common/login-strategies/sso-login.strategy.ts
```diff
@@ -136,7 +137,9 @@ export class SsoLoginStrategy extends LoginStrategy {
         );
       } else {
         const keyConnectorUrl = this.getKeyConnectorUrl(tokenResponse);
-        await this.keyConnectorService.setMasterKeyFromUrl(keyConnectorUrl, userId);
+        if (!(await this.configService.getFeatureFlag(FeatureFlag.UnlockKeyConnectorWithSdk))) {
```
JaredSnider-Bitwarden: ❓ , do you think it is worth taking a similar approach to [the unlock service PR for the password login strategy](https://github.com/bitwarden/clients/pull/19371/changes#diff-5fdcd81a147378433eda47db1b87b69f68a33254cb37c2f4974cc04d46edf8c5R45) in that we save this flag lookup in the data to keep it from changing in case of 2FA scenarios?

JaredSnider-Bitwarden: Just started review - will finish tomorrow

libs/auth/src/common/login-strategies/user-api-login.strategy.ts
JaredSnider-Bitwarden, line 66: ⚠️ : changing this implementation to use `response.canUnlockWithKeyConnector()` introduces an unflagged logic change where the condition goes from just a `response.apiUseKeyConnector` check to a `this.apiUseKeyConnector && this.accountKeysResponseModel != null` check.  

Something like:
```
protected override async setMasterKey(response: IdentityTokenResponse, userId: UserId) {
  const sdkHandled =
    response.canUnlockWithKeyConnector() &&
    (await this.configService.getFeatureFlag(FeatureFlag.UnlockKeyConnectorWithSdk));

  if (!sdkHandled && response.apiUseKeyConnector) {
    const env = await firstValueFrom(this.environmentService.environment$);
    const keyConnectorUrl = env.getKeyConnectorUrl();
    await this.keyConnectorService.setMasterKeyFromUrl(keyConnectorUrl, userId);
  }
}
```
Would restore the old code without applying new logic. 

This comment also applies to the `setUserKey` method, I think.

quexten: Okay. Updated.

libs/auth/src/common/login-strategies/sso-login.strategy.ts
quexten: Moved to the cache.

libs/auth/src/common/login-strategies/user-api-login.strategy.ts
```diff
@@ -76,6 +89,8 @@ export class UserApiLoginStrategy extends LoginStrategy {
         );
         await this.keyService.setUserKey(userKey, userId);
       }
+    } else {
+      await this.masterPasswordService.setMasterKeyEncryptedUserKey(response.key, userId);
```
JaredSnider-Bitwarden: ⚠️ : moving this here means we have unflagged logic changes. 

Could we please make this:
```
protected override async setUserKey(response: IdentityTokenResponse, userId: UserId): Promise<void> {
  const sdkHandledKeyConnector =
    response.canUnlockWithKeyConnector() &&
    (await this.configService.getFeatureFlag(FeatureFlag.UnlockKeyConnectorWithSdk));

  if (sdkHandledKeyConnector) {
    await this.unlockService.unlockWithKeyConnector(userId, response.intoKeyConnectorUnlockData());
    return;
  }

 // old code:
  if (response.key) {
    await this.masterPasswordService.setMasterKeyEncryptedUserKey(response.key, userId);
  }

  if (response.apiUseKeyConnector) {
    const masterKey = await firstValueFrom(this.masterPasswordService.masterKey$(userId));
    if (masterKey) {
      const userKey = await this.masterPasswordService.decryptUserKeyWithMasterKey(masterKey, userId);
      await this.keyService.setUserKey(userKey, userId);
    }
  }
}

File: libs/common/src/auth/models/response/identity-token.response.ts [hunk 6]
```diff
@@ -100,4 +101,19 @@ export class IdentityTokenResponse extends BaseResponse {
   hasMasterKeyEncryptedUserKey(): boolean {
     return Boolean(this.key);
   }
+
+  canUnlockWithKeyConnector(): boolean {
+    return this.apiUseKeyConnector && this.accountKeysResponseModel != null;
```
JaredSnider-Bitwarden: ⚠️ :  `apiUserKeyConnector` is only set on the server for client_credential or API key logins ([server code](https://github.com/bitwarden/server/blob/main/src/Identity/IdentityServer/RequestValidators/CustomTokenRequestValidator.cs#L150-L157)).  However, key connector users also use the SSO flow. I think this should be predicated upon the existence of the `this.userDecryptionOptions.keyConnectorOption` and the `this.key`: 

Something like:
```
canUnlockWithKeyConnector(): boolean {
  return (
    this.userDecryptionOptions?.keyConnectorOption != null &&  // ensures key connector member decryption type
    this.key != null &&   // ensures keyConnectorKeyWrappedUserKey is present
    this.accountKeysResponseModel != null  // ensures new account crypto state is present
  );
}

JaredSnider-Bitwarden (changes requested):

libs/auth/src/common/login-strategies/user-api-login.strategy.ts
quexten: Applied

libs/common/src/auth/models/response/identity-token.response.ts
quexten: TBH, the key connector option is present, but the key is null is a scenario that should never occur. This should just be key connector option being present.

Also, as a note for KM: We should move "key" to "keyConnectorKeyWrappedUserKey" in the key connector option.

JaredSnider-Bitwarden: Ok, if `this.key` is required and is a scenario that should never occur, I do think the code should reflect that, but I won't hold this up on that since you said it is an unrealistic scenario. 

> Also, as a note for KM: We should move "key" to "keyConnectorKeyWrappedUserKey" in the key connector option.

Good call!

JaredSnider-Bitwarden (approved):
Thank you for iterating on this! This looks great! 
Not sure what's going on with CI rust lint errors, but approving.

mzieniukbw (approved):

quexten:
Rust errors seem unrelated.

PR #19371, opened 2026-03-04
[PM-33173] Use unlock service for password login strategy

quexten:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-33173

## 📔 Objective

Migrates to use the unlock service instead of manually setting keys within auth code. This allows us to drop dependence on the master-key in a follow-up PR. This work is feature-flagged.

## 📸 Screenshots

https://github.com/user-attachments/assets/ed8ae9a0-dffb-4d5b-aafa-cf18b4e4e8bb


libs/auth/src/common/login-strategies/password-login.strategy.spec.ts
```diff
@@ -222,6 +229,27 @@ describe("PasswordLoginStrategy", () => {
     );
   });
 
+  it("uses master password unlock service when feature flag is enabled", async () => {
+    const userKey = new SymmetricCryptoKey(
+      new Uint8Array(64).buffer as unknown as CsprngArray,
+    ) as UserKey;
+
+    configService.getFeatureFlag.mockResolvedValue(true);
```
JaredSnider-Bitwarden: ⚠️ : Not specifying the flag can result in unexpected behavior - we just had feature flagged, in flight work that would be turned on for your test with this. 

```
    configService.getFeatureFlag
      .calledWith(FeatureFlag.UseUnlockServiceForPasswordLogin)
      .mockResolvedValue(true);
```

libs/auth/src/common/login-strategies/password-login.strategy.ts
```diff
@@ -76,6 +84,13 @@ export class PasswordLoginStrategy extends LoginStrategy {
       map((state) => state.tokenRequest.masterPasswordHash),
     );
     this.localMasterKeyHash$ = this.cache.pipe(map((state) => state.localMasterKeyHash));
+
+    // Set the feature flag state; this will run in the background since the constructor itself cannot be async
+    void (async () => {
+      this.unlockServiceForPasswordLogin = await this.configService.getFeatureFlag(
+        FeatureFlag.UseUnlockServiceForPasswordLogin,
+      );
+    })();
```
JaredSnider-Bitwarden: ❓ : can we move this into the `logIn` method and await its result because with this as written, there is no contract that requires this to have been completed before a `logIn` can start - even though our `configService` should always return this data before a user triggers a logIn. I consider an explicit contract of having it awaited in the login method to be safer.

JaredSnider-Bitwarden (changes requested):
This is looking great. I love the simplified API we are going to end up with. Login and then just leverage y'all's unlock service. 🥇 . I have two small items down 👇

libs/auth/src/common/login-strategies/password-login.strategy.ts
quexten: Okay, I was a bit worried about the setMasterKey usage, but it seems this should be fine since it's only called after login is finished.

libs/auth/src/common/login-strategies/password-login.strategy.spec.ts
quexten: @JaredSnider-Bitwarden your suggested implementation didn't work. `calledwith` just did not work as expected here. Maybe due to getFeatureFlag being async? I'm not sure.

```
    configService.getFeatureFlag.mockImplementation(async (flag: FeatureFlag) => {
      if (flag === FeatureFlag.UseUnlockServiceForPasswordLogin) {
        return true;
      }
      return false;
    });
```

i went with this now.

libs/auth/src/common/login-strategies/login.strategy.ts
```diff
@@ -262,8 +262,8 @@ export abstract class LoginStrategy {
     }
 
     await this.setMasterKey(response, userId);
-    await this.setUserKey(response, userId);
     await this.setAccountCryptographicState(response, userId);
+    await this.setUserKey(response, userId);
```
JaredSnider-Bitwarden, line 266: ⚠️ : This is an unflagged reordering, but after some investigation, I don't think it should introduce any issues.

libs/auth/src/common/login-strategies/password-login.strategy.ts
```diff
@@ -61,11 +63,17 @@ export class PasswordLoginStrategy extends LoginStrategy {
 
   protected cache: BehaviorSubject<PasswordLoginStrategyData>;
 
+  // Cache the feature flag state in the service, so that every time the service is accessed, the value is the same.
+  // Changing the feature flag state during runtime could lead to unintended behavior, since there are multiple times
+  // where this value is checked.
+  unlockServiceForPasswordLogin = false;
```
JaredSnider-Bitwarden: ⚠️ : I remembered that we cannot cache these types of values in memory on this service b/c of extension foreground context scenarios. We must add this to the `PasswordLoginStrategyData` so that it persists across foreground context loss.  

Here's an email 2FA example scenario where I log the value of the flag on login start, but since the email 2FA workflow requires you to close the popup,  the value of the flag is reset. 

https://github.com/user-attachments/assets/3055aa9e-478c-4734-8654-c146a147810c

JaredSnider-Bitwarden (changes requested):
Thank you for your patience. I saw one final issue that needs to be addressed before approval. Everything else is good!

libs/auth/src/common/login-strategies/password-login.strategy.ts
quexten: Moved.

JaredSnider-Bitwarden (approved):
LGTM as long as CI passes!

quexten:
@JaredSnider-Bitwarden Since this is entirely feature flagged, i'll merge it and we'll do QA on main. Please let me know if there are concerns with this and i'll revert it back out before `rc` cut.

PR #19372, opened 2026-03-04
[PM-33174] Remove master key from auth request login

quexten:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-33174

## 📔 Objective

We removed support for master-key-auth-request flows a long time ago. This removes dependence on the master-key in the login strategy.


libs/auth/src/common/login-strategies/auth-request-login.strategy.ts
```diff
@@ -80,31 +80,10 @@ export class AuthRequestLoginStrategy extends LoginStrategy {
...
-      await this.masterPasswordService.setMasterKeyEncryptedUserKey(response.key, userId);
-    }
-
-    if (authRequestCredentials.decryptedUserKey) {
-      await this.keyService.setUserKey(authRequestCredentials.decryptedUserKey, userId);
-    } else {
-      await this.trySetUserKeyWithMasterKey(userId);
-
-      // Establish trust if required after setting user key
-      await this.deviceTrustService.trustDeviceIfRequired(userId);
```
quexten, line 95: This seems like a bug was present before the change. Why would we only trust the device if we use the master-key unlock flow?

ike-kottlowski: I did some sleuthing and it's a little unclear to me why we would do that. The origin of this code is from the TDE flow that was added 3 years ago when this file was called `passwordless-login.strategy.ts`. @trmartin4 was the author of the PR that merged the changes in he may have some insight.

After interrogating Claude, the thought was that depending on from whence the request came could dictate assumptions about the current device. 
1. If the `UserKey` comes to us decrypted, then the current device already has device keys ergo already trusted?
2. If we must decrypt the `UserKey` then this device is new and must be trusted for future attempts. This one feels like a TDE flow where an unknown device is approved by an Admin possibly.

I'll let Todd add any information but that's my understanding.

ike-kottlowski (commented):
Is it possible to get a screen recording showing unchanged functionality?

libs/auth/src/common/login-strategies/auth-request-login.strategy.ts
trmartin4: If this is https://github.com/bitwarden/clients/commit/5665576147a4fbf30024f56358d424ff26a1a131 that you are referring to, that is the giant TDE feature branch merge, which included the entire TDE feature.  So I suspect that this was a bug introduced during TDE in an individual commit to that feature branch that got associated with the resulting merge commit.

quexten: > After interrogating Claude, the thought was that depending on from whence the request came could dictate assumptions about the current device. 
> 1. If the UserKey comes to us decrypted, then the current device already has device keys ergo already trusted?
> 2. If we must decrypt the UserKey then this device is new and must be trusted for future attempts. This one feels like a TDE flow where an unknown device is approved by an Admin possibly.

⚠️ Claude is very good at making up justifications for things that have no underlying reasoning. I believe what it suggested is just noise?

At any rate, a completely newly logged in device would not be trusted. It would just be logged in and never have had a user-key which it needs in order to trust. If you log in via auth request, that is login-with-device or an admin approval, for an admin approval you would always get a user-key, whereas for login-with-device you would sometimes get a user-key, and sometimes get a master-key, depending both on the client version of the approving device, and on the unlock method that was used to unlock the approving device.

Because of this, this means that depending on the client version, unlock method, and approval method, the device would either be trusted or not trusted. Which suggests, this is a bug.

quexten:
@ike-kottlowski 

https://github.com/user-attachments/assets/3a8d7b11-345a-4e24-b050-9e1aa323604b

ike-kottlowski (approved):

libs/auth/src/common/login-strategies/auth-request-login.strategy.ts
ike-kottlowski: I agree Claude created noise, it wasn't helpful.

I'll brush up on Key Management so I'm better able to detect this type of noise. Thank you.

PR #19375, opened 2026-03-04
[PM-33121][VULN-394] Use fromChromeEvent + onCommitted for earlier phishing interception

AlexRubik:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-33121

## 📔 Objective

Alternative and preferred fix to [PM-32047](https://bitwarden.atlassian.net/browse/PM-32047) ([PR #19118](https://github.com/bitwarden/clients/pull/19118)).

### What

Replaces `chrome.tabs.onUpdated` with `fromChromeEvent(chrome.webNavigation.onCommitted)` for phishing URL checks.

### Why

- **Earlier interception** — `onCommitted` fires after the server responds but before page content renders, earlier than `tabs.onUpdated "complete"` which waits for the full page load.
- **Platform team pattern** — Uses `fromChromeEvent()` for Chrome event → RxJS Observable conversion with automatic listener cleanup, matching the established pattern in `badge-browser-api.ts`, `account-switcher.service.ts`, etc.
- **Simpler code** — Eliminates the static `_tabUpdated$` Subject, `_handleTabUpdated` handler, `PhishingDetectionNavigationEvent` type, and manual `BrowserApi.removeListener` cleanup (~30 lines removed).

### Why `onCommitted` over `onBeforeNavigate`

`onBeforeNavigate` fires before DNS resolution — the earliest interception point — but `tabs.update()` called during `onBeforeNavigate` races with the in-progress navigation. The original page load can overwrite the redirect, especially when DNS/resources are cached (e.g., after "Continue to this site"). `onCommitted` fires after navigation commits, so `tabs.update()` reliably redirects. See commit `e7d2b82` for full details.

### Changes

| File | Change |
|------|--------|
| `phishing-detection.service.ts` | Replace `BrowserApi.addListener(chrome.tabs.onUpdated)` + Subject with `fromChromeEvent(chrome.webNavigation.onCommitted)`. Filter to `frameId === 0` (main frame only). Remove `ignored` from `distinctUntilChanged` to prevent duplicate events from breaking the "Continue to this site" flow. Fix one-time ignore leak by deleting hostname in `onContinueCommand$`. |
| `phishing-detection.service.spec.ts` | Rewrite tests to mock `fromChromeEvent` module. Test iframe filtering, extension page filtering, and dispose cleanup. |
| `phishing-data.service.ts` | Add `https://bitwarden.github.io/phishing-test-page/inf-load/` to test web addresses (org-hosted infinite-loading page for QA). |
| `phishing-data.service.spec.ts` | Add test for the new QA test address. |

### QA fix: "Continue to this site" re-blocking

Addressed the issue [reported by Olena](https://bitwarden.atlassian.net/browse/PM-33121?focusedCommentId=139318) where clicking "Continue to this site" immediately re-blocked the page. Unlike `tabs.onUpdated "complete"` which fires once per navigation, `onBeforeNavigate` can fire multiple times. The first event would delete the hostname from the ignore list, then a duplicate event would see it as no longer ignored and re-block. Fixed by removing `ignored` from the `distinctUntilChanged` comparison — duplicate events for the same url+tabId are now suppressed regardless of the ignore flag changing between fires.

### Browser compatibility

Works on Chrome, Firefox, Edge, and Opera. Safari is already excluded from phishing detection. The `webNavigation` permission is already declared in the manifest.

### Non-blocking behavior

The check is purely observational — the browser continues loading the page normally while the async IndexedDB lookup runs in parallel. If phishing is detected, `navigateTabToUrl` replaces the tab with the warning page.
---
### `onErrorOccurred` for HTTP error pages (`9ca513b`)

Chrome skips `onCommitted` for HTTP errors/network failures, firing `onErrorOccurred` instead. Added `onErrorOccurred` alongside `onCommitted` so error-page phishing URLs are caught in Chrome too.

### Per-tab "Continue to this site" tracking (`c1af5ff`)

Changed ignore tracking from global hostname to `hostname:tabId` pairs. The previous approach had a race condition in multi-tab scenarios where the hostname was deleted before all navigation events finished firing. Per-tab entries persist until extension restart. Also filters `about:` URLs from the pipeline.

[PM-32047]: https://bitwarden.atlassian.net/browse/PM-32047?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
---


ttalty (dismissed):

apps/browser/src/dirt/phishing-detection/services/phishing-data.service.ts
```diff
@@ -251,6 +251,7 @@ export class PhishingDataService {
       new URL("http://phishing.testcategory.com").href,
       new URL("https://phishing.testcategory.com").href,
       new URL("https://phishing.testcategory.com/block").href,
+      new URL("https://alexrubik.github.io/phishing-test-page/").href, // infinite loading page
```
ttalty: ❗ Let's change this to something Bitwarden hosted. 

Details to follow after identifying the process.

ttalty (changes requested):
❗ The developer hosted phishing testing link should be Bitwarden hosted.

apps/browser/src/dirt/phishing-detection/services/phishing-data.service.ts
AlexRubik: https://github.com/bitwarden/clients/pull/19375/commits/ecf71f7c51bca1ffb289575b16ea8a29c93f2081

https://github.com/bitwarden/phishing-test-page

https://bitwarden.github.io/phishing-test-page/inf-load/

apps/browser/src/dirt/phishing-detection/services/phishing-detection.service.ts
```diff
@@ -60,22 +52,18 @@ export class PhishingDetectionService {
...
-        (navEvent) =>
-          navEvent.changeInfo.status === "complete" &&
-          !!navEvent.tab.url &&
-          !this._isExtensionPage(navEvent.tab.url),
-      ),
-      map(({ tab, tabId }) => {
-        const url = new URL(tab.url!);
-        return { tabId, url, ignored: this._ignoredHostnames.has(url.hostname) };
+    // Intercept at the earliest navigation event, before DNS resolution begins
+    const onTabUpdated$ = fromChromeEvent(chrome.webNavigation.onBeforeNavigate).pipe(
```
Banrion: ⚠️ We previously used the `chrome.webNavigation.onBeforeNavigate` function. It was replaced with the `chrome.tabs.onUpdated` usage due to desire to support Safari. If we go back to using the navigation api, this may make it difficult to support Safari in the future. Can we confirm if Safari is planned to be supported in the near future? If so, we may need to reassess how to handle this situation.

ttalty (dismissed):

Banrion (dismissed):

Banrion (dismissed):

Banrion (dismissed):

Banrion (dismissed):

AlexRubik:
Reset branch onto current main and rewrote commits to resolve conflicts with #20079 (constructor refactor) and #19515 (delta sync). Same logical changes, clean history.

Here's the pre-reset backup for reference: https://github.com/bitwarden/clients/pull/20360

Banrion (approved):

apps/browser/src/dirt/phishing-detection/services/phishing-detection.service.spec.ts
```diff
@@ -8,277 +8,153 @@ import { PhishingDetectionSettingsServiceAbstraction } from "@bitwarden/common/d
...
+    emitErrorEvent(1, "http://akonaa.fr/", "net::ERR_HTTP_RESPONSE_CODE_FAILURE", 1);
 
-    it("closes the tab with the given tabId", async () => {
-      sendMessage(PHISHING_DETECTION_CANCEL_COMMAND.command, { tabId: 42 });
-      await flushPromises(); // Wait for all async operations to complete
-      expect(BrowserApi.closeTab).toHaveBeenCalledWith(42);
-    });
+    expect(phishingDataService.isPhishingWebAddress).not.toHaveBeenCalled();
   });
 });
```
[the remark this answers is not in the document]
AlexRubik: @claude https://github.com/bitwarden/clients/pull/19375/commits/2e484089ae13d19fba509652112ca66da8f27cd8

Banrion (approved):

PR #19392, opened 2026-03-05
[PM-29232] Add cookie acquisition

djsmith85:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-29232

## 📔 Objective

Adds a way for the SDK to acquire a cookie via the proxy-cookie-redirect-connector if needed.

On request routes the user to the `proxy-cookie-redirect-connector` (launches browser) and the connector will redirect to the `bw-server/api/sso-cookie-vendor` and the endpoint will prompt to authenticate against the configured IdP and on success will provide the cookie back to the desktop application via the `bitwarden://`schema.


libs/common/src/platform/services/server-communication-config/default-server-communication-config.service.ts
```diff
@@ -61,4 +61,13 @@ export class DefaultServerCommunicationConfigService implements ServerCommunicat
   async getCookies(hostname: string): Promise<Array<[string, string]>> {
     return this.client.cookies(hostname);
   }
+
+  async acquireCookie(hostname: string): Promise<void> {
```
coroiu: suggestion: can you rename hostname here to "url"?
```suggestion
  async acquireCookie(url: string): Promise<void> {
```

djsmith85: Addressed with b73ac6992587bb8884c22bf92396c0776f33e8a5

PR #19395, opened 2026-03-05
[PM-33198] Keeper direct importer

detunized:
## 🎟️ Tracking

N/A

## 📔 Objective

Add a direct importer for Keeper password manager, similar to the existing LastPass direct importer. This allows users to import their vault data directly from Keeper by providing their credentials, without needing to manually export a file first.

Key components:
- **Access module** (`libs/importer/src/importers/keeper/access/`): handles Keeper API authentication (including device approval, 2FA/TOTP/SMS, region redirect, WebSocket push notifications) and vault sync/decryption
- **Importer**: converts decrypted Keeper vault records (logins, cards, identities, secure notes, SSH keys, etc.) into Bitwarden's `ImportResult` format with folder structure preserved
- **UI dialogs**: device approval method selection and code entry, 2FA selection and code entry

## 📸 Screenshots

image
image
image
image
image

image


libs/importer/src/importers/keeper/access/services/crypto.ts
```diff
@@ -0,0 +1,343 @@
...
+  const arr = new Uint8Array(length);
+  crypto.getRandomValues(arr);
+  return arr;
+}
+
+export function generateEncryptionKey(): Uint8Array {
+  return getRandomBytes(32);
+}
+
+export async function decryptAesV1(data: Uint8Array, key: Uint8Array): Promise<Uint8Array> {
```
quexten:
```suggestion
/**
 * Decrypt an aes-v1 packet. Keeper aes-v1 is AES-CBC without any authentication (MAC).
 */
export async function decryptAesV1(data: Uint8Array, key: Uint8Array): Promise<Uint8Array> {
```

libs/importer/src/importers/keeper/access/services/crypto.ts
```diff
@@ -0,0 +1,343 @@
...
+  const iv = data.subarray(0, AES_BLOCK_SIZE);
+  const encrypted = data.subarray(AES_BLOCK_SIZE);
+  const cryptoKey = await crypto.subtle.importKey("raw", key, { name: "AES-CBC" }, false, [
+    "decrypt",
+  ]);
+  const decrypted = await crypto.subtle.decrypt({ name: "AES-CBC", iv }, cryptoKey, encrypted);
+  return new Uint8Array(decrypted);
+}
+
+export async function encryptAesV2(
```
quexten, line 54:
```suggestion
/**
 * Encrypt a packet with aes-v2. Keeper aes-v2 is AES-GCM. Optionally accepts an external nonce, if not provided the nonce will be generated automatically. WARNING: DO NOT REUSE THE NONCE. THIS WILL LEAD TO CATASTROPHIC CRYPTOGRAPHIC FAILURE.
 */
export async function encryptAesV2(
```

libs/importer/src/importers/keeper/access/services/crypto.ts
```diff
@@ -0,0 +1,343 @@
...
+  ]);
+  const encrypted = await crypto.subtle.encrypt(
+    { name: "AES-GCM", iv: nonceBuffer, tagLength: AES_GCM_TAG_SIZE * 8 },
+    cryptoKey,
+    data,
+  );
+  return concatUint8Arrays(nonceBuffer, new Uint8Array(encrypted));
+}
+
+export async function decryptAesV2(
```
quexten, line 80:
```suggestion
/**
 * Decrypt a packet with aes-v2. Keeper aes-v2 is AES-GCM. The packet has the format: Nonce|Plaintext|Tag where Plaintext|Tag are the output of AES-GCM.
 */
export async function decryptAesV2(
```

libs/importer/src/importers/keeper/access/services/crypto.ts
```diff
@@ -0,0 +1,343 @@
...
+
+export async function encryptEc(data: Uint8Array, publicKey: CryptoKey): Promise<Uint8Array> {
+  const { privateKey: ephemeralPrivate, publicKey: ephemeralPublic } = await generateEcKey();
+  const encryptionKey = await deriveEcdhKey(publicKey, ephemeralPrivate);
+  const encryptedData = await encryptAesV2(data, encryptionKey);
+  const ephemeralPublicBytes = await unloadEcPublicKey(ephemeralPublic);
+  return concatUint8Arrays(ephemeralPublicBytes, encryptedData);
+}
+
+export async function decryptEc(data: Uint8Array, privateKey: CryptoKey): Promise<Uint8Array> {
```
quexten, line 185:
```suggestion
/**
 * Decrypts an ECC-encrypted-packet. Keeper ECC encryption works by deriving an ephemeral content-encryption-key, then encrypting the payload using AES-GCM. The packet concatinates the ECC ephemeral public key, and the AES-GCM encrypted packet.
 */
export async function decryptEc(data: Uint8Array, privateKey: CryptoKey): Promise<Uint8Array> {
```

libs/importer/src/importers/keeper/access/services/crypto.ts
```diff
@@ -0,0 +1,343 @@
...
+  const sharedSecret = await crypto.subtle.deriveBits(
+    { name: "ECDH", public: publicKey },
+    privateKey,
+    256,
+  );
+  const hash = await crypto.subtle.digest("SHA-256", sharedSecret);
+  return new Uint8Array(hash);
+}
+
+export async function encryptEc(data: Uint8Array, publicKey: CryptoKey): Promise<Uint8Array> {
```
quexten, line 174:
```suggestion
/**
 * Encrypts an ECC-encrypted packet. Keeper ECC encryption works by deriving a shared key content-encrytpion-key using ECDH, then encrypting the content with under the content-encryption-key with AES-GCM
 */
export async function encryptEc(data: Uint8Array, publicKey: CryptoKey): Promise<Uint8Array> {
```

libs/importer/src/importers/keeper/access/services/crypto.ts
```diff
@@ -0,0 +1,343 @@
...
+
+export async function decryptEc(data: Uint8Array, privateKey: CryptoKey): Promise<Uint8Array> {
+  const ephemeralPublicBytes = data.subarray(0, 65);
+  const ephemeralPublic = await loadEcPublicKey(ephemeralPublicBytes);
+  const encryptionKey = await deriveEcdhKey(ephemeralPublic, privateKey);
+  const encryptedData = data.subarray(65);
+  return await decryptAesV2(encryptedData, encryptionKey);
+}
+
+export async function deriveKeyV1(
```
quexten, line 196:
```suggestion
/**
 * Derives a keeper master-key using a password, salt and a number of iterations.
 */
export async function deriveKeyV1(
```

libs/importer/src/importers/keeper/access/services/crypto.ts
```diff
@@ -0,0 +1,343 @@
...
+  password: string,
+  salt: Uint8Array,
+  iterations: number,
+): Promise<Uint8Array> {
+  const key = await deriveKeyV1(password, salt, iterations);
+  const hash = await crypto.subtle.digest("SHA-256", key);
+  return new Uint8Array(hash);
+}
+
+export async function decryptEncryptionParams(
```
quexten, line 239:
```suggestion
/**
 * Derives a key from the encryption parameters. The encryption parameters is a binary blob that contains the data needed to derive the key. It is structured as follows.
 * CORRUPTED[1]|Iterations[3]|Salt[16]|IV[16]|Data
 * The iterations, salt, key are used to derive a content-encryption-key. The IV and cek are used to decrypt the data, which consists of two 32-byte blocks. These two blocks need to match. They are decrypted with AES-CBC, using no pkcs7 padding.
 */
export async function decryptEncryptionParams(
```

libs/importer/src/importers/keeper/access/services/crypto.ts
```diff
@@ -0,0 +1,343 @@
...
+
+  const iterations = (encryptionParams[1] << 16) + (encryptionParams[2] << 8) + encryptionParams[3];
+  const salt = encryptionParams.subarray(4, 20);
+  const key = await deriveKeyV1(password, salt, iterations);
+  const iv = encryptionParams.subarray(20, 36);
+
+  // Decrypt data with no padding
+  // We need to manually handle this since Web Crypto always expects PKCS7 padding
+  const encryptedData = encryptionParams.subarray(36);
+  const decrypted = await decryptAesNoPadding(encryptedData, key, iv);
```
quexten:
```suggestion

  // Decrypt data with no padding
  // We need to manually handle this since Web Crypto always expects PKCS7 padding
  const iv = encryptionParams.subarray(20, 36);
  const encryptedData = encryptionParams.subarray(36);
  const decrypted = await decryptAesNoPadding(encryptedData, key, iv);
```

libs/importer/src/importers/keeper/access/services/crypto.ts
```diff
@@ -0,0 +1,343 @@
...
+
+  return first32;
+}
+
+async function decryptAesNoPadding(
+  data: Uint8Array,
+  key: Uint8Array,
+  iv: Uint8Array,
+): Promise<Uint8Array> {
+  // Web Crypto doesn't support no-padding mode, so we need a workaround.
```
quexten, line 277:
```suggestion
  // Web Crypto doesn't support no-padding mode, so we need a workaround.
  // AES-CBC chains individual aes encryption operation blocks using the previous ciphertext as the IV of the next block. In PKCS7, the last block needs to be valid padding. In the case that the full block is padding and contains no data, the entire block should contain 0x10. Because of this, a non-padded AES-CBC block can be converted to a correctly padded block, in the case that the plaintext length is a multiple of the block size. This is done by using the last ciphertext block as the IV to encrypt the padding block containing only 0x10. Appending this at the end, then results in a valid PKCS7-padded block.
```

quexten (dismissed):
I'm suggesting some docs here. It will be incredibly hard otherwise to come back to this later on.

detunized:
I added all the comments, @quexten

quexten:
Thanks @detunized , removing my blocking review.

itsadrago (commented):
## Code Review Summary

**PR: [PM-33198] Keeper direct importer** - Adds a direct import feature for Keeper password manager, allowing users to authenticate and pull vault data without manual file export.

### Overview

This is a large, well-structured community contribution that adds a full Keeper direct import flow: API client with auth/2FA/device approval, vault decryption, record parsing, Angular UI dialogs, and protobuf-based communication. The implementation follows the established pattern from the existing LastPass direct importer (`libs/importer/src/importers/lastpass/access/`). Test coverage is provided for vault decryption and the importer conversion logic.

### Findings

| Severity | Count | Description |
|----------|-------|-------------|
| :warning: IMPORTANT | 1 | Unvalidated server redirect URL could send credentials to arbitrary hosts |
| :art: SUGGESTED | 1 | Recursive folder path builder has no cycle guard |

### Notes

- The `crypto.ts` module (~400 lines) implements Keeper-specific crypto primitives (AES-CBC, AES-GCM, RSA PKCS#1 v1.5, ECDH, PBKDF2). The project CLAUDE.md states "new encryption logic should not be added to this repo," though the existing LastPass direct importer sets precedent for third-party crypto in the importer library. The Bitwarden maintainers should confirm this is acceptable.
- The `@bufbuild/protobuf` dependency is added for Keeper's protobuf-based API. Generated protobuf files are committed (~18k lines). Maintainers should verify the generation process is reproducible and the generated files can be audited.
- Patch test coverage is 22% per Codecov. The core business logic (importer, vault decryption) is tested, but the auth client, crypto, socket, and UI services lack tests.

libs/importer/src/importers/keeper/access/services/client.ts
```diff
@@ -0,0 +1,862 @@
...
+        switch (response.loginState) {
+          case LoginState.REQUIRES_AUTH_HASH:
+            response = await this.handleAuthHash(password, response);
+            break;
+
+          case LoginState.REGION_REDIRECT:
+            if (!response.stateSpecificValue) {
+              throw new Error("Region redirect without server URL");
+            }
+            this.server = response.stateSpecificValue;
```
itsadrago, line 129: :warning: **IMPORTANT: Unvalidated server redirect URL**

`this.server` is set directly from `response.stateSpecificValue` without any validation that it is a legitimate Keeper domain. All subsequent API calls (`executeRest`, `connectPushSocket`) will use this attacker-controllable server value, which means authentication hashes, device tokens, and encrypted payloads would be sent to the redirected host.

While the initial connection uses HTTPS to a hardcoded Keeper domain (from `KeeperRegion`), a compromised or rogue Keeper server could redirect the client to an arbitrary endpoint.

Consider validating that `stateSpecificValue` matches a known Keeper domain pattern (e.g., `*.keepersecurity.*`) before accepting the redirect:

```typescript
case LoginState.REGION_REDIRECT:
  if (!response.stateSpecificValue) {
    throw new Error("Region redirect without server URL");
  }
  if (!this.isValidKeeperDomain(response.stateSpecificValue)) {
    throw new Error("Region redirect to untrusted server");
  }
  this.server = response.stateSpecificValue;
```

libs/importer/src/importers/keeper/access/vault.ts
```diff
@@ -0,0 +1,420 @@
...
+      const name = sanitizeFolderName(folderNames.get(uid) ?? uid);
+      const parentUid = childToParent.get(uid);
+      if (!parentUid) {
+        paths.set(uid, name);
+        return name;
+      }
+      const parentPath = getPath(parentUid);
+      const path = joinPath(parentPath, name);
+      paths.set(uid, path);
+      return path;
```
itsadrago, line 434: :art: **SUGGESTED: Add cycle guard to recursive folder path resolution**

The `getPath` function recurses via `getPath(parentUid)` on line 336 and caches results in `paths` on line 338. However, the cache is written *after* the recursive call returns, so it does not protect against circular parent references. If a Keeper vault response contained a cycle (e.g., folder A's parent is B and B's parent is A), this would cause a stack overflow.

While Keeper's API is unlikely to produce such data, a defensive guard is straightforward:

```typescript
const visiting = new Set<string>();
const getPath = (uid: string): string => {
  if (paths.has(uid)) {
    return paths.get(uid)!;
  }
  if (visiting.has(uid)) {
    // Break the cycle
    const name = sanitizeFolderName(folderNames.get(uid) ?? uid);
    paths.set(uid, name);
    return name;
  }
  visiting.add(uid);
  // ... rest of existing logic
};
```

libs/importer/src/importers/keeper/access/services/client.ts
detunized: I can add a check for this, but I'm not sure if it's necessary. Keep in mind that all the outgoing requests are encrypted with public Keerper keys, which means they could only be decrypted by the holder of the matching Keeper private key. The incoming responses from the server are encrypted by the private keys. And before being used they are being decrypted with the matching public key. This makes any form of forgery and/or hijacking very unlikely. 

https://github.com/Keeper-Security/Commander/blob/9d88f7a6a0c27ed81e040f102c0b856687788de5/keepercommander/loginv3.py#L274

```py
            elif resp.loginState == APIRequest_pb2.REGION_REDIRECT:
                params.server = resp.stateSpecificValue
                logging.info('Redirecting to region: %s', params.server)
                LoginV3API.register_device_in_region(params, encryptedDeviceToken)
                resp = LoginV3API.startLoginMessage(params, encryptedDeviceToken
```

A snippet from the Keeper Commander tool. No checking. Also no checks in C#, Pything or JavaScript SDKs.

libs/importer/src/importers/keeper/access/vault.ts
detunized: Unlikely to happen. Keeper would have to return an invalid folder structure. But I added this check anyway.

apps/browser/src/_locales/en/messages.json
```diff
@@ -4590,6 +4590,132 @@
...
+  },
+  "keeperSelectApprovalMethod": {
+    "message": "Select approval method"
+  },
+  "keeperSelectApprovalMethodDesc": {
+    "message": "Choose how you want to approve this device."
+  },
+  "keeperApprovalMethodEmail": {
+    "message": "Email"
+  },
```
itsadrago, line 5114: We already have a translation for "Email" called "email" across web/browser/desktop

apps/browser/src/_locales/en/messages.json
```diff
@@ -4590,6 +4590,132 @@
...
+  "keeperApprovalMethodPush": {
+    "message": "Keeper Push"
+  },
+  "keeperDeviceApprovalEmailDesc": {
+    "message": "Check your email and click the approval link, or enter the verification code below."
+  },
+  "keeperVerificationCode": {
+    "message": "Verification code"
```
itsadrago: We already have a translation for "Verification code" called "verificationCode" across web/browser/desktop

apps/browser/src/_locales/en/messages.json
```diff
@@ -4590,6 +4590,132 @@
...
+  "keeperVerificationCodeHint": {
+    "message": "Enter the 6-digit code from the email, or click the link and this dialog will close automatically."
+  },
+  "keeperDeviceApprovalPushDesc": {
+    "message": "Approve the device request on your other Keeper device. This dialog will close automatically."
+  },
+  "keeperMFARequired": {
+    "message": "Keeper two-factor authentication required"
+  },
+  "keeperMFADesc": {
```
itsadrago: I see the same message for LastPass in `lastPassMFADesc`. We should create a generic key for this message and have both Keeper and LastPass use it instead across web/browser/desktop
```suggestion
"mfaDesc": {
    "message": "Enter your one-time passcode from your authentication app"
},
```

apps/browser/src/_locales/en/messages.json
```diff
@@ -4590,6 +4590,132 @@
...
+  "keeperSelectTwoFactorMethod": {
+    "message": "Select two-factor method"
+  },
+  "keeperSelectTwoFactorMethodDesc": {
+    "message": "Choose how you want to verify your identity."
+  },
+  "keeperTwoFactorTotp": {
+    "message": "Authenticator app (TOTP)"
+  },
+  "keeperTwoFactorSms": {
```
itsadrago: `keeperTwoFactorSms` and `keeperDuoSms` share the same message across web/browser/desktop. Let's have them both use a single generic message

```suggestion
"textMessageSms": {
      "message": "Text message (SMS)"
  },
```

apps/browser/src/_locales/en/messages.json
```diff
@@ -4590,6 +4590,132 @@
...
+    "message": "Text message (SMS)"
+  },
+  "keeperTwoFactorDuo": {
+    "message": "Duo Security"
+  },
+  "keeperTwoFactorRsa": {
+    "message": "RSA SecurID"
+  },
+  "keeperTwoFactorBackup": {
```
itsadrago:
```suggestion
"backupCodes": {
    "message": "Backup codes"
  },
```

apps/browser/src/_locales/en/messages.json
```diff
@@ -4590,6 +4590,132 @@
...
+  "keeperSelectTwoFactorMethodDesc": {
+    "message": "Choose how you want to verify your identity."
+  },
+  "keeperTwoFactorTotp": {
+    "message": "Authenticator app (TOTP)"
+  },
+  "keeperTwoFactorSms": {
+    "message": "Text message (SMS)"
+  },
+  "keeperTwoFactorDuo": {
```
itsadrago:
```suggestion
"duoSecurity": {
    "message": "Duo Security"
  },
```

apps/browser/src/_locales/en/messages.json
```diff
@@ -4590,6 +4590,132 @@
...
+  "keeperDuoPasscode": {
+    "message": "Passcode"
+  },
+  "keeperDuoUnknown": {
+    "message": "Unknown method"
+  },
+  "keeperDuoApprovalRequired": {
+    "message": "Duo verification"
+  },
+  "keeperDuoPushWaiting": {
```
itsadrago:
```suggestion
 "duoPushWaiting": {
    "message": "Approve the Duo push notification on your device."
  },
```

apps/browser/src/_locales/en/messages.json
```diff
@@ -4590,6 +4590,132 @@
...
+    "message": "Approve the Duo push notification on your device."
+  },
+  "keeperDuoSmsWaiting": {
```
itsadrago:
```suggestion
  "duoSmsWaiting": {
    "message": "A text message has been sent to your phone. Enter the code in the Duo app to continue."
  },
```

apps/browser/src/_locales/en/messages.json
```diff
@@ -4590,6 +4590,132 @@
...
+    "message": "A text message has been sent to your phone. Enter the code in the Duo app to continue."
+  },
+  "keeperDuoVoiceWaiting": {
```
itsadrago:
```suggestion
 "duoVoiceWaiting": {
    "message": "Answer the phone call from Duo and follow the instructions."
  },
```

apps/browser/src/_locales/en/messages.json
```diff
@@ -4590,6 +4590,132 @@
...
+  "keeperDuoVoice": {
+    "message": "Phone call"
+  },
+  "keeperDuoPasscode": {
+    "message": "Passcode"
+  },
+  "keeperDuoUnknown": {
+    "message": "Unknown method"
+  },
+  "keeperDuoApprovalRequired": {
```
itsadrago:
```suggestion
  "duoVerification": {
    "message": "Duo verification"
  },
```

apps/browser/src/_locales/en/messages.json
```diff
@@ -4590,6 +4590,132 @@
...
+  "keeperDuoSms": {
+    "message": "Text message (SMS)"
+  },
+  "keeperDuoVoice": {
+    "message": "Phone call"
+  },
+  "keeperDuoPasscode": {
+    "message": "Passcode"
+  },
+  "keeperDuoUnknown": {
```
itsadrago:
```suggestion
  "unknownMethod": {
    "message": "Unknown method"
  },
```

apps/browser/src/_locales/en/messages.json
```diff
@@ -4590,6 +4590,132 @@
...
+  "keeperDuoPush": {
+    "message": "Duo Push"
+  },
+  "keeperDuoSms": {
+    "message": "Text message (SMS)"
+  },
+  "keeperDuoVoice": {
+    "message": "Phone call"
+  },
+  "keeperDuoPasscode": {
```
itsadrago: This message already exist using the key `passcode` across web/browser/desktop and so it's unnecessary

apps/browser/src/_locales/en/messages.json
```diff
@@ -4590,6 +4590,132 @@
...
+  "keeperSelectDuoMethod": {
+    "message": "Select Duo method"
+  },
+  "keeperSelectDuoMethodDesc": {
+    "message": "Choose how you want to verify with Duo Security."
+  },
+  "keeperDuoPush": {
+    "message": "Duo Push"
+  },
+  "keeperDuoSms": {
```
itsadrago: Already exists in suggestion above using key `textMessageSms`

apps/browser/src/_locales/en/messages.json
```diff
@@ -4590,6 +4590,132 @@
   "lastPassYubikeyDesc": {
     "message": "Insert the YubiKey associated with your LastPass account into your computer's USB port, then touch its button."
   },
+  "importDirectlyFromKeeper": {
+    "message": "Import directly from Keeper"
+  },
+  "keeperEmail": {
+    "message": "Keeper Email"
+  },
+  "keeperRegion": {
```
itsadrago:
```suggestion
"dataCenterLocation": {
    "message": "Data Center Location"
  },
```

apps/browser/src/_locales/en/messages.json
```diff
@@ -4590,6 +4590,132 @@
...
+  "keeperMasterPassword": {
+    "message": "Keeper master password"
+  },
+  "keeperAuthRequired": {
+    "message": "Keeper authentication required"
+  },
+  "keeperDeviceApprovalRequired": {
+    "message": "Keeper device approval required"
+  },
+  "keeperSelectApprovalMethod": {
```
itsadrago:
```suggestion
"selectApprovalMethod": {
    "message": "Select approval method"
  },
```

apps/browser/src/_locales/en/messages.json
```diff
@@ -4590,6 +4590,132 @@
...
+    "message": "Select approval method"
+  },
+  "keeperSelectApprovalMethodDesc": {
```
itsadrago:
```suggestion
  "selectApprovalMethodDesc": {
    "message": "Choose how you want to approve this device."
  },
```

apps/browser/src/_locales/en/messages.json
```diff
@@ -4590,6 +4590,132 @@
...
+    "message": "Choose how you want to approve this device."
+  },
+  "keeperApprovalMethodEmail": {
+    "message": "Email"
+  },
+  "keeperApprovalMethodPush": {
+    "message": "Keeper Push"
+  },
+  "keeperDeviceApprovalEmailDesc": {
```
itsadrago:
```suggestion
"approvalEmailDesc": {
    "message": "Check your email and click the approval link, or enter the verification code below."
  },
```

itsadrago (changes requested):
There are some conventions that are generally followed in the messages json file...
1. The key name should mirror (to a reasonable degree) the message
2. When possible, write the message in a generic way such that it can be reused
3. Make use of placeholder replacement to enable reusable / dynamic content

libs/importer/src/importers/keeper/access/services/client.ts
```diff
@@ -0,0 +1,862 @@
...
+    [DuoMethod.Push, TwoFactorPushType.TWO_FA_PUSH_DUO_PUSH],
+    [DuoMethod.Sms, TwoFactorPushType.TWO_FA_PUSH_DUO_TEXT],
+    [DuoMethod.Voice, TwoFactorPushType.TWO_FA_PUSH_DUO_CALL],
+  ]);
+
+  private supported2faMethods: Set<TwoFactorMethod> = new Set([
+    TwoFactorMethod.Totp,
+    TwoFactorMethod.Sms,
+    TwoFactorMethod.Duo,
+    TwoFactorMethod.KeeperPush,
```
harr1424: ❓  Is this handled in the method below? Or will it fall through to default?

libs/importer/src/components/keeper/dialog/keeper-approval-method-select.component.ts
```diff
@@ -0,0 +1,53 @@
...
+} from "@bitwarden/components";
+
+import { DeviceApprovalChannel } from "../../../importers/keeper/access";
+
+type KeeperApprovalMethodSelectData = {
+  methods: DeviceApprovalChannel[];
+};
+
+// FIXME(https://bitwarden.atlassian.net/browse/CL-764): Migrate to OnPush
+// eslint-disable-next-line @angular-eslint/prefer-on-push-component-change-detection
```
harr1424: >Use the OnPush change detection strategy. We will eventually enforce this project wide for performance reasons.

https://contributing.bitwarden.com/contributing/code-style/web/angular

libs/importer/src/components/keeper/dialog/keeper-device-approval-prompt.component.ts
```diff
@@ -0,0 +1,72 @@
...
+} from "@bitwarden/components";
+
+export type KeeperDeviceApprovalVariant = "email" | "push";
+
+type KeeperDeviceApprovalPromptData = {
+  variant: KeeperDeviceApprovalVariant;
+};
+
+// FIXME(https://bitwarden.atlassian.net/browse/CL-764): Migrate to OnPush
+// eslint-disable-next-line @angular-eslint/prefer-on-push-component-change-detection
```
harr1424: >Use the OnPush change detection strategy. We will eventually enforce this project wide for performance reasons.

https://contributing.bitwarden.com/contributing/code-style/web/angular

libs/importer/src/components/keeper/dialog/keeper-duo-method-select.component.ts
```diff
@@ -0,0 +1,58 @@
...
+
+import { DuoMethod } from "../../../importers/keeper/access";
+
+type KeeperDuoMethodSelectData = {
+  methods: DuoMethod[];
+  phoneNumber: string;
+};
+
+// FIXME(https://bitwarden.atlassian.net/browse/CL-764): Migrate to OnPush
+// eslint-disable-next-line @angular-eslint/prefer-on-push-component-change-detection
```
harr1424: >Use the OnPush change detection strategy. We will eventually enforce this project wide for performance reasons.

https://contributing.bitwarden.com/contributing/code-style/web/angular

libs/importer/src/components/keeper/dialog/keeper-duo-push-prompt.component.ts
```diff
@@ -0,0 +1,50 @@
...
+} from "@bitwarden/components";
+
+import { DuoMethod } from "../../../importers/keeper/access";
+
+type KeeperDuoPushPromptData = {
+  method: DuoMethod;
+};
+
+// FIXME(https://bitwarden.atlassian.net/browse/CL-764): Migrate to OnPush
+// eslint-disable-next-line @angular-eslint/prefer-on-push-component-change-detection
```
harr1424: >Use the OnPush change detection strategy. We will eventually enforce this project wide for performance reasons.

https://contributing.bitwarden.com/contributing/code-style/web/angular

libs/importer/src/components/keeper/dialog/keeper-error.component.ts
```diff
@@ -0,0 +1,30 @@
...
+import {
+  DIALOG_DATA,
+  ButtonModule,
+  DialogModule,
+  DialogService,
+  TypographyModule,
+} from "@bitwarden/components";
+
+// FIXME(https://bitwarden.atlassian.net/browse/CL-764): Migrate to OnPush
+// eslint-disable-next-line @angular-eslint/prefer-on-push-component-change-detection
```
harr1424: >Use the OnPush change detection strategy. We will eventually enforce this project wide for performance reasons.

https://contributing.bitwarden.com/contributing/code-style/web/angular

libs/importer/src/components/keeper/dialog/keeper-multifactor-prompt.component.ts
```diff
@@ -0,0 +1,85 @@
...
+} from "@bitwarden/components";
+
+export type KeeperMultifactorPromptVariant = "totp" | "push";
+
+type KeeperMultifactorPromptData = {
+  variant: KeeperMultifactorPromptVariant;
+};
+
+// FIXME(https://bitwarden.atlassian.net/browse/CL-764): Migrate to OnPush
+// eslint-disable-next-line @angular-eslint/prefer-on-push-component-change-detection
```
harr1424: >Use the OnPush change detection strategy. We will eventually enforce this project wide for performance reasons.

https://contributing.bitwarden.com/contributing/code-style/web/angular

libs/importer/src/components/keeper/dialog/keeper-password-prompt.component.ts
```diff
@@ -0,0 +1,56 @@
...
+  ButtonModule,
+  DialogModule,
+  DialogService,
+  FormFieldModule,
+  IconButtonModule,
+  TypographyModule,
+} from "@bitwarden/components";
+
+// FIXME(https://bitwarden.atlassian.net/browse/CL-764): Migrate to OnPush
+// eslint-disable-next-line @angular-eslint/prefer-on-push-component-change-detection
```
harr1424: >Use the OnPush change detection strategy. We will eventually enforce this project wide for performance reasons.

https://contributing.bitwarden.com/contributing/code-style/web/angular

libs/importer/src/components/keeper/dialog/keeper-two-factor-method-select.component.ts
```diff
@@ -0,0 +1,66 @@
...
+} from "@bitwarden/components";
+
+import { TwoFactorMethod } from "../../../importers/keeper/access";
+
+type KeeperTwoFactorMethodSelectData = {
+  methods: TwoFactorMethod[];
+};
+
+// FIXME(https://bitwarden.atlassian.net/browse/CL-764): Migrate to OnPush
+// eslint-disable-next-line @angular-eslint/prefer-on-push-component-change-detection
```
harr1424: >Use the OnPush change detection strategy. We will eventually enforce this project wide for performance reasons.

https://contributing.bitwarden.com/contributing/code-style/web/angular

libs/importer/src/components/keeper/dialog/keeper-two-factor-method-select.component.ts
```diff
@@ -0,0 +1,66 @@
...
+      case TwoFactorMethod.U2f:
+        return "keeperTwoFactorU2f";
+      case TwoFactorMethod.WebAuthn:
+        return "keeperTwoFactorWebAuthn";
+      case TwoFactorMethod.KeeperPush:
+        return "keeperTwoFactorKeeperPush";
+      case TwoFactorMethod.KeeperDna:
+        return "keeperTwoFactorKeeperDna";
+      default:
+        return "keeperTwoFactorUnknown";
```
harr1424: @claude can you confirm if all of these 2FA methods are implemented, and review the implementations for correctness?

libs/importer/src/components/keeper/import-keeper.component.ts
```diff
@@ -0,0 +1,148 @@
...
+  protected emailHint$ = this.formGroup.controls.email.statusChanges.pipe(
+    map((status) => {
+      if (status === "PENDING") {
+        return this.i18nService.t("importingYourAccount");
+      }
+    }),
+  );
+
+  // FIXME(https://bitwarden.atlassian.net/browse/CL-903): Migrate to Signals
+  // eslint-disable-next-line @angular-eslint/prefer-output-emitter-ref
```
harr1424: >Use the OnPush change detection strategy. We will eventually enforce this project wide for performance reasons.

https://contributing.bitwarden.com/contributing/code-style/web/angular

libs/importer/src/components/keeper/import-keeper.component.ts
```diff
@@ -0,0 +1,148 @@
...
+        return {
+          errors: {
+            message: this.i18nService.t(this.getValidationErrorI18nKey(error)),
+          },
+        };
+      }
+    };
+  }
+
+  // TODO: Review these error messages!
```
harr1424: ❓ Have these been reviewed?

libs/importer/src/components/import.component.html
```diff
@@ -155,9 +155,23 @@ <h2 class="tw-font-medium" bitTypography="h6">{{ "data" | i18n }}</h2>
           click "Show me my data" to save the CSV file.
         </ng-container>
         <ng-container *ngIf="format === 'keepercsv'">
-          Log into the Keeper web vault (keepersecurity.com/vault). Click on your "account email"
-          (top right) and select "Settings". Go to "Export" and find the "Export to .csv File"
-          option. Click "Export" to save the CSV file.
+          <p bitTypography="body1">
+            Log into the Keeper web vault (keepersecurity.com/vault). Click on your "account email"
+            (top right) and select "Settings". Go to "Export" and find the "Export to .csv File"
+            option. Click "Export" to save the CSV file.
```
harr1424: ❓ Should this use a translation key?

libs/importer/src/importers/keeper/access/models/record-v3.ts
```diff
@@ -0,0 +1,11 @@
+import { VaultField } from "./vault-field";
+
+// TODO: Add RecordV2!
```
harr1424, line 3: ❓ Should this be followed up on?

libs/importer/src/importers/keeper/access/services/client.ts
harr1424, line 1: @claude please review this file from a security perspective, is sensitive information handled according to best practices? Please compare the implementation present in this file against the guidelines at this URL: https://contributing.bitwarden.com/architecture/cryptography/crypto-guide

libs/importer/src/importers/keeper/access/services/crypto.ts
harr1424, line 1: @claude please review this file from a security perspective, is sensitive information handled according to best practices? Please compare the implementation present in this file against the guidelines at this URL: https://contributing.bitwarden.com/architecture/cryptography/crypto-guide

libs/importer/src/services/import.service.abstraction.ts
```diff
@@ -24,4 +24,12 @@ export abstract class ImportServiceAbstraction {
...
     organizationId: string,
   ) => Importer;
+
+  // Import an already-parsed ImportResult directly.
+  importImportResult: (
+    importResult: ImportResult,
+    organizationId?: string,
+    selectedImportTarget?: FolderView | CollectionView,
+    canAccessImportExport?: boolean,
+  ) => Promise<ImportResult>;
```
harr1424, line 34: ❓ Why is this being added?

libs/importer/src/services/import.service.ts
```diff
@@ -144,6 +144,20 @@ export class ImportService implements ImportServiceAbstraction {
...
+      canAccessImportExport,
+    );
+  }
+
+  async importImportResult(
+    importResult: ImportResult,
+    organizationId: OrganizationId = null,
+    selectedImportTarget: FolderView | CollectionView = null,
+    canAccessImportExport: boolean = false,
+  ): Promise<ImportResult> {
```
harr1424, line 162: ❓ Same as in abstraction, I am curious why this is being added

harr1424 (commented):
In general, we try to avoid adding tech debt as part of new work, most of my comments involve [a contributor guideline involving the use of OnPush change detecton](https://contributing.bitwarden.com/contributing/code-style/web/angular-migration-guide#onpush-change-detection). 

I've also asked a few questions regarding TODOs and implementation details. Lastly, due to the size and scope of this PR, I've asked our Claude review bot some questions, I hope you understand, this is not intended to diminish or disrespect your contribution.

libs/importer/src/components/keeper/dialog/keeper-approval-method-select.component.ts
detunized: Addressed by @itsadrago in https://github.com/detunized/clients/pull/2

libs/importer/src/components/keeper/dialog/keeper-device-approval-prompt.component.ts
detunized: Addressed by @itsadrago in https://github.com/detunized/clients/pull/2

libs/importer/src/importers/keeper/access/services/client.ts
detunized: TOTP, SMS and Duo and 100% supported. KeeperPush is supported for as a new device approval method and I'm working on the 2FA bit. This will be added to this PR later. I paused it for a bit to address all the feedback I received so far. So technically this is valid. It shouldn't be in the supported 2FA methods.

libs/importer/src/components/keeper/dialog/keeper-two-factor-method-select.component.ts
detunized: No, not all of these methods are supported. This is a user facing string selection code and I wonder if it makes sense to have these strings added already to make it easier to add implementations in the followup PRs.  This maps to the enums in the Keeper code and it usually takes an extra effort to find those in the SDK or minimized JS code on the website. If it's against the policy, I can also take the strings out that are not actively used and just put them in the code as a comment to be added in the future.

libs/importer/src/components/keeper/import-keeper.component.ts
detunized: Addressed by @itsadrago

detunized: Actually just the TODO have been removed. It would make sense to validate that these strings map to the actual errors. I will add this to the TODO section in the README.

libs/importer/src/components/import.component.html
detunized: It looks like everything around there uses the same pattern. I didn't really change the text, I just put it inside a `p` tag.

libs/importer/src/importers/keeper/access/models/record-v3.ts
detunized: This is actually a bigger issue: how to test this with the actual legacy accounts which we do not have access to. All the new accounts create V3 records only. We would need to have access to an account created potentially a few years ago. 

Alternatively this can be just copied from the C#/Python SDK V2 handling code and just hope for the best. No live testing.

Already added to README.

libs/importer/src/importers/keeper/access/services/client.ts
detunized: The guy just leaves the :eyes: under each comment. Is it supposed to reply at some point? I can ask my Claude to do the same.

libs/importer/src/services/import.service.abstraction.ts
detunized: All importers, except for the LastPass direct importer, import data from a file (csv, json, whatever). The LastPass direct importer imports directly from the account using REST APIs and then it generates a CSV and passes it on to the CSV importer. The Keeper direct importer doesn't do that. The data format is too complicated to be serialized into a csv or something like that. So I added this method to the import service to ingest the `ImportResult` directly.

libs/importer/src/services/import.service.ts
detunized: All importers, except for the LastPass direct importer, import data from a file (csv, json, whatever). The LastPass direct importer imports directly from the account using REST APIs and then it generates a CSV and passes it on to the CSV importer. The Keeper direct importer doesn't do that. The data format is too complicated to be serialized into a csv or something like that. So I added this method to the import service to ingest the ImportResult directly.

libs/importer/src/importers/keeper/access/services/crypto.ts
detunized: ## Security Review: `crypto.ts`

Reviewed against the [Bitwarden Crypto Guide](https://contributing.bitwarden.com/architecture/cryptography/crypto-guide).

### Context

This file implements Keeper's proprietary encryption protocol for interoperability. It is not Bitwarden's own encryption logic. Most findings below are dictated by Keeper's protocol and cannot be changed.

### Findings

**Crypto in TypeScript, not SDK** — The guide states new cryptographic code should be in the SDK. However, this implements a third-party protocol (Keeper) for import purposes, so the SDK constraint doesn't practically apply.

**AES-CBC without MAC (`decryptAesV1`)** — No authentication, susceptible to padding oracle attacks in theory. This is Keeper's legacy "aes-v1" format — we can't change it.

**RSA PKCS#1 v1.5** — Vulnerable to Bleichenbacher-style attacks. Required by Keeper's protocol. Web Crypto doesn't support PKCS#1 v1.5 for encryption, hence he `node-forge` dependency.

**`encryptAesV2` accepts an external nonce** — The WARNING comment about nonce reuse is appropriate. All internal call sites generate fresh nonces (default path), so this is safe in practice.

**Non-constant-time comparison (`uint8ArrayEquals`)** — Used in `decryptEncryptionParams` to verify password correctness by comparing two decrypted 32-byte blocks. A constant-time comparison would be more defensively correct, though the practical risk is low (local comparison of decrypted plaintext, not a
network-facing MAC check).

**Key material not zeroed** — Keys (`Uint8Array`) remain in JS heap until GC'd. The crypto guide acknowledges this is an inherent TypeScript limitation: _"a critical security property absent in TypeScript implementations."_

### What's done well

- Uses Web Crypto API for all standard primitives (AES-GCM, ECDH P-256, PBKDF2-SHA256)
- ECIES pattern in `encryptEc`/`decryptEc` is correct (ephemeral ECDH → SHA-256 → AES-GCM)
- Random generation via `crypto.getRandomValues`
- Fresh nonce generated by default in `encryptAesV2`
- `decryptAesNoPadding` workaround for Web Crypto's mandatory PKCS7 is clever and correct

*^ this is what CC had to say about that*

libs/importer/src/importers/keeper/access/services/client.ts
detunized: ## Security Review: `client.ts`

Reviewed against the [Bitwarden Crypto Guide](https://contributing.bitwarden.com/architecture/cryptography/crypto-guide).

### Context

This file implements the Keeper login and vault sync protocol. It handles authentication, device registration, 2FA, and encrypted API communication.

### Findings

**Password as immutable string** — The password flows as a `string` through `login` → `handleAuthHash` → `deriveV1KeyHash`. JavaScript strings are immutable and cannot be zeroed from memory. Inherent TypeScript limitation acknowledged by the crypto guide.

**Device private key exported to JWK (`getPublicKeyFromPrivate`)** — The full private key is temporarily exported to JWK format to extract the public components. The private `d` component is correctly dropped before re-importing as a public key. The temporary JWK object containing `d` persists in JS memory until GC'd.

**Server URL from redirect** — `this.server` is set from `response.stateSpecificValue` during region redirect. If the Keeper server were compromised, this could redirect API calls to an attacker-controlled endpoint. This is inherent to Keeper's protocol design.

**No sensitive data in logs** — Error logging in the calling component (`import-keeper.component.ts`) logs error objects. These shouldn't contain vault data but could contain server error messages. Looks acceptable.

### What's done well

- Fresh 256-bit transmission key generated per API request
- 2FA token expiration set to `TWO_FA_EXP_IMMEDIATELY` (most secure option)
- All API payloads encrypted with AES-GCM before transmission
- Server responses decrypted and verified via AES-GCM authentication
- Device credentials use ECDH P-256 key pairs
- Login state machine has a bounded iteration limit (10) preventing infinite loops
- Push socket properly cleaned up in `finally` block

*^ this is what CC had to say about that*

apps/browser/src/_locales/en/messages.json
detunized: Replaced with `email`

apps/browser/src/_locales/en/messages.json
detunized: Replaced with `verificationCode`

apps/browser/src/_locales/en/messages.json
detunized: I renamed it to generic `mfaTotpDesc`. I did not change the LastPass importer or anything else. It would be good to do this in a follow-up PR. I can do that later if you'd like me to.

apps/browser/src/_locales/en/messages.json
detunized: Both renamed to `textMessageSms`

apps/browser/src/_locales/en/messages.json
detunized: Rename to `backupCodes`

apps/browser/src/_locales/en/messages.json
detunized: Renamed to `duoSecurity`

apps/browser/src/_locales/en/messages.json
detunized: Renamed to `duoPushWaiting`

apps/browser/src/_locales/en/messages.json
detunized: Renamed to `duoSmsWaiting`

apps/browser/src/_locales/en/messages.json
detunized: Renamed to `duoVerification`

apps/browser/src/_locales/en/messages.json
detunized: Renamed to `unknownMethod`

apps/browser/src/_locales/en/messages.json
detunized: Switched to existing `passcode`

apps/browser/src/_locales/en/messages.json
detunized: Renamed to `dataCenterLocation`

apps/browser/src/_locales/en/messages.json
detunized: Renamed to `duoVoiceWaiting`

apps/browser/src/_locales/en/messages.json
detunized: Renamed to `selectApprovalMethod`

apps/browser/src/_locales/en/messages.json
detunized: Renamed to `selectApprovalMethodDesc`

apps/browser/src/_locales/en/messages.json
detunized: Switched to `textMessageSms`

apps/browser/src/_locales/en/messages.json
detunized: Renamed to `approvalEmailDesc`

libs/importer/src/services/import.service.abstraction.ts
harr1424: Thank you for clarifying!

libs/importer/src/importers/keeper/access/services/client.ts
quexten: @harr1424 The crypto guide is aimed at developers implementing features for the e2ee password manager product, not for compatibility layers with external tooling such as passkeys, SSH, or in this case keeper. There are no guidelines or that, and we do not control how keeper encrypts their stuff.

The one thing I'll note here is that we usually prefer to do crypto - even compatibility with external products - in the SDK directly, though at this point it's too late to change that. It likely means that the node-forge dependency ownership will move to tools, since tools will be the only consumer after KM drops the last remmenants of it from KM TS code.

libs/importer/src/importers/keeper/access/services/crypto.ts
quexten: Same comment applies here. We do not control how keeper imports things, so without reviewing this in detail, all claude review on cryptographic design is fully irrelevant.

.github/renovate.json5
```diff
@@ -295,6 +295,7 @@
       matchPackageNames: [
         "@microsoft/signalr-protocol-msgpack",
         "@microsoft/signalr",
+        "@bufbuild/protobuf",
```
withinfocus, line 307: https://bitwarden.atlassian.net/browse/VULN-526 -- this is approved.

libs/importer/src/components/keeper/dialog/keeper-auth-stage-view.component.html
```diff
@@ -0,0 +1,315 @@
...
+        [primaryLabel]="'continue' | i18n"
+        (primary)="submitCode()"
+        [secondaryLabel]="'tryAnotherMethod' | i18n"
+        (secondary)="tryAnother()"
+        (close)="cancel()"
+      >
+        <div bitDialogContent>
+          <p bitTypography="body1">{{ "approvalEmailDesc" | i18n }}</p>
+          <form (ngSubmit)="submitCode()">
+            <bit-form-field class="!tw-mb-0">
```
mcamirault: ⛏️ You can use the `disableMargin` attribute here instead of setting the class. There's a couple other places in this file this applies but I'll only leave one comment

harr1424:
@claude please review

apps/web/src/locales/en/messages.json
```diff
@@ -13891,6 +13891,195 @@
...
+  },
+  "duoVoiceWaiting": {
+    "message": "Answer the phone call from Duo and follow the instructions."
+  },
+  "tryAnotherMethod": {
+    "message": "Try another method"
+  },
+  "keeperUnsupported2faMethod": {
+    "message": "Your Keeper account requires a two-factor authentication method that is not currently supported by the direct importer."
+  },
```
mcamirault, line 15046: The "noKeeperDataFound" key is missing from this file

libs/importer/src/importers/keeper/access/models/client-options.ts
```diff
@@ -0,0 +1,12 @@
+import { KeeperRegion } from "../enums";
+import { Ui } from "../ui";
+
+export interface ClientOptions {
+  region: KeeperRegion;
+  ui: Ui;
+
+  // Cache for testing. Not needed in production use.
```
mcamirault: 🎨 If not needed then this should be removed, or the fields commented out completely

libs/importer/src/components/keeper/keeper-direct-import-ui.service.ts
```diff
@@ -0,0 +1,330 @@
...
+    this.setStage({
+      kind: "approvalCode",
+      method,
+      variant,
+      previousCodeRejected: options?.previousCodeRejected ?? false,
+    });
+    return this.waitForUser<string | typeof Cancel | typeof Resend | typeof TryAnother>();
+  }
+
+  closeApprovalDialog(): void {
```
mcamirault: 🎨 This and the other no-op functions should be removed

mcamirault (changes requested):
A couple of small comments, only the missing I18N key is really a blocker. Also, I believe that the first screenshot is out of date with the described design with only a single Keeper entry and a sub-selector for the import method/format, can you update it?

detunized:
> lso, I believe that the first screenshot is out of date with the described design with only a single Keeper entry and a sub-selector for the import method/format, can you update it?

@mcamirault  Yes, I'll update it. We've been working on the UI a lot in the last weeks and things changed a lot. This screenshot is from the initial version of the importer. Do we even need the screenshots here? I cannot post all screens, there are too many. I can add a few, like now.

mcamirault:
> > lso, I believe that the first screenshot is out of date with the described design with only a single Keeper entry and a sub-selector for the import method/format, can you update it?
> 
> @mcamirault Yes, I'll update it. We've been working on the UI a lot in the last weeks and things changed a lot. This screenshot is from the initial version of the importer. Do we even need the screenshots here? I cannot post all screens, there are too many. I can add a few, like now.

@detunized You definitely don't need to add all of them, I'm mostly interested in the experience when you select the Keeper option and then direct/json/csv (basically just an updated version of the first pic in the PR). Apologies for the delayed response, I was traveling most of today

libs/importer/src/components/keeper/dialog/keeper-auth-stage-view.component.html
detunized: Done

libs/importer/src/importers/keeper/access/models/client-options.ts
detunized: Done

libs/importer/src/components/keeper/keeper-direct-import-ui.service.ts
detunized: Done

apps/web/src/locales/en/messages.json
detunized: `noKeeperDataFound` is removed as no longer needed

detunized:
> @claude please review

@harr1424 I fixed most. For 2. and 4. I'm waiting on the product team input.

detunized:
> @detunized You definitely don't need to add all of them, I'm mostly interested in the experience when you select the Keeper option and then direct/json/csv (basically just an updated version of the first pic in the PR). Apologies for the delayed response, I was traveling most of today

@mcamirault I updated the screenshots and removed the outdated ones

harr1424:
@claude please give this a re-review, have all findings and comments (resolved or otherwise) been addressed?

itsadrago:
@claude please give this a re-review, have all findings and comments (resolved or otherwise) been addressed?

harr1424 (dismissed):

mcamirault (dismissed):

dani-garcia (dismissed):

harr1424 (approved):

itsadrago (approved):

dani-garcia (approved):

PR #19398, opened 2026-03-05
fix(sso): [PM-33190] SSO Fails on Callback from Successful IdP Login

Patrick-Pimentel-Bitwarden:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-33190

## Root cause

Commit `944b3ffdc2` (\[PM-31406] TypeScript 5.9 type compatibility fixes) replaced `Utils.fromBufferToUrlB64()` with `Utils.fromArrayToUrlB64()` in the SSO component. The commit was described as “non-functional,” but the two code paths differ in padding behavior:

- **Old path** (`fromBufferToUrlB64`): calls `fromB64toUrlB64()` which explicitly strips padding via `.replace(/=/g, “”)` (`utils.ts:298`)
- **New path** (`fromArrayToUrlB64`): calls `arr.toBase64({ alphabet: “base64url” })` via the core-js polyfill, which defaults `omitPadding` to `false`

This means the PKCE code challenge sent during SSO login changed from unpadded (e.g., `abc-_dEf`) to padded (e.g., `abc-_dEf==`). The server's PKCE verification expects unpadded challenges per RFC 7636, so the mismatch causes `invalid_grant` on the `/connect/token` call.

## Fix

`libs/common/src/platform/misc/utils.ts:179` — added `omitPadding: true` to the `.toBase64()` call in `fromArrayToUrlB64`:

```typescript
// Before (broken):
return arr.toBase64({ alphabet: “base64url” });

// After (fixed):
return arr.toBase64({ alphabet: “base64url”, omitPadding: true });
```

This restores parity with the old `fromBufferToUrlB64` behavior. The fix is global to all callers of `fromArrayToUrlB64`, not just SSO. The existing unit test at `utils.spec.ts:484` (`expect(urlB64String).not.toContain(“=”)`) confirms the regression and validates the fix.

## 📸 Screenshots

https://github.com/user-attachments/assets/e4efe741-0012-4046-839a-5a5335c31e40

[PM-31406]: https://bitwarden.atlassian.net/browse/PM-31406?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


djsmith85:
FYI @quexten

Patrick-Pimentel-Bitwarden:
We are aligned that the `fromArrayToB64` should strip the padding then correct? @quexten

quexten:
@Patrick-Pimentel-Bitwarden As you correctly point out the previous function stripped the padding. The new function should have the same behavior, all that was supposed to change is the input type.

dereknance (approved):

Patrick-Pimentel-Bitwarden:
When I stop and think about this I don't understand why we were originally stripping the padding for these utility functions. That also feels like a red flag. I can put it back the way it was but that feels wrong.

libs/common/src/platform/misc/utils.ts
```diff
@@ -163,20 +163,22 @@ export class Utils {
     return arr.toBase64({ alphabet: "base64" });
   }
 
-  static fromArrayToUrlB64(arr: Uint8Array): string;
-  static fromArrayToUrlB64(arr: null): null;
   /**
-   * Converts a Uint8Array to a URL-safe Base64 encoded string.
+   * Converts a Uint8Array to a URL-safe Base64 encoded string. This means we are
+   * stripping the padding off so that the padding characters '=' don't affect the
+   * url negatively.
```
quexten:
```suggestion
   * Converts a Uint8Array to a URL-safe Base64 encoded string, while stripping padding.
```

Base64URL does allow for padding, so the reasoning as-is is not valid https://datatracker.ietf.org/doc/html/rfc4648#section-5. We are just making the choice to not pad in this function.

libs/common/src/platform/misc/utils.spec.ts
```diff
@@ -496,6 +496,80 @@ describe("Utils Service", () => {
...
+      expect(result).not.toContain("=");
+      expect(result).not.toContain("+");
+      expect(result).not.toContain("/");
+    });
+
+    it("should contain URL-safe replacement characters", () => {
+      const result = Utils.fromBufferToUrlB64(sha256DigestBytes.buffer);
+      // The input bytes are chosen to produce '-' and '_' in URL-safe encoding
+      expect(result).toMatch(/^[A-Za-z0-9_-]+$/);
+    });
```
quexten: I don't really understand why we are testing this detailed instead of just testing against a test vector, considering we have a known input.


```suggestion
    const TEST_VECTOR_URL_BASE64 = "u_-78f7vm_G-75vxvu-b8b7v2_G675vx_u-b8f7vm_E"
    it("should output the correct value for the test value", () => {
      const result = Utils.fromBufferToUrlB64(sha256DigestBytes.buffer);
      expect(result).toBe(TEST_VECTOR_URLBASE64);
    });
```

libs/common/src/platform/misc/utils.spec.ts
```diff
@@ -496,6 +496,80 @@ describe("Utils Service", () => {
...
+  });
+
+  describe("fromArrayToUrlB64(...) - SSO PKCE scenario", () => {
+    const sha256DigestBytes = new Uint8Array([
```
quexten, line 516: Same here, just use a test vector?

Patrick-Pimentel-Bitwarden: done with [376cc0f](https://github.com/bitwarden/clients/pull/19398/commits/376cc0f7532d9b842c091286a55df85e1fa0841e)

libs/common/src/platform/misc/utils.ts
Patrick-Pimentel-Bitwarden: changed with [376cc0f](https://github.com/bitwarden/clients/pull/19398/commits/376cc0f7532d9b842c091286a55df85e1fa0841e)

PR #19406, opened 2026-03-05
[PM-31778] Multi-step policy edit dialog

JaredScar:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-31778

## 📔 Objective

Implements a multi-step policy edit dialog component to be used for policies that have multi-steps


apps/web/src/app/vault/services/web-vault-prompt.service.ts
```diff
@@ -65,13 +61,13 @@ export class WebVaultPromptService {
...
-        firstTimeDialog: true,
-      },
-    });
+    //AutoConfirmPolicyDialogComponent.open(this.dialogService, {
+    //  data: {
+    //    policy: new AutoConfirmPolicy(),
+    //    organizationId: organization.id,
+    //    firstTimeDialog: true,
+    //  },
+    //});
```
nick-livefront: ❓ Accidental removal?

JaredScar: My bad, yup. I think it was erroring, so this was a quick way to get it to work and I forgot to add it back..

JaredScar: Doesn't exist yet and it seems this was introduced by @BTreston and is intentional. Tagging him for further clarity

apps/web/src/app/admin-console/organizations/policies/policy-edit-definitions/auto-confirm-policy.component.html
```diff
@@ -57,3 +91,19 @@
     </li>
   </ol>
 </ng-template>
+
+<ng-template #step0Footer>
```
BTreston: this should be `#step1Footer`

apps/web/src/app/admin-console/organizations/policies/base-policy-edit.component.ts
```diff
@@ -86,6 +87,12 @@ export abstract class BasePolicyEditComponent implements OnInit {
   // FIXME(https://bitwarden.atlassian.net/browse/CL-903): Migrate to Signals
   // eslint-disable-next-line @angular-eslint/prefer-signals
   @Input() policy: BasePolicyEditDefinition | undefined;
+  // FIXME(https://bitwarden.atlassian.net/browse/CL-903): Migrate to Signals
+  // eslint-disable-next-line @angular-eslint/prefer-signals
+  @Input() currentStep: Signal<number> = signal(0);
+  // FIXME(https://bitwarden.atlassian.net/browse/CL-903): Migrate to Signals
+  // eslint-disable-next-line @angular-eslint/prefer-signals
+  @Input() organizationId: string | undefined;
```
BTreston: please address these  `FIXME`s

multi-step-policy-plan.md
```diff
@@ -0,0 +1,304 @@
+🎯 Multi-step Policy Edit Dialog Component
```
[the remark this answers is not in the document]
sven-bitwarden: Have we decided to commit plans/are they going here in the the root of apps?

apps/web/src/app/admin-console/organizations/policies/policy-edit-definitions/auto-confirm-policy.component.html
```diff
@@ -57,3 +91,19 @@
...
+    bitFormButton
+    type="submit"
+  >
+    {{ "openExtension" | i18n }}
+    <i class="bwi bwi-external-link tw-ml-1" aria-hidden="true"></i>
+  </button>
+  <button bitButton buttonType="secondary" bitDialogClose type="button">
+    {{ "close" | i18n }}
+  </button>
+</ng-template>
```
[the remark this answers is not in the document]
sven-bitwarden: Have you seen the claude suggestions?

apps/web/src/app/admin-console/organizations/policies/policy-edit-dialog.component.ts
```diff
@@ -135,7 +135,8 @@ export class PolicyEditDialogComponent implements AfterViewInit {
       throw new Error("PolicyComponent not initialized.");
     }
 
-    if ((await this.policyComponent.confirm()) == false) {
+    const confirmable = this.policyComponent as { confirm?: () => Promise<boolean> | boolean };
+    if (typeof confirmable.confirm === "function" && (await confirmable.confirm()) == false) {
```
BTreston: the `confirm` method is being removed, this is no longer applicable.

apps/web/src/app/vault/services/web-vault-prompt.service.ts
BTreston: There are a few things that need to happen here. 

1. We need to add a new public property property `firstTimeDialog` to the `AutoConfirmPolicy` policy edit definition
2. This policy edit definition needs a constructor with a `firstTimeDialog: boolean = false` default param
3. The `AutoConfirmPolicyEditComponent` below it needs 
``` typescript
override policy: AutoConfirmPolicy | undefined;
```
This enables us to define some configuration on the actual policy object itself (e.g. the `firstTimeDialog` boolean) and lets the policy edit component instance access properties of concrete policy definition instance while still maintaining the same interface generic interface. This tightly couples the policy definition to the edit component, but I actually see that as a benefit.

4. The template needs to be updated to read this property from the `policy` object.
5. This caller needs to be updated to: 
```typescript
const config: DialogConfig<PolicyEditDialogData> = {
        data: {
          policy: new AutoConfirmPolicy(true),
          organizationId: organization.id,
        },
    };

    this.dialogService.open(MultiStepPolicyEditDialogComponent, config)
```
6. Finally, the side effect logic in the now deleted `auto-confirm-edit-policy-dialog` need to be re-implemented in the `auto-confirm-policy.component.ts` and properly defined as the `sideEffect` property for each step in `this.policySteps`.

multi-step-policy-plan.md
BTreston, line 1: I don't think we need to check this into src control. I included it originally as context for you when handing this off

apps/web/src/app/admin-console/organizations/policies/policy-edit-dialogs/multi-step-policy-edit-dialog.component.ts
```diff
@@ -0,0 +1,175 @@
...
+    );
+  }
+
+  override async ngAfterViewInit() {
+    const policyResponse = await this.load();
+    this.loading = false;
+    // Force re-render so the @else branch renders #policyForm and policyFormRef is resolved.
+    // #policyForm lives inside @if (loading) ... @else { ... }, so it doesn't exist until
+    // loading is false and change detection has run.
+    this.changeDetectorRef.detectChanges();
```
BTreston: If we migrate this to a signal then we should not have to do this.

apps/web/src/app/admin-console/organizations/policies/policy-edit-definitions/vnext-organization-data-ownership.component.ts
```diff
@@ -49,10 +73,60 @@ export class vNextOrganizationDataOwnershipPolicyComponent
...
+
+        const orgKey = orgKeys[this.organizationId as OrganizationId];
+        const request = await this.buildVNextRequest(orgKey);
+
+        await this.policyApiService.putPolicyVNext(
+          this.organizationId ?? "",
+          this.policy?.type ?? PolicyType.OrganizationDataOwnership,
+          request,
+        );
+      },
```
BTreston: lets define this as a private function and just reference it here rather than defining it inline.

BTreston (changes requested):

nick-livefront (dismissed):

apps/web/src/app/admin-console/organizations/policies/policy-edit-definitions/auto-confirm-policy.component.ts
```diff
@@ -41,21 +48,128 @@ export class AutoConfirmPolicy extends BasePolicyEditDefinition {
...
+    this.enabled.valueChanges.pipe(startWith(this.enabled.value)),
+  ]).pipe(map(([policyEnabled, value]) => !policyEnabled && !value));
+
+  readonly policySteps: PolicyStep[] = [
+    {
+      titleContent: this.step0Title,
+      bodyContent: this.step0Content,
+      footerContent: this.step0Footer,
+      disableSave: this.saveDisabled$,
+      sideEffect: async () => {
```
BTreston: please extract this into its own private function rather than defining it in line. Same for the other side effect

apps/web/src/app/admin-console/organizations/policies/policy-edit-definitions/auto-confirm-policy.component.ts
```diff
@@ -41,21 +48,128 @@ export class AutoConfirmPolicy extends BasePolicyEditDefinition {
...
+    },
+    {
+      titleContent: this.step1Title,
+      bodyContent: this.step1Content,
+      footerContent: this.step1Footer,
+      sideEffect: async (): Promise<PolicyStepResult> => {
+        await this.router.navigate(["/browser-extension-prompt"], {
+          queryParams: { url: "AutoConfirm" },
+        });
+        return undefined;
```
BTreston: unnecessary return

apps/web/src/app/admin-console/organizations/policies/policy-edit-definitions/vnext-organization-data-ownership.component.html
BTreston, line 1: Our default templates should handle this (according to the jsdoc) Do we have some sane defaults defined somewhere? I didn't see them in the PR. We should not have to specify these templates on the policy component, there should only be templates if there are special layouts (like for auto confirm)

apps/web/src/app/admin-console/organizations/policies/policy-edit-dialogs/multi-step-policy-edit-dialog.component.ts
```diff
@@ -0,0 +1,173 @@
...
+      keyService,
+    );
+  }
+
+  override async ngAfterViewInit() {
+    const policyResponse = await this.load();
+    this.loading = false;
+
+    // #policyForm is always rendered (outside any conditional) so policyFormViewRef is
+    // resolved by the time ngAfterViewInit fires — no manual detectChanges() needed.
```
BTreston: remove comment

apps/web/src/app/admin-console/organizations/policies/policy-edit-dialogs/multi-step-policy-edit-dialog.component.ts
```diff
@@ -0,0 +1,173 @@
...
+
+    // Set inputs using ComponentRef API
+    policyComponentRef.setInput("policyResponse", policyResponse);
+    policyComponentRef.setInput("policy", this.data.policy);
+    policyComponentRef.setInput("currentStep", this.currentStep);
+    policyComponentRef.setInput("organizationId", this.data.organizationId);
+
+    // Read step configuration from child component.
+    // The signal write schedules a re-render, which picks up both the updated policySteps
+    // and the loading=false change in the same cycle — no manual detectChanges() needed.
```
BTreston: remove comment

apps/web/src/app/admin-console/organizations/policies/policy-edit-dialogs/multi-step-policy-edit-dialog.component.ts
```diff
@@ -0,0 +1,173 @@
...
+  private readonly policyFormViewRef: Signal<ViewContainerRef | undefined> = viewChild(
+    "policyForm",
+    { read: ViewContainerRef },
+  );
+
+  protected readonly policySteps: WritableSignal<PolicyStep[]> = signal([]);
+  readonly currentStep: WritableSignal<number> = signal(0);
+
+  // eslint-disable-next-line @bitwarden/components/enforce-readonly-angular-properties -- mutable override of base-class field; reassigned by updateSaveDisabled() on each step transition
+  override saveDisabled$: Observable<boolean> = of(false);
```
BTreston: can we convert this to a signal? We can use rxjs `toSignal()` on the observable supplied by the policy

BTreston (changes requested):

apps/web/src/app/admin-console/organizations/policies/policy-edit-definitions/vnext-organization-data-ownership.component.ts
```diff
@@ -49,10 +66,37 @@ export class vNextOrganizationDataOwnershipPolicyComponent
   extends BasePolicyEditComponent
   implements OnInit
 {
+  protected readonly centralizeDataOwnershipEnabled$: Observable<boolean> = defer(() =>
+    from(
+      this.policyApiService.getPolicy(
+        this.organizationId ?? "",
```
BTreston: we should not need `?? ''` here since we're using `defer`. If the org is not available anyway, we'd want to know.

apps/web/src/app/admin-console/organizations/policies/policy-edit-dialogs/multi-step-policy-edit-dialog.component.html
```diff
@@ -0,0 +1,55 @@
...
+      @if (titleTemplate) {
+        <ng-container [ngTemplateOutlet]="titleTemplate"></ng-container>
+      } @else {
+        {{ "editPolicy" | i18n }}
+      }
+    </ng-container>
+
+    @if (!policySteps()[currentStep()]?.titleContent) {
+      <span bitDialogSubtitle>{{ policy.name | i18n }}</span>
+    }
```
BTreston: `bitDialogSubtitle` does not exist. This will not show up in the UI.

apps/web/src/app/admin-console/organizations/policies/policy-edit-dialogs/multi-step-policy-edit-dialog.component.html
```diff
@@ -0,0 +1,55 @@
...
+
+    <ng-container bitDialogContent>
+      @if (loading) {
+        <div>
+          <span class="tw-sr-only">{{ "loading" | i18n }}</span>
```
BTreston: I think we'd want the spinner here too.

apps/web/src/app/admin-console/organizations/policies/policy-edit-dialogs/multi-step-policy-edit-dialog.component.ts
```diff
@@ -0,0 +1,167 @@
...
+    ),
+    { initialValue: false },
+  );
+
+  constructor(
+    @Inject(DIALOG_DATA) data: PolicyEditDialogData,
+    accountService: AccountService,
+    policyApiService: PolicyApiServiceAbstraction,
+    i18nService: I18nService,
+    // Not stored — injected only to satisfy the base-class constructor.
```
BTreston: remove comment

apps/web/src/app/admin-console/organizations/policies/base-policy-edit.component.ts
BTreston: The FIXME comments are not addressed

apps/web/src/app/admin-console/organizations/policies/base-policy-edit.component.ts
```diff
@@ -105,6 +117,13 @@ export abstract class BasePolicyEditComponent implements OnInit {
     }
   }
 
+  /**
```
BTreston, line 133: Please add a TODO comment to clean this up during feature flag removal

apps/web/src/app/admin-console/organizations/policies/policy-edit-dialog.component.ts
```diff
@@ -135,9 +135,11 @@ export class PolicyEditDialogComponent implements AfterViewInit {
...
     }
 
-    if ((await this.policyComponent.confirm()) == false) {
-      this.dialogRef.close();
-      return;
+    if (this.policyComponent.confirm) {
+      const confirmed = await this.policyComponent.confirm();
+      if (!confirmed) {
+        return;
+      }
```
BTreston: lets also add a note here to make sure this is cleaned up during feature flag removal

BTreston (changes requested):
Thank you for continuing this, just a few more points of feedback

apps/web/src/app/admin-console/organizations/policies/policy-edit-dialogs/multi-step-policy-edit-dialog.component.html
```diff
@@ -1,5 +1,8 @@
 <form [formGroup]="formGroup" [bitSubmit]="submit">
-  <bit-dialog [loading]="loading">
+  <bit-dialog
+    [loading]="loading"
+    [subtitle]="!policySteps()[currentStep()]?.titleContent ? (policy.name | i18n) : undefined"
```
BTreston: If we're using `subtitle`, then I think we also need to use `title` when there's no custom content. Otherwise, the subtitle appears before the title. I suggest:

- defining up here`@let titleTemplate = policySteps()[currentStep()]?.titleContent`
- using `titleTemplate` for the [subtitle] and [title] ternaries
- inverting the `!` logic in the ternaries (bc its more clear)
- remove the `@else` block
- reuse `titleTemplate` for the `@if` block below

BTreston (changes requested):

apps/web/src/app/vault/services/web-vault-extension-prompt.service.ts
```diff
@@ -93,8 +93,15 @@ export class WebVaultExtensionPromptService {
...
-    const accountAgeDays = accountAgeMs / (1000 * 60 * 60 * 24);
+    // Use UTC midnight for both dates to get an exact calendar-day count,
+    // avoiding off-by-one errors caused by DST transitions.
+    const nowMidnightMs = Date.UTC(now.getFullYear(), now.getMonth(), now.getDate());
+    const creationMidnightMs = Date.UTC(
+      creationDate.getFullYear(),
+      creationDate.getMonth(),
+      creationDate.getDate(),
+    );
+    const accountAgeDays = (nowMidnightMs - creationMidnightMs) / (1000 * 60 * 60 * 24);
```
BTreston: why are we changing this? I would prefer not to make random changes in other team's files

apps/web/src/app/admin-console/organizations/policies/policy-edit-definitions/vnext-organization-data-ownership.component.ts
```diff
@@ -123,6 +167,31 @@ export class vNextOrganizationDataOwnershipPolicyComponent
...
+
+    const request = await this.buildVNextRequest(orgKey);
+
+    await this.policyApiService.putPolicyVNext(
+      this.organizationId() ?? "",
+      this.policy()?.type ?? PolicyType.OrganizationDataOwnership,
+      request,
+    );
+  }
+
```
BTreston: we should define this function in the `BasePolicyEditComponent` so it can be reused across policies and polices that need their own special logic (e.g. auto confirm) can override.

apps/web/src/app/admin-console/organizations/policies/policy-edit-definitions/vnext-organization-data-ownership.component.ts
```diff
@@ -49,10 +66,37 @@ export class vNextOrganizationDataOwnershipPolicyComponent
...
+  protected readonly saveDisabled$: Observable<boolean> = combineLatest([
+    this.centralizeDataOwnershipEnabled$,
+    this.enabled.valueChanges.pipe(startWith(this.enabled.value)),
+  ]).pipe(map(([policyEnabled, value]) => !policyEnabled && !value));
+
+  readonly policySteps: PolicyStep[] = [
+    {
+      disableSave: this.saveDisabled$,
+      sideEffect: this.savePolicy.bind(this),
```
BTreston: why this versus just `this.savePolicy`

apps/web/src/app/admin-console/organizations/policies/policy-edit-definitions/auto-confirm-policy.component.ts
```diff
@@ -41,21 +48,131 @@ export class AutoConfirmPolicy extends BasePolicyEditDefinition {
...
+    this.enabled.valueChanges.pipe(startWith(this.enabled.value)),
+  ]).pipe(map(([policyEnabled, value]) => !policyEnabled && !value));
+
+  readonly policySteps: PolicyStep[] = [
+    {
+      titleContent: this.step0Title,
+      bodyContent: this.step0Content,
+      footerContent: this.step0Footer,
+      disableSave: this.saveDisabled$,
+      sideEffect: () => this.savePolicyStep(),
```
BTreston: same question: can we just do `this.savePolicyStep`

apps/web/src/app/admin-console/organizations/policies/policy-edit-definitions/auto-confirm-policy.component.ts
```diff
@@ -41,21 +48,131 @@ export class AutoConfirmPolicy extends BasePolicyEditDefinition {
...
+    },
+    {
+      titleContent: this.step1Title,
+      bodyContent: this.step1Content,
+      footerContent: this.step1Footer,
+      sideEffect: () => this.navigateToExtensionPromptStep(),
```
BTreston, line 118: same question: `this.navigateToExtensionPromptStep`

apps/web/src/app/admin-console/organizations/policies/policy-edit-definitions/auto-confirm-policy.component.ts
```diff
@@ -41,21 +48,131 @@ export class AutoConfirmPolicy extends BasePolicyEditDefinition {
...
+    },
+  ];
+
+  private async savePolicyStep(): Promise<PolicyStepResult> {
```
BTreston: this would be a case where we would override the proposed `savePolicy` function from the base component.

BTreston (changes requested):

quexten (approved):

nick-livefront (approved):

BTreston (approved):

PR #19408, opened 2026-03-06
[PM-32085] - popup width migration

jaasen-livefront:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-32085

## 📔 Objective

This PR includes a script for migrating from the old to new popup widths that was previously missed in the [original PR](https://github.com/bitwarden/clients/pull/18376) where the `extra-wide` option was removed and the `narrow` option was added.

The migration will do the following:

`wide` -> `default`
`extra-wide` -> `wide`

## 📸 Screenshots

https://github.com/user-attachments/assets/019c5b4b-e3dc-4b3d-8345-46625971c212

### DIFF
--- a/apps/browser/src/platform/popup/layout/popup-size.service.ts
+++ b/apps/browser/src/platform/popup/layout/popup-size.service.ts
@@ -100,7 +100,7 @@ export class PopupSizeService {
       localStorage.setItem(PopupSizeService.LocalStorageKey, storedValue);
     }
 
-    void this.setStyle(storedValue as any);
+    void this.setStyle(PopupSizeService.toPopupWidthOption(migratedValue));
   }
 
   /**
@@ -112,5 +112,17 @@ export class PopupSizeService {
       return "wide";
     }
     return value;
+  }
+
+  private static isPopupWidthOption(value: string | null): value is PopupWidthOption {
+    return value != null && value in PopupWidthOptions;
+  }
+
+  private static toPopupWidthOption(value: string | null): PopupWidthOption {
+    const migrated = PopupSizeService.migrateOldWidthOption(value);
+    if (PopupSizeService.isPopupWidthOption(migrated)) {
+      return migrated;
+    }
+    return "default";
   }
 }
--- a/apps/browser/src/platform/popup/layout/popup-size.service.ts
+++ b/apps/browser/src/platform/popup/layout/popup-size.service.ts
@@ -91,7 +91,26 @@ export class PopupSizeService {
    * To keep the popup size from flickering on bootstrap, we store the width in `localStorage` so we can quickly & synchronously reference it.
    **/
   static initBodyWidthFromLocalStorage() {
-    const storedValue = localStorage.getItem(PopupSizeService.LocalStorageKey);
+    let storedValue = localStorage.getItem(PopupSizeService.LocalStorageKey);
+
+    // Migrate old width option keys that no longer exist
+    const migratedValue = PopupSizeService.migrateOldWidthOption(storedValue);
+    if (migratedValue !== storedValue && migratedValue != null) {
+      storedValue = migratedValue;
+      localStorage.setItem(PopupSizeService.LocalStorageKey, storedValue);
+    }
+
     void this.setStyle(storedValue as any);
+  }
+
+  /**
+   * Maps old popup width option keys to their new equivalents.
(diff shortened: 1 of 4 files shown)


djsmith85 (dismissed):
Platform-changes look good.

jaasen-livefront:
@willmartian Would you mind taking a look when you have a chance? Ty!

apps/browser/src/platform/popup/layout/popup-size.service.ts
willmartian, line 1: Would it be possible for this logic to live in the migrator itself? Or is that bad form? cc @djsmith85 for Platform opinion.

willmartian (changes requested):

apps/browser/src/platform/popup/layout/popup-size.service.ts
jaasen-livefront: Good question. The short answer: the localStorage migration probably can't live in the state migrator, and it's not bad form to keep it in the service.

Here's why:

1. **Different storage layers.** The state migrator operates on Bitwarden's state provider
   system (via `MigrationHelper`). It doesn't have access to browser `localStorage`. The
   service-level migration in `migrateOldWidthOption` handles the separate `localStorage`
   cache under `bw-popup-width`, which exists purely for synchronous bootstrap to prevent
   flickering.

2. **Different key formats.** The mappings are also slightly different — the state migrator
   maps `wide` → `default` and `extra-wide` → `wide` (kebab-case), while the service maps
   `extraWide` → `wide` (camelCase). This suggests the localStorage values used different
   naming than the state provider values.

3. **Timing matters.** `initBodyWidthFromLocalStorage` is called synchronously during popup
   bootstrap, before the state provider is even initialized. The migrator runs asynchronously
   later. If you removed the localStorage migration from the service and relied on the state
   migrator, users could see a flicker on the first load after upgrade.

So keeping the localStorage migration in the service is the right call — it's the only code that reads from that storage layer, and it needs to handle stale values at read-time. The state migrator handles its own domain (the state provider), and the service handles its own (localStorage).

willmartian: Sounds good! Thank you for the explanation.

willmartian (changes requested):
Could we use a type guard here to avoid the `as any` cast? 

```typescript
private static isPopupWidthOption(value: string | null): value is PopupWidthOption {
  return value != null && value in PopupWidthOptions;
}

private static toPopupWidthOption(value: string | null): PopupWidthOption {
  if (value === "extraWide") {
    return "wide";
  }
  if (PopupSizeService.isPopupWidthOption(value)) {
    return value;
  }
  return "default";
}
```

Then `initBodyWidthFromLocalStorage` can drop the cast, null check, and localStorage write entirely:

```typescript
static initBodyWidthFromLocalStorage() {
  const storedValue = localStorage.getItem(PopupSizeService.LocalStorageKey);
  void this.setStyle(PopupSizeService.toPopupWidthOption(storedValue));
}
```

jaasen-livefront:
> Could we use a type guard here to avoid the `as any` cast?
> 
> ```ts
> private static isPopupWidthOption(value: string | null): value is PopupWidthOption {
>   return value != null && value in PopupWidthOptions;
> }
> 
> private static toPopupWidthOption(value: string | null): PopupWidthOption {
>   if (value === "extraWide") {
>     return "wide";
>   }
>   if (PopupSizeService.isPopupWidthOption(value)) {
>     return value;
>   }
>   return "default";
> }
> ```
> 
> Then `initBodyWidthFromLocalStorage` can drop the cast, null check, and localStorage write entirely:
> 
> ```ts
> static initBodyWidthFromLocalStorage() {
>   const storedValue = localStorage.getItem(PopupSizeService.LocalStorageKey);
>   void this.setStyle(PopupSizeService.toPopupWidthOption(storedValue));
> }
> ```

Love it! It allows us to also reuse `migrateOldWidthOption`.  Fixed in https://github.com/bitwarden/clients/pull/19408/commits/f3e5e7bca4a198cbaae97d6d24f66a20f5899c5d

jaasen-livefront:
@willmartian @djsmith85 Ready for another look when you have some time. Ty!

djsmith85 (approved):

willmartian (approved):

Issue #19382, opened 2026-03-05
I can add my own account as an emergency contact for myself.


RokeJulianLockhart:
### Steps To Reproduce

1. Visit [`vault.bitwarden.com/#/settings/emergency-access`](https://vault.bitwarden.com/#/settings/emergency-access).

1. Enter the output of `bw status | jq -r '.userEmail'` as the e-mail address of the to-be-added emergency contact.

1. Conduct the verification procedure.

### Expected Result

I presume that I should not be able to add myself as my own emergency contact.

### Actual Result

I am able to add myself as my own emergency contact.

### Screenshots or Videos

<p></p><img width="1276" height="1076" alt="Image" src="https://github.com/user-attachments/assets/846f617f-cd08-416d-8c45-df8248461da1" /><p></p>

### Additional Context

This was originally discussed at [`community.bitwarden.com/t/94503/2`](https://community.bitwarden.com/t/whats-the-purpose-in-adding-oneself-as-their-own-emergency-contact/94503/2?u=rokejulianlockhart), where a knowledgeable contributor posited:

> To me, the real question is whether having this capability can cause any harm (in which case it would make sense to petition Bitwarden to implement a block to such self-assigned emergency access). A cursory review of the [whitepaper](https://bitwarden.com/help/bitwarden-security-white-paper/#6AIDXvqstFQZi3TBw1xldg) suggests that it could potentially allow an attacker who gains access to (or is able to crack) the user’s RSA public key to then obtain the User Symmetric Key.

### Operating System

Linux

### Operating System Version

[`fedora-release-identity-kde-desktop-43`](https://fedora.mirrorservice.org/fedora/linux/updates/43/Everything/x86_64/Packages/f/fedora-release-identity-kde-desktop-43-26.noarch.rpm)

### Web Browser

Firefox

### Browser Version

[`firefox-nightly-150.0a1-20260303095635`](https://packages.mozilla.org/rpm/firefox/Packages/ec16585b0892a65d6a1ce7dd4dabcf1f9916fe7783e8465aa7e38236598b63e6-firefox-nightly-150.0a1-20260303095635.x86_64.rpm)

### Build Version

> `2026.2.1`

### Issue Tracking Info

- [x] I understand that work is tracked outside of Github. A PR will be linked to this issue should one be opened to address it, but Bitwarden doesn't use fields like "assigned", "milestone", or "project" to track progress.

pamperer562580892423:
> Build Version
> This does not appear to be accessible.

[...]

PR #19412, opened 2026-03-06
PM-33168 - Emergency Access - Add / Edit dialog - fix missing error toasts

JaredSnider-Bitwarden:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-33168
Server PR: https://github.com/bitwarden/server/pull/7165
Relates to https://github.com/bitwarden/clients/issues/19382

## 📔 Objective

Update the Emergency Access add / edit dialog to show errors via the `ValidationService` which will show the new server error for when a grantor tries to invite themselves + any existing possible errors (that previously would have only been logged). 

## 📸 Screenshots

https://github.com/user-attachments/assets/4b7c7d81-cc79-46d0-8269-a6db820e30f8


Hinton:
FYI if you don't catch the error the bitSubmit will call the `ValidationService` service for you. https://github.com/bitwarden/clients/blob/main/libs/components/src/async-actions/bit-submit.directive.ts#L49

JaredSnider-Bitwarden:
> FYI if you don't catch the error the bitSubmit will call the `ValidationService` service for you. https://github.com/bitwarden/clients/blob/main/libs/components/src/async-actions/bit-submit.directive.ts#L49

@Hinton, thank you! I'll just remove the catch instead here in a few.

apps/web/src/app/auth/settings/emergency-access/emergency-access-add-edit.component.ts
```diff
@@ -61,10 +61,10 @@ export class EmergencyAccessAddEditComponent implements OnInit {
     private formBuilder: FormBuilder,
     private emergencyAccessService: EmergencyAccessService,
     private i18nService: I18nService,
-    private platformUtilsService: PlatformUtilsService,
     private logService: LogService,
     private dialogRef: DialogRef<EmergencyAccessAddEditDialogResult>,
     private toastService: ToastService,
+    private validationService: ValidationService,
```
Hinton: This service is unused right?

JaredSnider-Bitwarden: Oof, thank you. Moving too quickly on that clean up. Fixed.

JaredSnider-Bitwarden:
Passed QA

PR #19414, opened 2026-03-06
[PM-33225] Remove shared module from desktop

Hinton:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-33225
## 📔 Objective

`SharedModule` is only consumed by `AppModule`. Remove it and reference the required imports there.


apps/desktop/src/app/app.module.ts
```diff
@@ -3,48 +3,45 @@ import "zone.js";
...
 import { SearchComponent } from "./layout/search/search.component";
-import { SharedModule } from "./shared/shared.module";
+import { ServicesModule } from "./services/services.module";
 
+/**
+ * This is the `AppModule` for the Bitwarden desktop application.
+ *
+ * This file contains **ONLY** components that are used in `AppComponent`. You most likely
+ * **DO NOT** want to modify this file. Routable components are handled by the `AppRoutingModule`.
+ */
```
Hinton, line 32: @willmartian I'm interested if there is anything else we should mention in the `AppModule`.

willmartian: I think this is fine!

PR #19418, opened 2026-03-06
[CL-1088] only apply macOS extra top padding to nav logo in overlay mode

willmartian:
## Tracking

https://bitwarden.atlassian.net/browse/CL-1088

## Summary

- The macOS extra top padding on `bit-nav-logo` was unconditionally applied via a static class, causing it to affect all side nav states
- Padding is now conditional: the full `calc(0.5rem + var(--bit-sidenav-macos-extra-top-padding))` only applies when the side nav is in overlay mode (`isOverlay()`), otherwise falls back to `tw-pt-2`
- Uses `ngClass` object syntax for clarity

## Screencast

https://github.com/user-attachments/assets/e3d67453-ca5a-4431-bc27-da8efe0fd122

## Test plan

- [ ] On macOS desktop, open the app and verify the nav logo has correct top padding when the side nav is in overlay mode (open, not push mode)
- [ ] Verify the nav logo has standard `0.5rem` top padding when the side nav is in push mode or closed
- [ ] Verify no visual regressions on non-macOS platforms

### DIFF
--- a/libs/components/src/navigation/nav-logo.component.html
+++ b/libs/components/src/navigation/nav-logo.component.html
@@ -1,6 +1,6 @@
 <div
   class="tw-px-2 tw-bg-bg-sidenav"
-  [class]="{
+  [ngClass]="{
     'tw-sticky tw-top-0 tw-z-50 tw-pb-4': sideNavService.open(),
     'tw-pb-[calc(theme(spacing.8)_+_2px)]': !sideNavService.open(),
     'tw-pt-[calc(0.5rem_+_var(--bit-sidenav-extra-top-padding,_0px))]': sideNavService.isOverlay(),
--- a/libs/components/src/navigation/nav-logo.component.ts
+++ b/libs/components/src/navigation/nav-logo.component.ts
@@ -1,3 +1,4 @@
+import { NgClass } from "@angular/common";
 import { ChangeDetectionStrategy, Component, input, inject } from "@angular/core";
 import { RouterLinkActive, RouterLink } from "@angular/router";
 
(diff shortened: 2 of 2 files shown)


libs/components/src/navigation/nav-logo.component.html
```diff
@@ -1,10 +1,12 @@
 <div
-  class="tw-px-2 tw-pt-[calc(0.5rem_+_var(--bit-sidenav-macos-extra-top-padding,_0px))] tw-bg-bg-sidenav"
-  [class]="
-    sideNavService.open()
-      ? 'tw-sticky tw-top-0 tw-z-50 tw-pb-4'
-      : 'tw-pb-[calc(theme(spacing.8)_+_2px)]'
-  "
+  class="tw-px-2 tw-bg-bg-sidenav"
+  [ngClass]="{
```
Hinton: ngClass 🤮

willmartian: [ed0766e](https://github.com/bitwarden/clients/pull/19418/commits/ed0766e5927418d9ca49308a45daa692326c9b7f)

willmartian:
> Would it be nicer to extract the classes to ts code?

Yes, but I would like to do that alongside our migration to tailwind-variants.

libs/components/src/navigation/nav-logo.component.html
willmartian: Reverted [2f3dd4f](https://github.com/bitwarden/clients/pull/19418/commits/2f3dd4fa8d5a0a48b85917ec6eaa6c40f55db61b). I was under the assumption that [class] now did everything that `[ngClass]` can do--that is not so. (Caught with UI Review). This will be removed when we move to tailwind-variants.

PR #19425, opened 2026-03-06
[PM-31438] Send unsaved edits dialog

mcamirault:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-31438

## 📔 Objective

This PR adds a pop-up that displays on the Send page whenever you have a Send open with unsaved edits and attempt an action that would close the Send: navigating to another vault page, opening a different Send, closing the Send edit drawer, etc. To implement all of these I added a new config value when creating a dialog called `closePredicate`, which takes an asynchronous function and is executed every time we attempt to close the dialog. If the function returns true then the dialog closes, if it returns false then it doesn't. To do this I also had to make the `close` and `openDrawer` methods of the `DialogService` async (which is what pulled in most of the other teams besides Tools here) and now return a boolean from the former indicating whether the close was successful or not.

*Some details, mostly intended for the Tools team*
In order to allow the base `SendComponent` to trigger the form actions it needs to using the new `closePredicate` I had to lift the Send form out of the `SendFormComponent` and into the `SendFormService`. This service was used before solely to encapsulate some Send decryption and API call logic and as such wasn't intended to be called outside of the `SendFormComponent`, but I've expanded its scope a bit. I felt this was a better pattern than calling a static method on the `SendAddEditDialogComponent` plus some JavaScript hackery involving passing Observables through the dialog component's data, which was my initial approach. Props to @willmartian for answering all my questions about the dialog and suggesting the service approach :)

## 📸 Screenshots

*Video demoing three of the actions that can trigger the pop-up: closing the dialog, opening an existing Send, creating a new Send. Also demos that the changes are discarded if the user clicks the appropriate button*

https://github.com/user-attachments/assets/ac54b539-1667-4e74-9430-d15f94d512f4


apps/web/src/app/tools/send/send.component.ts
```diff
@@ -53,19 +62,48 @@ const BroadcasterSubscriptionId = "SendComponent";
...
+    SpinnerComponent,
   ],
   templateUrl: "send.component.html",
   providers: [DefaultSendFormConfigService],
 })
 export class SendComponent extends BaseSendComponent implements OnInit, OnDestroy {
-  private sendItemDialogRef?: DialogRef<SendItemDialogResult> | undefined;
+  /**
+   * Prevent browser tab from closing/refreshing if the Send form has unsaved edits.
+   * Shows a confirmation dialog if user tries to leave during an active upload.
```
harr1424: ⛏️  "upload" might be confusing to future work in this file, since files can be uploaded, but I think here "upload" refers to Send creation or editing an existing Send.

apps/web/src/app/tools/guards/unsaved-send-edits.guard.ts
```diff
@@ -0,0 +1,7 @@
+import { CanDeactivateFn } from "@angular/router";
+
+import { SendComponent } from "../send/send.component";
+
+export const unsavedSendEditsGuard: CanDeactivateFn<SendComponent> = async (component) => {
```
harr1424:
```suggestion
export const UnsavedSendEditsGuard: CanDeactivateFn<SendComponent> = async (component) => {
```
I think this should be Pascal Case

harr1424 (dismissed):
Looks good! One very minor nit-pick regarding terminology used in a comment.

mzieniukbw (dismissed):
LGTM form KM

r-tome (dismissed):
Approving AC owned changes

nick-livefront (dismissed):
✅ Vault

Banrion (dismissed):
DIRT file looks good

apps/web/src/app/tools/send/send.component.ts
mcamirault: I copied this approach from elsewhere in the code and I missed this when changing the comment. Removed in my latest commit

nick-livefront (dismissed):

Patrick-Pimentel-Bitwarden (dismissed):

harr1424 (dismissed):
Tools changes look good

nick-livefront (dismissed):

sbrown-livefront (dismissed):
Billing looks good ✅

mzieniukbw (dismissed):

Banrion (dismissed):

neuronull (dismissed):
Autofill Desktop looks good

neuronull (dismissed):

mzieniukbw (dismissed):

pamperer562580892423:
A user's perspective: looks good (--> joke, but: one of the biggest mysteries to me is why it's so easy to loose unsaved data when using BW, so that's a good direction to go IMHO... sorry for the intrusion, though)

nick-livefront (dismissed):

mcamirault:
@pamperer562580892423 No worries about intruding, early user feedback is invaluable for new features like this 😁. It also made me realize that I hadn't updated demo vid in the initial PR comment to show the latest changes so thanks for that!

willmartian (changes requested):
Making `DialogService.close` async causes a ton of churn (as you can see). I also get a little antsy mechanically adding a bunch of `await`s everywhere, and `void` is an API smell imo. So, what do you think of this alternative?

Return a `DialogCloseRef` synchronously that contains the promise:

```ts
// something like this?
type DialogCloseRef = {
  closed: Promise<boolean>;
}
```

That would allow us to revert most of the diffs in this PR. Interested subscribers can await `.closed` as needed. The only breaking change remaining would be making `openDrawer` async, but that has a much smaller surface area.

edit:
~Actually, for programmatic closing, should we even check the predicate? The main point of the close predicate seems like it should be for user-initiated dismissals (X button, navigation away), not programmatic ones like switching drawers.~

~I think it would be weird to show a "You have unsaved changes..." dialog in the event that a secondary drawer is opened programmatically. Perhaps the better approach there would be to bypass the predicate and force close? I will loop design in on this and see if they have an opinion.~

edit x2:
After more thought, I think it is fine for programmatic closing to check the predicate. Being inconsistent is probably more confusing. And most of our programmatic calls to `.close` were originally triggered by some user action (typically a submit button click).

willmartian:
(Moving to draft while @mcamirault and I suss out my above comment, since the outcome will alter most cross-team diffs. 🙇 )

mcamirault:
@willmartian and I discussed this and came to the conclusion that the `.close` methods should remain async but their return type should be made more flexible to allow future changes.

libs/components/src/dialog/dialog.service.ts
```diff
@@ -79,7 +80,14 @@ export type DialogConfig<D = unknown, R = unknown> = Pick<
   | "width"
   | "restoreFocus"
   | "closeOnNavigation"
->;
+> & {
+  closePredicate?: (result?: any) => Promise<boolean>;
```
willmartian: I believe we can type this as `result: R`?

willmartian (dismissed):
Strong work, thank you. One comment from me below, but otherwise good to send to QA.

libs/components/src/dialog/dialog.service.ts
mcamirault: I believe we'll also have to change the types of the configs on 167 and 292 to `DialogConfig<unknown, R>` and `DialogConfig<D,R>` respectively (it complains about the type of the close predicate on 295 if we don't do the second one). The former was a mistake on my part in the new code so it doesn't matter but the latter is a change to the existing type that pulls in about 16 more changes. Not a big deal since the PR is so wide already and I think this is more correct, but worth pointing out.

willmartian (dismissed):

r-tome (dismissed):
AC changes look good

harr1424:
@claude please give this a review

Patrick-Pimentel-Bitwarden (dismissed):

harr1424 (dismissed):

sbrown-livefront (dismissed):
✅ Billing looks good

neuronull (dismissed):
Desktop Native changes LGTM

nick-livefront (dismissed):

willmartian (dismissed):

Banrion (dismissed):

cd-bitwarden (dismissed):
Everything looks good 👍

theMickster:
Hello @mcamirault et. al, 

A local multi-agent Claude Code review produced the following report (I'm field testing a forthcoming addition to the plugin marketplace). Kindly review at your convenience and lemme know if the findings are legitimate or not. If not, it would be beneficial to know briefly what context Claude may be missing or misunderstanding to make better assessments of our code.

<details>
<summary># Code Review: [PM-31438] Send unsaved edits dialog (#19425)</summary> 

**Date:** 2026-04-16 | **Reviewed by:** Claude Code

## Summary

| Severity      | Count |
| ------------- | ----- |
| 🛑 Blocker    | 0     |
| ⚠️ Important  | 3     |
| ♻️ Refactor   | 2     |
| 💡 Suggestion | 0     |

This PR makes a significant infrastructure change — making `DialogRef.close()` async and adding `closePredicate` support — to implement an "unsaved edits" confirmation dialog for the Send feature. The core architecture is sound, and the ~100+ mechanical `await`/`void` updates are largely correct. However, there are three important bugs: a `void` operator discarding a drawer reference in the Access Intelligence page, a null dereference when the unsaved-edits confirmation dialog is dismissed via ESC, and a false-positive dirty check that always triggers the confirmation in "add" mode.

## Findings

### ⚠️ Important

#### `void` on `openDrawer` stores `undefined` in signal — Access Intelligence drawer can never be programmatically closed

`bitwarden_license/bit-web/src/app/dirt/access-intelligence/v2/access-intelligence-page/access-intelligence-page.component.ts:270`

  In the `setupDrawerSubscription` method, the PR changes:
  ```typescript
  this.currentDialogRef.set(
    void this.dialogService.openDrawer(AccessIntelligenceDrawerV2Component, {
      data: content,
    }),
  );
  ```

  The `void` operator always evaluates to `undefined`, so `this.currentDialogRef.set(undefined)` is always the result. All subsequent `this.currentDialogRef()?.close()` calls (in `ngOnDestroy`, `changeTab`, `handleCloseDrawer`, etc.) become no-ops because the signal value is always `undefined`. The drawer opens but the component loses its reference to it.

  TypeScript doesn't catch this because the signal type includes `| undefined`.

  Contrast with `RiskInsightsComponent` in the same PR, which correctly uses `switchMap(async ...)` to properly `await` and store the drawer reference.

  **Fix:** Convert the `.subscribe()` callback to async and use `await`, or use `.then()` to set the signal after the promise resolves.
  </details>

#### Null dereference in `promptForUnsavedEdits` when ESC/backdrop dismisses the unsaved edits dialog — causes fail-open that silently discards edits

`libs/tools/send/send-ui/src/send-form/services/default-send-form.service.ts:147`

  The chain of events:
  1. User has unsaved edits in a Send form drawer
  2. User tries to close the drawer; `closePredicate` calls `promptForUnsavedEdits()`
  3. The "Discard edits?" dialog opens (`UnsavedEditsDialogComponent`) — no `disableClose` set
  4. User presses ESC or clicks backdrop
  5. CDK dialog internally closes, emitting `undefined` on the `closed` observable
  6. `lastValueFrom(dialogRef.closed)` resolves to `undefined`
  7. `result.result` throws `TypeError: Cannot read properties of undefined`
  8. Error propagates back to `DrawerDialogRef.close()` where the closePredicate catch block logs it and falls through
  9. The drawer closes, silently discarding the user's unsaved edits

  This defeats the entire purpose of the unsaved edits feature — the guard itself has an escape hatch that allows data loss.

  **Fix:** Add optional chaining: `if (result?.result === UnsavedEditsDialogResult.Discard)`. When `result` is `undefined` (ESC/backdrop), the method returns `false`, keeping the drawer open. Alternatively, open the `UnsavedEditsDialogComponent` with `disableClose: true`.
  </details>

#### `sendFormHasEdits()` always returns `true` in "add" mode — false-positive "Discard edits?" prompt on untouched forms

`libs/tools/send/send-ui/src/send-form/services/default-send-form.service.ts:~100`

  In `initializeSendForm` for "add" mode:
  - `originalSendView` is set to `null`
  - `updatedSendView` is set to `new SendView()`

  In `sendFormHasEdits()`:
  ```typescript
  return (
    JSON.stringify(this.originalSendView, replacer) !==
    JSON.stringify(this.updatedSendView, replacer)
  );
  ```

  `JSON.stringify(null)` produces `"null"`, while `JSON.stringify(new SendView())` produces a full object string. These will never be equal. Additionally, child component `valueChanges` subscriptions call `patchSend` during initialization, further modifying `updatedSendView`.

  This means opening a new Send form and immediately closing it (without any user input) will always show the "Discard edits?" confirmation dialog. The `window:beforeunload` handler also fires incorrectly.

  **Fix:** Initialize `originalSendView` to a snapshot of the initial `updatedSendView` (after child forms have patched their defaults) instead of `null`, or handle the null case explicitly in `sendFormHasEdits()`.
  </details>

### ♻️ Refactor

#### Incorrect JSDoc comments on `UnsavedEditsDialogResult` — copy-paste from `SendItemDialogResult`

`libs/tools/send/send-ui/src/send-form/components/unsaved-edits-dialog/unsaved-edits-dialog.component.ts:8-11`

  ```typescript
  export const UnsavedEditsDialogResult = Object.freeze({
    /** The send item was created or updated. */  // Wrong — should be "discard unsaved edits"
    Discard: "discard",
    /** The send item was deleted. */              // Wrong — should be "go back to editing"
    Back: "back",
  } as const);
  ```

  The JSDoc comments are verbatim copies from `SendItemDialogResult` (`Saved: "saved"` / `Deleted: "deleted"`) and describe the wrong semantics for this dialog.

  **Fix:** Update to `/** The user chose to discard unsaved edits. */` and `/** The user chose to go back to editing. */`.
  </details>

#### Duplicate imports in web `SendComponent` — 5 imports appear twice

`apps/web/src/app/tools/send/send.component.ts:60-84`

  The PR adds a block of new imports at the top of the `imports` array without removing the existing entries below. Five imports appear twice: `FormsModule`, `AsyncPipe`, `I18nPipe`, `CalloutComponent`, and `IconComponent`.

  Angular handles duplicate imports gracefully at runtime (idempotent), so there is no functional impact. However, it's clearly unintentional and adds noise.

  **Fix:** Remove the 5 duplicate entries, keeping each import only once.
  </details>

## Reviewed and Dismissed

   <details><summary>🔍 2 initial findings dismissed after validation</summary>

   #### `@HostListener` on arrow function property will silently not work
   `apps/web/src/app/tools/send/send.component.ts:~91`
   **Dismissed because:** Angular's Ivy compiler generates host binding code that accesses the method via `ctx.handleBeforeUnloadEvent()` (instance access, not prototype). Arrow function class properties are defined on the instance during construction, which happens before any event fires. Additionally, this exact pattern (`@HostListener` on an arrow function property) already exists in production code at `libs/vault/src/cipher-view/attachments/attachments-v2.component.ts`.

   #### `closePredicate` error handler defaults to "allow close" (fail-open behavior)
   `libs/components/src/dialog/dialog.service.ts:~170`
   **Dismissed because:** The `closePredicate` is a UX convenience feature for preventing accidental loss of form edits — it is not a security mechanism. Fail-open is the correct default for UI dialogs: trapping a user in a dialog they can't close due to an internal error is a worse outcome than losing unsaved text edits. The error is properly logged. No vault data, encryption keys, or security invariants are affected.

   </details>

</details>

[PM-31438]: https://bitwarden.atlassian.net/browse/PM-31438?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

mzieniukbw (dismissed):

mcamirault:
Hey @theMickster, those all look like legitimate findings to me. I've fixed them along with a merge conflict that just popped up

nick-livefront (approved):

theMickster:
> Hey @theMickster, those all look like legitimate findings to me. I've fixed them along with a merge conflict that just popped up

Thank you, sir! I will note them as 'wins' for the multi-agent review field test.

r-tome (approved):

cd-bitwarden (approved):
nice work!

Patrick-Pimentel-Bitwarden (approved):

harr1424 (approved):

Banrion (approved):

willmartian (approved):

mzieniukbw (approved):

neuronull (approved):

cyprain-okeke (approved):
LGTM

PR #19426, opened 2026-03-06
Add fix-angular-fixmes skill to resolve Angular FIXME migration comments

JaredScar:
This new skill provides a structured approach to address recurring FIXME comments related to Angular modernization, specifically for CL-764 (Migrate to OnPush) and CL-903 (Migrate to Signals). It includes detailed steps for discovering FIXMEs, implementing necessary changes, and validating the updates. The skill emphasizes the use of Angular CLI for migrations and outlines a cleanup checklist to ensure all related comments and imports are properly managed.


Hinton:
This skill seems to have significant overlap with the Angular modernization skill. What differentiates it from the existing skill?

.claude/skills/fix-angular-fixmes/SKILL.md
```diff
@@ -0,0 +1,325 @@
+---
+name: fix-angular-fixmes
+description: Resolves eslint-disable suppression comments throughout the Bitwarden clients codebase by fixing the underlying issue. Use when the user asks to "fix FIXMEs", "fix eslint suppressions", "clean up eslint-disable-next-line", "resolve CL-764", "resolve CL-903", "migrate to OnPush", "migrate to Signals", or reduce linting suppressions.
+allowed-tools: Read, Write, StrReplace, Glob, Grep, Bash(npx ng generate:*), Bash(npm run lint:fix), Bash(npm run test)
+---
+
+# Fix ESLint Suppressions
+
+Resolves `eslint-disable-next-line` comments by fixing the underlying issue. Every suppression is technical debt — the goal is to fix the code and remove the comment, not just delete the suppression.
```
theMickster: ⚠️ This is nearly exactly the same as the frontmatter. Let's remove it because this won't help Claude (it helps humans but that's not the audience).

JaredScar:
> This skill seems to have significant overlap with the Angular modernization skill. What differentiates it from the existing skill?

Good catch! @Hinton I've now fixed up the PR to point to that skill to understand the best way to modernize the codebase. The `fix-angular-fixmes` skill now focuses on targeting the suppressions we have across the codebase to prevent ESLint from crying...

.claude/skills/fix-angular-fixmes/SKILL.md
theMickster, line 1: ⚠️ In all the places where we have Before & After code blocks, I recommend that those are two distinct code blocks with bold **Before** and **After** descriptions. I am not so sure that a single inline code comment noting before and after will suffice.

theMickster, line 1: ⚠️ Please take out all horizontal lines `---`. 
Makes this more human readable, but this is for Claude Code and not humans.

.claude/skills/fix-angular-fixmes/SKILL.md
```diff
@@ -0,0 +1,325 @@
...
+
+```bash
+npm run lint:fix
+```
+
+Fix any errors that remain. Run `npm run test` if behaviour-critical code was changed.
+
+---
+
+## Key rules
```
theMickster: ⚠️ Please move these to the top of the `SKILL.md` because we want Claude to read these first before making decisions on the work to do. 
Also, given how large the `SKILL.md` already is, I think Claude will just ignore these altogether if they are left at the bottom of the file.

theMickster (commented):
♻️ I have concerns about the usability and maintainability of the `SKILL.md` itself because there's a lot of individual bits and parts that aren't always coupled in code but are now coupled in a big skill.  

I wonder if these individual fixmes are better written as individual markdown files in an example folder that are then referenced in the skill. That would allow for better progressive disclosure by Claude only loading examples when required. 🤔

.claude/skills/fix-angular-fixmes/SKILL.md
```diff
@@ -0,0 +1,325 @@
+---
+name: fix-angular-fixmes
+description: Resolves eslint-disable suppression comments throughout the Bitwarden clients codebase by fixing the underlying issue. Use when the user asks to "fix FIXMEs", "fix eslint suppressions", "clean up eslint-disable-next-line", "resolve CL-764", "resolve CL-903", "migrate to OnPush", "migrate to Signals", or reduce linting suppressions.
```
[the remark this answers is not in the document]
theMickster: I agree that the frontmatter needs some trimming.

theMickster (dismissed):
Let's give this a run for it's money over the next few sprints iterating quickly if we find that Claude Code is making any mistakes while using it.

theMickster (approved):

PR #19427, opened 2026-03-06
[BEEEP] Desktop Native: napi type macro helpers

neuronull:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-33264

## 📔 Objective

note: macros can be controversial and there are tradeoffs to adding them. I believe I've made these documented and tested and that the usage is fairly straight forward. It was a BEEEP project I wanted to explore and I'm open to criticism of it.

Existing pattern: define a type in the napi module for a given desktop native crate’s correlate, and the type is a 1-1 mapping of the crate’s type, but with the napi macros on it.

The “problem”: either a boilerplate From is needs to be implemented, or just copying the struct fields in-line.

We want to preserve the boundary on napi, i.e., not creating a dependency on the source crates to napi.

In order to achieve this, I propose the introduction of a couple macros that are highly test-able and documented, and still make the callsite readable, and clear what is going on.

### Details:

Each type that needs to cross the Electron boundary is re-declared in the napi crate using napi_mirrors!, which generates both the #[napi(object/string_enum)]-annotated type and a From<SourceType> impl. The source crates (ssh_agent, chromium_importer, etc.) are untouched.

The macros eliminate the boilerplate of re-declaring types and writing From impls by hand, while keeping the code readable, editor-navigable, and free of cross-crate coupling.
  
Alternatives considered:
- Alternative 1: annotate types in the source crates directly.
    - Would require the source crates to depend on napi-rs. Those crates are pure Rust business logic and shouldn't know anything about Electron or the binding layer. Adding napi attributes there would permanently couple core logic to an Electron-specific dependency, breaking any future non-Electron consumer of those crates.

- Alternative 2: build.rs code generation.
    - A build script could parse source type definitions and emit the napi wrapper code. This works but adds meaningful build-time complexity: you need to parse Rust source (fragile against formatting/attribute changes), the generated output is invisible to the editor and harder to debug, and errors surface at build time with poor diagnostics. The macro approach keeps everything visible in normal source files with standard compiler errors.

### DIFF
--- a/apps/desktop/desktop_native/napi/index.d.ts
+++ b/apps/desktop/desktop_native/napi/index.d.ts
@@ -297,24 +297,20 @@ export declare namespace sshagent_v2 {
     unlock(): void
   }
   export type SSHAgentState = SshAgentState
-  /** SSH public key data */
   export interface PublicKey {
     alg: string
     blob: Array<number>
   }
-  /** SSH sign request fields. */
   export interface SignRequest {
     publicKey: PublicKey
     processName?: string
     isForwarding: boolean
     namespace?: SignRequestNamespace
   }
-  /** Data for a sign request, including vault cipher context. */
   export interface SignRequestData {
     signRequest: SignRequest
     cipherId?: string
   }
-  /** Namespace of a sign request. */
   export const enum SignRequestNamespace {
     Git = 'Git',
     File = 'File',
--- a/apps/desktop/desktop_native/napi/src/chromium_importer.rs
+++ b/apps/desktop/desktop_native/napi/src/chromium_importer.rs
@@ -2,12 +2,9 @@
 pub mod chromium_importer {
     use std::collections::HashMap;
 
-    use chromium_importer::{
-        chromium::{
-            DefaultInstalledBrowserRetriever, LoginImportResult as _LoginImportResult,
-            ProfileInfo as _ProfileInfo,
-        },
-        metadata::NativeImporterMetadata as _NativeImporterMetadata,
+    use chromium_importer::chromium::{
+        DefaultInstalledBrowserRetriever, LoginImportResult as _LoginImportResult,
+        ProfileInfo as _ProfileInfo,
     };
 
     #[napi(object)]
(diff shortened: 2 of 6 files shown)


Hinton:
> We want to preserve the boundary on napi, i.e., not creating a dependency on the source crates to napi.

Why? Can you elaborate on which benefits it provides. In https://github.com/bitwarden/clients/pull/18787#issuecomment-3968135720 we proposed removing the wrapped types instead.

neuronull:
> > We want to preserve the boundary on napi, i.e., not creating a dependency on the source crates to napi.
> 
> Why? Can you elaborate on which benefits it provides. In [#18787 (comment)](https://github.com/bitwarden/clients/pull/18787#issuecomment-3968135720) we proposed removing the wrapped types instead.

To be clear- I'm not advocating that the wrapper types are helpful to us in any specific way. I'm only advocating that the de-coupling of the DN source code to napi is something worth preserving. In this view, the wrapper types are a necessary side effect of that.

Currently the logic in desktop native is only accessed through napi for Electron. And currently there isn't a dependency of the source code in DN on napi. This means that the source code in DN could be leveraged with other transport mechanisms in the future (whether in the monorepo or not ...). By adding coupling the source code in DN with napi, we tie ourselves to that. It's more rigid.  

Granted, in this particular PR, the alternative I mentioned about having the macro callsites live on the source types in DN, would not be overtly invasive and could feasibly be undone in a simple PR. An example of what it would look like can be found below.

This PR can be modified to take that approach, and it would simplify the re-specification of the types in the napi modules we have. While it does technically add that coupling to napi, it is easy enough to remove so could be a reasonable tradeoff to accept for the benefit of not having to repeat the types.

```Rust

// chromium_importer/src/metadata.rs

#[napi_mirrors]
pub struct NativeImporterMetadata {
// ...
}
```

Hinton:
> Currently the logic in desktop native is only accessed through napi for Electron. And currently there isn't a dependency of the source code in DN on napi. This means that the source code in DN could be leveraged with other transport mechanisms in the future (whether in the monorepo or not ...). By adding coupling the source code in DN with napi, we tie ourselves to that. It's more rigid.

SDK-internal supports both WASM and mobile bindings. I don't see how adding an optional feature flagged dependency of internal libraries to napi ties outself to that.

neuronull:
>> "By adding coupling the source code in DN with napi, we tie ourselves to that. It's more rigid"

> SDK-internal supports both WASM and mobile bindings. I don't see how adding an optional feature flagged dependency of internal libraries to napi ties outself to that.

My statement was in reference to the "Alternative 1" posed in the PR description of this PR.

Hinton:
I'm still not following why the decoupling is worth preserving.

neuronull:
👋 we talked out of band on this one. I'd had this idea for moving the duplication of the napi structs to macros to make the code more readable while still preserving a separation between the napi crate from the source ones, for some time and had been eager to try it out on some beeep time.

While #18787 is concerned with a somewhat targeted problem with the passkeys work,it  sounds like the approaches discussed in #18787 are anticipated to reach some conclusions on a path forward to mimic the approaches and lessons learned from the sdk-internal.

Closing this PR to denote it as an educational exercise.

PR #19433, opened 2026-03-08
[PM-31112] Decouple local tools crypto from user key

mzieniukbw:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-31112
https://bitwarden.atlassian.net/browse/PM-31055
Requires https://github.com/bitwarden/sdk-internal/pull/829

## 📔 Objective

Decouples local tools crypto from direct use of user key into an indirect local user data key. Used by password generator history and email forwarders encryption.
This will allow for user key rotation without logout, since we just have to re-encrypt the local user data key with the new user key, without re-encrypting all other data.
The local user data key state is not exposed in SDK. For backwards compatibility the local user data key is user key wrapped user key - this allows us to use it as drop in replacement
The SDK's CryptoClient exposes two new WASM stop-gap functions for encryption with the local user data key - intended to be only used by tools at this moment.


libs/tools/generator/extensions/history/src/legacy-password-history-decryptor.ts
```diff
@@ -1,36 +1,34 @@
...
-      const encrypted = new EncString(item.password);
-      const decrypted = await this.encryptService.decryptString(encrypted, key);
+      const decrypted = await firstValueFrom(
+        this.sdkService.userClient$(this.userId).pipe(
+          map((sdk) => {
+            if (!sdk) {
+              throw new Error("SDK not available");
+            }
+            using ref = sdk.take();
+            return ref.value.crypto().decrypt_with_local_user_data_key(item.password);
```
quexten, line 25: Just a thought: I wonder whether we should make this interface more narrowly scoped to take a password history item? I don't know if we just want arbitrary data encrypted / decrypted by people using this new public interface?

quexten (dismissed):

libs/tools/generator/extensions/history/src/legacy-password-history-decryptor.ts
mzieniukbw: i abandoned that idea, as soon as i saw the codebase for tools. I would leave it up to tools discretion to do so, it's too much work.

quexten: Fair point. Let's monitor for now, if people do use this for new use-cases, we can re-evaluate.

quexten (dismissed):

mcamirault (dismissed):

mzieniukbw:
Looks like we can't use sdk-managed state. Need to make change to SDK first.

dani-garcia (dismissed):
Platform changes lgtm

mcamirault (dismissed):
Tools changes still look good

quexten (dismissed):

quexten (approved):

dani-garcia (approved):
No platform changes anymore, but approved anyway

mcamirault (approved):

PR #19436, opened 2026-03-08
[PM-33279] Clear SDK-managed state keys on logout

mzieniukbw:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-33279
Requires https://github.com/bitwarden/sdk-internal/pull/825

## 📔 Objective

The `WrappedLocalUserDataKey` state, which is SDK-managed repository, should be cleared on logout.

This is because on user key rotation (which logs out) and re-login, the SDK will fail to unwrap the key using the user key.


mzieniukbw:
no longer needed with client managed state

PR #19447, opened 2026-03-09
Proposal to not wrap napi types

Hinton:
## 📔 Objective

Now that there is no requirement for napi exports to all live in the same crate we can clean up and remove most wrapping types. The remaining types differs from the originals in some way be it intentional or unintentional and were not migrated in this effort.


iinuwa:
For context, I believe this was spawned from discussion on #18787.

PR #19336, opened 2026-03-03
[PM-28488] Move event logging code to DIRT folders

eliykat:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-28488

## 📔 Objective

Move event logging code to DIRT ownership.

This had a large impact on import statements, so I have re-exported all files from their old locations to avoid impacting every team in this PR. I will submit follow-up PRs to update each team's files. The new location uses barrel files which should ease this a bit in the future.

The main changes in this PR are split into 2 commits for ease of review:
- the first commit moves files only - so you can verify nothing substantive has changed
- the second commit re-exports the new files from their old locations

The following commits make minor changes only.

PR #19450, opened 2026-03-09
[PM-31475] Remove decline event log for org data ownership policy

JaredScar:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-31475

## 📔 Objective

* Remove the following server-side EventType: OrganizationUser_SelfRevoked = 1518, // User self-revoked due to declining organization data ownership policy
* Remove the call for the server-side Self-Revoke endpoint to generate an event log (it is handled client-side)
* Update associated test cases to remove the expectation of generating an event log

### DIFF
--- a/apps/web/src/app/dirt/event-logs/services/event.service.ts
+++ b/apps/web/src/app/dirt/event-logs/services/event.service.ts
@@ -362,6 +362,13 @@ export class EventService {
           this.getShortId(ev.organizationUserId),
         );
         break;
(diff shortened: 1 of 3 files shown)


djsmith85:
@JaredScar  Some conflicts due to https://github.com/bitwarden/clients/pull/19336 merging and I think once this is updated, the review from @bitwarden/team-platform-dev won't be necessary anymore.

PR #19452, opened 2026-03-09
[PM-31646] stop label collection from crossing into sibling form-field containers

audreyality:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-31646

## 📔 Objective

Adds a containsChildFormElement guard to left-label, right-label, and
recursive sibling traversal so that text inside a sibling DOM container
that holds its own input/select/textarea is not absorbed as label text
for the target field.  This prevents Yahoo's country-code dropdown
container ("Enter Country Code") from contaminating the username field's
label-left, which caused a false TOTP classification via the word "code".

See also: #19453

## 🎥 Demo

| Before | After |
|--------|-------|
| <video src="https://github.com/user-attachments/assets/e355e794-7fd7-4021-b352-9c02dc5aee52" width="250" /> | <video src="https://github.com/user-attachments/assets/da996c08-948b-4916-b7d3-8242a4fa9042" width="250" /> |


apps/browser/src/autofill/services/collect-autofill-content.service.ts
```diff
@@ -710,6 +714,15 @@ export class CollectAutofillContentService implements CollectAutofillContentServ
       : null;
   }
 
+  /**
+   * Checks whether an element contains child form fields (input, select, textarea).
+   * Used to prevent label collection from absorbing text from sibling DOM
+   * containers that hold their own form fields.
+   */
+  private containsChildFormElement(element: Node): boolean {
```
jprusik: Non-blocking nit; you might consider dropping "form" from this method name as it's a bit overloaded.

```suggestion
  private containsFieldInput(element: Node): boolean {
```

apps/browser/src/autofill/services/collect-autofill-content.service.ts
```diff
@@ -710,6 +714,15 @@ export class CollectAutofillContentService implements CollectAutofillContentServ
...
+    return nodeIsElement(element) && !!element.querySelector("input, select, textarea");
```
jprusik: Similar to `ExcludedInlineMenuTypes` in `apps/browser/src/autofill/services/autofill-constants.ts` I wonder if it makes sense to codify the relationship of these input types somewhere (if we haven't already).

audreyality: If I understand you correctly, you mean extract a constant?

audreyality: I assumed I understood you properly. Addressed by [e08ba2e](https://github.com/bitwarden/clients/pull/19452/commits/e08ba2e1a94f87d71c13d79e7f57e824a2f10c86).

apps/browser/src/autofill/services/collect-autofill-content.service.ts
audreyality: Addressed by [e08ba2e](https://github.com/bitwarden/clients/pull/19452/commits/e08ba2e1a94f87d71c13d79e7f57e824a2f10c86).

apps/browser/src/autofill/services/autofill-constants.ts
```diff
@@ -98,6 +98,8 @@ export class AutoFillConstants {
     ...AutoFillConstants.ExcludedAutofillTypes,
   ];
 
+  static readonly FieldTypes: string[] = ["input", "select", "textarea"];
```
jprusik: At the risk of miring us in pedantry; what is the relationship between these values? There are other inputs, why are we not adding them? Being explicit about the nature of that relationship helps us avoid improper consumption or extension of these values in the future.

That is, I presume these are "Fields we care about for inline menu and/or autofill" (we don't want to include, say `radio` or `color`)

audreyality: Because those are types of `input`. A comment would probably help here.

```html
<input type="radio" />
<input type="color" />
```

audreyality: Addressed by 1c1eb586b9e1b0e03729326441f35da0c3d375a9.

jprusik: > Because those are types of input

lol right; I'm jumping ahead to input type. Thanks

PR #19453, opened 2026-03-09
[PM-31646] prioritize usernames ahead of ambiguous TOTPs

audreyality:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-31646

## 📔 Objective

Split `isFillableTotpField` into `isFillableTotpField` (`TotpFieldNames` /
one-time-code autocomplete) and `maybeFillableTotpField` (AmbiguousTotpFieldNames
only) and reorder the classification switch so that a reliable TOTP signal
still wins unconditionally, but a username match beats a purely ambiguous TOTP
match.  This fixes Yahoo's login-username field being misclassified as TOTP
because 'code' appears in `AmbiguousTotpFieldNames`.

See also: #19452

## 🎥 Demo

| Before | After |
|--------|-------|
| <video src="https://github.com/user-attachments/assets/e355e794-7fd7-4021-b352-9c02dc5aee52" width="250" /> | <video src="https://github.com/user-attachments/assets/da996c08-948b-4916-b7d3-8242a4fa9042" width="250" /> |


apps/browser/src/autofill/services/autofill.service.ts
```diff
@@ -1035,29 +1035,36 @@ export default class AutofillService implements AutofillServiceInterface {
           return;
         }
 
-        const isFillableTotpField =
+        const isTotpEligible =
```
jprusik: nit; feel free to disregard if you don't prefer

```suggestion
        const isTotpCandidate =
```

apps/browser/src/autofill/services/autofill.service.ts
```diff
@@ -1035,29 +1035,36 @@ export default class AutofillService implements AutofillServiceInterface {
...
-          ]) ||
-            field.autoCompleteType === "one-time-code") &&
           !AutofillService.fieldIsFuzzyMatch(field, [...AutoFillConstants.RecoveryCodeFieldNames]);
 
+        const isFillableTotpField =
+          isTotpEligible &&
+          (AutofillService.fieldIsFuzzyMatch(field, AutoFillConstants.TotpFieldNames) ||
+            field.autoCompleteType === "one-time-code");
+
+        const maybeFillableTotpField =
```
jprusik: Nit; I think "Fillable" is the thing causing me to read this differently (is maybe a totp, not maybe fillable). In any case, similar thinking here around naming, but again, feel free to discard.

```suggestion
        const isTotpFieldPossibly =
```

apps/browser/src/autofill/services/autofill.service.ts
audreyality: Addressed by 1d2b1605dcc8c319c459f5aa402e76334b9a2b4d.

apps/browser/src/autofill/services/autofill.service.ts
audreyality: Addressed by 1d2b1605dcc8c319c459f5aa402e76334b9a2b4d. Rather than take your suggestion, I simplified all of the names.

PR #19454, opened 2026-03-09
[PM-32160] Use shared Send List component in web

mcamirault:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-32160

My first attempt at this was reviewed in [this PR](https://github.com/bitwarden/clients/pull/19258); it was simpler for me to start fresh on a new branch than to try and remove the changes I didn't need on the old branch.

## 📔 Objective

This PR makes the web client use the recently-created `SendListComponent` that desktop is currently using. To keep the change visually the same when the Send UI Refresh feature flag is off I had to add an input to the Send List to allow us to suppress the search bar, but we can remove this change at the same time we remove the feature flag and the old Send UI. I also cleaned up some imports, old code syntax, and outdated icon components while I was here.

## 📸 Screenshots

Should be the same as existing.

### DIFF
--- a/libs/vault/src/cipher-form/components/cipher-form.component.spec.ts
+++ b/libs/vault/src/cipher-form/components/cipher-form.component.spec.ts
@@ -10,6 +10,7 @@ import { ConfigService } from "@bitwarden/common/platform/abstractions/config/co
 import { I18nService } from "@bitwarden/common/platform/abstractions/i18n.service";
 import { CipherArchiveService } from "@bitwarden/common/vault/abstractions/cipher-archive.service";
 import { Cipher } from "@bitwarden/common/vault/models/domain/cipher";
+import { AttachmentView } from "@bitwarden/common/vault/models/view/attachment.view";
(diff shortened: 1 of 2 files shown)


apps/web/src/app/tools/send/send.component.ts
```diff
@@ -256,4 +225,73 @@ export class SendComponent extends BaseSendComponent implements OnInit, OnDestro
...
+    try {
+      await this.sendApiService.removePassword(send.id);
+      this.toastService.showToast({
+        variant: "success",
+        title: null,
+        message: this.i18nService.t("removedPassword"),
+      });
+    } catch (e) {
+      this.logService.error(e);
+    }
```
Hinton, line 282: Should this show an error toast if it fails? You have the `ValidationService.showError` which do this with toasts.

mcamirault: Added this for both `removePassword` and `deleteSend`

PR #19455, opened 2026-03-09
[CL-1082] card form controls

BryanCunningham:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/CL-1082

## 📔 Objective

- Minor style updates for checkbox, switch and radio inputs
- Updates Switch usage to use `bit-form-control` as a wrapper to match other controls implementation patterns
- Adds new `bit-form-control-card` version of checkbox, switch and radio buttons

#### NOTE:
@bitwarden/team-vault-dev @bitwarden/team-auth-dev The changes in your code are to update switch component usage. We Changed the component usage to be in line with how other form control components are used by consumers. 

#### Not enabled
image

#### Enabled
image

### Radio
image

#### Unchecked
image

#### Checked
image


libs/components/src/form-control/form-control-card.component.ts
```diff
@@ -0,0 +1,81 @@
...
+      const inputId = this.base.inputId();
+      const hasError = this.base.formControl().hasError;
+
+      const describedBy = hasError ? this.errorId : (this.hint()?.id ?? null);
+
+      if (this.switch()) {
+        // For SwitchComponent, use signals to set ARIA directly on the inner input,
+        // avoiding a querySelector race with Angular's property binding rendering cycle
+        this.switch().ariaLabelledBy.set(this.labelId);
+        this.switch().ariaDescribedBy.set(describedBy ?? undefined);
```
vleague2: Seeing typechecking failures for these lines

libs/components/src/checkbox/checkbox.component.ts
```diff
@@ -26,90 +26,92 @@ export class CheckboxComponent implements BitFormControlAbstraction {
     "tw-inline-block",
     "tw-align-sub",
     "tw-flex-none", // Flexbox fix for bit-form-control
-    "!tw-p-1",
-    // Give checkbox explicit height and width to fix iOS rendering bug
```
vleague2, line 30: We should probably check this on iOS

libs/components/src/form-control/form-control-card-group.component.html
```diff
@@ -0,0 +1,21 @@
...
+@if (!label()) {
+  <ng-container *ngTemplateOutlet="cardStack"></ng-container>
+}
+
+<ng-template #cardStack>
+  <div class="tw-flex tw-flex-col tw-gap-2 [&_bit-form-control-card]:!tw-mb-0">
+    <ng-content></ng-content>
+  </div>
+  <div class="tw-mt-2">
+    <ng-content select="bit-hint" ngProjectAs="bit-hint"></ng-content>
```
vleague2: For my own education, why do we need to project it? Is this getting consumed or wrapped by another component?

libs/components/src/form-control/form-control-card.component.html
```diff
@@ -0,0 +1,35 @@
+<div
+  class="tw-flex tw-items-center tw-relative tw-p-4 tw-border tw-border-border-base tw-rounded-xl tw-justify-between tw-gap-4 [&:has(label:hover)]:tw-border-border-brand-soft [&:has(input:focus-visible)]:tw-ring-2 [&:has(input:focus-visible)]:tw-ring-border-focus [&:has(input:focus-visible)]:tw-border-transparent [&:has(input:checked)]:tw-bg-bg-brand-soft"
```
vleague2: You can use tailwind's `has-` selector in the current version we are on

libs/components/src/form-control/form-control-card.component.html
```diff
@@ -0,0 +1,35 @@
...
+        }
+      </span>
+      <ng-content select="bit-hint"></ng-content>
+    </span>
+  </div>
+  <ng-content></ng-content>
+</div>
+@if (base.hasError) {
+  <div [id]="errorId" class="tw-mt-1 tw-text-fg-danger tw-text-xs tw-ms-0.5">
+    <i class="bwi bwi-error"></i> {{ base.displayError }}
```
vleague2: Not `bit-icon`?

libs/components/src/form-control/form-control-card.component.html
```diff
@@ -0,0 +1,35 @@
...
+
+    <span
+      class="tw-inline-flex tw-flex-col"
+      [class.tw-text-muted]="base.formControl().disabled"
+      [class.tw-text-main]="!base.formControl().disabled"
+    >
+      <span [id]="labelId" bitTypography="body2" class="tw-leading-5 tw-font-medium">
+        <ng-content select="bit-label"></ng-content>
+        @if (base.required) {
+          <span class="tw-text-xs tw-font-normal"> ({{ "required" | i18n }})</span>
```
vleague2: Do we have a typography variant that matches this usage?

libs/components/src/form-control/form-control-card.stories.ts
```diff
@@ -0,0 +1,135 @@
...
+
+          <bit-form-control-card>
+            <input type="checkbox" bitCheckbox formControlName="featureB" />
+            <bit-label>Feature B</bit-label>
+          </bit-form-control-card>
+        </bit-form-control-card-group>
+      </form>
+    `,
+  }),
+};
```
vleague2: Is there a way to do a story with an error showing? And with a required field?

libs/components/src/form-control/form-control.component.html
```diff
@@ -1,27 +1,28 @@
 <label
-  class="tw-transition tw-items-start [&:has(input[type='checkbox'])]:tw-gap-1 [&:has(input[type='radio'])]:tw-gap-1.5 tw-select-none tw-mb-0 tw-inline-flex tw-rounded has-[:focus-visible]:tw-ring has-[:focus-visible]:tw-ring-primary-600"
-  [ngClass]="[formControl().disabled ? 'tw-cursor-auto' : 'tw-cursor-pointer']"
+  class="tw-transition tw-items-start tw-gap-2 tw-select-none tw-mb-0 tw-inline-flex tw-rounded has-[:focus-visible]:tw-ring has-[:focus-visible]:tw-ring-border-focus"
+  [ngClass]="[base.formControl().disabled ? 'tw-cursor-auto' : 'tw-cursor-pointer']"
```
vleague2: Could we try removing `ngClass` (this and other instances) while we're in here?

libs/components/src/form-control/form-control.component.html
```diff
@@ -1,27 +1,28 @@
...
         <ng-content select="bit-hint"></ng-content>
       </span>
     }
   </span>
 </label>
-@if (hasError) {
+@if (base.hasError) {
   <div class="tw-mt-1 tw-text-danger tw-text-xs tw-ms-0.5">
-    <i class="bwi bwi-error"></i> {{ displayError }}
+    <i class="bwi bwi-error"></i> {{ base.displayError }}
```
vleague2: Same q about `bit-icon`

libs/components/src/form-control/form-control.component.ts
```diff
@@ -1,60 +1,25 @@
...
 
-import { BitFormControlAbstraction } from "./form-control.abstraction";
+import { FormControlBaseDirective } from "./form-control-base.directive";
 
 // FIXME(https://bitwarden.atlassian.net/browse/CL-764): Migrate to OnPush
 // eslint-disable-next-line @angular-eslint/prefer-on-push-component-change-detection
 @Component({
   selector: "bit-form-control",
   templateUrl: "form-control.component.html",
+  hostDirectives: [
```
vleague2, line 15: Can't leave a comment on the relevant line but can we migrate to onpush for this component while we're in here?

libs/components/src/icon-tile/icon-tile.component.ts
```diff
@@ -23,7 +23,7 @@ const sizeStyles: Record<IconTileSize, { container: string[]; icon: string[] }>
     icon: ["tw-text-sm"],
   },
   default: {
-    container: ["tw-w-8", "tw-h-8"],
+    container: ["tw-w-9", "tw-h-9"],
```
vleague2, line 36: Is this going to cause conflicts with the icon tile style changes?

libs/components/src/radio-button/radio-button-base.directive.ts
```diff
@@ -0,0 +1,42 @@
...
+import { RadioGroupComponent } from "./radio-group.component";
+
+let nextId = 0;
+
+@Directive()
+export class RadioButtonBaseDirective {
+  private groupComponent = inject(RadioGroupComponent);
+
+  readonly id = input(`bit-radio-button-${nextId++}`);
+  readonly value = input<unknown>();
```
vleague2: Is this supposed to be typed as `unknown`?

libs/components/src/radio-button/radio-button-card.component.html
```diff
@@ -0,0 +1,14 @@
+<bit-form-control-card disableMargin>
```
vleague2: Are we actually doing built-in margin on `bit-form-control-card`? I'd like to not do built-in margin on new components if we can help it.

libs/components/src/radio-button/radio-button-card.component.ts
```diff
@@ -0,0 +1,34 @@
+import { Component, HostBinding, inject } from "@angular/core";
+
+import { FormControlCardComponent } from "../form-control/form-control-card.component";
+
+import { RadioButtonBaseDirective } from "./radio-button-base.directive";
+import { RadioInputComponent } from "./radio-input.component";
+
+// FIXME(https://bitwarden.atlassian.net/browse/CL-764): Migrate to OnPush
```
vleague2: Can we make this OnPush?

libs/components/src/tw-theme.css
```diff
@@ -133,13 +133,13 @@
...
-  --color-orange-800: #8a2c0d;
-  --color-orange-900: #70240b;
-  --color-orange-950: #441306;
+  --color-orange-400: #ff894a;
+  --color-orange-500: #fa6216;
+  --color-orange-600: #e04b00;
+  --color-orange-700: #b23300;
+  --color-orange-800: #8a2203;
+  --color-orange-900: #631c04;
+  --color-orange-950: #441600;
```
vleague2, line 144: There are a lot of color changes in here. I feel like we keep running into this and tweaking colors in multiple PRs. Are these actually the correct ones we need? Can we ask design what's going on?

libs/components/src/form-control/form-control-card.component.html
BryanCunningham: This should probably just go away. We're moving to using an `*` instead

libs/components/src/icon-tile/icon-tile.component.ts
BryanCunningham: It _shouldn't_ assuming the icon tile updates are the same as what the Figma shows it should be

libs/components/src/radio-button/radio-button-card.component.html
BryanCunningham: Yeah, I'll remove this.

libs/components/src/form-control/form-control-card-group.component.html
BryanCunningham: Hmm, we may not need this. Probably a copy/paste situation

libs/components/src/icon-tile/icon-tile.component.ts
vleague2: I realized icon tile is already in `main` so I pulled latest into the milestone 3 branch, if you want to pull to this branch

libs/components/src/radio-button/radio-button-base.directive.ts
BryanCunningham: @vleague2 This was already how this was typed. I just moved it to a new file. Was one of those things I wasn't sure if I wanted to open a can of worms to figure it out or not

libs/components/src/checkbox/checkbox.component.ts
BryanCunningham: It still has an explicit height and width. I just removed the comment

libs/components/src/form-control/form-control-group.component.ts
```diff
@@ -0,0 +1,180 @@
...
+
+@Component({
+  selector: "bit-form-control-group, bit-radio-group, bit-radio-button-group",
+  templateUrl: "form-control-group.component.html",
+  changeDetection: ChangeDetectionStrategy.OnPush,
+  imports: [NgTemplateOutlet, I18nPipe, IconComponent],
+  host: {
+    "(focusout)": "onFocusOut($event)",
+  },
+})
```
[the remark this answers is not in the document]
BryanCunningham: We are already using a semantic `fieldset` this role is unnecessary

libs/components/src/form-control/form-control-base.directive.ts
```diff
@@ -0,0 +1,95 @@
...
+  );
+
+  readonly formControl = contentChild.required(BitFormControlAbstraction);
+  readonly formControlEl = contentChild.required(BitFormControlAbstraction, { read: ElementRef });
+  readonly ngControl = contentChild(NgControl);
+
+  readonly inputId = signal(this.id);
+
+  @HostBinding("class") get classes() {
+    return ([] as string[])
```
vleague2: Do we need the typecast?

libs/components/src/form-control/form-control-card.component.html
```diff
@@ -0,0 +1,41 @@
+<div
+  class="tw-flex tw-items-center tw-relative tw-p-4 tw-border tw-border-border-base tw-rounded-xl tw-justify-between tw-gap-4 has-[label:hover]:tw-border-border-brand-soft has-[input:focus-visible]:tw-ring-2 has-[input:focus-visible]:tw-ring-border-focus has-[input:focus-visible]:tw-border-transparent has-[input:checked]:tw-bg-bg-brand-soft"
+  (change)="inGroup && groupItem.notifyChange()"
+>
+  <label
+    [for]="base.inputId()"
+    class="tw-absolute tw-top-0 tw-left-0 tw-size-full tw-cursor-pointer"
+  ></label>
```
vleague2, line 10: Empty label?

libs/components/src/form-control/form-control-card.component.spec.ts
```diff
@@ -0,0 +1,162 @@
...
+
+import { CheckboxComponent } from "../checkbox/checkbox.component";
+import { I18nMockService } from "../utils/i18n-mock.service";
+
+import { FormControlCardComponent } from "./form-control-card.component";
+import { FormControlModule } from "./form-control.module";
+
+describe("FormControlCardComponent - aria-describedby reactivity", () => {
+  // TODO: Fix this the next time the file is edited.
+  // eslint-disable-next-line @angular-eslint/prefer-on-push-component-change-detection
```
vleague2: Can we get these TODOs fixed so we aren't adding new debt items?

libs/components/src/form-control/form-control-card.component.ts
```diff
@@ -0,0 +1,140 @@
...
+    this.base.disableMarginSignal.set(true);
+
+    effect(() => {
+      const hostEl = this.base.formControlEl().nativeElement;
+      const inputId = this.base.inputId();
+      const hasError = this.hasErrorSignal();
+
+      const describedBy = hasError ? this.errorId : (this.hint()?.id ?? null);
+
+      if (this.switch()) {
```
vleague2: If we store `this.switch()` as a const, you can avoid the `?` on the subsequent references

libs/components/src/form-control/form-control-card.component.ts
```diff
@@ -0,0 +1,140 @@
...
+        // For SwitchComponent, use signals to set ARIA directly on the inner input,
+        // avoiding a querySelector race with Angular's property binding rendering cycle
+        this.switch()?.ariaLabelledBy.set(this.labelId);
+        this.switch()?.ariaDescribedBy.set(describedBy ?? undefined);
+        hostEl.removeAttribute("aria-labelledby");
+        hostEl.removeAttribute("aria-describedby");
+      } else {
+        // For other controls (e.g. checkbox), the host element is the input itself
+        const inputEl = hostEl.id !== inputId ? hostEl.querySelector(`[id="${inputId}"]`) : null;
```
vleague2: Is there a way to get this element via Angular instead of a query selector? If so, we could also reference it in the effect on line 122

libs/components/src/form-control/form-control-group-item.directive.ts
```diff
@@ -0,0 +1,30 @@
...
+
+import { FormControlGroupComponent } from "./form-control-group.component";
+
+@Directive({
+  selector: "[bitFormControlGroupItem]",
+})
+export class FormControlGroupItemDirective {
+  readonly group = inject(FormControlGroupComponent, { optional: true });
+
+  // Optional — undefined is safe when used outside a group
```
vleague2: Why would this ever get used outside a group, if it is named `GroupItemDirective`?

libs/components/src/radio-button/radio-button.component.ts
```diff
@@ -1,58 +1,40 @@
...
+    return this.base.selected;
   }
 
-  get groupDisabled() {
-    return this.groupComponent.disabled;
-  }
-
-  get block() {
-    return this.groupComponent.block();
+  @HostBinding("class") get classList() {
```
vleague2: Possibly switch this to the `host` version while we're in here?

libs/components/src/radio-button/radio-input.component.ts
```diff
@@ -27,52 +27,63 @@ export class RadioInputComponent implements BitFormControlAbstraction {
...
+    "[&>label:hover]:before:tw-border-2",
+    "focus-visible:before:tw-border-2",
+    "[&>label:focus-visible]:before:tw-border-2",
 
     // if it exists, the parent form control handles focus
-    "[&:not(bit-form-control_*)]:focus-visible:tw-ring-2",
-    "[&:not(bit-form-control_*)]:focus-visible:tw-ring-offset-2",
-    "[&:not(bit-form-control_*)]:focus-visible:tw-ring-primary-600",
+    "[&:not(bit-form-control_*, bit-form-control-card)]:focus-visible:before:tw-ring-2",
+    "[&:not(bit-form-control_*, bit-form-control-card)]:focus-visible:before:tw-ring-offset-2",
```
vleague2: Do these need the child selector `_*`?

libs/components/src/switch/switch.component.ts
```diff
@@ -31,50 +29,130 @@ let nextId = 0;
...
+export class SwitchComponent implements ControlValueAccessor, BitFormControlAbstraction {
+  private readonly injector = inject(Injector);
+  private get ngControl(): NgControl | null {
+    return this.injector.get(NgControl, null, { self: true, optional: true });
+  }
+
+  readonly size = signal<"base" | "large">("base");
+
+  readonly ariaLabelledBy = signal<string | undefined>(undefined);
+  readonly ariaDescribedBy = signal<string | undefined>(undefined);
```
vleague2, line 54: Not important but I think if you did `signal<string>()`, that is the same end result typing

libs/components/src/form-control/form-control-card.component.html
```diff
@@ -0,0 +1,41 @@
+<div
+  class="tw-flex tw-items-center tw-relative tw-p-4 tw-border tw-border-border-base tw-rounded-xl tw-justify-between tw-gap-4 has-[label:hover]:tw-border-border-brand-soft has-[input:focus-visible]:tw-ring-2 has-[input:focus-visible]:tw-ring-border-focus has-[input:focus-visible]:tw-border-transparent has-[input:checked]:tw-bg-bg-brand-soft"
+  (change)="inGroup && groupItem.notifyChange()"
```
vleague2: Does this actually fire on a div?

vleague2 (commented):
Initial thoughts, I will need to do another pass. I also can't click the switch in the card view?

https://github.com/user-attachments/assets/8819e3dd-7bfe-42e1-a76e-cd09a4cd1aaf

libs/components/src/form-control/form-control-card.component.html
BryanCunningham: Yeah, this is how we make the whole card clickable. We then use the content that is passed in as aria-labelledby on the input

libs/components/src/form-control/form-control-card.component.html
BryanCunningham: It does. It bubbles up from the input and gets caught here. I don't love this though

libs/components/src/radio-button/radio-input.component.ts
BryanCunningham: They do not

libs/components/src/form-control/form-control-group-item.directive.ts
BryanCunningham: @vleague2 While it's possible, it's doesn't really make sense to do so. The `FormControlCardComponent` uses this as a host directive in all cases, regardless of if it's in a group or not. If no value is assigned to the item, the 'group' behavior is a noop. I updated this comment a little bit and added some more explicit instructions on the docs pages

libs/components/src/form-control/form-control-card.component.ts
BryanCunningham: I've updated this a little bit so that the abstract provides the input element instead

libs/components/src/checkbox/checkbox.mdx
```diff
@@ -36,6 +36,48 @@ import { CheckboxModule } from "@bitwarden/components";
...
+</bit-form-control-group>
+```
+
+### Opting out of group selection (not recommended)
+
+Omitting `[value]` while nested inside a group renders the card as a standalone control — it manages
+its own checked state via `formControlName` and does not interact with the group's selection. In
+most cases, if a card doesn't need group selection, it should be placed outside the group entirely.
+
+Choose checkboxes in a group when:
```
vleague2: I think this should go in the "using in a group" section

libs/components/src/form-control/form-control-base.directive.ts
```diff
@@ -0,0 +1,95 @@
...
+  selector: "[bitFormControlBase]",
+})
+export class FormControlBaseDirective {
+  readonly id = `bit-form-control-${++nextId}`;
+  readonly label = input<string>();
+
+  readonly inline = input(false, { transform: booleanAttribute });
+
+  readonly disableMargin = input(false, { transform: booleanAttribute });
+  readonly disableMarginSignal = signal(false);
```
vleague2, line 31: Is this because we can't make `disableMargin` a model with the boolean transform on it?

libs/components/src/form-control/form-control-card.component.spec.ts
```diff
@@ -0,0 +1,161 @@
...
+    });
+
+    it("updates aria-describedby to the error element", () => {
+      expect(inputEl.getAttribute("aria-describedby")).toBe(card.errorId);
+    });
+  });
+
+  describe("when there is no hint and no error", () => {
+    // TODO: Fix this the next time the file is edited.
+    // eslint-disable-next-line @angular-eslint/prefer-on-push-component-change-detection
```
vleague2: Only 1 left in this file!

libs/components/src/form-control/form-control-card.component.ts
```diff
@@ -0,0 +1,140 @@
...
+        // avoiding a querySelector race with Angular's property binding rendering cycle
+        switchElement.ariaLabelledBy.set(this.labelId);
+        switchElement.ariaDescribedBy.set(describedBy ?? undefined);
+        hostEl.removeAttribute("aria-labelledby");
+        hostEl.removeAttribute("aria-describedby");
+      } else {
+        const el = this.base.formControl().inputEl?.nativeElement ?? hostEl;
+
+        hostEl.removeAttribute("aria-labelledby");
+        hostEl.removeAttribute("aria-describedby");
```
vleague2: Looks like we remove these two attributes from the hostEl in both the if and else conditions. a) is that just in case a consumer erroneously attaches one to the host? b) we could move it up since it happens regardless of the if

libs/components/src/form-control/form-control-base.directive.ts
BryanCunningham: Nope. Removed

BryanCunningham: Also updated to use non-legacy host binding

libs/components/src/radio-button/radio-input.component.ts
BryanCunningham: Turns out we did need the child selector here. Updated

vleague2 (commented):
I started a review yesterday but didn't finish it, but here are the comments from it anyway lol

libs/components/src/form-control/form-control-base.directive.ts
BryanCunningham: Yeah

libs/components/src/form-control/form-control-card.component.ts
BryanCunningham: @vleague2 I refactored this a little bit. Let me know what you think

libs/components/src/form-control/form-control-card.component.ts
```diff
@@ -92,25 +80,34 @@ export class FormControlCardComponent {
...
     effect(() => {
       const hostEl = this.base.formControlEl().nativeElement;
-      const hasError = this.hasErrorSignal();
-
-      const describedBy = hasError ? this.errorId : (this.hint()?.id ?? null);
-
+      const hintId = this.inGroup
+        ? (this.groupItem.group?.hint()?.id ?? null)
+        : (this.hint()?.id ?? null);
+      const errorId = this.inGroup ? (this.groupItem.group?.errorId ?? null) : this.errorId;
```
vleague2: Trying to visualize how this works -- so if there are 3 cards in a group, and the group has an error, all 3 will read the same error, even if the group error only pertains to one of them or is an overall error (like an API error)?

libs/components/src/form-control/form-control-card.component.ts
```diff
@@ -0,0 +1,137 @@
...
+      const describedBy = [errorId, hintId].filter(Boolean).join(" ") || undefined;
+      const switchElement = this.switch();
+
+      // Always clear ARIA from host wrapper — inner element owns these
+      hostEl.removeAttribute("aria-labelledby");
+      hostEl.removeAttribute("aria-describedby");
```
vleague2: Do we have any stories where we're using the aria properties on the host? (I tried to find some but couldn't.) I.e. is this like, a supported actual property or are we just proactively removing them in case people add them erroneously? 

(Edit to add -- I know I keep coming back to these with different thoughts, idk I am just stuck on them haha)

libs/components/src/form-control/form-control-card.component.html
```diff
@@ -0,0 +1,49 @@
...
+>
+  <!-- label z-index to ensure label sits on top of switch thumb control(z-index: 2) -->
+  <label
+    [for]="base.inputId()"
+    class="tw-absolute tw-top-0 tw-left-0 tw-size-full tw-z-[3]"
+    [class]="base.formControl().disabled ? 'tw-cursor-default' : 'tw-cursor-pointer'"
+  ></label>
+
+  <div class="tw-flex tw-items-center tw-gap-4">
+    @if (icon()) {
```
vleague2: We should probably to `if (icon(); as icon)` and then `[icon]=icon` on line 19, so that the `if` check properly gates the icon tile

libs/components/src/form-control/form-control-group.component.html
```diff
@@ -0,0 +1,31 @@
+@if (label()) {
+  <fieldset class="tw-border-none tw-m-0 tw-p-0">
+    <legend class="tw-flex tw-gap-0.5 tw-items-center tw-mb-2 tw-text-sm/5 tw-font-medium">
+      <ng-content select="bit-label"></ng-content>
+      @if (required()) {
+        <sup class="tw-text-sm tw-text-fg-danger tw-relative tw-top-[-1px]">*</sup>
```
vleague2: `aria-hidden` for this one too?

libs/components/src/form-control/form-control-group.component.spec.ts
```diff
@@ -0,0 +1,240 @@
...
+import { FormControlGroupComponent } from "./form-control-group.component";
+import { FormControlModule } from "./form-control.module";
+
+describe("FormControlGroupComponent", () => {
+  let fixture: ComponentFixture<TestHostComponent>;
+  let host: TestHostComponent;
+  let group: FormControlGroupComponent;
+
+  // TODO: Fix this the next time the file is edited.
+  // eslint-disable-next-line @angular-eslint/prefer-on-push-component-change-detection
```
vleague2: Another todo to fix

libs/components/src/form-control/form-control-card.component.ts
BryanCunningham: @vleague2 This is really only for a scenario where we have a group of checkboxes and the entire group is validated together. Like if you have to chose at least one of the options. There should be no API errors on checkboxes as you can't really chose the wrong thing. 

While this logic _could_ also be applied to groups of switches, it is not recommended that you do that. As switches are meant to take effect immediately and should never be required. 

Switches could still be in a group if they are related and it makes sense to do so.

libs/components/src/form-control/form-control-card.component.ts
BryanCunningham: @vleague2 We do not have an example of this currently. And, I suspect this isn't a real scenario we'd run into often. That being said, I figured *if* the consumer did want to use an additional label and/or description, we should probably support that, just in case

libs/components/src/checkbox/checkbox.component.ts
```diff
@@ -26,90 +36,88 @@ export class CheckboxComponent implements BitFormControlAbstraction {
...
-    "!-tw-mt-1",
-    "!-tw-mb-1",
-    "!-tw-ms-1",
+    "tw-h-6",
+    "tw-w-6",
+    "tw-rounded-lg",
+    // negative margin to negate the positioning added by the sizing
+    "!-tw-mt-[1px]",
+    "!-tw-mb-[1px]",
+    "!-tw-ms-[1px]",
```
willmartian: nit: you can use `tw-mt-px` instead of `tw-mt-[1px]`

libs/components/src/checkbox/checkbox.mdx
```diff
@@ -36,6 +36,48 @@ import { CheckboxModule } from "@bitwarden/components";
 
 <Canvas of={stories.Custom} />
 
-### Intermediate
+### Indeterminate
```
willmartian, line 39: :)

libs/components/src/form-control/form-control.component.html
willmartian: Bump, now is a good time to remove `ngClass`

libs/components/src/form-control/form-control.component.html
```diff
@@ -1,27 +1,28 @@
...
-  [ngClass]="[formControl().disabled ? 'tw-cursor-auto' : 'tw-cursor-pointer']"
+  class="tw-transition tw-items-start tw-gap-2 tw-select-none tw-mb-0 tw-inline-flex tw-rounded has-[:focus-visible]:tw-ring has-[:focus-visible]:tw-ring-border-focus"
+  [ngClass]="[base.formControl().disabled ? 'tw-cursor-auto' : 'tw-cursor-pointer']"
 >
   <ng-content></ng-content>
+  <!-- minor padding adjustment for visual alignment -->
   <span
-    class="tw-inline-flex tw-flex-col"
-    [ngClass]="formControl().disabled ? 'tw-text-muted' : 'tw-text-main'"
+    class="tw-inline-flex tw-flex-col tw-pt-[1px]"
```
willmartian: `tw-pt-px`

libs/components/src/form-control/form-control.abstraction.ts
```diff
@@ -1,6 +1,10 @@
+import { ElementRef } from "@angular/core";
+
 export abstract class BitFormControlAbstraction {
   abstract disabled: boolean;
   abstract required: boolean;
   abstract hasError: boolean;
   abstract error: [string, any];
+  inputId?: string;
+  inputEl?: ElementRef<HTMLInputElement>;
```
willmartian: Why are these needed now but not before?

libs/components/src/form-control/form-control-group.component.ts
```diff
@@ -0,0 +1,191 @@
...
+
+  // ── Mode (auto-detected from child types) ─────────────────────────────────
+  // Radio children call registerRadioChild() in their constructor, which flips
+  // mode to 'single'. Checkbox/switch children never register, keeping 'multi'.
+  private readonly _hasRadioChildren = signal(false);
+  readonly mode = computed(() => (this._hasRadioChildren() ? "single" : "multi"));
+
+  registerRadioChild(): void {
+    this._hasRadioChildren.set(true);
+  }
```
willmartian, line 76: Having a register function like this feels unidiomatic. Could we instead use a `contentChild`?

willmartian (changes requested):
A few small comments. Also, I think Vicki still has some unaddressed comments--could you respond or mark them as resolved?

libs/components/src/form-control/form-control-card.component.ts
willmartian, line 1: Why does this component have so much more logic that `form-control.component`? Aren't they meant to be essentially the same component with two different templates?

libs/components/src/form-control/form-control-group.component.html
```diff
@@ -0,0 +1,33 @@
...
+    </legend>
+    <ng-container *ngTemplateOutlet="contentTpl"></ng-container>
+  </fieldset>
+}
+
+@if (!label()) {
+  <ng-container *ngTemplateOutlet="contentTpl"></ng-container>
+}
+
+<ng-template #contentTpl>
```
willmartian: Tpl is not an obvious shortening of template to me, could we expand it?

willmartian (commented):
I believe we want this to target `main`. I don't think it relies on anything else in M3?

willmartian (commented):
I was doing some code review with Claude and then it just turned into code. Chat about this in the AM? https://github.com/bitwarden/clients/pull/20192

BryanCunningham:
> I was doing some code review with Claude and then it just turned into code. Chat about this in the AM? #20192

@willmartian Happy to chat. From what I can tell so far, I like your solution much more. Far more simple.

libs/components/src/form-control/form-control-card.component.ts
willmartian: I understand now.

BryanCunningham:
> I believe we want this to target `main`. I don't think it relies on anything else in M3?

@willmartian I was under the impression that all form related updates would go out at the same time. I'm fine with this going to main first but we should double check that design is ok with it

vleague2:
> I was under the impression that all form related updates would go out at the same time. I'm fine with this going to main first but we should double check that design is ok with it

My unsolicited opinion is to keep all the form updates together in M3

willmartian (approved):
Great work! Lint needs to be fixed, but otherwise, LGTM.

apps/browser/src/vault/popup/settings/admin-settings.component.html
```diff
@@ -14,14 +14,15 @@
 
     <form [formGroup]="adminForm">
       <bit-card>
-        <bit-switch formControlName="autoConfirm">
+        <bit-form-control>
```
nick-livefront, line 17: ⛏️  Can `disableMargin` be added? There is extra space being added by the control to this setting

Image

BryanCunningham: @nick-livefront Yeah, if we need that, for sure. I'd really like for none of these components to have any built in margin on them. Maybe one day!

EDIT: `bit-form-control` has this available. You can just add it there and it'll work as you'd expect 
```html
<bit-form-control disableMargin>
```

JaredSnider-Bitwarden (approved):

nick-livefront (approved):

apps/browser/src/vault/popup/settings/admin-settings.component.html
BryanCunningham: For posterity: I'm going to fast follow with this change in a different PR. I don't want to force folks to approve again.

willmartian (approved):

PR #19461, opened 2026-03-10
[PM-33345] Bump typescript to 5.9

quexten:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-33345

## 📔 Objective

Updates the remaining uncaught type issues and bumps typescript to version 5.9

### DIFF
--- a/libs/eslint/components/require-theme-colors-in-svg.spec.mjs
+++ b/libs/eslint/components/require-theme-colors-in-svg.spec.mjs
@@ -4,7 +4,6 @@ import rule from "./require-theme-colors-in-svg.mjs";
 const ruleTester = new RuleTester({
   languageOptions: {
     parserOptions: {
-      project: [__dirname + "/../tsconfig.spec.json"],
       projectService: {
         allowDefaultProject: ["*.ts*"],
       },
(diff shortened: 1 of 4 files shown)


libs/common/src/key-management/sends/types/send-hashed-password.type.ts
```diff
@@ -1,4 +1,4 @@
 import { Opaque } from "type-fest";
 
 export type SendHashedPassword = Opaque<Uint8Array, "SendHashedPassword">;
```
mzieniukbw: ❓  shouldn't this also be `Uint8Array<ArrayBuffer>` ?
```suggestion
export type SendHashedPassword = Opaque<Uint8Array<ArrayBuffer>, "SendHashedPassword">;
```

package-lock.json
```diff
@@ -116,7 +116,7 @@
         "@types/retry": "0.12.5",
         "@types/zxcvbn": "4.4.5",
         "@typescript-eslint/rule-tester": "8.31.0",
```
mzieniukbw: Let's upgrade this one too 
```suggestion
        "@typescript-eslint/rule-tester": "8.57.0",
```

libs/common/src/key-management/sends/types/send-hashed-password.type.ts
quexten: It was not needed to make the ts compiler happy, but I've added it now. Depending on the usage, TS can infer the type, and in the case of the sendHashedPassword it was not used in a way where it was needed.

package-lock.json
quexten: Applied

PR #19469, opened 2026-03-10
feat(platform): extract vaultUrl from /api/config response

addisonbeck:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-33352

https://github.com/bitwarden/sdk-internal/pull/832

## 📔 Objective

Updates `DefaultConfigService` to extract the vault URL from the server's `/api/config` response and populate it in the `ServerCommunicationConfig` that gets passed to the SDK.

The `vaultUrl` field is extracted from `response.environment.vault` using optional chaining to safely handle cases where the environment field may be undefined. 

This enables the SDK to receive an explicit vault URL for SSO cookie acquisition redirects, which fixes two use cases:
1. **Port preservation**: Prevents `localhost:8000` being stripped to `localhost` in local development environments
2. **Multi-subdomain routing**: Ensures redirects use `vault.bitwarden.com` instead of `api.bitwarden.com` when the cookie domain differs from the vault hostname

### DIFF
--- a/libs/common/src/platform/services/config/default-config.service.ts
+++ b/libs/common/src/platform/services/config/default-config.service.ts
@@ -251,17 +251,18 @@ export class DefaultConfigService implements ConfigService {
 
   private parseBoostrapConfig(response: ServerConfigResponse) {
     const bootstrap = response.communication?.bootstrap ?? null;
+    const vaultUrl = response.environment?.vault;
 
     // Emit communication config so subscribers (e.g. ServerCommunicationConfigService) can persist it
     const communicationConfig: ServerCommunicationConfig =
-      bootstrap?.type === "ssoCookieVendor"
+      bootstrap?.type === "ssoCookieVendor" && vaultUrl != null
         ? {
             bootstrap: {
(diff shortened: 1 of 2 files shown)


libs/common/src/platform/services/config/default-config.service.ts
```diff
@@ -261,6 +261,7 @@ export class DefaultConfigService implements ConfigService {
               idpLoginUrl: bootstrap.idpLoginUrl,
               cookieName: bootstrap.cookieName,
               cookieDomain: bootstrap.cookieDomain,
+              vaultUrl: response.environment?.vault,
```
coroiu: issue: this is probably gonna cause a panic if the value is undefined/null. We might want to consider skipping the emission completely if we don't have this information

addisonbeck: implemented on https://github.com/bitwarden/clients/pull/19469/commits/d257289c2bb0afeeba1e2e76bf9263aefae91e9d I just set the communication type to direct if there is no vault url.

PR #19422, opened 2026-03-06
[PM-32424] Send Access Enumeration protection

ike-kottlowski:
## 🎟️ Tracking

[PM-32424](https://bitwarden.atlassian.net/browse/PM-32424)

Server PR: https://github.com/bitwarden/server/pull/7166
SDK PR: https://github.com/bitwarden/sdk-internal/pull/821

## 📔 Objective

Improve enumeration protection of Sends

## 📸 Screenshots

[PM-32424]: https://bitwarden.atlassian.net/browse/PM-32424?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #19471, opened 2026-03-10
[PM-31112] Local user data key state declaration

mzieniukbw:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-31112

## 📔 Objective

Adds local user data key state declaration.
Split from #19433, to unblock #19422.


libs/common/src/platform/services/sdk/client-managed-state.ts
```diff
@@ -93,4 +91,12 @@ export class RepositoryRecord<ClientType, SdkType> implements Repository<SdkType
   async removeAll(): Promise<void> {
     await this.getUserState().update(() => ({}));
   }
+
+  private getUserState() {
```
mzieniukbw, line 95: this was formatter moving the function, there is no functional change here.

coroiu (approved):

PR #19475, opened 2026-03-10
[PM-33372] hide prompt if org does not use my items

JaredScar:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-33372

## 📔 Objective

Hide the prompt "Prompt users to move individual vault items to the organization" from policy if organization does not enable "use my items" option.

## 📸 Screenshots

image


apps/web/src/app/admin-console/organizations/policies/policies.component.ts
```diff
@@ -130,13 +131,18 @@ export class PoliciesComponent {
       .subscribe();
   }
 
-  edit(policy: BasePolicyEditDefinition, organizationId: OrganizationId) {
+  edit(
+    policy: BasePolicyEditDefinition,
+    organizationId: OrganizationId,
+    organization: Organization,
+  ) {
```
BTreston: We should just use `organization` and infer the ID from the object rather than 2 separate properties.

apps/web/src/app/admin-console/organizations/policies/policy-edit-dialog.component.ts
```diff
@@ -40,41 +46,43 @@ export type PolicyEditDialogData = {
    * The organization ID for the policy.
    */
   organizationId: string;
+  /**
+   * The organization object, used to determine which policy options are available
+   * based on the org's feature entitlements.
+   */
+  organization?: Organization;
```
BTreston: Same as above, we should just use the whole object.

apps/web/src/app/admin-console/organizations/policies/base-policy-edit.component.ts
```diff
@@ -86,6 +86,9 @@ export abstract class BasePolicyEditComponent implements OnInit {
   // FIXME(https://bitwarden.atlassian.net/browse/CL-903): Migrate to Signals
   // eslint-disable-next-line @angular-eslint/prefer-signals
   @Input() policy: BasePolicyEditDefinition | undefined;
+  // FIXME(https://bitwarden.atlassian.net/browse/CL-903): Migrate to Signals
+  // eslint-disable-next-line @angular-eslint/prefer-signals
+  @Input() organization: Organization | undefined;
```
BTreston: Why do we need a new input here? The `policyResponse` should have the `organizationId`. I would imagine it makes more sense for the implementing class to get its own `organization$` object from `OrganizationService` using that ID rather than adding a new input to the base component and exposing it to all the components regardless of if they need it.

BTreston (changes requested):
Overall looks good, just a few points of feedback. This looks like it will cause conflicts with the multi step refactor, so be aware of that.

apps/web/src/app/admin-console/organizations/policies/policy-edit-dialogs/auto-confirm-edit-policy-dialog.component.ts
```diff
@@ -98,7 +98,11 @@ export class AutoConfirmPolicyDialogComponent
   private readonly openExtensionTitle: Signal<TemplateRef<unknown> | undefined> =
     viewChild("step1Title");
 
-  override policyComponent: AutoConfirmPolicyEditComponent | undefined;
+  protected saveDisabled$: Observable<boolean> | undefined;
+
+  private get typedPolicyComponent(): AutoConfirmPolicyEditComponent | undefined {
+    return this.policyComponent() as AutoConfirmPolicyEditComponent | undefined;
```
BTreston: We should use type narrowing here rather than `as`

apps/web/src/app/admin-console/organizations/policies/policy-edit-dialogs/organization-data-ownership-edit-policy-dialog.component.ts
```diff
@@ -83,7 +82,11 @@ export class OrganizationDataOwnershipPolicyDialogComponent
   private readonly policyForm = viewChild.required<TemplateRef<unknown>>("step0");
   private readonly policyFormTitle = viewChild.required<TemplateRef<unknown>>("step0Title");
 
-  override policyComponent: vNextOrganizationDataOwnershipPolicyComponent | undefined;
+  protected saveDisabled$: Observable<boolean> | undefined;
+
+  private get typedPolicyComponent(): vNextOrganizationDataOwnershipPolicyComponent | undefined {
+    return this.policyComponent() as vNextOrganizationDataOwnershipPolicyComponent | undefined;
```
BTreston: same feedback as above

BTreston (changes requested):
One more thing

BTreston (dismissed):

apps/web/src/app/vault/services/web-vault-extension-prompt.service.spec.ts
```diff
@@ -224,7 +224,7 @@ describe("WebVaultExtensionPromptService", () => {
       getFeatureFlag.mockResolvedValueOnce(minAgeDays);
 
       const exactDate = new Date();
-      exactDate.setDate(exactDate.getDate() - 30); // Exactly 30 days old
+      exactDate.setTime(exactDate.getTime() - 30 * 24 * 60 * 60 * 1000 - 1); // Exactly 30 days old in ms
```
BTreston: 🎨 Do we need to do this?

BTreston (dismissed):
non blocking question

BTreston (approved):
Thank you for making all the requested changes

gbubemismith (approved):

PR #19491, opened 2026-03-11
[PM-31115] Support ephemeral pin envelope state in SDK

quexten:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-31115
https://github.com/bitwarden/sdk-internal/pull/836
https://github.com/bitwarden/clients/pull/19519

## 📔 Objective

Client managed state registrations for ephemeral PIN unlock. This allows the SDK to manage the ephemeral PIN / set it up after unlock.

### DIFF
--- a/libs/common/src/key-management/ephemeral-pin-envelope-mapper.ts
+++ b/libs/common/src/key-management/ephemeral-pin-envelope-mapper.ts
@@ -4,8 +4,10 @@ import { EphemeralPinEnvelopeState } from "@bitwarden/sdk-internal";
 
 import { PIN_PROTECTED_USER_KEY_ENVELOPE_EPHEMERAL } from "./pin/pin.state";
 
-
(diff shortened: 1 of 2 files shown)


libs/common/src/key-management/ephemeral-pin-envelope-mapper.ts
```diff
@@ -0,0 +1,19 @@
+import { SdkRecordMapper } from "@bitwarden/common/platform/services/sdk/client-managed-state";
+import { UserKeyDefinition } from "@bitwarden/common/platform/state";
+
+import { PIN_PROTECTED_USER_KEY_ENVELOPE_EPHEMERAL } from "./pin/pin.state";
+import { EphemeralPinEnvelopeState } from "@bitwarden/sdk-internal";
+
+export class EphemeralPinEnvelopeMapper implements SdkRecordMapper<EphemeralPinEnvelopeState, EphemeralPinEnvelopeState> {
```
quexten: I wonder whether we can make a generic wrapper for things that don't need mapping so we don't have to re-write this in the future..

PR #19492, opened 2026-03-11
[PM-33396] Add middleware support to ApiService in clients

coroiu:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-33396

## 📔 Objective

Adds middleware pipeline support so that we can add cookies and retries to requests.


coroiu:
Closing as we'll pull this in with Daniels PR

PR #19496, opened 2026-03-11
[PM-33380] Fix Access Intelligence drawer not opening on row click

AlexRubik:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-33380

## 📔 Objective

Fix the "All Applications" tab row click handler that silently crashes when opening the at-risk members drawer.

The `showAppAtRiskMembers` method was converted from an arrow function to a regular method in [#18990](https://github.com/bitwarden/clients/pull/18990) ([PM-32127](https://bitwarden.atlassian.net/browse/PM-32127)). Because this method is passed as a callback to a child component via Angular input binding, the regular method loses `this` context when invoked, causing a TypeError.

This PR converts it back to an arrow function property, matching the identical pattern in `critical-applications.component.ts`.

[PM-32127]: https://bitwarden.atlassian.net/browse/PM-32127?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/bitwarden_license/bit-web/src/app/dirt/access-intelligence/all-applications/applications.component.spec.ts
+++ b/bitwarden_license/bit-web/src/app/dirt/access-intelligence/all-applications/applications.component.spec.ts
@@ -263,6 +263,27 @@ describe("ApplicationsComponent", () => {
     });
   });
 
+  describe("showAppAtRiskMembers", () => {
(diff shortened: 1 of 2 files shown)


lastbestdev (approved):
Working as expected. Good catch & fix 👍🏻

PR #19500, opened 2026-03-11
[PM-29796] Parameter validation with pattern match

enmande:
## 🎟️ Tracking

[PM-29796](https://bitwarden.atlassian.net/browse/PM-29796)

## 📔 Objective

- Adds boundary check for incoming IDs; provides a better user experience in the event an invite fails to parse.
- Updates pattern matching on Utils for outbound URLs.
  - This means all outbound requests will now be validated for additional unusual patterns, as described in `Utils.ts` updates.

### Authed
Invite fails to parse in an authenticated scenario

https://github.com/user-attachments/assets/668fe302-cc74-4fea-a920-2ccde9b24c4c

### Unauthed
Invite fails to parse in an unauthenticated scenario

https://github.com/user-attachments/assets/c3377e4c-5968-45f4-a1f6-5a6a111fbea7

[PM-29796]: https://bitwarden.atlassian.net/browse/PM-29796?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/web/src/app/auth/organization-invite/accept-organization.component.ts
```diff
@@ -43,6 +44,16 @@ export class AcceptOrganizationComponent extends BaseAcceptComponent {
...
+    if (invite === null) {
+      throw new Error(this.i18nService.t(`${this.failedMessage}`));
+      this.toastService.showToast({
+        message: this.i18nService.t(this.failedMessage),
+        variant: "error",
+        timeout: 10000,
+      });
+      await this.router.navigate(["/"]);
+      return;
+    }
```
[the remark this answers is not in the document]
enmande: Was evaluating throwing to `BaseAcceptComponent` vs full handling in this component. Pushed early; did not intend this to be reviewed as the authoritative work. Good catch though ;)

apps/web/src/app/auth/organization-invite/accept-organization.component.ts
```diff
@@ -61,11 +73,22 @@ export class AcceptOrganizationComponent extends BaseAcceptComponent {
       timeout: 10000,
     });
 
-    await this.router.navigate(["/vault"]);
+    await this.router.navigate(["/"]);
```
JaredSnider-Bitwarden, line 71: ❔ : this change was not super obvious to me?  Was this to just leave the redirection up to the `redirectGuard` as the previous vault navigation assumed that the user was unlocked?

apps/web/src/app/auth/organization-invite/accept-organization.component.ts
```diff
@@ -61,11 +73,22 @@ export class AcceptOrganizationComponent extends BaseAcceptComponent {
...
+    if (invite === null) {
+      // If invite is null (fromParams failed to validate, etc.), we must handle that case here in full.
+      // The unauthedHandler does not account for error handling in the BaseAcceptComponent.
+      this.toastService.showToast({
+        message: this.i18nService.t(this.failedMessage),
+        variant: "error",
+        timeout: 10000,
+      });
+      await this.router.navigate(["/"]);
+      return;
```
JaredSnider-Bitwarden: Non blocking nit (up to you): this could be a re-usable `handleInvalidInvite()` method.

libs/common/src/platform/misc/utils.ts
```diff
@@ -20,6 +20,21 @@ import { I18nService } from "../abstractions/i18n.service";
...
+  "\t",
+  "%09", // TAB (stripped by WHATWG URL parser during normalization)
+  "\n",
+  "%0a", // Line feed (stripped by WHATWG URL parser during normalization)
+  "\r",
+  "%0d", // Carriage return (stripped by WHATWG URL parser during normalization)
+  "\0",
+  "%00", // Null byte (can truncate strings in some parsers)
+];
+const queryDangerousPatterns = ["/", "%2f", "#", "%23"];
```
JaredSnider-Bitwarden: 🎨 :  is it worth documenting these just as you have done excellently above with the `pathTraversalPatterns`?

```
const queryDangerousPatterns = [
  "/", // Forward slash (could escape query context into path)
  "%2f", // URL-encoded "/"
  "#", // Hash (would terminate the query string and begin a fragment)
  "%23", // URL-encoded "#"
];

```

libs/common/src/platform/misc/utils.ts
```diff
@@ -20,6 +20,21 @@ import { I18nService } from "../abstractions/i18n.service";
...
+  "%5c", // Backslash (some parsers normalize to forward slash)
+  "\t",
+  "%09", // TAB (stripped by WHATWG URL parser during normalization)
+  "\n",
+  "%0a", // Line feed (stripped by WHATWG URL parser during normalization)
+  "\r",
+  "%0d", // Carriage return (stripped by WHATWG URL parser during normalization)
+  "\0",
+  "%00", // Null byte (can truncate strings in some parsers)
+];
```
JaredSnider-Bitwarden, line 47: 🎨 : docs are great, but it wasn't immediately obvious to me (b/c I don't have all the various url encoded chars memorized) that these were in groups of 2. Thoughts on something like for explicitness:
```
const pathTraversalPatterns = [
  "..", // Double-dot traversal (single-encoded resolves to ".." via decodeURIComponent)
  "%2e", // URL-encoded ".."
  "\\", // Backslash (some parsers normalize to forward slash)
  "%5c", // URL-encoded "\\"
  "\t", // TAB (stripped by WHATWG URL parser during normalization)
  "%09", // URL-encoded "\t"
  "\n", // Line feed (stripped by WHATWG URL parser during normalization)
  "%0a", // URL-encoded "\n"
  "\r", // Carriage return (stripped by WHATWG URL parser during normalization)
  "%0d", // URL-encoded "\r"
  "\0", // Null byte (can truncate strings in some parsers)
  "%00", // URL-encoded "\0"
];
```

apps/web/src/app/auth/organization-invite/accept-organization.component.ts
enmande: 😸 Thank you for noticing and asking this in particular. Yes, that's exactly the thinking. It's also the redirection expressed in the `authedHandler` error handling [in the superclass component](https://github.com/bitwarden/clients/blob/849bf53eca073f183ddf0e76a9245cc2e0e8de77/apps/web/src/app/common/base.accept.component.ts#L71). Because local testing showed the behavior to be the same, it made sense to me to align these behaviors if we were going to explicitly define them at this level, and allow higher-level concerns (e.g., `redirectGuard`) to set the standard.

JaredSnider-Bitwarden: Perfect. Thank you for the clarification!

libs/common/src/platform/misc/utils.ts
enmande: 🫡 Excellent feedback, ty. Clarified in a452ba1.

libs/common/src/platform/misc/utils.ts
enmande: 🫡 Excellent feedback, ty. Clarified in a452ba1.

apps/web/src/app/auth/organization-invite/accept-organization.component.ts
enmande: 😸 I always go back and forth on 2 vs 3 repetitions as a guideline for abstraction. This makes good sense here and increases readability. Refactored in 4813b3c.

PR #19507, opened 2026-03-11
[CL-1099] spike: bit-item policies list story

willmartian:
## Summary

Illustrative story showing what it might look like to use `bit-item` for the Admin Console policies list, using real policy copy.

- Adds a `PoliciesList` story to `libs/components/src/item/item.stories.ts`
- Uses actual policy names and descriptions from the web app's i18n strings
- Shows `On`/`Off` badges (`success`/`secondary` variants) in the end slot
- Wrapped in `bit-layout` for the square-corner appearance

## Outcome

Decided to use bit-item on policies page: https://github.com/bitwarden/clients/pull/19151#pullrequestreview-3932971426

## Screenshots

_See Storybook: Component Library/Item > Policies List_

### DIFF
--- a/apps/browser/src/_locales/en/messages.json
+++ b/apps/browser/src/_locales/en/messages.json
@@ -6118,8 +6118,8 @@
       }
     }
   },
-  "transferItemsToOrganizationContent": {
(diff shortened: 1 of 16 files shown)


willmartian:
@claude, since the badges are not interactive, they can be placed in the end slot of `bit-item-content` without the action wrapper

PR #17080, opened 2025-10-28
Auth/pm 23801/move prelogin request

Patrick-Pimentel-Bitwarden:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-23801

## 📔 Objective

1. Added in immediate prelogin call upon hitting continue button.
2. Added tests.

## Original Implementation
https://github.com/user-attachments/assets/1192cee8-1c42-41a9-bd09-a807adaa7eb6

## New Implementation
https://github.com/user-attachments/assets/2cda8c0b-96ac-4ceb-a717-a488f971fa60

## ⏰ Reminders before review

- Contributor guidelines followed
- All formatters and local linters executed and passed
- Written new unit and / or integration tests where applicable
- Protected functional changes with optionality (feature flags)
- Used internationalization (i18n) for all UI strings
- CI builds passed
- Communicated to DevOps any deployment requirements
- Updated any necessary documentation (Confluence, contributing docs) or informed the documentation team

## 🦮 Reviewer guidelines

- 👍 (`:+1:`) or similar for great changes
- 📝 (`:memo:`) or ℹ️ (`:information_source:`) for notes or general info
- ❓ (`:question:`) for questions
- 🤔 (`:thinking:`) or 💭 (`:thought_balloon:`) for more open inquiry that's not quite a confirmed issue and could potentially benefit from discussion
- 🎨 (`:art:`) for suggestions / improvements
- ❌ (`:x:`) or ⚠️ (`:warning:`) for more significant problems or concerns needing attention
- 🌱 (`:seedling:`) or ♻️ (`:recycle:`) for future improvements or indications of technical debt
- ⛏ (`:pick:`) for minor or nitpick changes

PR #19510, opened 2026-03-11
Auth/PM-33353 - Password Login - refine prefetching of password prelogin data

JaredSnider-Bitwarden:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-33353

## 📔 Objective

To fix a bug introduced in #17080 where the password prelogin API call was made twice when the feature flag was enabled, I've extracted all password prelogin logic into its own domain: `PasswordPreloginApiService`, `PasswordPreloginService` (with `DefaultPasswordPreloginService`), and `PasswordPreloginData`. The new service encapsulates caching and deduplication of the prelogin observable per email. It will enable us to swap to the SDK implementation more cleanly than we could before as well in follow up work. 

`LoginComponent` now attempts to populate the cache on `Continue` and attaches the resolved data to `PasswordLoginCredentials` on `Submit`. `PasswordLoginStrategy` consumes it directly via a private `makePasswordPreloginMasterKey` method, falling back to the `DefaultPasswordPreloginService` if no prefetched data is present. This lets `LoginStrategyService` shed all password scoped prelogin state, and removes the `PM23801_PrefetchPasswordPrelogin` feature flag entirely. 

⚠️ :  This will require QA testing on the feature.  ✅ done

## 📸 Screenshots

Note: the password prelogin call is only made once. 

https://github.com/user-attachments/assets/d064a0be-4cbe-428a-a279-db12cb2c8c06

Another demo showing that, when a password prelogin request is already in flight but not yet resolved, another request is not fired:

https://github.com/user-attachments/assets/c91372d4-b9f5-4b28-be70-378562a6eab4


libs/common/src/auth/password-prelogin/default-password-prelogin.service.ts
```diff
@@ -0,0 +1,28 @@
...
+      return this.currentPreloginData$;
+    }
+
+    this.currentEmail = normalized;
+    this.currentPreloginData$ = from(
+      this.passwordPreloginApiService.getPreloginData(new PasswordPreloginRequest(normalized)),
+    ).pipe(map(PasswordPreloginData.fromResponse), shareReplay({ bufferSize: 1, refCount: false }));
+
+    return this.currentPreloginData$;
+  }
```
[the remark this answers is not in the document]
JaredSnider-Bitwarden: Fixed with 98079946a4b2861680d847ce83cbaedca24a5764

libs/auth/src/common/services/login-strategies/login-strategy.service.ts
```diff
@@ -135,33 +100,34 @@ export class LoginStrategyService implements LoginStrategyServiceAbstraction {
...
+    private authRequestService: AuthRequestServiceAbstraction,
+    private userDecryptionOptionsService: InternalUserDecryptionOptionsServiceAbstraction,
+    private stateProvider: GlobalStateProvider,
+    private billingAccountProfileStateService: BillingAccountProfileStateService,
+    private vaultTimeoutSettingsService: VaultTimeoutSettingsService,
+    private kdfConfigService: KdfConfigService,
+    private taskSchedulerService: TaskSchedulerService,
+    private configService: ConfigService,
+    private accountCryptographicStateService: AccountCryptographicStateService,
+    private passwordPreloginService: PasswordPreloginService,
```
JaredSnider-Bitwarden, line 130: There are no subclasses for the `LoginStrategyService` so... no reason I could see to be protected.

libs/auth/src/common/services/login-strategies/login-strategy.state.spec.ts
JaredSnider-Bitwarden, line 1: All strict TS changes.

libs/auth/src/common/login-strategies/password-login.strategy.spec.ts
JaredSnider-Bitwarden, line 1: I added tests for the pre-fetching of password prelogin data (along with several other untested flows - which are now tested and documented heavily).

libs/common/src/auth/password-prelogin/password-prelogin.service.ts
```diff
@@ -0,0 +1,20 @@
...
+   * then call again with the same email to await the result. Returns the same
+   * in-flight observable for a given email, starting a fresh request if the email changes.
+   */
+  abstract getPreloginData$(email: string): Observable<PasswordPreloginData>;
+
+  /**
+   * Clears any cached prelogin data. Should be called after a successful password login
+   * to prevent stale KDF config from persisting in memory.
+   */
+  abstract clearCache(): void;
```
JaredSnider-Bitwarden, line 19: This defensively bounds the lifetime of the memory of this service to its intended use - regardless of the process reloads we have in place on our clients.

Patrick-Pimentel-Bitwarden (dismissed):
👍 Changes look good! had some discussion in slack about the caching getting cleared and the resolution looks great.

Patrick-Pimentel-Bitwarden (approved):

PR #19521, opened 2026-03-12
PM-31418 implemented password generator inside drawer

bmbitwarden:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-31418

## 📔 Objective

on web, puts Password Generator in the drawer

## 📸 Screenshots

https://github.com/user-attachments/assets/b9a00b2a-ed23-4e16-90e5-35576247f5fd


harr1424 (commented):
The code changes in this PR look good!  I am confused though because the linked Jira ticket is already in QA and doesn't mention anything about a drawer. 

Following recent feedback from product, this story should likely go through "Design Review" before code changes are approved and sent to QA.

bmbitwarden:
> The code changes in this PR look good! I am confused though because the linked Jira ticket is already in QA and doesn't mention anything about a drawer.
> 
> Following recent feedback from product, this story should likely go through "Design Review" before code changes are approved and sent to QA.

i am a bit confused too because the requirements have become a bit scattered across multiple jira tickets, pr comments and slack threads

I just moved https://bitwarden.atlassian.net/browse/PM-31418 back to code review so we can get this portion set before going back to qa status

jprusik (dismissed):

harr1424 (dismissed):

apps/browser/src/autofill/services/autofill-overlay-content.service.ts
```diff
@@ -1129,7 +1129,8 @@ export class AutofillOverlayContentService implements AutofillOverlayContentServ
     }
 
     autofillFieldData.inlineMenuFillType = CipherType.Login;
-    autofillFieldData.showPasskeys = autofillFieldData.autoCompleteType.includes("webauthn");
+    autofillFieldData.showPasskeys =
+      autofillFieldData.autoCompleteType?.includes("webauthn") ?? false;
```
blackwood, line 1135: Per offline discussion, this will be handled in [PM-33555](https://bitwarden.atlassian.net/browse/PM-33555) and [PM-29525](https://bitwarden.atlassian.net/browse/PM-29525), and isn't related to the work present otherwise.

blackwood (changes requested):

blackwood (approved):
approved for autofill concerns (via relocation)

jprusik (approved):
Approving for Autofill concerns (there no longer are any)

libs/tools/send/send-ui/src/add-edit/send-add-edit-dialog.component.html
```diff
@@ -1,35 +1,77 @@
...
+    <div [hidden]="showGenerator()" class="tw-flex tw-flex-row tw-items-center tw-gap-2 tw-flex-1">
+      <button bitButton type="submit" form="sendForm" buttonType="primary" #submitBtn>
+        {{ "save" | i18n }}
+      </button>
+      <button bitButton type="button" buttonType="secondary" bitDialogClose>
+        {{ "cancel" | i18n }}
+      </button>
+      <div class="tw-ml-auto">
+        <button
+          *ngIf="config?.mode !== 'add'"
```
mcamirault, line 66: ⛏️ Since you're using `@if` above, might as well do it here too to remove the last `*ngIf`. That might allow you to slim down the imports list for this component as well.

libs/tools/send/send-ui/src/add-edit/send-add-edit-dialog.component.html
```diff
@@ -1,35 +1,77 @@
...
+        (click)="useGeneratedPassword()"
+        [disabled]="!(generatorButtonLabel() && generatedValue())"
+      >
+        {{ generatorButtonLabel() }}
+      </button>
+      <button bitButton type="button" buttonType="secondary" (click)="closeGenerator()">
+        {{ "cancel" | i18n }}
+      </button>
+    }
+    <div [hidden]="showGenerator()" class="tw-flex tw-flex-row tw-items-center tw-gap-2 tw-flex-1">
```
mcamirault, line 57: 🎨 Unlike the form component I don't think it matters if these ones get destroyed and recreated. I'd suggest removing the `hidden` and putting the div into an `@else` block for the above `@if` instead, it's a little cleaner IMO

mcamirault (approved):
Nice job implementing this! One concern I have is that this change requires the `SendAddEditDialog` to have display logic for both the Send form and the password generator, and for it to reach deeply into its children to update the form with the generated password. To me this seems sub-optimal and is only necessary because we can only have one drawer open at a time. I wonder if there's a conversation to be had with UIF about this, maybe allowing for multiple drawers but stack them on top of each other so that only one is visible at a time?

That aside I have two suggestions but neither are blocking, they're more for consistency/code flow.

PR #19525, opened 2026-03-12
[PM-29524]Remove ts strict ignore in collect autofill content service

dan-livefront:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-29524
## 📔 Objective

This PR removes  ts strict ignore in collect autofill content service while avoiding scope keep and maintaining the changes local and pragmatic


apps/browser/src/autofill/services/collect-autofill-content.service.ts
```diff
@@ -519,13 +527,13 @@ export class CollectAutofillContentService implements CollectAutofillContentServ
    * Returns the attribute of an element as a lowercase value.
    * @param {ElementWithOpId<FormFieldElement>} element
    * @param {string} attributeName
-   * @returns {string}
+   * @returns {string | null}
```
bensbits91: Should this jsdoc tag be `@returns {string | undefined}`? 

If `getPropertyOrAttribute` returns `null`, `null?.toLowerCase()` evaluates to `undefined` (I think).

dan-livefront: Ope! updated here [update jsdoc](https://github.com/bitwarden/clients/pull/19525/commits/32ce6df7d7951367fe61d6806891fdc3d8eb5a84)

bensbits91 (dismissed):

bensbits91 (approved):

PR #19535, opened 2026-03-12
Uif/cl 205/add banner slot to bit layout

iivins-livefront:
## Tracking:

https://bitwarden.atlassian.net/browse/CL-205

## Overview:

`bit-layout` had no dedicated slot for banners and it forced developers to use negative margin overrides on `bit-banner`. This PR eliminates that coupling by adding a proper content projection slot to `bit-layout` that sits above the padded `<main>` within the CSS grid column.

## Key Changes

- Added a banner content projection slot to `bit-layout` that renders above the padded main area
- Slot collapses to zero height when no banner is projected, leaving unaffected layouts visually unchanged
- Removed the negative margin host class from `banner.component`
- Moved `app-vault-banners` from `vault.component` into `user-layout.component`
- Wired `app-vault-banners` into the new banner slot via the `slot="banner"` attribute
- Removed the stale padding-sync warning comment from the layout template
- Added a `WithBanner` Storybook story to document and verify the slot behaviour

## Summary

The banner slot is a minimal change to the layout component that removes a the coupling between `bit-layout` and `bit-banner`. Usage instances no longer need to worry about negative margins around the banner

## Screenshots

Screenshot 2026-03-12 at 5 25 12 PM


iivins-livefront:
> I'm seeing some diffs in Chromatic that look unintended for the kitchen sink layout spacing. I also know there are other places we use banners besides the web vault, and often those are nested a bit far down, which makes the top-level `user-layout` slot approach kind of complicated because we won't want to lift everything to the top level. I know we had a couple other solution ideas for this in the ticket -- what do you think?

Are you talking about the spacing when nested in the screenshot? That part was intended to level it out / even the vertical spacing - but I didn't see any additional "if nested do this" conditionals to add

It matches the Figma designs more than the Kitchen Sink test from what I'm seeing - the action slot has the top spacing if used but I don't think that's _technically_ wrong here ( unless I'm not understanding it right )

Screenshot 2026-03-16 at 5 21 28 PM

Screenshot 2026-03-16 at 5 14 15 PM

vleague2:
@iivins-livefront Yeah the one in the Kitchen sink where there is spacing above and to the sides of the banner. Afaik we want the banner to be edge-to-edge always, so the extra spacing wouldn't be desirable here. I haven't looked at the code enough to see why this is happening on Kitchen Sink and not in the bit-layout story you added.

Screenshot 2026-03-17 at 9 38 36 AM

For the other banner locations, I see some for admin console and some owned by the auth and billing teams. These are the ones where I'm not sure we'll be able to lift them all the way up to the layout level?

PR #19537, opened 2026-03-12
[PM-33526] honor hidden password permission in bw list items and bw get item

vlad-trofimov:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-4288

## 📔 Objective

The CLI's CipherResponse class was not respecting the viewPassword permission flag when serializing cipher data for output. The constructor calls new LoginExport(o.login) which unconditionally serializes all login fields including password, totp, and password history, regardless of collection-level access controls. This allows users with "View items, hide passwords" collection permission to see plaintext passwords and TOTP secrets via bw list items or bw get item, bypassing the intended security model.

This is not acceptable because:
- It violates collection-level access control policies set by organization admins
- It creates a security vulnerability where restricted users can access sensitive data through the CLI that they cannot access through other clients

This PR adds explicit redaction logic in the CipherResponse constructor that checks o.viewPassword === false and sets this.login.password = null and this.login.totp = null when the flag indicates passwords should be hidden. Password history is now only populated when o.viewPassword !== false. The logic uses strict equality checks (=== false and !== false) to ensure safe fail-open behavior for personal vault items where viewPassword defaults to true.

This is acceptable because:
- It enforces collection permissions consistently with other Bitwarden clients
- The fail-safe approach only redacts when viewPassword is explicitly false, preserving normal behavior for personal vault items
- It's a minimal, single-point fix in the CLI response layer that doesn't affect shared export functionality

## 📸 Screenshots

https://github.com/user-attachments/assets/c0db4c70-6beb-447c-bd27-e1353bd6aae3


apps/cli/src/vault/models/cipher.response.ts
```diff
@@ -30,11 +30,17 @@ export class CipherResponse extends CipherWithIdExport implements BaseResponse {
       this.creationDate = o.creationDate;
     }
     this.deletedDate = o.deletedDate;
-    if (o.passwordHistory != null) {
+    if (o.passwordHistory != null && o.viewPassword !== false) {
       this.passwordHistory = o.passwordHistory.map((h) => new PasswordHistoryResponse(h));
     }
     if (o.type === CipherType.Login && o.login != null) {
       this.login = new LoginResponse(o.login);
```
gbubemismith: Nice change. I think I'll prefer if we redact before the LoginResponse is fully constructed. Consider moving the redaction logic into LoginResponse itself by passing viewPassword as a constructor parameter. 

``` ts
this.login = new LoginResponse(o.login, o.viewPassword);
```

gbubemismith (changes requested):

apps/cli/src/vault/models/cipher.response.ts
vlad-trofimov: Thanks for the feedback! I agree this is cleaner. To fully implement this, I'll need to update both files:

  - login.response.ts — add viewPassword?: boolean to the constructor and apply
  redaction there
  - cipher.response.ts — pass o.viewPassword to the constructor and remove the
  post-construction redaction block

will submit commit momentarily

gbubemismith (approved):
Thanks for working on this. Changes look good

gbubemismith:
I'll move this to QA to test

gbubemismith:
@vlad-trofimov Thanks for working on this. Everything looks good 👍🏼

PR #19555, opened 2026-03-13
[PM-29525] Typefixes in services/autofill-overlay-content.service.ts

jprusik:
## 🎟️ Tracking

[PM-29525](https://bitwarden.atlassian.net/browse/PM-29525)

## 📔 Objective

Typing fixes, also resolves some console errors

[PM-29525]: https://bitwarden.atlassian.net/browse/PM-29525?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/browser/src/autofill/services/autofill-overlay-content.service.ts
+++ b/apps/browser/src/autofill/services/autofill-overlay-content.service.ts
@@ -1,3 +1,5 @@
+// FIXME: Update this file to be type safe and remove this and next line
+// @ts-strict-ignore
 import "@webcomponents/custom-elements";
 import "lit/polyfill-support.js";
 import { FocusableElement, tabbable } from "tabbable";


apps/browser/src/autofill/services/autofill-overlay-content.service.ts
```diff
@@ -1785,8 +1787,11 @@ export class AutofillOverlayContentService implements AutofillOverlayContentServ
     await this.updateMostRecentlyFocusedField(this.mostRecentlyFocusedField);
 
     const focusedFieldRectsTop = this.focusedFieldData?.focusedFieldRects?.top;
-    const focusedFieldRectsBottom =
-      focusedFieldRectsTop + this.focusedFieldData?.focusedFieldRects?.height;
+    const focusedFieldRectsHeight = this.focusedFieldData?.focusedFieldRects?.height;
+    if (!focusedFieldRectsTop || !focusedFieldRectsHeight) {
```
bensbits91, line 1791: Will `!focusedFieldRectsTop` incorrectly return `false` when `focusedFieldData?.focusedFieldRects?.top === 0`? Should it be more explicit, like `focusedFieldRectsTop == null`?

dan-livefront: Good call out, since 0 is falsy, a valid `top 0` could technically be treated as missing. It’s a pretty narrow case though, the field would have to be exactly at the top of the viewport and a reposition such as a scroll or resize would need to happen, since this check only matters during reposition and not on the initial open. That said, if we want to be extra cautious we can switch to an explicit == null check now.

I am ok keeping it as is but not opposed to being more explicit, dealers choice, Jon!

bensbits91: Makes sense. I'm fine with it as-is 👍

dan-livefront: I went ahead and added your rec on [null check](https://github.com/bitwarden/clients/pull/19562/commits/5437f2b4ac4306b638162a3c1953fa348135412b) on a separate PR that has these changes @bensbits91

PR #19557, opened 2026-03-13
[PM-24476] At Risk Password setting

nick-livefront:
## 🎟️ Tracking

[PM-24476](https://bitwarden.atlassian.net/browse/PM-24476)

## 📔 Objective

Adds a `Vault Customization` setting that will toggle the visibility of change password notifications for premium users
- Only visible for premium users
- Defaults to `true`

## 📸 Screenshots

<video src="https://github.com/user-attachments/assets/8440b73b-246f-4324-bbca-8495bcdcb226" />

[PM-24476]: https://bitwarden.atlassian.net/browse/PM-24476?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/common/src/vault/abstractions/vault-settings/vault-settings.service.ts
```diff
@@ -21,6 +21,11 @@ export abstract class VaultSettingsService {
    * for Autofill suggestions.
    */
   abstract clickItemsToAutofillVaultView$: Observable<boolean>;
+  /**
+   * An observable monitoring the state of the visibility of at-risk password
+   * notifications.
+   */
+  abstract showAtRiskPasswordNotifications$: Observable<boolean>;
```
shane-melton, line 32: ⛏️ (non-blocking) Can we add some more detail to this setting's description. I believe it would be beneficial to highlight that it only applies to items that are evaluated as at-risk in "realtime"; and not for organization owned ciphers that have been explicitly assigned a task.

There is often confusion on the difference between these two types of at-risk notifications because they present themselves identically to the enduser.

libs/vault/src/cipher-view/cipher-view.component.ts
```diff
@@ -264,9 +266,17 @@ export class CipherViewComponent {
   );
 
   readonly showChangePasswordLink = computed(() => {
-    return this.hasLoginUri() && (this.hadPendingChangePasswordTask() || this.passwordIsAtRisk());
+    return (
+      this.hasLoginUri() &&
+      (this.hadPendingChangePasswordTask() ||
+        (this.passwordIsAtRisk() && this.showAtRiskPasswordNotifications()))
+    );
   });
```
shane-melton, line 277: ⛏️ (non-blocking): Similarly here, more explanation on the expected logic could be beneficial for the next person to come through here and see this chain of logic.

shane-melton (dismissed):
Changes look good and thanks for additional tests! My only suggestions are around documentation clarity.

nick-livefront:
@shane-melton Added comments, let me know what you think about their wording!

shane-melton (dismissed):
Perfect, thanks!

pamperer562580892423:
This new menu entry is called "Show at-risk password **notifications**". Wouldn't it make more sense to add that to the "Settings --> **Notifications**" page then, rather than to "Appearance"? 🤔 (if BW discussed this internally before and I'm just late, then please ignore my unqualified remark)

JaredSnider-Bitwarden (dismissed):
Approving auth test file change

libs/common/src/vault/services/vault-settings/vault-settings.service.ts
```diff
@@ -52,14 +53,22 @@ export class VaultSettingsService implements VaultSettingsServiceAbstraction {
...
     );
 
+  private showAtRiskPasswordNotificationsState: ActiveUserState<boolean> =
+    this.stateProvider.getActive(SHOW_AT_RISK_PASSWORD_NOTIFICATIONS);
+  /**
+   * {@link VaultSettingsServiceAbstraction.showAtRiskPasswordNotifications$}
+   */
+  readonly showAtRiskPasswordNotifications$: Observable<boolean> =
+    this.showAtRiskPasswordNotificationsState.state$.pipe(map((x) => x ?? true));
+
```
jaasen-livefront, line 71: 🎨 Not a blocker but we could use `shareReplay({ bufferSize: 1, refCount: false }),` to avoid multiple reads from our subscribers.

jaasen-livefront (dismissed):
Nice work! Just one small suggestion.

nick-livefront:
> This new menu entry is called "Show at-risk password **notifications**". Wouldn't it make more sense to add that to the "Settings --> **Notifications**" page then, rather than to "Appearance"? 🤔 (if BW discussed this internally before and I'm just late, then please ignore my unqualified remark)

Hi @pamperer562580892423! No worries, a valid question! This setting in under the "Appearance > Vault customization" because it alters the way the vault presents information to the user. The Notification settings are oriented towards notifications that a user would see popup on a webpage, like when a "save item to vault" message is shown after logging in.

pamperer562580892423:
> Hi @pamperer562580892423! No worries, a valid question! This setting in under the "Appearance > Vault customization" because it alters the way the vault presents information to the user. The Notification settings are oriented towards notifications that a user would see popup on a webpage, like when a "save item to vault" message is shown after logging in.

@nick-livefront Thanks for your response! - And who am I to criticize this... on the other hand: though I can understand that reasoning from a developers perspective, I'm not sure a regular user would follow that logic, but perceive this as a "notification (setting)". - Or, I think, there are also valid reasons to see this as a "security setting", so to expect it in Settings --> "Account security" would also have made much sense to me.

Honestly, "Appearance" would have been the last tab I would have searched for such a "security notifications" (😉) setting... maybe... can't say exactly, now that I know. 😅

Anyway, thanks for incorporating that toggle at all now! 👍

libs/common/src/vault/services/vault-settings/vault-settings.service.ts
nick-livefront: Apologies, @jaasen-livefront I'm just seeing this. This has passed QA, it should be fine but to avoid anything unforeseen I'm going to leave as is.

nick-livefront:
@JaredSnider-Bitwarden @shane-melton @jaasen-livefront I had to fix a merged conflict, and this has passed QA!

JaredSnider-Bitwarden (approved):

shane-melton (approved):

PR #19559, opened 2026-03-13
[PM-33568] Add Claude guidance on injected content script exceptions

jprusik:
## 🎟️ Tracking

[PM-33568](https://bitwarden.atlassian.net/browse/PM-33568)

## 📔 Objective

Add Claude guidance on injected content script exceptions

[PM-33568]: https://bitwarden.atlassian.net/browse/PM-33568?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.claude/rules/autofill-content-scripts.md
+++ b/.claude/rules/autofill-content-scripts.md
@@ -1,6 +1,6 @@
 ---
 paths:
-  - "apps/browser/src/autofill/content/**/*.ts"
+  - "apps/browser/src/autofill/content/**"
 ---
 
 # Autofill Content Scripts - Critical Rules


jprusik:
> test

@claude what is this?

.claude/rules/autofill-content-scripts.md
audreyality, line 1: 🎨 Instead of adding `.claude` directories throughout the codebase, it might be a better idea to instead specialize using [path specific rules](https://code.claude.com/docs/en/memory#path-specific-rules). 

For the autofill content scripts directory:
```
---
paths:
  - "apps/browser/src/autofill/content/**/*.ts"
---
```

💭 Angular files use specialized extensions. That could be used to scope those rules, so they're not pulled into every context. If there's a skill for it, it might make sense to refer to the skill instead of stating the rules themselves. 

```
---
paths:
  - "**/*.module.ts"-   
  - "**/*.component.ts"
  - "**/*.component.html"
---

# Angular Usage

- Always use `Skill(angular-pattern-guide)` when editing angular components and modules.
```

🤔 I'm not sure how to scope the `BrowserApi` rule. Maybe a good way to do it is to hoist the content scripts into their own top-level directory? That way they're never targeted by the general-purpose rules at all?

```
---
paths:
  - "lib/**/*.ts"
  - "app/**/*.ts"
---
```

```
---
paths:
  - "content/**/*.ts"
---
```

cc: @SaintPatrck - this is probably something we should collaborate to figure out.

.claude/CLAUDE.md
```diff
@@ -40,7 +40,7 @@ main directories are:
 cross-dependencies that violate the intended modular structure. Always consult and respect the
 dependency rules defined in `eslint.config.mjs`, `nx.json`, and other configuration files.
 
-## Angular Architecture Patterns
+## Angular Architecture Patterns (Angular app and library code only)
```
theMickster: I don't think this change is needed. Claude knows what is Angular code and what is not. 
If you feel that it does need to be here (because you have often noticed Claude making bad recommendations), then I am not so sure that only this change will do. We may need to be more stern and direct with Claude and/or point to more specific definitions in the file.

.github/CODEOWNERS
```diff
@@ -154,11 +154,12 @@ nx.json @bitwarden/team-platform-dev
 
 ## Autofill team files ##
 apps/browser/src/autofill @bitwarden/team-autofill-dev
-apps/desktop/src/autofill @bitwarden/team-autofill-desktop-dev
 libs/common/src/autofill @bitwarden/team-autofill-dev
-apps/desktop/macos/autofill-extension @bitwarden/team-autofill-desktop-dev
 .github/workflows/test-browser-interactions.yml @bitwarden/team-autofill-dev
+
 # Autofill desktop files
+apps/desktop/macos/autofill-extension @bitwarden/team-autofill-desktop-dev
```
theMickster: Will you please consider this change for another PR? The current work is Claude focused for the Autofill team and this smells like something out of scope IMO

.github/CODEOWNERS
```diff
@@ -252,3 +254,4 @@ apps/desktop/native-messaging-test-runner/package-lock.json @bitwarden/team-plat
 .claude/ @bitwarden/team-ai-sme
 .github/workflows/respond.yml @bitwarden/team-ai-sme
 .github/workflows/review-code.yml @bitwarden/team-ai-sme
+apps/browser/src/autofill/content/.claude/CLAUDE.md @bitwarden/team-autofill-dev
```
theMickster: Two things. 
1. Are you 100% sure this is required? Are you sure the rule on line ~156 does not already cover this file?
2. I know we're are all about code ownership here, but I would like to discuss this before we go littering this file with 1-off Claude file ownership. That seems like unnecessary code churn IMO.

.claude/CLAUDE.md
jprusik: This was Claude guidance, so dropping it is fine with me

.github/CODEOWNERS
jprusik: > Are you sure the rule on line ~156 does not already cover this file?

It does not appear to, but I'm fine with deferring for now

.github/CODEOWNERS
jprusik: Yeah, this is just moving the autofill desktop rules down to their section, but I'm fine with deferring

.claude/rules/autofill-content-scripts.md
jprusik: The approach of localized guidance resonates with me for two reasons:

1) It gels with some of our recent training and recommendations on the tools
2) This is still useful contextual guidance for humans

That said, I don't feel strongly on this topic; these instructions are included in my local untracked rules.

.github/CODEOWNERS
jprusik: Per our offline conversation, I moved this up to the autofill team section, since GitHub did not pick up Autofill dev team ownership of `autofill/content/.claude/CLAUDE.md` (I added the team manually)

.claude/rules/autofill-content-scripts.md
audreyality: I prefer localized guidance as well. What I'm concerned about is how this scales.

* Adding `.claude` folders throughout the codebase makes it more difficult to find claude's rules.
* The angular and browser API patterns are not localized. We have them included in files where they aren't relevant (services, for example) and where they need to be actively countermanded (autofill directories).

Rules allow the angular guidance to target the specific files where it should be used. Placing the autofill-specific rules in the repo-root `.claude/rules` folder makes it easier to see the exceptions. They're not referenced within the folders where they matter, which is worse from a local discoverability point of view. On the other hand, their frontmatter allows them to be localized in practice by the tool most likely to read a rules file (Claude Code, as opposed to a tool like me 😉).

jprusik:
> **Inline finding for .github/CODEOWNERS line 160:**
> 
> ⚠️ **IMPORTANT**: CODEOWNERS entry is ineffective due to ordering regression
> 
> In the previous commit, this entry was placed **after** the general `.claude/ @bitwarden/team-ai-sme` rule at the end of the file (line 255), which correctly overrode it since CODEOWNERS uses last-matching-pattern precedence. The latest commit moved it to line 160, which is **before** the general `.claude/` rule on line 255.
> 
> As a result, `@bitwarden/team-ai-sme` now takes precedence as the code owner for `apps/browser/src/autofill/content/.claude/CLAUDE.md` instead of `@bitwarden/team-autofill-dev`.
> 
> **Fix:** Move this entry back to after line 255 (after the `.claude/` rule), or place it at the very end of the file.

I'm going to remove the rule entirely for now, but I'd like to find a way for teams to own claude guidance for their domain expertise/code ownership.

.claude/rules/autofill-content-scripts.md
SaintPatrck: Love the ideas getting thrown around. 

I agree with keeping things organized in the root `.claude/` dir.

🚀 This feels like the perfect opportunity to try out scoped rules. Would you be up for converting this into a rule scoped to `autofill/content/`? I don't think the granular filters are needed yet, given how small the rules are. If the rules continues to grow and/or starts negatively impacting context or behavior we can look in to optimizations. Thoughts? 

♻️ If Rules end up working well we can consider applying the same treatment to `apps/browser/CLAUDE.md` and other nested `CLAUDE.md` files.

theMickster: I am in support of trying the autofill instructions out as scoped rules. Please keep 👀 on Claude as you work and provide feedback in Slack whether Claude follows these or not. 

Let's pump the breaks on moving nested `CLAUDE.md` files at the time being. I would rather us spend effort maintaining the content in the files instead of reorganizing and tinkering with rules + paths.

PR #19562, opened 2026-03-13
[PM-33555]Remove @ts-strict-ignore in autofill overlay content service

dan-livefront:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-33555
## 📔 Objective

- Make userFilledFields nullable and guard reads/writes
- Add stable listener wrappers for keyup/focus/input so add/remove use the same ref
- Return HTMLElement | null from submit button finders and guard usage
- Tighten event/DOM narrowing (KeyboardEvent, FocusEvent, InputEvent, HTMLElement)
- Cast throttled handlers for useEventHandlersMemo compatibility
- Improve null handling in focus/submit helpers
- Fix getBoundingClientRectFromIntersectionObserver rect typing
- Guard isIgnoredField when field type is null
- Update spec expectations for nullable state and listener cleanup


apps/browser/src/autofill/services/autofill-overlay-content.service.ts
```diff
@@ -1788,7 +1822,7 @@ export class AutofillOverlayContentService implements AutofillOverlayContentServ
 
     const focusedFieldRectsTop = this.focusedFieldData?.focusedFieldRects?.top;
     const focusedFieldRectsHeight = this.focusedFieldData?.focusedFieldRects?.height;
-    if (!focusedFieldRectsTop || !focusedFieldRectsHeight) {
+    if (focusedFieldRectsTop == null || focusedFieldRectsHeight == null) {
```
jprusik, line 1825: because 0 ><

jprusik (dismissed):

jprusik (approved):

PR #16801, opened 2025-10-09
[PM-25206] Inject service instead of passing as param

quexten:
## 🎟️ Tracking
https://bitwarden.atlassian.net/browse/PM-25206
(This is part of a PR chain, and just a small clean up to make follow-up PRs nicer)

## 📔 Objective
Passing services in function args is an anti-pattern. It should be provided in the constructor / via DI instead.
## ⏰ Reminders before review

- Contributor guidelines followed
- All formatters and local linters executed and passed
- Written new unit and / or integration tests where applicable
- Protected functional changes with optionality (feature flags)
- Used internationalization (i18n) for all UI strings
- CI builds passed
- Communicated to DevOps any deployment requirements
- Updated any necessary documentation (Confluence, contributing docs) or informed the documentation team

## 🦮 Reviewer guidelines

- 👍 (`:+1:`) or similar for great changes
- 📝 (`:memo:`) or ℹ️ (`:information_source:`) for notes or general info
- ❓ (`:question:`) for questions
- 🤔 (`:thinking:`) or 💭 (`:thought_balloon:`) for more open inquiry that's not quite a confirmed issue and could potentially benefit from discussion
- 🎨 (`:art:`) for suggestions / improvements
- ❌ (`:x:`) or ⚠️ (`:warning:`) for more significant problems or concerns needing attention
- 🌱 (`:seedling:`) or ♻️ (`:recycle:`) for future improvements or indications of technical debt
- ⛏ (`:pick:`) for minor or nitpick changes

PR #19567, opened 2026-03-14
Handle unordered environment update events on logout

### LINKED ISSUE #18291: Browser Extension looses self-hosted URL, switches back to bitwarden.com
### Steps To Reproduce

Logging in to our self-hosted Bitwarden Instance worked fine in the past and basically works fine also now. But from time to time, the Browser Extension seems to loose it's memory about our own Bitwarden instance and switches back to bitwarden.com.

When this occurs, I can switch back to "self-hosted", enter "https://our.vault.domain.tld/" and save. Then it immediately shows "bitwarden.com" again - and at the same time the green "Environment URLs saved" banner.

Only thing to fix it temporarily is to restart Chrome, then it's working again.

### Expected Result

The Browser Extension memorizes permanently the choice made where to log in - and if that fails for whatever reason shows an error message.

### Actual Result

As described, after some seconds of delay it switches back to bitwarden.com instead of authenticating against our Bitwarden instance.

### Screenshots or Videos

_No response_

### Additional Context

The Setup (Mac, Chrome, Bitwarden Browser-Extension) worked flawlessly for the past months. A change of internet provider and/or VPN connection might have happened roughly at the same time where these issues started to show.

Is the Browser Extension somehow trying to reach the Bitwarden system to check whether it's reachable at all (and when it fails then "blocks" by silently falling back to bitwarden.com)?

### Operating System

macOS

### Operating System Version

Tahoe 26.1

### Web Browser

Chrome

### Browser Version

143.0.7499.170

### Environment Versions

impossible as I cannot log in to Bitwarden's Browser Extension right now.

Chrome v143.0.7499.170 (on MacOS)
Bitwarden Browser Extension v2025.12.0

### Issue Tracking Info

trmartin4:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-30715
fixes #18291

## 📔 Objective

`DefaultEnvironmentService.environment$` switches between `USER_ENVIRONMENT_KEY` and `GLOBAL_ENVIRONMENT_KEY` based on `activeAccountId$` - whether it is `null` or not. 

During logout, `USER_ENVIRONMENT_KEY` is cleared by `stateEventRunnerService.handleEvent("logout")` _before_ `accountService.clean()` fully propagates the `activeAccountId$` → `null` transition to the popup. Because the background and popup are separate JS contexts sharing `chrome.storage`, these writes arrive as `chrome.storage.onChanged` events with no ordering guarantee.

When `USER_ENVIRONMENT_KEY` clears first, `environment$` is still subscribed to user state (because account hasn't switched to `null` yet) and receives `null`, which `buildEnvironment(null, null)` resolves to the `US` default. 

When the login page loads in this state, any call to `setEnvironment(EU)` writes to `GLOBAL_ENVIRONMENT_KEY`, but `environment$` is watching user state, so the write produces no emission and the selector immediately reverts.

TL;DR: User state can become `null` while the account is still "active".

#### ❓ Why did this start with `2025.12.0` release?
The current theory is that https://github.com/bitwarden/clients/pull/16801 changed where the lock logic is defined - into the `LockService`, and that changed the likelihood of this occurring.  This fix feels like a more maintainable solution than relying on ordering of state being cleared, which seems brittle.

### Changes

`default-environment.service.ts`: `environment$` and `cloudWebVaultUrl$`
Both observables had the same pattern: subscribe to user-scoped state when a user is active and global-scoped state when not. Neither handled the case where user state is null while a user is still "active" per `activeAccountId$`.

This change added an inner `switchMap` that treats null user-scoped state as a signal to fall back to global state rather than passing `null` to `buildEnvironment`.

This is safe because `seedUserEnvironment` is always called before `switchAccount` during login ([login.strategy.ts:197-199](vscode-webview://10gbah056qtms93e7jo7m53jdau5t5teaju36ode42c788nvt4qh/libs/auth/src/common/login-strategies/login.strategy.ts#L197)), so user-scoped state is never legitimately `null` for an active logged-in user — null can only mean the logout cleanup arrived out of order.

`default-environment.service.spec.ts`
Added two regression tests:

- Verify `environment$` falls back to global state (`EU`) when user-scoped state is set to `null` while a user is still active.
- Verifies that a `setEnvironment(EU)` call during this window sticks (writes to global state, which `environment$` is now observing)

Both tests fail without the fix and pass with it.

## 📸 Screenshots

https://github.com/user-attachments/assets/2eb76323-a08f-45bd-89a4-a0e3947c978b

### DIFF
--- a/libs/common/src/platform/services/default-environment.service.ts
+++ b/libs/common/src/platform/services/default-environment.service.ts
@@ -153,6 +153,13 @@ export class DefaultEnvironmentService implements EnvironmentService {
       .getGlobal(GLOBAL_ENVIRONMENT_KEY)
       .state$.pipe(map((state) => this.buildEnvironment(state?.region, state?.urls)));
 
+    // The current environment emitted by environment$ can come from two sources: the "global" environment,
+    // which represents the environment used by the application before a user authenticates, and the "user" environment,
+    // which represents the environment of an authenticated account using the application.
+    // We preference the user environment, only falling back to the global environment if a user is not logged in.
+    // However, there does exist the possibility that - due to timing of state updates - the active userId state may
+    // be cleared after the user environment state.
+    // To handle this, we fall back to the global state as a last resort, if the user state is not set.
     const globalEnvState$ = this.stateProvider.getGlobal(GLOBAL_ENVIRONMENT_KEY).state$;
     this.environment$ = account$.pipe(
       switchMap((userId) => {


trmartin4:
@justindbaur tagging you for your experience with `EnvironmentService` and `StateProvider` to make sure that this wasn't something you suggest handling a different way.

PR #11900, opened 2024-11-07
[PM-14252] Switch to oo7 and drop libsecret

quexten:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-14252

## 📔 Objective

Oo7 (https://github.com/bilelmoussaoui/oo7) is the library for interacting with desktop Linux keyrings that many modern projects are switching towards. It supports accessing secret service directly, but also going via the secret portal in sandboxes (for flatpak/snap).
Note: Migration code is included. If there are clients that are on snap / flatpak, they will migrate from storing directly in secret service to the file backend + portal based approach.

The API is a lot more sane and the library is pure rust, which should reduce bugs we have with linux platform secrets management. Also drops dependency on compiling against c libraries.

(Also should make cross compiling for ARM possible)

## ⏰ Reminders before review

- Contributor guidelines followed
- All formatters and local linters executed and passed
- Written new unit and / or integration tests where applicable
- Protected functional changes with optionality (feature flags)
- Used internationalization (i18n) for all UI strings
- CI builds passed
- Communicated to DevOps any deployment requirements
- Updated any necessary documentation (Confluence, contributing docs) or informed the documentation team

## 🦮 Reviewer guidelines

- 👍 (`:+1:`) or similar for great changes
- 📝 (`:memo:`) or ℹ️ (`:information_source:`) for notes or general info
- ❓ (`:question:`) for questions
- 🤔 (`:thinking:`) or 💭 (`:thought_balloon:`) for more open inquiry that's not quite a confirmed issue and could potentially benefit from discussion
- 🎨 (`:art:`) for suggestions / improvements
- ❌ (`:x:`) or ⚠️ (`:warning:`) for more significant problems or concerns needing attention
- 🌱 (`:seedling:`) or ♻️ (`:recycle:`) for future improvements or indications of technical debt
- ⛏ (`:pick:`) for minor or nitpick changes

PR #2635, opened 2022-05-05
[BEEEP] [EC-141] Rust - Windows hello

Hinton:
## Type of change

- [ ] Bug fix
- [x] New feature development
- [ ] Tech debt (refactoring, code cleanup, dependency upgrades, etc)
- [ ] Build/deploy pipeline (DevOps)
- [ ] Other

## Objective

Follow up to the rust PR, which continues the migration away from our native modules to rust.

Replaces `@nodert-win10-rs4/windows.security.credentials.ui` and `forcefocus` with calling the proper Windows API using rust. The `UserConsentVerifierInterop` properly centers on the window and prevents blocking the UI.

## Before you submit

- [ ] I have checked for **linting** errors (`npm run lint`) (required)
- [ ] This change requires a **documentation update** (notify the documentation team)
- [ ] This change has particular **deployment requirements** (notify the DevOps team)

PR #19576, opened 2026-03-15
[PM-33582] Remove obsolete secure storage code

trmartin4:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-33582

## 📔 Objective

Removes obsolete checking in desktop native code that was checking for non-JSON values, as well as an obsolete method in the `unix.rs` for old secret data.

Since we were going to be testing secure storage, it felt like the appropriate time to clean both up.

📓 Read associated Jira ticket for background on removal timing.

Affected users would be:
- Windows desktop users who had not retrieved state stored in secure storage since July of 2022.
- Linux desktop users who had not retrieved state stored in secure storage since January of 2025.

### Timeline of this code
**Jul 25, 2022 — [0c1bd57c](https://github.com/bitwarden/clients/commit/0c1bd57c) (PR [#2635](https://github.com/bitwarden/clients/pull/2635) — EC-141 Rust Windows Hello)**
Switched from npm keytar to @bitwarden/desktop-native. For Windows, this switched from the secret being in UTF-8 to UTF-16.  This is also where the original existing comment and migration were introduced. `getPassword` got a try/catch fallback:
- Read with new native API.
- If not valid JSON, read via `getPasswordKeytar` (old UTF-8 path on Windows).
- Re-save in new format.

**Apr 18, 2023 — [830af7b0](https://github.com/bitwarden/clients/commit/830af7b0) (PR [#5234](https://github.com/bitwarden/clients/pull/5234) — Desktop Biometrics Rework)**
TypeScript migration fallback removed.  The `catch` block changed from "migrate and return" to `throw new Error("Password in bad format")`. `get_password_keytar` still existed in Rust but was now unreachable. The comment was left behind.

**Dec 4, 2024 — [80a898bd](https://github.com/bitwarden/clients/commit/80a898bd) (PR [#11900](https://github.com/bitwarden/clients/pull/11900) — Switch to oo7 and drop libsecret)**
`get_password_keytar` removed from Rust on all platforms. On Linux, get_password_legacy was introduced as a new migration shim — this time for libsecret → oo7, with the comment to remove after 2025.03.

### DIFF
--- a/apps/desktop/desktop_native/core/src/password/unix.rs
+++ b/apps/desktop/desktop_native/core/src/password/unix.rs
@@ -1,20 +1,12 @@
 use std::collections::HashMap;
 
 use anyhow::{anyhow, Result};
-use oo7::dbus::{self};
 use tracing::info;
 
 use crate::password::PASSWORD_NOT_FOUND;
 
 /// Retrieves a password from the Linux Secret Service keyring.
 pub async fn get_password(service: &str, account: &str) -> Result<String> {
-    match get_password_new(service, account).await {
-        Ok(res) => Ok(res),
-        Err(_) => get_password_legacy(service, account).await,
-    }
-}
-
-async fn get_password_new(service: &str, account: &str) -> Result<String> {
     let keyring = oo7::Keyring::new().await?;
     let _ = try_prompt(&keyring).await;
     let attributes = HashMap::from([("service", service), ("account", account)]);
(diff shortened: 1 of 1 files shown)


quexten:
@trmartin4 Since the write path already used only oo7 for a while, I anticipate there to not be much risk, but could we QA flatpak, snap *and* (some) unsandboxed install?

All of these work differently / the sandboxing uses the portal backend for oo7, the unsandboxed versions do not.

djsmith85 (approved):

PR #19588, opened 2026-03-16
[PM-27125] Add cookies to fetch requests

djsmith85:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-27125

## 📔 Objective

Includes work from https://github.com/bitwarden/clients/pull/19492

- Register a class during main process to listen to storages changes within `SERVER_COMMUNICATION_CONFIGS` and when present attach cookies to any fetch request using the Electron session. (`apps/desktop/src/main/sso-cookie.main.ts`)

- Register middleware to act on fetch redirects to acquire cookies for SSO (`libs/common/src/platform/services/server-communication-config/default-server-communication-config.service.ts`)

### DIFF
--- a/libs/common/src/platform/services/server-communication-config/default-server-communication-config.service.ts
+++ b/libs/common/src/platform/services/server-communication-config/default-server-communication-config.service.ts
@@ -98,11 +98,11 @@ export class DefaultServerCommunicationConfigService implements ServerCommunicat
         return response;
       }
       const needsBootstrap = await firstValueFrom(this.needsBootstrap$(hostname));
-
-      if (needsBootstrap) {
-        await this.acquireCookie(hostname);
+      if (!needsBootstrap) {
+        return response;
       }
 
+      await this.acquireCookie(hostname);
       // Retry with original request (follow redirect mode, cookies now in session)
       return next(request);
     };
--- a/libs/common/src/platform/services/server-communication-config/default-server-communication-config.service.spec.ts
+++ b/libs/common/src/platform/services/server-communication-config/default-server-communication-config.service.spec.ts
@@ -173,10 +173,9 @@ describe("DefaultServerCommunicationConfigService", () => {
       expect(result).toBe(redirect);
     });
 
-    it("retries without acquireCookie on opaqueredirect when bootstrap not needed", async () => {
+    it("returns opaqueredirect response early without retry when bootstrap not needed", async () => {
       const redirect = { status: 0, type: "opaqueredirect" } as Response;
-      const ok = { status: 200, type: "default" } as Response;
-      next.mockResolvedValueOnce(redirect).mockResolvedValueOnce(ok);
+      next.mockResolvedValueOnce(redirect);
       mockClientInstance.needsBootstrap.mockResolvedValue(false);
 
       const result = await middleware(new Request("https://vault.acme.com/"), next);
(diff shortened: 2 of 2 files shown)


libs/common/src/platform/services/server-communication-config/default-server-communication-config.service.ts
```diff
@@ -74,5 +78,32 @@ export class DefaultServerCommunicationConfigService implements ServerCommunicat
...
+    await new Promise((resolve) => setTimeout(resolve, 100));
+  }
+
+  private buildRedirectMiddleware(): FetchMiddleware {
+    return async (request: Request, next: FetchFn): Promise<Response> => {
+      const manualRequest = new Request(request.clone(), { redirect: "manual" });
+      const response = await next(manualRequest);
+
+      const isRedirect =
+        response.type === "opaqueredirect" || (response.status >= 300 && response.status < 400);
```
coroiu: suggestion: we should probably remove the status checks, the status is never returned and is always zero from what I can see


```suggestion
        response.type === "opaqueredirect";
```

libs/common/src/platform/services/server-communication-config/default-server-communication-config.service.ts
```diff
@@ -74,5 +78,32 @@ export class DefaultServerCommunicationConfigService implements ServerCommunicat
...
+
+      if (!isRedirect) {
+        return response;
+      }
+
+      const hostname = new URL(request.url).hostname;
```
coroiu: question: why this and not `Utils.getDomain`?

coroiu: issue: I think we need to add some protection here

- `new URL("about:blank").hostname` returns an empty string, we should probably not try to fetch cookies for that

libs/common/src/platform/services/server-communication-config/default-server-communication-config.service.ts
djsmith85: Addressed with 70aa9c7

libs/common/src/platform/services/server-communication-config/default-server-communication-config.service.ts
djsmith85: Addressed with 1c54d72

djsmith85: Addressed with 1c54d72
Needs to be the hostname, but now using the `Utils.getHostname`

libs/common/src/platform/services/server-communication-config/default-server-communication-config.service.ts
```diff
@@ -74,5 +79,32 @@ export class DefaultServerCommunicationConfigService implements ServerCommunicat
...
+        return response;
+      }
+      const needsBootstrap = await firstValueFrom(this.needsBootstrap$(hostname));
+
+      if (needsBootstrap) {
+        await this.acquireCookie(hostname);
+      }
+
+      // Retry with original request (follow redirect mode, cookies now in session)
+      return next(request);
```
coroiu, line 107: issue: sorry but I just noticed this: the middleware doesn't actually use `needsBootstrap` to check if we should retry, meaning that redirects in environments that don't use the ELB will always retry. We should fix this

```suggestion
  private buildRedirectMiddleware(): FetchMiddleware {
    return async (request: Request, next: FetchFn): Promise<Response> => {
      const manualRequest = new Request(request.clone(), { redirect: "manual" });
      const response = await next(manualRequest);

      if (response.type !== "opaqueredirect") {
        return response;
      }

      const hostname = Utils.getHostname(request.url);
      if (Utils.isNullOrWhitespace(hostname)) {
        return response;
      }

      const needsBootstrap = await firstValueFrom(this.needsBootstrap$(hostname));
      if (!needsBootstrap) {
        return response;
      }

      await this.acquireCookie(hostname);
      
      // Retry with original request (follow redirect mode, cookies now in session)
      return next(request);
```

djsmith85: Addressed with 601bc51

PR #19593, opened 2026-03-16
[PM-32180] Change when premium upgrade happens

jengstrom-bw:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-32180

## 📔 Objective

Pull the shared logic of when to show the premium upgrade prompt to a service.

### DIFF
--- a/libs/angular/src/vault/services/premium-upsell.service.spec.ts
+++ b/libs/angular/src/vault/services/premium-upsell.service.spec.ts
@@ -6,11 +6,10 @@ import { AccountService } from "@bitwarden/common/auth/abstractions/account.serv
 import { BillingAccountProfileStateService } from "@bitwarden/common/billing/abstractions";
 import { FeatureFlag } from "@bitwarden/common/enums/feature-flag.enum";
 import { ConfigService } from "@bitwarden/common/platform/abstractions/config/config.service";
+import { mockAccountServiceWith } from "@bitwarden/common/spec";
 import { UserId, CipherId } from "@bitwarden/common/types/guid";
 import { CipherService } from "@bitwarden/common/vault/abstractions/cipher.service";
 import { CipherData } from "@bitwarden/common/vault/models/data/cipher.data";
-
-import { mockAccountServiceWith } from "@bitwarden/common/spec";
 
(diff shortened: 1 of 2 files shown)


libs/angular/src/vault/services/premium-upsell.service.ts
```diff
@@ -0,0 +1,58 @@
...
+  private readonly activeUserId = toSignal(this.activeUserId$, { requireSync: true });
+  private hasPremium$ = this.activeUserId$.pipe(
+    switchMap((userId) => {
+      return this.billingAccountService.hasPremiumFromAnySource$(userId);
+    }),
+  );
+  private ciphers$ = this.cipherService.ciphers$(this.activeUserId());
+
+  private readonly ciphers = toSignal(this.ciphers$, { initialValue: {} });
+  private readonly hasPremium = toSignal(this.hasPremium$);
```
jaasen-livefront: ❌  we'll want to assign this with an initial value of `true` to avoid potentially flashing the premium upsell for premium users.

libs/angular/src/vault/services/premium-upsell.service.spec.ts
```diff
@@ -0,0 +1,159 @@
...
+
+import { AccountService } from "@bitwarden/common/auth/abstractions/account.service";
+import { BillingAccountProfileStateService } from "@bitwarden/common/billing/abstractions";
+import { FeatureFlag } from "@bitwarden/common/enums/feature-flag.enum";
+import { ConfigService } from "@bitwarden/common/platform/abstractions/config/config.service";
+import { UserId, CipherId } from "@bitwarden/common/types/guid";
+import { CipherService } from "@bitwarden/common/vault/abstractions/cipher.service";
+import { CipherData } from "@bitwarden/common/vault/models/data/cipher.data";
+
+import { mockAccountServiceWith } from "../../../../../libs/common/spec";
```
jaasen-livefront: 🎨 we can use an alias here

```suggestion
import { mockAccountServiceWith } from "@bitwarden/common/spec";
```

jengstrom-bw:
@jaasen-livefront Got those suggestions in, thanks!

PR #19598, opened 2026-03-16
SSH Agent v2: use in electron services

neuronull:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-30754

## 📔 Objective

- adds logic to utilize the desktop native's v2 napi exposure of the ssh agent to the main and render ssh agent services.
    - this was done in as additive a way as possible, to simplify readability and the eventual FF removal and v1 removal later.
- added unit test for the v2 implementation
- moved the IPC definitions for preload, out of Platform space and into @bitwarden/team-desktop-native-dev  ownership.

## Testing done

- confirmed list-keys still works as expected for v1
- confirmed by manually overriding the v2 feature flag, that the v2 code path is executed (which we expect to panic due to it not being implemented yet)

// ssh-add -L asked the BW agent

```
debug1: Next authentication method: publickey
debug3: ssh_get_authentication_socket_path: path '/Users/foo/.bitwarden-ssh-agent.sock'
debug1: get_agent_identities: bound agent to hostkey
debug1: get_agent_identities: agent returned 1 keys
debug1: Will attempt key: key1 ED25519 SHA256:hiYKWXr9u1wMsdMOolfZL9E+++Bmj9/eDSXri9DbMNM agent
debug1: Will attempt key: /Users/foo/.ssh/id_rsa RSA SHA256:8pF78Q+6S0E/i0u3/27s+BZdcDWwPdnIAhb/FgVkA50 explicit
debug2: pubkey_prepare: done
debug1: Offering public key: key1 ED25519 SHA256:hiYKWXr9u1wMsdMOolfZL9E+++Bmj9/eDSXri9DbMNM agent
debug3: send packet: type 50
debug2: we sent a publickey packet, wait for reply
debug3: receive packet: type 51
debug1: Authentications that can continue: publickey
```

// expected panic when v2 FF enabled
```
[Elec] thread 'tokio-runtime-worker' (2097706) panicked at /Users/foo/workspace/bw/clients/apps/desktop/desktop_native/ssh_agent/src/server/mod.rs:48:9:
[Elec] not yet implemented
[Elec] note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
[Elec] 15:41:43.873 › [NAPI] [DEBUG] desktop_napi::sshagent_v2::sshagent_v2: Signaling the agent to start the server.
[Elec] 15:41:43.873 › [NAPI] [DEBUG] ssh_agent::agent: Starting the server.
[Elec] 15:41:43.874 › SSH agent v2 encountered an error:  Error: not yet implemented
```


addisonbeck (approved):

quexten (approved):

coltonhurst (approved):
Looks good, thanks @neuronull 👍

coltonhurst:
One note:

In reviewing this I found we have at least 3 patterns for IPC on desktop; ideally we would only have one. If there are reasons for multiple, then we should have better documentation on which pattern developers should choose based on the pros and cons for each.

I've created a ticket for investigating this here: https://bitwarden.atlassian.net/browse/PM-34860

(Please note this PR seems to correctly follow existing SSH v1 patterns, so this comment is more of a bird's eye view comment about the desktop app in general, not a comment about this PR. Placing this here just because I did some digging during the review of this PR.)

neuronull:
> One note:
> 
> In reviewing this I found we have at least 3 patterns for IPC on desktop; ideally we would only have one. If there are reasons for multiple, then we should have better documentation on which pattern developers should choose based on the pros and cons for each.
> 
> I've created a ticket for investigating this here: https://bitwarden.atlassian.net/browse/PM-34860
> 
> (Please note this PR seems to correctly follow existing SSH v1 patterns, so this comment is more of a bird's eye view comment about the desktop app in general, not a comment about this PR. Placing this here just because I did some digging during the review of this PR.)

Good eye! I aimed to make this change as non-intrusive as possible to make the v2 effort as isolated to the agent implementation as possible, so opted to keep parity of the v1 patterns.

neuronull:
QA has validated these changes.

PR #19599, opened 2026-03-16
[PM-30101] subscription discounts in web checkout

kdenney:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-30101

## 📔 Objective

Add discount support to checkout for all checkout flows that offer premium and families plans.

## 📸 Screenshots

Discount badge added to cart summary component:
(used in several checkout flows)
image

And with multiple discounts:
image

Upgrade checkout flow with families:
(This is used by the nav button, upgrade dialog, new subscription page, etc.)
image

Create families organization from vault:
image

Trial initiation with 0-day trial:
image

And with 7-day trial:
image

And with multiple discounts eligible:
image


sukhleenb (dismissed):

eliykat (dismissed):

apps/web/src/app/billing/services/subscription-discount.service.ts
```diff
@@ -0,0 +1,80 @@
...
+
+    return featureFlag$.pipe(
+      switchMap((featureFlagEnabled) => {
+        if (!featureFlagEnabled) {
+          return of([]);
+        }
+
+        return merge(
+          this.fetchDiscounts$(),
+          this.refreshTrigger.pipe(switchMap(() => this.fetchDiscounts$(true))),
```
amorask-bitwarden, line 43: ❓/ ⛏️  This would result in two concurrent API calls doing the same thing off the bat right since we're using a `BehaviorSubject` right? I think if we used a `Subject` instead, it'd only emit on `.next()` which allow us to  prevent the duplicate call. Not a blocker.

apps/web/src/app/billing/services/subscription-discount.service.ts
```diff
@@ -0,0 +1,80 @@
...
+      map((discounts) => {
+        return discounts.filter((d) => d.tierEligibility[tier]);
+      }),
+    );
+
+  isDiscountExpiredError(error: unknown): boolean {
+    return (
+      error instanceof ErrorResponse &&
+      error.statusCode === 400 &&
+      error.message === DISCOUNT_EXPIRED_MESSAGE
```
amorask-bitwarden, line 60: 📓 Just noting that a modification to this error message in `server` will break the toast-based error handling + refresh for the various invocations of this method. Not a blocker, but something we'll need to keep in mind if Product or Design ever wants to change the error message.

libs/common/src/billing/enums/discount-tier-type.enum.ts
```diff
@@ -0,0 +1,6 @@
+export const DiscountTierType = {
```
amorask-bitwarden: ⛏️ [ADR](https://contributing.bitwarden.com/architecture/adr/ts-deprecate-enums#the-enum-like-pattern) calls for using `Object.freeze`, but it's the first time I've seen it so certainly non-blocking.

apps/web/src/app/billing/trial-initiation/trial-billing-step/trial-billing-step.component.ts
```diff
@@ -99,10 +141,17 @@ export class TrialBillingStepComponent implements OnInit, OnDestroy {
             !!billingAddress.country && !!billingAddress.postalCode,
         ),
       ),
+      this.eligibleDiscounts$,
     ]).pipe(
       debounceTime(500),
-      switchMap(([cadence, billingAddress]) =>
-        this.trialBillingStepService.getCosts(product, tier, cadence, billingAddress),
+      switchMap(([cadence, billingAddress, discounts]) =>
```
amorask-bitwarden: ⛏️ `discounts` is unused here.

amorask-bitwarden (dismissed):
Great work - everything makes sense to me. I only had a few picks if you want to address them (non-blocking) and one overarching question. Coupons are not used in the Premium > Organization upgrade path, which I assume is by design given the nature of the coupons we're rolling out, but I wanted to make sure that was the case.

kdenney:
> Great work - everything makes sense to me. I only had a few picks if you want to address them (non-blocking) and one overarching question. Coupons are not used in the Premium > Organization upgrade path, which I assume is by design given the nature of the coupons we're rolling out, but I wanted to make sure that was the case.

@amorask-bitwarden Yeah I realized partway through working this story that the premium>org upgrade path was not included in the ticket. I raised it with Micah and he said he'd create another ticket for us to circle back and add that support later.

apps/web/src/app/billing/services/subscription-discount.service.ts
kdenney: Yes and I don't love that we are looking for that exact error string but it's an edge case so hopefully it's very rare. And the good news is that the error string from the server is never displayed to the user so we can change what is displayed without having to change any server code to update that message.

apps/web/src/app/billing/trial-initiation/trial-billing-step/trial-billing-step.component.ts
kdenney: Good catch, thanks. Removed in https://github.com/bitwarden/clients/pull/19599/commits/12d189a482fe7bce9fc38996c0d47a3f3e9fafdc

libs/common/src/billing/enums/discount-tier-type.enum.ts
kdenney: Good call. Added in https://github.com/bitwarden/clients/pull/19599/commits/12d189a482fe7bce9fc38996c0d47a3f3e9fafdc

apps/web/src/app/billing/services/subscription-discount.service.ts
kdenney: Nice catch! I forgot that about BehaviorSubject so I'm glad you mentioned it. Changed to Subject in https://github.com/bitwarden/clients/pull/19599/commits/12d189a482fe7bce9fc38996c0d47a3f3e9fafdc.

eliykat (approved):

amorask-bitwarden (approved):

PR #19606, opened 2026-03-17
[PM-33781] fix(fido2): fall back to browser for hardware/CTAP1 keys not in vault

### LINKED ISSUE #19280: `wordpress.com/log-in/webauthn` does not accept CTAP1 keys registered via `my.wordpress.com/me/security/two-step-auth`.
### Steps To Reproduce

1. Register a (CTAP1) U2F key-pair, at [`my.wordpress.com/me/security/two-step-auth`](https://my.wordpress.com/me/security/two-step-auth#:~:text=browser%2E%20Learn%20more-,Your%20security%20keys,-Register%20key).

1. Attempt to authenticate, at [`wordpress.com/log-in/webauthn`](https://wordpress.com/log-in/webauthn).

### Expected Result

The authentication should be successful.

### Actual Result

Authentication fails:

<p></p>Image<p></p>

### Screenshots or Videos

https://github.com/user-attachments/assets/e986c953-0457-4cdd-89bc-ca488580748e

### Additional Context

[`issues/8971`](https://github.com/bitwarden/clients/issues/8971#issuecomment-2104844822) consistently reproduces, too.

### Operating System

Linux

### Operating System Version

[`fedora-release-identity-kde-desktop-43`](http://mirror.slu.cz/fedora/linux/updates/43/Everything/x86_64/Packages/f/fedora-release-identity-kde-desktop-43-26.noarch.rpm)

### Web Browser

Firefox

### Browser Version

[`https://packages.mozilla.org/rpm/firefox/Packages/6c1b3699b02b5c789f1f90413aa359ae54d47644d573bdf36fe374cd2674d5a8-firefox-nightly-150.0a1-20260226092447.x86_64.rpm`](https://packages.mozilla.org/rpm/firefox/Packages/6c1b3699b02b5c789f1f90413aa359ae54d47644d573bdf36fe374cd2674d5a8-firefox-nightly-150.0a1-20260226092447.x86_64.rpm)

### Environment Versions

<blockquote>

~~~YAML
Version: 2026.1.1
SDK: 'main (8e5efd8)'
Server version: 2026.2.0
~~~

</blockquote>

### Issue Tracking Info

nancysangani:
## 🎟️ Tracking
Fixes #19280
Internal: https://bitwarden.atlassian.net/browse/PM-32855

## 📔 Objective
When all entries in `allowCredentials` explicitly specify non-`internal` transports (e.g. `usb`, `nfc`), the credentials cannot be Bitwarden passkeys. This PR makes the extension fall back immediately to the native browser WebAuthn handler in that case, allowing hardware/CTAP1 keys registered outside Bitwarden to work correctly.
The approach is safe as the RP already knows the transports of their own credentials — no vault information is leaked. As noted by the maintainer, if an RP does not send correct transports a Bitwarden passkey would not be usable, but this is accepted as an RP issue.


jprusik (dismissed):

micahblut:
We will not be accepting this contribution. Introducing this behavior allows a compromised site to enumerate what credentials a user does or does not have stored in their Bitwarden vault, and this violates Bitwarden's security objectives.

nancysangani:
@micahblut Thank you for the feedback. I understand the credential enumeration concern with the vault-lookup based fallback. I've revised the approach to only fall back based on the `transports` field provided by the RP in `allowCredentials` — specifically when all entries explicitly specify non-`internal` transports (e.g. `usb`, `nfc`). This is safe because the RP already knows the transports of their own credentials, so no vault information is leaked. Would you be open to reviewing a revised PR with this approach?

micahblut:
> @micahblut Thank you for the feedback. I understand the credential enumeration concern with the vault-lookup based fallback. I've revised the approach to only fall back based on the `transports` field provided by the RP in `allowCredentials` — specifically when all entries explicitly specify non-`internal` transports (e.g. `usb`, `nfc`). This is safe because the RP already knows the transports of their own credentials, so no vault information is leaked. Would you be open to reviewing a revised PR with this approach?

Yes! Having reviewed that approach internally we're happy to review a revised PR with that approach. If an RP does not send the correct transports, a passkey stored in Bitwarden would not be usable, but we accept this as unlikely and fixable by the RP.

nancysangani:
@micahblut I've updated the PR to remove the vault-lookup based fallback as discussed. The fix now only falls back to the native browser handler when all `allowCredentials` entries explicitly specify non-`internal` transports (e.g. `usb`, `nfc`), which is safe as the RP already knows its own credential transports. Please review it when you get a chance. Thanks!

libs/common/src/platform/services/fido2/fido2-client.service.ts
```diff
@@ -325,6 +325,19 @@ export class Fido2ClientService<
...
+    if (
+      params.allowedCredentialIds.length > 0 &&
+      transports != null &&
+      transports.length === params.allowedCredentialIds.length &&
+      transports.every((t) => t.length > 0 && !t.includes("internal"))
+    ) {
+      this.logService?.info(
+        `[Fido2Client] All allowCredentials entries specify non-internal transports only — falling back to browser.`,
+      );
+      throw new FallbackRequestedError();
```
coroiu, line 344: suggestion: consider adding a comment to explain why we are doing this. We're gonna need this when we migrate away from our custom TS code to `passkey-rs`

```suggestion
      // Spec reference: https://www.w3.org/TR/webauthn-3/#sctn-discover-from-external-source step 20.
      // The spec says the client should determine if "no authenticator will become available" by
      // examining transports. Since Bitwarden is an internal-only authenticator, if all
      // allowCredentials entries specify only non-internal transports, we cannot satisfy the
      // request. We throw FallbackRequestedError (instead of the spec's NotAllowedError) to let
      // the browser's native WebAuthn handler contact the hardware authenticator.
      this.logService?.info(
        `[Fido2Client] All allowCredentials entries specify non-internal transports only — falling back to browser.`,
      );
      throw new FallbackRequestedError();
```

libs/common/src/platform/services/fido2/fido2-client.service.spec.ts
```diff
@@ -688,6 +729,7 @@ describe("FidoAuthenticatorService", () => {
     function createParams(params: Partial<AssertCredentialParams> = {}): AssertCredentialParams {
       return {
         allowedCredentialIds: params.allowedCredentialIds ?? [],
+        allowedCredentialTransports: params.allowedCredentialTransports,
```
coroiu: suggestion: merge these into one object.

----

The spec models `allowCredentials` as a list of `PublicKeyCredentialDescriptor` objects where each descriptor bundles `{id, type, transports}` together (see [section 5.8.3](https://www.w3.org/TR/webauthn-3/#dictionary-credential-descriptor)). This PR introduces `allowedCredentialTransports` as a separate parallel array alongside the existing `allowedCredentialIds`, which means the two arrays must be kept in sync by convention.

The reason `allowedCredentialIds` is a flat `string[]` today is that we previously only needed the IDs from each descriptor. Now that we also need `transports`, we should take the opportunity to align with the spec's structure rather than adding a second parallel array. Consider replacing both with a single list of descriptor objects in `AssertCredentialParams`:

```typescript
export interface AssertCredentialParams {
  allowedCredentials: { id: string; transports?: ("ble" | "hybrid" | "internal" | "nfc" | "usb")[] }[];
  // ... rest of fields
}
```

This would eliminate the need for the defensive `transports.length === params.allowedCredentialIds.length` check, and it mirrors the spec's `PublicKeyCredentialDescriptor` structure more closely. The mapping in `webauthn-utils.ts` would become a single `.map()` call producing `{ id, transports }` objects instead of two separate `.map()` calls.

nancysangani:
@coroiu Updated as suggested — replaced the parallel `allowedCredentialIds` / `allowedCredentialTransports` arrays with a single `allowedCredentials` descriptor list matching the WebAuthn spec structure. Also added a comment explaining the transport-based fallback logic referencing the W3C spec. Please review it when you get a chance. Thanks!

libs/common/src/platform/services/fido2/fido2-client.service.spec.ts
nancysangani: Updated as suggested — replaced the parallel `allowedCredentialIds` / `allowedCredentialTransports` arrays with a single `allowedCredentials` descriptor list matching the WebAuthn spec structure. Also added a comment explaining the transport-based fallback logic referencing the W3C spec.

libs/common/src/platform/services/fido2/fido2-client.service.ts
nancysangani: Added the comments!

jprusik (approved):

coroiu (approved):
Looks good, thank you for the change! I've marked this PR as ready for QA

bensbits91 (approved):

kmancusi0105:
@nancysangani , hello!
Currently we have a bug involving Safari.
In short, when I go to authenticate with a passkey that isn't in the vault, the extension prompt still displays and I never get the browser prompt displaying
1. Set up the extension, but don't log in
2. Go to webauthn.io and register a passkey to Yubikey or other hardware
3. Log into the extension
4. Fill in the passkey value and click 'Authenticate'
Expected: the browser prompt would authenticate your passkey
Actual: the Bitwarden prompt displays and closing it **doesn't** fallback to the browser

https://github.com/user-attachments/assets/a704116d-2890-4266-99ac-baecea46c2af

coroiu:
Merged the PR as it passed QA. Thank you for your contribution! 🎉

PR #19607, opened 2026-03-17
[PM-14820] CLI: Replace legacy generator service with credential generator service

harr1424:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-14820

## 📔 Objective

The CLI needs to be migrated to the new credential generator service. 

Doing so allows the CLI  to work with all credential types (usernames, emails, and passwords).

It also enables the CLI to validate its inputs against generator policy.

The generator should continue using the flags and commands it already uses, extending the existing logic to include the new capabilities listed above.


apps/cli/src/tools/generate.command.ts
```diff
@@ -1,130 +1,427 @@
...
+      if ("calibrate" in policyConstraints) {
+        adjustResult = policyConstraints.calibrate(requestedSettings).adjust(requestedSettings);
+      } else {
+        adjustResult = policyConstraints.adjust(requestedSettings);
+      }
+      const policyAdjustedSettings = adjustResult.state;
+      const applied = adjustResult.applied;
+
+      // Hard engine limits cannot be overridden, even with --force
+      if (hardLimits.length > 0) {
```
mcamirault, line 83: ⛏️ Small optimization, we can do this check and return right after line 77's call right? No reason to do more work if the hard limits have already been exceeded.

mcamirault (dismissed):
One tiny optimization opportunity but you can take it or leave it. Looks good otherwise, especially the plethora of new tests!

mcamirault (dismissed):

libs/tools/generator/core/src/policies/dynamic-password-policy-constraints.ts
```diff
@@ -60,10 +60,26 @@ export class DynamicPasswordPolicyConstraints implements DynamicStateConstraints
...
   calibrate(state: PasswordGeneratorSettings): StateConstraints<PasswordGeneratorSettings> {
     // decide which constraints are active
-    const lowercase = state.lowercase || this.constraints.lowercase?.requiredValue || false;
-    const uppercase = state.uppercase || this.constraints.uppercase?.requiredValue || false;
-    const number = state.number || this.constraints.number?.requiredValue || false;
-    const special = state.special || this.constraints.special?.requiredValue || false;
+    const lowercase =
+      state.lowercase ||
+      this.constraints.lowercase?.requiredValue ||
+      this.constraints.minLowercase?.min > 0 ||
```
audreyality: ⚠️ Introducing the `minLowerCase.min` and related checks on the other lines disables the `maybe` checks on lines 86-92. This results in a user being unable to disable a checkbox when there's a value in the spinner box. The appropriate behavior is for the checkbox to always override the presence of a minimum value.

audreyality (changes requested):

libs/common/src/tools/types.ts
```diff
@@ -93,6 +93,11 @@ export type WithConstraints<State> = {
 
   /** the constraints enforced upon the type. */
   readonly constraints: Constraints<State>;
+
+  /** constraints that were actually applied (i.e. changed a value).
+   *  When absent, no constraints were applied.
+   */
+  readonly applied?: Constraints<State>;
```
audreyality: 🎨 I suggest always having an object here in order to ease `undefined` checks everywhere in the codebase. These types are all designed to be immutable, so you could create a placeholder type, freeze it, and write it into the `applied` field to avoid allocation overhead.

```ts
// in a utility file

// freezing `EMTPY` protects all callers from aliased edits creating spooky action at a distance.
const EMPTY = Object.freeze({});

/** The empty set of constraints */
export function unconstrained<T>() : Constraints<T> {
   return EMPTY;
}

// at usage site
const foo: WithConstraints<State> = { ..., applied: unconstrained<State>() };
```

🤔 It may be necessary for the body of `unconstrained` to apply a type assertion. Prefer `satisfies` if the compiler lets you do it, as that will fail to compile if the compiler can't prove that `T` allows unconstrained.

libs/common/src/tools/state/user-state-subject.ts
```diff
@@ -44,7 +44,7 @@ import { isDynamic } from "./state-constraints-dependency";
 import { UserStateSubjectDependencies } from "./user-state-subject-dependencies";
 import { UserStateSubjectDependencyProvider } from "./user-state-subject-dependency-provider";
 
-type Constrained<State> = { constraints: Readonly<Constraints<State>>; state: State };
+type Constrained<State> = WithConstraints<State>;
```
audreyality: ⛏️ These types are identical. You should replace `Constrained<State>` with `WithConstraints<State>` to avoid tech debt from an unnecessary alias.

libs/common/src/tools/state/user-state-subject.ts
```diff
@@ -292,23 +292,20 @@ export class UserStateSubject<
         if (!loadedState && !this.objectKey?.initial) {
           this.log.debug("no value; bypassing adjustment");
           return {
-            constraints: {} as Constraints<State>,
+            constraints: {},
             state: null,
-          } satisfies Constrained<State>;
+          } as Constrained<State>;
```
audreyality: ⚠️ This significantly weakens the type guarantee here; `satisfies` protects you against accidental incorrect constraint definitions.

🎨 It's a place where the placeholder I noted would provide proper type safety.

apps/cli/src/tools/generate.command.ts
```diff
@@ -1,130 +1,419 @@
...
-      ),
-    );
-    const enforcedOptions = shouldEnforceOptions
-      ? (await this.passwordGenerationService.enforcePasswordGeneratorPoliciesOnOptions(options))[0]
-      : options;
-
-    const password = await this.passwordGenerationService.generatePassword(enforcedOptions);
-    const res = new StringResponse(password);
-    return Response.success(res);
+  async run(cmdOptions: GenerateCmdOptions): Promise<Response> {
```
audreyality, line 51: 💭 I didn't take a close look at this logic the first time through. Now when I look at it it seems very strange to me. 

It basically inlines all of the operations from `CrendentialGeneratorService.generate$`, replicating most of the things it already handles. This includes a full reimplementation of the algorithm selection logic, including its metadata, in `resolveAlgorithm`. (`resolveAlgorithm` is necessary; I'm observing that it could be more tightly focused on returning a `CredentialAlgorithm` that is forwarded to `generate$`.)

The policy logic has a similar issue. It's performed by `settingsSubject` and can be read out using `settingsSubject.withConstraints$`. This logic extracts the value from the subject, loads the policy manually out of the metadata, and invokes the constraints instead.

❓ What decision(s) drove this implementation? Are there specific reasons why `generate$` and `withConstraints$` didn't work? If so, what are they?

We can take this conversation offline. It's probably easier to explore this in a meeting than through code review, especially since it's largely driven by curiosity. 🙂

🤔 There's also a code smell involved in the `try` block starting on line 70. If you work within the observable, then an error tears down the pipeline. That ensures that your `try` is limited to a single line--an `await firstValueFrom(yourPipeline$)`.

libs/common/src/tools/types.ts
```diff
@@ -72,6 +72,14 @@ export type Constraints<T> = {
   [Key in keyof T]?: Simplify<Constraint<T[Key]>>;
 };
 
+/** Utility type to provide an empty set of constraints
+ * Freezing `EMPTY` protects all callers from aliased edits creating spooky action at a distance.
+ */
+const EMPTY = Object.freeze({});
+export function unconstrained<T>(): Constraints<T> {
+  return EMPTY;
+}
```
audreyality: ⛏️ This type is missing documentation; did you mean to attach the JSDoc to it, and not `EMPTY`?

⛏️ This file contains types; you've added data and a function. Those probably are better suited for a nearby `util.js`.

libs/common/src/tools/state/user-state-subject.ts
```diff
@@ -292,17 +290,18 @@ export class UserStateSubject<
         if (!loadedState && !this.objectKey?.initial) {
           this.log.debug("no value; bypassing adjustment");
           return {
+            state: null as State,
```
audreyality, line 294: 💭 When I wrote this I remember being upset that I needed to `null as State` for the typecheck to work. I see now that the reason that's necessary is because `WithConstraints<>.state` does not allow `null` and wonder how that might be addressed. It's out of scope of this work, and also an interesting type problem.

libs/common/src/tools/types.ts
audreyality: 📝 Just a note - these nits are blocking my ✅. The other stuff might be worth addressing, but I don't expect it to be.

mcamirault: Done

audreyality (approved):

apps/cli/src/tools/generate.command.ts
audreyality: > What decision(s) drove this implementation?

Per conversation offline, the likely cause is limited shelf-life of the code as SDK integration proceeds.

👍🏻 I'm leaving the feedback unresolved for reference purposes; it doesn't block review. Thanks for circling back to getting the CLI using the new logic!

itsadrago (approved):

itsadrago (approved):
## Summary

**PR: CLI Generator Migration to CredentialGeneratorService**

**Risk assessment:** Low-Medium. The CLI command is a self-contained consumer; shared library changes (`WithConstraints`, constraint classes) have broader impact but are well-tested.

### What this PR does

- Replaces the legacy `PasswordGenerationServiceAbstraction` with the new `CredentialGeneratorService` in the CLI `generate` command
- Extends the CLI to support username, catchall email, subaddress email, and forwarded email generation (in addition to existing password/passphrase)
- Modifies the `StateConstraints` interface so `adjust()` and `fix()` return `WithConstraints<State>` (including the new `applied` field) instead of bare `State`, enabling the CLI to report which specific settings were changed by policy
- Adds the `unconstrained()` utility for a shared frozen empty constraints object
- Fixes a correctness issue in `PasswordPolicyConstraints` where include-flags (e.g., `special`) are now derived from the post-constraint `minSpecial` value rather than the pre-constraint `state.minSpecial`, ensuring policy-mandated minimums correctly activate their corresponding flags

### Review findings

No critical or important issues found. The prior reviews by @audreyality and @mcamirault were thorough, and the identified issues have been addressed. Key observations:

- **Security:** API keys for forwarder services are passed via CLI flags, which is standard CLI practice. No secrets are logged or persisted inappropriately.
- **Correctness:** Policy enforcement correctly distinguishes org policies (non-overridable) from default policies (overridable with `--force`). Hard engine limits are always enforced regardless of `--force`.
- **Breaking changes:** The `StateConstraints.adjust()` and `fix()` return type change is a breaking interface change, but all implementations in the codebase have been updated consistently.
- **Test coverage:** 597 lines of new tests cover password/passphrase/username/catchall/subaddress/forwarder flag mapping, policy enforcement, hard limits, saved settings merging, and error handling.
- **Resource management:** The `BehaviorSubject` created for `account$` is properly completed in a `finally` block.

**Verdict:** Clean PR. Well-structured migration with comprehensive test coverage.

audreyality (approved):

mcamirault (approved):

PR #19609, opened 2026-03-17
[CL-1119] toggle berry update

BryanCunningham:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/CL-1119

#####  Changes you may see in your code
- update to use the berry component inside toggles rather than the badge component
 
## 📔 Objective

- Replace instances of badge inside toggle to use berry
- Set variant based on selected state

## 📸 Screenshots
image

### DIFF
--- a/libs/components/src/stories/kitchen-sink/components/kitchen-sink-toggle-list.component.ts
+++ b/libs/components/src/stories/kitchen-sink/components/kitchen-sink-toggle-list.component.ts
@@ -10,11 +10,11 @@ import { KitchenSinkSharedModule } from "../kitchen-sink-shared.module";
   template: `
     <div class="tw-my-6">
       <bit-toggle-group [(selected)]="selectedToggle" aria-label="Company list filter">
-        <bit-toggle value="all"> All <span bitBadge variant="info">3</span> </bit-toggle>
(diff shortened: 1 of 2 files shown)


bitwarden_license/bit-web/src/app/admin-console/providers/manage/members.component.html
```diff
@@ -27,21 +27,15 @@
   >
     <bit-toggle [value]="undefined">
       {{ "all" | i18n }}
-      @if (dataSource.activeUserCount; as allCount) {
-        <span bitBadge variant="info">{{ allCount }}</span>
-      }
+      <bit-berry [value]="dataSource.activeUserCount" variant="primary"></bit-berry>
```
vleague2: For the consumer code, do we need to specify the variant since primary is the default already?

BryanCunningham: No, not really. We could remove it here if we want

BryanCunningham: Removed [here](https://github.com/bitwarden/clients/pull/19609/changes/548cc1c29d54fa8cd8a8e2f86e2fffc6b8ab4983)

PR #19611, opened 2026-03-17
[PM-33787] [Defect] Enable Premium Users to Upgrade to Free Organization

sbrown-livefront:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-33787

## 📔 Objective

Improves readability of `selectableProduct` and `selectablePlan` logic and allows premium user to upgrade to a free organization using the existing organization creation path

## 📸 Screenshots

https://github.com/user-attachments/assets/33f0499e-000f-4545-8955-5478bb8d931e

### DIFF
--- a/apps/desktop/desktop_native/ssh_agent/src/agent.rs
+++ b/apps/desktop/desktop_native/ssh_agent/src/agent.rs
@@ -31,8 +31,8 @@ where
 
 impl<K, H> BitwardenSSHAgent<K, H>
 where
-    K: KeyStore + Send + Sync,
-    H: ApprovalRequester,
+    K: KeyStore + Send + Sync + 'static,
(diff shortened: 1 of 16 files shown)


sbrown-livefront:
> This looks great. Only one question for me. We added this line to our `submit`: `if (canUpgradeFromPremium && this.selectedPlan()?.type !== PlanType.Free)`, which is correct. Do we need to do the same thing for `refreshSalesTax`? I'm thinking of this flow:
> 1. Select "Teams"
> 2. Fill out billing address (Sales tax fetched)
> 3. Switch to "Free" (Address still valid)
> 4. Sales tax fetch fails cuz we can't get tax for a free org

Great catch, I'll return early since org name could trigger it since address isn't visible. 👌🏾

PR #19614, opened 2026-03-17
remove default icons on non-semantic variants

BryanCunningham:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/CL-1118

## 📔 Objective

- Remove default badge icons on non-semantic variants

## 📸 Screenshots
image

### DIFF
--- a/libs/components/src/badge/badge.component.ts
+++ b/libs/components/src/badge/badge.component.ts
@@ -74,7 +74,7 @@ const commonStyles = [
 ];
 
 const defaultIconMap: Record<BadgeVariant, BitwardenIcon | null> = {
-  info: "bwi-info-circle",
+  info: null,
   subtle: null,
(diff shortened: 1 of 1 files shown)


BryanCunningham:
> Should info badges keep the icon?

@willmartian I'm not sure. Was a little bit torn on that myself

willmartian:
> > Should info badges keep the icon?
> 
> @willmartian I'm not sure. Was a little bit torn on that myself

I lean towards keeping them. It feels like a clear line we can draw: info is semantic, primary is not. Does that create any additional migration effort, etc?

BryanCunningham:
@willmartian Actually, the 'info' variant is legacy and will be removed. Being kept around for now to reduce churn in consumer code. We are good to remove the icon here

PR #19618, opened 2026-03-17
PM-27343 Mitigate build warnings

dan-livefront:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-27343
## 📔 Objective

Removed 267 build warnings by adding a `tsconfig.build.json` so TypeScript aligns with what Webpack is already bundling. This doesn’t change the build output, everything should continue compiling as expected, including all entry points. Also set asyncFunction: true in Webpack to remove the async/await and WASM warning. Open to other approaches if this one is not best in terms of maintainability. 

### Before 
Screenshot 2026-03-17 at 4 39 45 PM

### After 
Screenshot 2026-03-17 at 4 38 34 PM


apps/browser/webpack.base.js
```diff
@@ -372,6 +372,9 @@ module.exports.buildConfig = function buildConfig(params) {
       webassemblyModuleFilename: "assets/[modulehash].wasm",
       path: params.outputPath,
       clean: true,
+      environment: {
+        asyncFunction: true,
+      },
```
dan-livefront, line 377: set asyncFunction: true to remove the async/await and WASM warning

addisonbeck (approved):

PR #19622, opened 2026-03-17
[PM-27368] Injected Autofill experiences should respect the user's "show animations" setting

### LINKED ISSUE #17580: Reduce motion for autofill inline menu
### Steps To Reproduce

1. Go to a login page.
2. Click on the autofill button in the webpage textbox.
3. The autofill iframe and menu show transitions.

### Expected Result

If the user has [prefers-reduced-motion](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/At-rules/@media/prefers-reduced-motion) enabled, transitions should be disabled for accessibility and OS preference.

### Actual Result

Autofill iframe and menu have transitions.

### Screenshots or Videos

_No response_

### Additional Context

_No response_

### Operating System

Windows, macOS, Linux

### Operating System Version

_No response_

### Web Browser

Chrome

### Browser Version

_No response_

### Environment Versions

Version: 2025.11.0

SDK: 'main (8ef7951)'

Server version: 2025.11.0

### Issue Tracking Info

jprusik:
## 🎟️ Tracking

[PM-27368](https://bitwarden.atlassian.net/browse/PM-27368)

## 📔 Objective

This PR adds granular service-level controls to `AnimationControlService` to inform a class of animations for injected content script experiences (see below). The states driving these controls are intentionally proxying the existing `enableRoutingAnimation$` state. This was done because the user setting ui controlling this state is presented as "Show animations" (`Settings -> Appearance`) ; by splitting the state concerns, we make future setting UI updates easy, while delivering upon user expectations with the current experience.

### Limitations and Additional notes

- These settings do not extend to SVG animations at this time (loaders, countdowns)
- Instead of pulling `AnimationControlService` into each of the affected autofill-related services, I chose to proxy the `AnimationControlService` via `AutofillService`, since that reduced the service injection footprint and more tightly scoped those particular state concerns to autofill (rather than animation settings; that is, we don't care about other animation settings in each of these contexts).
- We have some partial animation control in `fadeOutNotification` for notifications. This value is hardcoded and only controls fade _out_ animations, so is probably better represented by `enableNotificationAnimation$`. I opted to defer that task, however to keep the scope of this task small.
- Animation setting updates do not propgate to the injected experiences (notification, inline menu) until they are rebuilt. Since these experiences are emphemeral and affected primarily at their entry anyhow, it didn't seem worth the messaging complexity to update those experiences on the fly (the next inline menu or notification open will use the new setting value).

## 📸 Screenshots

https://github.com/user-attachments/assets/b4ee5dea-3240-4423-8bfc-e0a6bc46aad2

### Affected animations

| Animations ON | Animations OFF |
| --- | --- |
| autofill-on | autofill-off |
| inline-menu-on | inline-menu-off |
| notification-on | notification-off |

### **Unaffected** animations (these are NOT affected by the animation setting)

**Notification bar button loading state**
notification-action-button-loading

**Inline Menu Passkey loading state**
passkey-loading

**Inline Menu TOTP countdown**
totp-expiry

[PM-27368]: https://bitwarden.atlassian.net/browse/PM-27368?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/common/src/platform/abstractions/animation-control.service.ts
```diff
@@ -26,6 +44,11 @@ export class DefaultAnimationControlService implements AnimationControlService {
     map((state) => state ?? this.defaultEnableRoutingAnimation),
   );
 
+  // Proxying `enableRoutingAnimation` for now to enable future decoupling
+  enableAutofillAnimation$ = this.enableRoutingAnimation$;
+  enableNotificationAnimation$ = this.enableRoutingAnimation$;
+  enableInlineMenuAnimation$ = this.enableRoutingAnimation$;
```
jprusik: An additional upside to this approach; the owning team doesn't need to bring in the autofill team if/when discrete controls for these concerns are added.

apps/browser/src/autofill/background/overlay.background.spec.ts
```diff
@@ -195,6 +197,10 @@ describe("OverlayBackground", () => {
     inlineMenuFieldQualificationService = new InlineMenuFieldQualificationService();
     themeStateService = mock<ThemeStateService>();
     themeStateService.selectedTheme$ = selectedThemeMock$;
+    enableNotificationAnimationMock$ = new BehaviorSubject(true);
+    enableInlineMenuAnimationMock$ = new BehaviorSubject(true);
+    autofillService.enableNotificationAnimation$ = enableNotificationAnimationMock$;
+    autofillService.enableInlineMenuAnimation$ = enableInlineMenuAnimationMock$;
```
audreyality: ⛏️ These lines are populating a mock declared on line 175. Moving them near that line would make it more clear that they're not breaking encapsulation of the actual `AutofillService`.

~~🎨 `mock<T>(...)` supports passing an object. This can be used to populate the fields of `autofillService, enabling the fields to become readonly, better protecting encapsulation.~~

apps/browser/src/autofill/content/abstractions/autofill-init.ts
```diff
@@ -13,6 +13,7 @@ export type AutofillExtensionMessage = {
   subFrameUrl?: string;
   subFrameId?: string;
   pageDetailsUrl?: string;
+  showAnimations?: boolean;
```
audreyality, line 16: ♻️ There are 19 total properties on `AutofillExtensionMessage`, and all but 1 are optional. This type should probably be specialized. Consider replacing it with a [discriminated union](https://www.typescriptlang.org/docs/handbook/2/narrowing.html#discriminated-unions).

apps/browser/src/autofill/content/autofill-init.ts
```diff
@@ -131,7 +131,7 @@ class AutofillInit implements AutofillInitInterface {
     await this.sendExtensionMessage("updateIsFieldCurrentlyFilling", {
       isFieldCurrentlyFilling: true,
     });
-    await this.insertAutofillContentService.fillForm(fillScript);
+    await this.insertAutofillContentService.fillForm(fillScript, showAnimations);
```
audreyality: 🎨 Consider converting `showAnimations` to a boolean value to eliminate the `undefined` test cases.

```ts
    await this.insertAutofillContentService.fillForm(fillScript, !!showAnimations);
```

apps/browser/src/autofill/content/autofill-init.spec.ts
```diff
@@ -324,6 +325,7 @@ describe("AutofillInit", () => {
           );
           expect(autofillInit["insertAutofillContentService"].fillForm).toHaveBeenCalledWith(
             fillScript,
+            undefined,
```
audreyality: ❓ Should there be an autofill-init test that expects `true` as its 2nd argument?
♻️ If `insertAutofillContentService` were injected into `AutofillInit`, then this test could inspect the mock instead of breaking encapsulation.

apps/browser/src/autofill/notification/bar.ts
```diff
@@ -293,6 +293,7 @@ function handleCloseNotification(e: Event) {
   e.preventDefault();
   sendPlatformMessage({
     command: "bgCloseNotificationBar",
+    // @TODO This value should be replaced with the resolved user preference
```
audreyality: ❓ Was this `TODO` left in place intentionally? 
⛏️ If so replace with `FIXME`, which is a stronger indicator of technical debt. (It's used with disabled lint markers.)

apps/browser/src/autofill/overlay/inline-menu/abstractions/autofill-inline-menu-iframe.service.ts
```diff
@@ -5,6 +5,7 @@ export type AutofillInlineMenuIframeExtensionMessage = {
   portKey?: string;
   generatedPassword?: string;
   refreshPassword?: boolean;
+  showAnimations?: boolean;
```
audreyality, line 8: 🤔 Can this type declaration be made stronger? Having a definitely-typed value for the property can eliminate variation in default values possible when reading an optional field.

🎨 A static factory method can ensure a consistent value is initialized in the codebase.

📝 This also applies to other messages with an optional `showAnimations` member.

libs/common/src/platform/abstractions/animation-control.service.ts
```diff
@@ -8,6 +8,24 @@ export abstract class AnimationControlService {
...
+   * Whether notification bar animations are enabled.
+   * Currently backed by the routing animation state; will be decoupled in the future.
+   */
+  abstract enableNotificationAnimation$: Observable<boolean>;
+
+  /**
+   * Whether inline menu animations are enabled.
+   * Currently backed by the routing animation state; will be decoupled in the future.
+   */
+  abstract enableInlineMenuAnimation$: Observable<boolean>;
```
audreyality, line 30: 🎨 Consider explaining when these observables emit and/or complete in their documentation.

🥳 Fun fact! Duplicate emissions are among the most common case of spurious emissions in rxjs. You can deduplicate emissions using [distinctUntilChanged()](https://rxjs.dev/api/operators/distinctUntilChanged).

libs/common/src/platform/abstractions/animation-control.service.ts
audreyality:
```suggestion
  enableAutofillAnimation$ = this.enableRoutingAnimation$.pipe(distinctUntilChanged());
  enableNotificationAnimation$ = this.enableRoutingAnimation$.pipe(distinctUntilChanged());
  enableInlineMenuAnimation$ = this.enableRoutingAnimation$.pipe(distinctUntilChanged());
```

audreyality (dismissed):
A few nits; nothing blocking. Most of my comments are things to pay attention as the code continues evolving.

apps/browser/src/autofill/background/overlay.background.spec.ts
jprusik: I did attempt the `mock<T>(...)` pattern, but it doesn't appear to work for property values, only functions returning observables. I poked around the codebase and online for prior art/alternate patterns, but wasn't able to expediently find a working solution. If you have a working pattern, I'm down to implement it.

apps/browser/src/autofill/content/autofill-init.ts
jprusik: Good feedback - one small tweak. Because this setting is on by default, an undefined case should default to `true`, so `showAnimations ?? true` over `!!showAnimations`

apps/browser/src/autofill/notification/bar.ts
jprusik: The `@TODO` is [a jsdoc convention](https://jsdoc.app/tags-todo), but I see we've adopted to FIXME more widely, so I'll switch to that (I prefer the more authoritative language as well)

addisonbeck (dismissed):

apps/browser/src/autofill/background/overlay.background.spec.ts
audreyality: 🤦🏻‍♀️ Nope. You're right. The proxy it uses doesn't let it override properties.

apps/browser/src/autofill/notification/bar.ts
```diff
@@ -293,6 +293,10 @@ function handleCloseNotification(e: Event) {
   e.preventDefault();
   sendPlatformMessage({
     command: "bgCloseNotificationBar",
+    /**
+     * FIXME (PM-33879): This value should be replaced with the resolved
+     * user preference and/or removed entirely
+     */
```
audreyality: ⚠️ Avoid `/* */` within the body of a function; it can play havoc with nesting

```suggestion
    // FIXME (PM-33879): This value should be replaced with the resolved
    // user preference and/or removed entirely
```

audreyality (approved):

addisonbeck (approved):

PR #19624, opened 2026-03-18
[PM-33820] fix lastpass no elements error

itsadrago:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-33820

## 📔 Objective

When importing a LastPass vault (either through a csv file or direct import) a user will see the following error for any vaults that include a credit card item: "no elements in sequence". 

As a result of the import failure, no items will have been imported from the LastPass vault

The screenshot below was from a test done using the Chrome extension, although any client that offers a LastPass vault import is affected

## 📸 Screenshots

image


harr1424 (approved):

pamperer562580892423:
@itsadrago:

> When using the BW Chrome browser extension to direct import a LastPass vault, a user will see the following error for any vaults that include a credit card item: "no elements in sequence".

Sorry for hopping in here. As I read this, I just wanted to "throw in", that it seems at least also to be an issue when using the importer of the **web vault** (and not only an issue of the browser extensions), like this "report" suggests: https://community.bitwarden.com/t/trying-to-migrate-from-lastpass-to-bitwarden-every-import-attempt-fails-with-the-error-no-elements-in-sequence/94776 (but maybe you already had this on the radar...)

itsadrago:
> @itsadrago:
> 
> > When using the BW Chrome browser extension to direct import a LastPass vault, a user will see the following error for any vaults that include a credit card item: "no elements in sequence".
> 
> Sorry for hopping in here. As I read this, I just wanted to "throw in", that it seems at least also to be an issue when using the importer of the **web vault** (and not only an issue of the browser extensions), like this "report" suggests: https://community.bitwarden.com/t/trying-to-migrate-from-lastpass-to-bitwarden-every-import-attempt-fails-with-the-error-no-elements-in-sequence/94776 (but maybe you already had this on the radar...)

Yes, the issue here affects any clients that offer a LastPass vault import. I've changed the description to reflect that. Thanks for pointing this out!

PR #19644, opened 2026-03-18
[PM-33446] Ensure attachments are cleared on clone

shane-melton:
## 🎟️ Tracking

[PM-33446](https://bitwarden.atlassian.net/browse/PM-33446)

## 📔 Objective

Similar https://github.com/bitwarden/clients/pull/19642, we need to ensure existing attachments are cleared when performing a clone operation. This was a bug before SDK encryption, however the TS clients happily ignored the mismatch of original attachments with a new cipher; but it was ignored by the server (cloning attachments is not supported). However, when passing in the incorrectly "cloned" attachments with exiting key values to the SDK, the SDK expectedly fails to decrypt the original attachment key values with the newly generated cipher key. This rightfully throws an exception and prevents the bad attachment meta data being sent to the server (even though it would have been ignored).

## 📸 Screenshots

https://github.com/user-attachments/assets/887b5808-fa5b-4d32-8cac-5b696958e337

[PM-33446]: https://bitwarden.atlassian.net/browse/PM-33446?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/vault/src/cipher-form/components/cipher-form.component.spec.ts
+++ b/libs/vault/src/cipher-form/components/cipher-form.component.spec.ts
@@ -10,6 +10,7 @@ import { ConfigService } from "@bitwarden/common/platform/abstractions/config/co
 import { I18nService } from "@bitwarden/common/platform/abstractions/i18n.service";
 import { CipherArchiveService } from "@bitwarden/common/vault/abstractions/cipher-archive.service";
 import { Cipher } from "@bitwarden/common/vault/models/domain/cipher";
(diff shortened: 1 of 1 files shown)


shane-melton:
@nikwithak Agreed, it should probably be clearer that attachments will be dropped. I'll create a separate ticket for improving that UX and get feedback/input from design.

PR #19651, opened 2026-03-19
[PM-32453] handle different encodings for lastpass private key

itsadrago:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-32453

## 📔 Objective

The encoding of the LastPass private key is different depending on the kind of LastPass login (federated or not) used when importing a LastPass vault. With the fixes in this branch, the importer client now uses the appropriate decoder (Hex or Base64) depending on the LastPass login used.

Without the fixes in this branch, the importer client is only supporting hex encoding, which is used for non-federated logins. Currently, when a federated login is used to import a LastPass vault, the import process will fail with an error and no items will imported from the LastPass vault.

### DIFF
--- a/apps/browser/src/_locales/en/messages.json
+++ b/apps/browser/src/_locales/en/messages.json
@@ -6169,9 +6169,6 @@
   "enterMultipleEmailsSeparatedByComma": {
     "message": "Enter multiple emails by separating with a comma."
   },
-  "emailsRequiredChangeAccessType": {
-    "message": "Email verification requires at least one email address. To remove all emails, change the access type above."
-  },
   "emailPlaceholder": {
     "message": "user@bitwarden.com , user@acme.com"
   },
--- a/apps/browser/src/vault/popup/components/vault/intro-carousel/intro-carousel.component.ts
+++ b/apps/browser/src/vault/popup/components/vault/intro-carousel/intro-carousel.component.ts
@@ -24,14 +24,14 @@ import { IntroCarouselService } from "../../../services/intro-carousel.service";
   ],
 })
 export class IntroCarouselComponent {
-  protected itemTypes = ItemTypes;
-  protected loginCards = LoginCards;
-  protected noCredentials = NoCredentialsIcon;
-  protected secureDevices = DevicesIcon;
+  protected readonly itemTypes = ItemTypes;
+  protected readonly loginCards = LoginCards;
+  protected readonly noCredentials = NoCredentialsIcon;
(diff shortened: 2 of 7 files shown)


libs/importer/src/importers/lastpass/access/services/client.ts
```diff
@@ -63,8 +63,20 @@ export class Client {
       );
 
       let privateKey: Uint8Array = null;
+      let iv: Uint8Array = null;
```
harr1424: ⛏️  If I were to work on this in the future, I would be confused by this variable name. Does it represent incoming vault, import vector?

itsadrago: I also added more error checking. Take another look :)

libs/importer/src/importers/lastpass/access/services/client.ts
```diff
@@ -63,8 +63,40 @@ export class Client {
...
+          encryptedPrivateKey = Utils.fromHexToArray(session.encryptedPrivateKey);
+          initVec = key.subarray(0, 16);
+        } else if (clientInfo.login == LastpassLoginType.Federated) {
+          // LastPass private key format is !<base64>|<base64>
+          // Private key use AES-CBC encryption, with the first base64 string
+          // being the initialization vector and the second base64 string
+          // containing the actual encrypted key
+          const parts = session.encryptedPrivateKey.split("|");
+          if (parts.length !== 2) {
+            throw new Error("Wrong lastpass private key format, no | separator");
```
mcamirault: ⛏️ Can we capitalize LastPass and replace "Wrong" with "Invalid" here and in all places below?

libs/importer/src/importers/lastpass/access/models/client-info.ts
```diff
@@ -1,13 +1,19 @@
...
   description: string;
+  login: LastpassLoginType;
 
   static createClientInfo(id: string): ClientInfo {
-    return { platform: Platform.Desktop, id, description: "Importer" };
+    return {
+      platform: Platform.Desktop,
+      id,
+      description: "Importer",
+      login: LastpassLoginType.MasterPassword,
```
mcamirault: 🎨 Can we move this to a parameter instead? That way the logic for setting it can live in the LastPassDirectImportService `creatClientInfo` method alongside the ID logic instead of being a bit hidden in the`Vault`

itsadrago: Great suggestion. Done

PR #19655, opened 2026-03-19
[PM-29480] Add dialog to cookie acquisition

djsmith85:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-29480

## 📔 Objective

Majority of the changes are moving all logic regarding the SSO cookie acquisition and attaching to fetch requests into `apps/desktop` instead of `libs/common`.

Adds a dialog to inform the user that the browser will need to launch to retrieve a sso cookie. This is acquired by the user authenticating with their IdP and the cookie being sent back down to the app.

- Clicking on "Launch browser" will open a browser and direct the user to their Idp
- Clicking on "Later" will cancel all operations, but as this dialog is triggered by having an invalid or having no cookie attached to fetch requests, it might show again shortly after (multiple Api-calls or a background sync for example)

## 📸 Screenshots

image

### DIFF
--- a/apps/desktop/src/platform/services/server-communication-config/server-communication-config-platform-api.service.ts
+++ b/apps/desktop/src/platform/services/server-communication-config/server-communication-config-platform-api.service.ts
@@ -37,12 +37,7 @@ export class ServerCommunicationConfigPlatformApiService implements ServerCommun
     timeoutId: NodeJS.Timeout;
   } | null = null;
 
-  // Listen for callback messages from app.component.ts
-  private callbackSubscription: Subscription | null = this.messageListener
-    .messages$(SSO_COOKIE_VENDOR_CALLBACK_COMMAND)
-    .subscribe((msg) => {
-      this.handleCallback(msg.urlString);
-    });
+  private callbackSubscription: Subscription | null = null;
 
   private static readonly TIMEOUT_MS = 5 * 60 * 1000; // 5 minutes
 
@@ -52,6 +47,15 @@ export class ServerCommunicationConfigPlatformApiService implements ServerCommun
     private logService: LogService,
     private dialogService: DialogService,
   ) {}
+
+  // Listen for callback messages from app.component.ts
+  init() {
+    this.callbackSubscription = this.messageListener
+      .messages$(SSO_COOKIE_VENDOR_CALLBACK_COMMAND)
+      .subscribe((msg) => {
+        this.handleCallback(msg.urlString);
+      });
+  }
 
   async acquireCookies(vaultUrl: string): Promise<AcquiredCookie[] | undefined> {
     if (this.pendingHostname !== null) {
--- a/apps/desktop/src/platform/services/server-communication-config/default-server-communication-config.service.ts
+++ b/apps/desktop/src/platform/services/server-communication-config/default-server-communication-config.service.ts
@@ -5,11 +5,9 @@ import { ConfigService } from "@bitwarden/common/platform/abstractions/config/co
 import { SdkLoadService } from "@bitwarden/common/platform/abstractions/sdk/sdk-load.service";
 import { FetchMiddleware, FetchFn } from "@bitwarden/common/platform/misc/fetch-middleware";
 import { Utils } from "@bitwarden/common/platform/misc/utils";
-import {
-  ServerCommunicationConfigClient,
-  ServerCommunicationConfigPlatformApi,
-} from "@bitwarden/sdk-internal";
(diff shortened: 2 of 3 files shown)


apps/desktop/src/platform/services/server-communication-config/server-communication-config-platform-api.service.ts
```diff
@@ -0,0 +1,170 @@
...
+          "Cookie acquisition already in progress for hostname, returning existing promise",
+        );
+        return this.pendingPromise;
+      }
+      // Different hostname - cancel previous and start new
+      this.logService.warning("Cancelling previous cookie acquisition for different hostname");
+      this.cancelPendingAcquisition();
+    }
+
+    const normalizedVaultUrl = vaultUrl.startsWith("https://") ? vaultUrl : `https://${vaultUrl}`;
```
coroiu, line 74: suggestion: I think we can skip this. Either that or throw an error if the url doesn't start with `https://`. The way this is currently written we might end up with a url that looks like this: `https://http://vault.something.com`

apps/desktop/src/platform/services/server-communication-config/server-communication-config-platform-api.service.ts
```diff
@@ -0,0 +1,170 @@
...
+ * - Listens for callbacks via MessageListener subscribing to SSO_COOKIE_VENDOR_CALLBACK_COMMAND
+ * - Deduplicates concurrent calls across both the dialog phase and callback-waiting phase
+ * - 5-minute timeout for safety
+ * - Returns undefined on timeout/cancellation (not error)
+ *
+ */
+export class ServerCommunicationConfigPlatformApiService implements ServerCommunicationConfigPlatformApi {
+  // Tracks the hostname and shared promise for the full flow (dialog + callback).
+  // Used for deduplication: a second call for the same hostname returns this promise directly.
+  private pendingHostname: string | null = null;
```
coroiu, line 30: suggestion: consider renaming this to `pendingVaultUrl`

apps/desktop/src/platform/services/server-communication-config/server-communication-config-platform-api.service.ts
```diff
@@ -0,0 +1,170 @@
...
+        return this.pendingPromise;
+      }
+      // Different hostname - cancel previous and start new
+      this.logService.warning("Cancelling previous cookie acquisition for different hostname");
+      this.cancelPendingAcquisition();
+    }
+
+    const normalizedVaultUrl = vaultUrl.startsWith("https://") ? vaultUrl : `https://${vaultUrl}`;
+    const url = `${normalizedVaultUrl}/proxy-cookie-redirect-connector.html`;
+    this.logService.info(`Opening browser for cookie acquisition: ${url}`);
```
coroiu, line 76: nit: this comment is in the wrong place, we haven't show the dialog yet at this point, the browser is opened after that

apps/desktop/src/platform/services/server-communication-config/server-communication-config.repository.spec.ts
```diff
@@ -1,6 +1,6 @@
 import { AcquiredCookie, ServerCommunicationConfig } from "@bitwarden/sdk-internal";
 
-import { FakeAccountService, FakeStateProvider } from "../../../../spec";
+import { FakeAccountService, FakeStateProvider } from "../../../../../../libs/common/spec";
```
coroiu: issue: we should probably import from `@bitwarden/common`

apps/desktop/src/platform/services/server-communication-config/server-communication-config-platform-api.service.ts
```diff
@@ -0,0 +1,170 @@
...
+    resolve: (cookies: AcquiredCookie[] | undefined) => void;
+    timeoutId: NodeJS.Timeout;
+  } | null = null;
+
+  // Listen for callback messages from app.component.ts
+  private callbackSubscription: Subscription | null = this.messageListener
+    .messages$(SSO_COOKIE_VENDOR_CALLBACK_COMMAND)
+    .subscribe((msg) => {
+      this.handleCallback(msg.urlString);
+    });
```
coroiu: issue: subscriptions should be created in `init()` functions

apps/desktop/src/platform/services/server-communication-config/server-communication-config.repository.spec.ts
djsmith85: Addressed with b9120c5

PR #19669, opened 2026-03-19
[CL-1123] fix ghost button inactive styles

BryanCunningham:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/CL-1123

## 📔 Objective

- Update in active ghost button styles to match Figma design

## 📸 Screenshots
image

### DIFF
--- a/libs/components/src/shared/base-button.directive.ts
+++ b/libs/components/src/shared/base-button.directive.ts
@@ -12,15 +12,13 @@ export const focusRing = [
   "focus-visible:tw-z-10",
 ];
 
-type ButtonColorStylesArgs = {
-  buttonType: ButtonType;
-  isDisabled: boolean;
-};
-
 export const getButtonColorStyles = ({
   buttonType,
   isDisabled,
-}: ButtonColorStylesArgs): string[] => {
+}: {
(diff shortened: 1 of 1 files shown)


libs/components/src/shared/base-button.directive.ts
```diff
@@ -12,7 +12,15 @@ export const focusRing = [
...
-export const getButtonColorStyles = (buttonType: ButtonType): string[] => {
+type ButtonColorStylesArgs = {
+  buttonType: ButtonType;
+  isDisabled: boolean;
+};
+
+export const getButtonColorStyles = ({
+  buttonType,
+  isDisabled,
+}: ButtonColorStylesArgs): string[] => {
```
willmartian: Can we just inline this type since it isn't reused?

willmartian (changes requested):

libs/components/src/shared/base-button.directive.ts
BryanCunningham: Sure thing. Fixed [here](https://github.com/bitwarden/clients/pull/19669/changes/f90a2a79b8b28470be79fe4e417960d08e523539)

Issue #18061, opened 2025-12-19
macOS ssh agent won't respond after awake from sleep


gengjiawen:
### Steps To Reproduce

macOS go to sleep, ssh agent won't working have to restart bitwarden to make it work

### Expected Result

always work

### Actual Result

have to restart bitwarnden

### Operating System

macOS

### Operating System Version

26.2 (25C56)

### Installation method

Mac App Store

### Build Version

Version 2025.12.0 (52522)

### Issue Tracking Info

- [ ] I understand that work is tracked outside of GitHub. A PR will be linked to this issue should one be opened to address it, but Bitwarden doesn't use fields like "assigned", "milestone", or "project" to track progress.

Krychaz:
Hi there,

Thank you for your report! 

I was able to reproduce this issue, and I have flagged this to our engineering team.

If you wish to add any further information/screenshots/recordings etc., please feel free to do so at any time - our engineering team will be happy to review these.

Thanks once again!

wrobelda:
I'd like to point out that not only is agent unresponsive, it causes an instantaneous logout. I have to log back into Bitwarden each time this happens, several times a day.

Krychaz:
Hi,

@gengjiawen @wrobelda We are looking into this issue now and since it was reproduced last month, we cannot do it anymore. Can you please confirm if the issue is still happening on your side while running 2025.12.0 and if so could you please provide with your replication steps.

Thank you!

wrobelda:
@Krychaz also can't reproduce this issue anymore, seems like it got fixed along the way.

coltonhurst:
@wrobelda thanks for the response.

I will close this issue, if it crops up again please let us know!

PR #19673, opened 2026-03-19
[PM-33915] [PM-33876] Fix: macOS SSH agent stops working after sleep/inactivity

### LINKED ISSUE #19653: MacOS ssh-agent stop working after some period of inactivity
### Steps To Reproduce

MacOS, ssh-agent configured with SSH_AUTH_SOCK=~/.bitwarden-ssh-agent.sock
It works, `ssh-add -l` shows all my keys
Then after some inactivity time - usually a few hours, much longer than vault lock timeout `ssh-add -l` and any other ssh related command start hanging. 
The ` ~/.bitwarden-ssh-agent.sock` is still present, but any access to it hangs.
I have to restart Bitwarden in order to make it working again.

This seems to be the same issue as https://github.com/bitwarden/clients/issues/18061 but I cannot reopen the original one

### Expected Result

ssh-agent always works

### Actual Result

ssh-agent stop working and all ssh related commands hangs

### Screenshots or Videos

_No response_

### Additional Context

_No response_

### Operating System

macOS

### Operating System Version

26.3.1

### Installation method

Homebrew

### Build Version

2026.2.1 (57189)

### Issue Tracking Info

tdragon:
## 🎟️ Tracking

Fixes #19653 
Related to 
- #18061 
- https://bitwarden.atlassian.net/browse/PM-33876

## 📔 Objective

**Root cause:** 
`bitwarden_russh::serve()` accept loop uses `tokio::select!` with pattern `Some(Ok((stream, info))) = listener.next()`. 

When `accept()` returns any error (transient OS error after sleep/resume), pattern doesn't match, `select!` disables the listener branch and waits on cancellation token that never fires. 
Accept loop is dead from this point.

Can't fix in `bitwarden_russh` directly - repo is deprecated and not accepting PRs.

**Changes:**

1. `PeercredUnixListenerStream` - absorb accept errors at stream level. Log warning, schedule re-poll with `wake_by_ref()`, return `Poll::Pending`. The buggy `serve()` loop never sees any errors.

2. `BitwardenDesktopAgent::stop()` - cancel the `CancellationToken`. Before this it only set `is_running = false` but the accept loop kept running because token was never cancelled.

## 📸 Screenshots

_not applicable - no UI changes_

### DIFF
--- a/apps/browser/src/_locales/en/messages.json
+++ b/apps/browser/src/_locales/en/messages.json
@@ -6169,9 +6169,6 @@
   "enterMultipleEmailsSeparatedByComma": {
     "message": "Enter multiple emails by separating with a comma."
   },
-  "emailsRequiredChangeAccessType": {
-    "message": "Email verification requires at least one email address. To remove all emails, change the access type above."
-  },
   "emailPlaceholder": {
     "message": "user@bitwarden.com , user@acme.com"
   },
--- a/apps/desktop/src/locales/en/messages.json
+++ b/apps/desktop/src/locales/en/messages.json
@@ -4702,9 +4702,6 @@
   "enterMultipleEmailsSeparatedByComma": {
     "message": "Enter multiple emails by separating with a comma."
   },
-  "emailsRequiredChangeAccessType": {
-    "message": "Email verification requires at least one email address. To remove all emails, change the access type above."
-  },
   "emailPlaceholder": {
     "message": "user@bitwarden.com , user@acme.com"
   },
--- a/apps/web/src/app/tools/send/send.component.html
+++ b/apps/web/src/app/tools/send/send.component.html
@@ -1,75 +1,42 @@
 <app-header>
-  <ng-container slot="title-suffix">
-    <small #actionSpinner [appApiAction]="actionPromise">
-      <ng-container *ngIf="$any(actionSpinner).loading">
-        <i
-          class="bwi bwi-spinner bwi-spin tw-text-muted"
-          title="{{ 'loading' | i18n }}"
-          aria-hidden="true"
-        ></i>
-        <span class="tw-sr-only">{{ "loading" | i18n }}</span>
-      </ng-container>
-    </small>
-  </ng-container>
-  <tools-new-send-dropdown *ngIf="!disableSend"></tools-new-send-dropdown>
+  @if (!disableSend()) {
+    <tools-new-send-dropdown></tools-new-send-dropdown>
+  }
 </app-header>
 
-<bit-callout type="warning" title="{{ 'sendDisabled' | i18n }}" *ngIf="disableSend">
-  {{ "sendDisabledWarning" | i18n }}
-</bit-callout>
+@if (disableSend()) {
+  <bit-callout type="warning" title="{{ 'sendDisabled' | i18n }}">
+    {{ "sendDisabledWarning" | i18n }}
+  </bit-callout>
+}
 
 @if (SendUIRefresh$ | async) {
-  <div class="tw-mb-4">
-    <bit-search
-      [(ngModel)]="searchText"
-      [placeholder]="'searchSends' | i18n"
-      (ngModelChange)="searchTextChanged()"
-      appAutofocus
-    />
-  </div>
-
-  <div class="tw-mb-4">
-    <bit-toggle-group [selected]="selectedToggleValue" (selectedChange)="onToggleChange($event)">
-      <bit-toggle [value]="'all'">{{ "allSends" | i18n }}</bit-toggle>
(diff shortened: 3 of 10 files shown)


tdragon:
Running the fixed version for a few days not-stop - there were not a single hang. Previously I had to restart Bitwarden several times a day, so I guess the fix is working.

neuronull:
> Running the fixed version for a few days not-stop - there were not a single hang. Previously I had to restart Bitwarden several times a day, so I guess the fix is working.

Hi @tdragon , it sounds like you upgraded your Desktop client and are no longer facing this issue? If so, I think we can safely close this PR and the issue that was opened?

tdragon:
> > Running the fixed version for a few days not-stop - there were not a single hang. Previously I had to restart Bitwarden several times a day, so I guess the fix is working.
> 
> Hi @tdragon , it sounds like you upgraded your Desktop client and are no longer facing this issue? If so, I think we can safely close this PR and the issue that was opened?

No, I build a version with this fix and run it to prove that the fix is working. The nature of the bug is quite random, so it is hard to reproduce it. The first report of this bug was closed as "cannot reproduce". Please, consider to accepting my fix. 
This is a critical bug blocking us from migrating to Bitwarden from 1PW.

neuronull:
> > > Running the fixed version for a few days not-stop - there were not a single hang. Previously I had to restart Bitwarden several times a day, so I guess the fix is working.
> > 
> > 
> > Hi @tdragon , it sounds like you upgraded your Desktop client and are no longer facing this issue? If so, I think we can safely close this PR and the issue that was opened?
> 
> No, I build a version with this fix and run it to prove that the fix is working. The nature of the bug is quite random, so it is hard to reproduce it. The first report of this bug was closed as "cannot reproduce". Please, consider to accepting my fix. This is a critical bug blocking us from migrating to Bitwarden from 1PW.

Ah, thanks for clarifying that. I will try to reproduce the issue. We'll want to ensure a reproducible state in order to increase confidence in the fix.

I'll also state that we're in the process of a v2 re-write of the ssh agent implementation at the moment, so this code is very much under construction.

tdragon:
> Ah, thanks for clarifying that. I will try to reproduce the issue. We'll want to ensure a reproducible state in order to increase confidence in the fix.
> 
> I'll also state that we're in the process of a v2 re-write of the ssh agent implementation at the moment, so this code is very much under construction.

Thanks, @neuronull ! I hope the v2 will not have this issue, but if it is not planned for the next release I think my fix worth it.
Reproducing may be difficult - the error depends on macOS invalidating the Unix socket after sleep/wake, which is timing-dependent and may not occur in debug builds.

However, the bug is evident from reading the code: in peercred_unix_listener_stream.rs, a transient accept() error returns Poll::Ready(Some(Err(err))), which terminates the stream and permanently kills the SSH agent listener. No logging exists on this path, so the failure is completely silent. The fix is defensive — log and retry instead of terminating. I'd suggest reviewing the error handling path in the code rather than trying to reproduce.

apps/desktop/desktop_native/core/src/ssh_agent/mod.rs
```diff
@@ -205,6 +205,7 @@ impl BitwardenDesktopAgent {
 
         self.is_running
             .store(false, std::sync::atomic::Ordering::Relaxed);
+        self.cancellation_token.cancel();
```
neuronull, line 208: While cancelling the token here is the correct thing to do, the `stop()` function is actually dead code.

This problem will be addressed in the v2.

neuronull:
> However, the bug is evident from reading the code: in peercred_unix_listener_stream.rs, a transient accept() error returns Poll::Ready(Some(Err(err))), which terminates the stream and permanently kills the SSH agent listener. No logging exists on this path, so the failure is completely silent.

Hey @tdragon  thanks for working with me on this! I left a comment in the ticket that was opened, perhaps we could converge the discussion of the problem there. I'd suggest the following: if between my comment and the questions I asked, you're able to get a reliable way to reproduce the issue, I can a/b (before/after) test it and this would be the ideal scenario. In absence of that, something that would help us is if you provided a/b output of the console logs. If your code change fixes the issue you are facing, then the tracing log should be present in the output.

neuronull:
(FYI)  

https://github.com/bitwarden/clients/commit/50cf5bde996e394560e8e272360a4209b958008f

neuronull:
👋  hey @tdragon  , I'll go ahead and close this PR for the time being.

We've got some great data in https://github.com/bitwarden/clients/issues/19653

If we get a reproduction in the OS sleep scenario, and the instrumentation demonstrates an issue in the listener stream wrapper, we can re-open this PR and re-evaluate it.

And in the meantime, we'll focus on improving this area in the v2 of the feature!

PR #19676, opened 2026-03-19
[PM-33877] - handle blank custom field values in cipher form

### LINKED ISSUE #19654: new Custom Field value is undefined
### Steps To Reproduce

Reproduced in Bitwarden version 2026.2.1 and 2026.1.1 , on Windows 11 .

2. Open 'Bitwarden desktop'
3. Click on 'Login' and open an existing Login item.
4. Scroll down to 'Custom Fields'
5. Click on '+ add'
6. Give it a type (e.g. Text) and name (e.g. "Site space") and click Save.
7. Observe you now see the Login item with new field, but a blank value.
8. Do NOT focus the new custom field or enter a text here, just click Save for the Login item.
9. Observe that the value of the custom field now says **undefined**.
10. Observe that using the COPY button here does say it did a copy, but in reality does nothing. There is no undefined text on clipboard.

Looks like a programming bug.

### Expected Result

Blank value, not a text "undefined". Looks like a programming error.

### Actual Result

The value is actually blank but shown as "undefined" text. 

### Screenshots or Videos

Image

### Additional Context

_No response_

### Operating System

Windows

### Operating System Version

11 25H2

### Installation method

### Build Version

Version 2026.2.1 SDK ' ()' Shell 39.2.6 Renderer 142.0.7444.226 Node 22.21.1 Architecture x64

### Issue Tracking Info

jaasen-livefront:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-33877

Fixes #19654

## 📔 Objective

Custom fields in the cipher form allow blank values for Text and Hidden field types which results in `undefined` being applied to the form values. This PR properly handles those values.

## 📸 Screenshots

https://github.com/user-attachments/assets/700cb13c-8212-4f18-b5a9-313986d52e0c

### DIFF
--- a/libs/vault/src/cipher-view/custom-fields/custom-fields-v2.component.html
+++ b/libs/vault/src/cipher-view/custom-fields/custom-fields-v2.component.html
@@ -14,7 +14,7 @@
         <textarea
           readonly
           bitInput
-          [value]="field.value"
+          [value]="field.value ?? ''"
           aria-readonly="true"
           vaultAutosizeReadOnlyTextArea
         ></textarea>
@@ -35,7 +35,7 @@
           readonly
           bitInput
           type="password"
-          [value]="field.value"
+          [value]="field.value ?? ''"
           aria-readonly="true"
           class="tw-font-mono"
           *ngIf="!revealedHiddenFields.includes(i)"


pamperer562580892423:
@jaasen-livefront As I just saw your preview video, with two blank custom fields that have the same label, I just remembered that exactly this condition (and probably similar ones) caused crashes on the Android app: https://github.com/bitwarden/android/issues/6486

Maybe totally unrelated to your work here, and/or maybe you're already counting that in -- but anyway, as I could reproduce it back then myself, I thought I wanted to point this out here. (and if there was a connection now to your fix here, maybe this would solve that mobile app issue as well)

nick-livefront (approved):

jaasen-livefront:
> @jaasen-livefront As I just saw your preview video, with two blank custom fields that have the same label, I just remembered that exactly this condition (and probably similar ones) caused crashes on the Android app: [bitwarden/android#6486](https://github.com/bitwarden/android/issues/6486)
> 
> Maybe totally unrelated to your work here, and/or maybe you're already counting that in -- but anyway, as I could reproduce it back then myself, I thought I wanted to point this out here. (and if there was a connection now to your fix here, maybe this would solve that mobile app issue as well)

@pamperer562580892423 Thanks for pointing that out. While these are certainly related the fix here will have no affect the android bug. Good news is there appears to be a fix that has just been applied [here](https://github.com/bitwarden/android/pull/6743)!

PR #19686, opened 2026-03-20
[CL-1124] Increase badge max-width

willmartian:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/CL-1124

## 📔 Objective

Updates the default `maxWidthClass` input on `BadgeComponent` from `tw-max-w-40` (10rem/160px) to `tw-max-w-[30ch]` to prevent badges from truncating too soon, particularly when an icon is present.

### DIFF
--- a/libs/components/src/badge/badge.component.ts
+++ b/libs/components/src/badge/badge.component.ts
@@ -98,7 +98,7 @@ export class BadgeComponent implements FocusableElement {
    * Tailwind max-width class to apply when truncating is enabled.
    * Must be a valid Tailwind max-width utility class (e.g., "tw-max-w-40", "tw-max-w-xs").
    */
-  readonly maxWidthClass = input<`tw-max-w-${string}`>("tw-max-w-40");
+  readonly maxWidthClass = input<`tw-max-w-${string}`>("tw-max-w-[30ch]");
 
   protected readonly classList = computed(() => {
     return [


libs/components/src/badge/badge.component.ts
```diff
@@ -116,55 +88,59 @@ export class BadgeComponent {
...
-      return null;
-    }
-
-    return this.startIcon() || getDefaultIconForVariant(this.variant());
-  });
-
-  protected readonly iconSizeStyles = computed(() => {
-    return sizeStyles[this.size()]?.icon;
-  });
+  readonly maxWidthClass = input<`tw-max-w-${string}`>("tw-max-w-[30ch]");
```
BryanCunningham, line 101: @willmartian I think we need this to be dynamic based on if a start icon is shown or not. We should also probably factor in the X padding into this to ensure it's actually 30 chars that are showing. Alternatively, we could probably apply this to the content container instead and those calculations are no longer necessary

BryanCunningham: Think this might do the trick for us actually
```js
<span class="tw-px-1 tw-text-start tw-min-w-0 tw-flex-1" [class]="truncate() ? maxWidthClass() : null" [class.tw-truncate]="truncate()"> 
  <ng-content></ng-content>
</span>
```

BryanCunningham: And remove it from the host classes, of course

PR #19690, opened 2026-03-20
[PM-29927] update reseller notifications

kdenney:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-29927

## 📔 Objective

Updates reseller-managed organization renewal notification copy

  - Removes the reseller name from banner copy, replacing it with the hardcoded phrase "your Bitwarden provider"
  - Suppresses the issued state banner (shown when a renewal invoice is issued but the subscription is still active)
  - Changes the past_due banner type from warning to info for consistency with the component template and requirements

New i18n keys resellerRenewalWarningMsgV2 and resellerPastDueWarningMsgV2 are introduced rather than modifying the existing keys, as deployed i18n strings are immutable.

## 📸 Screenshots
image

image


cyprain-okeke (approved):
Looks good

PR #19691, opened 2026-03-20
Refactor member dialog to use chip input for email entry

JaredScar:
## 🎟️ Tracking

N/A

## 📔 Objective

Adds a new component to UI library called `chip-input` which can be used for comma separated type inputs within the app to better modernize the Bitwarden app.

## 📸 Screenshots

image

image

image

### DIFF
--- a/apps/web/src/app/admin-console/organizations/members/components/member-dialog/member-dialog.component.html
+++ b/apps/web/src/app/admin-console/organizations/members/components/member-dialog/member-dialog.component.html
@@ -27,7 +27,10 @@
             <p bitTypography="body1">{{ "inviteUserDesc" | i18n }}</p>
             <bit-form-field *ngIf="{ seats: remainingSeats$ | async } as remaining">
               <bit-label>{{ "email" | i18n }}</bit-label>
-              <input id="emails" type="text" appAutoFocus bitInput formControlName="emails" />
+              <bit-chip-input
+                formControlName="emails"
+                placeholder="{{ 'typeEmailAndPressComma' | i18n }}"
+              ></bit-chip-input>
 
               <bit-hint *ngIf="remaining.seats > 1">{{
                 "inviteMultipleEmailDesc" | i18n: remaining.seats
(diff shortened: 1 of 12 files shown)


willmartian (changes requested):
hey Jared! Are you familiar with https://components.bitwarden.com/?path=/story/component-library-form-multi-select--collections ? 

If so, how should/do these differ?

JaredScar:
> hey Jared! Are you familiar with https://components.bitwarden.com/?path=/story/component-library-form-multi-select--collections ?
> 
> If so, how should/do these differ?

Was not aware of it! @willmartian Essentially the idea here is to allow adding writable inputs too for selection. It seems the multi-select doesn't support that, but maybe I should aim a PR at changing that instead? 🤔

willmartian:
> Essentially the idea here is to allow adding writable inputs too for selection. It seems the multi-select doesn't support that, but maybe I should aim a PR at changing that instead? 🤔

Hmm, I'm a little hesitant to move forward _right now:_

- We are in the middle of rebuilding many of our DS components, including chip and multi select. I am hesitant to introduce additional churn in any components we are actively touching.
- Relatedly, I think Product/Design sign-off is extra important for changes to the CL. Is this pattern flexible enough to meet their needs elsewhere? 
 
Let me sync up with design, and I will get back to you. 

Slight tangent:
I wonder if the UI in the screenshot should use the [access selector component](https://62a88a6de5b807fa98886113-xjwoqczwup.chromatic.com/?path=/story/web-organizations-access-selector--group-members-access)? Does having that table below the input change any necessary affordance of the input itself?

eliykat (commented):
I agree with @willmartian that this PR is a bit premature - even as a BEEEP project it needs some discussion with the UIF team and our designer first. Have you discussed it with Sukhleen at all?

I still like the idea of using chips here, but it's largely design's call to make.

JaredScar:
> I agree with @willmartian that this PR is a bit premature - even as a BEEEP project it needs some discussion with the UIF team and our designer first. Have you discussed it with Sukhleen at all?
> 
> I still like the idea of using chips here, but it's largely design's call to make.

I did discuss this with Sukhleen and she liked the idea of this for the `Generate Invite Link` changes! May be worth a overall discussion with her and the UIF team though? 🤔 

@eliykat @willmartian

eliykat:
@JaredScar Yes I think so - Jillian has taken over from Sukhleen as our designer, so please talk to her - particularly because we have other changes lined up for this component (for the invite link work). She will need to approve any UI changes here. Also include @willmartian because we need to decide whether chips will live in the component library. A post in our public team slack channel is probably the easiest way to ping everyone at once.

Please also make and link a Jira ticket in the PR description.

PR #19693, opened 2026-03-20
[PM-33139] Targeting Rules initial implementation

jprusik:
## 🎟️ Tracking

[PM-33139](https://bitwarden.atlassian.net/browse/PM-33139)

## 📔 Objective

Implements Fill Assist Targeting Rules which allow the browser extension to use curated CSS selector rules to locate form fields on web pages instead of relying solely on heuristics.

When enabled (opt-in user setting within the Autofill Settings view of the extension), the extension fetches a set of form/field locators (grouped by host) from a server-specified data provider. Targeting rules are resolved by matching selectors against the DOM, and fills targeted fields directly, bypassing heuristic field qualification. If no rules exist for the current page, normal heuristics-driven logic will be used. However, when rules do exist and fail (the selector does not resolve an element), targeting rules do NOT fall back to heuristic strategies.

## Notes

- This feature is currently behind the feature flag `fill-assist-targeting-rules`; we'll be undergoing some internal testing and iteration to verify stability/viability before any public release.
- The current default data provider uses a "Maps" metaphor. As a consumer, with the context of implemented use, this same data is represented as "Rules".
- Because targeting rules leverage a human-curated experience, we treat the items targeted on the host page as "vetted". This means allowing some field handling behaviours that would otherwise be considered risky or incomplete (which is why this is an opt-in feature). The rules are a description of the _page_, not anything in the vault or Bitwarden experience. Note: this assumption does not apply to the user's _ciphers_; all cipher URI matching rules remain in place. Some additional scenarios:
  - If a field is hidden, we presume the rule author would already know that (the concern of vetting changes to the pages targeted is a separate concern).
  - For this same reason, if the targeted rules experience fails, we do not fall back to the heuristic approach since the expectations of the page failed and need to be re-reviewed by the rules author.
  - We do not respect equivalent domains in targeting rules since they are specifically vetted experiences (an author may create rules for each host in equivalent domain matches).
  - Blocked domains remain unaffected; a blocked domain will continue to prevent content script injection on specified sites, regardless of the presence of a targeted rule for those sites.
- We expect to allow optional server customization of the rules resource location (self-hosted, enterprise, etc). Consequently, the data resource-handling code is somewhat defensive.

## Changes Summary

- Data fetching (`apps/browser/src/autofill/services/targeting-rules-data.service.ts`)
  - periodic (currently every 6 hours) fetch from configured or default remote endpoint (a la phishing domains)
  - this also doesn't presume data provider host; default can easily be switched out later for an in-house CDN without logical changes
  - CDN cache-busting via timestamp query parameter
  - we intentionally clear the cached local state of rules on failure; stale rules represent more risk than temporarily falling back to heuristics.
- Data model & types (`libs/common/src/autofill/types/index.ts`)
  - annotated with codified intent/expectations
- Host lookup (`libs/common/src/autofill/services/domain-settings.service.ts`)
  - `getTargetingRulesForUrl()` is the primary method in which call-sites should be getting rules; note, a return of `[]` from this method indicates the URL is blocklisted (no relevant forms, suppress autofill), whereas `null` indicates that no rules exist for the passed URI (fall through to heuristics). This is mostly to help normalize out-of-spec data providers.
- Punycode/IDN support (`libs/common/src/autofill/utils/punycode.ts`)
  - new util to handle punycode conversions; `punycodeToUnicode(host)` converts `xn--` labels back to unicode for matching unicode-keyed rules; somewhat optional if we're able to assume data providers will always author entries consistently without unicode key names
  - this file was heavily developed by an agent using test-driven development methodology. Given the use-case, I felt it made sense to try a small implementation over bringing in a new dependency. Very much open to thoughts/alternatives on this point.
- Field collection (`apps/browser/src/autofill/services/collect-autofill-content.service.ts`)
  - `getTargetedPageDetails()` iterates form selectors, queries DOM via `queryDeepSelector`, builds minimal `AutofillField` entries
  - targeted fields are marked with `targeted = true` and `fieldQualifier`
  - targeted fields are always treated as `viewable = true` (rules may deliberately target hidden fields)
- Inline menu integration (`apps/browser/src/autofill/services/autofill-overlay-content.service.ts`)
  - `setTargetedFieldFillType()` maps targeted field qualifiers to cipher types (Login, Card, Identity)
  - targeted fields bypass `isIgnoredField` heuristic checks (authoritative override)
- Fill script generation (`apps/browser/src/autofill/services/autofill.service.ts`)
  - `generateTargetedFillScript()` maps cipher values to targeted field types
  - `getValueForTargetedFieldType()` resolves cipher property values for all defined field types
  - targeted fields skip untrusted iframe check, for the same reason we allow hidden input fills; rules are vetted/authoritative
- Shadow DOM support (`apps/browser/src/autofill/services/dom-query.service.ts`)
  - `queryDeepSelector()` splits on `>>>` combinator, traverses into (both closed and open) shadow roots
- User setting UI toggle (`apps/browser/src/autofill/popup/settings/`)

## Tech Debt / Limitations

In order to keep the scope of this initial implementation ~~small~~ less huge, I've deferred some tasks which will be handled in independent tasks:

- UI for the feature in the Autofill Settings view is placeholder and will be handled as a separate task ([PM-32245](https://bitwarden.atlassian.net/browse/PM-32245)).
- The `form` selector type is parsed from rules but not used; this is intended to enable cases where a `<form>` tag exists but is otherwise ambiguous. It additionally enables creating an explicit relationship between a form tag and described inputs ([PM-34491](https://bitwarden.atlassian.net/browse/PM-34491)).
- The `FormContent.category` type represents an optional indicator of form purpose; this is intended to enable skipping heuristics for determining form purpose categories (e.g. login vs signup forms that otherwise look similar in the DOM) ([PM-34492](https://bitwarden.atlassian.net/browse/PM-34492)).
- Targeted fills inside closed shadow roots may not trigger add/update cipher notifications ([PM-34493](https://bitwarden.atlassian.net/browse/PM-34493)).
- Targeted-rule pages still run the full heuristic mutation observer setup ([PM-34495](https://bitwarden.atlassian.net/browse/PM-34495)).
- Targeted fields are subject to viewport checks despite potentially being off-screen by design (tabbed forms, fields revealed by interaction) ([PM-34496](https://bitwarden.atlassian.net/browse/PM-34496)).
- The Inline menu needs to be fully implemented for totp cases, which requires special handling (see: the Selector Sequence concept in the Forms Map) ([PM-34743](https://bitwarden.atlassian.net/browse/PM-34743))
- `queryDeepSelector` currently only handles shadow DOM boundaries; iframe `>>>` traversal works for fill, but inline menu positioning is incorrect because it relies on per-frame content script injection for coordinate calculation ([PM-34497](https://bitwarden.atlassian.net/browse/PM-34497)):
  Screenshot 2026-03-31 at 12 32 01 PM

## Related Work

- https://github.com/bitwarden/server/pull/7137

## Screenshots

Screenshot 2026-03-06 at 1 57 21 PM

[PM-33139]: https://bitwarden.atlassian.net/browse/PM-33139?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[PM-34491]: https://bitwarden.atlassian.net/browse/PM-34491?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[PM-32245]: https://bitwarden.atlassian.net/browse/PM-32245?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/common/src/autofill/services/domain-settings.service.ts
```diff
@@ -156,6 +206,12 @@ export class DefaultDomainSettingsService implements DomainSettingsService {
       map((x) => x ?? UriMatchStrategy.Domain),
     );
 
+    this.enableFillAssistState = this.stateProvider.getGlobal(ENABLE_FILL_ASSIST);
+    this.enableFillAssist$ = this.enableFillAssistState.state$.pipe(map((x) => x ?? false));
+
+    this.targetingRulesState = this.stateProvider.getGlobal(TARGETING_RULES);
```
MGibson1: these being global is incompatible with the self-host targeting rules settings you have planned.

libs/common/src/autofill/services/domain-settings.service.ts
```diff
@@ -218,4 +274,92 @@ export class DefaultDomainSettingsService implements DomainSettingsService {
 
     return domains$;
   }
+
+  async setEnableFillAssist(newValue: boolean): Promise<void> {
+    await this.enableFillAssistState.update(() => newValue);
```
MGibson1: best practice is to include `shouldUpdate` a callback to avoid writes that are unnecessary.

libs/common/src/autofill/services/domain-settings.service.ts
```diff
@@ -218,4 +274,92 @@ export class DefaultDomainSettingsService implements DomainSettingsService {
...
+    const enableFillAssist = await firstValueFrom(this.enableFillAssist$);
+    if (!enableFillAssist) {
+      return null;
+    }
+
+    // Fill Assist will not be applied when the user is logged out
+    const activeAccount = await firstValueFrom(this.accountService.activeAccount$);
+    if (!activeAccount) {
+      return null;
+    }
```
MGibson1, line 321: active is not the same as logged in, you want to check auth service's active account status

apps/browser/src/autofill/services/targeting-rules-data.service.ts
```diff
@@ -0,0 +1,199 @@
...
+  },
+);
+
+/**
+ * Browser-specific service responsible for fetching and syncing targeting rules
+ * from an external source. Fetches rules on initialization and periodically
+ * refreshes them in the background.
+ */
+export class TargetingRulesDataService {
+  static readonly UPDATE_INTERVAL = 6 * 60 * 60 * 1000; // 6 hours
```
MGibson1, line 55: we probably want this to be configurable server-side

apps/browser/src/autofill/services/targeting-rules-data.service.ts
MGibson1, line 1: how does this handle user switches and adds?

apps/browser/src/autofill/services/autofill-overlay-content.service.ts
```diff
@@ -1116,6 +1127,67 @@ export class AutofillOverlayContentService implements AutofillOverlayContentServ
...
+      AutofillTargetingRuleTypes.identityAddress3,
+      AutofillTargetingRuleTypes.identityCity,
+      AutofillTargetingRuleTypes.identityState,
+      AutofillTargetingRuleTypes.identityPostalCode,
+      AutofillTargetingRuleTypes.identityCountry,
+      AutofillTargetingRuleTypes.identityCompany,
+      AutofillTargetingRuleTypes.identityPhone,
+      AutofillTargetingRuleTypes.identityEmail,
+      AutofillTargetingRuleTypes.identityUsername,
+    ];
```
MGibson1: **nit** shouldn't be redefined every call

libs/common/src/autofill/constants/index.ts
```diff
@@ -149,4 +149,35 @@ export const CLEAR_NOTIFICATION_LOGIN_DATA_DURATION = 60 * 1000; // 1 minute
 
 export const MAX_DEEP_QUERY_RECURSION_DEPTH = 4;
 
+export const DEEP_QUERY_SELECTOR_COMBINATOR = ">>>";
+
+export const AutofillTargetingRuleTypes = {
```
MGibson1, line 158: This type is weird. why have a class with every property equal to the property name?

apps/browser/src/autofill/services/targeting-rules-data.service.ts
jprusik: Agreed - added a follow up task: [PM-34562](https://bitwarden.atlassian.net/browse/PM-34562)

libs/common/src/autofill/constants/index.ts
jprusik: So we can use a shared reference instead of repeating string values all over the place which can fall out of sync are susceptible to undetected typos. It also enables enum-style typing (see `AutofillTargetingRuleType`) [without using enums](https://contributing.bitwarden.com/architecture/adr/ts-deprecate-enums).

apps/browser/src/autofill/services/autofill-overlay-content.service.ts
jprusik: addressed in 8ba46410500a1b1cebeeb63e20a0628fcfdc076f

apps/browser/src/autofill/services/autofill.service.ts
```diff
@@ -848,6 +855,136 @@ export default class AutofillService implements AutofillServiceInterface {
...
+    // Login fields
+    if (fieldType === "username") {
+      return cipher.login?.username ?? null;
+    }
+    if (fieldType === "password" || fieldType === "newPassword") {
+      return cipher.login?.password ?? null;
+    }
+    if (fieldType === "totp") {
+      return cipher.login?.totp ?? null;
+    }
```
jprusik: probably need to rethink this:
- `totp` won't yield the generated value like this; I think we need to pass it through `totpService`
- Maps will need to be able to specify when a value is split between multiple fields, like TOTP (or in map terms, inputs which are grouped to represent a split-value vs single value inputs represented in multiple places)

jprusik: resolved in `5624f1c7eb578c85d7fba143f009955227c8a81d`

well, removed, but will be revisited in [PM-34743](https://bitwarden.atlassian.net/browse/PM-34743)

libs/common/src/autofill/services/domain-settings.service.ts
jprusik: addressed in 5e2f0da35c43f7e3a0599665f84e0745be9f88ee

libs/common/src/autofill/services/domain-settings.service.ts
jprusik: addressed in 5e2f0da35c43f7e3a0599665f84e0745be9f88ee

apps/browser/src/autofill/services/targeting-rules-data.service.ts
jprusik: addressed in 5e2f0da35c43f7e3a0599665f84e0745be9f88ee

addisonbeck (dismissed):
Platform updates are simple extensions of config and scheduled jobs. LGTM, but I didn't review any of the other code. Thanks for using the new scheduling lib! I think this is the first time it's been changed since it was cut, glad to see it working.

differsthecat (dismissed):

differsthecat (dismissed):

jprusik:
Rebased to resolve another round of merge conflicts

differsthecat (dismissed):

dereknance (dismissed):
Approving changes to Platform-owned files on Addison's behalf after rebase.

differsthecat (approved):

dereknance (approved):

PR #19695, opened 2026-03-20
[PM-33952] Fix cipher key encryption logic when editing ciphers

nikwithak:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-33952

## 📔 Objective

When cipher-key-encryption and SDK encryption are both off, editing a cipher will rollback the cipher key. This also has potential to cause cipher corruption  with some a server versions.

Note: This code path is dead with `pm-22136-sdk-cipher-encryption` turned on, but worth fixing just in case.


jaasen-livefront (approved):
Ah nice catch!

PR #9513, opened 2024-06-05
[PM-5156] [PM-5216] Duo v2 removal

ike-kottlowski:
## 🎟️ Tracking

[PM-5216](https://bitwarden.atlassian.net/browse/PM-5216)

This needs to be merged with the complementary [Server PR](https://github.com/bitwarden/server/pull/4126).

## 📔 Objective

Remove Duo SDK v2 dependencies

### Login with Duo

https://github.com/bitwarden/clients/assets/137194738/19125142-aee6-48bc-b6c9-a5a3b084dea0

## ⏰ Reminders before review

- Contributor guidelines followed
- All formatters and local linters executed and passed
- Written new unit and / or integration tests where applicable
- Protected functional changes with optionality (feature flags)
- Used internationalization (i18n) for all UI strings
- CI builds passed
- Communicated to DevOps any deployment requirements
- Updated any necessary documentation (Confluence, contributing docs) or informed the documentation team

## 🦮 Reviewer guidelines

- 👍 (`:+1:`) or similar for great changes
- 📝 (`:memo:`) or ℹ️ (`:information_source:`) for notes or general info
- ❓ (`:question:`) for questions
- 🤔 (`:thinking:`) or 💭 (`:thought_balloon:`) for more open inquiry that's not quite a confirmed issue and could potentially benefit from discussion
- 🎨 (`:art:`) for suggestions / improvements
- ❌ (`:x:`) or ⚠️ (`:warning:`) for more significant problems or concerns needing attention
- 🌱 (`:seedling:`) or ♻️ (`:recycle:`) for future improvements or indications of technical debt
- ⛏ (`:pick:`) for minor or nitpick changes

[PM-5216]: https://bitwarden.atlassian.net/browse/PM-5216?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #19699, opened 2026-03-22
[PM-33963] Remove legacy Duo Web SDK v2 from desktop

trmartin4:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-33963

## 📔 Objective

Removes unused `duo.js` script, which was partially removed in #9513 but not fully.  The script was removed from the `/browser` directory (https://github.com/bitwarden/clients/pull/9513/changes#diff-194e7017881c648ee6936efa0111eb58772f5e1df4c973cf95ddbf233edd2677), but it was not removed from `/desktop`.

This change removes the script from the desktop folder as well.

We know that this is not used, because Duo removed support for Web SDK v2 on March 30, 2024 (https://duo.com/docs/duoweb-v2).


rr-bw (commented):
❓-  It looks like usages of `id="duo-frame"` were cleaned up in https://github.com/bitwarden/clients/pull/9513. Can we also clean up `#duo-frame` as part of this work?

See:
- [apps/browser/src/popup/scss/tailwind.css](https://github.com/bitwarden/clients/blob/main/apps/browser/src/popup/scss/tailwind.css#L123)
- [apps/desktop/src/scss/misc.scss](https://github.com/bitwarden/clients/blob/main/apps/desktop/src/scss/misc.scss#L233)

PR #19719, opened 2026-03-23
flatten type-specific data into `openNotificationBar` message

audreyality:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-32531

## 📔 Objective

Introduces a type guard that verifies message schema for at-risk password notification to validate the message format and improve type safety.

## 🎥 Demo

| Before | After |
|--------|-------|
| <video src="https://github.com/user-attachments/assets/04fb1ecf-474c-4b56-877f-1dffb675ce13" width="250" /> | <video src="https://github.com/user-attachments/assets/7f0d623b-5e03-4966-b244-9599adf56008" width="250" /> |

Note: Since webtests.dev has a [well-known password change URL](https://www.w3.org/TR/change-password-url/), the "change password" button displays. Clicking on the button takes you to the password change page.

Sites without this URL display a popup that resembles the "Before" rendering. Unlike that rendering, they now always include the name of the organization requesting the change.


apps/browser/src/_locales/en/messages.json
```diff
@@ -2827,7 +2827,7 @@
       }
     }
   },
-  "atRiskChangePrompt": {
+  "atRiskChangePromptV2": {
```
jprusik: Why does this key need to change?

apps/browser/src/_locales/en/messages.json
```diff
@@ -2837,8 +2837,8 @@
     },
     "description": "Notification body when a login triggers an at-risk password change request and the change password domain is known."
   },
-  "atRiskNavigatePrompt": {
-    "message": "$ORGANIZATION$ wants you to change this password because it is at-risk. Navigate to your account settings to change the password.",
+  "atRiskNavigatePromptV2": {
+    "message": "Your password for this site is at-risk. $ORGANIZATION$ has requested that you navigate to your account settings and change it.",
```
jprusik, line 2877: non-blocking nit; this copy is so close to `atRiskChangePrompt` now, I wonder if we could make a case for consolidating the two entries?

jprusik (dismissed):
A few non-blocking nits; the rest looks good!

apps/browser/src/_locales/en/messages.json
jprusik: > $ORGANIZATION$ has requested that you navigate to your account settings and change it.

@sukhleenb I'm not sure I follow this copy- you don't change your password in your account settings, you change it in the cipher directly, no?

audreyality: * These copy changes were requested by @sukhleenb, so I think your nit is a question for her, not me. 
* Here, "account settings" is talking about the website's account settings.

apps/browser/src/_locales/en/messages.json
audreyality: Originally both messages were changed. This one was reverted but I didn't notice it was a revert. Addressed by 376af1a98e4cbd6c08bdaa8104fd830afa368008.

apps/browser/src/autofill/content/components/notification/at-risk-password/container.ts
```diff
@@ -30,7 +33,7 @@ export function AtRiskNotification({
 }: AtRiskNotificationProps) {
   const { passwordChangeUri, organizationName } = params;
   const riskMessage = chrome.i18n.getMessage(
-    passwordChangeUri ? "atRiskChangePrompt" : "atRiskNavigatePrompt",
+    passwordChangeUri ? "atRiskChangePromptV2" : "atRiskNavigatePromptV2",
```
jprusik:
```suggestion
    passwordChangeUri ? "atRiskChangePrompt" : "atRiskNavigatePromptV2",
```

apps/browser/src/autofill/content/components/notification/at-risk-password/container.spec.ts
```diff
@@ -0,0 +1,122 @@
...
+    it("uses atRiskChangePrompt when passwordChangeUri is present in params", () => {
+      AtRiskNotification({
+        ...baseProps,
+        params: {
+          organizationName: "Acme Co.",
+          passwordChangeUri: "https://example.com/.well-known/change-password",
+        },
+      });
+
+      expect(chrome.i18n.getMessage).toHaveBeenCalledWith("atRiskChangePromptV2", "Acme Co.");
```
jprusik:
```suggestion
      expect(chrome.i18n.getMessage).toHaveBeenCalledWith("atRiskChangePrompt", "Acme Co.");
```

jprusik (changes requested):
Small change needed to update `atRiskChangePromptV2` to `atRiskChangePrompt` after message catalog key revert

jprusik (dismissed):

apps/browser/src/autofill/background/notification.background.ts
```diff
@@ -1813,6 +1828,87 @@ export default class NotificationBackground {
...
+   * re-derives it from the cipher's URIs via the well-known change-password protocol,
+   * preventing an attacker-controlled iframe from injecting a malicious URL.
+   *
+   * @param _message - The extension message (unused)
+   * @param sender - The contextual sender of the message
+   */
+  private async handleOpenChangePasswordUrlMessage(
+    _message: NotificationBackgroundExtensionMessage,
+    sender: chrome.runtime.MessageSender,
+  ) {
```
audreyality, line 1844: 📝 @jprusik - this is the handler that presents scope concerns; it ended up more self-contained than I expected.

🤔 Since this is in the extension context, `TemporaryNotificationChangeLoginService` might be unnecessary. Do you think that this logic can invoke [DefaultChangeLoginPasswordService](https://github.com/bitwarden/clients/blob/main/libs/vault/src/services/default-change-login-password.service.ts)?

apps/browser/src/_locales/en/messages.json
```diff
@@ -2846,8 +2846,8 @@
     },
     "description": "Notification body when a login triggers an at-risk password change request and the change password domain is known."
   },
-  "atRiskNavigatePrompt": {
-    "message": "$ORGANIZATION$ wants you to change this password because it is at-risk. Navigate to your account settings to change the password.",
+  "atRiskNavigatePromptV2": {
```
jprusik, line 2876: I think this needs to get updated in the storybook as well:
https://695ffc4bef53d3a5ae4c8067-enjnmrwfuq.chromatic.com/?path=/story/components-notifications-at-risk-notification--default

audreyality: Addressed in 639b140350c7e2ed7be706fec7dceb56948613c5.

apps/browser/src/autofill/background/notification.background.ts
```diff
@@ -1826,6 +1828,87 @@ export default class NotificationBackground {
...
+          ]);
+        }),
+        switchMap(([ciphers, tasks]) => {
+          if (!ciphers?.length || !tasks?.length) {
+            return EMPTY;
+          }
+
+          // FIXME: When multiple ciphers match the tab URL and each has a pending security
+          // task, this returns the first match — which may not be the cipher the user was
+          // originally notified about. Consider carrying cipher ID through a trusted
```
jprusik, line 1868: FYI; in the past, it was also possible to have more than one security task associated with a given cipher. I think that has been largely mitigated upstream somewhere, but iirc we still defensively handle that in the past by explicitly dismissing all security tasks for the cipher in the update confirmation.

apps/browser/src/autofill/content/components/lit-stories/notification/at-risk-notification/container.lit-stories.ts
```diff
@@ -18,8 +18,9 @@ export default {
   args: {
     theme: ThemeTypes.Light,
     handleCloseNotification: () => alert("Close notification action triggered"),
+    handleChangePasswordClick: () => alert("Change password action triggered"),
     params: {
-      passwordChangeUri: "https://webtests.dev/.well-known/change-password", // Remove to see "navigate" version of notification
+      hasPasswordChangeUri: true, // Set to false to see "navigate" version of notification
       organizationName: "Acme Co.",
     },
     i18n: mockI18n,
```
jprusik, line 26: I think you need to update the changed key name in the mock catalog (`mockI18n`) as well

apps/browser/src/autofill/background/notification.background.ts
jprusik: I would think so; give it a try (or follow up)?

jprusik (dismissed):

apps/browser/src/autofill/background/notification.background.ts
audreyality: The good news for that, at least, is that the cipher is what matters for the lookup.

The main place I could see this FIXME resulting in a bug is if there's a website where there are several matching ciphers for a given login URL that each have a different change URL. 

It's certainly possible--I can imagine such a circumstance being created using URL rewriting and matching rules--but would be significantly outside the obvious, especially given that the well-known URL is, at present, being read from a fixed path.

apps/browser/src/autofill/content/components/lit-stories/notification/at-risk-notification/container.lit-stories.ts
audreyality: Addressed by 7a3edbbcb9df100c1b6b7208ad9546c948b9d499.

apps/browser/src/autofill/background/notification.background.ts
audreyality: I think follow up. I want to get this moving, there's a FIXME to note the debt, and there's multiple places we'd fix.

jprusik (dismissed):

jprusik (approved):

PR #19725, opened 2026-03-24
[CL-954] Upgrade to Angular 21

willmartian:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/CL-954

## 📔 Objective

Upgrades Angular from v20 to v21

**Changes:**
- Angular core, CLI, and related packages bumped to v21 via `ng update @angular/core@21 @angular/cli@21`
- Storybook upgraded to v10
- Updated `@ng-select/ng-select`, `@angular-eslint/schematics`, `@ngtools/webpack`, and `@angular/cdk` to Angular 21-compatible versions
- Pinned `jest-environment-jsdom@29.7.0` as an explicit dependency (previously installed transitively via a jest override that was removed; upgrading to v30 broke `window.location` mocking in tests)
- Added `jest-environment-jsdom` to Renovate grouping and overrode it to `29.7.0` inside `@angular-devkit/build-angular` context (optional peer dep conflict)
- Changed `moduleResolution` from `node` to `bundler` in `tsconfig.base.json` — required for Angular CDK 21 which dropped legacy directory-based entry points in favour of `exports` field only; `libs/nx-plugin` retains `node` as it uses CommonJS
- Added `isolatedModules: true` to spec tsconfigs in Angular libs that used the old `node10` pattern (`libs/auto-confirm`, `libs/logging-angular`, `libs/pricing`, `libs/subscription`) to prevent ts-jest from failing on Angular 21 subpath imports
- Fixed `@HostListener` event parameter types (`EventTarget` / `Event` instead of `HTMLElement` / `KeyboardEvent`) in 5 components — Angular 21's compiler now enforces these more strictly
- Fixed type-only imports/exports in autofill Lit components and `libs/logging` — Storybook v10 upgraded Vite v7→v8 (Rolldown), which does strict export analysis and correctly surfaces `import { Type }` vs `import type { Type }` inconsistencies
- Fixed `vault-wrapper.component.ts` type error — `VaultComponent` is generic while `VaultOrigComponent` is not; annotated `computed<Type<unknown>>()` explicitly since `*ngComponentOutlet` accepts `Type<unknown>`

Unrelated but was annoying me:
- Added `.claude/worktrees/` to `.gitignore`

## 📸 Screenshots

N/A — dependency update, no UI changes.


package.json
```diff
@@ -157,15 +157,15 @@
     "webpack-node-externals": "3.0.0"
   },
   "dependencies": {
-    "@angular/animations": "20.3.18",
+    "@angular/animations": "21.2.5",
     "@angular/cdk": "20.2.14",
```
Hinton: CDK should be upgraded. I suspect these all have peer or dependencies on angular 20 which would maintain them in the lockfile.

package.json
Hinton: Ng-select should also be on 21.

Hinton, line 208: Toastr should be on 20

willmartian: Toastr actually does not need to be updated as part of this. Current version supports Angular >= 17.x < 23.x

Hinton: Hmm, in the past they've only supported one version. I guess since it's archived and we're moving away from it less moving pieces are better.

jengstrom-bw (dismissed):
Vault changes look good.

connerbw (dismissed):

jengstrom-bw (dismissed):

dani-garcia (dismissed):

jrmccannon (dismissed):
Looks good from Admin Console perspective. 

> Added isolatedModules: true to spec tsconfigs in Angular libs that used the old node10 pattern (libs/auto-confirm, libs/logging-angular, libs/pricing, libs/subscription) to prevent ts-jest from failing on Angular 21 subpath imports

Are there changes that need to be made in the `libs/auto-confirm` library to address this call out?

willmartian:
> > Added isolatedModules: true to spec tsconfigs in Angular libs that used the old node10 pattern (libs/auto-confirm, libs/logging-angular, libs/pricing, libs/subscription) to prevent ts-jest from failing on Angular 21 subpath imports
> 
> Are there changes that need to be made in the `libs/auto-confirm` library to address this call out?

@jrmccannon Not to be knowledge, beyond what I have already made here. I will chat with @bitwarden/team-platform-dev to see if we should update of nx lib generator.

dan-livefront (dismissed):

BryanCunningham (dismissed):

jengstrom-bw (dismissed):
vault files look good

dani-garcia (dismissed):

jrmccannon (dismissed):

BryanCunningham (approved):

dani-garcia (approved):

sbrown-livefront (approved):

dan-livefront (approved):

jengstrom-bw (approved):

jrmccannon (approved):

PR #19726, opened 2026-03-24
Magnify Glue Part 1

coltonhurst:
Magnify Glue Part 1

Contains the following changes:

- Settings component changes
- Local active user state
- Bitwarden Render Process --ipc--> Main Process
- Keyboard shortcut

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -246,6 +246,205 @@
         "@napi-rs/cli": "3.5.1"
       }
     },
+    "apps/desktop/node_modules/@inquirer/confirm": {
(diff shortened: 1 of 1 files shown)


package-lock.json
ike-kottlowski, line 1: Did you mean to include the `package-lock.json`?

coltonhurst: It did change when running `npm i` at root, but I don't really want that to be part of this PR, so reverted in db3c4cedf82301d4e2a0f680658bf213ac3d16fc, thanks!

PR #19734, opened 2026-03-24
[PM-33431] Inline menu icon is showing up on the view item screen on vault.bitwarden.com

dan-livefront:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-33431
## 📔 Objective

The inline menu did not recognize readonly and disabled inputs as non editable. Added checks for readonly, aria readonly, and aligned with cached field data as needed.

## 📸 Screenshots

Screenshot 2026-03-24 at 12 37 44 PM


blackwood (approved):
LGTM!

PR #19735, opened 2026-03-24
[PM-34012] - Replace image in welcome dialog with extension prompt

jaasen-livefront:
## 🎟️ Tracking

cherry pick of [2c0346e](https://github.com/bitwarden/clients/commit/2c0346e52a9057ffeb225c7b32ca63b67683a062)


apps/web/src/app/vault/components/web-vault-extension-prompt/web-vault-extension-prompt-dialog.component.html
```diff
@@ -1,11 +1,13 @@
 <div
   class="tw-max-w-3xl tw-p-9 tw-border tw-border-solid tw-border-border-base tw-rounded-xl tw-bg-background tw-text-center"
 >
-  <img
-    src="/images/vault/extension-mock-login.png"
-    [alt]="'extensionPromptImageAlt' | i18n"
-    class="tw-mb-[1.875rem]"
-  />
+  <div [attr.alt]="'extensionPromptImageAlt' | i18n" class="tw-mb-[1.875rem]">
```
nick-livefront, line 4: There is a duplicate `alt` here with the `svg` and `div`, with my quick testing it doesn't impact the screen reader experience so I don't think this needs to be held up. Can you raise a PR to remove from `main` though?

jaasen-livefront: @nick-livefront Ah nice catch, not sure how that slipped in. I'll fix that in main.

PR #19737, opened 2026-03-24
Implement the wrapper service

cyprain-okeke:

### DIFF
--- a/apps/browser/src/autofill/services/autofill-triage.service.spec.ts
+++ b/apps/browser/src/autofill/services/autofill-triage.service.spec.ts
@@ -196,9 +196,7 @@ describe("AutofillTriageService", () => {
 
       const result = service.triageField(mockField, mockPageDetails);
 
-      result.conditions.forEach((condition) => {
-        expect(condition.subConditions).toEqual([]);
-      });
+      expect(result.conditions.length).toBeGreaterThan(0);
     });
 
     it("should call qualification service for each check", () => {


apps/browser/src/autofill/services/autofill-triage.service.ts
```diff
@@ -0,0 +1,275 @@
...
+   * @returns Flat condition result with pass/fail status
+   */
+  private checkCondition(description: string, fn: () => boolean): AutofillTriageConditionResult {
+    const passed = fn();
+    return {
+      description,
+      passed,
+      subConditions: [], // PoC limitation: no sub-condition visibility
+    };
+  }
```
kdenney: This whole function seems overkill since all it does is evaluate the function and map. Perhaps this could be replaced by a `AutofillTriageConditionResult.from()` method? Or honestly even just building the condition result object inline in the `triageField` method above.

apps/browser/src/autofill/types/autofill-triage.ts
```diff
@@ -0,0 +1,75 @@
...
+  /**
+   * Whether this condition passed (true) or failed (false).
+   */
+  passed: boolean;
+
+  /**
+   * Optional nested sub-conditions that contributed to this result.
+   * Allows for hierarchical visibility into complex checks.
+   */
+  subConditions: AutofillTriageConditionResult[];
```
kdenney: IMO we should just remove subConditions for now since this solution doesn't require it and we get quite a bit of good data this way so it might not even be needed in the future.

apps/browser/src/autofill/services/autofill-triage.service.ts
```diff
@@ -0,0 +1,275 @@
...
+      return "creditCard";
+    }
+    if (conditions.find((c) => c.description === "Is for identity form" && c.passed)) {
+      return "identity";
+    }
+    if (conditions.find((c) => c.description === "Is for account creation form" && c.passed)) {
+      return "accountCreation";
+    }
+
+    // Check field-level qualifications (fallback)
```
kdenney: I don't think this fallback section below is needed. The actual browser extension autofill code in `AutofillOverlayContentService` only relies on the top-level methods so that should be good enough for this imo.

apps/browser/src/autofill/services/autofill-triage.service.ts
```diff
@@ -0,0 +1,275 @@
...
+    return {
+      description,
+      passed,
+      subConditions: [], // PoC limitation: no sub-condition visibility
+    };
+  }
+
+  /**
+   * Derives a human-readable qualification category based on which checks passed.
+   * Priority: login > creditCard > identity > accountCreation > ineligible
```
kdenney: This is slightly different from the actual priority order in `AutofillOverlayContentService`. That service uses this priority: login > creditCard > accountCreation > identity

This should match `AutofillOverlayContentService` to be consistent.

apps/browser/src/autofill/services/autofill-triage.service.ts
kdenney: After we remove the fallback section, all that is left is the 4 checks and the "ineligible" else condition. I would prefer if this didn't find by the strings "Is for login form" etc. 

What do you think about moving this logic to the `triageField` method above and using local boolean const vars to keep track of what each of the 4 methods returned and then build the `qualifiedAs` there? Or, if you have another idea to avoid finding by these magic string values?

PR #19745, opened 2026-03-24
[PM-33301] Prevent Unverified Bank Account from Upgrade to Premium

sbrown-livefront:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-33301

## 📔 Objective

Adds explicit handling of unsupported bank account payment methods for upgrading through vault or through the subscription modal.

* The `showBankAccountOption` input is set to `false` in the payment method display, ensuring users cannot select a bank account as a payment method during the upgrade.

* Added explicit checks to prevent upgrades using unverified bank accounts, displaying a clear error toast if attempted. The upgrade process now also validates the presence of an organization name, billing address, and payment method, throwing descriptive errors if any are missing.

Note:
There is no way to add a bank account for a premium user through the UI, but paths to upgrade using an unverified account are blocked in the client and the server.

## Premium User Cannot Upgrade Using Bank Account Option

https://github.com/user-attachments/assets/f6c9246c-eef2-4d5c-b344-b1c3b007f806

## Premium User with Paypal Upgrades with New Payment Method in New Organization Path

https://github.com/user-attachments/assets/f03eff5b-75a0-472e-b69f-c04454e596d5

## Premium User with Account Credit Upgrades with New Payment Method  New Organization Path

https://github.com/user-attachments/assets/9d74b750-939c-4a8d-82d6-3e928341cab2

## Organization User can Use Bank Account For Upgrade

https://github.com/user-attachments/assets/831388c4-0308-44f9-a3e1-8487a4e60e43


amorask-bitwarden (commented):
This PR looks great, but I just wanted to clarify one thing that I might have not mentioned. If the customer already has a **verified** bank account, they can definitely use that for the upgrade.

The problem stems from us having to handle bank account verification (async / depends on user action) in what is essentially an entirely synchronous operation. If their bank account is already set up, we don't have that issue.

You can check whether the customer has a verified bank account by getting their masked payment method and checking if the `hostedVerificationUrl` exists:
https://github.com/bitwarden/clients/blob/main/apps/web/src/app/billing/payment/types/masked-payment-method.ts#L42-L47

Let me know if that makes sense!

amorask-bitwarden (dismissed):

amorask-bitwarden (approved):

PR #19754, opened 2026-03-25
Innovation sprint/autotriage/context menu

kdenney:


apps/browser/src/autofill/background/context-menus.background.ts
```diff
@@ -39,6 +40,11 @@ export default class ContextMenusBackground {
             });
           }
         }
+
+        // the autofill triage popout component calls this to fetch the triage result
+        if (msg.command === "getAutofillTriageResult") {
+          sendResponse(this.contextMenuClickedHandler.latestTriageResult ?? null);
```
cyprain-okeke, line 46: I think Chrome extension message listeners require returning true when using sendResponse asynchronously, otherwise the message port closes. This could cause race conditions.

  Maybe we should add return true; after calling sendResponse, or restructure to handle synchronously.

kdenney: Good call, thanks! Added in latest push.

PR #19761, opened 2026-03-25
[PM-33580] fix: skip change-password notification when no ciphers exist for URL

bensbits91:
## 🎟️ Tracking

[PM-33580](https://bitwarden.atlassian.net/browse/PM-33580)

## 📔 Objective

Resolve issue where empty "Update existing login" popup appears after generating a password for a site for which you have not saved a cipher. This PR changes the behavior for sites without a saved cipher--prompts user to save a new cipher--while retaining the behavior for sites with a saved cipher--prompts to update existing cipher.

[PM-33580]: https://bitwarden.atlassian.net/browse/PM-33580?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## 📸  Screenshots

The following videos demonstrate the issue before this PR and the after fix introduced by this PR.

### Before this PR

https://github.com/user-attachments/assets/56bc920b-effb-432d-aefe-b4d3539017ac

### After this PR

https://github.com/user-attachments/assets/e6147a91-816f-413a-91b0-a319aa11b233


apps/browser/src/autofill/background/notification.background.ts
```diff
@@ -1052,23 +1052,40 @@ export default class NotificationBackground {
...
-          loginDomain,
-          // @TODO handle empty strings / incomplete data structure
-          data.newPassword || data.password,
-          tab,
-        );
+        if (ciphersForURL.length > 0 && changePasswordNotificationIsEnabled) {
+          await this.pushChangePasswordToQueue(
+            ciphersForURL.map((c) => c.id),
+            loginDomain,
+            // @TODO handle empty strings / incomplete data structure
```
audreyality, line 1065: ⛏️ Using `FIXME` aligns with department practices for identifying tech debt.

```suggestion
            // FIXME: handle empty strings / incomplete data structure
```

apps/browser/src/autofill/background/notification.background.ts
```diff
@@ -1052,23 +1052,40 @@ export default class NotificationBackground {
...
-        await this.pushChangePasswordToQueue(
-          ciphersForURL.map((c) => c.id),
-          loginDomain,
-          // @TODO handle empty strings / incomplete data structure
-          data.newPassword || data.password,
-          tab,
-        );
+        if (ciphersForURL.length > 0 && changePasswordNotificationIsEnabled) {
+          await this.pushChangePasswordToQueue(
+            ciphersForURL.map((c) => c.id),
```
audreyality, line 1063: 🎨 Consider extracting this `map` into a variable so that it can be inspected easily in the debugger.

apps/browser/src/autofill/background/notification.background.ts
```diff
@@ -1052,23 +1052,40 @@ export default class NotificationBackground {
...
+          );
+
+          return true;
+        }
 
-        return true;
+        // No existing ciphers for this URL — offer to save the generated password as a new login.
+        // The cipher may lack a username, but that is easier for the user to fix than losing
+        // a generated password they cannot easily retrieve.
+        if (ciphersForURL.length === 0 && newLoginNotificationIsEnabled) {
```
audreyality, line 1076: ⛏️ The length check is unnecessary. An array cannot have fewer than 0 elements.

```suggestion
        if (newLoginNotificationIsEnabled) {
```

audreyality (approved):
Some nits and a suggestion. Nothing blocking.

apps/browser/src/autofill/background/notification.background.ts
bensbits91: 🤔 I think this check is still needed. The preceding conditional `if (ciphersForURL.length > 0 && changePasswordNotificationIsEnabled) {...}` only fires when both ciphersForURL.length > 0 && changePasswordNotificationIsEnabled. If that block doesn't fire — either because there are no ciphers, or because changePasswordNotificationIsEnabled is false — we fall through to the === 0 check. In that case ciphersForURL.length could still be > 0 (the disabled-notification scenario), so the === 0 guard is important. Let me know if I'm missing something (always very possible).

audreyality: 🤦🏻‍♀️ You're correct; I can't believe I missed that.

PR #19767, opened 2026-03-25
Wire up the magnify render process ipc to the main process

coltonhurst:
Wire up the magnify render process IPC with an example chrome version call

### DIFF
--- a/apps/desktop/scripts/start.js
+++ b/apps/desktop/scripts/start.js
@@ -4,13 +4,19 @@ const { exec } = require("node:child_process");
 const concurrently = require("concurrently");
 const rimraf = require("rimraf");
 
+const BUILD_MAGNIFY = "nx run desktop:build-magnify";
+const BUILD_MAGNIFY_PRELOAD = "nx run desktop:build-magnify-preload";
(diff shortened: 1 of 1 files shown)


apps/desktop/scripts/start.js
```diff
@@ -8,45 +8,76 @@ const args = process.argv.splice(2);
...
+        // eslint-disable-next-line no-console
+        console.error("Could not build Magnify preload: ", err);
+        reject(err);
+        return;
+      }
+
+      // eslint-disable-next-line no-console
+      console.log("Magnify preload built successfully: \n", output);
+      resolve();
+    });
```
neuronull, line 49: nit, feel free to ignore as this is POC work: it seems this is duplicate logic to the result handling from the other exec call, could potentially be extracted to a helper

coltonhurst: Yeah, agreed. I was refactoring this a tiny bit and then hopped on a call with @eliykat and didn't finish the refactoring haha.

I did a bit in 476a8b869cca87c3eeb0df314aaf7c179971bf7a, we can refactor it further as needed

PR #19769, opened 2026-03-25
[PM-34139] upload page

itsadrago:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-34139

## 📔 Objective

First attempt at the upload page, using file input instead of drop zone per internal discussion


apps/web/src/app/tools/receive/receive-file-upload.component.html
```diff
@@ -0,0 +1,30 @@
+<bit-section>
+  <bit-form-field>
+    <bit-label for="file">{{ "file" | i18n }}</bit-label>
+    <div class="tw-flex tw-mt-2 tw-mb-1">
+      <button bitButton type="button" buttonType="secondary" (click)="fileSelector.click()">
+        {{ "chooseFile" | i18n }}
+      </button>
+      <span
+        class="tw-flex tw-items-center tw-pl-3"
+        [ngClass]="fileName() ? 'tw-text-main' : 'tw-text-muted'"
```
Hinton: **suggestion:** Swap out `ngClass` for `class` and you can drop `CommonModule` from the file.

apps/web/src/app/tools/receive/receive-file-upload.component.ts
```diff
@@ -0,0 +1,113 @@
...
+  FormFieldModule,
+  SectionComponent,
+  ToastService,
+} from "@bitwarden/components";
+
+@Component({
+  selector: "app-receive-upload",
+  templateUrl: "receive-file-upload.component.html",
+  changeDetection: ChangeDetectionStrategy.OnPush,
+  imports: [ButtonModule, CommonModule, FormFieldModule, JslibModule, SectionComponent],
```
Hinton: What do we use from JslibModule? It's deprecated and we should avoid it.

itsadrago: It was being used to enable i18n functionality in the component html, but I'm going to switch to depending on `I18nPipe` directly

PR #19771, opened 2026-03-25
refactor(scheduling): extract @bitwarden/scheduling Nx leaf library

addisonbeck:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-34200

## 📔 Objective

Extract the scheduling domain from `libs/common/src/platform/scheduling/` into a dedicated Nx leaf library at `libs/scheduling/` (`@bitwarden/scheduling`).

This is a purely structural refactor — no behavior changes.

### Design decisions

- This re-exports the migrated files from their original locations to avoid updating callers. This is conventional for migrating code out of the common libraries.
- `@bitwarden/logging` is technically not a new dependency, but previously we were importing those files through shims. Now we are directly referencing the logging lib.


libs/scheduling/src/default-task-scheduler.service.spec.ts
```diff
@@ -1,7 +1,7 @@
 import { mock, MockProxy } from "jest-mock-extended";
 
-import { LogService } from "../abstractions/log.service";
-import { ScheduledTaskNames } from "../scheduling/scheduled-task-name.enum";
+import { LogService } from "@bitwarden/common/platform/abstractions/log.service";
```
djsmith85: ⚠️  This should also use `import { LogService } from "@bitwarden/logging";`

djsmith85 (changes requested):
Nice one. Changes are looking good. Just one small nit

libs/scheduling/src/default-task-scheduler.service.spec.ts
addisonbeck: Good catch, thanks. Fixed on https://github.com/bitwarden/clients/pull/19771/commits/d121976a138ff73baefc648676b20e2e7e83bc72

libs/scheduling/src/default-task-scheduler.service.spec.ts
```diff
@@ -1,7 +1,7 @@
 import { mock, MockProxy } from "jest-mock-extended";
 
-import { LogService } from "../abstractions/log.service";
-import { ScheduledTaskNames } from "../scheduling/scheduled-task-name.enum";
+import { LogService } from "@bitwarden/logging";
+import { ScheduledTaskNames } from "@bitwarden/scheduling";
```
djsmith85: Sorry, found another nit:
As the imported file is directly located next to this it should be 
`import { ScheduledTaskName } from "./scheduled-task-name.enum";` instead of `"@bitwarden/scheduling"`

djsmith85 (changes requested):

libs/scheduling/src/default-task-scheduler.service.spec.ts
addisonbeck: No worries, I appreciate it. Fixed on https://github.com/bitwarden/clients/pull/19771/commits/a83db78cf8eea16533fa40075e2a1b9b650bda9d

djsmith85 (dismissed):

addisonbeck:
@djsmith85 sorry, had to remove an extra newline. Mind giving this another review?

djsmith85 (approved):

PR #19774, opened 2026-03-26
Magnify/desktop ipc handler

ike-kottlowski:


apps/desktop/src/autofill/models/magnify-command.ts
```diff
@@ -0,0 +1,81 @@
...
+ * Fields that can be copied from a vault item via the CopyVaultItemField command.
+ * CRITICAL: The resolved field value is copied to clipboard inside the Bitwarden
+ * renderer — decrypted values never cross IPC to the Magnify window.
+ */
+export const VaultItemField = Object.freeze({
+  Password: "password",
+  Username: "username",
+  Totp: "totp",
+  CardNumber: "cardNumber",
+  CardCode: "cardCode",
```
neuronull, line 22: (opinion): It could be nice to comment these out (and add a code comment) or remove until there is implementation to support them. Mostly because it can be misleading as a reader of the code, to have to ascertain that the plumbing is there but it's not hooked up yet. In our case it seems fairly small changes to add the support for future vault items to this layer.

apps/desktop/src/autofill/models/magnify-command.ts
```diff
@@ -0,0 +1,81 @@
...
+} as const);
+export type VaultItemField = (typeof VaultItemField)[keyof typeof VaultItemField];
+
+/**
+ * A slim, display-only representation of a vault item for the Magnify UI.
+ * CRITICAL: Never include decrypted passwords, notes, or encryption keys.
```
neuronull, line 28: Is this by chance a claude doc comment?  this could be done in a follow up post poc but I find the claude style documentation comments more distracting than helpful oftentimes.

apps/desktop/src/autofill/models/magnify-command.ts
```diff
@@ -0,0 +1,81 @@
...
+  name: string;
+  username?: string;
+  uri?: string;
+};
+
+/**
+ * Input payload for the VaultSearch command.
+ */
+export type VaultSearchInput = {
+  query: string;
```
neuronull, line 41: do we want to impose any rules on the query at this level?

apps/desktop/src/autofill/models/magnify-command.ts
```diff
@@ -0,0 +1,81 @@
...
+};
+
+/**
+ * Internal request forwarded from Main process to the Bitwarden renderer.
+ * Includes a correlationId for matching responses.
+ */
+export type MagnifyRequest = {
+  command: MagnifyCommand;
+  input: MagnifyCommandInput;
+  correlationId: string;
```
neuronull, line 70: nit: "correlationId" is hard for my brain to grok. Would `requestId` be representative? To me, "correlation" is the action done by the code that uses the data structures, and the correlation is being done based on the request.

apps/desktop/src/autofill/services/magnify-command-handler.service.ts
```diff
@@ -0,0 +1,182 @@
...
+    const filter =
+      input.cipherTypes?.length > 0 ? (c: CipherView) => input.cipherTypes.includes(c.type) : null;
+
+    const matched = await this.searchService.searchCiphers(userId, input.query, filter, allCiphers);
+
+    const results: MagnifyVaultItemResult[] = matched.slice(0, MAX_SEARCH_RESULTS).map((c) => ({
+      id: c.id,
+      name: c.name,
+      username: c.login?.username,
+      uri: c.login?.uris?.[0]?.uri,
```
neuronull, line 96: The first URI in the array is used in other production bits of code but just calling it out here as a ux/product note ; I guess if the URI is only displayed as informational that this is ok

apps/desktop/src/autofill/services/magnify-command-handler.service.ts
```diff
@@ -0,0 +1,182 @@
...
+    const allCiphers = await this.cipherService.getAllDecrypted(userId);
+    const cipher = allCiphers.find((c) => c.id === input.vaultItemId);
+
+    if (cipher == null) {
+      this.sendResponse({
+        command: request.command,
+        correlationId: request.correlationId,
+        results: [],
+        error: "Vault item not found",
+      });
```
neuronull, line 130: nit/suggestions (can be post POC, non blocking): constructing an error case response happens multiple times in the file, we could have a helper that does that and takes the `request` as an argument, along with an error string

apps/desktop/src/magnify/magnify-preload.ts
```diff
@@ -0,0 +1,16 @@
...
+
+const magnify = {
+  /**
+   * Send a command to the Bitwarden renderer and get results back.
+   */
+  sendCommand: (
+    command: string,
+    input: string,
+  ): Promise<{ results: MagnifyCipherResult[]; error?: string }> =>
+    ipcRenderer.invoke("autofill.magnifyCommand", { command, input }),
```
neuronull: nit: "magnifyCommand" is also a raw string in the ipc channels ts file.  We could define these as consts somewhere.

apps/desktop/src/autofill/models/magnify-command.ts
```diff
@@ -0,0 +1,81 @@
...
+
+/**
+ * Response sent from the Bitwarden renderer back through the Main process to Magnify.
+ */
+export type MagnifyResponse = {
+  command: MagnifyCommand;
+  correlationId: string;
+  results: MagnifyVaultItemResult[];
+  error?: string;
+};
```
neuronull, line 81: I find it slightly confusing to have the two magnify request types but one Response type 🤔 maybe its just me though. I wonder if more descriptive names of the types could help. I presume the two request types was an intentional choice to avoid making requestId optional, which I get.

apps/desktop/src/autofill/services/magnify-command-handler.service.ts
```diff
@@ -0,0 +1,182 @@
...
+
+/**
+ * Angular service in the Bitwarden renderer process that handles
+ * Magnify commands relayed from the Main process.
+ *
+ * This service owns all vault operations for Magnify — decrypted data
+ * never leaves this process. Clipboard writes happen here, not in Magnify.
+ */
+@Injectable()
+export class MagnifyCommandHandlerService {
```
neuronull, line 32: could we add some logging (debug / info / error) to the logic in this service?

apps/desktop/src/autofill/models/magnify-command.ts
```diff
@@ -0,0 +1,81 @@
...
+
+/**
+ * Response sent from the Bitwarden renderer back through the Main process to Magnify.
+ */
+export type MagnifyResponse = {
+  command: MagnifyCommand;
+  correlationId: string;
+  results: MagnifyVaultItemResult[];
+  error?: string;
```
neuronull, line 80: I guess the same logic for descriminated unions could apply for the response type? Here the `results` and `error` decide the state. We could instead have a `MagnifySuccessResponse` and a `MagnifyFailureResponse`

apps/desktop/src/autofill/services/magnify-command-handler.service.ts
```diff
@@ -0,0 +1,182 @@
...
+      case VaultItemField.Totp:
+        return cipher.login?.totp;
+      case VaultItemField.CardNumber:
+        return cipher.card?.number;
+      case VaultItemField.CardCode:
+        return cipher.card?.code;
+      default:
+        return undefined;
+    }
+  }
```
neuronull, line 172: 💭  would this make sense to put in the magnify commands TS file instead, and other consumers can use it as a helper

apps/desktop/src/autofill/services/magnify-command-handler.service.ts
```diff
@@ -0,0 +1,182 @@
...
+
+  private async getActiveUserId(): Promise<UserId | null> {
+    const account = await firstValueFrom(this.accountService.activeAccount$);
+    return account?.id ?? null;
+  }
+
+  private sendResponse(response: MagnifyResponse): void {
+    ipc.autofill.sendMagnifyCommandResponse(response);
+  }
```
neuronull, line 181: this one-liner function isn't super helpful, I wonder if it could be modified or if it makes sense to remove.

apps/desktop/src/autofill/models/magnify-command.ts
ike-kottlowski: It is, I'll take a look.

ike-kottlowski:
Abandoned branch for Magnify.

PR #19785, opened 2026-03-26
[PM-34131] Editing ciphers with change at risk password banner fails on web

jengstrom-bw:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-34131

## 📔 Objective

Fixes changePasswordUrl resource error state blocking the edit cipher UI.


libs/vault/src/cipher-view/cipher-view.component.ts
```diff
@@ -284,12 +284,24 @@ export class CipherViewComponent {
...
+        return await this.changeLoginPasswordService.getChangePasswordUrl(params.cipher);
+      } catch (e: any) {
+        this.logService.error(e.message);
+        return undefined;
       }
-      return await this.changeLoginPasswordService.getChangePasswordUrl(params.cipher);
     },
   });
 
+  readonly changePasswordLink = computed(() => {
```
shane-melton, line 298: ❓ I might be misunderstanding `resource()` but why do we need to wrap it in another `computed()` signal?

shane-melton (approved):

PR #19787, opened 2026-03-26
[INNO] - Passkey Report

jrmccannon:
## Associated PRs
- Server : https://github.com/bitwarden/server/pull/7317

## 📔 Objective
Adds a passkey report that shows users which of their logins could be upgraded with passkey authentication but haven't been yet. It pulls a directory of passkey-capable services from the server, matches them against the user's login items by domain, and flags any that are missing FIDO2 credentials. The report shows whether each service supports passwordless login, MFA, or both, and links out to setup instructions where available. There's an individual version that covers your personal vault and org memberships (with filtering), and an org-level version for admins. Requires premium and the PasskeyLoginReport feature flag.

## 📸 Screenshots
Vault report page
image

Vault Report
image

Admin Console Report Page
image

Admin Console Report
image


apps/web/src/app/admin-console/organizations/reporting/reports-home.component.ts
```diff
@@ -117,7 +123,7 @@ export class ReportsHomeComponent implements OnInit, AfterViewInit, OnDestroy {
     }
   }
 
-  private buildReports(productType: ProductTierType): ReportEntry[] {
+  private buildReports(productType: ProductTierType, passkeyReportEnabled: boolean): ReportEntry[] {
```
BTreston: 🎨 probably easier for clean up to have this function `async`, change its caller to use `switchMap` and define the feature flag in the body rather than pass it in as a param. Non-blocking though

apps/web/src/app/dirt/reports/pages/organizations/org-passkey-report.component.ts
```diff
@@ -0,0 +1,263 @@
...
+    this.organization$
+      .pipe(
+        concatMap(async (org) => {
+          const userId = await firstValueFrom(this.userId$);
+          this.manageableCiphers.set(await this.cipherService.getAll(userId));
+          await this.load(org);
+        }),
+        takeUntilDestroyed(),
+      )
+      .subscribe();
```
BTreston: `effect` is the modern angular version of a `subscribe` to drive reactive side-effects
```typescript
effect(() => {
    const org = this.organization();
    untracked(async () => {
      const userId = await firstValueFrom(this.userId$);
      this.manageableCiphers.set(await this.cipherService.getAll(userId));
      await this.load(org);
    });
  });
```

apps/web/src/app/dirt/reports/pages/organizations/org-passkey-report.component.ts
```diff
@@ -0,0 +1,263 @@
...
+        switchMap((userId) =>
+          this.organizationService.organizations$(userId).pipe(getById(params.organizationId)),
+        ),
+        filter((organization): organization is Organization => organization != null),
+        shareReplay({ refCount: true, bufferSize: 1 }),
+      ),
+    ),
+  );
+
+  protected readonly organization = toSignal(this.organization$);
```
BTreston: Considering the previous suggestion, might be better to just wrap this into a single class member

apps/web/src/app/dirt/reports/pages/organizations/org-passkey-report.component.ts
```diff
@@ -0,0 +1,263 @@
...
+  protected readonly hasLoaded = signal(false);
+  protected readonly ciphers = signal<PasskeyCipherRow[]>([]);
+  protected readonly dataSource = new TableDataSource<PasskeyCipherRow>();
+
+  // Private state
+  private readonly manageableCiphers = signal<Cipher[]>([]);
+  private readonly passkeyServices = signal<Map<string, PasskeyServiceEntry>>(new Map());
+
+  // Observable streams
+  private readonly userId$ = this.accountService.activeAccount$.pipe(getUserId);
```
BTreston: This one is probably also a good candidate to be wrapped in a signal. That way you can just do `userId()` instead of all the `await firstValueFrom`

apps/web/src/app/dirt/reports/pages/organizations/org-passkey-report.component.ts
```diff
@@ -0,0 +1,263 @@
...
+    try {
+      if (this.passkeyServices().size === 0) {
+        const userId = await firstValueFrom(this.userId$);
+        const entries = await this.passkeyDirectoryApiService.getPasskeyDirectory(userId);
+        this.passkeyServices.set(
+          entries
+            .filter((x) => x.domainName != null)
+            .reduce(
+              (map, entry) => map.set(entry.domainName, entry),
+              new Map<string, PasskeyServiceEntry>(),
```
BTreston: :art: this should be in the definition

apps/web/src/app/dirt/reports/pages/organizations/org-passkey-report.component.ts
```diff
@@ -0,0 +1,263 @@
...
+            ),
+        );
+      }
+    } catch (e) {
+      this.logService.error("[OrgPasskeyReportComponent] Failed to load passkeys", e);
+    }
+
+    if (this.passkeyServices().size === 0) {
+      return;
+    }
```
BTreston, line 189: :art: we can return early inside that `try/catch` block above if `entries` is empty.

apps/web/src/app/dirt/reports/pages/passkey-report.component.ts
```diff
@@ -0,0 +1,321 @@
...
+  protected readonly selectedFilterChip = "0";
+  protected readonly filterOrgStatus$ = new BehaviorSubject<number | string>(0);
+  private readonly maxItemsToSwitchToChipSelect = 5;
+
+  // Organization state
+  private readonly organizations = signal<Organization[]>([]);
+  protected readonly organizations$ = this.accountService.activeAccount$.pipe(
+    getUserId,
+    switchMap((userId) => this.organizationService.organizations$(userId)),
+  );
```
BTreston: same feedback as the other file, just one class member signal for this:

```typescript
protected readonly organizations = toSignal(this.accountService.activeAccount$.pipe(
    getUserId,
    switchMap((userId) => this.organizationService.organizations$(userId)),
  ));
```

apps/web/src/app/dirt/reports/pages/passkey-report.component.ts
```diff
@@ -0,0 +1,321 @@
...
+
+  // Private state
+  private readonly passkeyServices = signal<Map<string, PasskeyServiceEntry>>(new Map());
+  private readonly userId$ = this.accountService.activeAccount$.pipe(getUserId);
+  private readonly currentFilterStatus = signal<number | string>(0);
+
+  constructor() {
+    this.organizations$
+      .pipe(takeUntilDestroyed())
+      .subscribe((orgs) => this.organizations.set(orgs));
```
BTreston: We can delete this with the signal based approach

apps/web/src/app/dirt/reports/pages/passkey-report.component.ts
```diff
@@ -0,0 +1,321 @@
...
+  // Organization state
+  private readonly organizations = signal<Organization[]>([]);
+  protected readonly organizations$ = this.accountService.activeAccount$.pipe(
+    getUserId,
+    switchMap((userId) => this.organizationService.organizations$(userId)),
+  );
+
+  // Private state
+  private readonly passkeyServices = signal<Map<string, PasskeyServiceEntry>>(new Map());
+  private readonly userId$ = this.accountService.activeAccount$.pipe(getUserId);
```
BTreston: also same feedback here:

```typescript
private readonly userId = toSignal(this.accountService.activeAccount$.pope(getUserId));
```

apps/web/src/app/dirt/reports/pages/passkey-report.component.ts
```diff
@@ -0,0 +1,321 @@
...
+            ),
+        );
+      }
+    } catch (e) {
+      this.logService.error("[PasskeyReportComponent] Failed to load passkeys", e);
+    }
+
+    if (this.passkeyServices().size === 0) {
+      return;
+    }
```
BTreston, line 142: also same feedback from the other file.

BTreston: Actually, this whole block looks identical to the other file. Can we review and consolidate all the duplicated logic between these two components into a shared service? Claude should be able to do it relatively easily.

apps/web/src/app/dirt/reports/pages/passkey-report.component.ts
```diff
@@ -0,0 +1,321 @@
...
+  buildPasskeyCipherRow,
+  getPasskeyServiceMatch,
+  processPasskeyCiphers,
+} from "./passkey-report.utils";
+
+@Component({
+  changeDetection: ChangeDetectionStrategy.OnPush,
+  selector: "app-passkey-report",
+  templateUrl: "passkey-report.component.html",
+  standalone: false,
```
BTreston: If possible, we should make this component standalone as well

libs/common/src/dirt/models/response/passkey-directory-entry.response.ts
```diff
@@ -0,0 +1,16 @@
+import { BaseResponse } from "../../../models/response/base.response";
+
+export class PasskeyDirectoryEntryResponse extends BaseResponse {
+  domainName: string;
+  instructions: string;
```
BTreston: I think `string | undefined` would fit better than `?? "";`

libs/common/src/dirt/services/abstractions/passkey-directory-api.service.ts
BTreston, line 1: I believe the naming convention we use on the front-end is:

- abstraction: `PasskeyDirectoryApiService`
- concrete implementation: `DefaultPasskeyDirectoryApiService`

can we rename these files and classes to align with that? TY

BTreston (changes requested):
Overall looks really good, just a lot of "modern angular" nitpicks and some general clean up for when/if this gets merged.

apps/web/src/app/admin-console/organizations/reporting/reports-home.component.ts
```diff
@@ -73,7 +76,7 @@ export class ReportsHomeComponent implements OnInit, AfterViewInit, OnDestroy {
           .organizations$(userId)
           .pipe(getOrganizationById(params.organizationId)),
       ),
-      map((org) => this.buildReports(org.productTierType)),
+      switchMap(async (org) => await this.buildReports(org.productTierType)),
```
BTreston: 🎨 No need for `async` and `await`. You can just do:
`      switchMap((org) => this.buildReports(org.productTierType)),`

apps/web/src/app/dirt/reports/pages/organizations/org-passkey-report.component.ts
```diff
@@ -0,0 +1,267 @@
...
+      const organization = this.organization();
+
+      if (organization == null) {
+        return;
+      }
+
+      await untracked(async () => {
+        await this.load(organization);
+      });
+    });
```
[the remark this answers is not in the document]
BTreston: The robot is correct

apps/web/src/app/dirt/reports/pages/passkey-report.component.ts
```diff
@@ -0,0 +1,319 @@
...
+
+  // Organization state
+  protected readonly organizations = toSignal(
+    this.accountService.activeAccount$.pipe(
+      getUserId,
+      switchMap((userId) => this.organizationService.organizations$(userId)),
+    ),
+  );
+
+  private readonly userIdNullable = toSignal(this.accountService.activeAccount$.pipe(getUserId));
```
BTreston: Ok, after doing some more reading I think we actually can use `requireSync` here. The reason being because this is stream is backed by a `BehaviorSubject` which DOES actually emit a synchronous value on subscription. The same, however is NOT true for `organizations$` which was what threw me off. Apologies for sending you through this hoop.

Also generally we should avoid type assertions using `as` to satisfy TS errors.

apps/web/src/app/dirt/reports/pages/passkey-report.component.ts
```diff
@@ -0,0 +1,319 @@
...
+    }
+
+    try {
+      const entries = (await this.passkeyDirectoryApiService.getPasskeyDirectory(this.userId()))
+        .filter((x) => x.domainName != null)
+        .reduce(
+          (map, entry) => map.set(entry.domainName, entry),
+          new Map<string, PasskeyServiceEntry>(),
+        );
+
```
BTreston: ❓ have we done away with the early return in favor of `.size` check at all the call sites?

apps/web/src/app/dirt/reports/pages/passkey-report.service.ts
```diff
@@ -0,0 +1,144 @@
...
+
+    return [...currentRows];
+  }
+
+  /**
+   * Builds a PasskeyCipherRow from a cipher and its matching passkey service entry.
+   */
+  buildPasskeyCipherRow(cipher: CipherView, match: PasskeyServiceEntry): PasskeyCipherRow {
+    return {
+      cipher,
```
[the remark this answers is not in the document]
BTreston: The robot is also correct here

apps/web/src/app/dirt/reports/pages/passkey-report.service.ts
```diff
@@ -0,0 +1,144 @@
...
+
+export interface PasskeyServiceEntry {
+  instructions: string | undefined;
+  supportsPasskeyLogin: boolean;
+  supportsPasskeyMfa: boolean;
+}
+
+export interface PasskeyCipherRow {
+  cipher: CipherView;
+  instructions: string | null;
```
BTreston: ⛏️ prefer `undefined` over `null`;

BTreston (changes requested):
Just a few more little quality of life fixes and 1 questions. Otherwise looks good

BTreston (dismissed):
LGTM 👍

BryanCunningham (dismissed):

apps/web/src/app/dirt/reports/pages/passkey-report.component.html
```diff
@@ -0,0 +1,148 @@
...
+      {{ "passkeyLoginReportHowToUse" | i18n }}
+    </a>
+  </p>
+  @if (!hasLoaded() && loading()) {
+    <div>
+      <i
+        class="bwi bwi-spinner bwi-spin tw-text-muted"
+        title="{{ 'loading' | i18n }}"
+        aria-hidden="true"
+      ></i>
```
lastbestdev: Nit: I belive we prefer use of the `bit-icon` component here instead of `i` tag with `bwi` class

apps/web/src/app/dirt/reports/pages/passkey-report.component.html
```diff
@@ -0,0 +1,148 @@
...
+      rel="noreferrer"
+    >
+      {{ "passkeyLoginReportHowTheyWork" | i18n }}
+    </a>
+    {{ "passkeyLoginReportAnd" | i18n }}
+    <a bitLink href="https://bitwarden.com/help/storing-passkeys/" target="_blank" rel="noreferrer">
+      {{ "passkeyLoginReportHowToUse" | i18n }}
+    </a>
+  </p>
+  @if (!hasLoaded() && loading()) {
```
lastbestdev: Do we need to check on both `hasLoaded` and `loading` here? Feels like one of the two may be redundant

apps/web/src/app/dirt/reports/pages/passkey-report.component.html
```diff
@@ -0,0 +1,148 @@
...
+              <a
+                bitLink
+                href="#"
+                appStopClick
+                (click)="selectCipher(row.cipher)"
+                title="{{ 'editItemWithName' | i18n: row.cipher.name }}"
+                >{{ row.cipher.name }}</a
+              >
+              @if (row.cipher.organizationId) {
+                <i
```
lastbestdev: Nit: same as other, prefer `bit-icon`

apps/web/src/app/dirt/reports/pages/passkey-report.component.html
```diff
@@ -0,0 +1,148 @@
...
+                  class="bwi bwi-collection-shared tw-ml-1"
+                  appStopProp
+                  title="{{ 'shared' | i18n }}"
+                  aria-hidden="true"
+                ></i>
+                <span class="tw-sr-only">{{ "shared" | i18n }}</span>
+              }
+              @if (row.cipher.hasAttachments) {
+                <i
```
lastbestdev: Nit: same as other, prefer `bit-icon`

apps/web/src/app/dirt/reports/pages/passkey-report.component.html
```diff
@@ -0,0 +1,148 @@
...
+                  aria-hidden="true"
+                ></i>
+                <span class="tw-sr-only">{{ "attachments" | i18n }}</span>
+              }
+              <br />
+              <small>{{ row.cipher.subTitle }}</small>
+            </td>
+            <td bitCell>
+              @if (row.supportsPasskeyLogin) {
+                <i
```
lastbestdev: Nit: same, prefer `bit-icon`

apps/web/src/app/dirt/reports/pages/passkey-report.component.html
```diff
@@ -0,0 +1,148 @@
...
+                  class="bwi bwi-passkey tw-mr-1"
+                  title="{{ 'passwordlessLogin' | i18n }}"
+                  aria-hidden="true"
+                ></i>
+                <span class="tw-sr-only">{{ "passwordlessLogin" | i18n }}</span>
+              }
+              @if (row.supportsPasskeyMfa) {
+                <i
```
lastbestdev: Nit: same, prefer `bit-icon`

apps/web/src/app/dirt/reports/pages/organizations/org-passkey-report.component.ts
```diff
@@ -0,0 +1,262 @@
...
+  private readonly passkeyReportService = inject(PasskeyReportService);
+  private readonly passwordRepromptService = inject(PasswordRepromptService);
+  private readonly syncService = inject(SyncService);
+  private readonly adminConsoleCipherFormConfigService = inject(
+    AdminConsoleCipherFormConfigService,
+  );
+
+  // Reactive state
+  protected readonly loading = signal(false);
+  protected readonly hasLoaded = signal(false);
```
lastbestdev: Having two variables for loading state is redundant, we should remove one of them

apps/web/src/app/dirt/reports/pages/passkey-report.component.ts
```diff
@@ -0,0 +1,311 @@
...
+  private readonly userId = toSignal(this.accountService.activeAccount$.pipe(getUserId), {
+    requireSync: true,
+  });
+
+  protected readonly currentFilterStatus = signal<number | string>(0);
+  private readonly passkeyServices = signal<Map<string, PasskeyServiceEntry>>(new Map());
+
+  constructor() {
+    void this.init();
+  }
```
lastbestdev: As opposed to calling an `async` function in the synchronous `constructor`, we should make the component implement `OnInit` by using `async ngOnInit` instead for this logic.

apps/web/src/app/dirt/reports/pages/passkey-report.component.ts
```diff
@@ -0,0 +1,311 @@
...
+    } else {
+      this.dataSource.filter = () => true;
+    }
+  }
+
+  protected async filterOrgToggleChipSelect(filterId: string | null) {
+    await this.filterOrgToggle(filterId ?? 0);
+  }
+
+  private async init() {
```
lastbestdev: We could just change this method to `ngOnInit`

apps/web/src/app/dirt/reports/pages/passkey-report.service.ts
```diff
@@ -0,0 +1,160 @@
...
+        rows[index] = updatedRow;
+      } else if (updatedRow == null && index > -1) {
+        rows.splice(index, 1);
+      }
+
+      return rows;
+    }
+
+    return [...currentRows];
+  }
```
lastbestdev, line 147: Not critical: IMO this method feels a bit out of place in the service, since it is highly coupled to the table rows in the Passkey Report itself. I think this logic (determining when to remove rows (ciphers) from the table after they are updated) is better suited in the report component.

apps/web/src/app/dirt/reports/pages/passkey-report.component.ts
```diff
@@ -0,0 +1,311 @@
...
+  protected readonly hasLoaded = signal(false);
+  protected readonly error = signal(false);
+  protected readonly ciphers = signal<PasskeyCipherRow[]>([]);
+  protected readonly allCiphers = signal<PasskeyCipherRow[]>([]);
+  protected readonly dataSource = new TableDataSource<PasskeyCipherRow>();
+
+  // Filter state
+  protected readonly filterStatus = signal<(number | string)[]>([0]);
+  protected readonly showFilterToggle = signal(false);
+  protected readonly vaultMsg = signal("vault");
```
lastbestdev: Does this need to be a signal? Is there a reason it can't be a plain string or included in an english translation directly?

apps/web/src/locales/en/messages.json
```diff
@@ -2993,9 +2993,52 @@
...
+    "message": "Logins that support passkeys"
+  },
+  "passkeyLoginFoundReportDesc": {
+    "message": "We found $COUNT$ login(s) in your $VAULT$ that support passkeys. To further secure these accounts, consider setting up a passkey where available. Sourced from passkeys.2fa.directory",
+    "placeholders": {
+      "count": {
+        "content": "$1",
+        "example": "8"
+      },
+      "vault": {
```
lastbestdev: Can we just put "vault" directly in the message and remove this variable? It looks to me like that is always the value passed in by the components anyway

apps/web/src/app/dirt/reports/pages/organizations/org-passkey-report.component.html
lastbestdev, line 1: Nit: several instances of `i` with `bwi` class to be converted to `bit-icon` use in this file

apps/web/src/app/dirt/reports/pages/passkey-report.component.ts
```diff
@@ -0,0 +1,313 @@
...
+    } else if (typeof status === "string") {
+      const orgId = status as OrganizationId;
+      this.dataSource.filter = (r: PasskeyCipherRow) => r.cipher.organizationId === orgId;
+    } else {
+      this.dataSource.filter = () => true;
+    }
+  }
+
+  protected async filterOrgToggleChipSelect(filterId: string | null) {
+    await this.filterOrgToggle(filterId ?? 0);
```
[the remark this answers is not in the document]
lastbestdev: Bug fixed in latest

lastbestdev (approved):
Addressed review items from my earlier review. LGTM 👍🏻

r-tome (approved):

AlexRubik (approved):

vleague2 (approved):
UIF code looks fine and this is a large PR with existing approvals, so I am approving, but the fingerprint icon contrast with the blue background color is poor (according to the screenshots). Can you please get with @bostonjillianc to swap that out for a different SVG as a fast-follow before the next rc cut? (And whatever other design review notes she has!)

lastbestdev:
> UIF code looks fine and this is a large PR with existing approvals, so I am approving, but the fingerprint icon contrast with the blue background color is poor (according to the screenshots). Can you please get with @bostonjillianc to swap that out for a different SVG as a fast-follow before the next rc cut? (And whatever other design review notes she has!)

@vleague2 I will cut a follow up ticket to address this icon update. Thank you for the review!

PR #19790, opened 2026-03-26
Fix autofill prompt to save new credentials when the vault is locked

differsthecat:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-34100

## 📔 Objective

Fixes a bug where the autofill notification was not showing when the vault is locked.

## 📸 Screenshots

After fix:

https://github.com/user-attachments/assets/bac4ec08-464c-4b43-81a8-c724c081ee0d


differsthecat:
This is expected behavior following a rewrite of the notification-triggering logic. This scenario has proven to be very noisy for users in the past.
Closing.

PR #19804, opened 2026-03-27
[CL-1138] Add stacked drawer support with back button

willmartian:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/CL-1138

## 📔 Objective

Adds the ability to stack multiple drawers on top of each other with a back button, without closing the current drawer.

**API:**
- `dialogService.openDrawer(Component, config)` — unchanged; returns a `DrawerRef`
- `drawerRef.stack(Component, config)` — pushes a new component onto the drawer stack; `closeOnNavigation` is excluded from config since it is inherited from the root
- `dialogService.closeAll()` — now also closes all open drawers (previously only CDK dialogs)
- `DrawerRef` is now a public exported class and can be injected inside drawer components:
  ```ts
  private drawerRef = inject(DrawerRef, { optional: true });
  drawerRef?.stack(ChildComponent, { data: { ... } });
  ```
- `DialogRef.closePredicate` is a public, settable instance field (merged from main's PM-31364), so a drawer component can gate its own close based on its own state (e.g., unsaved changes)

**Behaviour:**
- The back button (`bwi-angle-left`) appears automatically in `bit-dialog` when stack depth > 1
- Back button and footer buttons (`bitDialogClose`) pop the top drawer and fire its `closed` observable
- The X button closes the entire stack top-to-bottom; if a drawer's `closePredicate` rejects, that drawer and any below it remain open (drawers above are already closed)
- `openDrawer` shares the same sequential close behavior for its upfront cleanup when called while drawers are open
- `closeOnNavigation` is evaluated on the root ref and governs the entire stack
- Auth-lock correctly fires `closed` on all drawer refs before clearing

**Refactors:**
- `DialogRef`, `DrawerRef`, `DialogConfig`, `CdkDialogRef` extracted to `dialog-ref.ts` to break the circular dependency between `DrawerService` and `DialogService`
- `DrawerService` now owns the `DrawerRef[]` stack (previously `Portal[]`) — single source of truth; `DialogService` no longer maintains a parallel stack
- `DrawerService.closeAll()` walks the stack top-to-bottom calling each `ref.close()`, stopping at the first rejection; `closeOnNavigation` subscription moved from `DialogService` into `DrawerService`
- `DrawerRef.close()` pops the top drawer; the X button calls `drawerService.closeAll()`

## 📸 Screenshots

https://github.com/user-attachments/assets/c1c956fa-5fe4-42d5-b3e1-03ecfb435886


libs/components/src/dialog/drawer.service.ts
```diff
@@ -19,18 +36,49 @@ export class DrawerService {
...
 
+  /** Clear the entire stack, firing each ref's closed observable from top to bottom. */
+  closeAll(): void {
+    const refs = [...this.stack()].reverse();
+    this.stack.set([]);
+    this.pushWidthPx.set(0);
+    this.isPushMode.set(false);
+    for (const ref of refs) {
+      ref.close();
+    }
```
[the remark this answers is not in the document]
willmartian: Fixed in [54eb6f2](https://github.com/bitwarden/clients/pull/19804/commits/54eb6f2f0fbd91859036cf947da7b608736665c8)

harr1424 (dismissed):
Tools owned file looks good! Appreciate the great comments throughout the new code.

BryanCunningham (dismissed):

libs/components/src/dialog/dialog.service.stories.ts
```diff
@@ -201,6 +215,45 @@ class LargeDrawerContentComponent {
   }
 }
 
+// FIXME(https://bitwarden.atlassian.net/browse/CL-764): Migrate to OnPush
+// eslint-disable-next-line @angular-eslint/prefer-on-push-component-change-detection
```
vleague2: Can we do onPush to avoid adding another fixme?

BryanCunningham (dismissed):
The Claude comment about the close function seems like it might need a quick look but, otherwise, seems good to go to me

willmartian:
@claude review please

libs/components/src/dialog/dialog.service.stories.ts
```diff
@@ -201,6 +218,54 @@ class LargeDrawerContentComponent {
   }
 }
 
+@Component({
+  template: `
+    <bit-dialog [title]="'Level ' + level + ' Drawer'">
+      <span bitDialogContent>
+        This is level {{ level }} of the drawer stack.
+        @if (level < 3) {
+          <br /><br />
```
BryanCunningham: Nit: Can we just use flex/gap or margin here? I know this is just a story but this just feels so wrong :)

libs/components/src/dialog/dialog.service.ts
```diff
@@ -355,36 +174,52 @@ export class DialogService {
...
-     * Similar to `this.open`, we get around this with mutability.
+     * Circular: we need the ref for the injector before we have the portal,
+     * and we need the portal to complete the ref. Solved with mutability (same
+     * pattern as openDialog / CdkDialogRef).
      */
-    this.activeDrawer = new DrawerDialogRef(this.drawerService, this.logService, config);
+    const ref: DrawerRef<R, C> = new DrawerRef<R, C>(
+      () => this.drawerService.pop(),
+      () => this.drawerService.isTop(ref),
+      (c, cfg) => this.stackDrawer(c, cfg),
```
BryanCunningham: Nit: Can we use more descriptive arg names here? I'm assuming this is component and config but, might be easier to understand at a glance with more explicit names

BryanCunningham (dismissed):
Couple pretty minor nits but, looks good overall!

BryanCunningham (approved):

harr1424 (approved):

shane-melton (approved):

PR #19805, opened 2026-03-27
Api service for receive

Hinton:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-34140
## 📔 Objective

Scaffolded ApiService for Receives.

### DIFF
--- a/libs/common/src/tools/receive/models/data/receive.data.ts
+++ b/libs/common/src/tools/receive/models/data/receive.data.ts
@@ -5,7 +5,7 @@ import { ReceiveFileData } from "./receive-file.data";
 export class ReceiveData {
   id: string;
   name: string;
-  file?: ReceiveFileData;
+  files: ReceiveFileData[] = [];
   userKeyWrappedSharedContentEncryptionKey: string;
   userKeyWrappedPrivateKey: string;
   scekWrappedPublicKey: string;
@@ -28,8 +28,6 @@ export class ReceiveData {
     this.revisionDate = response.revisionDate;
     this.expirationDate = response.expirationDate;
 
-    if (response.file != null) {
-      this.file = new ReceiveFileData(response.file);
-    }
+    this.files = response.files.map((f) => new ReceiveFileData(f));
   }
 }
--- a/libs/common/src/tools/receive/models/domain/receive.ts
+++ b/libs/common/src/tools/receive/models/domain/receive.ts
@@ -15,7 +15,7 @@ export class Receive {
   userKeyWrappedSharedContentEncryptionKey: EncString;
   userKeyWrappedPrivateKey: EncString;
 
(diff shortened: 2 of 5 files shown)


libs/common/src/tools/receive/models/domain/receive.ts
```diff
@@ -0,0 +1,39 @@
...
+  id: ReceiveId;
+
+  // Encrypted shared content, encrypted by the sharedContentEncryptionKey (SCEK).
+  name: EncString;
+  scekWrappedPublicKey: EncString;
+
+  userKeyWrappedSharedContentEncryptionKey: EncString;
+  userKeyWrappedPrivateKey: EncString;
+
+  file?: ReceiveFile[];
```
Thomas-Avery:
```suggestion
  files?: ReceiveFile[];
```
⛏️

libs/common/src/tools/receive/models/data/receive.data.ts
```diff
@@ -0,0 +1,35 @@
+import { ReceiveResponse } from "../response/receive.response";
+
+import { ReceiveFileData } from "./receive-file.data";
+
+export class ReceiveData {
+  id: string;
+  name: string;
+  file?: ReceiveFileData;
```
harr1424:
```suggestion
  files?: ReceiveFileData[];
```

libs/common/src/tools/receive/models/response/receive.response.ts
```diff
@@ -0,0 +1,36 @@
...
+    );
+    this.userKeyWrappedPrivateKey = this.getResponseProperty("UserKeyWrappedPrivateKey");
+    this.scekWrappedPublicKey = this.getResponseProperty("ScekWrappedPublicKey");
+    this.secret = this.getResponseProperty("Secret");
+    this.uploadCount = this.getResponseProperty("UploadCount");
+    this.creationDate = this.getResponseProperty("CreationDate");
+    this.revisionDate = this.getResponseProperty("RevisionDate");
+    this.expirationDate = this.getResponseProperty("ExpirationDate");
+
+    const file = this.getResponseProperty("File");
```
harr1424:
```suggestion
    const files = this.getResponseProperty("Files");
```

❓ I think we will want to update the response model to extract the `Files` property similar to how `Name` is being handled in `Send` and `Receive` currently

libs/common/src/tools/receive/models/response/receive.response.ts
```diff
@@ -0,0 +1,36 @@
...
+    this.file = file != null ? new ReceiveFileApi(file) : undefined;
```
harr1424: We likely want to update this to support an enumerable type of Files, the server can send an empty enumerable as a default, but this isn't currently implemented

PR #19808, opened 2026-03-27
More Simple IPC Test

coltonhurst:
A more simple (maybe, maybe not) version of ipc based on @ike-kottlowski's PR here: https://github.com/bitwarden/clients/pull/19774

Shout out to @ike-kottlowski and @neuronull on the pairing session yesterday!

### DIFF
--- a/apps/desktop/src/autofill/models/magnify-commands.ts
+++ b/apps/desktop/src/autofill/models/magnify-commands.ts
@@ -35,5 +35,5 @@ export type MagnifyCommandResponse =
 export type MagnifyLoginItem = {
   id: string;
   name: string;
-  email: string;
+  username: string;
 };
--- a/apps/desktop/src/autofill/services/desktop-magnify.service.ts
+++ b/apps/desktop/src/autofill/services/desktop-magnify.service.ts
@@ -87,9 +87,9 @@ export class DesktopMagnifyService implements OnDestroy {
     const response: MagnifyCommandResponse = {
       type: MagnifyCommand.SearchVault,
       results: [
-        { id: "a1b2c3", name: "Netflix", email: "user@gmail.com" },
-        { id: "d4e5f6", name: "Netflix Family", email: "family@gmail.com" },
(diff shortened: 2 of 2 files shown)


apps/desktop/src/magnify/app/app.component.ts
```diff
@@ -1,13 +1,16 @@
...
 })
-export class AppComponent {
-  constructor() {
-    // eslint-disable-next-line no-console
-    console.log("chrome version" + window.ipc.chrome());
+export class AppComponent implements OnInit {
+  constructor(private readonly commandService: CommandService) {}
+
+  async ngOnInit(): Promise<void> {
+    await this.commandService.searchVault("Netfli");
```
neuronull, line 14: Is this just temporary because we don't have everything hooked up yet?

apps/desktop/src/autofill/models/magnify-commands.ts
```diff
@@ -0,0 +1,39 @@
...
+  | { type: typeof MagnifyCommand.SearchVault; results: MagnifyLoginItem[] }
+  | { type: typeof MagnifyCommand.CopyPassword; result: string };
+
+/*
+  Magnify Item: Login
+*/
+export type MagnifyLoginItem = {
+  id: string;
+  name: string;
+  email: string;
```
neuronull: should `email` be `username` ?

apps/desktop/src/magnify/app/app.component.ts
coltonhurst: Yes, just here as example usage for the team working on the UI side!

apps/desktop/src/autofill/models/magnify-commands.ts
coltonhurst: Ah yes, good call. Email should be username!

id = the item id (in this case, cipher id)
name = the item name (in this case, cipher name field)
email should be username, which is the cipher username field

Updated in 979815afa1259e3f7a754e48954b0b277b9df336

PR #19810, opened 2026-03-27
[PM-33524] Not able to set new Master Password in a previously TDE org

enmande:
## 🎟️ Tracking

[PM-33524](https://bitwarden.atlassian.net/browse/PM-33524)

## 📔 Objective

Add edge-case handling for supplying salt for a user who is transitioning from a master password-less state (TDE) to one where a Master Password is required (e.g., TDE offboarding).

Added tests to describe these scenarios.

🔍  Comments have been added across `emailToSalt` callsites as part of this update to enable task tracking as the greater epic works forward, hence most of the cross-team review requests.

[PM-33524]: https://bitwarden.atlassian.net/browse/PM-33524?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## 📸  Screenshots

Charlie is being offboarded from TDE; organization has changed from TDE to Master Password encryption. Is able to successfully set a master password and access vault.

https://github.com/user-attachments/assets/7088b4ad-4321-4b7a-a6a7-bdbd88112a3b


JaredSnider-Bitwarden (dismissed):
LGTM!

r-tome (dismissed):

libs/common/src/key-management/master-password/services/master-password.service.ts
```diff
@@ -118,11 +118,27 @@ export class MasterPasswordService implements InternalMasterPasswordServiceAbstr
...
-              return unlockData.salt;
+              // No unlock data. Determine whether this is a hydration failure
+              // or a user who legitimately has no master password yet
+              // (e.g., TDE offboarding).
+              return from(this.userHasMasterPassword(userId)).pipe(
+                switchMap((hasMp) => {
+                  if (hasMp) {
+                    throw new Error("Master password unlock data not found for user.");
+                  }
+                  // TODO: PM-32059 — Edge case: user does not have a master password (e.g., TDE offboarding).
```
quexten, line 122: Just noting, the comment here may be misleading. If we generate new salts after removing email support, then we should not do that transparently as a side-effect of saltForUser$, but instead just have a "generateSalt" function, that the tde offboarding *deliberately* calls.

quexten (dismissed):

libs/common/src/key-management/master-password/services/master-password.service.ts
enmande: Thank you for this callout. It's super important to be precise anyway, but in particular here -- the salt work is long-lived and complex; don't want this to be misinterpreted when we return to it in next stages. We got hit by a merge conflict, so I took the opportunity to update this comment in 3188c50.

JaredSnider-Bitwarden (approved):

quexten (approved):

r-tome (approved):

PR #19818, opened 2026-03-27
Magnify/frontend UI merge to main

JaredScar:
## 🎟️ Tracking

Magnify

## 📔 Objective

Frontend changes for magnify innovation sprint project

### DIFF
--- a/.github/workflows/build-web-target.yml
+++ b/.github/workflows/build-web-target.yml
@@ -41,4 +41,5 @@ jobs:
       pull-requests: write
       id-token: write
       security-events: write
+      packages: write
 
--- a/.github/workflows/sdk-update.yml
+++ b/.github/workflows/sdk-update.yml
@@ -196,6 +196,11 @@ jobs:
 
           echo "✅ No stop-updates label found. Proceeding."
 
+      - name: Configure git identity
+        run: |
+          git config user.name "$_BOT_NAME"
+          git config user.email "$_BOT_EMAIL"
+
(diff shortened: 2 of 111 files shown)


BryanCunningham:
@JaredScar Is there a design or screenshots for this? We're in the process of updating to a new color palette. The new palette is already in main and able to be used. Was just curious if that should/could be applied here. I noticed all the colors used were from the legacy palette, was just wondering if that is what the design called for or if this was being built without much design input.

cd-bitwarden:
> @JaredScar Is there a design or screenshots for this? We're in the process of updating to a new color palette. The new palette is already in main and able to be used. Was just curious if that should/could be applied here. I noticed all the colors used were from the legacy palette, was just wondering if that is what the design called for or if this was being built without much design input.

@BryanCunningham 

Is there a link, or information somewhere on the new palate colors?

vleague2:
> > @JaredScar Is there a design or screenshots for this? We're in the process of updating to a new color palette. The new palette is already in main and able to be used. Was just curious if that should/could be applied here. I noticed all the colors used were from the legacy palette, was just wondering if that is what the design called for or if this was being built without much design input.
> 
> @BryanCunningham
> 
> Is there a link, or information somewhere on the new palate colors?

 @cd-bitwarden 

We have it documented here currently: https://components.bitwarden.com/?path=/docs/documentation-colors--docs I don't think we've made any eng-wide announcements about it due to the update still being in-progress, but designers are working with it for future-looking designs. I know this is an innovation sprint project though

PR #19829, opened 2026-03-27
[PM-19168] Remove Archive Feature Flag

nick-livefront:
## 🎟️ Tracking

[PM-19168](https://bitwarden.atlassian.net/browse/PM-19168)

## 📔 Objective

Removes the Archive feature flag and refactor all implementation points to use the "true" path

## 📸 Screenshots

N/A

[PM-19168]: https://bitwarden.atlassian.net/browse/PM-19168?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/web/src/app/vault/components/vault-items/vault-cipher-row.component.spec.ts
+++ b/apps/web/src/app/vault/components/vault-items/vault-cipher-row.component.spec.ts
@@ -5,12 +5,15 @@ import { RouterModule } from "@angular/router";
 import { mock } from "jest-mock-extended";
 import { BehaviorSubject } from "rxjs";
 
+import { PremiumBadgeComponent } from "@bitwarden/angular/billing/components/premium-badge";
 import { JslibModule } from "@bitwarden/angular/jslib.module";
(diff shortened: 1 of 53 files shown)


apps/web/src/app/vault/components/vault-items/vault-cipher-row.component.spec.ts
jengstrom-bw, line 1: ❓ Are the additions of the premium components and services part of this feature flag removal?

nick-livefront: Good question, I should have called this out. I removed `archiveEnabled` which was default to `false` in the test. With the removal it was now treated as `true` which had downstream impacts of the `app-premium-badge` now being present in the test runs.

These imports were needed for that to render correctly

jengstrom-bw (approved):
Thanks for clarifying that question. Looks good!

PR #19859, opened 2026-03-30
[PM-34380] Remove SASS from apps/web

Hinton:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-34380
## 📔 Objective

Removes SASS from `apps/web` build system. The icons and multi select SASS files are compiled into CSS that are then included.


djsmith85:
FYI @Hinton: https://github.com/bitwarden/clients/pull/19865 removes the usage of the `mfaTypes` all together and we can then remove old-mfa.scss

libs/angular/src/scss/bwicons/styles/style.css
```diff
@@ -0,0 +1,709 @@
+/**
+ * This is the source file, use it to generate `style.css`.
```
vleague2: The comments being copied to the css files are confusing 😩

libs/components/src/multi-select/scss/bw.theme.scss
```diff
@@ -1,3 +1,9 @@
+/**
+ * This is the source file, use it to generate `bw.theme.css`.
```
vleague2: Ok so these have to be run manually to generate the css files?

Hinton: Correct. We seldom if ever change the files so some slight pain for that process is worth streamlining it for everyone.

libs/angular/src/scss/bwicons/styles/style.css
Hinton: Yes, i rewrote the comments and I think they work better in both files now.

apps/web/src/css/tailwind.css
```diff
@@ -79,6 +81,23 @@
...
+      width: 100%;
+    }
+  }
+
+  @import "./braintree.css";
+  @import "./vault-filters.css";
+
+  @import "../../../../libs/angular/src/scss/bwicons/styles/style.css";
+  @import "../../../../libs/angular/src/scss/icons.css";
+  @import "../../../../libs/components/src/multi-select/scss/bw.theme.css";
```
vleague2: Shouldn't these imports be at the top of the file?

vleague2 (approved):

willmartian (approved):

vleague2 (approved):

PR #19864, opened 2026-03-30
[CL-1124] updated badge max width

BryanCunningham:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/CL-1124

## 📔 Objective

- Slightly increase default badge max width
- Move max width class to content container so that icon render does not effect how many characters are seen


libs/components/src/badge/badge.component.ts
```diff
@@ -130,8 +130,11 @@ export class BadgeComponent {
   /**
    * Tailwind max-width class to apply to constrain badge width.
```
vleague2: Can we update this to say it will constrain the width of the badge text?

BryanCunningham: Sure!

BryanCunningham: Updated [here](https://github.com/bitwarden/clients/pull/19864/changes/bfd4905f0cab5f60c3460bfcf3e6aac4c752dd73)

vleague2 (approved):

libs/pricing/src/components/cart-summary/cart-summary.component.html
```diff
@@ -17,7 +17,7 @@
         </h2>
         <span bitTypography="h3">&nbsp;</span>
         @if (!hideTerm) {
-          <span bitTypography="body1" class="tw-text-muted tw-ms-1"> / {{ term | i18n }} </span>
+          <span bitTypography="body1" class="tw-text-muted"> / {{ term | i18n }} </span>
```
BryanCunningham, line 20: @bitwarden/team-billing-dev removed this margin as it was not rendering/applying correctly previously. Now that it is, the spacing is too large and is no longer needed as the `&nbsp;` above provides the correct spacing

amorask-bitwarden (approved):

cd-bitwarden (approved):
LGTM!

PR #19874, opened 2026-03-30
[PM-33928] Fix: Can view MyItems Passwords in Org Vault Health Reports

Banrion:
## 🎟️ Tracking

[[PM-33928] Can view MyItems Passwords in Org Vault Health Reports](https://bitwarden.atlassian.net/browse/PM-33928)

## 📔 Objective

Org vault health reports include items from users' **My Items** collections, the personal `DefaultUserCollection` created by the _Enforce organization data ownership_ policy. The `canManageCipher` method in each org report component was not accounting for these collections, making those rows clickable and editable by admins even though admins cannot manage items in another member's personal collection.

The fix loads the org's shared (non-default) collections via `CollectionService.decryptedCollections$` on init and builds a `sharedCollectionIds` set. `canManageCipher` now returns false for any cipher whose collectionIds have no overlap with that set meaning the cipher exists only in a member's **My Items** collection. Those report rows now render as non-clickable text.

The `manageableCiphers` array lookup was also converted to a set for more performant per-row lookups.

### Note 

These changes are scoped to fix the bug. The `manageableCiphers` improvement was a small enough change for a performance boost that I chose to include. Other improvements to this area of code are out of scope for this ticket, but comments on follow up improvements are welcome.

## 📸 Screenshots

After changes, "My Item in Test Org" row cannot be clicked:

image

[PM-33928]: https://bitwarden.atlassian.net/browse/PM-33928?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/web/src/app/dirt/reports/pages/organizations/exposed-passwords-report.component.ts
+++ b/apps/web/src/app/dirt/reports/pages/organizations/exposed-passwords-report.component.ts
@@ -2,6 +2,7 @@ import { Component, OnInit } from "@angular/core";
 import { ActivatedRoute } from "@angular/router";
 import { firstValueFrom, takeUntil, tap } from "rxjs";
 
+import { CollectionService } from "@bitwarden/admin-console/common";
 import { AuditService } from "@bitwarden/common/abstractions/audit.service";
 import { OrganizationService } from "@bitwarden/common/admin-console/abstractions/organization/organization.service.abstraction";
 import { AccountService } from "@bitwarden/common/auth/abstractions/account.service";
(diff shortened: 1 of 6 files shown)


apps/web/src/app/dirt/reports/pages/organizations/exposed-passwords-report.component.ts
```diff
@@ -54,7 +55,8 @@ export class ExposedPasswordsReportComponent
   extends BaseExposedPasswordsReportComponent
   implements OnInit
 {
-  private manageableCiphers: Cipher[] = [];
+  private manageableCipherIds = new Set<string>();
+  private sharedCollectionIds = new Set<string>();
```
AlexRubik, line 59: The collection-loading setup and canManageCipher logic is identical across all 5 org report components (~25 lines each). Would it make sense to extract this into a shared base class or utility as a follow-up?

AlexRubik (approved):
Approved + follow up comment

apps/web/src/app/dirt/reports/pages/organizations/exposed-passwords-report.component.ts
Banrion: 👍 There are a number of improvements that could be made on the older reports that we adopted. The plan is to review that area of our product and analyze for overall improvements when we have capacity after our team has had our standards meetings.

PR #19880, opened 2026-03-31
[no ticket] Improve matching of dev extension

quexten:
## 🎟️ Tracking

[no ticket]

## 📔 Objective

Sometimes the desktop app does not pick up the dev extension. In my case on linux/chromium, the matching code does not work and I have to manually override on every startup. This PR fixes it by adding an additional matcher for an alternative way to grab the extension ID.


dani-garcia (approved):
Sounds reasonable, lgtm!

PR #19894, opened 2026-03-31
[PM-15489] 2fa account recovery

kspearrin:
# 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-15489

# 📔 Objective

Allow account recovery to selectively recover password and/or 2FA on org members.

# Screenshots

image

image


kspearrin:
This is the Claude plan for the frontend work on the 2fa recovery feature that was implemented here: https://github.com/bitwarden/server/pull/7139

apps/web/claude-plan.md
```diff
@@ -0,0 +1,178 @@
...
+New signature:
+
+```ts
+async recoverAccount(
+  organizationUserId: string,
+  organizationId: OrganizationId,
+  resetMasterPassword: boolean,
+  resetTwoFactor: boolean,
+  newMasterPassword?: string,
+  email?: string,
```
BTreston: This is alot of params, a single object would be cleaner

BTreston: Also in general, this function is pretty large and does a lot of stuff. While we are here, It would be nice to have Claude break this out into smaller private methods along the following lines:

- `recoverAccount`: Public orchestrator; validates flags, routes work, calls API
- `buildMasterPasswordRequest`: Picks legacy (default) vs. V2 (flagged) crypto path
- `buildResetPasswordRequestV2`: New-API master password request construction
- `decryptUserKey`: Two-step RSA decryption to recover the user key
- `buildKdfConfig`: KDF config construction from API response 
- `resetMasterPassword`: Thin backward-compat wrapper around `recoverAccount` to avoid updating all callers.

I think this would better serve the implementation for this feature

apps/web/claude-plan.md
BTreston, line 1: Can we add test coverage considerations to the plan?

BTreston (changes requested):
Plan looks good, just some minor feedback.

BTreston (dismissed):
Plan looks good to me 👍

apps/web/src/locales/en/messages.json
```diff
@@ -9694,6 +9694,30 @@
   "recoverAccount": {
     "message": "Recover account"
   },
+  "recoverAccountWarning": {
+    "message": "When you reset the password or two-step login for this member, they'll be notified of the change and logged out."
+  },
+  "accountRecoveryPolicyDescV2": {
```
kspearrin, line 9725: Should I delete the old strings?

apps/web/src/app/admin-console/organizations/members/components/account-recovery/account-recovery-dialog.component.ts
```diff
@@ -88,43 +95,70 @@ export class AccountRecoveryDialogComponent {
     switchMap((userId) => this.policyService.masterPasswordPolicyOptions$(userId)),
   );
 
+  /** True when the org has the Require Two-Step Login policy enabled. */
+  twoFactorPolicyEnabled$ = this.accountService.activeAccount$.pipe(
+    getUserId,
+    switchMap((userId) => this.policyService.policies$(userId)),
```
BTreston: we should use `policyAppliesToUser$` or `policiesByType$` here instead of `policies$`

apps/web/src/app/admin-console/organizations/members/services/organization-user-reset-password/organization-user-reset-password.service.ts
```diff
@@ -94,127 +105,83 @@ export class OrganizationUserResetPasswordService implements UserKeyRotationKeyR
...
+      } else {
+        const passwordRequest = await this.buildMasterPasswordRequest(
+          request.newMasterPassword,
+          request.email,
+          kdfConfig,
+          existingUserKey,
+        );
+        apiRequest.newMasterPasswordHash = passwordRequest.newMasterPasswordHash;
+        apiRequest.key = passwordRequest.key;
+      }
```
BTreston: A ternary statement would be less noisy here

apps/web/src/locales/en/messages.json
BTreston: I think we will need to keep them alongside a v2 of the components that show this new UI so it can be properly feature flagged. They can be removed after feature flag cleanup. The `vnext-organization-data-ownership-component` is a good example to feed Claude for the feature flagging pattern for policy UI. (also potentially a good idea to build a skill from)

apps/web/src/app/admin-console/organizations/members/components/account-recovery/account-recovery-dialog.component.html
BTreston, line 1: favor `@if` rather than `*ngIf`

apps/web/src/app/admin-console/organizations/members/services/organization-user-reset-password/organization-user-reset-password.service.ts
```diff
@@ -94,127 +105,83 @@ export class OrganizationUserResetPasswordService implements UserKeyRotationKeyR
...
-      // Create authentication and unlock data
-      const authenticationData =
-        await this.masterPasswordService.makeMasterPasswordAuthenticationData(
-          newMasterPassword,
-          kdfConfig,
-          salt,
+  async recoverAccount(request: RecoverAccountRequest): Promise<void> {
+    const apiRequest = new OrganizationUserResetPasswordRequest();
+    apiRequest.resetMasterPassword = request.resetMasterPassword;
+    apiRequest.resetTwoFactor = request.resetTwoFactor;
```
BTreston: We should stick with the `newConstructor` function and build the request once we have all our data initialized rather than property bagging. 

Optionally, you could also refactor to promote `newConstructor` to be the **actual** `constructor`, but that may be out of scope. If Claude comes up with an easy path forward for it though, I'd say its worth a go.

BTreston (changes requested):
Looking good so far, thank you for breaking up the `resetMasterPassword` function.

apps/web/src/locales/en/messages.json
kspearrin: Working on the feature flagged UI now

apps/web/src/app/admin-console/organizations/members/components/account-recovery/account-recovery-dialog.component.ts
kspearrin: I tried this, but it appears that those functions use the current user to determine if the policy applies are not. This causes problems when the current user is an admin, for which the policy does not apply, but it would still apply to the user you are recovering the account of.

BTreston: We can continue to use `policies$` in that case. I'll bring this up to the team as part of some ongoing work to policies, this use case has come up in the past a few times.

apps/web/src/app/admin-console/organizations/members/components/account-recovery/account-recovery-dialog-v2.component.ts
BTreston, line 1: Can we address the `FIXME`s in this file?

apps/web/src/app/admin-console/organizations/policies/policy-edit-definitions/reset-password.component.html
BTreston, line 1: We also need a v2 of the policy definition components or the feature flag check in the template here, otherwise this copy will be visible when the ff is turned off.

BTreston (changes requested):

BTreston (dismissed):

BTreston (dismissed):

kspearrin:
resolved conflicts and fixed CI errors

BTreston (approved):

apps/web/claude-plan.md
rr-bw: @kspearrin Looks like this plan `.md` file was unintentionally merged to `main`?

kspearrin: Yes, it can be deleted.

PR #19896, opened 2026-03-31
[PM-34225] generate keypair even if there is no current one on password login

jlf0dev:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-34225
## 📔 Objective

The original keypair regeneration service was intended to fix corrupted keypairs, not generate a keypair for accounts missing it. This means it skipped accounts that didn't have a keypair yet, as there was code in the login strategies that generated it for new accounts. 

We have since removed the code to generate the keypair for those missing it, leaving a small number of accounts without the ability to generate them. This alters the keypair regeneration service to create a new keypair if its missing.

Note that this will only work for SSO users, as password logins are blocked.

We still don't have intentions to enable this flag for every user, and this will remain targeted functionality.

### DIFF
--- a/libs/key-management/src/user-asymmetric-key-regeneration/services/default-user-asymmetric-key-regeneration.service.ts
+++ b/libs/key-management/src/user-asymmetric-key-regeneration/services/default-user-asymmetric-key-regeneration.service.ts
@@ -1,4 +1,4 @@
-import { combineLatest, firstValueFrom, map } from "rxjs";
+import { firstValueFrom, map } from "rxjs";
 
 import { ApiService } from "@bitwarden/common/abstractions/api.service";
 import { FeatureFlag } from "@bitwarden/common/enums/feature-flag.enum";
@@ -42,9 +42,8 @@ export class DefaultUserAsymmetricKeysRegenerationService implements UserAsymmet
       }
     } catch (error) {
       this.logService.error(
-        "[UserAsymmetricKeyRegeneration] An error occurred: " +
-          error +
-          " Skipping regeneration for the user.",
+        "[UserAsymmetricKeyRegeneration] An error occurred. Skipping regeneration for the user.",
+        error,
       );
     }
   }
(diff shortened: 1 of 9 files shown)


libs/key-management/src/user-asymmetric-key-regeneration/services/default-user-asymmetric-key-regeneration.service.ts
```diff
@@ -68,13 +67,28 @@ export class DefaultUserAsymmetricKeysRegenerationService implements UserAsymmet
...
     );
 
+    let publicKeyResponse = null;
+    try {
+      publicKeyResponse = await this.apiService.getUserPublicKey(userId);
+    } catch (e) {
+      if ((e as any)?.statusCode !== 404) {
+        throw e;
+      }
+    }
```
jlf0dev, line 81: If the public key wasn't able to be retrieved, it would previously throw. This change catches the 404 instead and allows the code to continue.

libs/key-management/src/user-asymmetric-key-regeneration/services/default-user-asymmetric-key-regeneration.service.ts
```diff
@@ -42,9 +42,8 @@ export class DefaultUserAsymmetricKeysRegenerationService implements UserAsymmet
       }
     } catch (error) {
       this.logService.error(
-        "[UserAsymmetricKeyRegeneration] An error occurred: " +
-          error +
-          " Skipping regeneration for the user.",
+        "[UserAsymmetricKeyRegeneration] An error occurred. Skipping regeneration for the user.",
+        error,
```
jlf0dev, line 46: The inline error here would previously just show [Object object] instead of the error. This will allow the error to show in the console instead.

Thomas-Avery (approved):
LGTM, we should make a follow up ticket for updating the data recovery tool.

PR #19897, opened 2026-03-31
[PM-29455] safari unlock during autofill does not close

audreyality:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-29455

## 📔 Objective

Automatically close the popup window after using an autofill UX trigger to unlock in Safari.

## 🎥 Demo

| Before | After |
|--------|-------|
| <video src="https://github.com/user-attachments/assets/6dbd4acb-12b1-43d0-ab59-a5b912b12502" width="240" /> | <video src="https://github.com/user-attachments/assets/cb304eef-9d9e-44f3-a416-6383a9ae0526" width="240" /> |


apps/browser/src/autofill/background/notification.background.ts
```diff
@@ -1913,8 +1913,9 @@ export default class NotificationBackground {
     }
 
     const extensionUrl = BrowserApi.getRuntimeURL("popup/index.html");
-    const unlockPopoutTabs = (await BrowserApi.tabsQuery({ url: `${extensionUrl}*` })).filter(
-      (tab) => tab.url?.includes(`singleActionPopout=${AuthPopoutType.unlockExtension}`),
+    const allTabs = await BrowserApi.tabsQuery({ url: `${extensionUrl}*` });
+    const unlockPopoutTabs = allTabs.filter((tab) =>
+      tab.url?.includes(`singleActionPopout=${AuthPopoutType.unlockExtension}`),
     );
```
audreyality: 📝 This change isn't strictly necessary; it was added to facilitate debugging.
🎨 Renamed `allTabs` to `extensionTabs` in the latest commit.

bensbits91 (dismissed):
Looks great Audrey 😄 

One thing that comes to mind, but I'm not sure how valuable, so take it or leave it:

Now that we're expecting `openUnlockPopout` to execute before `addToLockedVaultPendingNotifications`, should we update the test to assert that expectation?

Currently overlay.background.spec.ts:2895 only has `expect(openUnlockPopoutSpy).toHaveBeenCalled();`

It might be worth asserting the call order. I'm no Jest expert, but something like this?

`expect(openUnlockPopoutSpy.mock.invocationCallOrder[0])
  .toBeLessThan(tabSendMessageDataSpy.mock.invocationCallOrder[0]);`

Would this help guard against future regression? Again: take it or leave it.

audreyality:
> Would this help guard against future regression? Again: take it or leave it.

I have mixed feelings about this. I'm really hesitant to lock down something like function execution order, especially across 4 different callsites. On the other hand, you're right. It would protect against regression. 🤔

bensbits91:
> > Would this help guard against future regression? Again: take it or leave it.
> 
> I have mixed feelings about this. I'm really hesitant to lock down something like function execution order, especially across 4 different callsites. On the other hand, you're right. It would protect against regression. 🤔

I do NOT have strong feelings and I understand your hesitation. Feel free to disregard.

apps/browser/src/auth/popup/utils/auth-popout-window.ts
```diff
@@ -35,6 +43,15 @@ async function openUnlockPopout(senderTab: chrome.tabs.Tab) {
...
   });
+
+  if (pendingNotification) {
+    await BrowserApi.tabSendMessageData(
+      senderTab,
+      "addToLockedVaultPendingNotifications",
+      pendingNotification,
+    );
+  }
+
```
audreyality, line 54: @bensbits91 - I decided to address [your feedback](https://github.com/bitwarden/clients/pull/19897#pullrequestreview-4045997262) by inlining the notification into `openUnlockPopout`. With that done, there's only 1 test on the ordering. 😄

enmande (dismissed):
Auth changes LGTM; thank you for the added test and clear, helpful comments ✨

apps/browser/src/auth/popup/utils/auth-popout-window.ts
bensbits91: Nice 😁 Looks good

bensbits91 (approved):

enmande (approved):

PR #19898, opened 2026-03-31
[SM-1927] Updating code so that if the machine acct name is corrupted data

cd-bitwarden:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/SM-1927

## 📔 Objective

Updating code so that if the machine acct name is corrupted data it doesn't lock users out of the list views


bitwarden_license/bit-web/src/app/secrets-manager/service-accounts/service-account.service.ts
```diff
@@ -203,4 +218,16 @@ export class ServiceAccountService {
...
+
+  private async decryptField(
+    encString: EncString,
+    organizationKey: SymmetricCryptoKey,
+  ): Promise<{ value: string; error: boolean }> {
+    try {
+      const decrypted = await this.encryptService.decryptString(encString, organizationKey);
+      return { value: decrypted, error: false };
+    } catch {
+      return { value: DECRYPT_ERROR, error: true };
```
justindbaur, line 233: We probably want to log the error here so we aren't suppressing possibly useful information that could be used to diagnose the core problem.

justindbaur (changes requested):
One small thing, other than that looks good to go.

justindbaur (approved):

justindbaur (approved):

PR #19905, opened 2026-04-01
[Shared Unlock] [PM-34510] Remove desktop-side browser ipc fingerprint

quexten:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-34510

## 📔 Objective

We are replacing browser-initiated biometrics with shared unlock + biometrics-over-sdk-ipc. This will allow us to drop all legacy typescript code around biometrics IPC.

As part of this, we are no longer supporting IPC fingerprints. There is no security goal actually achieved by these, for a security definition we have defined / that is reasonable, and it increases user friction for users who enable the setting (without understanding what it is for).

This PR removes the desktop side only. After a compatibility period, the browser side can be removed too. This is how the extension / desktop app behave with different version combinations:

Old extension + New Desktop
- No fingerprint shown, since the desktop never asks for the fingerprint. Only the desktop enforces fingerprints by initiating the fingerprinting process.
New extension + Old Desktop
- Fingerprint shown. The new extension still supports showing fingerprints, until the backward compatibility period is over. 

### Security Impact
Please consider the security definition here: https://github.com/bitwarden/sdk-internal/blob/969f43fcaa86cca7e6d6e751e12ad56c5d560041/crates/bitwarden-shared-unlock/src/lib.rs#L93-L106 and https://github.com/bitwarden/sdk-internal/blob/26ab7e59b7c4cde815e529af7d1979a2a22f2684/crates/bitwarden-ipc/src/crypto_provider/noise/mod.rs#L51-L66. Under these security definitions, browser ipc fingerprinting serves no security goal, and it can be removed.


djsmith85 (changes requested):
As components have been removed or updated, please also remove any keys from the `en/messages.json` that aren't needed anymore.

The rest of the changes look good.

quexten:
Thanks @djsmith85. Removed the now unused i18n keys.

djsmith85 (approved):

eligrubb (approved):
LGTM!

PR #19907, opened 2026-04-01
[Shared Unlock] [PM-34508] Default enable native messaging permission on browser

quexten:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-34508

## 📔 Objective

Native messaging was optional so far. This increases user friction and adoption of biometric unlock (and soon shared unlock), and increases risk of bugs while providing no benefit. 

This PR moves the permission into the mandatory, required permissions, and removes support code around handling what happens when native messaging is not a permission that was granted.

**_The goal from the UX perspective is that you just install both apps, and shared unlock works._**


quexten:
Converting to draft while discussing with product.

apps/browser/src/auth/popup/settings/account-security.component.ts
```diff
@@ -210,12 +209,6 @@ export class AccountSecurityComponent implements OnInit, OnDestroy {
           }
 
           // Biometrics status shouldn't be checked if permissions are needed.
```
JaredSnider-Bitwarden: This comment is now orphaned and should be deleted?

JaredSnider-Bitwarden (commented):
Everything looks great but the orphaned comment here 👇

apps/browser/src/auth/popup/settings/account-security.component.ts
quexten: Removed the comment!

JaredSnider-Bitwarden (approved):

mzieniukbw (approved):

PR #16921, opened 2025-10-17
[BEEEP] [POC] Desktop <-> Browser IPC Framework channel

coroiu:
## 📔 Objective

I decided to play around some more with the IPC framework. I was able to implement Desktop Renderer <-> Desktop Main <-> Browser Background IPC channels. I also played around with forwarding so in this implementation Desktop Main acts as a relay between browser and renderer so that the renderer and extension can talk directly. All of it works by piggybacking on the awesome IPC proxy work implemented by @Dani (biometrics needs to be enabled in both clients because that’s what enables IPC)

## ⏰ Reminders before review

- Contributor guidelines followed
- All formatters and local linters executed and passed
- Written new unit and / or integration tests where applicable
- Protected functional changes with optionality (feature flags)
- Used internationalization (i18n) for all UI strings
- CI builds passed
- Communicated to DevOps any deployment requirements
- Updated any necessary documentation (Confluence, contributing docs) or informed the documentation team

## 🦮 Reviewer guidelines

- 👍 (`:+1:`) or similar for great changes
- 📝 (`:memo:`) or ℹ️ (`:information_source:`) for notes or general info
- ❓ (`:question:`) for questions
- 🤔 (`:thinking:`) or 💭 (`:thought_balloon:`) for more open inquiry that's not quite a confirmed issue and could potentially benefit from discussion
- 🎨 (`:art:`) for suggestions / improvements
- ❌ (`:x:`) or ⚠️ (`:warning:`) for more significant problems or concerns needing attention
- 🌱 (`:seedling:`) or ♻️ (`:recycle:`) for future improvements or indications of technical debt
- ⛏ (`:pick:`) for minor or nitpick changes

PR #19908, opened 2026-04-01
IPC Desktop <-> Browser transport

coroiu:
## 🎟️ Tracking

Supersedes https://github.com/bitwarden/clients/pull/16921

## 📔 Objective

Implements an IPC transport that lets the browser extension and the desktop app exchange messages directly over the existing native messaging channel, and routes messages between the desktop's Electron **main** and **renderer** processes. This wires the previously-stubbed IPC service into a working desktop ↔ browser pipe so SDK-based IPC clients can address `DesktopMain`, `DesktopRenderer`, and `BrowserBackground` endpoints.

### What changes

**Desktop native (Rust / NAPI)**
- Add per-client targeted messaging to the IPC server. The server now tracks a `client_id → sender` map and exposes `send_to(client_id, message)` so a message can be delivered to one specific connected client instead of broadcasting to all. Senders are registered on connect and removed on disconnect.
- Surface `sendTo` through the NAPI binding (`index.d.ts` / `ipc.rs`).

**Desktop main process**
- New `IpcMainService` replaces `NoopIpcService`. It bridges three directions:
  - Browser → desktop: subscribes to native messaging messages, delivers `DesktopMain`-bound payloads to the local IPC client and forwards `DesktopRenderer`-bound payloads to the renderer.
  - Renderer → browser/main: handles `ipc.send` from the renderer, routing to the local client or out to the correct browser client via `sendTo`.
  - Main → renderer: pushes messages to the window over `ipc.onMessage`.
- `NativeMessagingMain` now exposes a `messages$` observable and a `sendTo(clientId, message)` method for targeted replies.
- Wire `IpcMainService` into `main.ts`.

**Desktop renderer process**
- New `IpcRendererService` (replacing `NoopIpcService` in `services.module.ts`) that talks to the main process through the preload bridge.
- Add an `ipcService` bridge (`onMessage` / `send`) to `preload.ts`.

**Browser extension background**
- `IpcBackgroundService` now connects to the desktop app over native messaging, performs a discover handshake, and forwards `DesktopMain` / `DesktopRenderer`-bound messages to it. Includes automatic reconnection (10s interval) and a discover timeout (5s), plus port cleanup on failed handshake to avoid leaking the native port / `desktop_proxy` process.
- `IpcContentScriptManagerService` converts its constructor-time observable subscription into an explicit `init()` call; content-script IPC is now always enabled (still gated by the feature flag and MV3).

**Shared (`libs/common`)**
- Add `ForwardedIpcMessage` type + `isForwardedIpcMessage` guard so a message relayed through the desktop can carry its `originalSource`. Harden `isIpcMessage` against null/malformed input.
- Switch the browser background and desktop services to SDK in-memory sessions (`IpcClient.newWithSdkInMemorySessions`); `IpcSessionRepository` is no longer wired up (marked as currently unused).

## 📸 Screenshots

N/A — no UI changes.


apps/browser/src/platform/ipc/ipc-content-script-manager.service.ts
```diff
@@ -31,6 +29,8 @@ export class IpcContentScriptManagerService {
           await BrowserApi.registerContentScriptsMv3([
             {
               id: IPC_CONTENT_SCRIPT_ID,
+              // WARNING: This means that all websites can talk to the IPC layer.
+              // Before sharing unlock state, this needs to be fixed.
```
coroiu: issue: irrelevant since we've decoupled this from unlock state.

apps/browser/src/platform/ipc/ipc-background.service.ts
```diff
@@ -60,29 +113,94 @@ export class IpcBackgroundService extends IpcService {
...
+
+  private scheduleReconnect() {
+    if (this.reconnectTimer != null) {
+      return;
+    }
+    this.reconnectTimer = setTimeout(() => {
+      this.reconnectTimer = undefined;
+      void this.connectToDesktop();
+    }, RECONNECTION_INTERVAL_MS);
+  }
```
coroiu, line 211: suggestion: we should re-factor this to use a timer observable which is subscribed to in `init`

apps/browser/src/platform/ipc/ipc-content-script-manager.service.ts
```diff
@@ -14,8 +13,7 @@ export class IpcContentScriptManagerService {
       return;
     }
 
-    configService
-      .getFeatureFlag$(FeatureFlag.ContentScriptIpcChannelFramework)
+    of(true)
```
coroiu: issue: there is no point in having an observable here anymore, just call unregister and register directly. Also, since this change would trigger the BrowserApi immediately on initialization we should move this code into an `init()` function and call it from the appropriate init service

apps/desktop/desktop_native/napi/src/lib.rs
coroiu, line 24: question: why do we a have a deprecated comment here?

apps/desktop/src/main/native-messaging.main.ts
```diff
@@ -17,6 +18,16 @@ export class NativeMessagingMain {
...
 
+  private _messages$ = new Subject<ipc.IpcMessage>();
+  readonly messages$ = this._messages$.asObservable();
+
+  /**
+   * Optional callback that returns true if the listener should be kept alive
+   * even when a specific integration is being disabled. This prevents one
+   * integration from stopping the listener while another still needs it.
+   */
+  shouldKeepListening?: () => Promise<boolean>;
```
coroiu: There is a follow-up ticket to refactor this file after biometrics has been moved out of it: [PM-34514 [Follow Up] Refactor native messaging after biometrics has been migrated](https://bitwarden.atlassian.net/browse/PM-34514)

apps/desktop/src/main.ts
```diff
@@ -356,29 +371,21 @@ export class Main {
...
-          this.nativeMessagingMain
-            .listen()
-            .catch((err) =>
-              this.logService.error("Error while starting native message listener", err),
-            );
+          // Start native messaging when shared unlock is enabled at runtime
+          await this.nativeMessagingMain.generateManifests();
+          await this.nativeMessagingMain.listen();
+        } catch (err) {
+          this.logService.error("Error while setting up native messaging:", err);
```
coroiu: issue: this doesn't make sense anymore since shared-unlock is no longer part of the changeset. We should just remove any conditions and generateManifests and start listening regardless (if that doesn't cause any issues with DDG)

apps/desktop/src/main.ts
```diff
@@ -295,6 +298,18 @@ export class Main {
       app.getAppPath(),
     );
 
+    // TODO: Rewrite this to use an observable pattern instead of a callback, to avoid potential issues with multiple listeners and to better handle async operations
+    this.nativeMessagingMain.shouldKeepListening = async () => {
+      return Promise.resolve(true);
+    };
```
coroiu: suggestion: I think we can just remove this completely and always continue listening (remove the `shouldKeepListening` function entirely, I don't think it's needed)

libs/common/src/platform/ipc/ipc.service.ts
```diff
@@ -133,6 +139,12 @@ export abstract class IpcService {
    * - Implementations may override `init` but should call this helper exactly once.
    */
   protected async initWithClient(client: IpcClient): Promise<void> {
+    // TODO: Remove, this is only used for development
+    // Example: Open electron renderer devtools console and call `await discover({ BrowserBackground: { id: "Own" } })`
+    // You might need to reload the browser extension if the desktop app was started after it (or if it was restarted while the extenion was running)
+    (globalThis as any).discover = async (endpoint: Endpoint) =>
+      await ipcRequestDiscover(this.client, endpoint);
+
```
coroiu: issue: dev code that needs to be removed

coroiu (commented):
self-review

apps/web/src/app/platform/ipc/web-ipc.service.ts
```diff
@@ -70,21 +70,27 @@ export class WebIpcService extends IpcService {
...
+
+      // Ensure the browser extension is present
+      const version = await ipcRequestDiscover(
+        this.client,
+        { BrowserBackground: { id: "Own" } },
+        AbortSignal.timeout(DISCOVER_MESSAGE_TIMEOUT_MS),
+      );
+      this.logService.info(
+        `[IPC] Connected to Bitwarden Browser Extension with version ${version.version}`,
+      );
```
coroiu: issue: This was dev code and should be removed

apps/desktop/src/main.ts
```diff
@@ -356,29 +366,20 @@ export class Main {
...
-          if (browserIntegrationEnabled) {
-            this.nativeMessagingMain
-              .generateManifests()
-              .catch((err) => this.logService.error("Error while generating manifests", err));
-          }
           if (ddgIntegrationEnabled) {
-            this.nativeMessagingMain
-              .generateDdgManifests()
-              .catch((err) => this.logService.error("Error while generating DDG manifests", err));
+            await this.nativeMessagingMain.generateDdgManifests();
```
coroiu, line 416: **note**: we are now always enabling native messaging regardless of settings.

issue?: Do we need to disable DDG integration somewhere else so it doesn't just rely on native messaging not being enabled when the integration shouldn't be active

coroiu (commented):
uncommited review comments

apps/browser/src/platform/ipc/ipc-background.service.ts
coroiu: Tried this refactor — the timer-observable version didn't end up being an improvement over the current `setTimeout`-based reconnect, so leaving it as-is.

apps/desktop/src/main.ts
coroiu: Brought this up with desktop native and got the go-ahead to continue with the code as-is

> I don't think there will be a problem with the native messaging host service running while the DDG setting is off; IIRC, it is also turned on when biometrics are enabled, so that should be an expected potential case.
> I think the only potential issue is having that service always running from a security perspective. Not sure if that opens any new vectors. My gut answer is it would not.
>
> [Alright, so ok to proceed as-is then?]
>
> I think so, as long as testing for the DDG integration passes; I can help out with that before QA does full regression on it

dereknance (approved):
This looks excellent 🎉

Issue #18612, opened 2026-01-27
Invalid `ssh-add` flag causes ssh-agent to become unresponsive


Bazoogle:
### Steps To Reproduce

From Windows:
1. Open services.msc on windows
2. Go to `OpenSSH Authentication Agent` and disable the service
3. Open Bitwarden Desktop App
4. Select `File` in the toolbar
5. Select `Settings`
6. Check `Enable SSH Agent`
7. Open `cmd` on windows
8. Type `ssh-add -L` to confirm functionality
9. Type `ssh-add -?` or any other invalid flag
10. Type `ssh-add -L` again or try to `ssh` into a machine

### Expected Result

It should execute the SSH command. When type `ssh-add -L` it should list your SSH keys. When typing `ssh <server>` it should initiate an SSH connection into a server.

### Actual Result

The command is left hanging, with no output whatsoever. `ssh-agent` process seems to crash. To get it to respond, you must exit Bitwarden desktop entirely, and restart the app.

### Additional Context

The `app.log` file quickly fills up with the repeating the following lines:
``` log
[2026-01-27 15:54:02.108] [info]  [NAPI] [INFO] desktop_core::ssh_agent::platform_ssh_agent::named_pipe_listener_stream: [SSH Agent Native Module] Incoming connection
[2026-01-27 15:54:02.122] [info]  [NAPI] [INFO] desktop_core::ssh_agent::platform_ssh_agent::named_pipe_listener_stream: Waiting for connection
[2026-01-27 15:54:24.967] [info]  [NAPI] [INFO] desktop_core::ssh_agent::platform_ssh_agent::named_pipe_listener_stream: [SSH Agent Native Module] Incoming connection
[2026-01-27 15:54:24.976] [error] [NAPI] [ERROR] desktop_core::ssh_agent::platform_ssh_agent::named_pipe_listener_stream: Failed getting process info error=Failed to get process pid=10952
[2026-01-27 15:54:24.978] [info]  [NAPI] [INFO] desktop_core::ssh_agent::platform_ssh_agent::named_pipe_listener_stream: Waiting for connection
[2026-01-27 15:54:24.979] [info]  [NAPI] [INFO] desktop_core::ssh_agent::platform_ssh_agent::named_pipe_listener_stream: [SSH Agent Native Module] Incoming connection
[...]

PR #19909, opened 2026-04-01
[PM-34520] fix: prevent SSH agent infinite loop on Windows when peer info lookup fails

sjroesink:
## 🎟️ Tracking

Resolves https://github.com/bitwarden/clients/issues/18612

## 📔 Objective

Fix the SSH agent becoming permanently unresponsive on Windows after a failed connection (e.g. `ssh-add -?` or any command where the process exits before peer info can be gathered).

**Root cause:** When `get_peer_info()` or `GetNamedPipeClientProcessId()` failed, the listener loop did `continue` without creating a new named pipe. The old `listener` stayed in a "connected" state, causing `listener.connect()` to return immediately on the next iteration — creating an infinite loop that flooded the logs    and made the SSH agent unresponsive to all subsequent commands.

**Fix:** Restructure the connection handler so a new pipe is always created after each connection, regardless of whether peer info gathering succeeded. Failed connections are gracefully dropped while the listener resets to a clean state.


pamperer562580892423:
@sjroesink FYI: https://community.bitwarden.com/t/ssh-agent-whats-changing-and-what-to-expect/95308

sjroesink:
> @sjroesink FYI: https://community.bitwarden.com/t/ssh-agent-whats-changing-and-what-to-expect/95308

Got it! I'll keep using my forked version until the v2 agent is released, thanks for pointing this out

neuronull:
Thanks for reporting you're still facing the problem @sjroesink , and for looking into a fix !

I can circle back once the v2 is in, to let you know, and likely you'll be able to re-open the PR on the v2 implementation.

PR #19910, opened 2026-04-01
Update primitive colors

BryanCunningham:

### DIFF
--- a/libs/components/src/tw-theme.css
+++ b/libs/components/src/tw-theme.css
@@ -137,7 +137,7 @@
   --color-orange-300: #fdba8c;
   --color-orange-400: #ff894a;
   --color-orange-500: #fa6216;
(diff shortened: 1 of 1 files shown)


libs/components/src/tw-theme.css
```diff
@@ -137,7 +137,7 @@
   --color-orange-300: #fdba8c;
   --color-orange-400: #ff894a;
   --color-orange-500: #fa6216;
-  --color-orange-600: #cc4400;
+  --color-orange-600: #CC4400;
```
vleague2: Isn't this the same as before?

BryanCunningham: It is. Switched back to lowercase

vleague2 (approved):

PR #19915, opened 2026-04-01
Magnify: support for credit card item and actions

neuronull:
## 📔 Objective

Adds support for new item type: credit card, and three actions: copy card number, expiration, and CVV. 

In order to support it, discriminated union was added back. This is so that the code sending the requests from the front end doesn't have to send request for each item type and merge the results.

Added an "Extract" type in the IPC dispatcher for the sendCommand, on "type". This allows us compile time type safety checks that the response type matches the request type.


apps/desktop/src/autofill/models/magnify-commands.ts
```diff
@@ -24,13 +32,48 @@ export type MagnifyCommand = (typeof MagnifyCommand)[keyof typeof MagnifyCommand
 export type MagnifyCommandRequest =
   | { type: typeof MagnifyCommand.SearchVault; input: string }
   | { type: typeof MagnifyCommand.CopyPassword; id: string }
-  | { type: typeof MagnifyCommand.ViewInBitwarden; itemId: string };
+  | { type: typeof MagnifyCommand.ViewInBitwarden; itemId: string }
+  | { type: typeof MagnifyCommand.CopyCardNumber; itemId: string }
+  | { type: typeof MagnifyCommand.CopyCardExpiration; itemId: string; format?: string }
+  | { type: typeof MagnifyCommand.CopyCardCode; itemId: string };
```
coltonhurst, line 46: One other small nit - we have both `id` and `itemId` here. Don't feel pressure to change anything but we could converge all in on just `id` or `itemId` for consistency. Just mentioning, not a big deal at all.

neuronull: ah, yeah, I can address this. IIRC , "itemId" is used in other parts of the repo as a move towards that away from cipherId. So if you're alright with itemId, I'll use that.

coltonhurst: Sounds good to me 👍

libs/common/src/key-management/vault-timeout/services/vault-timeout-settings.service.ts
```diff
@@ -280,24 +293,48 @@ export class VaultTimeoutSettingsService implements VaultTimeoutSettingsServiceA
...
             currentVaultTimeoutAction,
             maxSessionTimeoutPolicyData,
           );
 
-          // As a side effect, set the new value determined by determineVaultTimeout into state if it's different from the current
-          // We want to avoid having a null timeout action always so we set it to the default if it is null
-          // and if the user becomes subject to a policy that requires a specific action, we set it to that
-          if (vaultTimeoutAction !== currentVaultTimeoutAction) {
+          // As a side effect, persist the determined action back to state when needed.
+          const oneActionAvailable = availableVaultTimeoutActions.length === 1;
```
quexten, line 312: Looks like you accidentally pulled in some changes from main and are merging them into `magnify/main` via this PR. I don't think my review is required.

https://github.com/bitwarden/clients/pull/19594/changes#diff-f3f77512e75f15bf8e74e8e486ae2e65b9f104b8aed99d1db4bc6ffe56cad1a6

quexten: I suggest pulling in main on `magnify/main` and then rebasing or merging `magnify/main` into this PR to eliminate the changes.

PR #19921, opened 2026-04-01
[CL-1130] Fix a11y and storybook errors in UIF-owned code

vleague2:
## 🎟️ Tracking

[CL-1130](https://bitwarden.atlassian.net/browse/CL-1130)

## 📔 Objective

We are working through Storybook errors related to accessibility and general console errors (usually story setup issues). This PR fixes current errors in UIF-owned code and also makes some setup/testing changes for this workflow.

Setup/testing changes:
- Add overlay container to the list of elements that the a11y plugin should check
- Exclude `.cdk-visually-hidden` because it has a11y violations that we cannot fix because we do not control the cdk implementation
- Fix `test-stories` script so that it doesn't timeout frequently, and turn on the "fail on console errors" option

Fixes:
- Prod
  - Landmarks and progress bars must have unique accessible names. I added translation keys where needed for these.
  - Scrollable regions must be reachable by tab
  - Properly set a role for icons and svgs depending on whether they are intended to be perceived by screenreader users
  - Don't set `nav` landmark in side-nav to `dialog` role, because this is not allowed. Move all dialog-related attributes to the wrapper `div`
- Story-only
  - Remove `aria-label` from elements that don't support it
  - Separate stories that were mounting multiple `main`s in one story, since this is a violation and it is more work to try to maneuver around that than to just have them be in separate stories
  - Added a story for callouts with multiple no-title variants together so that we can validate that the fallback accessible unique name approach works
  - Fix missing mocks
  - Headings must not skip levels

Billing/Auth/DIRT:
- I tried to split all these PRs by team ownership area, but naturally some of the CL changes have propagated out to code in your areas. It should be just translation additions to your stories/tests!

[CL-1130]: https://bitwarden.atlassian.net/browse/CL-1130?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


BryanCunningham (approved):

Banrion (approved):
✅ DIRT files

sbrown-livefront (approved):
🔥 Billing looks good!

rr-bw (approved):
Approving Auth file.

PR #19600, opened 2026-03-16
[PM-31941] Access Intelligence Refactor Integration

Banrion:
# Note: This pull request will be broken down into a series of focused pull requests rather than using one overwhelming pull request. 

## 🎟️ Tracking

[PM-31941 Access Intelligence Refactor Integration](https://bitwarden.atlassian.net/browse/PM-31941)

---

## 📔 Objective

This PR ships V2 of Access Intelligence (formerly Risk Insights) behind the feature flag `pm-31936-access-intelligence-new-architecture`. V1 remains fully functional when the flag is off.

### Data Models

A new Access Intelligence model family (`AccessReport`, `ApplicationHealth`, `AccessReportSettings`, `AccessReportSummary`, `MemberRegistryEntry`, `CipherHealth`).

The `AccessReport` domain model stores member data in a registry (one entry per member, referenced by ID from cipher records) instead of duplicating member data per cipher. This significantly reduces encrypted payload size for large organizations.

### Services

Seven new focused services replace the monolithic orchestrator:

- `AccessIntelligenceDataService` — coordinates report loading and state for components
- `AccessReportEncryptionService` — encrypts and decrypts report payloads
- `ReportPersistenceService` — loads and saves reports to the API
- `ReportGenerationService` — builds report data from cipher and member inputs
- `MemberCipherMappingService` — maps members to their associated ciphers
- `CipherHealthService` — evaluates cipher health (weak passwords, reuse, etc.)
- `DrawerStateService` — manages UI drawer open/close and selection state
- Versioning services — upgrade V1 encrypted payloads to V2 format on load, and downgrade back if the flag is disabled

`SecurityTasksService` is now a proper abstraction with a default V2 implementation and a legacy bridge that preserves V1 behavior. `LegacyRiskInsightsEncryptionService` supports reading V1-format data and downgrading from V2 when needed.

### V2 UI Components (behind feature flag)

- `AccessIntelligencePageComponent` — main page: tab navigation, drawer management, CSV export
- `AllActivityV2Component` — dashboard: summary metrics, critical apps, at-risk members
- `ApplicationsV2Component` — applications list with search, filter, and critical app management
- `ApplicationsTableV2Component` — reusable table presentation layer
- `AccessIntelligenceDrawerV2Component` — detail drawer for members and applications
- `NewApplicationsDialogV2Component` — review and triage new applications
- `ReviewApplicationsViewV2Component` — application selection inside the dialog
- `PasswordChangeMetricV2Component` — password change metric card
[...]

PR #19922, opened 2026-04-01
[PM-31942] Handle load/save Access Intelligence reports as files (pt. 1)

lastbestdev:
## 🎟️ Tracking
Part 1 of https://bitwarden.atlassian.net/browse/PM-31942

## 📔 Objective
- Includes model updates for making requests to the updated Access Intelligence report file APIs (changes [here](https://github.com/bitwarden/server/pull/7228))
- Adds AccessIntelligenceApiService for calls to new APIs. Includes default implementation + tests 

## 📸 Screenshots
N/A


bitwarden_license/bit-common/src/dirt/access-intelligence/models/api/access-report-create.api.ts
```diff
@@ -0,0 +1,13 @@
+import { AccessReportMetricsApi } from "./access-report-metrics.api";
+
+/**
+ * Request body for POST /reports/organizations/{organizationId}
+ *
+ */
+export class AccessReportCreateApi {
```
Banrion: ⚠️ As per the [API model documentation](https://contributing.bitwarden.com/architecture/clients/data-model/#api), api models should be reusable and shared across methods. This model defines the contract between the api service for one method. 

- Move to `access-intelligence-api.service.ts` file
- Rename to `AccessReportCreateRequest`

bitwarden_license/bit-common/src/dirt/access-intelligence/services/abstractions/access-intelligence-api.service.ts
```diff
@@ -0,0 +1,65 @@
...
+  AccessReportSummaryApi,
+} from "../../models";
+
+export abstract class AccessIntelligenceApiService {
+  /** GET /reports/organizations/{orgId}/latest */
+  abstract getLatestReport$(orgId: OrganizationId): Observable<AccessReportApi>;
+
+  /**
+   * POST /reports/organizations/{orgId}
+   */
```
Banrion, line 52: ⚠️ JSDoc comment suggestions across the files:

**Comments to explain "what" not "how"**
The JSDoc comments on these methods in the class describe the HTTP route rather than the method's purpose. Since implementations could use different endpoints or backends, the comment couples the abstraction to one specific implementation detail. Consider replacing with a description of what the method does (ex, Creates a new access report for the given organization) and moving route documentation to the implementation instead.

**Parameters and returns**
Consider adding _@param_ and _@returns_ tags to the JSDoc so callers get parameter descriptions and return type context on hover (e.g., what the Observable emits, or what happens when no report exists for the given org).

**Examples**
Examples in JSDoc (_@example_) are worth adding when the usage isn't obvious from the signature. For instance, showing the expected date range format for `getSummaryDataByDateRange$`, or clarifying that `createReport$` should be followed by `uploadReportFile$` to complete the upload flow.

bitwarden_license/bit-common/src/dirt/access-intelligence/services/abstractions/access-intelligence-api.service.ts
```diff
@@ -0,0 +1,65 @@
...
+import { OrganizationId, OrganizationReportId } from "@bitwarden/common/types/guid";
+
+import {
+  AccessReportApi,
+  AccessReportCreateApi,
+  AccessReportFileApi,
+  AccessReportSummaryApi,
+} from "../../models";
+
+export abstract class AccessIntelligenceApiService {
```
Banrion, line 39: ⚠️ Missing class JSDoc

bitwarden_license/bit-common/src/dirt/access-intelligence/services/implementations/api/default-access-intelligence-api.service.ts
```diff
@@ -0,0 +1,150 @@
...
+    reportId: OrganizationReportId,
+    file: File,
+    reportFileId: string,
+  ): Observable<void> {
+    const formData = new FormData();
+    formData.append("file", file, file.name);
+
+    const response = this.apiService.send(
+      "POST",
+      `/reports/organizations/${orgId}/${reportId}/file/report-data?reportFileId=${reportFileId}`,
```
Banrion: ❌ Response route changed in server pull request.

bitwarden_license/bit-common/src/dirt/access-intelligence/services/abstractions/access-intelligence-api.service.ts
Banrion, line 39: ❌ Missing endpoints and implementation:

- PATCH {organizationId}/{reportId} (full report update)
- GET {organizationId}/{reportId}/file/download (self hosted report file download)

Banrion (changes requested):

bitwarden_license/bit-common/src/dirt/access-intelligence/services/abstractions/access-intelligence-api.service.ts
lastbestdev: Added endpoints in latest. I also needed to update the ApiService to handle blob response types, so you'll see a change for that as well.

bitwarden_license/bit-common/src/dirt/access-intelligence/services/implementations/api/default-access-intelligence-api.service.ts
lastbestdev: Updated

bitwarden_license/bit-common/src/dirt/access-intelligence/services/abstractions/access-intelligence-api.service.ts
lastbestdev: Added in latest

bitwarden_license/bit-common/src/dirt/access-intelligence/services/abstractions/access-intelligence-api.service.ts
lastbestdev: Great point about not including implementation details in the documentation for the abstract class. 

Docs are added in latest

bitwarden_license/bit-common/src/dirt/access-intelligence/models/api/access-report-create.api.ts
lastbestdev: Done

Banrion: ⚠️ We can remove this file since the model was moved

bitwarden_license/bit-common/src/dirt/access-intelligence/models/api/access-report-file.api.ts
```diff
@@ -0,0 +1,29 @@
+import { BaseResponse } from "@bitwarden/common/models/response/base.response";
+import { FileUploadType } from "@bitwarden/common/platform/enums";
+
+import { AccessReportApi } from "./access-report.api";
+
+/**
+ * Response model returned when creating a report with the Access Intelligence V2 feature flag
```
Banrion: ⛏️ No need to mention the "Access Intelligence V2 feature flag" since v1 will be cleaned up and the comment will be out of date. In general, when adding comments we can leave out "architecture" guidance or refactoring comments unless it's important to context. In this case, sharing When it's used is not necessary for documenting What the model is for.

bitwarden_license/bit-common/src/dirt/access-intelligence/models/api/access-report.api.ts
```diff
@@ -24,6 +26,8 @@ export class AccessReportApi extends BaseResponse {
   memberRegistry: string = "";
```
Banrion, line 26: 📝  Member registry is actually included in the "reports" property. The individual keys on AccessReport models haven't been updated to reflect the new verbiage from the [glossary](https://github.com/bitwarden/clients/pull/19600/changes#diff-d0d10f5de03251e172538674d23cf9d529af0c289a9d00e0bdbc1e8ef4bc02e3). While it's not necessary for this pull request, I call it out to follow up on updating the model, which I'll track in a new Jira ticket.

bitwarden_license/bit-common/src/dirt/access-intelligence/services/abstractions/access-intelligence-api.service.ts
```diff
@@ -0,0 +1,151 @@
...
+  ): Observable<{ blob: Blob; fileName: string }>;
+
+  /**
+   * Updates an existing Access Intelligence report.
+   * @param orgId - the ID of the Organization the report belongs to
+   * @param reportId - the ID of the report to update
+   * @param request - the data to update on the report
+   * @returns observable emitting the updated Access Intelligence report
+   */
+  abstract updateReport$(
```
Banrion: ⚠️ Suggesting a change from `updateReport$` to `updateReportSettings$` since this route is intended to only update the settings, which ends up updating the summary and metrics. We should not allow updating the actual report. I suggest updating the comments too.

bitwarden_license/bit-common/src/dirt/access-intelligence/services/abstractions/access-intelligence-api.service.ts
```diff
@@ -0,0 +1,151 @@
...
+  reportData?: string;
+  contentEncryptionKey?: string;
+  summaryData?: string;
+  applicationData?: string;
+  metrics?: AccessReportMetricsApi;
+  fileSize?: number;
+}
+
+export interface AccessReportUpdateRequest {
+  reportData?: string;
```
Banrion: ⚠️ Pending server changes, remove the **reportData** from the update request. This was missed in previous reviews but we should not update the actual report and passing in this data will cause similar issues for the report load size.

bitwarden_license/bit-common/src/dirt/access-intelligence/services/abstractions/access-intelligence-api.service.ts
```diff
@@ -0,0 +1,151 @@
...
+export interface AccessReportCreateRequest {
+  reportData?: string;
+  contentEncryptionKey?: string;
+  summaryData?: string;
+  applicationData?: string;
+  metrics?: AccessReportMetricsApi;
+  fileSize?: number;
+}
+
+export interface AccessReportUpdateRequest {
```
Banrion: ⚠️ Rename to `AccessReportSettingsUpdateRequest` to match change request in https://github.com/bitwarden/clients/pull/19922/changes#r3080507340

bitwarden_license/bit-common/src/dirt/access-intelligence/services/implementations/api/default-access-intelligence-api.service.spec.ts
```diff
@@ -0,0 +1,401 @@
...
+    it("should propagate API errors", async () => {
+      mockApiService.send.mockRejectedValue(new Error("Download failed"));
+
+      await expect(firstValueFrom(service.downloadReportFile$(orgId, reportId))).rejects.toThrow(
+        "Download failed",
+      );
+    });
+  });
+
+  describe("updateReport$", () => {
```
Banrion: ⚠️ Update variable name to match change request: https://github.com/bitwarden/clients/pull/19922/changes#r3080507340

bitwarden_license/bit-common/src/dirt/access-intelligence/services/implementations/api/default-access-intelligence-api.service.ts
```diff
@@ -0,0 +1,180 @@
...
+      true,
+    );
+    return from(response).pipe(map((response) => new AccessReportFileApi(response)));
+  }
+
+  updateSummaryData$(
+    orgId: OrganizationId,
+    reportId: OrganizationReportId,
+    summaryData: string,
+    metrics?: Record<string, number>,
```
Banrion: ⚠️ Use **AccessReportMetricsApi**. The original method in `risk-insights-api.service.ts` uses the v1 version, **RiskInsightsMetricsData** (which is actually the incorrect model layer used). They should be identical and should match the `OrganizationReportMetricsData` model on the server.

bitwarden_license/bit-common/src/dirt/access-intelligence/services/abstractions/access-intelligence-api.service.ts
```diff
@@ -0,0 +1,151 @@
...
+   * @param reportId - the ID of the report to update
+   * @param summaryData - the encrypted summary data to store on the report
+   * @param metrics - optional map of metric names to their values
+   * @returns observable emitting the updated Access Intelligence report
+   */
+  abstract updateSummaryData$(
+    orgId: OrganizationId,
+    reportId: OrganizationReportId,
+    summaryData: string,
+    metrics?: Record<string, number>,
```
Banrion: ⚠️ Use `AccessReportMetricsApi`

bitwarden_license/bit-common/src/dirt/access-intelligence/services/abstractions/access-intelligence-api.service.ts
Banrion: ⚠️ I'll also flag that these properties are optional. I started a discussion for the server side to see if we want to make changes there. For the client side, we should enforce the required properties even if the server endpoint allows optional fields. The required fields should be **summaryData**, **applicationData**, and **metrics**. The **contentEncryptionKey** is currently reused for the same report and shouldn't be allowed to be changed currently since the report data is not updated. Let's remove that one from the request.

bitwarden_license/bit-common/src/dirt/access-intelligence/services/abstractions/access-intelligence-api.service.ts
```diff
@@ -0,0 +1,151 @@
...
+} from "../../models";
+
+export interface AccessReportCreateRequest {
+  reportData?: string;
+  contentEncryptionKey?: string;
+  summaryData?: string;
+  applicationData?: string;
+  metrics?: AccessReportMetricsApi;
+  fileSize?: number;
+}
```
Banrion, line 18: ⚠️ Similar to the [comment](https://github.com/bitwarden/clients/pull/19922/changes#r3080883939) on the model below, we want to enforce properties on the client side even if the server supports optional values (though a server side discussion has been started. Remove **reportData** and make other fields required.

```suggestion
export interface AccessReportCreateRequest {
  contentEncryptionKey: string;
  summaryData: string;
  applicationData: string;
  metrics: AccessReportMetricsApi;
  fileSize: number;
}
```

⚠️ In addition, the above defines the wiring for the create report endpoint expecting to save the report to a file. For legacy support, I suggest creating two different methods to capture the different model wiring, even though they use the same endpoint path under the hood, rather than supporting two different types as input for `createReport$`

Add a new type for legacy creations: 
```export interface AccessReportLegacyCreateRequest {
  reportData: string;
  contentEncryptionKey: string;
  summaryData: string;
  applicationData: string;
  metrics: AccessReportMetricsApi;
}
```

Then, add the legacy method to accept the Legacy type:
```
abstract createLegacyReport$(
    orgId: OrganizationId,
    request: AccessReportLegacyCreateRequest,
  ): Observable<AccessReportFileApi>;   // <---- Verify type of v1 endpoint return type
```

Banrion (changes requested):
Requesting a few changes for clarity and server change matches. I also suggested a new method to handle legacy report creations. It will only be needed for backwards compatibility while the file changes are being adopted and migrated, but makes it easier for cleanup and differences in the response and requests in the meantime. Happy to discuss if needed!

bitwarden_license/bit-common/src/dirt/access-intelligence/models/api/access-report-file.api.ts
lastbestdev: Updated doc comment

bitwarden_license/bit-common/src/dirt/access-intelligence/services/abstractions/access-intelligence-api.service.ts
lastbestdev: Updated in latest. 👍🏻 

I am a bit confused about when we would use this, since I believe we currently update the Summary and Application data properties separately, hence the 2 methods in this service for doing them separate (`updateSummaryData$` & `updateApplicationData$`)

bitwarden_license/bit-common/src/dirt/access-intelligence/services/abstractions/access-intelligence-api.service.ts
lastbestdev: Properties updated in the interface accordingly. reportData and contentEncryptionKey both removed

bitwarden_license/bit-common/src/dirt/access-intelligence/services/implementations/api/default-access-intelligence-api.service.ts
lastbestdev: Good catch, I didn't see this model to use here!

bitwarden_license/bit-common/src/dirt/access-intelligence/services/abstractions/access-intelligence-api.service.ts
lastbestdev: Added in latest 👍🏻 

This makes sense, since we are going to have a state potentially where V2 Access Intelligence is enabled, but report file storage is not yet. We would want this API service to support the old create report logic without file storage.

bitwarden_license/bit-common/src/dirt/access-intelligence/models/api/report-file.api.ts
```diff
@@ -0,0 +1,20 @@
+import { BaseResponse } from "@bitwarden/common/models/response/base.response";
+
+/**
+ * Metadata for an uploaded report file.
+ */
+export class ReportFileApi extends BaseResponse {
+  id: string | undefined;
```
Banrion: ⛏️ I think id's are required, we shouldn't get a response without one. Prefer removing undefined or the `id?: string` if needed.

Banrion (dismissed):

Banrion (approved):

PR #19923, opened 2026-04-01
[PM-34230] Blumira integration changes

voommen-livefront:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-34230

## 📔 Objective

Allow integration into Blumira using HTTP Event Collector (HEC) using token auth.
Renamed Huntress specific integration to connect-via-token-auth (in order to keep it generic)
The example URL will be obtained from the integration

## 📸 Screenshots

image

image


voommen-livefront:
A new PR #19914 replaces this PR.

PR #19932, opened 2026-04-02
[Shared Unlock] [PM-34073] Disable process reload in dev mode

quexten:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-34073

## 📔 Objective

Debugging gets harder when process reload is active, and process reload is not really needed in dev mode. This PR disables process reload in dev mode. This makes debugging a few things around lock (and shared unlock) significantly easier.

### DIFF
--- a/apps/browser/src/background/main.background.ts
+++ b/apps/browser/src/background/main.background.ts
@@ -1319,7 +1319,7 @@ export default class MainBackground {
       this.accountService,
       this.logService,
       this.authService,
-      this.platformUtilsService
+      this.platformUtilsService,
     );
 
     // Background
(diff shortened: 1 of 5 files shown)


dani-garcia (commented):
Sounds fairly reasonable, but I worry that the behavior difference between dev and prod will cause some developers to miss some bugs if they're not aware. 

Maybe we can change the it to also check some env variable and tell devs to use a different start command? Kinda like how we have `npm run electron:ignore` if you want to disable SSL validation, rather than just doing it by default

quexten:
@dani-garcia are you saying the default should still be that process reload is disabled? Or should process reload be enabled by default, but disabled by choice for some testing?

dani-garcia:
I was thinking enabled by default to match prod, but with an easy way to explicitly disable it

PR #19933, opened 2026-04-02
[Shared Unlock] [PM-34073] Navigate when unlock happens in the background while in the lock component

quexten:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-34073

## 📔 Objective

If the account is unlocked while the user sitting on the lock component, but without going via a lock component flow (an example here is the new shared unlock feature), then the client should automatically continue navigation. This PR listens to user-key changes and navigates when the user-key becomes present (the vault unlocks).


libs/key-management-ui/src/lock/components/lock.component.ts
```diff
@@ -268,6 +275,32 @@ export class LockComponent implements OnInit, OnDestroy {
...
+          if (this.unlockingViaManualFlow) {
+            return;
+          }
+          await this.pinService.userUnlocked(this.activeAccount!.id);
+          await this.doContinue({});
+        }),
+        takeUntil(this.destroy$),
+      )
+      .subscribe();
+  }
```
Thomas-Avery, line 302: 💭 
I'm concerned with how this will work with unlocked account switching which we put in a work around here https://github.com/bitwarden/clients/pull/12875 for.

If I'm reasoning about this correctly the unlocked account switch would now fire?: 

```
          await this.pinService.userUnlocked(this.activeAccount!.id);
          await this.doContinue({});
```

Which is a different behavior. I wonder if there is a better way to express these two cases to make it clearer what's happening.

Thomas-Avery (changes requested):

libs/key-management-ui/src/lock/components/lock.component.ts
quexten: Hm. Good point. My line of thinking was that a user should never remain on the lock screen, if their account is in-fact unlocked. Maybe, to keep it simple, the better way is to offer an observable in the shared unlock service, that emits events when shared unlock causes an unlock?

Thomas-Avery: > Maybe, to keep it simple, the better way is to offer an observable in the shared unlock service, that emits events when shared unlock causes an unlock?

Yes I think that is the correct approach here. Great suggestion.

quexten: Moving to draft until we get the SDK changes merged. The clients changes will then follow the approach outlined above.

PR #19952, opened 2026-04-02
Add shield icon switching

coltonhurst:
- Show the shield when there is no text
- Show "esc" when there is text, and show the shield in the bottom right in the action bar


apps/desktop/src/magnify/app/components/action-bar/action-bar.component.css
```diff
@@ -7,6 +7,16 @@
   border-top: 1px solid var(--border-color);
 }
 
+.action-bar-logo {
+  margin-left: auto;
+  flex-shrink: 0;
+  width: 9px;
+  height: 11px;
+  color: var(--icon-color);
+  opacity: 0.4;
```
BryanCunningham, line 16: @coltonhurst Why would we need an opacity on a logo? I'm assuming this is to change the color? Could probably just change the fill color and avoid this

coltonhurst:
Hey @BryanCunningham thanks for the comments! This is actually for the innovation sprint and isn't going to be merged to main yet - just to `magnify/main`

BryanCunningham:
> Hey @BryanCunningham thanks for the comments! This is actually for the innovation sprint and isn't going to be merged to main yet - just to `magnify/main`

No worries! Yeah, I seen that. That's why I just approved it. But, feels like the kind of thing we'd want to fix before going to main, that's all.

PR #19953, opened 2026-04-02
[PM-34119] Web New Item Dialog

nick-livefront:
## 🎟️ Tracking

[PM-34119](https://bitwarden.atlassian.net/browse/PM-34119)

## 📔 Objective

Adds new item selection dialog in preparation for the addition of Bank Account, Driver's License, and Passport cipher types.
- Adds `AddItemGridComponent` that stores all logic for for which item types are visible and is cross platform compatible. 
- When `PM32009NewItemTypes` feature flag is enabled the new button on the web vault and admin console will open the dialog rather than the dropdown menu.
- @bitwarden/team-ui-foundation I made a few tweak to the dialog component, I'll make an inline comment there. 

## 📸 Screenshots

|Dark Mode | Light Mode|
|-|-|
|<video src="https://github.com/user-attachments/assets/57a68e60-1c7e-464c-8ffe-8359f4b6e4d8" />|<video  src="https://github.com/user-attachments/assets/fc5afc95-8fb0-49e1-90c7-4ef7f9ce75be" />

[PM-34119]: https://bitwarden.atlassian.net/browse/PM-34119?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/components/src/dialog/dialog/dialog.component.html
nick-livefront, line 1: @bitwarden/team-ui-foundation One change here to support the designs of the new item dialog. [Figma](https://www.figma.com/design/8TTVF6pZE4Lxu8FNvYN5VV/Item-types?node-id=123-42776&t=6mrh8xPsnAVkyBBP-4)

- The dialog without a footer currently has extra padding on the bottom. I added a `hasFooter` conditional to only render the wrapping `<div>` when present. This also why there are storybook changes for existing implementations of the dialog.

vleague2 (dismissed):
UIF changes look fine!

libs/vault/src/components/add-item-dialog/add-item-dialog.component.ts
```diff
@@ -0,0 +1,61 @@
...
+@Component({
+  selector: "vault-add-item-dialog",
+  templateUrl: "./add-item-dialog.component.html",
+  changeDetection: ChangeDetectionStrategy.OnPush,
+  imports: [DialogModule, I18nPipe, AddItemGridComponent],
+})
+export class AddItemDialogComponent {
+  constructor(
+    private readonly dialogRef: DialogRef<AddItemDialogCloseResult>,
+    @Inject(DIALOG_DATA) readonly data: AddItemDialogData,
```
jrmccannon: Newer pattern is to use the `inject()` function and then `data` could be set as `protected readonly` instead of public.

See [DecryptionFailureDialog](https://github.com/bitwarden/clients/blob/a07c9ebf6bf3e87cc1d8710f119cb82434d5e1b3/libs/vault/src/components/decryption-failure-dialog/decryption-failure-dialog.component.ts#L36) as an example.

apps/web/src/app/admin-console/organizations/collections/vault-header/vault-header.component.ts
```diff
@@ -109,6 +109,9 @@ export class VaultHeaderComponent {
   // eslint-disable-next-line @angular-eslint/prefer-output-emitter-ref
   @Output() searchTextChanged = new EventEmitter<string>();
 
+  /** Emits an event when the add item dialog should be opened */
+  onOpenAddItemDialog = output();
```
jrmccannon: Missing `readonly` (vault version has it)

apps/web/src/app/admin-console/organizations/collections/vault-header/vault-header.component.ts
```diff
@@ -218,6 +221,10 @@ export class VaultHeaderComponent {
     this.onAddCipher.emit(cipherType);
   }
 
+  openAddItemDialog(): void {
```
jrmccannon: Missing `protected` (vault version has it)

jrmccannon (changes requested):
Looks good. Just a few comments.

libs/vault/src/components/new-cipher-menu/new-cipher-menu.component.ts
```diff
@@ -24,18 +26,10 @@ import { I18nPipe } from "@bitwarden/ui-common";
...
-  // FIXME(https://bitwarden.atlassian.net/browse/CL-903): Migrate to Signals
-  // eslint-disable-next-line @angular-eslint/prefer-signals
-  canCreateFolder = input(false);
-  // FIXME(https://bitwarden.atlassian.net/browse/CL-903): Migrate to Signals
-  // eslint-disable-next-line @angular-eslint/prefer-signals
-  canCreateCollection = input(false);
-  // FIXME(https://bitwarden.atlassian.net/browse/CL-903): Migrate to Signals
-  // eslint-disable-next-line @angular-eslint/prefer-signals
-  canCreateSshKey = input(false);
+  readonly canCreateCipher = input(false);
```
shane-melton, line 29: 👍 Thanks for fixing these!

libs/vault/src/components/add-item-grid/add-item-grid.component.ts
```diff
@@ -0,0 +1,82 @@
...
+  imports: [CommonModule, I18nPipe, IconTileComponent, IconComponent, ItemModule, TypographyModule],
+})
+export class AddItemGridComponent {
+  readonly canCreateFolder = input(false);
+  readonly canCreateCollection = input(false);
+  readonly canCreateSshKey = input(false);
+
+  readonly cipherSelected = output<CipherType>();
+  readonly folderSelected = output();
+  readonly collectionSelected = output();
```
shane-melton: ❓ Instead of having different outputs for each "type" that we have to register separate handlers for in the templates; could we create one output like `output<AddItemDialogCloseResult>()`.  I think this would reduce the API surface area for the component and makes it easier to discover all the "items" that can be selected.

shane-melton (commented):
Looks good, nice work! Just have one question/suggestion around the public API of the `add-item-grid` component.

libs/vault/src/components/add-item-dialog/add-item-dialog.component.ts
nick-livefront: [ea3d9c2](https://github.com/bitwarden/clients/pull/19953/commits/ea3d9c245e75f87587dd736e47ca2c9fd80ca952)

libs/vault/src/components/add-item-grid/add-item-grid.component.ts
nick-livefront: Great call! I ended up moving the result type from the dialog into grid. The dialog still exports a separate type but currently it is equivalent to the `AddItemGridResult` type.

[5cb3243](https://github.com/bitwarden/clients/pull/19953/commits/5cb3243dddafbe0351fad6d4fe53805e53afa5fb)

apps/web/src/app/admin-console/organizations/collections/vault-header/vault-header.component.ts
nick-livefront: [6383f63](https://github.com/bitwarden/clients/pull/19953/commits/6383f6380688ee23fc6f48132fd80c2b93d18e3e)

apps/web/src/app/admin-console/organizations/collections/vault-header/vault-header.component.ts
nick-livefront: [6383f63](https://github.com/bitwarden/clients/pull/19953/commits/6383f6380688ee23fc6f48132fd80c2b93d18e3e)

shane-melton (approved):
Looks great, nice work!

vleague2 (approved):

jrmccannon (approved):
Looks good! Thanks!

PR #19968, opened 2026-04-03
Switch to the new RequestSignForWindow

Hinton:
## 📔 Objective

Updates `windows-rs` to include the Windows SDK bump from https://github.com/microsoft/windows-rs/issues/3896. This allows us to use `RequestSignForWindow` which lets us pass in a window handle to ensure the windows hello prompt appears ontop of the application window.

This was implemented conditionally so that old windows versions would keep working with the older APIs.

### DIFF
--- a/apps/desktop/desktop_native/core/src/biometric_v2/linux.rs
+++ b/apps/desktop/desktop_native/core/src/biometric_v2/linux.rs
@@ -53,7 +53,7 @@ impl super::BiometricTrait for BiometricLockSystem {
         polkit_is_bitwarden_policy_available().await
     }
 
-    async fn enroll_persistent(&self, _user_id: &str, _key: &[u8]) -> Result<()> {
+    async fn enroll_persistent(&self, _user_id: &str, _key: &[u8], _hwnd: &[u8]) -> Result<()> {
         // Not implemented
         Ok(())
     }


apps/desktop/desktop_native/core/src/biometric_v2/windows.rs
```diff
@@ -232,34 +242,52 @@ async fn windows_hello_authenticate(message: String) -> Result<bool> {
...
-            break;
-        }
-    });
-    // Only stop focusing once this function exits. The focus MUST run both during the initial
-    // creation with RequestCreateAsync, and also with the subsequent use with RequestSignAsync.
-    let _guard = scopeguard::guard((), |_| {
-        stop_focusing.store(true, std::sync::atomic::Ordering::Relaxed);
-    });
+    // Newer Windows SDK builds include support focusing the dialogs
+    let available = ApiInformation::IsMethodPresent(
```
quexten, line 251: I don't think this will work correctly for unlocking via the  extension. In the existing code we are doing:
https://github.com/bitwarden/clients/blob/ea447325383217774d2b9fa263812c1c3723a9cd/apps/desktop/desktop_native/core/src/biometric_v2/windows.rs#L223
which will get the hwnd of the browser in case the browser is in the foreground, an the desktop app in case the desktop app is in the foreground. However, if the desktop app is minimized or closed, I'm not sure this will work at all, and if the browser and desktop app are on display, but you are unlocking via the browser, the prompt would show on the wrong window (on the desktop window).

Can we switch the hwnd to just be the foreground window?

mzieniukbw: From what i remember, the new Api brings the window into focus, even if hidden  or minimised. Providing browser as the handle to this api, brings the browser extension popup / popout (if separate process id) or the browser into focus. The Windows Hello popup is then locked into the provided process id, which means you can't use the browser extension / browser / desktop until you confirm. I think this will work as expected.

Hinton: @mzieniukbw do we want to try and switch to `GetForegroundWindow`? Not sure exactly when I get time to work on this again so if KM gets to it before me feel free to take ownership of the PR.

PR #19982, opened 2026-04-03
[PM-24795] When logged out, "(empty)" options appear under File

jengstrom-bw:
## 🎟️ Tracking

[Jira](https://bitwarden.atlassian.net/browse/PM-24795)

## 📔 Objective

Fix `Lock Vault` and `Log Out` menu item enabled state when there are no lockable or authenticated accounts


apps/desktop/src/main/menu/menu.first.ts
shane-melton: 🎨 `visible` can be left off or always set to `false` due to the `!account.isAuthenticated` condition above.

shane-melton, line 1: 🎨  It could be helpful to setup some tests for this file to avoid this again in the future.

shane-melton (commented):
Logic changes look good, I'd just suggest we add some tests to prevent this from happening again in the future.

jengstrom-bw:
Added tests and removed unneeded `visible`

apps/desktop/src/main/menu/menu.first.spec.ts
shane-melton, line 1: 🎉 Thanks!

shane-melton (approved):
Looks good!

PR #20004, opened 2026-04-06
[PM-31119] Run side-effects in sdk unlock service

quexten:
## 🎟️ Tracking

https://github.com/bitwarden/sdk-internal/pull/905
https://bitwarden.atlassian.net/browse/PM-31119

## 📔 Objective

Runs side-effects that provide the user-key to the auto-key (never lock / session key) system, and the biometrics subsystem. PIN is handled separately, in a separate PR.


libs/unlock/src/default-unlock.service.ts
```diff
@@ -224,4 +244,36 @@ export class DefaultUnlockService implements UnlockService {
...
+    if (await firstValueFrom(this.biometricStateService.biometricUnlockEnabled$(userId))) {
+      await this.biometricsService.setBiometricProtectedUnlockKeyForUser(userId, userKey);
+    }
+    if (await this.shouldStoreSessionKey(userId)) {
+      await this.stateService.setUserKeyAutoUnlock(userKey.toBase64(), { userId: userId });
+    }
+    await this.stateProvider.setUserState(USER_EVER_HAD_USER_KEY, true, userId);
+  }
+
+  private async shouldStoreSessionKey(userId: UserId): Promise<boolean> {
```
mzieniukbw: ❓  this name is somewhat confusing. Maybe something along the lines of should store user key auto unlock key ?

libs/unlock/src/default-unlock.service.ts
```diff
@@ -224,4 +244,36 @@ export class DefaultUnlockService implements UnlockService {
...
+      return true;
+    }
+
+    const vaultTimeout = await firstValueFrom(
+      this.stateProvider
+        .getUserState$(VAULT_TIMEOUT, userId)
+        .pipe(filter((timeout) => timeout != null)),
+    );
+
+    return vaultTimeout == VaultTimeoutStringType.Never;
```
mzieniukbw, line 262: 🎨  not much test coverage for this

libs/unlock/src/default-unlock.service.ts
```diff
@@ -224,4 +244,36 @@ export class DefaultUnlockService implements UnlockService {
...
+    }
+    if (await this.shouldStoreSessionKey(userId)) {
+      await this.stateService.setUserKeyAutoUnlock(userKey.toBase64(), { userId: userId });
+    }
+    await this.stateProvider.setUserState(USER_EVER_HAD_USER_KEY, true, userId);
+  }
+
+  private async shouldStoreSessionKey(userId: UserId): Promise<boolean> {
+    if (this.platformUtilsService.getClientType() === ClientType.Cli) {
+      return true;
```
mzieniukbw, line 253: 🎨  not much test coverage for this

libs/unlock/src/default-unlock.service.ts
quexten: Renamed.

libs/unlock/src/default-unlock.service.ts
quexten: Added coverage for this, and unrelated code in this file that was not covered.

mzieniukbw (dismissed):

eligrubb (approved):

libs/unlock/src/default-unlock.service.ts
```diff
@@ -209,4 +229,36 @@ export class DefaultUnlockService implements UnlockService {
...
+  // When unlocking, certain side-effects must be run, such as setting the never-lock key and the biometrics key.
+  // Currently this does not happen from within the SDK but form here instead.
+  private async runOnUnlockSideEffects(
+    userId: UserId,
+    client: Ref<PasswordManagerClient>,
+  ): Promise<void> {
+    const userKey = SymmetricCryptoKey.fromString(
+      await client.value.crypto().get_user_encryption_key(),
+    );
+    if (await firstValueFrom(this.biometricStateService.biometricUnlockEnabled$(userId))) {
```
djsmith85, line 242: ⚠️  This line could be moved into the new method being called on the `biometricsService`. It would just early return if the user did not have biometric unlock enabled. This also means the `UnlockService` no longer requires a dependency on `biometricStateService`.

🎨 If this patter is extended, then I would suggest creating an unlock event (which might already gets emitted by the account-object) and adding subscriptions in the consuming services to that unlock event. This shifts the responsibility to set/remove keys on to the subscribed services. The biometricsService knows best if it needs to set something and how. StateService/-Provider are a little low-level for this, but easy enough to create a service for that. 
This pattern also greatly reduces the dependencies the unlockService requires.

djsmith85 (changes requested):
Thank you for adding documentation to the USER_EVER_HAD_USER_KEY

libs/unlock/src/default-unlock.service.ts
quexten: First, thank you for the concern about the technical debt incurred by KM here.

I've looked into this, and I don't think we want to move forward with moving the condition into the biometrics service from the KM perspective right now in this PR. I know we just had a 1:1 sync on this, so I'll explain why.

The biometrics service's public interface currently expects the caller to know when to store the key. This is evident both by the name: `setBiometricProtectedUnlockKeyForUser`, and also by the fact that the caller - electronKeyService - currently handles the same exact check where it queries the biometrics state service https://github.com/bitwarden/clients/blob/daf03aa13edf82423247e4f4e1bd03bdb86f8ad5/apps/desktop/src/key-management/electron-key.service.ts#L57.

Moving it into the biometrics service is at minimum a breaking change, and as cleanup requires removing the check in electron key service, and a few dependency changes in e.g. ElectronKeyService, and multiple tests that are broken / now invalid, *and requires QA of all of biometrics in general before this PR can be merged*.

Further, the name is no longer valid, and should then really be "provideUserKey" or something similar, to signify that the biometrics system itself determines (based on the setting) whether or not to store the key, because it would be unintuitive to call `setBiometricProtectedUnlockKeyForUser` without any effect.

Essentially, the comment is scope-creeping this ticket which is only supposed to port existing behavior into the unlock service into a refactor of the biometrics system, which is not the intent of this ticket.

*If* we do want to make this change, then I want at least other KM devs to agree that the scope creep and interface change here is accepted and wanted. My local git diff where I apply the suggested move has 13 files changed https://github.com/bitwarden/clients/pull/20092.

Also:
>  If this patter is extended, then I would suggest creating an unlock event (which might already gets emitted by the account-object) and adding subscriptions in the consuming services to that unlock event. 

I don't know if this applies throughout all teams, but service-to-service subscriptions are a pattern that KM does not want to use. (I had a DM with @Thomas-Avery about this; https://bitwarden.slack.com/archives/D07GV7LTB9A/p1756910855832469, for his reference).

I'll bring this up in the KM dev sync, but I suspect that the cost of having three service dependencies is cheaper than the QA effort and extra effort if we did refactor. At any rate that refactor should not be part of this PR.

cc @Thomas-Avery

djsmith85: @quexten: That's understandable and I agree this should not be solved with this ticket. I'll be approving to unblock this work. 

I'm glad though we discussed this and it has uncovered some other design issues, than can be discussed by KM/others in a follow-up. Thanks for the detailed investigation and explanation, including a PR your team can use as basis for discussion.

If KM has decided that service-to-service subscriptions are not something they want to follow, that's fine with me it was just a suggestion to decouple service-functionality and reduce required dependencies.

djsmith85 (approved):

PR #20005, opened 2026-04-06
Resolve Inquirer Dependency Issues

coltonhurst:
When running `npm ci` at root, I receive errors that indicate the `package.json` is out of sync with the `package-lock.json` file.

In https://github.com/bitwarden/clients/pull/18433 the version string for `@napi-rs/cli` was updated, but it seems the resolved package entries for the dependencies weren't also updated. I may have this wrong.

This PR aims to fix the `npm ci` issues.


coltonhurst:
Closing this as the issue was due to different npm versions.

PR #20006, opened 2026-04-06
[CL-978] Fields updates

BryanCunningham:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/CL-978

## 📔 Objective
Updates all form fields to use refreshed design language

#### NOTE:
This is being merged to a feature branch

@bitwarden/team-vault-dev Changes in your code are just to remove the `disableReadOnlyBorder` input from some fields as this input is no longer supported/readonly inputs do not have any borders in the new style

@bitwarden/team-admin-console-dev Changes in your code are just a minor spacing adjustment that was no longer necessary

### Field
image
#### Large
image

#### Inactive
image

#### Readonly
image

### Select
image

### Multi-select
image

### Textarea
image

### Search
image


apps/web/src/app/billing/shared/self-hosting-license-uploader/self-hosting-license-uploader.component.html
```diff
@@ -4,8 +4,14 @@
 <form [formGroup]="form" [bitSubmit]="submit">
   <bit-form-field>
     <bit-label>{{ description | i18n }}</bit-label>
-    <div class="tw-my-1">
-      <button bitButton type="button" buttonType="secondary" (click)="fileSelector.click()">
+    <div class="tw-text-main tw-flex tw-items-center tw-h-full tw-gap-2">
```
vleague2: Should we be using one of the newer color variables instead?

bitwarden_license/bit-web/src/app/secrets-manager/settings/porting/sm-import.component.html
```diff
@@ -3,8 +3,16 @@
 <form [formGroup]="formGroup" [bitSubmit]="submit" class="tw-max-w-xl">
   <bit-form-field>
     <bit-label>{{ "fileUpload" | i18n }}</bit-label>
-    <div class="file-selector tw-pt-2 tw-pb-1 tw-break-words">
-      <button bitButton type="button" buttonType="secondary" (click)="fileSelector.click()">
+    <div
+      class="file-selector tw-text-main tw-flex tw-items-center tw-h-full tw-gap-2 tw-break-words"
```
vleague2: Hmm I wonder if it's worth even making these fixes in this PR, if the feature branch is also going to include migrating these to the new file selector anyway

libs/components/src/form-field/form-field.component.html
```diff
@@ -1,119 +1,43 @@
...
-            }
-          </label>
-        </div>
-        <div
-          class="tw-min-w-3 tw-grow tw-border-l-0 group-focus-within/bit-form-field:tw-border-l-0 !tw-rounded-e-lg"
-          [ngClass]="inputBorderClasses"
-        ></div>
-      </div>
-    </div>
+<div [class]="fieldContainerClasses">
```
vleague2: Ok my one gripe is that this is `fieldContainerClasses` but the `bitFieldContainer` is a different element

libs/components/src/form-field/form-field.component.html
```diff
@@ -1,119 +1,43 @@
...
+      <sup class="tw-text-sm tw-text-fg-danger tw-relative tw-top-[-1px]">*</sup>
+      <span class="tw-sr-only"> ({{ "required" | i18n }})</span>
+    }
+  </label>
+  <div bitFieldContainer [size]="size()" [hasError]="input().hasError" [readOnly]="readOnly">
     <div
-      class="tw-gap-1 tw-bg-background tw-rounded-lg tw-flex tw-min-h-11 [&:has(input:read-only,textarea:read-only):not(:has(button:not([aria-disabled='true'])))]:tw-bg-secondary-100"
+      #prefixContainer
+      [hidden]="!prefixHasChildren()"
+      class="tw-flex tw-items-center tw-border tw-border-border-base tw-px-3 tw-bg-bg-quaternary tw-rounded-s-[11px]"
```
vleague2: 11px seems so specific

libs/components/src/input/input.directive.ts
```diff
@@ -136,21 +161,20 @@ export class BitInputDirective implements BitFormFieldControl {
...
       "tw-rounded-lg",
       // Hover
-      this.hasError ? "hover:tw-border-danger-700" : "hover:tw-border-primary-600",
+      this.hasError ? "hover:tw-border-border-danger" : "hover:tw-border-border-brand",
       // Focus
-      "focus:hover:tw-border-primary-600",
-      "disabled:tw-bg-secondary-100",
-      "disabled:hover:tw-border-secondary-500",
-      "focus:tw-border-primary-600",
+      // "focus:hover:tw-border-primary-600",
```
vleague2: Old?

vleague2 (commented):
Quick first look at the field itself

apps/web/src/app/billing/shared/self-hosting-license-uploader/self-hosting-license-uploader.component.html
BryanCunningham: Yes, good call

libs/components/src/form-field/form-field.component.html
BryanCunningham: It is. It's necessary to match the outer container's border-radius. If it's the same it's too much. Had to remove 1 px

apps/web/src/app/billing/shared/self-hosting-license-uploader/self-hosting-license-uploader.component.html
BryanCunningham: I agree with your comment below. Might not even be worth making these fixes. I'll update these usages to use the new fileupload

bitwarden_license/bit-web/src/app/secrets-manager/settings/porting/sm-import.component.html
BryanCunningham: Probably not worth it if we are updating to the new component

libs/components/src/form-field/form-field.component.html
BryanCunningham: Updated the class name

libs/components/src/stories/colors.mdx
```diff
@@ -104,7 +104,7 @@ export const BackgroundCard = ({ name, primitiveColor }) => {
         <BackgroundCard name="primary" primitiveColor="white" />
         <BackgroundCard name="secondary" primitiveColor="gray-050" />
         <BackgroundCard name="tertiary" primitiveColor="gray-050" />
-        <BackgroundCard name="quaternary" primitiveColor="gray-200" />
+        <BackgroundCard name="quaternary" primitiveColor="gray-100" />
         <BackgroundCard name="gray" primitiveColor="gray-300" />
         <BackgroundCard name="disabled" primitiveColor="gray-100" />
```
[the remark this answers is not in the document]
vleague2: Seems like a legit comment

apps/web/src/app/billing/shared/self-hosting-license-uploader/self-hosting-license-uploader.component.html
vleague2: I don't think it's in `main` yet so I'd just revert these and let them be borked in the feature branch till we have it available

libs/components/src/form-field/form-field.component.ts
```diff
@@ -17,17 +15,22 @@ import { I18nPipe } from "@bitwarden/ui-common";
...
 import { BitHintDirective } from "../form-control/hint.directive";
 import { BitLabelComponent } from "../form-control/label.component";
-import { inputBorderClasses } from "../input/input.directive";
 
 import { BitErrorComponent } from "./error.component";
+import { BitFieldContainerDirective, FieldContainerSize } from "./field-container.directive";
 import { BitFormFieldControl } from "./form-field-control";
 
 // FIXME(https://bitwarden.atlassian.net/browse/CL-764): Migrate to OnPush
 // eslint-disable-next-line @angular-eslint/prefer-on-push-component-change-detection
```
vleague2: Can't remember if I already left a comment about fixing these while we're in here

libs/components/src/input/input.directive.ts
```diff
@@ -61,9 +79,9 @@ export class BitInputDirective implements BitFormFieldControl {
 
   readonly id = input(`bit-input-${nextId++}`);
 
-  @HostBinding("attr.aria-describedby") ariaDescribedBy?: string;
+  ariaDescribedBy?: string;
```
vleague2: Wonder if this should become a signal?

libs/components/src/input/input.directive.ts
```diff
@@ -35,21 +34,40 @@ export function inputBorderClasses(error: boolean) {
...
       "tw-border-none",
       "focus:tw-outline-none",
-      "[&:is(input,textarea):disabled]:tw-bg-secondary-100",
+      "tw-bg-transparent",
+      "tw-text-fg-heading",
+      "[&:is(textarea)]:tw-resize-none",
+      "[&:is(input,textarea):disabled]:tw-bg-bg-secondary",
+      "[&:is(input,textarea):disabled]:!tw-placeholder-fg-inactive",
+      "[&:is(input,textarea):disabled]:!tw-text-fg-inactive",
+      "[&:-webkit-autofill]:tw-shadow-[inset_0_0_0_1000px_var(--color-bg-secondary)]",
```
vleague2: Can you speak more to the webkit-autofill styles?

libs/components/src/input/input.directive.ts
```diff
@@ -85,18 +103,32 @@ export class BitInputDirective implements BitFormFieldControl {
...
+  ngAfterViewInit() {
+    this.adjustTextareaHeight();
+  }
+
+  protected onInput() {
     this.ngControl?.control?.markAsUntouched();
+    this.adjustTextareaHeight();
+  }
+
+  private adjustTextareaHeight() {
```
vleague2, line 88: What do we need this function for? (If it's to always grow the textarea as the scroll area becomes larger -- do we actually want that?)

libs/components/src/search/search.component.html
```diff
@@ -1,43 +1,32 @@
 <form
+  bitFieldContainer
```
vleague2, line 2: So using the form field with prefix/suffix didn't end up working for search? (I know we've talked about it, I just can't remember haha)

vleague2 (commented):
Idk where to put these Storybook findings:
- Dropdown arrow is missing on Field > Select story
- Should the dropdown arrow in selects become a lighter gray when it's inactive? (Actually in the figma, there's no arrow at all for inactive?)
- There is an X with weird styling at the end of the multi select field when selections are made

libs/components/src/form-field/form-field.component.html
BryanCunningham: For posterity: We talked about this outside of PR review and decided this needs to stay a magic number

libs/components/src/input/input.directive.ts
BryanCunningham: `-webkit-autofill` puts a blue background on an input that has been auto filled. I just sort of assumed we didn't want that

libs/components/src/input/input.directive.ts
BryanCunningham: I was told that is what we wanted. But, there should probably be some sort of max-height involved

libs/components/src/search/search.component.html
BryanCunningham: It did not. The form-field prefix is rendered in a dark gray box. And, we didn't want that for the search component I guess. So, abstracted what I could and reused it for search

apps/web/src/app/billing/shared/self-hosting-license-uploader/self-hosting-license-uploader.component.html
BryanCunningham: Reverted

bitwarden_license/bit-web/src/app/secrets-manager/settings/porting/sm-import.component.html
BryanCunningham: Reverted

libs/components/src/input/input.directive.ts
vleague2: Ooh interesting. I wonder if there is any a11y implication for that?

libs/components/src/search/search.component.html
vleague2: Ok cool makes sense!

libs/components/src/form-field/form-field.component.html
```diff
@@ -1,119 +1,43 @@
...
-    >
-      <ng-container *ngTemplateOutlet="labelContent"></ng-container>
-    </label>
     <div
-      class="tw-gap-1 tw-flex tw-min-h-[1.85rem] tw-border-0 tw-border-solid tw-items-center"
-      [ngClass]="{
-        'tw-border-secondary-300/50 tw-border-b tw-pb-[2px]': !disableReadOnlyBorder,
-        'tw-border-transparent tw-pb-[3px]': disableReadOnlyBorder,
-      }"
+      class="tw-flex tw-w-full tw-items-center tw-pe-3 has-[[biticonbutton]:last-child]:tw-pe-1 has-[bit-select]:tw-pe-0 has-[bit-multi-select]:tw-pe-0 has-[textarea]:!tw-items-start"
```
vleague2, line 21: `has-[[biticonbutton]:last-child]` Maybe I'm missing something, but isn't this functionally the same as `has-[[biticonbutton]]` since if there's at least 1, it would be the last child?

libs/importer/src/components/import.component.html
```diff
@@ -467,8 +467,16 @@ <h2 class="tw-font-medium" bitTypography="h6">{{ "data" | i18n }}</h2>
...
+          <div
+            class="file-selector tw-text-main tw-flex tw-items-center tw-h-full tw-gap-2 tw-break-words"
+          >
+            <button
+              bitButton
+              type="button"
+              size="small"
+              buttonType="secondary"
+              (click)="fileSelector.click()"
+            >
```
vleague2: There's a couple more of these

libs/vault/src/cipher-view/read-only-cipher-card/read-only-cipher-card.component.ts
```diff
@@ -1,31 +1,14 @@
...
-      // Delay model update until next change detection cycle
-      setTimeout(() => {
-        if (this.formFields) {
-          this.formFields.last.disableReadOnlyBorder = true;
-        }
-      });
-    }
-  }
-}
+export class ReadOnlyCipherCardComponent {}
```
vleague2, line 14: We can probably either remove this wrapper component or make a follow-up for vault to do so. I think its only purpose was to do that border code

libs/components/src/form-field/form-field.component.html
BryanCunningham: @vleague2 So, the idea here is: Only apply this specific padding-inline-end if the last child is an icon button. This selector may not be _quite_ right though now that I'm thinking about it.

libs/vault/src/cipher-view/read-only-cipher-card/read-only-cipher-card.component.ts
BryanCunningham: Yeah, it can likely be removed.

libs/components/src/form-field/form-field.component.html
vleague2: Ooh yeah I think it is not applying to the child because it's a `has` selector, not a child target

libs/vault/src/cipher-form/components/attachments/cipher-attachments.component.html
```diff
@@ -124,7 +124,7 @@ <h2 class="tw-sr-only" id="attachments">{{ "attachments" | i18n }}</h2>
           {{ "maxFileSizeSansPunctuation" | i18n }}
         </p>
       } @else {
-        <bit-progress [barWidth]="uploadProgress()" class="tw-block tw-mt-2" />
+        <bit-progress-bar [value]="uploadProgress()" class="tw-block tw-mt-2" />
```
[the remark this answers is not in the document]
BryanCunningham: Brought up to date with main

libs/vault/src/cipher-view/read-only-cipher-card/read-only-cipher-card.component.ts
BryanCunningham: There are still some styles in here on the bit-card so I'm going to leave it for now I think

libs/components/src/input/input.directive.ts
BryanCunningham: @vleague2 Seems like there may be. I'll remove this

libs/components/src/form-field/form-field.component.html
BryanCunningham: @vleague2 I double checked this. This has selector is what we need to accomplish the goal here

libs/components/src/input/input.directive.ts
BryanCunningham: @vleague2 Added a max-height for the non-readonly version. We'll let the readonly version expand as much as it needs to show all it's content

libs/components/src/input/input.directive.ts
vleague2: Are we able to migrate this one while we're in here, or no because of the `ngControl` stuff?

BryanCunningham: I can try it out. Seems these migrations usually open up a can of worms though

vleague2: Yeah. I think as much of these as we can do while we're in here is good, but if it's blowing the scope, we can keep it as a separate task

BryanCunningham: I vote for waiting. Mostly out of not wanting to complicate this PR any further

libs/components/src/input/input.directive.ts
BryanCunningham: Updated [here](https://github.com/bitwarden/clients/pull/20006/changes/c48ba6c28a750de9b111cd27c0e654a7ae8e0169)

libs/components/src/input/input.directive.ts
BryanCunningham: @vleague2 I went ahead and updated this [here](https://github.com/bitwarden/clients/pull/20006/changes/c35d0c47a92a09494b7b6d88032618068b0e9b3e)

libs/components/src/form-field/field-container.directive.ts
```diff
@@ -0,0 +1,76 @@
+import { booleanAttribute, computed, Directive, input } from "@angular/core";
+
+export type FieldContainerSize = "base" | "large";
+
+@Directive({
+  selector: "[bitFieldContainer]",
+  standalone: true,
```
vleague2: I think this is true by default so you shouldn't need this

libs/components/src/input/input.directive.ts
```diff
@@ -35,71 +39,78 @@ export function inputBorderClasses(error: boolean) {
...
+  protected readonly hasSuffix = input(false);
 
-  readonly showErrorsWhenDisabled = input<boolean>(false);
+  protected readonly showErrorsWhenDisabled = input<boolean>(false);
 
-  get labelForId(): string {
-    return this.id();
-  }
+  private readonly destroyRef = inject(DestroyRef);
+  private readonly controlEvent = signal<unknown>(null);
```
vleague2: I'm not following what this is for since it looks like a new addition overall?

libs/components/src/multi-select/multi-select.component.ts
```diff
@@ -87,13 +101,33 @@ export class MultiSelectComponent implements OnInit, BitFormFieldControl, Contro
   // eslint-disable-next-line @angular-eslint/prefer-output-emitter-ref
   @Output() onItemsConfirmed = new EventEmitter<any[]>();
```
vleague2: Me here to ask for more migrations 😂

libs/components/src/input/input.directive.ts
BryanCunningham: @vleague2 This was needed as part of the work to make 'required' a signal. I'm def open to suggestions if there is a more simple approach. It's possible I over complicated it

libs/components/src/multi-select/multi-select.component.ts
BryanCunningham: @vleague2 Updated. Also moved the component to use OnPush

libs/components/src/input/input.directive.ts
```diff
@@ -136,21 +169,17 @@ export class BitInputDirective implements BitFormFieldControl {
     return this.elementRef.nativeElement.readOnly;
   }
 
-  get standaloneInputClasses() {
+  protected get standaloneInputClasses() {
```
willmartian: Getters will not reliably trigger change detection in OnPush components; please update this and elsewhere to use computed signals.

(Even thought this is a directive and not OnPush itself, it's getters are public and are accessed by other OnPush components.

libs/components/src/input/input.directive.ts
willmartian: Yeah, sadly, this is necessary until the signal-based form API is released.

libs/components/src/form-field/field-container.directive.ts
willmartian, line 1: What are you thoughts on keeping this as is vs adding a `hideLabel` input to `bit-form-field` and using it within `bit-search` directly?

libs/components/src/form-field/form-field-control.ts
```diff
@@ -13,11 +13,11 @@ export type InputTypes =
...
 export abstract class BitFormFieldControl {
-  abstract ariaDescribedBy?: string;
+  abstract ariaDescribedBy: WritableSignal<string | undefined>;
   abstract id: Signal<string>;
   abstract labelForId: string;
-  abstract required: boolean;
-  abstract hasError: boolean;
+  abstract required: Signal<boolean>;
+  abstract hasError: Signal<boolean>;
   abstract error: [string, any];
```
willmartian: Non signal properties should be marked as readonly for better safety in `OnPush` contexts

libs/components/src/form-field/form-field.stories.ts
willmartian, line 1: Why rename `disabled` --> `inactive`? Aren't these still disabled?

libs/components/src/multi-select/multi-select.component.ts
```diff
@@ -175,34 +211,17 @@ export class MultiSelectComponent implements OnInit, BitFormFieldControl, Contro
 
   /**Implemented as part of NG_VALUE_ACCESSOR */
   protected onChange(items: SelectItemView[]) {
-    if (!this.notifyOnChange) {
-      return;
-    }
-
-    this.notifyOnChange(items);
+    this.selectedItems.set(items);
```
willmartian, line 189: Was this not being set before a bug?

libs/components/src/search/search.component.ts
willmartian, line 35: Could you update this component to OnPush while we are here?

libs/components/src/form-field/form-field-control.ts
willmartian: Is the implementation of this abstract class always the same?

If so, do we want to convert this to a host directive, and move the repeated logic there? We are duplicating a lot now with all of the controlEvent additions.

libs/components/src/input/input.directive.ts
```diff
@@ -35,71 +39,78 @@ export function inputBorderClasses(error: boolean) {
...
+    this.controlEvent();
+    return (
+      this.requiredInput() || (this.ngControl?.control?.hasValidator(Validators.required) ?? false)
+    );
+  });
 
-  readonly hasPrefix = input(false);
-  readonly hasSuffix = input(false);
+  protected readonly hasPrefix = input(false);
+  protected readonly hasSuffix = input(false);
```
willmartian: I don't think these are used anywhere and can be removed 🤔

libs/components/src/input/input.directive.ts
```diff
@@ -35,71 +39,78 @@ export function inputBorderClasses(error: boolean) {
...
 
-  readonly showErrorsWhenDisabled = input<boolean>(false);
+  protected readonly showErrorsWhenDisabled = input<boolean>(false);
```
willmartian: Weird feature to have 🤔

willmartian (changes requested):
Nice work thus far!

libs/components/src/form-field/field-container.directive.ts
BryanCunningham: @willmartian I'm open to doing that if we think that is the cleaner solution.

BryanCunningham: @willmartian There are some other minor issues with that though. We'd also need a way to style the prefix differently. In bit-form-field the prefix is rendered in a container with a gray background. That is not wanted in the search component

BryanCunningham: The more I think about it. This abstraction feels like the right way. Along with the prefix styling there are other wrinkles. Claude had some opinions about it as well:
- role="search" doesn't fit bit-form-field's structure. bit-search wraps everything in <form role="search"> for accessibility semantics. bit-form-field renders a <label> above a <div bitFieldContainer>, so there's no clean way to nest the <form> without breaking the label/input association or producing invalid HTML (<label> outside the <form> that wraps the input it references).
- `bit-form-field requires` a `BitFormFieldControl`. It uses `contentChild.required(BitFormFieldControl)`, so the search input would need to implement that full interface (ariaDescribedBy signal, hasError signal, etc.) just to satisfy the wrapper, gaining nothing from it since search doesn't show errors or hints.

libs/components/src/form-field/form-field.stories.ts
BryanCunningham: @willmartian Yep, still disabled in the technical sense. But, we've been migrating away from the term in documentation and token names. We've chosen to go with 'Inactive' in these cases as it's more inclusionary/less ableist

libs/components/src/input/input.directive.ts
BryanCunningham: @willmartian Yeah, agreed. It's used in one place currently. 

Unrelated, I do think a 'inactiveReasonText' input might be nice. Could provide more context as to why an input is inactive if it's not clear from the context. Or, even if we _think_ it's clear from the context, more info is always better IMHO

libs/components/src/multi-select/multi-select.component.ts
BryanCunningham: 🤔 It wasn't a bug. It was being assigned in the writeValue function previously. This change was made as part of an OnPush migration plan. It may not be necessary here. But, I also don't think it hurts anything by being here either. Open to thoughts about it though!

libs/components/src/form-field/form-field-control.ts
BryanCunningham: @willmartian Changed this to a host directive [here](https://github.com/bitwarden/clients/pull/20006/changes/db5cf5aee4d5db3fdce031f0a95086c3a375cdb5)

Let me know what you think of this approach

libs/components/src/input/input.directive.ts
BryanCunningham: Updated to a computed signal

libs/components/src/form-field/field-container.directive.ts
willmartian: > role="search" doesn't fit bit-form-field's structure

Not a problem, the form field could still be wrapped in a form.

>  the search input would need to implement that full interface 

Not a problem, since we could just use `bitInput`.

---

I think the main reason my suggestion isn't a slam dunk is, as you point out, the prefix styling. If we could use the same prefix styling as `bit-form-field`, that would yield some good code deduplication. However, I asked design about that, and they prefer the bespoke style for search, so I think we are good to leave this as is. Thanks for thinking about it with me!

libs/components/src/form-field/form-field-control.ts
```diff
@@ -12,15 +27,56 @@ export type InputTypes =
...
-  abstract spellcheck?: ModelSignal<boolean | undefined>;
-  abstract readOnly?: boolean;
-  abstract focus?: () => void;
+@Directive({
+  standalone: true,
+  host: {
+    "[attr.aria-describedby]": "ariaDescribedBy()",
+  },
+})
+export class BitFormFieldControl implements AfterViewInit {
```
willmartian: Can we rename this `BitFormFieldControlDirective`, and the file name as well?

libs/components/src/form-field/form-field-control.directive.ts
```diff
@@ -12,15 +27,56 @@ export type InputTypes =
...
+@Directive({
+  standalone: true,
+  host: {
+    "[attr.aria-describedby]": "ariaDescribedBy()",
+  },
+})
+export class BitFormFieldControl implements AfterViewInit {
+  protected readonly ngControl = inject(NgControl, { optional: true, self: true });
+  private readonly destroyRef = inject(DestroyRef);
+  private readonly controlEvent = signal<unknown>(null);
```
willmartian, line 44: Can you add a comment on this? I will forget what it is used for / why it is needed.

libs/components/src/input/input.directive.ts
```diff
@@ -29,129 +26,104 @@ export function inputBorderClasses(error: boolean) {
...
+export class BitInputDirective implements AfterViewInit {
+  private readonly ngControl = inject(NgControl, { optional: true, self: true });
+  private readonly elementRef = inject<ElementRef<HTMLInputElement>>(ElementRef);
+  private readonly parentFormField = inject(BitFormFieldComponent, { optional: true });
+  readonly formFieldControl = inject(BitFormFieldControl);
+
+  readonly id = input(`bit-input-${nextId++}`);
+
+  constructor() {
+    effect(() => this.formFieldControl.labelForId.set(this.id()));
```
willmartian: We can move this `id` input onto the directive as well, since it is duplicated between all components.

libs/components/src/input/input.directive.ts
```diff
@@ -29,129 +26,104 @@ export function inputBorderClasses(error: boolean) {
...
+    effect(() => this.formFieldControl.readOnly.set(this.elementRef.nativeElement.readOnly));
+  }
+
+  protected classList() {
```
willmartian: This should be a computed signal instead of a method

libs/components/src/input/input.directive.ts
```diff
@@ -29,129 +26,104 @@ export function inputBorderClasses(error: boolean) {
...
+  private readonly elementRef = inject<ElementRef<HTMLInputElement>>(ElementRef);
+  private readonly parentFormField = inject(BitFormFieldComponent, { optional: true });
+  readonly formFieldControl = inject(BitFormFieldControl);
+
+  readonly id = input(`bit-input-${nextId++}`);
+
+  constructor() {
+    effect(() => this.formFieldControl.labelForId.set(this.id()));
+    effect(() => this.formFieldControl.readOnly.set(this.elementRef.nativeElement.readOnly));
+  }
```
willmartian: This can also be moved to the directive, right? Also, in its current form, this will never update after the initial read. Effects only update if an internal signal is used. I'm not sure exactly what the correct workaround is, or if there is an API for this outside of signal-based forms, which are about to launch but haven't yet. Could you look into it?

libs/components/src/input/input.directive.ts
```diff
@@ -29,129 +26,104 @@ export function inputBorderClasses(error: boolean) {
...
   }
 
-  readonly id = input(`bit-input-${nextId++}`);
-
-  @HostBinding("attr.aria-describedby") ariaDescribedBy?: string;
-
-  @HostBinding("attr.aria-invalid") get ariaInvalid() {
-    return this.hasError ? true : undefined;
+  protected get ariaInvalid() {
+    return this.formFieldControl.hasError() ? true : undefined;
```
willmartian: This should be a computed signal instead of a getter

libs/components/src/select/select.component.ts
```diff
@@ -61,17 +66,28 @@ export class SelectComponent<T> implements BitFormFieldControl, ControlValueAcce
...
+        ?.searchInput()
+        .nativeElement.setAttribute(
+          "aria-describedby",
+          this.formFieldControl.ariaDescribedBy() ?? "",
+        );
+    });
+  }
+
+  ngAfterViewInit() {
+    // intentionally empty — shared reactive logic is in BitFormFieldControl host directive
```
willmartian: I think we can remove this then?

libs/components/src/input/input.directive.ts
BryanCunningham: I updated readonly to be a signal on the host directive. [This work?](https://github.com/bitwarden/clients/pull/20006/commits/dfcd24fa34f5a592f881aa784713e7942d71240d)

vleague2 (approved):

jengstrom-bw (approved):
Vault items look good

willmartian (approved):

JaredScar (approved):

PR #20008, opened 2026-04-06
[PM-34230] Blumira Integration using HEC

voommen-livefront:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-34230

## 📔 Objective

1. Create a blumira integration using the skill `create-hec-event-integration`
2. Rename Huntress specific components to `connect-via-hec-token`
3. Created a new property to called `urlHelperLinkText` to configure prompts for each integration

## 📸 Screenshots

**Blumira in Light mode**
image

**Blumira connection dialog with helper text**
image

**Blumira in Dark mode**
Screenshot 2026-04-01 at 2 10 21 PM


bitwarden_license/bit-web/src/app/dirt/organization-integrations/integration-card/integration-card.component.ts
```diff
@@ -362,24 +366,25 @@ export class IntegrationCardComponent implements AfterViewInit, OnDestroy {
...
 
-  async saveHuntress(result: HuntressConnectDialogResult) {
-    // Huntress uses "Splunk" scheme for HEC protocol compatibility
+  async saveConnectViaHecTokenIntegration(result: ConnectViaHecTokenDialogResult) {
+    // create the Hec configuration
     const config = OrgIntegrationBuilder.buildHecConfiguration(
       result.url,
       result.token,
-      OrganizationIntegrationServiceName.Huntress,
+      this.integrationSettings().name as OrganizationIntegrationServiceName,
```
Banrion, line 374: ❓ I'm curious about the what differences between the `Integration` and `OrganizationIntegration`. They look very similar.

In regards to this line of code, it looks like the `serviceName: OrganizationIntegrationServiceName` is provided through the `organizationIntegration` property on an `Integration`. Is the integration `name` and organization  `serviceName` ever different? Using the `serviceName` would remove the need for typecasting. It's a minor improvement, but I'm asking to know if the `Integration` and `OrganizationIntegration` types might be able to be consolidated into one in the future.

Banrion (approved):
The Huntress to Hec abstraction refactoring looks good!

bitwarden_license/bit-web/src/app/dirt/organization-integrations/integration-card/integration-card.component.ts
voommen-livefront: `Integration` is an array of integrations, which could be either `event-integrations`, `user-provisioning`, `single-signon` and so on. It contains properties that define an "integration".
`OrganizationIntegration` is the configuration of an integration, which is a configuration and template (template is what the message that will be sent to Datadog, huntress, Crowdstrike, etc) and this comes from the database. Not all Integrations have OrganizationIntegration.
Also, Organization Integration (configuration and template) are in the database as `dbo.OrganizationIntegration` (Configuration) and `dbo.OrganizationIntegrationConfiguration` (template)

PR #20010, opened 2026-04-06
[CL-999] Update tabs

lxiong-livefront:
## 🎟️ Tracking

[JIRA](https://bitwarden.atlassian.net/browse/CL-999?atlOrigin=eyJpIjoiMzUzMmVlODE0YWQ3NDY3MGJmOTJiNWMwZGZlYjgzNGYiLCJwIjoiaiJ9)

## 📔 Objective

Updated horizontal tab styles and added responsive behavior
- Check 'Overflow Tabs' story for responsive behavior 
- Responsive behavior is implemented in `tab-group.component.ts`

NOTE FOR CONSUMER TEAM CODE CHANGES: 
Changes to consumer code were largely to remove usage of the `disablePadding` input. As the tabs no longer apply padding within the group component. 

## 📸 Screenshots

Screenshot 2026-04-09 at 12 52 03 PM

https://github.com/user-attachments/assets/69e3db54-8259-4280-9a51-d31483b6e3c2


libs/components/src/tabs/shared/tab-list-item.directive.ts
```diff
@@ -46,53 +51,40 @@ export class TabListItemDirective implements FocusableOption {
...
       "tw-border-transparent",
       "tw-border-solid",
       "tw-bg-transparent",
-      "hover:tw-underline",
-      "focus-visible:tw-z-10",
-      "focus-visible:tw-outline-none",
-      "focus-visible:tw-ring-2",
-      "focus-visible:tw-ring-primary-600",
+      "tw-outline-none",
+      "tw-group",
```
vleague2: Can you give this a name? We run into styling conflicts when more than one component is using `tw-group`. (Example -- `tw-group/tab-list-item`, usage in children as selector `group-focus-visible/tab-list-item:tw-whatever`)

libs/components/src/tabs/tab-group/tab-group.component.ts
```diff
@@ -64,55 +75,112 @@ export class TabGroupComponent implements AfterContentChecked, AfterViewInit {
...
+    }
+
+    return undefined;
+  });
+
   /** The index of the active tab. */
   // TODO: Skipped for signal migration because:
   //  Accessor inputs cannot be migrated as they are too complex.
   // FIXME(https://bitwarden.atlassian.net/browse/CL-903): Migrate to Signals
   // eslint-disable-next-line @angular-eslint/prefer-signals
```
vleague2: If we have time to solve this TODO, that would be cool

libs/components/src/tabs/shared/tab-list-item.directive.ts
```diff
@@ -46,53 +51,40 @@ export class TabListItemDirective implements FocusableOption {
...
 
   get baseClassList(): string[] {
     return [
       "tw-block",
       "tw-relative",
-      "tw-py-2",
-      "tw-px-4",
+      "tw-shrink-0",
+      "tw-whitespace-nowrap",
+      "tw--mb-px",
```
vleague2: Extra dash?

libs/components/src/tabs/tab-group/tab-group.component.ts
```diff
@@ -64,55 +75,112 @@ export class TabGroupComponent implements AfterContentChecked, AfterViewInit {
...
+  /** Cached tab widths measured before any hiding, keyed by index. */
+  private readonly tabWidths = signal<number[]>([]);
+
+  /** Cached More button width measured before any hiding. */
+  private readonly moreButtonWidth = signal(0);
+
+  /** Whether the tab list has been rendered. Used to hide or display tab list container, preventing layout shifts. */
+  protected readonly tabListRendered = signal(false);
+
+  /** Determines the index at which tabs start overflowing into the "More" menu. */
```
vleague2: The wrench in this will be that we always want to display the active tab, even if it's after the index where the overflow will begin

libs/components/src/tabs/shared/tab-list-container.directive.ts
```diff
@@ -6,7 +6,7 @@ import { Directive } from "@angular/core";
 @Directive({
   selector: "[bitTabListContainer]",
   host: {
-    class: "tw-inline-flex tw-flex-wrap tw-leading-5",
+    class: "tw-inline-flex tw-flex-nowrap tw-w-full tw-leading-5 tw-gap-6",
```
vleague2: I wonder if we should export a const for the gap so that you don't have to measure it in the `tab-group` math? Maybe one less thing to measure? Ooh but maybe you can't since it's rem?

vleague2 (commented):
Overall this is making sense to me, couple random thoughts

libs/components/src/tabs/shared/tab-list-item.directive.ts
lxiong-livefront: That's what I thought at first too, but it's actually intentional (-1px) margin

libs/components/src/tabs/shared/tab-list-container.directive.ts
lxiong-livefront: Exporting a const for the gap would be ideal! I can see if there's similar cases of this elsewhere in the codebase. If anything, might have to switch from rem to px?

libs/components/src/tabs/tab-group/tab-group.component.ts
lxiong-livefront: The comment scared me away, but I can revisit this 😅

libs/components/src/tabs/tab-group/tab-group.component.ts
lxiong-livefront: Yeah, that part makes all of this more complex... hoping to discuss more in dev sync with Emily as well. The logic here will most likely need to be updated to accommodate that behavior

libs/components/src/tabs/shared/tab-list-item.directive.ts
lxiong-livefront: (taken from pre-existing `activeClassList` line 86)

libs/components/src/tabs/shared/tab-list-item.directive.ts
lxiong-livefront: Named in [427cda8](https://github.com/bitwarden/clients/pull/20010/commits/427cda87c46193a88e6ed85057d65e10a48b96f0)

libs/components/src/tabs/shared/tab-list-container.directive.ts
lxiong-livefront: Exported a const for this gap in [427cda8](https://github.com/bitwarden/clients/pull/20010/commits/427cda87c46193a88e6ed85057d65e10a48b96f0). I do prefer this approach over needing to measure it

libs/components/src/tabs/tab-group/tab-group.component.ts
lxiong-livefront: Done in [522f091](https://github.com/bitwarden/clients/pull/20010/commits/522f0918751f191e94ba64c19e0bf4bb9c58c0ff)

libs/components/src/tabs/shared/tab-list-item.directive.ts
```diff
@@ -37,7 +32,6 @@ export class TabListItemDirective implements FocusableOption {
     this.elementRef.nativeElement.click();
   }
 
-  @HostBinding("class")
   get classList(): string[] {
```
vleague2: I wonder if this should be a computed signal now that it's not a host binding, since it accesses the `this.active()` signal

(Huh probably `textColorClassList` too.)

libs/components/src/tabs/tab-group/tab-group.component.ts
lxiong-livefront: This logic has been updated in [67e74d0](https://github.com/bitwarden/clients/pull/20010/commits/67e74d049e54b9287911ffd17d76c56646736fd5)

libs/components/src/tabs/shared/tab-list-container.directive.ts
```diff
@@ -1,12 +1,15 @@
 import { Directive } from "@angular/core";
 
+/** Gap between tab items in pixels */
+export const tabListContainerGap = 24;
```
vleague2: We should probably do like `TAB_LIST_CONTAINER_GAP` to represent that it's a static shared constant

vleague2 (commented):
Random thoughts, haven't been able to do another full review yet

libs/components/src/tabs/shared/tab-list-item.directive.ts
lxiong-livefront: These are both using `disabled` which is still a `@Input` so unfortunately can't update these to computed signals.

libs/components/src/tabs/shared/tab-list-container.directive.ts
lxiong-livefront: renamed in [6708ad2](https://github.com/bitwarden/clients/pull/20010/commits/6708ad2c152020e98b9dca47fdeb859ada39aa95)

libs/components/src/tabs/shared/tab-list-item.directive.ts
vleague2: Boooooo okay makes sense!

libs/components/src/tabs/shared/tab-header.component.ts
```diff
@@ -9,7 +9,7 @@ import { Component } from "@angular/core";
   selector: "bit-tab-header",
   host: {
     class:
-      "tw-min-h-16 tw-ps-4 tw-bg-background-alt tw-flex tw-items-end tw-border-0 tw-border-b tw-border-solid tw-border-secondary-100",
+      "tw-min-h-16 tw-mx-6 tw-flex tw-items-end tw-border-0 tw-border-b tw-border-solid tw-border-secondary-100",
```
BryanCunningham: Do we always want this margin? Feels like that should be controlled by the consumer

libs/components/src/tabs/shared/tab-list-item.directive.ts
```diff
@@ -46,53 +45,34 @@ export class TabListItemDirective implements FocusableOption {
...
+    "tw-block",
+    "tw-relative",
+    "tw-shrink-0",
+    "tw-whitespace-nowrap",
+    "tw--mb-px",
+    "tw-pb-3",
+    "tw-text-sm",
+    "tw-font-medium",
+    "tw-transition",
+    "tw-border-0",
```
BryanCunningham: Probably don't need this `tw-border-0`. Think the following `tw-border-b-2` only applies border-bottom. But, I could be wrong

libs/components/src/tabs/tab-group/tab-group.component.html
```diff
@@ -1,32 +1,70 @@
+@let tabLabelContentClasses =
+  "tw-flex tw-items-center tw-gap-1.5 tw-rounded group-focus-visible/tab-list-item:tw-ring-2 group-focus-visible/tab-list-item:tw-ring-border-focus";
+
+<ng-template #tabLabelContent let-tab="tab" let-truncate="truncate">
```
BryanCunningham: Can these be moved out of the template declaration like this? 
```html
@let tab() = tab;
@let truncate() = truncate;
```

libs/components/src/tabs/tab-group/tab-group.component.html
```diff
@@ -1,32 +1,70 @@
...
+      bitTabListItem
+      type="button"
+      role="tab"
+      [hidden]="sortedTabs().overflow.length === 0"
+      [attr.aria-hidden]="sortedTabs().overflow.length === 0 ? 'true' : null"
+      [attr.tabindex]="sortedTabs().overflow.length === 0 ? null : -1"
+      [bitMenuTriggerFor]="overFlowTabsMenu"
+    >
+      <span [class]="tabLabelContentClasses">
+        More
```
BryanCunningham: We'll need to add this string to messages.json to be localized

libs/components/src/tabs/shared/tab-header.component.ts
lxiong-livefront: I asked Emily about this and she said to update it to 24px. I hear you though, but I think it's good to have a default margin especially for tabs in dialogs that have a dialog heading. If anything, the consumer can override if needed?

libs/components/src/tabs/tab-group/tab-group.component.html
lxiong-livefront: These are both declared within the `for` loop and depend on the specific index of the tab

libs/components/src/tabs/shared/tab-header.component.ts
BryanCunningham: I can see this causing visual regressions. We only had the padding start previously so that we had this little gap on the left of the first tab. Now, these will appear 8px more indented from the left of whatever container they are in. Have we checked the usage in the app? Does this cause anything weird to happen?

image


Now we may see something like this where the tabs are not visually aligned with their surrounding content

image

lxiong-livefront: Removed spacing from tab group as well as tab body content and updated in 'Header' and 'Dialog' [eb7b19b](https://github.com/bitwarden/clients/pull/20010/commits/eb7b19b084ad0feb66581ce20c910175fcb732bc)

libs/components/src/tabs/tab-nav-bar/tab-link.component.ts
```diff
@@ -37,34 +58,52 @@ export class TabLinkComponent implements FocusableOption, AfterViewInit {
   };
 
   readonly route = input<string | any[]>();
+  readonly label = input<string>();
+  readonly berryValue = input<number>();
+  readonly trailingIcon = input<BitwardenIcon>();
+
+  // eslint-disable-next-line @angular-eslint/prefer-signals
+  @ContentChild(TabLabelDirective) templateLabel?: TabLabelDirective;
```
vleague2: Is there a blocker for using the updated signal selector here?

lxiong-livefront: No blocker here! Updated in [4e36d94](https://github.com/bitwarden/clients/pull/20010/commits/4e36d944f5cff491de6c6a35899a08aade4eb829)

libs/components/src/tabs/tab-group/tab-group.component.ts
```diff
@@ -117,6 +117,15 @@ export class TabGroupComponent implements AfterContentChecked, AfterViewInit {
 
   private readonly _selectedIndex = signal<number | null>(null);
 
+  /**
+   * Guards the effect's "no active tab" emission path against first render.
+   * `ngAfterContentChecked` defers `isActive` updates to a microtask, so the
+   * effect fires with all tabs inactive before they've been initialized — emitting
+   * a spurious `selectedTabChange`. A plain boolean avoids re-triggering the effect
```
vleague2: Can we use a different word than `spurious` 😂

libs/components/src/tabs/tab-group/tab.component.ts
```diff
@@ -1,28 +1,37 @@
...
   },
+  changeDetection: ChangeDetectionStrategy.OnPush,
 })
 export class TabComponent implements OnInit {
+  readonly id = nextTabId++;
+
   readonly disabled = input(false);
   readonly textLabel = input("", { alias: "label" });
+  readonly berryValue = input<number>();
+  readonly trailingIcon = input<BitwardenIcon>();
```
willmartian: This should be `endIcon`, to match the naming scheme established in buttons and links.

libs/components/src/tabs/tabs.mdx
```diff
@@ -30,12 +45,51 @@ Tabs can be used in dialogs to separate related content.
 
 <Canvas of={dialogStories.TabContent} />
 
+## Leading and Trailing Icons in Tabs
+
+Tabs can have a leading and trailing icon as well as a berry count. A trailing icon can be added
+using the `[trailingIcon]` input. A berry count can be added using the `[berryValue]` input. Leading
+icons can be added by using the `[bitTabLabel]` directive.
+
```
willmartian, line 55: Can/should we add a `startIcon` input too? 

And then maybe remove `bitTabLabel`? Or still leave it as a backup option?

libs/components/src/tabs/shared/tab-list-container.directive.ts
```diff
@@ -1,12 +1,15 @@
...
+export const TAB_LIST_CONTAINER_GAP = 24;
+
 /**
  * Directive used for styling the container for bit tab labels
  */
 @Directive({
   selector: "[bitTabListContainer]",
   host: {
-    class: "tw-inline-flex tw-flex-wrap tw-leading-5",
+    class: "tw-inline-flex tw-flex-nowrap tw-w-full tw-leading-5 tw-gap-6", // Ensure that `tw-gap-` class and `TAB_LIST_CONTAINER_GAP` value are equivalent
```
willmartian: Rather than a comment ensuring that they stay in sync (which will be forgotten about or deleted by AI at some point), would it make sense to use a CSS variable instead?

```
host: {
    '[style.gap.px]': 'TAB_LIST_CONTAINER_GAP',
  }
```

```
protected readonly TAB_LIST_CONTAINER_GAP = 24;
```

libs/components/src/tabs/tabs.stories.ts
```diff
@@ -130,14 +136,12 @@ export const NavigationTabs: Story = {
...
-        <bit-tab-link [route]="['active']">Active</bit-tab-link>
+        <bit-tab-link [route]="['active']" [berryValue]="3" [trailingIcon]="'bwi-check-circle'">Active</bit-tab-link>
         <bit-tab-link [route]="['item-2']">Item 2</bit-tab-link>
         <bit-tab-link [route]="['item-3']">Item 3</bit-tab-link>
         <bit-tab-link [route]="['item-with-child-counter']">
           Item With Counter
-          <div slot="end" class="tw-ps-2 tw-text-muted">
-            42
-          </div>
+          <div slot="end" class="tw-ps-2 tw-text-muted">42</div>
```
willmartian, line 141: Is this end slot still supported, or has it been supplanted by the new inputs? Should it be removed either now or in a future ticket?

Image

libs/components/src/tabs/tab-nav-bar/tab-nav-bar.component.html
```diff
@@ -1,5 +1,54 @@
...
+        <bit-icon name="bwi-angle-down" class="tw-text-base" />
+      </span>
+    </button>
   </nav>
 </bit-tab-header>
+
+<bit-menu #overflowTabsMenu>
+  @for (i of sortedTabs().overflow; track i) {
+    @let tab = tabLabels()[i];
+    <button
```
willmartian: Should these be anchors?

libs/components/src/tabs/tab-nav-bar/tab-link.component.ts
```diff
@@ -37,34 +58,51 @@ export class TabLinkComponent implements FocusableOption, AfterViewInit {
...
   @HostListener("keydown", ["$event"]) onKeyDown(event: KeyboardEvent) {
     if (event.code === "Space") {
       this.tabItem().click();
     }
   }
 
   get active() {
-    return this.routerLinkActive()?.isActive ?? false;
+    return this.isActive();
   }
```
willmartian: Do we still need this getter?

libs/components/src/tabs/tab-nav-bar/tab-link.component.ts
```diff
@@ -1,33 +1,54 @@
...
+  readonly elementRef = inject(ElementRef);
 
   readonly tabItem = viewChild.required(TabListItemDirective);
   readonly routerLinkActive = viewChild.required<RouterLinkActive>("rla");
+  private readonly labelText = viewChild<ElementRef>("labelText");
+
+  /** Display text for the overflow menu. Uses `label` input if provided, otherwise reads projected text content. */
+  readonly displayText = computed(
+    () => this.label() ?? this.labelText()?.nativeElement.textContent?.trim() ?? "",
+  );
```
willmartian, line 59: I'm not following here. So we support both a label and projected content for the text? Why have two options?

libs/components/src/tabs/tab-nav-bar/tab-nav-bar.component.html
BryanCunningham: @willmartian I would think they should be buttons, we aren't navigating to a new page so buttons make more sense probably

BryanCunningham: Well... these do 'navigate', it seems... I'm not sure then

libs/components/src/tabs/tabs.mdx
BryanCunningham: There are a couple usages of `bitTabLabel` in the codebase currently. They seem to contain both an icon and some text and a count' We could maybe refactor these to use a `startIcon` along with a `berryCount`.

libs/components/src/tabs/tab-nav-bar/tab-nav-bar.component.html
willmartian: I think it would be proper, `[routerLink]` is essentially `href`, so having it on anything other than an anchor smells to me.

libs/components/src/tabs/tab-nav-bar/tab-link.component.ts
BryanCunningham: Nope, removed

libs/components/src/tabs/tabs.mdx
BryanCunningham: I've removed this and updated the consumer code usages of it

libs/components/src/tabs/shared/tab-list-container.directive.ts
BryanCunningham: Good call. Updated

libs/components/src/tabs/tab-nav-bar/tab-nav-bar.component.html
BryanCunningham: I made these anchor tags.

libs/components/src/tabs/tab-nav-bar/tab-link.component.ts
BryanCunningham: For posterity: We met and chatted about this. Will be resolved in a followup ticket

libs/components/src/shared/base-button.directive.ts
```diff
@@ -231,7 +231,7 @@ export const getButtonColorStyles = ({
   const baseDisabledStyles = [
     "aria-disabled:!tw-text-fg-disabled",
     "hover:!tw-text-fg-disabled",
-    "aria-disabled:tw-cursor-not-allowed",
+    "aria-disabled:tw-cursor-default",
```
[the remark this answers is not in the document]
BryanCunningham: Deliberate change. Will address other usages of cursor not allowed later

apps/web/src/app/tools/send/shared/send-success-drawer-dialog.component.html
```diff
@@ -1,4 +1,4 @@
-<bit-dialog dialogSize="large" disablePadding="false" background="alt">
+<bit-dialog dialogSize="large" background="alt">
```
adudek-bw, line 1: Dumb question as someone still learning a lot about front end stuff: since disablePadding by default is false, would this not be functionally the same?

BryanCunningham: @adudek-bw Not a dumb question! And, yes, it would have been functionally the same as just not including it at all 😄 

That input just no longer exists so, I just ripped it out of here

adudek-bw (approved):
LGTM :)

apps/web/src/app/tools/send/shared/send-success-drawer-dialog.component.html
willmartian: That is correct! Removing this here is no functional difference and not strictly required. It is just less confusing to go ahead and remove it, since `disablePadding` is no longer an input.

Now, _if_ this was bound as a dynamic value, `[disablePadding]="false"`, then removing it would be _mandatory_, because Angular typechecks bindings against available inputs, so this would cause a type error.

willmartian (approved):

apps/web/src/locales/en/messages.json
```diff
@@ -98,15 +98,6 @@
   "cannotRestoreAccessError": {
     "message": "Cannot restore organization access"
   },
-  "allApplicationsWithCount": {
```
Banrion, line 101: ❓ Just asking a general question if anyone happens to know: I know that messages for keys can't be changed. If a key is removed and then later added again, does it pick up the new message if it's different?

Banrion (approved):
✅ DIRT files look good!

apps/web/src/locales/en/messages.json
BryanCunningham: @Banrion Hmm... I hadn't really thought of that to be honest. Do you think we should keep in here just in case? Even if it's no longer used?
@willmartian Any context/guidance here?

willmartian: I am also unsure. @djsmith85 could you weigh in?

cd-bitwarden (approved):
LGTM

apps/web/src/locales/en/messages.json
djsmith85: Sure, when a PR merges to `main` we push to Crowdin and that deletes all existing translations in Crowdin for that key. If the key gets re-introduced by another PR, then it'll treat it as new. Once a week a sync PR is created that pulls the changes from Crowdin, which will remove/replace all usages in the translation files, other than `en/messages.json`

willmartian: Sounds like our change here is fine then. Thanks @djsmith85!

Banrion: Thanks for the replies! That's how I suspected it worked, but wanted to confirm.

r-tome (approved):

PR #20021, opened 2026-04-06
[CL-989] updating popover component

iivins-livefront:
**Overview:**

This PR updates the `bit-popover` component to match the design specs and modernize it's angular usage. Migrating to semantic color tokens, adding a fixed-width border, and introducing two new optional content slots ( header and footer ) for richer layouts. The component is also modernized to use OnPush change detection and a signal-based closed output, with no breaking changes required for existing consumers.

**Key Changes:**
- Updated `popover.component.html` and styles to match semantic tokens
- Added an optional `bit-popover-header` for image/media content above the title, no empty space when unused
- Added an optional `bit-popover-footer` for action buttons below body content, no empty space when unused
- Added `PopoverHeaderComponent` and `PopoverFooterComponent`.
- Migrated `CoachmarkComponent` to use the new `bitPopoverFooter` for its step counter and Back/Next buttons, web and browser
- Modernized `PopoverComponent` to `ChangeDetectionStrategy.OnPush` with signal based inputs/outputs
- All 6 existing consumer templates continue to work without modification
- Updated `popover.stories.ts` to reflect the agreed story list
- Title and body text updated to use `bitTypography` with new semantic color tokens

**Summary:**

The popover component has been visually updated and structurally extended with two new optional content slots. Angular modernization is applied throughout and the `CoachmarkComponent` ( only consumer that needed to get update ) got updated

**Screenshots:**

Screenshot 2026-04-06 at 5 34 54 PM
Screenshot 2026-04-06 at 5 35 17 PM
Screenshot 2026-04-06 at 5 35 34 PM
Screenshot 2026-04-06 at 5 35 46 PM


shane-melton (dismissed):
Vault change looks good!

libs/components/src/popover/popover.component.html
```diff
@@ -5,26 +5,41 @@
...
-    <div class="tw-overflow-hidden tw-rounded-xl tw-shadow-lg">
-      <div
-        class="tw-relative tw-z-20 tw-w-72 tw-break-words tw-bg-background tw-pb-4 tw-pt-2 tw-text-main"
-      >
-        <div class="tw-me-2 tw-flex tw-items-start tw-justify-between tw-gap-4 tw-ps-4">
-          <h2 bitTypography="h5" class="tw-font-medium tw-mt-1">
+    <div
+      class="tw-overflow-hidden tw-rounded-xl tw-shadow-lg tw-border tw-border-solid tw-border-border-base"
+    >
+      <div class="tw-relative tw-z-20 tw-w-[352px] tw-break-words tw-bg-background tw-text-fg-body">
```
BryanCunningham: @iivins-livefront This width feels very specific and a little bit magic. What's the reasoning for this specific number?

libs/components/src/popover/popover.component.html
```diff
@@ -5,26 +5,41 @@
...
+        @if (headerSlot()) {
+          <div
+            class="tw-h-[160px] tw-mt-2 tw-mx-2 tw-overflow-hidden tw-rounded-lg tw-bg-bg-secondary tw-flex tw-items-center tw-justify-center"
```
BryanCunningham: @iivins-livefront Do we want to set an actual height on this? Or, does a max-height make more sense here?

libs/components/src/popover/popover.component.css
```diff
@@ -1,20 +1,25 @@
...
 }
 
 .bit-popover-right-start .bit-popover-arrow,
 .bit-popover-left-start .bit-popover-arrow {
   @apply tw-top-6 -tw-translate-y-1/2;
 }
 
+.bit-popover-has-header.bit-popover-right-start .bit-popover-arrow,
+.bit-popover-has-header.bit-popover-left-start .bit-popover-arrow {
+  @apply tw-top-[184px];
```
BryanCunningham: @iivins-livefront This number feels kind of magic. What if the popover isn't 184px tall? Can we find a more future proof way to position this?

libs/components/src/popover/popover.component.html
```diff
@@ -5,26 +5,32 @@
...
+              (click)="closed.emit()"
+              size="small"
+            ></button>
+          </div>
+          <div class="tw-flex tw-flex-col tw-gap-6">
+            <div class="tw-flex tw-flex-col tw-gap-3">
+              <div bitTypography="h4" noMargin class="tw-text-fg-heading">
+                {{ title() }}
+              </div>
+              <p bitTypography="body2" noMargin class="tw-text-fg-body">
```
vleague2: I don't think this `noMargin` is doing anything. I'm seeing a lot of space underneath

Image

libs/components/src/popover/popover.stories.ts
```diff
@@ -296,28 +414,26 @@ export const LeftCenter: Story = {
...
   }),
   play: async (context) => {
     const canvasEl = context.canvasElement;
     const button = getByRole(canvasEl, "button");
-
     await userEvent.click(button);
   },
 };
+
 export const LeftEnd: Story = {
```
vleague2, line 417: We will either need to not use the full header/footer setup or give ourselves more space to render these stories, because now the popover is too tall to show in the end position 

Image

libs/components/src/popover/popover.stories.ts
```diff
@@ -465,29 +569,26 @@ export const AboveStart: Story = {
...
   }),
   play: async (context) => {
     const canvasEl = context.canvasElement;
     const button = getByRole(canvasEl, "button");
-
     await userEvent.click(button);
   },
 };
 
 export const AboveCenter: Story = {
```
vleague2, line 562: All of the above stories can't fit above the trigger

Image

libs/components/src/popover/popover.stories.ts
```diff
@@ -533,205 +631,12 @@ export const AboveEnd: Story = {
...
   }),
   play: async (context) => {
     const canvasEl = context.canvasElement;
     const button = getByRole(canvasEl, "button");
-
     await userEvent.click(button);
   },
 };
-
-export const SpotlightTour: Story = {
```
vleague2, line 547: Why remove this story?

libs/components/src/popover/popover.component.html
```diff
@@ -5,26 +5,32 @@
...
+        <ng-content select="bit-popover-header"></ng-content>
+        <div class="tw-p-6">
+          <div class="tw-absolute tw-top-2 tw-right-2">
+            <button
+              type="button"
+              bitIconButton="bwi-close"
+              [label]="'close' | i18n"
+              (click)="closed.emit()"
+              size="small"
+            ></button>
```
vleague2: I wonder if this button should go above the new header slot in the DOM order?

libs/components/src/popover/popover.component.html
```diff
@@ -5,26 +5,32 @@
     class="tw-relative"
     role="dialog"
     aria-modal="true"
-    [attr.aria-label]="title()"
+    [attr.aria-label]="title() || null"
```
vleague2: Not sure if this will do anything because the title is defaulting to an empty string. We should probably solve that problem in the component file

libs/components/src/popover/popover.component.html
```diff
@@ -5,26 +5,32 @@
...
+              bitIconButton="bwi-close"
+              [label]="'close' | i18n"
+              (click)="closed.emit()"
+              size="small"
+            ></button>
+          </div>
+          <div class="tw-flex tw-flex-col tw-gap-6">
+            <div class="tw-flex tw-flex-col tw-gap-3">
+              <div bitTypography="h4" noMargin class="tw-text-fg-heading">
+                {{ title() }}
```
vleague2, line 29: Not actually helpful thought -- I feel like in most of our other usages of a header component, the title goes in the header. I know that's not how the design works for this one, but it is confusing to me to have a header that does not contain the main title

libs/components/src/popover/popover.component.html
BryanCunningham: @vleague2 Removed the `p` tag and the `noMargin` input that was useless :)

libs/components/src/popover/popover.stories.ts
BryanCunningham: @vleague2 Not sure how/why this was removed. Will add it back

libs/components/src/popover/popover.component.html
BryanCunningham: Yeah, it is a little bit weird. I agree. Not sure what else we'd call this section though

libs/components/src/popover/popover.component.html
BryanCunningham: @vleague2 I spoke to Emily about this. I think we are going to swap button variants in the event that the `bit-popover-header` is actually rendered. Solves for the possible issue where the button would be hard to see

libs/components/src/popover/popover.stories.ts
BryanCunningham: @vleague2 Used the shorted content on all these and solved this issue. Thanks for the heads up

libs/components/src/popover/popover.stories.ts
BryanCunningham: Used shorter content/no header or footer on these to fix

libs/components/src/popover/popover.component.html
vleague2: Do we want to do that in this ticket or in the follow-up we made?

BryanCunningham: I'll just do it in this PR real quick

apps/web/src/app/vault/components/coachmark/coachmark.component.html
```diff
@@ -17,8 +17,7 @@
     </a>
   }
 
-  <!-- Footer with navigation -->
-  <div class="tw-flex tw-items-center tw-justify-between tw-pt-2">
+  <bit-popover-footer slot="footer" class="tw-flex tw-w-full tw-items-center tw-justify-between">
```
vleague2: I think you don't need the footer slot now that it is selecting on the component selector

libs/components/src/popover/popover.stories.ts
```diff
@@ -66,15 +68,24 @@ export default {
...
+const defaultContent = /*html*/ `
+  <bit-popover [title]="'We\\'ve just released a new dashboard'" #myPopover>
+    A new and improved dashboard is now live! Enjoy a smoother, more intuitive experience.
+  </bit-popover>
+`;
+
+const withMediaAndFooterContent = /*html*/ `
+  <bit-popover [title]="'We\\'ve just released a new dashboard'" #myPopover>
+    <bit-popover-header>
+      <img slot="media" src="https://placehold.co/352x160/e2e8f0/64748b?text=Media+Slot" alt="" />
```
vleague2: This is convenient but I wonder if we should just do a div with some css for this slot rather than make an external request, or use an existing svg from our lib

libs/components/src/popover/popover.stories.ts
```diff
@@ -95,25 +106,59 @@ export const Default: Story = {
...
+        <button
+          type="button"
+          class="tw-border-none tw-bg-transparent tw-text-primary-600 tw-p-0"
+          [bitPopoverTriggerFor]="myPopover"
+          #triggerRef="popoverTrigger"
+          aria-label="Open popover"
+          title="Open popover"
+          bitLink
+        >
+          <i class="bwi bwi-question-circle"></i>
```
vleague2: Maybe we can update this to `bit-icon` while we're in here (it's in a couple places)

vleague2 (approved):

jaasen-livefront (approved):
Very nice!

PR #20022, opened 2026-04-06
[PM-34135] User preferences service - Innovation Sprint

shane-melton:
## 🎟️ Tracking

[PM-34135](https://bitwarden.atlassian.net/browse/PM-34135)

## 📔 Objective

> [!NOTE]
> This is an early PoC and will require changes/improvements around the encryption implementation and migrating it over to the SDK to follow current best practices. This should NOT be merged / released as-is.

Introduce a `PreferenceSyncService` that subscribes to registered state provider settings and syncs them to the server in an encrypted format when the setting is enabled. During sync, when the new sync preferences settings is enabled, decrypt and then apply the preferences in the synced user data.

## 📸 Screenshots

https://github.com/user-attachments/assets/9f0f07da-7c61-4085-b92d-d379fa12a11e

[PM-34135]: https://bitwarden.atlassian.net/browse/PM-34135?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/browser/src/_locales/en/messages.json
+++ b/apps/browser/src/_locales/en/messages.json
@@ -443,7 +443,7 @@
     "message": "Sync settings to other devices"
   },
   "syncSettingsDescription": {
-    "message": "When available settings will sync wherever you're logged in"
+    "message": "When available, settings will sync wherever you're logged in"
   },
   "syncNow": {
     "message": "Sync now"
--- a/apps/desktop/src/locales/en/messages.json
+++ b/apps/desktop/src/locales/en/messages.json
@@ -2093,7 +2093,7 @@
     "message": "Sync settings to other devices"
   },
   "syncSettingsDescription": {
-    "message": "When available settings will sync wherever you're logged in"
+    "message": "When available, settings will sync wherever you're logged in"
   },
   "enableMenuBar": {
     "message": "Show menu bar icon"
--- a/apps/web/src/locales/en/messages.json
+++ b/apps/web/src/locales/en/messages.json
@@ -2539,7 +2539,7 @@
     "message": "Sync settings to other devices"
   },
   "syncSettingsDescription": {
-    "message": "When available settings will sync wherever you're logged in"
+    "message": "When available, settings will sync wherever you're logged in"
   },
   "language": {
     "message": "Language"


libs/common/src/platform/sync/preferences/preference-sync.service.ts
```diff
@@ -0,0 +1,306 @@
...
+
+  private async encryptBlob(prefs: SyncedPreferences, userId: UserId): Promise<string | undefined> {
+    const key = await this.getUserKey(userId);
+    if (key == null) {
+      this.logService.warning("PreferenceSyncService: no user key available for encryption");
+      return undefined;
+    }
+
+    const json = JSON.stringify(prefs);
+    const encString = await this.encryptService.encryptString(json, key);
```
quexten, line 274: Question: Why not do this in the SDK with data blobs? (Considering this PR has a ticket attached I assume it's meant for merge).

libs/common/src/platform/sync/preferences/preference-sync.service.ts
```diff
@@ -0,0 +1,306 @@
...
+  SyncScope,
+  SyncedKeyEntry,
+  PREFERENCE_SYNC_ENABLED,
+  isGlobalEntry,
+} from "./synced-preferences-registry";
+import { UserPreferencesRequest } from "./user-preferences.request";
+import { UserPreferencesResponse } from "./user-preferences.response";
+
+/** Callback to retrieve the user's symmetric encryption key. */
+export type UserKeyProvider = (userId: UserId) => Promise<SymmetricCryptoKey | null>;
```
quexten, line 25: Please don't do this. Preferably use the SDK for encryption. If that's not possible at all, then at least use the key service.

libs/common/src/platform/sync/preferences/preference-sync.service.ts
quexten: Note: `encryptString` is deprecated. While currently this is a SHOULD, instead of  MUST, it does say:

https://github.com/bitwarden/clients/blob/ca0e4435925fa219ad3a8bfb78163d2342a47ab9/libs/common/src/key-management/crypto/abstractions/encrypt.service.ts#L9

libs/common/src/platform/sync/preferences/preference-sync.service.ts
```diff
@@ -0,0 +1,306 @@
...
+  private isEntryRelevant(entry: SyncedKeyEntry): boolean {
+    if (entry.scope === SyncScope.Shared) {
+      return true;
+    }
+    // Device-scoped: relevant if no section specified (all devices) or section matches
+    return entry.device == null || entry.device === this.clientType;
+  }
+
+  private async encryptBlob(prefs: SyncedPreferences, userId: UserId): Promise<string | undefined> {
+    const key = await this.getUserKey(userId);
```
quexten, line 267: Blocking: The user-key may be rotated during key rotation. This will entirely break the settings and render them undecryptable.

Further, depending on the size you expect these settings to be, they should really have a content encryption key, and new data should generally not be encrypted by the user-key, only other keys should be wrapped by the user-key, so that a key-rotation only re-wraps keys.

This will require PR's to both the SDK and the client implementations of key-rotation, since the client implementation is still the only active implementation.

shane-melton:
@quexten My apologies, I should have made it more clear in the PR description that this was just a PoC from the innovation sprint and is not intended to be merged/released as-is. Definitely agree with all your points and a more polished solution should absolutely be using the SDK and none of the deprecated TS APIs.

PR #20044, opened 2026-04-07
[CL-1128] Fix OnPush Lit Ignore Kitchen Sink Stories

iivins-livefront:
**Overview:**

The goal of this PR is to migrate all eight Kitchen Sink story components from the default change detection strategy to OnPush, resolving the long-standing CL-764 lint suppressions.

**Key Changes:**
- Added `ChangeDetectionStrategy.OnPush` to all 8 components
- Removed all `// FIXME(CL-764)` and `// eslint-disable-next-line @angular-eslint/prefer-on-push-component-change-detection` comment pairs
- Converted `selectedToggle` in `KitchenSinkToggleListComponent` from a plain property to an Angular signal
- Replaced constructor injection with `inject()` in `KitchenSinkDialogComponent`, `KitchenSinkDialogWithAutofocusComponent`, and `KitchenSinkMainComponent`
- Added protected readonly visibility modifiers to class members where applicable

**Summary:**

The kitchen sink story components previously suppressed the prefer-on-push-component-change-detection lint rule while awaiting a migration, that migration is now. The toggle list component required a signal conversion to remain compatible with `OnPush` since the two-way binding shorthand `[( )]` does not work with signal based state. All changes are confined to `libs/components/src/stories/kitchen-sink/components/` with no impact on prod

### DIFF
--- a/libs/components/src/stories/kitchen-sink/components/dialog-virtual-scroll-block.component.ts
+++ b/libs/components/src/stories/kitchen-sink/components/dialog-virtual-scroll-block.component.ts
@@ -1,5 +1,5 @@
 import { ScrollingModule } from "@angular/cdk/scrolling";
-import { Component, OnInit } from "@angular/core";
+import { ChangeDetectionStrategy, Component, OnInit, inject } from "@angular/core";
 
 import { DialogModule, DialogService } from "../../../dialog";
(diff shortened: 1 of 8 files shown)


vleague2 (approved):
Thank you! Can you update the PR title before merging?

PR #20045, opened 2026-04-07
[CL-962] update banner component styles

iivins-livefront:
**Overview:**

https://bitwarden.atlassian.net/browse/CL-962

The goal of this PR is to update the Banner component by introducing a structured slot based content model that supports a title, body text, and actions as distinct projection areas.

The component API was modernized with cleaner naming and semantic design tokens - all current consumers across the client applications were updated to match.

**Key Changes**

- Banner input renamed from bannerType to variant, with "info" -> "primary" and "premium" -> "success"
- Title slot, default body slot, and actions slot ( actions only render when a title is projected )
- BannerTitleDirective added as a content detection marker, ( gates action slot )
- Rendering bit-icon-tile with size responsiveness
- Update the Storybook stories and markdown
- All existing consumers updated from bannerType to variant with corrected value names
 
**Breaking Changes Addressed:**

- apps/web/src/app/auth/settings/verify-email.component.html
- apps/web/src/app/billing/organizations/warnings/components/organization-free-trial-warning.component.ts
- apps/web/src/app/billing/organizations/warnings/components/organization-reseller-renewal-warning.component.ts
- apps/web/src/app/billing/warnings/components/tax-id-warning.component.ts
- apps/web/src/app/vault/individual-vault/vault-banners/vault-banners.component.html
- apps/browser/src/vault/popup/components/at-risk-callout/at-risk-password-callout.component.html
- apps/browser/src/platform/popup/layout/popup-layout.stories.ts
- libs/components/src/stories/kitchen-sink/components/kitchen-sink-main.component.ts

**Screenshots:**
Screenshot 2026-04-07 at 4 54 57 PM
Screenshot 2026-04-07 at 4 54 36 PM
Screenshot 2026-04-07 at 4 54 22 PM


libs/components/src/banner/banner-title.directive.ts
```diff
@@ -0,0 +1,18 @@
...
+ *
+ * @example
+ * ```html
+ * <bit-banner>
+ *   <span slot="title">Title text</span>
+ * </bit-banner>
+ * ```
+ */
+@Directive({
+  selector: "[slot='title']",
```
vleague2: I think this will match with any slot named `title`, not just the banner title slot. The `simple-dialog.component.ts` does something similar to what you're trying to do here with needing some way to reference a slot through Angular content children -- maybe that approach will work for you and keep it scoped to the banner component?

libs/components/src/banner/banner.component.html
```diff
@@ -1,17 +1,54 @@
...
-    <ng-content></ng-content>
-  </span>
-  <!-- Overriding hover and focus-visible colors for a11y against colored background -->
+  <div class="tw-flex tw-flex-1 tw-gap-3" [class]="alignClass()">
+    @if (displayIcon(); as icon) {
+      <bit-icon-tile
+        [icon]="icon"
+        [variant]="variant()"
+        size="base"
+        [class]="showActions() ? '@5xl:tw-hidden' : ''"
```
vleague2: Is the only difference between this one and the one below it the size? And they are mutually exclusive? Could we conditionally pass the size input instead of having the icon tile twice?

libs/components/src/banner/banner.component.html
```diff
@@ -1,17 +1,54 @@
...
+    <div
+      class="tw-flex tw-flex-col @5xl:tw-flex-row @5xl:tw-items-center tw-gap-1 @5xl:tw-gap-3 tw-min-w-0"
+      [class]="showActions() ? '@5xl:tw-flex-1' : ''"
+    >
+      <div
+        class="tw-flex tw-flex-col tw-gap-1 tw-min-w-0"
+        [class]="showActions() ? '@5xl:tw-flex-1' : ''"
+      >
+        <div class="tw-text-lg tw-font-medium tw-leading-6 tw-text-main tw-m-0">
+          <ng-content select="[slot='title']"></ng-content>
```
vleague2: Thoughts on doing an input vs a slot here? I see we are applying some text styles to match the design, but the slot means anything can be passed in with different text styles. If we want to enforce it, an input would be easier (but prevents more complex content from going in the title). Maybe worth a check with design on what could go in banner titles?

libs/components/src/banner/banner.component.html
```diff
@@ -1,17 +1,54 @@
...
+        </div>
+        <!-- Overriding focus-visible color for link buttons for a11y against colored background -->
+        <span
+          class="tw-text-sm tw-leading-5 [&>button[bitlink]:focus-visible:before]:!tw-ring-text-main"
```
vleague2: Do we have a typography variant that would work here instead of customizing the size and leading?

libs/components/src/banner/banner.component.html
```diff
@@ -1,17 +1,54 @@
...
+      [class]="showActions() ? '@5xl:tw-flex-1' : ''"
+    >
+      <div
+        class="tw-flex tw-flex-col tw-gap-1 tw-min-w-0"
+        [class]="showActions() ? '@5xl:tw-flex-1' : ''"
+      >
+        <div class="tw-text-lg tw-font-medium tw-leading-6 tw-text-main tw-m-0">
+          <ng-content select="[slot='title']"></ng-content>
+        </div>
+        <!-- Overriding focus-visible color for link buttons for a11y against colored background -->
```
vleague2: Aren't the focus rings already black? `text-main` is also from the old palette, not the new one

libs/components/src/banner/banner.component.ts
```diff
@@ -1,89 +1,108 @@
...
+import { BannerTitleDirective } from "./banner-title.directive";
 
-const defaultIcon: Record<BannerType, string> = {
-  premium: "bwi-star",
-  info: "bwi-info-circle",
+export type BannerVariant = "primary" | "success" | "warning" | "danger";
+
+const defaultIcon: Record<BannerVariant, BitwardenIcon> = {
+  primary: "bwi-info-circle",
+  success: "bwi-star",
```
vleague2, line 16: We might want to check these defaults with design, I am not seeing it spec'd out in the Figma

libs/components/src/banner/banner.stories.ts
```diff
@@ -45,84 +47,145 @@ export default {
...
+      props: args,
+      template: /*html*/ `
+        <bit-banner
+          [variant]="variant"
+          [showClose]="showClose"
+        >
+          <span slot="title">Integration is the key</span>
+          Bitwarden is the most trusted password manager. With many tools to make your work even more efficient.
+          <ng-container slot="actions">
+            <button bitButton type="button" buttonType="secondary">Cancel</button>
```
vleague2: I think the buttons in the Figma examples are the small size, we should do that in the stories too. The button variants should also match the variant of the banner

libs/components/src/banner/banner.stories.ts
```diff
@@ -45,84 +47,145 @@ export default {
...
-        Bruce
+    template: /*html*/ `
+      <bit-banner [variant]="variant" [showClose]="showClose">
+        <span slot="title">
+          <span bitBadge variant="primary">New</span>
+          Integration updates
+        </span>
+        Bitwarden is the most trusted password manager. With many tools to make your work even more efficient.
+        <ng-container slot="actions">
+          This change affects all organization members. <a bitLink linkType="primary">Read the release notes</a>
```
vleague2: I think we should keep the `actions` slot as buttons only in these examples

libs/components/src/banner/banner.stories.ts
```diff
@@ -45,84 +47,145 @@ export default {
...
-        Bruce
-      </bit-banner>
-      `,
+    `,
   }),
+  args: {
+    variant: "primary",
+    showClose: true,
+  },
 };
```
vleague2, line 244: Maybe we should add a story for hiding the icon, while we're in here?

libs/components/src/banner/banner.mdx
```diff
@@ -8,44 +8,56 @@ import * as stories from "./banner.stories";
...
-### Info
+Renders a heading above the body text. When a title is projected, the actions slot becomes visible
+and the layout shifts to its expanded form. Apply `slot="title"` to any element.
 
-<Canvas of={stories.Info} />
+### Actions slot (`slot="actions"`)
 
-Used to communicate release notes, server maintenance or other informative event.
+Only visible when a title is projected. At wide widths the actions are right-aligned; at narrow
+widths they appear below the body text. Wrap action elements in `ng-container slot="actions"`.
```
vleague2, line 32: We can add a note here about choosing button variants based on the banner variant

libs/components/src/banner/banner.component.html
iivins-livefront: From my understanding, CSS container queries switch layouts at runtime. Since `size` is a TS input resolved at render time, a single element can't change its size in response to a breakpoint

libs/components/src/banner/banner.component.html
iivins-livefront: The thought process behind this was to allow the consumer to be as customizable as possible while also keeping the typography integrity. I can't find a message that says it, but I can check with Emily to confirm if design _would ever want_ it to be different

libs/components/src/banner/banner-title.directive.ts
iivins-livefront: I'm doing something similar on the Popover PR - going to switch the approach there as well

libs/components/src/banner/banner.component.html
iivins-livefront: `tw-leading-6` was intentionally placed on the component's wrapper div so it's no longer on the consumer's title element. That way consumers get the correct line-height automatically and without having to add it themselves

Closest typography I could find
`<div bitTypography="h4" noMargin class="tw-leading-6">`

libs/components/src/banner/banner-title.directive.ts
iivins-livefront: Per Emily from comments on the the Jira Ticket https://bitwarden.atlassian.net/browse/CL-962

"That’s a good question. I could see a future world in which the title could have a badge, but I’m not aware of any specific use cases for that at the moment.

If there aren’t any big implementation considerations with making it a title slot, I’m all for having some flexibility there!"

libs/components/src/banner/banner.component.html
vleague2: Well that is a bummer haha! Maybe we can include a quick comment here about this in case other people like me get confused 😂

libs/components/src/banner/banner.component.html
vleague2: Hmm okay interesting. Can you ask design if the banner text is supposed to be lined up with our typography variants, or if it's custom on purpose? We did the typography updates a while before starting milestone one, so I wonder if some updates are needed if these banner text styles are actually supposed to be the correct ones

libs/components/src/banner/banner.stories.ts
```diff
@@ -45,84 +45,136 @@ export default {
...
-export const Base: Story = {
+export const BannerBase: Story = {
   render: (args) => {
     return {
       props: args,
       template: /*html*/ `
-        <bit-banner ${formatArgsForCodeSnippet<BannerComponent>(args)}>
-          Content Really Long Text Lorem Ipsum Ipsum Ipsum
-          <button type="button" bitLink linkType="secondary">Button</button>
+        <bit-banner
```
willmartian: We should continue to use `formatArgsForCodeSnippet`, to improve the rendering in the Storybook UI.

willmartian (changes requested):

libs/components/src/banner/banner.component.html
willmartian: Bump

libs/components/src/banner/banner.stories.ts
iivins-livefront: Using `formatArgsForCodeSnippet` for the 4 available Banner stories:
- `BannerBase`
- `BannerBaseLargeText`
- `BannerSimple`
- `BannerSimpleLargeTest`

libs/components/src/banner/banner.component.ts
iivins-livefront: These have been updated in the Figma designs to match

BryanCunningham (dismissed):

libs/components/src/banner/banner.component.html
vleague2: I will move this to a follow-up ticket to discuss with design

sbrown-livefront (dismissed):
Billing looks good!

libs/components/src/banner/banner.component.html
```diff
@@ -1,24 +1,63 @@
 <div
-  class="tw-flex tw-items-center tw-gap-4 tw-p-2 tw-ps-4 tw-text-main tw-border-transparent tw-bg-clip-padding tw-border-solid tw-border-b tw-border-0"
+  class="tw-flex tw-gap-3 tw-bg-clip-padding tw-border-solid tw-border-b tw-border-0 tw-min-h-14 tw-p-4"
```
BryanCunningham: @willmartian Maybe worth looking into. I doubt we need this `tw-bg-clip-padding` class. And, I'm not for sure we want this `tw-min-h-14` either. Should probably just let the content set the height here

JaredSnider-Bitwarden (approved):
Auth file change LGTM!

libs/components/src/banner/banner.component.html
willmartian: Good catch. Fixed here [547f826](https://github.com/bitwarden/clients/pull/20045/commits/547f826036fdc2a44979c065e3b7673e45d0fb29)

BryanCunningham (approved):

nikwithak (approved):
✅ Vault changes

JaredSnider-Bitwarden (approved):

sbrown-livefront (approved):
✅ Billing looks good

JimmyVo16 (approved):

PR #20058, opened 2026-04-08
[PM-31270] New default argon2id in change kdf component

mzieniukbw:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-31270

## 📔 Objective

Adjusted default Argon2 configuration when user manually selects Argon2Id in change kdf component to:
Iterations: 6
Memory: 32MB
Parallelism: 4

Also refactored `PBKDF2KdfConfig` and `Argon2KdfConfig`:
- explicit defaults with `createDefault` function, instead of relying on the constructor with all optional arguments.
- kdf fields are readonly, immutable - prevents `change-kdf.component.ts` from modifying the global const, like it was before.

## 📸 Screenshots

New Argon2Id defaults when changing KDF

https://github.com/user-attachments/assets/15e45995-cbda-45a4-a9c2-122705a1c35e

### DIFF
--- a/apps/web/src/app/auth/emergency-access/services/emergency-access.service.ts
+++ b/apps/web/src/app/auth/emergency-access/services/emergency-access.service.ts
@@ -313,8 +313,8 @@ export class EmergencyAccessService implements UserKeyRotationKeyRecoveryProvide
       case KdfType.Argon2id:
         config = new Argon2KdfConfig(
           takeoverResponse.kdfIterations,
-          takeoverResponse.kdfMemory,
-          takeoverResponse.kdfParallelism,
+          takeoverResponse.kdfMemory!,
+          takeoverResponse.kdfParallelism!,
         );
         break;
     }
--- a/libs/common/src/auth/password-prelogin/password-prelogin.model.ts
+++ b/libs/common/src/auth/password-prelogin/password-prelogin.model.ts
@@ -19,7 +19,11 @@ export class PasswordPreloginData {
     const kdfConfig =
       response.kdf === KdfType.PBKDF2_SHA256
         ? new PBKDF2KdfConfig(response.kdfIterations)
-        : new Argon2KdfConfig(response.kdfIterations, response.kdfMemory, response.kdfParallelism);
+        : new Argon2KdfConfig(
+            response.kdfIterations,
+            response.kdfMemory!,
+            response.kdfParallelism!,
+          );
     kdfConfig.validateKdfConfigForPrelogin();
     return new PasswordPreloginData(kdfConfig);
   }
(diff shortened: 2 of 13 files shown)


quexten (dismissed):

addisonbeck (dismissed):

Thomas-Avery (dismissed):
LGTM

apps/web/src/app/auth/emergency-access/services/emergency-access.service.ts
```diff
@@ -313,8 +313,8 @@ export class EmergencyAccessService implements UserKeyRotationKeyRecoveryProvide
       case KdfType.Argon2id:
         config = new Argon2KdfConfig(
           takeoverResponse.kdfIterations,
-          takeoverResponse.kdfMemory,
-          takeoverResponse.kdfParallelism,
+          takeoverResponse.kdfMemory!,
+          takeoverResponse.kdfParallelism!,
```
rr-bw, line 317: @mzieniukbw Question: Now that the `PBKDF2KdfConfig` and `Argon2KdfConfig` constructors no longer fall back to defaults for undefined properties, what happens in scenarios like the above if `kdfIterations` or the `kdfMemory`/`kdfParallelism` on the response is undefined. Is that even possible? How would undefined values be caught to prevent acting on a malformed KDF? I noticed we have runtime checks in `KdfConfigResponse` ([see here](https://github.com/bitwarden/clients/blob/main/libs/common/src/key-management/models/response/kdf-config.response.ts#L12-L39)), and I'm not sure if these checks are now necessary elsewhere.

libs/common/src/auth/password-prelogin/password-prelogin.model.ts
```diff
@@ -19,7 +19,11 @@ export class PasswordPreloginData {
     const kdfConfig =
       response.kdf === KdfType.PBKDF2_SHA256
         ? new PBKDF2KdfConfig(response.kdfIterations)
-        : new Argon2KdfConfig(response.kdfIterations, response.kdfMemory, response.kdfParallelism);
+        : new Argon2KdfConfig(
+            response.kdfIterations,
+            response.kdfMemory!,
+            response.kdfParallelism!,
```
rr-bw, line 25: Similar question as [here](https://github.com/bitwarden/clients/pull/20058/changes#r3061090951).

apps/web/src/app/auth/emergency-access/services/emergency-access.service.ts
mzieniukbw: I made them required in https://github.com/bitwarden/clients/pull/20058/commits/d5697e4d18eb541fe7946c98e4728a1dc9d659e6. If null or undefined is provided, it will throw.
If a malformed KDF is provided in the server response, then that's ok. There is nothing we can do anyway, we need to be strict in those cases.

libs/common/src/auth/password-prelogin/password-prelogin.model.ts
mzieniukbw: Fixed in https://github.com/bitwarden/clients/pull/20058/commits/d5697e4d18eb541fe7946c98e4728a1dc9d659e6

quexten (approved):

Thomas-Avery (approved):

rr-bw (approved):

dani-garcia (approved):

PR #20072, opened 2026-04-09
[PM-33554] Fix emergency access fingerprint

quexten:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-33554

## 📔 Objective

The wrong ID was used in emergency access fingerprints, which meant that the wrong fingerprint is displayed. This PR fixes that.


Thomas-Avery (approved):
LGTM

PR #20080, opened 2026-04-09
Change SDK update PR to create new branch to avoid merge conflicts

trmartin4:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-34876

## 📔 Objective

Problem: When a manual SDK update was merged to `main`, the automation's rebase would conflict and fail, requiring manual branch deletion to recover.

Solution: Instead of rebasing existing branches, always create a fresh branch from the base branch.

What changed:
  1. Replaced "Switch to branch" step with "Check for existing branch", which now uses `git ls-remote` to check if the remote branch exists without switching to it
  2. Removed the rebase step entirely
  3. Added "Create fresh branch from base" step, which deletes local branch if it exists, creates new branch from base branch

So, the result will be:
  1. If the branch exists and the last commit is by the bot, it will be deleted and re-created, avoiding any merge conflicts.
  2. If the branch exists and the last commit is not by the bot, it will not update anything or delete any branches.  This would allow for cases where users are manually fixing breaking changes on the branch.  **This means that any automatic PR generation will be missed while someone has taken control of the branch with manual commits.**
  3. If the branch doesn't exist, it will be created.

### DIFF
--- a/.github/workflows/sdk-update.yml
+++ b/.github/workflows/sdk-update.yml
@@ -154,29 +154,32 @@ jobs:
             exit 1
           fi
 
-      - name: Switch to branch
-        id: switch-branch
+      - name: Check for existing branch
+        id: check-branch
+        env:
+          GH_TOKEN: ${{ steps.app-token.outputs.token }}
+          BASE_BRANCH: ${{ inputs.base-branch }}
         run: |
           BRANCH_NAME="sdlc/sdk-update"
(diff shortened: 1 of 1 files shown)


trmartin4:
@quexten tagging you as you brought up this issue, to make sure you agree this will fix it.

djsmith85 (approved):
Changes are looking good.

quexten (approved):
@trmartin4 Thanks for the ping, yes this seems like it would resolve the issue.

PR #20085, opened 2026-04-09
[CL-963] Update breadcrumb styles

lxiong-livefront:
## 🎟️ Tracking

[JIRA](https://bitwarden.atlassian.net/browse/CL-963)

## 📔 Objective

Updated the styling of current breadcrumb component and bifurcate into small and base size variants. Added support for active styles when the current page is represented in the list of breadcrumbs, by hooking into the Angular router. 

## 📸 Screenshots

Screenshot 2026-04-10 at 9 35 27 PM
<br/>
<br/>

https://github.com/user-attachments/assets/7c726c68-5e66-40a0-9dd2-a9d7dd18076d


libs/components/src/breadcrumbs/breadcrumb.component.html
```diff
@@ -1,6 +1,6 @@
 <ng-template>
-  @if (icon(); as icon) {
```
vleague2, line 2: The `as icon` is one of those "not enforced but seems important" Angular things -- if you were doing this in typescript, it would error because the 2nd function call of `icon()` on line 3 could be `undefined` technically. Storing it as the `icon` variable means the `if` check works properly.

vleague2 (commented):
This is looking great! I see some Storybook failures, and I have only one comment so far 😂

libs/components/src/breadcrumbs/breadcrumb.component.html
lxiong-livefront: added back in as `BitwardenIcon` [3ea0b29](https://github.com/bitwarden/clients/pull/20085/commits/3ea0b29c02b1bb031ae73e6db2306de77e41a411)

libs/components/src/breadcrumbs/breadcrumbs.component.ts
```diff
@@ -55,5 +80,23 @@ export class BreadcrumbsComponent {
   });
 
   /** The last breadcrumb, shown after the overflow menu */
-  protected readonly afterOverflow = computed(() => this.breadcrumbs().slice(-1));
+  protected readonly afterOverflow = computed(() => this.breadcrumbs().at(-1));
+
+  protected readonly baseStyles = [
+    "tw-inline-block",
+    "!tw-m-0",
+    "focus-visible:!tw-text-fg-brand",
```
BryanCunningham, line 104: Do we need all the `!important` declarations here? Or, can those be removed?

lxiong-livefront: I was able to remove a few `!important` declarations, but the others are necessary to override styles [309ca7a](https://github.com/bitwarden/clients/pull/20085/commits/309ca7ad907d0c68536e0737bc08b0d382727360)

BryanCunningham: @lxiong-livefront Awesome! If these are the 'base' styles, what is overriding them? I would think, if anything, we _might_ have to use `!important` on a more specific style/variant. But, even then, I usually don't find using 'important' necessary

lxiong-livefront: @BryanCunningham  The margin has a default from `bitTypography`, and I believe the focus-visible ones have defaults from button and link

BryanCunningham: Got it. Thanks for the context!

BryanCunningham (dismissed):

libs/components/src/breadcrumbs/breadcrumb.component.html
vleague2: Sorry, I don't think I explained it well! It's not a typecast, it's a variable assignment that stores the result of `icon()` for use. So `@if (icon(); as icon)` ensures that whatever depends on the result of this `if` check is using the same value as the `if` check (rather than invoking the function a 2nd time and possibly getting a different result). You would use it on line 3 `[name]="icon"`. Docs here on the syntax: https://angular.dev/guide/templates/control-flow#referencing-the-conditional-expressions-result

lxiong-livefront: Ah gotcha, thanks for clarifying! Added back in [48310c7](https://github.com/bitwarden/clients/pull/20085/commits/48310c705822183b3d1df27769e58f15b79b2025)

Apologies for being too quick with it (derogatory) fixed the `[name]` in [5de46de](https://github.com/bitwarden/clients/pull/20085/commits/5de46de14a4d7b26cde16c1a9a21aa9bdb0ad79b)

vleague2 (commented):
This looks good to me, just the missing translations erroring is the only thing I see!

lxiong-livefront:
> This looks good to me, just the missing translations erroring is the only thing I see!

@vleague2 Added translations in [dc26184](https://github.com/bitwarden/clients/pull/20085/commits/dc26184688055ba779c0cd23a8ef96492bb4a232)

BryanCunningham (approved):

BryanCunningham (approved):

jengstrom-bw (approved):
vault changes look good

BTreston (approved):

PR #20081, opened 2026-04-09
[PM-33797] AIV2: Standardize Models and Services: Wire Up

Banrion:
## 🎟️ Tracking

[PM-33797](https://bitwarden.atlassian.net/browse/PM-33797) — Access Intelligence V2: Standardize Models and Services
Source / reference: Draft PR #19600

This is part of a series of incremental changes to introduce the V2 architecture for Access Intelligence. The full series introduces, in order:

1. V2 data model family and encryption service abstraction ✅ merged
2. Type guards expansion ✅ merged
3. V2 service layer (bit-common) ✅ merged
4. Access security tasks service refactor (bit-web) ✅ merged
5. V2 shared UI components ✅ merged
6. V2 page-level UI components ✅ merged
7. Module/routing wiring and V2 tab additions (current)
8. Model renames and old model cleanup

## 📔 Objective

Wires the V2 architecture into the Angular module and routing so the V2 page renders behind the `AccessIntelligenceNewArchitecture` feature flag. Also adds the two remaining V2 tab components (`all-applications-tab`, `critical-applications-tab`) needed for Milestone 11 support, and fixes a backwards-compatibility issue in the bit-common service layer.

### Dependencies

⚠️ Depends on: all six prior PRs in the series ✅ merged

### Module and routing wiring

`access-intelligence-routing.module.ts` registers the V2 page component behind `AccessIntelligenceNewArchitecture` using `featureFlaggedRoute`. `access-intelligence.module.ts` adds all V2 service providers, declares `AccessIntelligencePageComponent`, and maps the `AdminTaskService` abstract token to `DefaultAdminTaskService` (required by `DefaultAccessSecurityTasksService`).

### bit-common service fixes

`RiskInsightsEncryptionService` replaced with `LegacyRiskInsightsEncryptionService` in both `risk-insights-orchestrator.service.ts` and `risk-insights-report.service.ts`. The concrete service was renamed in PR 3; this update brings the injected token into alignment and preserves backwards-compatible decryption of existing vault data.

### Milestone 11 V2 tab additions

[...]

PR #20090, opened 2026-04-10
[PM-32654] Access Intelligence: Applications table column width adjustments

AlexRubik:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-32654

## 📔 Objective

Fixes the Applications table column widths shifting during scroll in the Access Intelligence page.

The table uses virtual scrolling (`bit-table-scroll`) with `table-layout: auto`, which causes the browser to recalculate column widths based on currently rendered rows. As the user scrolls, rows with different application name lengths enter the viewport, causing columns to resize.

This applies the same fixed-layout pattern used by the 5 existing report tables (weak passwords, exposed passwords, reused passwords, unsecured websites, inactive two-factor):

- `layout="fixed"` on `bit-table-scroll` to lock column widths
- `tw-w-12` on checkbox and icon `<th>` headers to constrain utility columns
- `tw-w-1/4` on the Application `<th>` to allocate 25% of table width
- `tw-truncate tw-max-w-0` on the application name `<td>` to force text truncation within the fixed column boundary

Changes applied to all 3 applications table templates:
- `app-table-row-scrollable.component.html` (pre-M11, used when `Milestone11AppPageImprovements` is off)
- `app-table-row-scrollable-m11.component.html` (M11)
- `applications-table-v2.component.html` (V2, behind `AccessIntelligenceNewArchitecture` flag)

### Before (from jira ticket, dynamic width of application column)

https://github.com/user-attachments/assets/7e32b39f-eade-4697-ad6a-b0574858c501

### After (fixed width)

https://github.com/user-attachments/assets/9cf26f90-95a2-4859-a224-a8bff7f6a215

## Note on scope

This PR includes the same fix across all 3 table templates. The V2 template (`applications-table-v2.component.html`) shares the same table structure but uses signals/outputs instead of legacy input patterns. The `AccessIntelligenceNewArchitecture` flag is not consumed in the codebase yet, so the V2 table is not currently reachable at runtime. Is it acceptable to include the V2 change here, or should it be split into a separate PR?


Banrion:
Let's hold on this pr until https://github.com/bitwarden/clients/pull/20081 is merged so that we can apply changes and verify the storybooks.

PR #20095, opened 2026-04-10
[PM-34830][POC] Atomic writes for SDK client managed state

mzieniukbw:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-34830

## 📔 Objective

A minimal, scoped to SDK client managed state change, where a repository / state can choose whether the write should be optimistic (current behaviour) or atomic (as of, guaranteed to retrieve the value via get, immediately after write).

When `optimisticWrite` is false (atomic) on the repository registration, all writes will wait for up to 1000ms, until the change is reflect in `get`. This makes sure any subsequent `get` will reflect the value, immediately after writing (i.e. by `set`).

By default atomic, instead of optimistic (a choice) - to prevent race condition bugs in SDK, as we’ve seen plenty in TS cli.

Does not require an architectural change in the state or storage, does not change `state$` from hot to cold observable globally, does not affect TS state, only SDK.

Tested on LocalUserDataKey state, commented out the [cleanupDelayMs](https://github.com/bitwarden/clients/blob/1b63466144351ea6066611b2e8e0cf0855ad565b/libs/common/src/platform/services/key-state/local-user-data-key.state.ts#L11). I don't see the race condition in SDK, that we've seen as part of https://bitwarden.atlassian.net/browse/PM-33850 anymore. When i turn on the `optimisticWrite` to true on the repository registration, i can see the race condition back again.


libs/common/src/platform/services/sdk/client-managed-state.ts
coroiu, line 1: suggestion: Instead of checking equality on a single ID, what about comparing the entire state?

```
    const newState = await this.getUserState().update((state) => ({
      ...(state ?? {}),
      [id]: newValue,
    }));
    await this.waitUntilChanged(newState);
```

This should get rid of the loops and let you do a single equality check even for batch changes

mzieniukbw: Done, but i went with targeted update in form of a predicate function, to avoid full-state comparison (`JSON.stringify` on every `state$`emission would be expensive at 1k+ items (i.e. ciphers)). Unless you know better way to compare objects without over-engineering.

mzieniukbw:
I did some benchmarking with setting and getting the `LocalUserDataKey` in SDK, with `optimisticWrite` true and false. I saw no difference.
In fact, the entire `initialize_user_crypto` takes up to `5ms` for my test vault.

coroiu (approved):

PR #20096, opened 2026-04-10
[PM-31364] discard edits dialog to policies page

JaredScar:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-31364

## 📔 Objective

Adds a discard edits modal when navigating away with changes in the opened policy drawer.

## 📸 Screenshots

image


BryanCunningham (dismissed):

libs/components/src/dialog/simple-dialog/simple-configurable-dialog/simple-configurable-dialog.component.ts
```diff
@@ -52,6 +53,10 @@ export class SimpleConfigurableDialogComponent {
     ];
   }
 
+  get acceptButtonType(): ButtonType {
+    return this.simpleDialogOpts.type === "danger" ? "danger" : "primary";
```
sven-bitwarden: This is going to change the button for every usage of the dialog that has type === "danger" - is that acceptable?

BTreston (commented):
This seems like a pretty wide reaching change to support this. Are we in general moving towards this pattern for all dialogs or is this just supposed to be for policies? I worked with Claude on reeling in the scope of this change to just policies and we can accomplish the same behavior without affecting our dialog code. (aka you can just hook into the cdk dialog ref inside the component and do your logic based on the form state)

Credit to @sven-bitwarden for pointing it out

libs/components/src/dialog/simple-dialog/simple-configurable-dialog/simple-configurable-dialog.component.ts
JaredScar: Great catch, thanks! Resolving!

apps/web/src/app/admin-console/organizations/policies/policy-edit-dialog.component.ts
```diff
@@ -160,6 +208,8 @@ export class PolicyEditDialogComponent implements AfterViewInit {
         variant: "success",
         message: this.i18nService.t("editedPolicyId", this.i18nService.t(this.data.policy.name)),
       });
+      // Clear the predicate so the drawer closes immediately after a successful save.
+      this.dialogRef.closePredicate = undefined;
```
BTreston: I don't think you answered the question on my prior review. My big concern is that this pattern is very not obvious. Maybe we need some product input here to determine if its worth building out this capability in a more generic/reusable way, but as far as I can tell I don't think we should be changing the base dialog components and services to support this.

sven-bitwarden:
There are two introduced places where we're fire-and-forgetting a promise, between the TryClose and the observable calling Edit. I think we could remove those if we removed the API expansion on the base dialog (TryClose) and kept it local to policies. 
Rather than calling TryClose and not knowing if it closed or not, we should be able to observe the attempt to close in the policies dialog and display the confirmation, the result of which is calling Close knowing that it's actually the intention.

BTreston (changes requested):
Tests are failing, otherwise approved.

kdenney (approved):
Billing changes look good.

BTreston (dismissed):

BryanCunningham (dismissed):

sven-bitwarden (dismissed):
+1

BryanCunningham (dismissed):

kdenney (dismissed):

sven-bitwarden (approved):

kdenney (approved):

BryanCunningham (approved):

PR #20097, opened 2026-04-10
Improve secure key env var match default case

neuronull:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-35122

## 📔 Objective

Improves the logging of the default case for the secure key env var check.

I noticed this while observing tracing output from ssh agent. It reads:

```
2026-04-07T21:00:21.928081Z  INFO desktop_core::secure_memory::secure_key:  is not a valid secure key container backend, using automatic selection
```

, notably, the default match case is currently doing two things that could be improved: if the env var is not set, the log statement doesn't reflect that because it's an empty string. It doesn't delineate between the Ok(string) case and the Err case. Even though both result in a None return, it's more clear to read the logs what is going on.

Sine the existing code had `info` for both cases, I'm assuming the env var is not required/expected (hence using info for the Err case, instead of something like error or warn).

### DIFF
--- a/apps/desktop/desktop_native/core/src/secure_memory/secure_key/mod.rs
+++ b/apps/desktop/desktop_native/core/src/secure_memory/secure_key/mod.rs
@@ -8,7 +8,7 @@
 //! The implementations include DPAPI on Windows, `keyctl` on Linux, and `memfd_secret` on Linux,
 //! and a fallback implementation using mlock.
 
-use tracing::info;
+use tracing::{debug, info, warn};
 
 mod crypto;
 #[cfg(target_os = "windows")]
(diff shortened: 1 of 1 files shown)


apps/desktop/desktop_native/core/src/secure_memory/secure_key/mod.rs
```diff
@@ -171,10 +173,18 @@ fn get_env_forced_container() -> Option<CrossPlatformSecureKeyContainer> {
...
+            None
+        }
+        Err(error) => {
             info!(
-                "{} is not a valid secure key container backend, using automatic selection",
-                env_var.unwrap_or_default()
+                %error,
+                env_var = ENV_VAR_SECURE_KEY_CONTAINER_BACKEND,
+                "failed to read environment variable, using automatic selection"
             );
```
Thomas-Avery, line 195: 💭  The hot path here is users not setting `SECURE_KEY_CONTAINER_BACKEND` and us using automatic selection. 

I would suggest matching on `std::env::VarError::NotPresent` and just having a `debug!` or no log for that. Then having the other error options being what you have here.

Feels like unhelpful noise to have something like: 

```
INFO error=environment variable not found env_var=SECURE_KEY_CONTAINER_BACKEND failed to read environment variable, using automatic selection
```

Thomas-Avery (changes requested):

apps/desktop/desktop_native/core/src/secure_memory/secure_key/mod.rs
neuronull: addressed in https://github.com/bitwarden/clients/pull/20097/commits/f29f80b72f33f3a79c9792a934eb0479bee5ede5

also changed two info logs to warn, but LMK if you'd prefer different log level.

Thomas-Avery (approved):
LGTM

PR #20099, opened 2026-04-10
[PM-851] custom field reordering not keyboard accessible

### LINKED ISSUE #2658: Desktop / Browser: Accessibility - Custom field reordering not keyboard accessible
### Steps To Reproduce

1. Go to a cipher/item with more than one custom field
2. Edit the item
3. Note the ability to reorder the custom fields (the three-line "grabber" control)

### Expected Result

It should be possible to reorder fields using the keyboard.

### Actual Result

Currently, custom field reordering only works with a mouse/touchscreen. The three-line reordering control is not keyboard focusable, and there are no alternative controls to move an item up/down.

### Screenshots or Videos

The issue is present both on the browser extension

a series of custom fields, with one field currently being dragged with the mouse to reorder it

and on the desktop application

a series of custom fields, with one field currently being dragged with the mouse to reorder it

### Additional Context

One potential approach is to make the reordering drag handle three-lines control keyboard focusable, and once it does receive focus, display two additional focusable arrow items (arrow up, arrow down) that a keyboard user can focus and then activate to move the item up or down one slot.

I would recommend NOT trying to implement this purely as a keyboard event handler checking for up/down cursor keys, as these will be ignored/overridden for users that also navigate via keyboard and a screen reader, as the cursor keys will be processed directly by - Windows-based - screen readers to move focus (while on macOS they would still work, since VoiceOver doesn't use pure cursor keys, but needs the VO key to be pressed as well)

Incidentally, 1Password shows two arrows (up and down) when editing an item that has multiple websites (rather than using drag'n'drop). But ironically, these controls in 1Password are also not keyboard focusable/operable. So the approach (showing these controls) is good, but needs to be taken further to actually still then make them work for keyboard users.

screenshot from 1Password, showing two website fields - the one in focus is preceded by up and down arrow controls

### Operating System

Windows

### Operating System Version

11

### Web Browser

Chrome

### Browser Version

101.0.4951.54 (Official Build) (64-bit)

### Build Version

1.58.0 (extension) / Version 1.33.0 Shell 16.1.0 Renderer 96.0.4664.174 Node 16.9.1 Architecture x64 (desktop)

jengstrom-bw:
## 🎟️ Tracking

[Jira](https://bitwarden.atlassian.net/browse/PM-851)

## 📔 Objective

Enable use of only the keyboard to reorder custom fields and URIs.
fixes #2658  

## 📸 Screenshots

https://github.com/user-attachments/assets/c4015aa3-b047-4be5-a695-3b32250ac37c


apps/desktop/src/locales/en/messages.json
```diff
@@ -1876,6 +1876,12 @@
     "message": "Default match detection",
     "description": "Default URI match detection for auto-fill."
   },
+  "showMatchDetectionNoPlaceholder": {
```
jengstrom-bw, line 1978: These are added since they were missing for desktop

nikwithak (dismissed):

patrickhlauke:
potentially cleaner to wrap the controls in `role="application"` which forces passthrough of most key presses even when an SR is running. then just pass on instructions to let SR users know they can use cursor keys

jengstrom-bw:
> potentially cleaner to wrap the controls in `role="application"` which forces passthrough of most key presses even when an SR is running. then just pass on instructions to let SR users know they can use cursor keys

I thought about doing that, but I noticed the user gets kind of "trapped" on the button. They would have to toggle back to browse mode to use the arrow keys to navigate again. Since they would need to toggle either way if they use arrow key navigation, I didn't want to trap them on the button

patrickhlauke:
> I thought about doing that, but I noticed the user gets kind of "trapped" on the button. They would have to toggle back to browse mode to use the arrow keys to navigate again. Since they would need to toggle either way if they use arrow key navigation, I didn't want to trap them on the button

Regular Windows SR users will be familiar with this already though, without the need for instructions on how to do the opposite. And as they're editing, they'll most likely be `Tab`bing through things anyway (which is not intercepted by application mode)

And for macOS / VO users, an explanation of needing to go into/out of focus mode is unnecessary as VO doesn't have that concept. So the proposed instructions would not make sense to them.

So something like...

```
<div role="application">
  <button ... aria-label="... Use cursor keys to reorder ..."> ... </button>
</div>
```

may be more generic/appropriate.

(beyond that, once SR users *are* able to reorder, you'll probably need to consider using an ARIA live region to give them feedback when things *have* moved, what position they're in, or similar ... unless they've messed with it, GitHub itself had a decent implementation of this)

patrickhlauke:
quick side note: i see from the latest file changes that you have the `role="application"` directly on the button. this works, but be aware that this then suppresses/doesn't announce the button role when the control receives focus (but that shouldn't be a big problem for SR users, just something to be aware of).

I mentioned GitHub's drag-and-drop before ... full disclosure, we (my day job) worked with the GH team on some of this, and they did a nice write-up/blog post about it here - in particular the bit about announcing the changes https://github.blog/engineering/user-experience/exploring-the-challenges-in-creating-an-accessible-sortable-list-drag-and-drop/#challenge-announcing-movements-in-rapid-succession (the full "adding a whole dialog as an alternative" would obviously be a bit overkill here) /ht @kendallgassner

nikwithak (approved):

PR #20102, opened 2026-04-10
[CL-1141] Improve tailwind config structure across apps and libs

vleague2:
## 🎟️ Tracking

[CL-1141](https://bitwarden.atlassian.net/browse/CL-1141)

## 📔 Objective

The way tailwind styles are compiled is that tailwind uses the `tailwind.config.js` file to figure out which file paths to scan for tailwind classes. We have multiple of these config files because the client apps have their own code that should not be included in the other clients, but all of them share lib code.

The way it was working before, we were setting a shared base with `tailwind.config.base.js` and importing that into client configs. Unfortunately, the client configs were actually overriding the content paths from the base, and this led to a lot of base paths being specified again all over the place. Additionally, we had a `libs/components` config that Storybook was using, and that one was also overriding content paths and not including all the libs/clients properly.

This PR cleans up how we share and set the content paths. The `tailwind.config.base.js` now includes all shared lib paths. Teams adding a new lib will add it to only one place now -- here. The 3 clients import that base, and then concatenate their own local paths to the base content paths. The clients also store their local paths in a separate property for use in the `clients` root config, which pulls together the base and all 3 clients so that Storybook has access to the styles for all the clients. This makes sure that any team-owned code in Storybook, regardless of what its client path is (if it's not in a lib), will get represented with proper styles in Storybook. I deleted the `libs/components` config because: a) that is where Storybook looks first for the relevant tailwind config to use, b) we shouldn't be importing client code into a lib to pull it together for Storybook, and c) that config was not doing anything that the base and root configs couldn't do more cleanly. Now, Storybook will use the clients root config.

[CL-1141]: https://bitwarden.atlassian.net/browse/CL-1141?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/components/tailwind.config.base.js
```diff
@@ -1,20 +1,33 @@
...
 const plugin = require("tailwindcss/plugin");
+const path = require("path");
 
 function rgba(color) {
   return "rgb(var(" + color + ") / <alpha-value>)";
 }
 
+const libContent = [
+  // shared lib paths; add app-specific paths to the app's tailwind.config.js, not here
+  path.resolve(__dirname, "./src/**/*.{html,ts,mdx}"),
```
willmartian: Why do we need path.resolve here? These paths were working before without it, right?

If they are needed, maybe something like:

```ts
const libContent = [
  "./foo",
  // ...
].map(p => path.resolve(__dirname, p));
```

willmartian (dismissed):

libs/components/tailwind.config.base.js
vleague2: The paths were being overridden by the team configs so they weren't actually getting used. I didn't test without the path.resolve, but now that these are being properly imported and used across the clients, it seemed safer to resolve it.

willmartian (approved):
Didn't think of this earlier, but why not capture all libs `"../../libs/**/*.{html,ts,mdx}"`?

PR #20116, opened 2026-04-11
[PM-34931] Keeper di angular

itsadrago:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-33198

## 📔 Objective

Modernize some of the Angular code:
  - *ngFor / *ngIf → @for / @if control flow syntax
  - @Output() EventEmitter → output<T>() signal output
  - CommonModule removed from imports where no longer needed
  - ChangeDetectionStrategy.OnPush added, removing the FIXME comments and eslint-disable suppressions
  - inject() function replaces constructor injection in all components and services
  - @Inject(DIALOG_DATA) replaced with inject<DataType>(DIALOG_DATA) in all dialog components
  - Visibility modifiers applied: protected for template-accessible members, private for internal implementation
  - emailHint$ → emailHint signal using toSignal() in ImportKeeperComponent, removing the async pipe and CommonModule dependency
  - _parentFormGroup → signal<FormGroup | null>(null)
  - submit arrow functions marked protected readonly since they're bound to templates via [bitSubmit]


itsadrago:
Wrong target repo, meant for detunized:keeper-direct-importer

PR #20127, opened 2026-04-13
[PM-31061] Implement biometrics migration

quexten:
## 🎟️ Tracking

https://github.com/bitwarden/sdk-internal/pull/947
https://bitwarden.atlassian.net/browse/PM-31061

## 📔 Objective

Implements automatic migration of persistent biometrics protected user keys. Ephemeral biometric keys are already updated as a side-effect of the unlock service.

For this, a new "key id" value is stored which stores the current key id of the user-key that's stored in the biometric system. This allows the clients to determine which key (no key id = v1) is stored. If a v1 user upgrade to v2, then we would have a null -> key-id transition, and if a v2 user rotates their keys (somehow without logging out, which we don't currently support), then this case is covered too since the key id changes.

> [!NOTE]  
> This is intended to be run in the encrypted migrator *after* a v2 migration.


apps/desktop/src/key-management/biometrics/main-biometrics.service.ts
```diff
@@ -143,7 +147,15 @@ export class MainBiometricsService extends DesktopBiometricsService {
   }
 
   async enrollPersistent(userId: UserId, key: SymmetricCryptoKey): Promise<void> {
-    return await this.osBiometricsService.enrollPersistent(userId, key);
+    await this.osBiometricsService.enrollPersistent(userId, key);
+    await SdkLoadService.Ready;
+    const keyId = CryptoClient.get_key_id_for_symmetric_key(key.toEncoded());
```
Thomas-Avery: ❓  Is there a reason we are using this like PureCrypto vs using `this.sdkService.client$.pipe`

libs/common/src/key-management/encrypted-migrator/migrations/biometric-persistent-encryption-migration.ts
```diff
@@ -0,0 +1,78 @@
...
+    if (!(await this.biometricsService.hasPersistentKey(userId))) {
+      return "noMigrationNeeded";
+    }
+
+    const userKey = await firstValueFrom(this.keyService.userKey$(userId));
+    if (userKey == null) {
+      return "noMigrationNeeded";
+    }
+
+    const currentKeyId = CryptoClient.get_key_id_for_symmetric_key(userKey.toEncoded());
```
Thomas-Avery: ⚠️ Missing `await SdkLoadService.Ready;`. Consider moving to `this.sdkService.client$.pipe`

libs/common/src/key-management/encrypted-migrator/migrations/biometric-persistent-encryption-migration.ts
```diff
@@ -0,0 +1,78 @@
...
+    }
+
+    this.logService.info(
+      `[BiometricPersistentMigration] Re-enrolling biometric keys for user ${userId}`,
+    );
+
+    await this.biometricsService.enrollPersistent(userId, userKey);
+    await this.biometricsService.setBiometricProtectedUnlockKeyForUser(userId, userKey);
+
+    const keyId = CryptoClient.get_key_id_for_symmetric_key(userKey.toEncoded());
```
Thomas-Avery: ⚠️ Missing `await SdkLoadService.Ready;`

libs/common/src/key-management/encrypted-migrator/default-encrypted-migrator.ts
```diff
@@ -24,6 +32,10 @@ export class DefaultEncryptedMigrator implements EncryptedMigrator {
     readonly configService: ConfigService,
     readonly masterPasswordService: MasterPasswordServiceAbstraction,
     readonly syncService: SyncService,
+    readonly keyService: KeyService,
+    readonly biometricsService: BiometricsService,
+    readonly biometricStateService: BiometricStateService,
+    readonly platformUtilsService: PlatformUtilsService,
```
Thomas-Avery: It seems weird to me one of these is `private` and the rest are public? I would think they could all be private.

libs/common/src/key-management/encrypted-migrator/migrations/biometric-persistent-encryption-migration.spec.ts
```diff
@@ -0,0 +1,120 @@
...
+  describe("needsMigration", () => {
+    it("should return 'noMigrationNeeded' when biometric unlock is not enabled", async () => {
+      mockBiometricStateService.biometricUnlockEnabled$.mockReturnValue(of(false));
+
+      const result = await sut.needsMigration(mockUserId);
+
+      expect(result).toBe("noMigrationNeeded");
+    });
+
+    it("should return 'needsMigration' when enrolled key ID does not match current key ID (v2 to v2 migration)", async () => {
```
Thomas-Avery, line 58: Can we add a test where the key IDs do match? For v2 users that will be a hot path.

libs/key-management/src/biometrics/biometric.state.ts
```diff
@@ -69,6 +69,19 @@ export const FINGERPRINT_VALIDATED = new KeyDefinition<boolean>(
...
 
+/**
+ * Stores the key ID of the user key that was last enrolled in the biometric system.
+ * Used to detect when the user key has changed and biometric keys need re-enrollment.
+ */
+export const BIOMETRIC_ENROLLED_KEY_ID = new UserKeyDefinition<string | null>(
+  BIOMETRIC_SETTINGS_DISK,
+  "biometricEnrolledKeyId",
+  {
+    deserializer: (obj: any) => obj,
```
Thomas-Avery:
```suggestion
    deserializer: (obj) => obj,
```

Looks like most of these don't use `:any`?

apps/desktop/src/key-management/biometrics/desktop.biometrics.service.ts
```diff
@@ -9,8 +8,6 @@ import { BiometricsService } from "@bitwarden/key-management";
 export abstract class DesktopBiometricsService extends BiometricsService {
   abstract deleteBiometricUnlockKeyForUser(userId: UserId): Promise<void>;
   abstract setupBiometrics(): Promise<void>;
-  abstract enrollPersistent(userId: UserId, key: SymmetricCryptoKey): Promise<void>;
-  abstract hasPersistentKey(userId: UserId): Promise<boolean>;
```
Thomas-Avery, line 13: ❓ Why are we moving from client specific abstractions to general no-op implementations on BiometricsService?

libs/key-management/src/biometrics/biometric.service.ts
```diff
@@ -45,4 +45,22 @@ export abstract class BiometricsService {
...
+
+  /**
+   * Checks if the given user has a persistent biometric key enrolled.
+   * Default returns false for platforms that don't support persistent biometric keys.
+   * @param userId the user to check
+   * @returns true if a persistent key is enrolled
+   */
+  async hasPersistentKey(userId: UserId): Promise<boolean> {
+    return false;
+  }
```
Thomas-Avery: With these being actual implementations I don't think the code in `MainBiometricsService` is going to run? They don't `override` these implementations. I would keep implementations out of `abstract class`

Thomas-Avery (changes requested):

apps/desktop/src/key-management/biometrics/main-biometrics.service.ts
quexten: I suppose we can change the SDK api with another PR to the sdk. However, there is nothing here that actually needs a client instance; it is a pure function that does not depend on any SDK state. It's essentially a `static function`.

quexten: Maybe in other words, we basically have:

```
Wasm:
[ - Big Wasm Blob of Memory -
Client A
User Client B
  - Key store for user
  - Api client for user
  functions implemented on type instances
  static functions
User Client C
  - ...
User Client D
  - ...
IPC Client
]
/// TS land
TS then has a bunch of references to the structs in wasm memory.
```

You should implement functions on these clients, if they need access to memory in those clients. That is, because we do not need a client struct here, and have no state, it is fine for this to be a static function. As such, we don't need to have the `client$` instance. 

In this case, this is a Pure function, which does not require a client struct.
So there is no technical reason to requires implementing on a client instance. If there is a conventional reason, then I'm not aware of this being written somewhere, but we can certainly adopt it. Please let me know if I should update this, then we need a SDK PR.

apps/desktop/src/key-management/biometrics/desktop.biometrics.service.ts
quexten: Because this is a desktop-only migration, and we don't have a pattern for doing platform-specific encrypted migrations. The migrations don't have access to desktop-specific services. Do we want to scope-increase to figure out a pattern to do platform-specific encrypted migrations in a more general way? I expect this to only apply to desktop.

libs/key-management/src/biometrics/biometric.service.ts
quexten: You don't need the `override` keyword to override a function if that's what you are asking. All `override` does is tell typescript to fail compilation if the method does not exist in the base class, which prevents accidental cases of where you change the base class method, but don't change the override.

apps/desktop/src/key-management/biometrics/main-biometrics.service.ts
Thomas-Avery: Had a synchronous meeting on this.

apps/desktop/src/key-management/biometrics/desktop.biometrics.service.ts
Thomas-Avery: Had a synchronous meeting on this.

libs/key-management/src/biometrics/biometric.service.ts
Thomas-Avery: Classic TypeScript, looks like there is a way to for it to function like other languages https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-3.html#override-and-the---noimplicitoverride-flag but we don't have that turn on.

libs/common/src/key-management/encrypted-migrator/default-encrypted-migrator.ts
quexten: This is another "Classic typescript". Marking these private will make these fields, while having no classifier means that they are just parameters to the constructor. These should not be fields on the encrypted-migrator, since they are only used to construct the concrete migrations.

LogService is different because it is accessed outside of the constructor, and must be a field on the class.

If we make them all private fields, they will all have an unused warning. They should not be fields.

quexten: I suppose we could re-write it to be more clear. Rather than making this field implicit, make it explicit together with the other fields and remove the private classifier?

libs/key-management/src/biometrics/biometric.service.ts
quexten: Removed the default impl

quexten:
Waiting for https://github.com/bitwarden/sdk-internal/pull/982 to proceed.

libs/common/src/key-management/encrypted-migrator/default-encrypted-migrator.ts
Thomas-Avery: The `readonly` modifier already promotes these to class field.

apps/desktop/src/key-management/biometrics/main-biometrics.service.ts
```diff
@@ -142,8 +143,12 @@ export class MainBiometricsService extends DesktopBiometricsService {
     return true;
   }
 
+  /**
+   * Please note, this should only be called via the renderer service, and never directly.
```
quexten, line 147: Would have liked to not have to do this, however the scope increase is quite large, because none of the required services for the SDK service are present in electron main.

We don't have any callers and don't expect any callers from main that don't go via renderer, so this feels fine.

apps/desktop/src/key-management/biometrics/renderer-biometrics.service.ts
```diff
@@ -80,7 +86,16 @@ export class RendererBiometricsService extends DesktopBiometricsService {
   }
 
   async enrollPersistent(userId: UserId, key: SymmetricCryptoKey): Promise<void> {
-    return await ipc.keyManagement.biometric.enrollPersistent(userId, key.toBase64());
+    await ipc.keyManagement.biometric.enrollPersistent(userId, key.toBase64());
+    const keyId = (await firstValueFrom(this.sdkService.client$))
+      .crypto()
+      .get_key_id_for_symmetric_key(key.toEncoded());
```
Thomas-Avery, line 92: ❓ Is there a reason we aren't using the pipe map pattern we use else where?

Thomas-Avery (approved):
LGTM one non blocking question from me.

apps/desktop/src/key-management/biometrics/renderer-biometrics.service.ts
quexten: The pipe pattern is only required / recommended for user-scoped clients, since they have a limited lifetime / require the ref pattern. It's not required for non-scoped clients.

There is a comment here: https://github.com/bitwarden/clients/blob/cdffce8c9219e15882b823db55ace301d417d769/libs/common/src/platform/abstractions/sdk/sdk.service.ts#L119 for the user scoped client.

PR #20128, opened 2026-04-13
Pm 31929 new send policy for deletion days

bmbitwarden:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-31929

## 📔 Objective

Admin Console

Policy applies to non-owners and admins.

Allow administrators to specify deletion days (based on end-user dropdown)

End-user

When creating a new Send, the deletion day field is disabled and replaced with hint text that ‘[Org name] has set this deletion date’.

When editing an existing Send, the field will continue to be disabled with the same hint text

For any existing Sends that are not compliant, they will be marked as disabled and the UI will give the user the option to create a copy of the Send with a compliant deletion date.

## 📸 Screenshots

https://github.com/user-attachments/assets/684f404d-a90d-4e67-94c5-90c36f6f49c8


harr1424:
@claude please review

libs/tools/send/send-ui/src/send-form/components/send-details/send-details.component.ts
```diff
@@ -330,6 +299,32 @@ export class SendDetailsComponent implements OnInit {
...
+      this.sendPolicyService.deletionDatePolicyInfo$,
+      this.accountService.activeAccount$.pipe(
+        getUserId,
+        switchMap((userId) => this.orgService.organizations$(userId)),
+      ),
+    ])
+      .pipe(takeUntilDestroyed())
+      .subscribe(([policyInfo, organizations]) => {
+        const deletionDateControl = this.sendDetailsForm.get("selectedDeletionDatePreset");
+        if (policyInfo.deletionHours != null) {
```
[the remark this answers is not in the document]
mcamirault: Fixed

libs/tools/send/send-ui/src/services/send-policy.service.ts
```diff
@@ -136,7 +136,7 @@ export class SendPolicyService {
               return { deletionHours, orgId };
             }),
           )
-        : of(null),
+        : of({ deletionHours: null, orgId: null }),
```
harr1424: ❓ It isn't immediately clear that this is safe, I think a comment explaining this is intentional and why it is safe would be a cheque written to our future selves to better understand this. 

🎨 The last review also suggested adding a test to ensure this isn't accidentally broken in the future.

mcamirault: Since this value was accessed in a subscribe block the error wouldn't actually crash the entire component and with the feature flag off there's no form behavior to detect, just a console error message, so testing this would be tricky. I was mostly doing this return value to avoid having to do optional chaining in the component so I've changed it back to `of(null)` to match the other observables in the SendPolicyService

harr1424 (dismissed):

sven-bitwarden (approved):

libs/tools/send/send-ui/src/send-form/components/send-details/send-text-details.component.html
```diff
@@ -6,7 +6,7 @@
       id="text"
       rows="3"
       formControlName="text"
-      [readOnly]="formDisabled()"
+      [readonly]="formDisabled()"
```
harr1424, line 9: 🎨 good catch

harr1424 (approved):

PR #20129, opened 2026-04-13
PM-32187 implemented restrict send type

bmbitwarden:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-32187

## 📔 Objective

New send policy to restrict text vs. file

End-user: 

hide drop down and keep button as New or New Send in empty state

Update form header from New Text Send or New File Send to New Send 

## 📸 Screenshots

*The Policy editing experience*

https://github.com/user-attachments/assets/a7e07607-0731-46d5-8689-3e4d850c3357

*Send disabled by this policy change in web client*

https://github.com/user-attachments/assets/5b0d5b78-70c1-4ca1-b3c0-1a8a688788c3

*Send disabled by this policy change in browser client*

https://github.com/user-attachments/assets/97330ab8-d03f-4f1e-b71a-142ea152d7a2


harr1424:
@claude please give this a review

JimmyVo16 (dismissed):

harr1424 (dismissed):
Once concern I have is the following finding from Claude: 

> restrictedSendType is SendTypeRestriction | null (values 0 = TextOnly, 1 = FileOnly), but the comparison is against SendType.File. The code currently works only by coincidence that the numeric values overlap with SendType (Text=0, File=1). If either enum gains a value or its numbering shifts, this silently breaks and the wrong Send type would be created.

Otherwise everything looks good

mcamirault:
> Once concern I have is the following finding from Claude:
> 
> > restrictedSendType is SendTypeRestriction | null (values 0 = TextOnly, 1 = FileOnly), but the comparison is against SendType.File. The code currently works only by coincidence that the numeric values overlap with SendType (Text=0, File=1). If either enum gains a value or its numbering shifts, this silently breaks and the wrong Send type would be created.
> 
> Otherwise everything looks good

Yeah that's a good catch, caused by me losing track of which parts of the UI are duplicated between clients and which use unified components :D. Should be fixed now

mcamirault:
Apologies for the unintentional reviewer additions, the merge went bad somehow and added a ton of import format changes.

@harr1424 I changed up the approach here a little with how we store the allowed Send types; I believe this new method will play nicer with the approach we're discussing with regards to temporary item sharing

JimmyVo16:
Sorry I couldn’t get to this today. I’ll tackle it first thing tomorrow morning.

apps/web/src/app/tools/send/shared/send-success-drawer-dialog.component.ts
```diff
@@ -20,6 +23,19 @@ export class SendSuccessDrawerDialogComponent {
   readonly sendLink = signal<string>("");
   readonly activeSendIcon = ActiveSendIcon;
 
+  private readonly sendPolicyService = inject(SendPolicyService);
+  private readonly allowedSendTypes = toSignal(this.sendPolicyService.allowedSendTypes$, {
+    initialValue: [SendType.Text, SendType.File],
+  });
+
+  get dialogTitle(): string {
+    // TODO:ME - is this necessary?
```
harr1424: ❓ Is this an open question?

harr1424 (dismissed):

JimmyVo16 (dismissed):

apps/web/src/app/tools/send/shared/send-success-drawer-dialog.component.ts
mcamirault: Not anymore, removed

JimmyVo16 (approved):

libs/tools/send/send-ui/src/new-send-dropdown-v2/new-send-dropdown-v2.component.ts
```diff
@@ -3,20 +3,22 @@ import { toSignal } from "@angular/core/rxjs-interop";
 import { map, of, switchMap } from "rxjs";
 
 import { PremiumBadgeComponent } from "@bitwarden/angular/billing/components/premium-badge";
-import { JslibModule } from "@bitwarden/angular/jslib.module";
```
harr1424, line 6: 🎨 Nice!

libs/tools/send/send-ui/src/send-list-filters/send-list-filters.component.ts
```diff
@@ -1,31 +1,38 @@
 import { CommonModule } from "@angular/common";
 import { Component, OnDestroy } from "@angular/core";
+import { toSignal } from "@angular/core/rxjs-interop";
 import { ReactiveFormsModule } from "@angular/forms";
 import { Observable, of, switchMap } from "rxjs";
 
-import { JslibModule } from "@bitwarden/angular/jslib.module";
```
harr1424, line 6: 🎨 nice!

harr1424 (approved):

PR #20134, opened 2026-04-13
[Security] Update rand to 0.9.3

dani-garcia:
## 🎟️ Tracking

https://rustsec.org/advisories/RUSTSEC-2026-0097

## 📔 Objective

The current version of rand has an open advisory for undefined behavior, which is patched in 0.9.3. Looking into the conditions required for this behavior to trigger we should be safe, as we don't meet 3 out of the 5 conditions.

This PR updates our direct dependency and adds an ignore entry to the deny config, as some of our dependencies are pinning an older version of rand.

The failing lint: https://github.com/bitwarden/clients/actions/runs/24349721186/job/71100403550?pr=20132


dani-garcia:
Also tagging @bitwarden/team-desktop-native-dev as you added the initial deny config

coltonhurst (approved):
Looks good, assuming CI passes. Thanks @dani-garcia

neuronull:
FYI, renovate opened a PR that also would have addressed this: https://github.com/bitwarden/clients/pull/20133

apps/desktop/desktop_native/deny.toml
```diff
@@ -5,6 +5,9 @@ unmaintained = "workspace"
 ignore = [
   # Vulnerability in `rsa` crate: https://rustsec.org/advisories/RUSTSEC-2023-0071.html
   { id = "RUSTSEC-2023-0071", reason = "There is no fix available yet." },
+  # rand 0.8.5 pulled transitively by num-bigint-dig (via rsa, ssh-key, oo7). No updated version available.
+  # Not affected: no custom log::Log logger uses rand in this codebase.
+  { id = "RUSTSEC-2026-0097", reason = "Transitive dep via num-bigint-dig; no fix available. Not exploitable: no custom logger uses rand." },
```
neuronull, line 10: I believe you can specify `crate` , to restrict this to the affected crate, which would be more secure.

https://embarkstudios.github.io/cargo-deny/checks/advisories/cfg.html

neuronull (dismissed):

dani-garcia:
> FYI, renovate opened a PR that also would have addressed this: #20133

Yeah I triggered that run, but there are other crates using the affected 0.9.2 directly, like `oo7`, so we'd need to update anyway unfortunately.

apps/desktop/desktop_native/deny.toml
dani-garcia: Strangely using either id or crate seems to work for me:
```
  { id = "RUSTSEC-2026-0097", reason = "" },
  { crate = "rand", reason = "" },
```

but both combined gives me an error:
```
  { id = "RUSTSEC-2026-0097", crate = "rand", reason = "" },
```

```
❯ cargo deny --log-level error --all-features check all
error[unexpected-keys]: found 1 unexpected keys, expected: ["id", "reason"]
   ┌─ /Users/dani/Desktop/Code/clients/apps/desktop/desktop_native/deny.toml:10:31
   │
10 │   { id = "RUSTSEC-2026-0097", crate = "rand", reason = "" },
   │                               ─────

2026-04-13 16:25:24 [ERROR] failed to deserialize config from '/Users/dani/Desktop/Code/clients/apps/desktop/desktop_native/deny.toml'
```

So I guess it's either one or the other? I've added the ID to match the other ignore, but I can switch to crate-only if you think it's more secure.

neuronull: Ahh you're right. I thought they could be combined.

I think what you have is the way to go 👍

neuronull (approved):

djsmith85 (approved):

PR #20136, opened 2026-04-13
[PM-34918] use sdk for collection decryption

JaredScar:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-34918

## 📔 Objective

Implements code implementation for SDK encryption and decryption for collections page


JaredScar:
**Note:** Still need to implement the encryption function in the SDK. Will do on Monday of next week ideally

sven-bitwarden:
@JaredScar Build failing

JaredScar:
> @JaredScar Build failing

@sven-bitwarden Failures are expected until https://github.com/bitwarden/sdk-internal/pull/976 is merged in 😬

libs/admin-console/src/common/collections/services/default-collection-encryption.service.ts
```diff
@@ -0,0 +1,97 @@
...
+        }),
+        catchError((error: unknown) => {
+          this.logService.error(`Failed to decrypt collection: ${error}`);
+          throw error;
+        }),
+      ),
+    );
+  }
+
+  async decryptMany(collections: Collection[], userId: UserId): Promise<CollectionView[]> {
```
sven-bitwarden, line 27: I don't immediately see a big difference - can we consolidate decrypt to call decryptMany? This way there's no chance of accidental variance between the two implementations

libs/common/src/admin-console/models/collections/collection.ts
```diff
@@ -76,6 +78,49 @@ export class Collection extends Domain {
...
+        : ("" as CollectionId),
+      organizationId: uuidAsString(sdkCollection.organizationId) as OrganizationId,
+      name: new EncString(sdkCollection.name as unknown as string),
+    });
+    collection.externalId = sdkCollection.externalId;
+    collection.hidePasswords = sdkCollection.hidePasswords;
+    collection.readOnly = sdkCollection.readOnly;
+    collection.manage = sdkCollection.manage;
+    collection.defaultUserCollectionEmail = sdkCollection.defaultUserCollectionEmail;
+    collection.type =
```
sven-bitwarden: This isn't a safe assumption - we should be exhaustive with the mapping, otherwise when we add new collection types, areas like this may silently break. We could add a test as well that verifies that every member of CollectionTypes is mapped as well

libs/common/src/admin-console/models/collections/collection.view.ts
```diff
@@ -171,6 +173,61 @@ export class CollectionView implements View, ITreeNodeObject {
...
+      name: sdkView.name,
+    });
+
+    view.externalId = sdkView.externalId;
+    view.hidePasswords = sdkView.hidePasswords;
+    view.readOnly = sdkView.readOnly;
+    view.manage = sdkView.manage;
+    view.assigned = true;
+    view.defaultUserCollectionEmail = sourceCollection.defaultUserCollectionEmail;
+    view.type =
```
sven-bitwarden: Same concern with the mapping bit

libs/common/src/admin-console/models/collections/collection.ts
sven-bitwarden: I ended up doing something similar in our database queries on the server - we had a lot of queries assume one or the other, and had to change it to be a direct comparison

sven-bitwarden (dismissed):

sven-bitwarden (dismissed):

libs/vault/src/cipher-form/components/new-item-nudge/new-item-nudge.component.ts
```diff
@@ -21,11 +21,14 @@ import { CipherType } from "@bitwarden/sdk-internal";
...
   readonly configType = input.required<CipherType | null>();
   activeUserId$ = this.accountService.activeAccount$.pipe(getUserId);
-  showNewItemSpotlight$ = combineLatest([
-    this.activeUserId$,
-    toObservable(this.configType).pipe(map((cipherType) => this.mapToNudgeType(cipherType))),
-  ]).pipe(
-    switchMap(([userId, nudgeType]) => this.nudgesService.showNudgeSpotlight$(nudgeType, userId)),
+  showNewItemSpotlight$ = combineLatest([this.activeUserId$, toObservable(this.configType)]).pipe(
+    switchMap(([userId, cipherType]) => {
+      if (cipherType === null) {
```
nikwithak: Consider  using `==` instead of `===` here, which will also catch `cipherType: undefined`.

nikwithak (changes requested):
One minor callout, but otherwise LGTM

nikwithak (approved):

sven-bitwarden (approved):

PR #20141, opened 2026-04-13
[PM-35055] fix account recovery policy config checkbox states

kspearrin:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-35055

## 📔 Objective

When the account recovery policy is not turned on, the "auto entroll members" checkbox is disabled and unchcked.

## 📸 Screenshots
image


JimmyVo16 (changes requested):
The changes look good. Can you add some test coverage based on the Codecov report?

JimmyVo16 (approved):

PR #20142, opened 2026-04-13
[PM-34910] badges replace div with span and set a tab index

voommen-livefront:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-34910

## 📔 Objective

`<bit-badge>` or `span[bitBadge]` uses a `div` which makes it non-focusable. Replaced this with a `span`
Also there was no tab index - and now this is set to 0
An `aria-label` is provided to set a text, if the text on the badge is insufficient for a screen reader

## 📸 Screenshots

** changes to the bit-badge **
https://github.com/user-attachments/assets/20077988-701b-4a89-8a66-b0033b93dc59

** usage in the reports **
https://github.com/user-attachments/assets/891d8489-7e73-40f2-baba-04c532723768


voommen-livefront:
Update from UIL. This change is not required anymore as the existing functionality of bit-badge is intentional.

PR #20147, opened 2026-04-13
PM-19944 - remove feature flag pm-19941-migrate-cipher-domain-to-sdk

nikwithak:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-19944

## 📔 Objective

Removes feature flag `pm-19941-migrate-cipher-domain-to-sdk`


nick-livefront (approved):
👏

libs/common/src/enums/feature-flag.enum.ts
```diff
@@ -71,7 +70,6 @@ export enum FeatureFlag {
   AccessIntelligenceNewArchitecture = "pm-31936-access-intelligence-new-architecture",
 
   /* Vault */
```
quexten, line 76: Very cool! A question on this: We still have a few usages of the TS based crypto, which makes your other PR's (for the bank account item type) much larger. Do you have plans to get rid of these usages / migrate them, so we can delete all of the TS crypto?

gbubemismith (approved):

jaasen-livefront (approved):

libs/common/src/enums/feature-flag.enum.ts
nikwithak: Yes we plan to eventually move all the TS crypto calls over to the SDK. Once we finish migrating all of the operations over, we'll iterate to see what spots are still using the TS crypto and build a plan to migrate them.

PR #20159, opened 2026-04-14
[PM-35138] Update targeting rules data on vault sync

jprusik:
## 🎟️ Tracking

[PM-35138](https://bitwarden.atlassian.net/browse/PM-35138)

## 📔 Objective

We want to decouple fetching all data for targeting rules from checking for changes in the data. This enables us to check for data updates more aggressively (or on-demand) without incurring potentially costly payload transport.

[PM-35138]: https://bitwarden.atlassian.net/browse/PM-35138?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

Note: this update depends on [PM-35128](https://bitwarden.atlassian.net/browse/PM-35128) being implemented by the data provider

[PM-35128]: https://bitwarden.atlassian.net/browse/PM-35128?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/browser/src/autofill/services/targeting-rules-data.service.ts
+++ b/apps/browser/src/autofill/services/targeting-rules-data.service.ts
@@ -2,11 +2,11 @@ import {
   catchError,
   defer,
   EMPTY,
-  exhaustMap,
   firstValueFrom,
   from,
   retry,
   Subject,
+  switchMap,
   takeUntil,
   tap,
   timer,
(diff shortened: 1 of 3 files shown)


apps/browser/src/autofill/services/targeting-rules-data.service.ts
```diff
@@ -117,12 +136,12 @@ export class TargetingRulesDataService {
     this._destroy$.complete();
   }
 
-  private _backgroundUpdate() {
+  private _backgroundUpdate(skipCacheCheck = false) {
```
MGibson1: **nit**: parameter name implies a force download, but you're just skipping age check

apps/browser/src/autofill/services/targeting-rules-data.service.ts
```diff
@@ -176,22 +195,51 @@ export class TargetingRulesDataService {
...
+      throw new Error(
+        `Failed to fetch manifest: ${manifestResponse.status} ${manifestResponse.statusText}`,
+      );
+    }
 
-    const response = await this.apiService.nativeFetch(new Request(sourceUrl.href));
+    const manifest = await manifestResponse.json();
+
+    // Locate the forms map entry for our target version
+    const formsEntry = manifest?.maps?.forms?.[TARGET_FORMS_VERSION];
```
MGibson1, line 224: do you want to be able to fallback to earlier versions?

MGibson1 (commented):
This looks good to me for now. I do think there's improvement opportunities with fallback manifest version handling, but not necessary, yet.

apps/browser/src/autofill/services/targeting-rules-data.service.ts
jprusik: Earlier major versions of the schema? No, the client code can't inherently handle breaking changes, which is why I thought it might be a good idea to pin the schema - if a new version of a schema emerges, we have time to migrate our code before flipping to it.

MGibson1 (approved):

apps/browser/src/autofill/services/targeting-rules-data.service.ts
jprusik: addressed d69fe4b469d2ba27edf4dc2d8ca6f09b800df350

bensbits91 (approved):

PR #20170, opened 2026-04-15
Add new SendApiServer that uses the SDK

adudek-bw:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-31068

## 📔 Objective

Switches between the default SendAPIService and SendSdkApiService based on if the feature flag is present.

Needed additional SDK updates: https://github.com/bitwarden/sdk-internal/pull/961


libs/common/src/tools/send/services/send-api-service.selector.ts
```diff
@@ -0,0 +1,119 @@
...
+
+  async postSendFile(sendId: string, fileId: string, data: FormData): Promise<any> {
+    return (await this.getService()).postSendFile(sendId, fileId, data);
+  }
+
+  async putSend(id: string, request: SendRequest): Promise<SendResponse> {
+    return (await this.getService()).putSend(id, request);
+  }
+
+  async putSendRemovePassword(id: string): Promise<SendResponse> {
```
harr1424, line 138: 🎨 Since this is V2, maybe we should document that this will in fact remove all auth from a Send

https://github.com/bitwarden/server/blob/44b818acc69ec7eb52d2cdb89ae3d3bec46a29f4/src/Api/Tools/Controllers/SendsController.cs#L439-L463

libs/common/src/tools/send/services/send-sdk-api.service.ts
```diff
@@ -0,0 +1,534 @@
...
+        catchError((error: unknown) => {
+          this.logService.error(`Failed to delete send: ${error}`);
+          throw error;
+        }),
+      ),
+    );
+    await this.sendService.delete(id);
+  }
+
+  async removePassword(id: string): Promise<any> {
```
harr1424, line 125: 🎨 Consider adding documentation regarding absolute auth removal

libs/common/src/tools/send/services/send-sdk-api.service.ts
```diff
@@ -0,0 +1,534 @@
...
+  async postSendFile(sendId: string, fileId: string, data: FormData): Promise<any> {
+    return this.apiService.send("POST", "/sends/" + sendId + "/file/" + fileId, data, true, false);
+  }
+
+  async putSend(id: string, request: SendRequest): Promise<SendResponse> {
+    const r = await this.apiService.send("PUT", "/sends/" + id, request, true, true);
+    return new SendResponse(r);
+  }
+
+  async putSendRemovePassword(id: string): Promise<SendResponse> {
```
harr1424: 🎨 Consider documentation and error message changes to reflect this will remove all auth

harr1424 (dismissed):

harr1424:
@claude please review

libs/common/src/tools/send/services/send-sdk-api.service.ts
```diff
@@ -0,0 +1,494 @@
...
+      ),
+    );
+  }
+
+  private async uploadFile(
+    createResult: SdkCreateFileSendView,
+    userId: UserId,
+    file: ArrayBuffer,
+  ): Promise<void> {
+    const encryptedBytes = new Uint8Array(file);
```
harr1424: ❓ This was flagged during the Claude review, can you explain why it is in fact safe?  As i review this, I think the finding is a legitimate one, see comment immediately below in this file.

libs/common/src/tools/send/services/send-sdk-api.service.ts
```diff
@@ -0,0 +1,494 @@
...
+        await firstValueFrom(
+          this.sdkService.userClient$(userId).pipe(
+            switchMap(async (sdk) => {
+              if (!sdk) {
+                throw new Error("SDK not available");
+              }
+              using ref = sdk.take();
+              await ref.value
+                .sends()
+                .upload_send_file(sendId, fileId, encryptedFileName, encryptedBytes);
```
harr1424: ⚠️ `encryptedBytes` is a new key here? This seems to contradict your doc comments [here](https://github.com/bitwarden/sdk-internal/blob/c5fd548658767646c10deba6be89b79aae875c87/crates/bitwarden-send/src/create_file_send.rs#L169)

libs/common/src/tools/send/services/send-sdk-api.service.ts
```diff
@@ -0,0 +1,494 @@
...
+            }
+          : null,
+    };
+  }
+
+  private sdkSendViewToSendResponse(view: SdkSendView): SendResponse {
+    return new SendResponse(this.sdkSendViewToRawJson(view));
+  }
+
+  // FIXME: SendData stores encrypted field values from the server response, but the SDK
```
harr1424: ❓ Wasn't this also flagged during the Claude review? 

>The "transitional limitation" comment is not enough; this is a correctness bug under the feature flag, not just a tech-debt note. At minimum: gate the upsert path so SendApiServiceSelector.save() can't write SDK-shaped SendData into the same InternalSendService that SendApiService writes encrypted data into. Better: add an SDK-only state slice, or only call upsert after re-encrypting.

Storing expected cipher-text as plain-text will throw when [EncryptService.decryptBytes](https://github.com/bitwarden/clients/blob/3a56f2e832ac373a038408466690449a77b3ea58/libs/common/src/key-management/crypto/abstractions/encrypt.service.ts#L58) is passed anything other than an `EncString` won't it?   

The Claude review outlines the flow, if you trace it, it's accurate, and will result in the situation described above. `decrypt()` expects an `EncString`.

libs/common/src/tools/send/services/send-sdk-api.service.ts
```diff
@@ -0,0 +1,494 @@
...
+        catchError((error: unknown) => {
+          this.logService.error(`Failed to get send: ${error}`);
+          throw error;
+        }),
+      ),
+    );
+    return this.sdkSendViewToSendResponse(view);
+  }
+
+  async postSendAccess(
```
harr1424: Confirmed critical finding in Claude review that seems intentionally ignored, is there some reason why? 

>postSendAccess, postSendAccessV2, getSendFileDownloadData, getSendFileDownloadDataV2 accept apiUrl and ignore it. The legacy implementation honors it (send-api.service.ts:37-114) — it's used so a recipient on the US instance can open a Send hosted on the EU instance (or self-hosted), see how the param flows in. If the SDK can't accept a target URL, this needs to be fixed in the SDK layer; otherwise cross-region/cross-instance Send access silently breaks under the flag.

libs/common/src/tools/send/services/send-sdk-api.service.ts
```diff
@@ -0,0 +1,494 @@
...
+    id: string,
+    request: SendAccessRequest,
+    apiUrl?: string,
+  ): Promise<SendAccessResponse> {
+    const sdk: PasswordManagerClient = await firstValueFrom(this.sdkService.client$);
+    const view = await sdk.sends().access_send_v1(id, request.password ?? undefined);
+    return new SendAccessResponse(view);
+  }
+
+  async postSendAccessV2(
```
harr1424: Confirmed critical finding in Claude review that seems intentionally ignored, is there some reason why? 

>postSendAccess, postSendAccessV2, getSendFileDownloadData, getSendFileDownloadDataV2 accept apiUrl and ignore it. The legacy implementation honors it (send-api.service.ts:37-114) — it's used so a recipient on the US instance can open a Send hosted on the EU instance (or self-hosted), see how the param flows in. If the SDK can't accept a target URL, this needs to be fixed in the SDK layer; otherwise cross-region/cross-instance Send access silently breaks under the flag.

libs/common/src/tools/send/services/send-sdk-api.service.ts
```diff
@@ -0,0 +1,494 @@
...
+        }),
+        catchError((error: unknown) => {
+          this.logService.error(`Failed to delete send: ${error}`);
+          throw error;
+        }),
+      ),
+    );
+  }
+
+  async getSendFileDownloadData(
```
harr1424, line 209: Confirmed critical finding in Claude review that seems intentionally ignored, is there some reason why? 

>postSendAccess, postSendAccessV2, getSendFileDownloadData, getSendFileDownloadDataV2 accept apiUrl and ignore it. The legacy implementation honors it (send-api.service.ts:37-114) — it's used so a recipient on the US instance can open a Send hosted on the EU instance (or self-hosted), see how the param flows in. If the SDK can't accept a target URL, this needs to be fixed in the SDK layer; otherwise cross-region/cross-instance Send access silently breaks under the flag.

libs/common/src/tools/send/services/send-sdk-api.service.ts
```diff
@@ -0,0 +1,494 @@
...
+    apiUrl?: string,
+  ): Promise<SendFileDownloadDataResponse> {
+    const sdk: PasswordManagerClient = await firstValueFrom(this.sdkService.client$);
+    const data = await sdk
+      .sends()
+      .get_file_download_data_v1(send.id, send.file.id, request.password ?? undefined);
+    return new SendFileDownloadDataResponse(data);
+  }
+
+  async getSendFileDownloadDataV2(
```
harr1424, line 222: Confirmed critical finding in Claude review that seems intentionally ignored, is there some reason why? 

>postSendAccess, postSendAccessV2, getSendFileDownloadData, getSendFileDownloadDataV2 accept apiUrl and ignore it. The legacy implementation honors it (send-api.service.ts:37-114) — it's used so a recipient on the US instance can open a Send hosted on the EU instance (or self-hosted), see how the param flows in. If the SDK can't accept a target URL, this needs to be fixed in the SDK layer; otherwise cross-region/cross-instance Send access silently breaks under the flag.

harr1424 (changes requested):
I've stopped reviewing after confirming none of the "critical" findings from the Claude review have been addressed. At first, I thought they must not be valid, but I've confirmed all of them, is there some reason why you didn't address any of them?

adudek-bw:
Hey @harr1424 — sorry about the timing crossover. I pushed f33ba2dbed (and follow-up 3b58b788c7 with routing tests). All three critical findings should be addressed on the current branch:
  
  - File-send key mismatch — selector routes new file sends to legacy unconditionally; the SDK file-upload path is removed.
  - Plaintext into InternalSendService — SDK mutations now refetch the wire-encrypted form via legacy getSend before upserting; read methods (getSend, getSends, putSendRemovePassword) route to legacy unconditionally; the broken sdkSendViewToSendData/SendResponse helpers are deleted.
  - apiUrl ignored — selector routes apiUrl-bearing access calls to legacy; the SDK methods no longer advertise the param.

  Most of the threads you opened are now marked outdated by GitHub because the line numbers moved. Also picked up the V2 auth-removal docs from your April comments, added explicit throws in buildSendAuth/buildSendViewType for invalid payloads, cached the feature-flag observable on the selector, and handled the password-preserve and post-mutation-refetch-failure edge cases.

  Could you take another look?

libs/common/src/tools/send/services/send-sdk-api.service.ts
```diff
@@ -0,0 +1,337 @@
...
+    // SDK's create_file_send generates its own send key; a pre-encrypted file buffer
+    // from the caller wouldn't match. SendApiServiceSelector routes new file sends to
+    // the legacy service — this guard catches direct callers.
+    if (send.id == null && send.type === SendType.File) {
+      throw new Error(
+        "SendSdkApiService does not support creating file sends; use the legacy SendApiService.",
+      );
+    }
+    const sendView = await send.decrypt(userId);
+    await this.preserveExistingPasswordOnEdit(sendView);
```
harr1424: ❓ Is this meant to copy the password hash from an existing password protected Send into the edited version? 

❓ If so, is there logic downstream from here that will result in the hash being re-hashed?

libs/common/src/tools/send/services/send-sdk-api.service.ts
harr1424: ❓ Why did these functions expect `apiUrl` originally? Was removing this parameter the ideal approach, or should it have been used within the functions? 

EDIT: I think I understand now, the selector service you wrote handles this now, correct?

libs/common/src/tools/send/services/send-sdk-api.service.ts
```diff
@@ -0,0 +1,337 @@
...
+    return Promise.reject(
+      new Error(
+        "SendSdkApiService.getSend is not supported: returning SendResponse requires wire-encrypted fields the SDK does not expose. Use SendApiService.",
+      ),
+    );
+  }
+
+  // `apiUrl` is intentionally omitted: the SDK uses its configured environment and
+  // cannot target a per-call URL. SendApiServiceSelector routes `apiUrl`-bearing
+  // callers (e.g. the CLI receiving cross-instance Sends) to the legacy service.
```
harr1424: ❓ What is a cross-instance Send?

libs/common/src/tools/send/services/send-sdk-api.service.ts
```diff
@@ -0,0 +1,337 @@
...
+  async postSendAccessV2(accessToken: SendAccessToken): Promise<SendAccessResponse> {
+    const sdk: PasswordManagerClient = await firstValueFrom(this.sdkService.client$);
+    const view = await sdk.sends().access_send(accessToken.token);
+    return new SendAccessResponse(view);
+  }
+
+  getSends(): Promise<ListResponse<SendResponse>> {
+    return Promise.reject(
+      new Error(
+        "SendSdkApiService.getSends is not supported: returning SendResponse requires wire-encrypted fields the SDK does not expose. Use SendApiService.",
```
harr1424: 🎨 Something like this should be placed in a doc comment, so that IDE's will display this information before  the method is used and an error returned. Would it make sense to even deprecate this, or is there some reason why it exists at all?

libs/common/src/tools/send/services/send-sdk-api.service.ts
```diff
@@ -0,0 +1,337 @@
...
+      ),
+    );
+  }
+
+  putSendRemovePassword(_id: string): Promise<SendResponse> {
+    return Promise.reject(
+      new Error(
+        "SendSdkApiService.putSendRemovePassword is not supported: returning SendResponse requires wire-encrypted fields the SDK does not expose. Use SendApiService.",
```
harr1424: 🎨 Something like this should be placed in a doc comment, so that IDE's will display this information before the method is used and an error returned. Would it make sense to even deprecate this, or is there some reason why it exists at all?

libs/common/src/tools/send/services/send-sdk-api.service.ts
```diff
@@ -0,0 +1,337 @@
...
+    if (
+      sendView.id == null ||
+      sendView.authType !== AuthType.Password ||
+      sendView.password != null
+    ) {
+      return;
+    }
+    const existing = await firstValueFrom(this.sendService.get$(sendView.id));
+    if (existing?.password) {
+      sendView.password = existing.password;
```
harr1424: ⚠️ We need to be extremely careful that whatever ends up consuming this copied hash does NOT re-hash it later

libs/common/src/tools/send/services/send-sdk-api.service.ts
harr1424: ❓ Why did these functions expect apiUrl originally? Was removing this parameter the ideal approach, or should it have been used within the functions?

libs/common/src/tools/send/services/send-sdk-api.service.ts
```diff
@@ -39,30 +29,67 @@ import { SendView } from "../models/view/send.view";
...
   async save(sendData: [Send, EncArrayBuffer]): Promise<Send> {
     const userId = await firstValueFrom(this.accountService.activeAccount$.pipe(getUserId));
-    const [send, encBuffer] = sendData;
+    const [send] = sendData;
+    // SDK's create_file_send generates its own send key; a pre-encrypted file buffer
+    // from the caller wouldn't match. SendApiServiceSelector routes new file sends to
+    // the legacy service — this guard catches direct callers.
+    if (send.id == null && send.type === SendType.File) {
+      throw new Error(
+        "SendSdkApiService does not support creating file sends; use the legacy SendApiService.",
```
harr1424: 🎨 May also be worth placing in a doc comment, the existing doc comment is very verbose and personally I wouldn't read it, can you de-Claude that into something more readable by human developers?

libs/common/src/tools/send/services/send-sdk-api.service.ts
```diff
@@ -104,89 +131,46 @@ export class SendSdkApiService implements SendApiServiceAbstraction {
...
-          return await ref.value.sends().get(asUuid<SdkSendId>(id));
-        }),
-        catchError((error: unknown) => {
-          this.logService.error(`Failed to get send: ${error}`);
-          throw error;
-        }),
+  getSend(_id: string): Promise<SendResponse> {
+    return Promise.reject(
+      new Error(
+        "SendSdkApiService.getSend is not supported: returning SendResponse requires wire-encrypted fields the SDK does not expose. Use SendApiService.",
```
harr1424: 🎨 Something like this should be placed in a doc comment, so that IDE's will display this information before the method is used and an error returned. Would it make sense to even deprecate this, or is there some reason why it exists at all?

libs/common/src/tools/send/services/send-sdk-api.service.ts
adudek-bw: You were right — confirmed by reading the SDK Rust source. `bitwarden_crypto::pbkdf2` is applied unconditionally in `SendAuthType::auth_data` (`crates/bitwarden-send/src/send.rs:160-163`), which is what `SendAddRequest`/`SendEditRequest` go through (`create.rs:92`, `edit.rs:117`). The SDK treats `password` as plaintext and always derives the keyB64 itself.

Our flow does the PBKDF2 client-side in `SendService.encrypt` (`send.service.ts:103-108`) and stores the resulting keyB64 on `Send.password`. Passing that through to the SDK gives `PBKDF2(keyB64)` — a different hash from what the access flow would produce.

`preserveExistingPasswordOnEdit` made this visible on edit-without-password-change, but the same double-hash happens for new sends and password changes under the SDK path — every password-protected save was broken. Fixed in `58cfc4f593`: the selector now routes `authType === Password` saves to the legacy service, and the broken `preserveExistingPasswordOnEdit` helper is removed. Added selector tests for the new routing.

Follow-up needed on the SDK side: either a `PreservedPassword { key_b64 }` variant on `SendAuthType` or a way to skip the internal PBKDF2 — I'll file that so password-protected sends can move back to the SDK path later.

libs/common/src/tools/send/services/send-sdk-api.service.ts
adudek-bw: Confirmed against the Rust source — your concern was correct. The SDK applies PBKDF2 to whatever string is in `SendAuthType::Password.password` (`crates/bitwarden-send/src/send.rs:160-163`), with no preserve / skip-hash mechanism on the public request types. Copying the already-hashed `Send.password` value would result in a double-hashed password landing on the server.

Same root cause as the line-70 thread: every password-protected save under the SDK flag was broken, not just the preserve case. Fix shipped in `58cfc4f593` — the selector routes `authType === Password` saves to legacy and the `preserveExistingPasswordOnEdit` helper is gone.

libs/common/src/tools/send/services/send-api-service.selector.ts
```diff
@@ -47,16 +48,29 @@ export class SendApiServiceSelector implements SendApiServiceAbstraction {
...
+   *   under a client-derived key, but the SDK's `create_file_send` generates its own key
+   *   and would leave those bytes undecryptable.
+   * - **Password-protected sends.** `SendService.encrypt` already PBKDF2-derives
+   *   `Send.password` to its wire `keyB64` form, but the SDK's `SendAuthType::auth_data`
+   *   (`bitwarden-send/src/send.rs:160-163`) unconditionally re-applies PBKDF2 to the
+   *   `auth.password` string, double-hashing the value and leaving the server with a
+   *   password the user can't supply. The SDK has no `PreservedPassword`/skip-hash
+   *   variant on `SendAddRequest`/`SendEditRequest`, so this affects every
+   *   password-protected save (create, change, preserve-on-edit).
+   */
```
harr1424, line 59: Optional: I would re-write this in my own words. It documents an iterative development process with an LLM, but if you were reading this for the first time, in a doc comment served by an IDE, would you find it helpful or useful?

libs/common/src/tools/send/services/send-sdk-api.service.ts
harr1424: ⛏️ Is this still used?

harr1424, line 83: ❓ Why `as unknown as string` ?

harr1424 (approved):
~~No blockers~~ one nit and open question below. Please consider reviewing the doc comments and writing them in your own words, from the POV of a developer reading them in an IDE during development. Again completely optional, but please see requested changes regarding null safety below.

libs/common/src/tools/send/services/send-api-service.selector.ts
harr1424: I think this was lost along the way, please consider adding a doc comment.

libs/common/src/tools/send/services/send-sdk-api.service.ts
harr1424, line 1: ~~❓ Should this file have unit test coverage?~~

Most of the logic is used to forward requests onto the SDK, which should have its own unit tests. `save()` has a lot of conditional logic but it seems to be covered in the selector unit tests.

libs/common/src/tools/send/services/send-sdk-api.service.ts
```diff
@@ -0,0 +1,332 @@
...
+
+  private buildSendAddRequest(sendView: SendView): SendAddRequest {
+    return {
+      name: sendView.name,
+      notes: sendView.notes ?? undefined,
+      viewType: this.buildSendViewType(sendView),
+      maxAccessCount: sendView.maxAccessCount ?? undefined,
+      disabled: sendView.disabled,
+      hideEmail: sendView.hideEmail,
+      deletionDate: sendView.deletionDate.toISOString(),
```
harr1424: Could this be null? How will `.toISOString()` handle that? 

https://github.com/bitwarden/clients/blob/c8d507757d7eda29bd62df432600d2001411955d/libs/common/src/tools/send/models/view/send.view.ts#L27

libs/common/src/tools/send/services/send-sdk-api.service.ts
```diff
@@ -0,0 +1,332 @@
...
+
+  private buildSendEditRequest(sendView: SendView): SendEditRequest {
+    return {
+      name: sendView.name,
+      notes: sendView.notes ?? undefined,
+      viewType: this.buildSendViewType(sendView),
+      maxAccessCount: sendView.maxAccessCount ?? undefined,
+      disabled: sendView.disabled,
+      hideEmail: sendView.hideEmail,
+      deletionDate: sendView.deletionDate.toISOString(),
```
harr1424: Could this be null? How will `.toISOString()` handle that? 

https://github.com/bitwarden/clients/blob/c8d507757d7eda29bd62df432600d2001411955d/libs/common/src/tools/send/models/view/send.view.ts#L27

harr1424 (changes requested):
Please make minor changes to address null safety ~~and unit test coverage~~, and please consider the other suggested changes, which I don't consider blockers.

libs/common/src/tools/send/services/send-sdk-api.service.ts
adudek-bw: SdkSendId = Tagged<Uuid, "SendId"> and the SDK declares Uuid = unknown; TypeScript blocks unknown → string directly, so the double-cast is the standard idiom for unbranding the SDK's opaque UUID type.

harr1424: Thank you for explaining

adudek-bw: good catch, deleting now

harr1424 (dismissed):
Looks good! Thanks for your diligence in making the changes.

harr1424 (dismissed):

harr1424 (approved):

PR #20171, opened 2026-04-15
[PM-33314] Add FlightRecorder at @bitwarden/logging

dani-garcia:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-33314

## 📔 Objective
- Add `FlightRecorder` wrapper around the SDK's `FlightRecorderClient` at `@bitwarden/logging`
    - `FlightRecorder` can't depend on `SdkLoadService` directly because `@bitwarden/common` already depends on `@bitwarden/logging` and would create a circular dependency. Instead, callers pass `SdkLoadService.Ready` at the DI layer.

- Polyfill `TextEncoder`/`TextDecoder` in jsdom test environments: the SDK WASM module requires these globals at load time, and importing `@bitwarden/common` now transitively triggers the SDK module via the export. The polyfill is applied in two places:
  - `test.environment.ts`: for generator projects that use the custom jsdom environment directly
  - `jest.config.angular.js` via `setupFiles`: for all Angular test projects
  
- Replace `<rootDir>`-based paths in the shared jest configs (`jest.config.angular.js`, `jest.config.ts.js`) with `path.resolve(__dirname, ...)`: `<rootDir>` resolves relative to the consuming project's directory, which breaks for projects at different nesting levels.


libs/shared/test.environment.ts
```diff
@@ -24,5 +26,9 @@ export default class FixJSDOMEnvironment extends JSDOMEnvironment {
     this.global.Headers = Headers;
     this.global.Request = Request;
     this.global.Response = Response;
+
+    // SDK/WASM code relies on TextEncoder/TextDecoder being globally available
+    this.global.TextEncoder = TextEncoder;
+    this.global.TextDecoder = TextDecoder as typeof globalThis.TextDecoder;
```
djsmith85: ❓ Is this needed? This does the same as the polyfill-node-globals.ts which is loaded on https://github.com/bitwarden/clients/pull/20171/changes#diff-646376040e3c784ef3028bf3cde3599023180d1241055660e7c5228c8d8b6a62R14

libs/shared/jest.config.angular.js
```diff
@@ -16,6 +18,7 @@ const presetConfig = createCjsPreset({
 module.exports = {
   ...presetConfig,
   testMatch: ["**/+(*.)+(spec).+(ts)"],
+  setupFiles: [path.resolve(__dirname, "polyfill-node-globals.ts")],
```
djsmith85, line 22: Tests are failing and I think it might not be picking up the added polyfill.

I'm also not sure about using path.resolve here. It changes how nested projects detect locations.

Maybe it also needs to be `<rootDir>/../../libs/shared/polyfill-node-globals.ts`

djsmith85 (changes requested):

libs/shared/test.environment.ts
dani-garcia: I think the reason I added that was because there's two jest config files in `libs/tools/generator` that weren't extending the shared one here, for now I added the `polyfill-node-globals.ts` polyfill to those two directly, but maybe they should be inheriting from the shared one?

libs/shared/jest.config.angular.js
dani-garcia: I've tried that but it doesn't seem to work for the angular one, I get an error like: 
```
npm run test

> @bitwarden/clients@0.0.0 test
> jest

● Validation Error:

  Module <rootDir>/../../libs/shared/polyfill-node-globals.ts in the setupFiles option was not found.
         <rootDir> is: /Users/dani/Desktop/Code/clients/libs/dirt/card

  Configuration Documentation:
  https://jestjs.io/docs/configuration
```

Strangely, that same thing works perfectly fine on the non-angular `jest.config.ts.js`, so maybe I'm missing something 🤔

dani-garcia: Okay this seems to be because this library is nested one folder deeper than the rest: https://github.com/bitwarden/clients/tree/main/libs/dirt/card, which uses the angular shared config and so it's causing the path resolution to fail (that library would need an extra `../`)

Seems like the reason it hasn't failed before is that the library doesn't have any spec files, and so the other `<rootDir>/../../libs/shared/es2020-transformer.ts` above probably doesn't get evaluated?

djsmith85 (approved):

PR #20173, opened 2026-04-15
[PM-34751] inline menu autofill bypasses untrusted iframe check

dan-livefront:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-34751
## 📔 Objective

To achieve feature parity, add an alert for cross-origin forms when their URLs differ. Additionally, update the alert to display the full href, including ports, to provide greater clarity for users.

I avoided recommendations due to the short circuit giving a dead inline menu that silently short circuited. 

> [!NOTE]
> Additional work will follow up this which is being tracked here https://bitwarden.atlassian.net/browse/PM-35071

## 📸 Screenshots

Screenshot 2026-04-15 at 8 10 08 AM

### DIFF
--- a/apps/browser/src/autofill/services/autofill.service.ts
+++ b/apps/browser/src/autofill/services/autofill.service.ts
@@ -1566,8 +1566,17 @@ export default class AutofillService implements AutofillServiceInterface {
       return false;
     }
 
-    // Check the pageUrl against cipher URIs using the configured match detection.
-    // Remember: if we are in this function, the tabUrl already matches a saved URI for the login.
+    // Different origin than the top level tab so we will treat as untrusted.
+    try {
+      if (new URL(pageUrl).origin !== new URL(options.tabUrl).origin) {
+        return true;
+      }
+    } catch {
+      // Fall through to legacy check for unparseable URLs
+    }
+
+    // Otherwise, check the pageUrl against cipher URIs using the configured match detection.
+    // Remember: if we are in this part of the function, the tabUrl already matches a saved URI for the login.
     // We need to verify the pageUrl also matches.
     const equivalentDomains = await firstValueFrom(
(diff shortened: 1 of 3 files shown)


apps/browser/src/autofill/services/autofill.service.ts
```diff
@@ -1566,8 +1566,17 @@ export default class AutofillService implements AutofillServiceInterface {
...
+    try {
+      if (new URL(pageUrl).origin !== new URL(options.tabUrl).origin) {
+        return true;
+      }
+    } catch {
+      // Fall through to legacy check for unparseable URLs
+    }
+
+    // Otherwise, check the pageUrl against cipher URIs using the configured match detection.
+    // Remember: if we are in this part of the function, the tabUrl already matches a saved URI for the login.
```
jprusik, line 1579: ...or was not parseable. We might want an early check/validation/return instead of the try/catch

apps/browser/src/autofill/services/insert-autofill-content.service.ts
```diff
@@ -112,7 +112,7 @@ class InsertAutofillContentService implements InsertAutofillContentServiceInterf
 
     const confirmationWarning = [
       chrome.i18n.getMessage("autofillIframeWarning"),
-      chrome.i18n.getMessage("autofillIframeWarningTip", [globalThis.location.hostname]),
+      chrome.i18n.getMessage("autofillIframeWarningTip", [globalThis.location.href]),
```
jprusik: I think we want `protocol + host + pathname`; `href` is going to include query string and hash values, which is probably too precise for a matching pattern and too verbose for a dialog

jprusik (changes requested):
On hold, as we verify Product expectations of some additional scenarios

PR #20186, opened 2026-04-15
[PM-32019] add diamond action chip to archive menu option in desktop

jengstrom-bw:
## 🎟️ Tracking

[Jira](https://bitwarden.atlassian.net/browse/PM-32019)

## 📔 Objective

Replace the sidebar `Upgrade` chips to a bitIconButton.

## 📸 Screenshots

Screenshot 2026-04-22 at 1 37 16 PM


apps/desktop/src/billing/components/premium-badge-desktop.component.ts
jengstrom-bw, line 1: Let me know if this file should live somewhere else, put it here since `premium-badge.component.ts` is in `/libs/angular/src/billing/components/premium-badge/premium-badge.component.ts`

libs/components/src/chips/chip-action/chip-action.component.html
jengstrom-bw, line 1: updated `chip-action` and `chip-content` to allow for an icon only chip without having extra space behind the icon. If there is a better way to accomplish please let me know.

BryanCunningham: 👋 Hey @jengstrom-bw , Icon only chips are intentionally not supported. Happy to hop on a call or Slack thread with you and design to figure out a path forward.

Current options would be:
- Leave this as a chip and have a visible label
- Use a size small icon button. This option would require a small design change as that color style is not a supported icon button variant.

willmartian: (I just looped in Missy on this as well.)

harr1424 (dismissed):
Changes to Tools owned component look good

apps/desktop/src/billing/components/premium-badge-desktop.component.ts
cyprain-okeke: Nice work on this component! but I looked into whether we could extend the existing `PremiumBadgeComponent` in `libs/angular/src/billing/components/premium-badge/` instead of creating a separate   desktop-specific one, and I think we can do it safely.                                                                                                                                                                                                                                                                                                                                   
Every existing consumer (18 usages across web, browser, desktop, and shared libs) uses `<app-premium-badge>` with no inputs other than the occasional `organizationId` — no one passes `variant`,  label, or tooltip. So if we add those as optional inputs with defaults matching the current behavior, nothing changes for existing consumers:                                                                                                                                                                                             
  ```
readonly variant = input<ChipVariant>("accent-primary");  // existing behavior                                                                                                               
  readonly label = input<string>("upgrade");                 // existing behavior  
  readonly tooltip = input<string>("");                      // no tooltip, same as now  
```                                                                                                                                                                                                                                                                                                  
 Then the desktop sidebar usages become:
  `<app-premium-badge variant="primary" label="" tooltip="upgradeToPremium" /> `                                                                                                                                                                                                                                                                                                           
This avoids maintaining two components with identical business logic that could drift apart over time. It also lets us apply the `OnPush` fix (CL-764) and the `inject()` modernization in one  place.                                                                                                                                                                                              
_What do you think? Happy to discuss if there's a design reason to keep them separate._

cyprain-okeke (changes requested):

apps/desktop/src/billing/components/premium-badge-desktop.component.ts
jengstrom-bw: I like that change! Design is looking at exactly which component we need to use, because a `chip-action` must have a label. If it is possible to reuse the existing component I will, but if it uses something totally different component-wise it might need to stay an independent file. Appreciate the suggestion!

jengstrom-bw: Had to change a little more than just the variant, label, and tooltip. I consolidated with the existing component but if the changes are enough to want to split out to its own component, please let me know.

nick-livefront (approved):

cyprain-okeke (dismissed):

harr1424 (dismissed):

harr1424 (approved):

nick-livefront (approved):

cyprain-okeke (approved):

PR #20189, opened 2026-04-15
[PM-33933] Update autofill dialog for 'never' match setting

nikwithak:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-33933

## 📔 Objective

Updates the Autofill modal to give special messaging when the Autofill criteria is set to "Never". Note: This is rebased onto #20071 , and consumes several style changes included in that PR, so that PR will have to go before this one can merge in.

## 📸 Screenshots

image


gbubemismith:
Looks good! just one question, if I have a cipher with `https://other.com/secret` set to Never and `https://example.com` set to Exact and then on `https://example.com/path`, the Exact rule fails so the dialog opens and isNeverStrategy() and currentUrlMatchesSavedUri() both return true, both checks are true but for different URIs, so the title, body, green callout, and Never match badge all fire for example.com. Is this intended or should the "Never" branch check that the URI matching the current hostname is itself the Never one

https://github.com/user-attachments/assets/f7b69832-4d20-4448-b93b-7be56e2fa148

nikwithak:
Great catch @gbubemismith! Fixed in [4e7c4fa](https://github.com/bitwarden/clients/pull/20189/changes/4e7c4fac7b58b60dd8f4196e0352819ea9dcc9e7)

gbubemismith (dismissed):

gbubemismith (approved):

PR #20197, opened 2026-04-16
[PM-33315] Create @bitwarden/logging

dani-garcia:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-33315

## 📔 Objective

Create the `@bitwarden/logging-angular` library, which contains an angular-specific wrapper of `FlightRecorder` for dependency injection.


djsmith85 (approved):
Changes are looking good!

PR #20209, opened 2026-04-16
[PM-35246] Fix IdentityTokenResponse kdfConfig error

rr-bw:
## 🎟️ Tracking

[PM-35246](https://bitwarden.atlassian.net/browse/PM-35246)

## 📔 Objective

The changes in [bitwarden/clients#20058](https://github.com/bitwarden/clients/pull/20058) have uncovered a pre-existing bug/gap on the `IdentityTokenResponse`.

When constructing a `new IdentityTokenResponse()` from a server response, it is possible for the `Kdf` on the response to be absent (refresh token scenario).

The existing ternary operator would then see that the `kdf` is not `KdfType.PBKDF2_SHA256`, so it would fall back and try to construct `new Argon2KdfConfig()` with nullish properties. But as of https://github.com/bitwarden/clients/pull/20058 , the `PBKDF2KdfConfig()` and `Argon2KdfConfig()` constructors now requires all arguments to be explicitly passed/non-nullish - which catches this bug.

This PR creates a separate `RefreshAccessToken` class to handle a refresh token scenario, containing just the fields on a refresh token. The `ApiService.refreshAccessToken()` method then builds this `new RefreshAccessToken()` instead of a `new IdentityTokenResponse()`.

This PR also updates to the `IdentityTokenResponse` to use a `switch` on the `kdf` response to more specifically handle each `KdfType`. The `switch` includes a `default` defensive runtime check that throws if a `kdf` was not found on the response.

## 📸 Screenshots

https://github.com/user-attachments/assets/e587e565-a16b-4672-ae0c-64394f2ada6a

[PM-35246]: https://bitwarden.atlassian.net/browse/PM-35246?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/auth/src/common/login-strategies/login.strategy.ts
```diff
@@ -231,6 +231,10 @@ export abstract class LoginStrategy {
       tokenResponse.refreshToken, // Note: CLI login via API key sends undefined for refresh token.
     );
 
+    if (tokenResponse.kdfConfig == null) {
```
mzieniukbw: ❓ Do we need to apply the same in [sso-login.strategy.ts](https://github.com/bitwarden/clients/blob/8730a27b9e98868bd3ab42e54de9bc5f8776b5c9/libs/auth/src/common/login-strategies/sso-login.strategy.ts#L137) ?

mzieniukbw (dismissed):

libs/auth/src/common/login-strategies/login.strategy.ts
rr-bw: This change is now outdated as of commit https://github.com/bitwarden/clients/pull/20209/changes/bcf68b275b965acba8d0019ab4c989b9c86b687f - which makes the property required again on `IdentityResponseToken` and adds a defensive runtime check to throw if it isn't defined.

JaredSnider-Bitwarden (approved):
Thank you for your work to fix this! Excellent job!

PR #20216, opened 2026-04-16
[PM-35290] feat: auto-prefix https:// and improve validation UX for self-hosted environment URLs

sander-adamse:
## 🎟️ Tracking

Users in our pilot kept running into issues setting up the self-hosted environment, because they forgot to include `https://`.

## 📔 Objective

The `https://` prefix is basically always required for self-hosted setups, so having users type it manually just adds friction. When you type a bare domain like `passwords.company.com`, the validator fires immediately telling you to use `HTTPS` — but that's exactly what it's going to be anyway. Updated the validator to only show an error when `http://` is explicitly typed, and added auto-prefixing on save so the protocol gets added automatically if it's missing.

### Input
{20319345-9656-4DDD-8176-5660F2EFA302}

---

### Output
{BBEC1F65-FA78-44BD-8B8F-1DE47BA61E3E}


djsmith85 (commented):
Hey @sander-adamse thank you for your contribution.

Usually we don't accept any changes to `messages.json` other than `en/messages.json` as that is the source for all translations, in this case I'm fine with the changes as it's only removing the https-prefix in the hint and would retain the existing translations.

As @bitwarden/team-platform-dev was tagged because of these changes I'd be okay with approving this, but I'll leave it to Product management and @bitwarden/team-auth-dev, which might have further requested changes, before doing so.

djsmith85:
🎨  An alternative solution could be to use `<span bitPrefix>https://</span>` on the input fields which would simplify the overall approach and might not require changing the hints 🤷 ?

PR #20235, opened 2026-04-17
[PM-35224] Add route provider parameters support to featureFlaggedRoute

Banrion:
## 🎟️ Tracking

[[Client] Add route provider parameters support to featureFlaggedRoute](https://bitwarden.atlassian.net/browse/PM-35224)

## 📔 Objective

Extends `featureFlaggedRoute` to support scoped service providers on each side of a feature flag. Previously, any services exclusive to the flagged or default experience had to be registered at the module level and were instantiated unconditionally. Route-level providers create a child injector Angular only activates when the route is entered, enabling true lazy initialization.

**Changes:**

- Adds optional `flaggedRouteProviders` and `defaultRouteProviders` to `FeatureFlaggedRouteConfig`
- Adds a second `@example` showing scoped-provider usage
- Fully backward compatible, existing callers are unaffected


libs/angular/src/platform/utils/feature-flagged-route.ts
```diff
@@ -37,17 +44,54 @@ type FeatureFlaggedRouteConfig = {
...
+ *      featureFlag: FeatureFlag.GroupsComponentRefactor,
+ *      routeOptions: {
+ *        path: "groups",
+ *        canActivate: [OrganizationPermissionsGuard],
+ *      },
+ *      defaultRouteProviders: [
+ *        // Only instantiated when the feature flag is OFF
+ *        safeProvider({ provide: LegacyGroupsService, useClass: LegacyGroupsService, deps: [...] }),
+ *      ],
+ *      flaggedRouteProviders: [
```
djsmith85, line 63: ❓ What's the specific use case for this? Couldn't the existing and flagged component just create their own providers? As for example: 
https://github.com/bitwarden/clients/blob/main/bitwarden_license/bit-web/src/app/dirt/reports/member-access-report/member-access-report.component.ts#L59

Then using this or the https://github.com/bitwarden/clients/blob/main/libs/angular/src/platform/guard/feature-flag.guard.ts would be sufficient.

I still wonder why we have two of these 🤷

Banrion: An example is in this pull request for Access Intelligence refactoring in the [routing module](https://github.com/bitwarden/clients/pull/20081/changes#diff-2ac71d4288d0a0759bb21345ecd31627cbff0ebf9b1341eebc4dfbd3a3c3cf8f).

Two main reason: performance and unintended background execution

Sever Access Intelligence services have been refactored that start observable pipelines in the constructor. Route level providers scope the instantiation to when that route is actually activated, so only when the feature flag is on. Many of these services are also needed in many components rather than just one. The main service that benefits from this approach is the `AccessIntelligenceDataService` replacing the `RiskInsightsOrchestratorService`.

We could work around this by changing the services by exposing an `initialize` method to start the service, but I feel this adds a lot of overhead. The proposed change felt clean and clear to me

djsmith85 (approved):

PR #20243, opened 2026-04-17
[BEEEP] Improve speed performance of Admin Console Collections page

JaredScar:
## 🎟️ Tracking

N/A

## 📔 Objective

Optimize performance of vault component and services by running heavy operations outside Angular's zone. Implement chunked decryption for ciphers and collections to prevent UI blocking. Introduce a long-lived Web Worker for managing the Lunr search index, improving search responsiveness and reducing main thread load. Update virtual scrolling cache size in vault items component for better rendering efficiency.

## 📸 Screenshots

https://vimeo.com/1184239975?share=copy&fl=sv&fe=ci


eliykat (commented):
Some nice work here and good insights, particularly running work outside of the ngZone. Let's just hold for now while we wait for feedback from Platform on how multithreading is going to be handled with the SDK.

libs/common/src/vault/services/search.worker.ts
```diff
@@ -0,0 +1,184 @@
+/**
+ * Long-lived Web Worker that owns the Lunr full-text search index for the entire session.
+ *
+ * WHY LONG-LIVED: The previous design built the index in a worker then serialised it and
```
quexten, line 4: @JaredScar KM has a concurrent piece of work that eliminates search indexing entirely in most cases, and probably already achieves your goal. It also restructures the search service massively, and conflict with this work. We should sync up on how to proceed

https://github.com/bitwarden/clients/pull/19251

Unless you actually use lunr search queries, we should never build the search index, which is what the above work implements.  **Most users do not use lunr.**

Two more notes here:
- @eliykat If we do move to the SDK, we can achieve much better performance than LUNR, even without multi-threading. LUNR is just incredibly inefficient at indexing. I would always suggest optimizing the actual work first before bringing in threading
- If we do want to merge both works, I would suggest merging https://github.com/bitwarden/clients/pull/19251 first, then adapting this work to work on top of the extracted lunr service. However, I don't actually know what this is intended to achieve because https://github.com/bitwarden/clients/pull/19251 already eliminates search indexing entirely except for the first LUNR query.
- If we do care about first LUNR query performance, then I suggest that we look into re-writing LUNR queries and indexes in the SDK
  - Actually, this also applies to general LUNR query performance

However, a service worker always introduces a lot of complexity, and is probably just not needed with the approach taken in https://github.com/bitwarden/clients/pull/19251. Could you test if https://github.com/bitwarden/clients/pull/19251 already solves your issue?

quexten (changes requested):

libs/admin-console/src/common/collections/services/default-collection-admin.service.ts
```diff
@@ -134,24 +134,28 @@ export class DefaultCollectionAdminService implements CollectionAdminService {
...
-      return await CollectionAdminView.fromCollectionResponse(
-        c,
-        this.encryptService,
-        orgKeys[organizationId as OrganizationId],
+    const orgKey = orgKeys[organizationId as OrganizationId];
+
+    // Promise.all(8K) floods the microtask queue — decrypt in chunks with event-loop
+    // yields so the browser can paint and handle input between each batch.
+    const DECRYPT_CHUNK_SIZE = 2_000;
+    const results: CollectionAdminView[] = [];
```
quexten, line 142: Note: Porting state to the SDK will give better speed improvements.

libs/common/src/vault/services/search.worker.ts
quexten: I'll also note that https://github.com/bitwarden/clients/pull/19251 has been through multiple review rounds at this point, and partially through QA.

eliykat: Already discussed in Slack, but just for completeness - I agree we want to progress #19251 and rethink reintroducing multithreading per KM Team's advice.

PR #20248, opened 2026-04-17
[PM-34832] Fix account recovery strings

kspearrin:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-34832

## 📔 Objective

Fix 2FA recovery string, per spec.

### DIFF
--- a/apps/web/src/locales/en/messages.json
+++ b/apps/web/src/locales/en/messages.json
@@ -6668,7 +6668,7 @@
     }
   },
   "eventAdminPasswordReset": {
-    "message": "Master password reset for user $ID$.",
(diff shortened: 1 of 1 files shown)


kspearrin:
How should I handle the lint failure for the locale strings here? Some of this is changing strings for stuff we haven't even released yet, but its failing because I am modifying an existing string.

BTreston (changes requested):
We can't mutate existing copy strings (see CI error), these will need new keys. Also just a fyi: the reviewer team you'll want to tag in GH is `team-admin-console-dev`. That way its more visible to the team, TY 👍

PR #20254, opened 2026-04-18
Move Jest dependency ownership to Platform

trmartin4:
## 📔 Objective

As a major cross-cutting dependency, Jest should be managed by Platform, so we can apply the same cross-team migration patterns.  This PR proposes to move Jest under Platform ownership and also limit to just major updates creating PRs, as is the convention with all Platform-owned dependencies.


coroiu (approved):

cd-bitwarden (approved):
LGTM

PR #20259, opened 2026-04-20
[PM-35330] Fix state not being updated on change kdf

quexten:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-35330

## 📔 Objective

Some state values were not updated after using the KDF change service to update the KDF settings for a user. This leaves the user unable to change password.

### DIFF
--- a/libs/common/src/key-management/kdf/change-kdf.service.ts
+++ b/libs/common/src/key-management/kdf/change-kdf.service.ts
@@ -7,6 +7,7 @@ import { KdfConfig, KdfConfigService, KeyService } from "@bitwarden/key-manageme
 
 import { KdfRequest } from "../../models/request/kdf.request";
 import { SdkService } from "../../platform/abstractions/sdk/sdk.service";
+import { EncString } from "../crypto/models/enc-string";
 import { InternalMasterPasswordServiceAbstraction } from "../master-password/abstractions/master-password.service.abstraction";
 import {
   fromSdkAuthenticationData,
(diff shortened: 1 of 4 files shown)


libs/common/src/key-management/kdf/change-kdf.service.ts
```diff
@@ -67,5 +69,11 @@ export class DefaultChangeKdfService implements ChangeKdfService {
       unlockData.kdf,
     );
     await this.masterPasswordService.setMasterKey(masterKey, userId);
+    await this.masterPasswordService.setMasterPasswordUnlockData(unlockData, userId);
+    await this.masterPasswordService.setMasterKeyEncryptedUserKey(
+      new EncString(unlockData.masterKeyWrappedUserKey),
+      userId,
+    );
+    await this.kdfConfigService.setKdfConfig(userId, kdf);
```
Thomas-Avery, line 77: This makes the `setKdfConfig` in `libs/common/src/key-management/encrypted-migrator/migrations/minimum-kdf-migration.ts` redundant just for the encrypted migration. I think that is okay but it is a little weird.

Thomas-Avery (approved):
LGTM

libs/common/src/key-management/kdf/change-kdf.service.ts
quexten: I think we should pull it out of the migration. The KDF config should just always be set locally when setting the KDF and pulling the state into the migration does not make much sense.

PR #20275, opened 2026-04-20
[PM-35369] Suppress warnings 403 that forces JIT-SSO users to log out

amorask-bitwarden:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-35369

Resolves the user-visible symptom of [PM-35336](https://bitwarden.atlassian.net/browse/PM-35336) (JIT-SSO-confirmed user logged out immediately after being added to an org).

## 📔 Objective

Preflight `GET /organizations/{organizationId:guid}/billing/vnext/warnings` in `OrganizationWarningsService.readThroughWarnings$` and skip the call when the caller's access token does not carry an organization grant for the target org.

**Why:** a freshly JIT-SSO-confirmed user holds an access token that pre-dates their confirmation, so the JWT has no `orgowner`/`orgadmin`/`orgcustom`/`orguser` claim for the org. The server's `MemberOrProviderRequirement` denies the call with a 403, and `ApiService.handleApiRequestError` (libs/common) fires `logoutCallback("invalidAccessToken")` *inside* `send()` for any authenticated 401 or 403 — so downstream `catchError`/`try-catch` cannot prevent the logout. The only way to avoid the logout is to not make the call in the first place.

**The preflight**:

- Checks the four org-role claims (`orgowner`, `orgadmin`, `orgcustom`, `orguser`), accepting both the array and single-string shapes IdentityServer emits.
- Bypasses the preflight for provider users (`organization.isProviderUser`) because the server authorizes them via a DB lookup rather than JWT claims, and a preflight would incorrectly gate them out.
- Gates on `accountService.activeAccount$` and `tokenService.hasAccessToken$` so teardown races (no active user / no token) short-circuit to an empty response and log at `info` — preserving the PM-35369 case as the only `warning`-level signal.
- Does **not** cache the empty short-circuit, so a subsequent subscription after a token refresh hits the API normally.

**What this does *not* fix** (tracked separately):

- Server-side stale-token authorization (owned by Auth team) — `MemberOrProviderRequirement` should fall back to a DB membership check when the JWT claim is absent.
- Platform-side `ApiService.handleApiRequestError` collapsing 401 and 403 (owned by Platform team) — a 403 should not be session-fatal.

This PR addresses the symptom from Billing's scope so the RC cut isn't blocked by the regression while the root-cause fixes land.

## 📸 Screenshots

N/A — no UI changes. Behavior changes are observable via browser console (skip logs) and network tab (no `/warnings` call when the JWT lacks the grant).

## Test coverage

47 unit tests in `organization-warnings.service.spec.ts`:

- All four claim keys × both shapes (array and single-string)
- No grant for target org, cross-org claims
- Provider-user bypass
- No active account / no access token gates
- Fail-open on decode exception (with warning log)
- Short-circuit result is not cached (next call with grant hits the API)

Live-verified end-to-end against a fresh JIT-SSO-style account: without a grant the call is skipped (no 403, no logout); with a grant the call succeeds (200 OK).

[PM-35336]: https://bitwarden.atlassian.net/browse/PM-35336?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


kdenney (approved):
Nice fix! Looks good.

PR #20284, opened 2026-04-20
Add Windows WebAuthn Plugin FFI safe type wrappers [PM-29785]

iinuwa:
## 🎟️ Tracking

[PM-29785](https://bitwarden.atlassian.net/browse/PM-29785)

## 📔 Objective

Add wrapper types for Windows plugin authenticator

[PM-29785]: https://bitwarden.atlassian.net/browse/PM-29785?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/desktop/desktop_native/Cargo.lock
abergs, line 1: Have we double checked that nothing snuck into to the cargo.lock like last time? :)

apps/desktop/desktop_native/win_webauthn/src/api/plugin/mod.rs
```diff
@@ -3,3 +3,385 @@
...
+///     dark_theme_logo_svg: None,
+///     authenticator_info,
+///     supported_rp_ids: None,
+/// };
+/// WebAuthnPlugin::add_authenticator(&options).unwrap();
+///
+/// // Register this process to receive COM messages.
+/// let plugin = WebAuthnPlugin::new(clsid);
+/// plugin.register_server(authenticator).unwrap();
+/// ```
```
abergs, line 259: Do we really wanna keep all those comments above?

apps/desktop/desktop_native/win_webauthn/src/api/plugin/crypto.rs
```diff
@@ -275,6 +276,8 @@ impl<'a> From<&'a OwnedRequestHash> for RequestHash<'a> {
         RequestHash::new(&value.0)
     }
 }
+
+#[derive(Debug)]
```
abergs, line 280: Are we adding the Debug derives for the tracing?

abergs (approved):
Couple of nit like comments, but marking approved in case they are no-ops

apps/desktop/desktop_native/Cargo.lock
iinuwa: yeah, it's a one-line diff just adding a reference to `base64` for `win_webauthn`, which already existed at the root workspace.

apps/desktop/desktop_native/win_webauthn/src/api/plugin/crypto.rs
iinuwa: Yes, particularly for when `OwnedRequestHash` is stored inside another struct where it would be helpful to have `Debug`. The request hash is not sensitive, so I think that's OK.

apps/desktop/desktop_native/win_webauthn/src/api/plugin/mod.rs
iinuwa: I think it's helpful. It's a demonstration of how to use the API, which is useful if you generate docs or hover over in the IDE. Even if it doesn't get public usage, I think that it is clearer to see the flow of the API than the Bitwarden implementation, which can help onboarding.

Issue #20107, opened 2026-04-10
Autofill performance regression, even when autofill is disabled


toupeira:
### Summary

The autofill performance degradation from https://github.com/bitwarden/clients/issues/17090 has returned in 2026.3.0.

It originally showed up in 2025.10.0 and was fixed in 2025.12.0, and 2026.2.0 still seems good too. AFAIK the problem was introduced in 6129ca5366 and related to component-heavy sites using many shadow DOMs, and this commit was reverted in c04c1757ea for the fix.

I see there was some recent work around shadow DOMs again in f9dff532a6, along with performance specs, and while it does seem a bit less egregious than last time it's still very noticeable for me in the form of cursor lag on sites like YouTube or Reddit.

### Steps To Reproduce

Example 1:
 
1. Go to https://www.youtube.com/
2. Focus the search field and hold down a key, e.g. `i`
3. Observe the input lagging every second or so

Example 2:

1. Go to a random YouTube video
2. Move the mouse over the timeline to show the preview images
3. Observe the preview images lagging every second or so

Note: you might need to test on an older machine or throttle your CPU to see this

### Expected Result

The autofill code shouldn't degrade performance.

What really bugs me is that I don't even use autofill:

- "Show autofill suggestions on form fields" is disabled
- "Autofill on page load" is disabled

So it would be really neat if the extension could not run these expensive autofill watchers at all in this case.

I do use the fill keyboard shortcut and filling from the extension window, but surely the extension could just find the form fields **once** when those actions are triggered?

### Actual Result

The autofill code introduces lags.

### Screenshots or Videos

Firefox profile: https://share.firefox.dev/41tdVUx

The offending call stack looks similar as in https://github.com/bitwarden/clients/issues/17090#issuecomment-3470662755:

```
nodeIsElement [Extension "Bitwarden Password Manager" (ID: {446900e4-71c2-419f-a6a7-df9c091e268b}): moz-extension://<URL>:5148:23]
getShadowRoot [Extension "Bitwarden Password Manager" (ID: {446900e4-71c2-419f-a6a7-df9c091e268b}): moz-extension://<URL>:8878:18]
buildTreeWalkerNodesQueryResults [Extension "Bitwarden Password Manager" (ID: {446900e4-71c2-419f-a6a7-df9c091e268b}): moz-extension://<URL>:8920:37]
[...]

Issue #18771, opened 2026-02-05
Performance degradation over time in web applications when extension is enabled (Edge & Firefox)


gitbro23:
### Steps To Reproduce

1.	Enable the Bitwarden browser extension in Edge or Firefox.
2.	Open and work in one or more web applications.
3.	Observe performance for a while (minutes to hours).
4.	Note the slowdown in the web application's responsiveness.
5.	Close the tab and restart the session; performance is temporarily restored.

### Expected Result

The web application should maintain consistent performance regardless of whether the extension is enabled or how long the session lasts.

### Actual Result

The web application becomes increasingly slow over time. When the extension is disabled, the issue disappears completely, confirming a direct relation to the Bitwarden extension.

### Screenshots or Videos

<img width="675" height="413" alt="Image" src="https://github.com/user-attachments/assets/c0b26712-019f-4601-8780-ed29a5b6776f" />

### Operating System

Windows

### Operating System Version

Windows 11 23H2 x64

### Web Browser

Microsoft Edge

### Browser Version

latest: Version 144.0.3719.104 (64-Bit)

### Environment Versions

Version: 2025.12.1

SDK: ‘main (93a331f)’

Server version (third-party): 2025.12.0

Connected to third-party server implementation, Vaultwarden. Please verify errors with the official server or report them to the third-party server.

### Issue Tracking Info

- [x] I understand that work is tracked outside of Github. A PR will be linked to this issue should one be opened to address it, but Bitwarden doesn't use fields like "assigned", "milestone", or "project" to track progress.

NovaSilentium:
Hi there,

I am unable to reproduce this issue, it has been escalated for further investigation. If you have more information that can help us, please add it below.

Thanks!

pamperer562580892423:
Another user here.

As there were opened some similar issues here recently for the browser extensions, you probably should also check if your issue already was reported:

- https://github.com/bitwarden/clients/issues/18187
- https://github.com/bitwarden/clients/issues/17441
- https://github.com/bitwarden/clients/issues/17359
- https://github.com/bitwarden/clients/issues/17128 
- https://github.com/bitwarden/clients/issues/14143

gitbro23:
Hi!
What kind of further information should I send to you?
We can reproduce it, by using any web application. The performance is getting bad after using it a couple of minutes and we see a high memory usage for the specific tab. When the Bitwarden browser extension is disabled, this is not happening. --> less memory usage and best performance

[...]

Issue #17128, opened 2025-10-30
Slow input in jupyterlab when using extension


ilyasdc:
### Steps To Reproduce

1. Disable all extensions except for Bitwarden
2. Open a (large ?) jupyter notebook 
3. Type something (In the screncast I keep the key "a" pressed)

### Expected Result

I can type quickly

### Actual Result

 The input is very slow 

### Screenshots or Videos

 https://github.com/user-attachments/assets/32475a5b-4d6e-4c17-85bf-3f681b082ac1

### Additional Context

I have several passords attached to localhost. Can it be that bitwarden tries to parse the input field constantly ?

The same happens on chromium so I suppose tis

### Operating System

Linux

### Web Browser

Firefox

### Browser Version

144.0.2

### Environment Versions

Version: 2025.10.0

SDK: 'main (d0262d3)'

Server version: 2025.10.1

### Issue Tracking Info

- [x] I understand that work is tracked outside of Github. A PR will be linked to this issue should one be opened to address it, but Bitwarden doesn't use fields like "assigned", "milestone", or "project" to track progress.

NaLogo:
This also happens on GitHub, I noticed it on the form for creating a pull request. Don't mind the typo in the recording ;)

https://github.com/user-attachments/assets/c189ca30-0940-448d-88a0-95c77e1d27af

NovaSilentium:
Hi there,

I am unable to reproduce this issue, it has been escalated for further investigation. If you have more information that can help us, please add it below. 

Thanks!

yosinski:
+1. I'm also seeing this issue today for the first time.

Platform: Firefox on Mac
Bitwarden version: 2025.10.0

Disabling the Bitwarden extension: typing is fast as usual
Enabling the Bitwarden extension, Bitwarden is logged out: typing is slow in a text box
Enabling the Bitwarden extension, Bitwarden is logged in: typing is slow in a text box

NaLogo:
> Hi there,
> 
> I am unable to reproduce this issue, it has been escalated for further investigation. If you have more information that can help us, please add it below.
> 
> Thanks!

I didn't attach it in my original comment because I thought it might be easily reproducible, but here's a screen cap from when I ran the Firefox profiler and then typed in the PR description field.

<img width="648" height="442" alt="Image" src="https://github.com/user-attachments/assets/ca57cfb9-c2d4-4a46-bca3-f6da2d40d5ec" />

[...]

Issue #13558, opened 2025-02-25
Performance issues with Chrome plugin


mmeeks:
### Steps To Reproduce

We implement an on-line office suite with Collabora Online - that involves quite some complexity, DOM manipulation on key-press, and quite a complex set of JS. We spend a far amount of time optimizing things.

The more Chrome profiles I read - the more I see the Chrome Bitwarden plugin consuming CPU as an example - it takes more CPU time (and worse latency - cf. a 30ms block there) continually re-scanning what is after all an Office application looking for auto-fill opportunities - than it does rendering the application.

![Image](https://github.com/user-attachments/assets/7a7e0be7-7ac8-455d-8aec-abda3a1ae601)

Which is not great.

### Expected Result

Would love a way to turn off this continual DOM scanning. Ideally an autocomplete="off" somewhere that completely disables this.

  /**
   * Handles the processing of all mutations in the mutations queue. Will trigger
   * within an idle callback to help with performance and prevent excessive updates.
   */
  private processMutations = () => {
    const queueLength = this.mutationsQueue.length;

    if (!this.domQueryService.pageContainsShadowDomElements()) {
      this.checkPageContainsShadowDom();
    }

I guess this scan-whole-world from [clients](https://github.com/bitwarden/clients/tree/main)/[apps](https://github.com/bitwarden/clients/tree/main/apps)/[browser](https://github.com/bitwarden/clients/tree/main/apps/browser)/[src](https://github.com/bitwarden/clients/tree/main/apps/browser/src)/[autofill](https://github.com/bitwarden/clients/tree/main/apps/browser/src/autofill)/[services](https://github.com/bitwarden/clients/tree/main/apps/browser/src/autofill/services)
/collect-autofill-content.service.ts is the problem (?)

### Actual Result

Slowness affecting all apps that dynamically alter the DOM with bitwarden browser plugin enabled.

### Operating System

Linux

### Web Browser

Chrome

### Build Version

2025.1.2

### Issue Tracking Info

- [x] I understand that work is tracked outside of Github. A PR will be linked to this issue should one be opened to address it, but Bitwarden doesn't use fields like "assigned", "milestone", or "project" to track progress.

moritzpf:
I have the same issue. I have a page with about 20 lottie animations, which trigger a very high amount of constant dom changes. The plugin is eating more performance than the page itself. The page is literally unusable, please assign this ticket a higher priority.

cbw23:
Hi there,

[...]

PR #20286, opened 2026-04-20
[PM-35399] Issue-20107: Performance issue due to observer cascade and shadow DOMs in modern apps/websites.

golddragon007:
## 🎟️ Tracking

This MR should fix performance issues with browser freezes and input lag like:
https://github.com/bitwarden/clients/issues/20172
https://github.com/bitwarden/clients/issues/20107
https://github.com/bitwarden/clients/issues/18771 (potentially)
https://github.com/bitwarden/clients/issues/17128 (?)
https://github.com/bitwarden/clients/issues/13558
https://github.com/bitwarden/clients/issues/12031
https://github.com/bitwarden/clients/issues/10151

## 📔 Objective

Fixes freeze and input lag whenever the addon is enabled (no login required).

Tested on JIRA create ticket page, where in WYSIWYG I had like 3-4 sec lag between selecting text. After these changes this is gone, and functions as normal, tab CPU usage does not exceed 20%. (previously it was ~100%)

TBD:
- [ ] test to login
- [x] potentially fix tests
- [ ] fix broken stuff due to this PR (if there is any)
- [ ] code coverage(?)

# Shadow DOM Performance Optimizations
**1. Skip shadow root probing on pages without shadow DOM**
Where: buildTreeWalkerNodesQueryResults and checkForNewShadowRoots
A pageContainsShadowDom boolean is cached at page load. Any code path that would scan for shadow roots now checks this flag first and returns immediately if the page has none. This means zero shadow DOM overhead on the vast majority of ordinary pages.
 
**2. Only probe custom elements for shadow roots**
Where: buildTreeWalkerNodesQueryResults and queryShadowRoots
Per the HTML spec, only custom elements (those with a hyphen in their tag name, e.g. <my-widget>) can host a shadow root. Previously, getShadowRoot (which calls the synchronous Chrome extension API chrome.dom.openOrClosedShadowRoot) was called on every single DOM node — divs, spans, inputs, etc. Now there's a guard:
```
if (this.pageContainsShadowDom && element.tagName.includes("-")) {
    const shadow = this.getShadowRoot(element);
    ...
}
```
This eliminated thousands of unnecessary cross-boundary API calls per mutation cycle.
 
**3. Replaced :defined selector with * + tag name filter**
Where: queryShadowRoots
The old code used querySelectorAll(":defined") to find shadow root candidates. The :defined pseudo-class matches all registered built-in HTML elements — potentially thousands on a complex page like Jira. It was replaced with querySelectorAll("*") filtered by tagName.includes("-"), so getShadowRoot is only called on actual custom elements.
 
**4. Cache known shadow roots to avoid repeated full-page scans**
Where: deepQueryElements
A knownShadowRoots Set caches already-discovered shadow roots. On subsequent calls (e.g. every intersection event or page detail refresh), the code reuses this cache instead of doing a fresh querySelectorAll("*") + getShadowRoot scan of the entire DOM. The cache is only cleared when the mutation observer is rebuilt.

# Mutation observer changes
Before change:
```
mutation fires
  → pushed to mutationsQueue
  → debouncedProcessMutations fires (1 idle callback)
    → processMutations loops over N batches
      → schedules N idle callbacks (one per batch)
        → each batch: processMutationRecords(mutations)
          → schedules M idle callbacks (one per MutationRecord in the batch)
            → each record: processMutationRecord(mutation)
        → if last batch && domRecentlyMutated:
          → updateAutofillElementsAfterMutation()
```

After change:
```
mutation fires
  → pushed to mutationsQueue
  → debouncedProcessMutations fires (1 idle callback)
    → processMutations flattens all batches, clears queue immediately
      → schedules exactly 1 idle callback
        → processes all records synchronously
        → calls updateAutofillElementsAfterMutation once if needed
```

So for a page that delivers a single burst of 50 DOM mutations batched into 3 observer callbacks, the original would schedule 1 + 3 + 50 = 54 idle callbacks in a cascading chain. The new code schedules 2 (the debounce callback + the single processing callback).


audreyality:
👍🏻 It tickles me that I was selected to review this PR. I'm working on this problem too!

📝 I can't guarantee that we'll accept this PR, but I'll take it under advisement as I investigate this problem internally. If it's a good candidate for merge, then you can look forward to me cleaning it up for you! If not, then I'll add you as a co-author on the one I prepare. (I'd also traced some issues to shadow root detection.)

❌ In the meantime, there's quite a few broken unit tests. Also look for a [BIT report](https://github.com/bitwarden/browser-interactions-testing). These tests make sure that your performance enhancement hasn't come at the cost of breaking UX on other websites.

golddragon007:
I have to use the pipeline to run the tests as I have failed test on windows... and it takes a lot of time on this machine...
I changed shadow root detection a bit to see if this will fill the tests and fixed some other tests that had configuration issue.
Let's see now what it says.

golddragon007:
Hello @audreyality ,

It seems the tests have been fixed. I amended the last commit to put it in the ESLint suggestion. With all of this, the only remaining thing to do is to have code coverage and do manual testing if necessary.
I let both of those go to you as I have no clue how your policies want this to be covered, and likely I would just make tests that do nothing, just cover the code due to the type of lines that it complains about...

(I tried to build master, but the end result is non-functional, I can't pass the 2FA... which I don't think relates to my changes; however, just enabling the build addon does not cause issues anymore, unlike before the changes)

audreyality:
Thanks @golddragon007! 

There's a bunch of background work that I need to align, so this PR will likely lay dormant for a hot minute. I'm actively working on it, and will keep you informed as things move along.

golddragon007:
Uh, that's sad, as my browser causes issues because of this... well, I need to find a way to somehow build this to be functional, maybe from the last release.

golddragon007:
Hello @audreyality ,

I checked out the browser-v2026.3.0 tag (without my changes), and did an `npm install` then went into the app/browser and did an `npm run build:prod:chrome`, but the end result is non functional. I have problem with the login 2FA and then after the the vault locks I have problem with signing back... Can you help me how to build a local build that works? Does it require some special build settings to build the project to be functional? (it does not need to be packed for me)

audreyality:
You should be able to use the instructions in the [contributors guide](https://contributing.bitwarden.com/getting-started/clients/browser/#chrome-and-chromium-based-browsers) to run a build. There's nothing special about an official build (besides distribution). I use those same instructions to run my builds.

In general, I recommend running a development build of the Bitwarden server to test against when debugging. You can log in to your official Bitwarden vault, but a local build ensures that you won't run into things like API rate limits.

golddragon007:
> You should be able to use the instructions in the [contributors guide](https://contributing.bitwarden.com/getting-started/clients/browser/#chrome-and-chromium-based-browsers) to run a build. There's nothing special about an official build (besides distribution). I use those same instructions to run my builds.
> 
> In general, I recommend running a development build of the Bitwarden server to test against when debugging. You can log in to your official Bitwarden vault, but a local build ensures that you won't run into things like API rate limits.

Hmmm... Interesting because even if I build the released tag, it does not function for me (without my patch). I have issues with it around the authentication.

Well, even downloading the artifact from the pipeline does not work... I'm genuinely unable to sign in and pass the password, there's some package issue somewhere as I don't even see on the network tab the authentication request, nor the SignalR in the console for the tunel, nothing... weird.

audreyality:
If you're using the production environment (vault.bitwarden.com) to log in, you may be running into safeguards we apply on that environment. Have you tried setting up a local server? You don't need to build the whole thing. You could instead connect to [Bitwarden Lite](https://bitwarden.com/help/install-and-deploy-lite/).

You might also try asking for help setting up your environment in the [code](https://community.bitwarden.com/c/contribute-development/github-contributions/7/none) forum, so that this ticket stays focused on the PR.

golddragon007:
I using self hosted instance, I believe this one: https://github.com/dani-garcia/vaultwarden
Whatever comes with [Home Assistant](https://www.home-assistant.io/) by default.

Indeed I sent the right link, as here is clearly written that it used that one: https://github.com/hassio-addons/app-vaultwarden

apps/browser/src/autofill/services/collect-autofill-content.service.ts
```diff
@@ -1208,8 +1208,16 @@ export class CollectAutofillContentService implements CollectAutofillContentServ
...
+      /**
+       * Mutations to node attributes NOT on this list will not be observed!
+       * `class` is intentionally excluded: it changes extremely frequently on
+       * modern pages (animations, hover states, framework re-renders) and fires
+       * the handler for every element in the entire subtree. The cached
+       * `htmlClass` value is refreshed on the next full page-detail collection.
+       */
+      attributeFilter: Object.values(AUTOFILL_ATTRIBUTES).filter(
+        (attr) => attr !== AUTOFILL_ATTRIBUTES.CLASS,
+      ),
```
jprusik, line 1220: Please drop `CLASS` from `AUTOFILL_ATTRIBUTES` instead of filtering (we discussed internally and agree the perf gain is worth any potential class-only mutation miss)

audreyality:
@golddragon007 - we're going to take your contribution and iterate on it within our repository. 

Thank you for your contribution! I'll be sure to give you credit when the squash commit hits `main`!

audreyality:
@golddragon007 - Merged! https://github.com/bitwarden/clients/commit/6f3d3239c29966574b86003cb01817051103139a

PR #20289, opened 2026-04-20
[PM-31942] Add FileReportPersistenceService implementation to store Access Intelligence reports as files (pt. 2)

lastbestdev:
## 🎟️ Tracking
Part 2 of https://bitwarden.atlassian.net/browse/PM-31942 

## 📔 Objective
Includes the implementation of the ReportPersistenceService abstraction for handling Access Intelligence report file storage. Supports both cloud file uploads (Azure) and Bitwarden self-hosted file uploads (direct to API server).

Does not include any changes to hook into the V2 implementation of Access Intelligence logic. Those changes will come last in part 3, after #20081 is merged to `main`.

## 📸 Screenshots
N/A


bitwarden_license/bit-common/src/dirt/access-intelligence/services/implementations/api/default-access-intelligence-api.service.ts
```diff
@@ -197,4 +197,38 @@ export class DefaultAccessIntelligenceApiService extends AccessIntelligenceApiSe
 
     return from(response).pipe(map((response) => new AccessReportApi(response)));
   }
+
+  downloadReportFileAzure$(url: string): Observable<{ blob: Blob; fileName: string }> {
```
Banrion, line 198: ⚠️ New `downloadReportFileAzure$` implementation has no unit tests. The fetch logic itself is untested in the api service test file

bitwarden_license/bit-common/src/dirt/access-intelligence/services/implementations/api/default-access-intelligence-api.service.ts
```diff
@@ -197,4 +197,38 @@ export class DefaultAccessIntelligenceApiService extends AccessIntelligenceApiSe
...
+            const { done, value } = await reader.read();
+            if (done) {
+              break;
+            }
+            if (value) {
+              chunks.push(value);
+            }
+          }
+
+          const blob = new Blob(chunks);
```
Banrion: ⚠️ Use `response.blob()`

bitwarden_license/bit-common/src/dirt/access-intelligence/services/implementations/api/default-access-intelligence-api.service.ts
```diff
@@ -197,4 +197,38 @@ export class DefaultAccessIntelligenceApiService extends AccessIntelligenceApiSe
...
+          const contentDisposition = response.headers.get("Content-Disposition");
+          const fileName =
+            contentDisposition?.match(/filename="?([^";\n]+)"?/i)?.[1]?.trim() ??
+            url.split("/").pop() ??
+            "report";
+
```
Banrion: ⚠️ Instead of the full url, use the pathname

```suggestion
            const urlPath = new URL(url).pathname;
            const fallback = urlPath.split("/").pop() || "report";
```

bitwarden_license/bit-common/src/dirt/access-intelligence/services/implementations/persistence/file-report-persistence.service.ts
```diff
@@ -0,0 +1,262 @@
...
+
+            const upload$ = from(reportFile.arrayBuffer()).pipe(
+              switchMap((buffer) =>
+                from(
+                  this.fileUploadService.upload(
+                    { url: result.reportFileUploadUrl, fileUploadType: result.fileUploadType },
+                    new EncString(""),
+                    { buffer: new Uint8Array(buffer) } as unknown as EncArrayBuffer,
+                    {
+                      postDirect: () =>
```
Banrion: ❌ `postDirect` callback discard the `FormData` produced by `BitwardenFileUploadService`

Use the pattern of generating the methods like in [send-api.service.ts](https://github.com/bitwarden/clients/blob/fe4ebd458e985d872dc8b3bfa8f959ce0843b9dd/libs/common/src/tools/send/services/send-api.service.ts#L223) and [cipher-file-upload.service.ts](https://github.com/bitwarden/clients/blob/fe4ebd458e985d872dc8b3bfa8f959ce0843b9dd/libs/common/src/vault/services/file-upload/cipher-file-upload.service.ts#L71) for the `postDirect`, `renewFileUploadUrl` and `rollback` callbacks.

For observablility patterns, you can use `firstValueFrom` or `from`
```
postDirect: (fd: FormData) =>
  firstValueFrom(
    this.accessIntelligenceApiService.uploadReportFileFormData$(
      organizationId,
      reportId,
      fd,
      reportFileId,
    ),
  ),
```

Update the `uploadReportFile$` to accept `FormData` instead of `File` to support the `postDirect` callback.

```
uploadReportFile$(
  orgId: OrganizationId,
  reportId: OrganizationReportId,
  formData: FormData,
  reportFileId: string,
): Observable<void> {
  return from(
    this.apiService.send(
      "POST",
      `/reports/organizations/${orgId}/${reportId}/file?reportFileId=${reportFileId}`,
      formData,
      true,
      false,
    ),
  );
}

File: bitwarden_license/bit-common/src/dirt/access-intelligence/services/implementations/persistence/file-report-persistence.service.ts [hunk 5]
```diff
@@ -0,0 +1,262 @@
...
+
+            return upload$.pipe(map(() => ({ id: reportId, contentEncryptionKey })));
+          }),
+        );
+      }),
+    );
+  }
+
+  saveApplicationMetadata$(view: AccessReportView): Observable<void> {
+    this.logService.debug("[DefaultReportPersistenceService] Saving application metadata", {
```
Banrion: ⚠️ Using `DefaultReportPersistenceService` instead of `FileReportPersistenceService` for logs

bitwarden_license/bit-common/src/dirt/access-intelligence/services/implementations/persistence/file-report-persistence.service.ts
```diff
@@ -0,0 +1,262 @@
...
+          }),
+        );
+      }),
+    );
+  }
+
+  loadReport$(
+    organizationId: OrganizationId,
+  ): Observable<{ report: AccessReportView; hadLegacyBlobs: boolean } | null> {
+    this.logService.debug("[DefaultReportPersistenceService] Loading report", { organizationId });
```
Banrion: ⚠️ Using `DefaultReportPersistenceService` instead of `FileReportPersistenceService` for logs

bitwarden_license/bit-common/src/dirt/access-intelligence/services/implementations/persistence/file-report-persistence.service.ts
```diff
@@ -0,0 +1,262 @@
...
+            }
+
+            // V2: reportData lives in a file. Determine download strategy from the URL:
+            //   - Azure blob URL → unauthenticated GET (SAS token in URL handles auth)
+            //   - Server URL → authenticated API call
+            // V1 fallback: reportData is inline in the response.
+            let reportData$: Observable<string>;
+            if (apiResponse.reportFileDownloadUrl) {
+              const isAzure = new URL(apiResponse.reportFileDownloadUrl).hostname.includes(
+                "blob.core.windows.net",
```
Banrion: ❌ Rather than checking the `reportFileDownloadUrl`, we should be using a `fileDownloadType` from the response. 

Update response type and change to `const isAzure =  apiResponse.fileDownloadType === FileUploadType.Azure`. Server changes may be required.

bitwarden_license/bit-common/src/dirt/access-intelligence/services/implementations/persistence/file-report-persistence.service.ts
```diff
@@ -0,0 +1,262 @@
...
+              switchMap((reportData) => {
+                // Convert API → Data → Domain → View (following 4-layer architecture)
+                const data = new AccessReportData();
+                data.id = apiResponse.id;
+                data.organizationId = apiResponse.organizationId;
+                data.reports = reportData;
+                data.summary = apiResponse.summary;
+                data.applications = apiResponse.applications;
+                data.creationDate = apiResponse.creationDate;
+                data.contentEncryptionKey = apiResponse.contentEncryptionKey;
```
Banrion: ⚠️ Use `const data = new AccessReportData(apiResponse)` instead of manually setting each field.

bitwarden_license/bit-common/src/dirt/access-intelligence/services/implementations/persistence/file-report-persistence.service.ts
```diff
@@ -0,0 +1,262 @@
...
+      organizationId: view.organizationId,
+      applicationCount: view.applications.length,
+    });
+
+    return from(firstValueFrom(getUserId(this.accountService.activeAccount$))).pipe(
+      switchMap((userId) => {
+        if (!userId) {
+          throw new Error("User ID not found");
+        }
+
```
Banrion: ⚠️ Can be removed since `getUserId` guards the `userId`

bitwarden_license/bit-common/src/dirt/access-intelligence/services/implementations/persistence/file-report-persistence.service.ts
```diff
@@ -0,0 +1,262 @@
...
+    this.logService.debug("[FileReportPersistenceService] Saving report", {
+      organizationId,
+    });
+
+    return from(firstValueFrom(getUserId(this.accountService.activeAccount$))).pipe(
+      switchMap((userId) => {
+        if (!userId) {
+          throw new Error("User ID not found");
+        }
+
```
Banrion: ⚠️ Can be removed since `getUserId` guards the `userId`

bitwarden_license/bit-common/src/dirt/access-intelligence/services/implementations/persistence/file-report-persistence.service.ts
```diff
@@ -0,0 +1,262 @@
...
+    organizationId: OrganizationId,
+  ): Observable<{ report: AccessReportView; hadLegacyBlobs: boolean } | null> {
+    this.logService.debug("[DefaultReportPersistenceService] Loading report", { organizationId });
+
+    return from(firstValueFrom(getUserId(this.accountService.activeAccount$))).pipe(
+      switchMap((userId) => {
+        if (!userId) {
+          throw new Error("User ID not found");
+        }
+
```
Banrion: ⚠️ Can be removed since `getUserId` guards the `userId`

bitwarden_license/bit-common/src/dirt/access-intelligence/services/implementations/persistence/file-report-persistence.service.ts
```diff
@@ -0,0 +1,262 @@
...
+              map((result) => ({
+                result,
+                reportFile,
+                contentEncryptionKey: domain.contentEncryptionKey!,
+              })),
+            );
+          }),
+          switchMap(({ result, reportFile, contentEncryptionKey }) => {
+            const reportId = result.reportResponse.id as OrganizationReportId;
+            const reportFileId = result.reportResponse.reportFile?.id ?? "";
```
Banrion: ⚠️ `reportFileId` should never be null when using file storage. If it is, throw an error instead of defaulting to an empty string.

This led me to looking at the `ReportFileApi` model. The Api layer's job is to de-serialize the response format and doesn't necessarily enforce the business contract. This means that we should initialize `id` in `ReportFileApi` and throw an error if it isn't valid when using it in business logic.

```
const reportFileId = result.reportResponse.reportFile?.id;
if (!reportFileId) {
  throw new Error("Report file ID missing from server response");
}
```

Banrion (changes requested):

bitwarden_license/bit-common/src/dirt/access-intelligence/services/implementations/api/default-access-intelligence-api.service.ts
lastbestdev: Great catch & simplification 👍🏻

bitwarden_license/bit-common/src/dirt/access-intelligence/services/implementations/persistence/file-report-persistence.service.ts
lastbestdev: You are correct in that server changes would be required for this logic. A file download type is not currently included in the Get latest report endpoint response. 

CC @prograhamming, are we able to add the file download type to the response for retrieving the latest report?

bitwarden_license/bit-common/src/dirt/access-intelligence/services/implementations/persistence/file-report-persistence.service.ts
lastbestdev: Great point, since file uploads will always fail should it be missing. Updated

bitwarden_license/bit-common/src/dirt/access-intelligence/services/implementations/persistence/file-report-persistence.service.ts
lastbestdev: Updated in latest

Banrion:
@claude Review this pr please

bitwarden_license/bit-common/src/dirt/access-intelligence/services/implementations/persistence/file-report-persistence.service.ts
lastbestdev: This has been updated in the latest commits. Thank you @prograhamming for adding the field!

bitwarden_license/bit-common/src/dirt/access-intelligence/services/implementations/persistence/file-report-persistence.service.ts
```diff
@@ -0,0 +1,265 @@
...
+
+            const updateSummaryCall = this.accessIntelligenceApiService.updateSummaryData$(
+              view.organizationId,
+              view.id,
+              data.summary,
+              metrics as AccessReportMetricsApi,
+            );
+
+            return forkJoin([updateApplicationsCall, updateSummaryCall]).pipe(
+              map(() => undefined as void),
```
Banrion: ⚠️ The api service contains an endpoint for updating the applications and summary separately for backwards compatibility with v1 when the endpoints were separate. In this persistence service, prefer using the endpoint that updates both together to make one call instead of two. Use `updateReportSettings$` in the api service.

bitwarden_license/bit-common/src/dirt/access-intelligence/services/implementations/persistence/file-report-persistence.service.ts
```diff
@@ -0,0 +1,265 @@
...
+            const upload$ = from(reportFile.bytes()).pipe(
+              switchMap((buffer) =>
+                from(
+                  this.fileUploadService.upload(
+                    {
+                      url: createReportResponse.reportFileUploadUrl,
+                      fileUploadType: createReportResponse.fileUploadType,
+                    },
+                    reportFile.name as unknown as EncString,
+                    { buffer } as unknown as EncArrayBuffer,
```
Banrion: ⚠️ Agreeing with claude that we should avoid typecasting as unknown.

Banrion (changes requested):
Two comments added. Otherwise, this pull request is looking good!

bitwarden_license/bit-common/src/dirt/access-intelligence/services/implementations/persistence/file-report-persistence.service.ts
lastbestdev: Great callout, updated in latest

bitwarden_license/bit-common/src/dirt/access-intelligence/services/implementations/persistence/file-report-persistence.service.ts
lastbestdev: Fixed, but required more changes to existing file upload service methods to support using types other than EncString and EncArrayBuffer, which we are not using either of for this implementation.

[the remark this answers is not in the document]
lastbestdev: addressed in latest

Banrion: If the fix for typecasting was to update the existing services to support a new type, we are probably missing a step between having the encrypted file and uploading it. We should be able to use `EncArrayBuffer`, which was not originally used for encrypting reports due to them not being treated as files before.

Rather than updating the file services to include a non-encrypted type (the report file should be encrypted so this can be confusing from the outside perspective), we should add support for the correct type as output for an encrypted report.

lastbestdev:
@bitwarden/team-platform-dev @dereknance please disregard the review request. files changed have been updated to only include DIRT owned files 👍🏻

bitwarden_license/bit-common/src/dirt/access-intelligence/services/implementations/persistence/file-report-persistence.service.ts
```diff
@@ -0,0 +1,273 @@
...
+                    throw new Error(
+                      "Report file ID was not found in create report response. Unable to upload report as file",
+                    );
+                  }
+                }),
+                map((createReportResponse) => ({ createReportResponse, encryptedData })),
+              );
+            }),
+            switchMap(({ createReportResponse, encryptedData }) => {
+              const reportId = createReportResponse.reportResponse.id as OrganizationReportId;
```
Banrion, line 94: 📝  After locally reviewing with Claude to verify test cases, one missing case was suggested. Also not blocking for this pull request.

🧪 No spec coverage for the explicit throw when `reportFile.id` is missing from the create response.
`saveReport$` performs an explicit guard inside tap before the upload step:

```
tap((createReportResponse) => {
  const reportFileId = createReportResponse.reportResponse.reportFile?.id;
  if (!reportFileId) {
    throw new Error(
      "Report file ID was not found in create report response. Unable to upload report as file",
    );
  }
}),
```

The `reportFileId!` non-null assertion in `generateFileUploadCallbacks` depends on this throw firing. The accompanying spec covers the happy path, encryption failure, create failure, and upload failure — but no case asserts that a create response with `reportFile = undefined` (or `reportFile.id = undefined`) surfaces the documented error message.

Suggested addition:
```
it("should throw when create response is missing reportFile.id", async () => {
  const view = createRiskInsights({ organizationId });
  const responseWithoutFileId = new AccessReportFileApi({
    ReportFileUploadUrl: "https://storage.example.com/upload",
    FileUploadType: FileUploadType.Azure,
    ReportResponse: { Id: reportId, OrganizationId: organizationId },
    // ReportFile intentionally omitted
  });
  mockEncryptionService.encryptReportFile$.mockReturnValue(of(makeFileEncryptedData()));
  mockApiService.createReport$.mockReturnValue(of(responseWithoutFileId));

  await expect(firstValueFrom(service.saveReport$(view, organizationId))).rejects.toThrow(
    /Report file ID was not found/,
  );
});
```

This rounds out the error-path coverage and pins the contract for future maintainers (the ! assertion stays safe only as long as this guard runs first).

bitwarden_license/bit-common/src/dirt/access-intelligence/services/implementations/persistence/default-access-report-encryption.service.ts
```diff
@@ -226,6 +228,181 @@ export class DefaultAccessReportEncryptionService extends AccessReportEncryption
...
+            : from(this.keyGeneratorService.createKey(512))
+        ).pipe(
+          catchError((error: unknown) => {
+            this.logService.error(
+              "[DefaultAccessReportEncryptionService] Failed to get encryption key",
+              error,
+            );
+            throw new Error("Failed to get encryption key");
+          }),
+        );
```
Banrion, line 262: 🎨 Duplicate boilerplate for key resolution. This could be extracted into a helper function. Not blocking for the pull request.

Banrion (approved):
✅ LGTM!

bitwarden_license/bit-common/src/dirt/access-intelligence/services/implementations/persistence/default-access-report-encryption.service.ts
lastbestdev: will address in #20441

bitwarden_license/bit-common/src/dirt/access-intelligence/services/implementations/persistence/file-report-persistence.service.ts
lastbestdev: will address in #20441

PR #20298, opened 2026-04-21
Add warnings around org key subscriptions

quexten:
## 📔 Objective

Some teams invoke getting the orgkeys in a loop. Specifically, when importing to an organization, both the cipher encryption, and collection encryption each trigger a subscription for each encrypted item. Each subscription is not cached and will trigger a decryption of all organization and provider keys. This may mean tens or hundreds-of-thousands of RSA operations when importing a large vault.

This operation then takes multiple minutes instead of <1 second. This PR adds logs and deprecation notes that make this bad access pattern visible. Users should either make a bulk decryption operation, getting the keys once, or just move to the SDK which holds the keys in the key store.

### DIFF
--- a/libs/key-management/src/key.service.ts
+++ b/libs/key-management/src/key.service.ts
@@ -891,8 +891,10 @@ export class DefaultKeyService implements KeyServiceAbstraction {
                   continue;
                 }
                 decrypted = await encrypted.decrypt(this.encryptService, providerKeys!);
+                this.logService.info("WARNING: Decrypting provider key. This is slow and should only happen once on unlock! If this happens frequently in a loop, please investigate.");
               } else {
                 decrypted = await encrypted.decrypt(this.encryptService, userPrivateKey);
+                this.logService.info("WARNING: Decrypting organization key. This is slow and should only happen once on unlock! If this happens frequently in a loop, please investigate.");
               }
 
               result[orgId] = decrypted;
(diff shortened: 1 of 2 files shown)


libs/key-management/src/key.service.ts
```diff
@@ -891,8 +891,14 @@ export class DefaultKeyService implements KeyServiceAbstraction {
...
                 }
                 decrypted = await encrypted.decrypt(this.encryptService, providerKeys!);
+                this.logService.info(
+                  "WARNING: Decrypting provider key. This is slow and should only happen once on unlock! If this happens frequently in a loop, please investigate.",
+                );
               } else {
                 decrypted = await encrypted.decrypt(this.encryptService, userPrivateKey);
+                this.logService.info(
+                  "WARNING: Decrypting organization key. This is slow and should only happen once on unlock! If this happens frequently in a loop, please investigate.",
+                );
```
Thomas-Avery, line 901: I'm good with the deprecation notices but disagree with spamming the console log with these statements. There is overhead to writing to the console in the first place and  53 occurrences of calling `orgKeys$(` in the app. Most of these are valid but the user will see N number of WARNINGs for every N organization membership they have for one valid operation.

For the loop case it will be N times loop iterations occurrences.

Thomas-Avery (changes requested):

PR #20299, opened 2026-04-21
[PM-35835] Use SDK cipher decryption for admin-console

quexten:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-35835

## 📔 Objective

Cuts decryption time by 90% for admin console access for large vaults. (Measured ~38-40 seconds to 5 seconds).

### DIFF
--- a/libs/common/src/vault/services/cipher.service.ts
+++ b/libs/common/src/vault/services/cipher.service.ts
@@ -825,16 +825,11 @@ export class CipherService implements CipherServiceAbstraction {
       return [];
     }
 
+    const userId = await firstValueFrom(this.accountService.activeAccount$.pipe(map((a) => a?.id)));
     const ciphers = response.data.map((cr) => new Cipher(new CipherData(cr)));
-    const key = await this.keyService.getOrgKey(organizationId);
-    const decCiphers: CipherView[] = await Promise.all(
-      ciphers.map(async (cipher) => {
-        return await cipher.decrypt(key);
-      }),
-    );
-
-    decCiphers.sort(this.getLocaleSortingFunction());
-    return decCiphers;
+    let [decrypted, _] = await this.cipherEncryptionService.decryptManyLegacy(ciphers, userId);
+    decrypted.sort(this.getLocaleSortingFunction());
+    return decrypted;
   }
 
   async getLastUsedForUrl(


libs/common/src/vault/services/cipher.service.ts
```diff
@@ -829,16 +825,11 @@ export class CipherService implements CipherServiceAbstraction {
...
-    const key = await this.keyService.getOrgKey(organizationId);
-    const decCiphers: CipherView[] = await Promise.all(
-      ciphers.map(async (cipher) => {
-        return await cipher.decrypt(key);
-      }),
-    );
-
-    decCiphers.sort(this.getLocaleSortingFunction());
-    return decCiphers;
+    let [decrypted, _] = await this.cipherEncryptionService.decryptManyLegacy(ciphers, userId);
```
quexten, line 830: @eliykat AC question: How are decryption failures handled / do the need to be handled, or is it fine to just drop them?

eliykat: Not sure - I am used to the old error handling where a failed EncString decryption would return `[Error: Failed to decrypt]`. Do we no longer follow that pattern? How do we handle decryption failures generally in the vault?

eliykat, line 830: I notice we're using the deprecated legacy interface - is that correct?

quexten: I don't think vault offers another function here. There list view decryption returns list views which just have less information / are not applicable to AC. The individual decrypt is significantly slower.

I think ideally vault (or AC) would offer a functionality where the API request along with the decryption is moved to the SDK, and all you would have to do is call one SDK function.

shane-melton: Correct, we don't have an alternative here yet that isn't the more limited list view implementation.

shane-melton: The Individual Vault shows a [DecryptionFailureDialog](https://github.com/bitwarden/clients/blob/a07c9ebf6bf3e87cc1d8710f119cb82434d5e1b3/libs/vault/src/components/decryption-failure-dialog/decryption-failure-dialog.component.ts#L38) when any of failed/corrupted ciphers are detected. It informs the user that the provided cipherIds could not be decrypted and that they should reach out to CS. I don't see any reason you cannot re-use the same flow/dialog here in the AC, but that depends on how you all would prefer to surface those issues.

shane-melton, line 830: 🎨 I'd suggest putting this behind a feature flag so it can be more easily toggled in case organizations have some unexpected data format that breaks the SDK implementation. We've addressed all known issues found for individual vaults, but there is still the small chance there are unique unassigned items that could cause unexpected behaviors that is unique to organization vaults.

shane-melton (commented):
Approach looks good, though I'd recommend a feature flag for an added layer of safety.

PR #20300, opened 2026-04-21
PM-35270 moved eventcollection to phishing detection service

voommen-livefront:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-35270

## 📔 Objective

Event collection for the Phishing blocker was in the component. 
Moved this logic to the phishing-detection.service.
The service is long-lived whereas the compoent is destroyed as soon as the user navigates to or away from the phishing blocker site.

## 📸 Screenshots

Screen shots attached to the story to aid QA


apps/browser/src/dirt/phishing-detection/services/phishing-detection.service.ts
```diff
@@ -49,11 +67,34 @@ export class PhishingDetectionService {
...
+    ): Promise<void> => {
+      try {
+        const orgs = await getOrgsToNotify();
+        // intentionally keeping this sequential
+        // using Promise.all creates a race condition in eventCollectionService
+        for (const org of orgs) {
+          await eventCollectionService.collect(eventType, undefined, uploadImmediately, org.id);
+        }
+      } catch {
+        // Event collection failure should not block the user action
```
Banrion: 🎨 A warning log here would not block the user but give notice in the console on error

Banrion (dismissed):

Banrion (approved):

PR #17731, opened 2025-11-29
[PM-27283] [BEEEP] Reactive `availableVaultTimeoutActions$` in vault timeout settings

mzieniukbw:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-27283

## 📔 Objective

Make `availableVaultTimeoutActions$` reactive, connected to underlying state.
Required changes to biometric state and pin state services, to expose two functions to be reactive too.
For backwards compatibility, the `UserId` arg is not required and works as before.
To minimize the impact and QA testing, upstream changes (callers of changes functions) have been kept to minimum - this is separate work.

## 📸 Screenshots

https://github.com/user-attachments/assets/ece4f117-a86d-4461-8c37-d4c71b827fe3

https://github.com/user-attachments/assets/101e1c40-f99a-4718-9771-8653793ecfb0

https://github.com/user-attachments/assets/4dac0d47-4d4a-4240-ac01-56069af4747b

## ⏰ Reminders before review

- Contributor guidelines followed
- All formatters and local linters executed and passed
- Written new unit and / or integration tests where applicable
- Protected functional changes with optionality (feature flags)
- Used internationalization (i18n) for all UI strings
- CI builds passed
- Communicated to DevOps any deployment requirements
- Updated any necessary documentation (Confluence, contributing docs) or informed the documentation team

## 🦮 Reviewer guidelines

- 👍 (`:+1:`) or similar for great changes
- 📝 (`:memo:`) or ℹ️ (`:information_source:`) for notes or general info
- ❓ (`:question:`) for questions
- 🤔 (`:thinking:`) or 💭 (`:thought_balloon:`) for more open inquiry that's not quite a confirmed issue and could potentially benefit from discussion
- 🎨 (`:art:`) for suggestions / improvements
- ❌ (`:x:`) or ⚠️ (`:warning:`) for more significant problems or concerns needing attention
- 🌱 (`:seedling:`) or ♻️ (`:recycle:`) for future improvements or indications of technical debt
- ⛏ (`:pick:`) for minor or nitpick changes

PR #20309, opened 2026-04-21
Auth/PM-35336 - TokenService - prevent stale access token retrieval to fix logout on org user confirm v1

JaredSnider-Bitwarden:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-35336

## 📔 Objective

To solve the below bug: 

## Bug Description

When a user is JIT-provisioned via SSO into an **MP encryption org** (non-TDE), they are unexpectedly
logged out immediately after being confirmed to the organization. This happens in both auto-confirm
and manual confirm flows.

<details>
<summary>Root Cause</summary>

All four token types share a dual-storage pattern: a `_MEMORY` key and a `_DISK` key. Every getter
reads Memory first and falls back to Disk. But no setter cleared the other location on write —
meaning Memory and Disk could silently diverge. In the `TokenService`, we erroneously counted on external clears instead of internally enforcing the single token storage location invariant.

The divergence was triggered by the following sequence:

1. **Token written to Memory on login**: New users JIT-provisioned via SSO use the default vault
   timeout settings (`LogOut + 15min`), causing the access token to be written to
   `ACCESS_TOKEN_MEMORY` only. Note: the fact that we have [dev overrides](https://github.com/bitwarden/clients/blob/main/apps/web/src/app/core/core.module.ts#L193-L198) for the default vault timeout made this originally unable to be reproduced locally. 

2. **JIT provisioning triggers a background token migration**: Once JIT-provisioned and after the user set's an initial MP, the user's password becomes available, making `Lock` a valid vault timeout action. Since
   https://github.com/bitwarden/clients/pull/17731 made `availableVaultTimeoutActions$`
   fully reactive and https://github.com/bitwarden/clients/pull/19594 introduced token migration as a side effect, this availability change is observed mid-session — firing a `null → Lock`
   transition that calls `migrateTokenStorage` and copies the token to `ACCESS_TOKEN_DISK`.
   Memory is **not** cleared.

3. **`refreshIdentityToken` writes the new token to Disk only**: The org confirmation triggers a
   `SyncOrgKeys` push notification → `fullSync` → `refreshIdentityToken()`, which writes the new
   JWT (containing the new org claim) to `ACCESS_TOKEN_DISK`. Memory is **not** updated.

4. **Every `getAccessToken` call returns the stale Memory value**: The getter's Memory-first read
   always returns the old token — the one without the org claim. This causes a 403 on subsequent
   API calls, which triggers logout.

</details>

## 📸 Screenshots

https://github.com/user-attachments/assets/c4689c16-07be-417b-916f-07f1a2ac9cdf


enmande (dismissed):
Thank you for the discussion today -- LGTM.

quexten:
@eligrubb Swapped reviewers here since @JaredSnider-Bitwarden requested a review since this is a review blocker.

libs/common/src/auth/services/token.service.ts
```diff
@@ -359,6 +376,12 @@ export class TokenService implements TokenServiceAbstraction {
             userId,
           );
 
+          const newEncryptedAccessTokenInState = this.waitForStateValue(
```
quexten: Thought (non-blocking): Instead of putting the wait + write logic separately in your implementation, you could make a private method:
```
  private writeAndWait<T>(
    userId: UserId,
    keyDefinition: UserKeyDefinition<T>,
    value: T,
  ): Promise<T> {
    ...
  }
```
that internally does all of the:
- Create promise
- Start write
- Await promise

This would make the actual logic that wants to do the write more readable and easier to reason about.

This pattern seems to be used in at least two places and it feels worth it to extract.

libs/common/src/auth/services/token.state.ts
```diff
@@ -13,11 +13,16 @@ import {
 export const ACCESS_TOKEN_DISK = new UserKeyDefinition<string>(TOKEN_DISK, "accessToken", {
   deserializer: (accessToken) => accessToken,
   clearOn: [], // Manually handled
+  // Disable the ReplaySubject cache so every subscriber always reads the access token
+  // directly from storage. A cached value could be stale if a write has not yet
+  // propagated, causing downstream callers to send an outdated JWT on authenticated requests.
+  cleanupDelayMs: 0,
```
quexten: Comment (non-blocking): Disk storage can take 100+msec to read. This means that access token access may take 100+ msec to access each time, which means a stark performance regression. This is not proven, and non-blocking from my side, but something co consider.

In general, I want to propose a simplified architecture for the access token storage. This is similar to how biometrics now works.

1. There is a memory storage location (possibly secure memory). This storage location is *always* used for reading.
2. There is a disk storage location, it is used for filling the memory storage location on application startup, but otherwise never read

Operations:
4. Write:
 - If allow persist to disk, write to both memory, disk
 - If not allow persist to disk, write to memory, wipe disk
5. Read:
 - Read from memory. If empty, read from disk and fill memory
6. Switch persistence mode (due to vault timeout)
 - Delete disk / write to disk

This would alleviate most of these bugs at their root, because memory is always a reliable location to read from. You are no longer mutating active state / swapping locations which is a "dangerous" migration, you are merely changing the way it gets restort on app restart.

This de-risks the token service, while at the same time ensuring quick reads since we don't have to round-trip to disk.

libs/common/src/auth/services/token.service.ts
```diff
@@ -453,11 +514,19 @@ export class TokenService implements TokenServiceAbstraction {
       await this.clearAccessTokenKey(userId);
     }
 
-    // Clear tokens from disk storage (all platforms)
+    // Clear from disk and memory.
+    await this.clearAccessTokenDiskLocation(userId);
```
quexten, line 507: Should this also block until the state is propagated? (I'm not sure)

libs/common/src/auth/services/token.service.ts
```diff
@@ -601,13 +670,8 @@ export class TokenService implements TokenServiceAbstraction {
...
           decryptedRefreshToken = refreshToken;
 
-          // TODO: PM-6408
-          // 2024-02-20: Remove refresh token from memory and disk so that we migrate to secure storage over time.
-          // Remove these 2 calls to remove the refresh token from memory and disk after 3 months.
-          await this.singleUserStateProvider.get(userId, REFRESH_TOKEN_DISK).update((_) => null, {
-            shouldUpdate: (previousValue) => previousValue !== null,
-          });
-          await this.singleUserStateProvider.get(userId, REFRESH_TOKEN_MEMORY).update((_) => null);
+          await this.clearRefreshTokenDiskLocation(userId);
```
quexten, line 662: Should this also block?

libs/common/src/auth/services/token.service.ts
```diff
@@ -622,21 +686,28 @@ export class TokenService implements TokenServiceAbstraction {
...
       }
-      case TokenStorageLocation.Disk:
-        return await this.singleUserStateProvider
+      case TokenStorageLocation.Disk: {
+        const newRefreshToken = await this.singleUserStateProvider
           .get(userId, REFRESH_TOKEN_DISK)
           .update((_) => refreshToken, {
             shouldUpdate: (previousValue) => previousValue !== refreshToken,
           });
+        await this.clearRefreshTokenMemoryLocation(userId);
```
quexten, line 689: Seeing how many times we repeat this pattern, I do strongly recommend extracting the atomic write to a function, then using it in all places. This function should also use it.

libs/common/src/auth/services/token.service.ts
```diff
@@ -743,13 +822,17 @@ export class TokenService implements TokenServiceAbstraction {
     );
 
     if (storageLocation === TokenStorageLocation.Disk) {
-      return await this.singleUserStateProvider
+      const newClientId = await this.singleUserStateProvider
```
quexten, line 814: Should this wait for the propagation?

libs/common/src/auth/services/token.service.ts
```diff
@@ -817,13 +908,17 @@ export class TokenService implements TokenServiceAbstraction {
     );
 
     if (storageLocation === TokenStorageLocation.Disk) {
-      return await this.singleUserStateProvider
+      const newClientSecret = await this.singleUserStateProvider
```
quexten, line 900: Should this wait for the propagation?

quexten (dismissed):
I'll not block over the reviews, since this is a release blocker, but I do think we can:
- Clean up the current implementation by extracting the atomic writing into a separate function and using it everywhere

I also think the architecture of the token-service is risky as-is with swapping storage locations. I have outlined a better approach that reduces risk / makes it much easier to reason about.

libs/common/src/auth/services/token.state.ts
JaredSnider-Bitwarden: Hmm... technically, that `cleanupDelayMs` change isn't needed to fix this bug. It might be worth reverting to avoid performance issues on desktop. Thank you for the detailed suggestion on how to re-architect this. I will note this in a plan for a future overhaul.

libs/common/src/auth/services/token.service.ts
JaredSnider-Bitwarden: Great point. Done with 3a24a922f8f3d8bb184155620e8701cda1eb37fb

libs/common/src/auth/services/token.state.ts
JaredSnider-Bitwarden: Reverted with 7ae7819a13f25937ceddb67b69bbec161cc732ad

libs/common/src/auth/services/token.service.ts
JaredSnider-Bitwarden: Claude Answer: 

> No, blocking isn't needed here. The race condition `writeAndWait` guards against is: write a valid value to Location A, then clear Location B before A's state$ has propagated — leaving a window where both locations appear null. That window is dangerous because the token should be valid.
> 
> In `clearAccessToken` there's no new write — both locations are being set to null. Between the two clears a concurrent reader checks memory first; if memory hasn't been cleared yet it returns the still-valid token. Once both are null the reader gets null as intended. No dangerous null window exists.
> 

I think I agree. I'm not seeing the same danger on clear vs on set.

JaredSnider-Bitwarden:
@quexten I'm closing this in favor of https://github.com/bitwarden/clients/pull/20334 because we ascertained that none of the atomicity changes were required to fix the bug, but I want to keep the history here in case we do run into needing them in the future.

PR #20310, opened 2026-04-21
[BEEEP] Refactor dialog to use beforeClose$ rather than imperative predicates

BTreston:

### DIFF
--- a/libs/components/src/dialog/dialog.service.ts
+++ b/libs/components/src/dialog/dialog.service.ts
@@ -24,7 +24,6 @@ import {
 
 import { AuthService } from "@bitwarden/common/auth/abstractions/auth.service";
 import { AuthenticationStatus } from "@bitwarden/common/auth/enums/authentication-status";
-import { LogService } from "@bitwarden/logging";
 
 import { isAtOrLargerThanBreakpoint } from "../utils/responsive-utils";
 
@@ -53,14 +52,21 @@ class CustomBlockScrollStrategy implements ScrollStrategy {
   detach() {}
 }
 
+export type BeforeCloseEvent<R = unknown> = {
+  /** The result passed to close() */
+  result: R | undefined;
+  /** Call this synchronously to prevent the close */
+  cancel(): void;
+};
+
 export abstract class DialogRef<R = unknown, C = unknown> implements Pick<
   CdkDialogRef<R, C>,
-  "close" | "closed" | "disableClose" | "componentInstance"
+  "closed" | "disableClose" | "componentInstance"
 > {
   abstract readonly isDrawer?: boolean;
 
   // --- From CdkDialogRef ---
-  abstract close(result?: R, options?: DialogCloseOptions): Promise<DialogCloseRef>;
+  abstract close(result?: R, options?: DialogCloseOptions): DialogCloseRef;
   abstract readonly closed: Observable<R | undefined>;
   abstract disableClose: boolean | undefined;
   /**
@@ -68,6 +74,11 @@ export abstract class DialogRef<R = unknown, C = unknown> implements Pick<
    * Does not work with drawer dialogs.
    **/
   abstract componentInstance: C | null;
+
+  /**
+   * Emits before the dialog closes. Subscribers may call event.cancel() to prevent the close.
+   */
+  abstract readonly beforeClose$: Observable<BeforeCloseEvent<R>>;
 }
 
 export type DialogConfig<D = unknown, R = unknown> = Pick<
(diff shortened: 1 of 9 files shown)


libs/tools/send/send-ui/src/add-edit/send-add-edit-dialog.component.ts
```diff
@@ -111,6 +109,23 @@ export class SendAddEditDialogComponent {
...
+      if (!this.sendFormService.sendFormHasEdits()) {
+        return;
+      }
+      event.cancel();
+      void this.sendFormService.promptForUnsavedEdits().then((canClose) => {
+        if (canClose) {
+          this.dialogRef.close(event.result);
+        }
+      });
+    });
```
BTreston, line 128: IMO This is a much nicer way to be able to encapsulate our validation logic and define it directly in the caller rather than have to bind it to a predicate function in the dialog config.

libs/components/src/dialog/dialog.service.ts
```diff
@@ -302,38 +315,52 @@ export class DialogService {
...
       scrollStrategy: this.defaultScrollStrategy,
       positionStrategy: config?.positionStrategy ?? new ResponsivePositionStrategy(),
       closeOnNavigation: config?.closeOnNavigation,
       injector,
-      ...otherConfig,
+      ...config,
+      disableClose: true,
     };
 
     ref.cdkDialogRefBase = this.dialog.open<R, D, C>(componentOrTemplateRef, _config);
```
BTreston, line 340: This bit here is my least favorite part of this work, but I think it can be worked around in better way. I just did not want to spend the extra time at this stage.

libs/components/src/dialog/dialog.service.ts
```diff
@@ -196,27 +209,38 @@ class DrawerDialogRef<R = unknown, C = unknown> implements DialogRef<R, C> {
...
+    let cancelled = false;
+    this._beforeClose$.next({
+      result,
+      cancel: () => {
+        cancelled = true;
+      },
+    });
+    if (cancelled) {
+      return { closed: false };
     }
```
eliykat, line 242: I was initially confused as to why the `cancel()` callback would be called before it's evaluated in the next `if` block. As far as I can tell, it does work because `_beforeClose$.next` executes all subscriber callbacks synchronously, _and_ the subscriber callback in your example is also synchronous. But that's not something I want to rely on - it's mixing reactive and imperative code in a way that is confusing and seems brittle. e.g. if the subscriber does not call `cancel` synchronously (e.g. it's mixing it with another async observable), then you do have a race condition.

I'm also not sure that `next` executing subscriber callbacks synchronously is something to be relied on as part of rxjs' public API - it feels like an implementation detail (I could be wrong on that though).

libs/tools/send/send-ui/src/add-edit/send-add-edit-dialog.component.ts
```diff
@@ -111,6 +109,23 @@ export class SendAddEditDialogComponent {
...
+      // If there are no unsaved edits, allow close immediately
+      if (!this.sendFormService.sendFormHasEdits()) {
+        return;
+      }
+      event.cancel();
+      void this.sendFormService.promptForUnsavedEdits().then((canClose) => {
+        if (canClose) {
+          this.dialogRef.close(event.result);
+        }
+      });
```
eliykat, line 127: This code flow is unintuitive - I think mostly because we are trying to make async flows sync.

1. if this is the second time round, close
2. do synchronous business logic, close or proceed
3. cancel closing because now we're going to do async stuff
4. do async stuff
5. based on the result, call close again
6. now it hits the early exit in step 1

This is non-obvious for all callers to implement.

libs/components/src/dialog/dialog.service.ts
```diff
@@ -196,27 +209,38 @@ class DrawerDialogRef<R = unknown, C = unknown> implements DialogRef<R, C> {
...
+
+    let cancelled = false;
+    this._beforeClose$.next({
+      result,
+      cancel: () => {
+        cancelled = true;
+      },
+    });
+    if (cancelled) {
+      return { closed: false };
```
eliykat, line 241: This has a subtle and I think negative impact on the API: this method can return `closed: false`, but then close asynchronously later (due to the subscriber on `beforeClose$` invoking it).

At least with it being async, it properly awaits the custom business logic, and the `closed` value accurately reflects whether close was actually cancelled or not.

eliykat (commented):
Unfortunately I suspect that this really is an async flow that needs to be async. Happy to discuss/iterate further though if you want to keep exploring these ideas.

PR #20311, opened 2026-04-21
[PM-33473] Remove pm-29594-update-individual-subscription-page feature flag

amorask-bitwarden:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-33473

## 📔 Objective

Remove the `pm-29594-update-individual-subscription-page` feature flag and its flagged code paths per the Feature Flags Lifecycle process.

Changes:
- Remove the `PM29594_UpdateIndividualSubscriptionPage` enum member and its defaults-map entry from `libs/common/src/enums/feature-flag.enum.ts`.
- Delete `postReinstatePremium` and `postAccountStorage` (abstract + implementation) from `libs/common/src/abstractions/api.service.ts` and `libs/common/src/services/api.service.ts`. Drop the now-orphaned `StorageRequest` import.
- Simplify `subscription.component.ts` — `showSubscriptionPageLink$` now always calls `accountBillingClient.getSubscription()` on cloud. The prior flag branch and the dead self-hosted fallback (unreachable because the parent template gates on `@if (!selfHosted)`) are removed. `ConfigService`, `LogService`, `BillingAccountProfileStateService`, and `FeatureFlag` deps drop with them.
- Update `individual-billing-routing.module.ts` — remove the `isSubscriptionPageEnabled` `canMatch` guard and collapse the `/user-subscription` route tree to two routes (self-hosted via `canMatch: [isSelfHosted]`, cloud as default), matching the existing two-route `/premium` pattern in the same file.
- Delete `UserSubscriptionComponent` (`.ts` and `.html`) and drop it from module declarations.
- Clean up `AdjustStorageDialogComponent`: tighten `organizationId` from optional to required (matches the sole remaining caller `OrganizationSubscriptionCloudComponent`), remove the now-dead `postAccountStorage` else-branch, remove the `apiService` dependency and orphaned imports, and drop the redundant protected field.

Paired with the companion server PR that removes the corresponding endpoints and keeps `GET /accounts/subscription` as self-hosted-only for the `SelfHostedAccountSubscriptionComponent`.


sbrown-livefront (approved):
🧹 ✨

PR #20316, opened 2026-04-22
[PM-35491] Replace start-to-tray with autostart detection

quexten:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-35491

## 📔 Objective

There are three ways to start the desktop app, through the desktop icon, through the CLI/executable and through auto-start. For the former two, the user expects to see the window. For the latter, it is annoying / bad UX for the desktop window to appear. This PR removes the "start to tray" setting, and replaces it by detecting the above cases, and starting to tray automatically for the auto-start case.

Specifically, the behavior is as follows:
- If started through the CLI, the executable, or the desktop shortcut, the window always shows
- If started by auto-start, the app always starts to tray

Autostart is modified to always provide the `--autostart` parameter. If this parameter is present the app starts to the tray. Otherwise, the app is started in the foreground. Thus, running via the CLI, or the `.desktop` file will show the app.

For autostart, there are three mechanisms:
- Electron
  - Already had support for parameters, we just add the parameter
- Flatpak autostart portal
  - The autostart portal code already had support for parameters, and we just add the parameter
- Snap
  - Snap in electron builder-built electron apps runs a custom .desktop app in a pre-defined path that we cannot control directly. Thus, we simply re-write this file to add the auto-start parameter.

This also fixes QA-automation of the desktop app, which is currently broken, since launch to tray is default enabled.


apps/desktop/src/main/messaging.main.ts
```diff
@@ -130,14 +131,16 @@ export class MessagingMain {
   private addOpenAtLogin() {
     if (process.platform === "linux") {
       if (isFlatpak()) {
-        autostart.setAutostart(true, []).catch((e) => {});
+        autostart.setAutostart(true, [AUTOSTART_FLAG]).catch((e) => {});
+      } else if (isSnapStore()) {
+        this.updateSnapAutostartExec();
```
quexten, line 136: Note, snap currently constantly auto-starts, so we don't actually have a check here and just update it. There is a follow-up feature request to make it possible to disable snap auto-start. (This is not already the case because of how electron-builder handles auto-start).

dani-garcia (approved):

PR #20332, opened 2026-04-22
[PM-35354] Fix empty error toast when DisableHideEmail policy is not respected on browser

harr1424:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-35354

## 📔 Objective

It was discovered that when the Send related policy with option “Always show member’s email address” was enabled, the error toast displayed in the browser extension when saving a send that did not adhere to the policy was empty.

~~This missing `i18n` translation key `saveSendEditsFailed` has been added to browser, although this won't be used for the error specific to this situation. Instead,~~ a more descriptive `i18n` translation key `hideEmailPolicyInEffect` has been introduced to browser, web, and desktop. 

During resolution of the defect described above it was also determined that on browser, when the `DisableHideEmail` policy was enabled, creating a new Send still exposed a checkbox allowing the Send creator to attempt to hide their email. This PR also reworks how policy detection logic is used in the template, ensuring that when the policy is active, the disallowed option is not presented to the user. 

The screenshots below demonstrate the new `i18n` translation key in the browser extension and web client. 

### Web
Screenshot 2026-04-22 at 16 31 11

### Desktop
Screenshot 2026-04-22 at 16 27 01

### Browser
Screenshot 2026-04-22 at 16 26 41


apps/browser/src/_locales/en/messages.json
```diff
@@ -6418,5 +6418,13 @@
   "progressBar": {
     "message": "Progress bar",
     "description": "This is the name of a page component that displays progress to the user. This string will be used to label the progress bar for a screenreader when multiple progress bars are on a page, like 'Progress bar 1' and 'Progress bar 2'."
+  },
+  "saveSendEditsFailed": {
```
itsadrago, line 6422: Is this needed? I don't see it referenced in the modified files

apps/browser/src/_locales/en/messages.json
```diff
@@ -6418,5 +6418,13 @@
...
+    "message": "Saving Send edits failed",
+    "description": "'Send' is a noun and the name of a feature called 'Bitwarden Send'. It should not be translated."
+  },
+  "hideEmailPolicyInEffect": {
+    "message": "Your organization requires email visibility. Uncheck \"Hide your email from viewers\" to save.",
+    "description": "'Send' is a noun and the name of a feature called 'Bitwarden Send'. It should not be translated."
```
itsadrago: "Send" isn't mentioned in the message. The same is true for the new translation in the other files

apps/browser/src/_locales/en/messages.json
harr1424: Formerly this was missing from browser, which is why the toast was empty, ~~but it has been replaced by `hideEmailPolicyInEffect`, I've pushed a commit removing this from the three localization files,~~ and have also removed the `description` field that provided unnecessary context regarding the noun "Send".

itsadrago (approved):

PR #20334, opened 2026-04-22
Auth/PM-35336 - TokenService - prevent stale access token retrieval to fix logout on org user confirm

JaredSnider-Bitwarden:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-35336

Original attempted fix: https://github.com/bitwarden/clients/pull/20309
We initially thought this was due to StateProvider providing state data so there's a bunch of atomicity improvements in that PR that proved unnecessary to fix the actual bug at play. We swapped to this PR to strip down the changes to only what is absolutely necessary to fix the bug for the next release as it is a release blocker. 

## 📔 Objective

To solve the below bug: 

## Bug Description

When a user is JIT-provisioned via SSO into an **MP encryption org** (non-TDE), they are unexpectedly
logged out immediately after being confirmed to the organization. This happens in both auto-confirm
and manual confirm flows.

<details>
<summary>Root Cause</summary>

All four token types share a dual-storage pattern: a `_MEMORY` key and a `_DISK` key. Every getter
reads Memory first and falls back to Disk. But no setter cleared the other location on write —
meaning Memory and Disk could silently diverge. In the `TokenService`, we erroneously counted on external clears instead of internally enforcing the single token storage location invariant.

The divergence was triggered by the following sequence:

1. **Token written to Memory on login**: New users JIT-provisioned via SSO use the default vault
   timeout settings (`LogOut + 15min`), causing the access token to be written to
   `ACCESS_TOKEN_MEMORY` only. Note: the fact that we have [dev overrides](https://github.com/bitwarden/clients/blob/main/apps/web/src/app/core/core.module.ts#L193-L198) for the default vault timeout made this originally unable to be reproduced locally. 

2. **JIT provisioning triggers a background token migration**: Once JIT-provisioned and after the user set's an initial MP, the user's password becomes available, making `Lock` a valid vault timeout action. Since
   https://github.com/bitwarden/clients/pull/17731 made `availableVaultTimeoutActions$`
   fully reactive and https://github.com/bitwarden/clients/pull/19594 introduced token migration as a side effect, this availability change is observed mid-session — firing a `null → Lock`
   transition that calls `migrateTokenStorage` and copies the token to `ACCESS_TOKEN_DISK`.
   Memory is **not** cleared.

3. **`refreshIdentityToken` writes the new token to Disk only**: The org confirmation triggers a
   `SyncOrgKeys` push notification → `fullSync` → `refreshIdentityToken()`, which writes the new
   JWT (containing the new org claim) to `ACCESS_TOKEN_DISK`. Memory is **not** updated.

4. **Every `getAccessToken` call returns the stale Memory value**: The getter's Memory-first read
   always returns the old token — the one without the org claim. This causes a 403 on subsequent
   API calls, which triggers logout.

</details>

The correct fix will be to have the `TokenService` enforce storing data in a single slot only at a time by clearing the other locations on sets. 

## Detailed Changes

<details>
<summary>Abstractions</summary>

**`libs/common/src/auth/abstractions/token.service.ts`**
1. Updated JSDoc for `setAccessToken`, `setClientId`, and `setClientSecret` to document the new single-storage-location invariant — each setter now guarantees the alternate storage location is cleared after writing, so no token can exist in both memory and disk simultaneously.
2. Corrected the `setAccessToken` note about secure storage: previously described as storing tokens in secure storage, now accurately describes encrypting the token with a key held in secure storage and writing the ciphertext to disk.

</details>

<details>
<summary>Services</summary>

**`libs/common/src/auth/services/token.service.ts`**
1. `setAccessToken` — added cross-location clearing after every write path (SecureStorage happy path, SecureStorage key-failure fallback, SecureStorage error fallback, Disk, and Memory) so only one location holds the access token at any time; before this change, switching from Memory→Disk or Disk→Memory could leave a stale value in the evicted location.
2. Extracted private `clearAccessTokenMemoryLocation(userId)` and `clearAccessTokenDiskLocation(userId)` helpers to DRY up the repeated clear patterns and removed the stale `// TODO: PM-6408` inline eviction comment in the SecureStorage path that was partially doing this job.
3. `setRefreshToken` — same cross-location clearing pattern applied to all write paths, plus extraction of `clearRefreshTokenMemoryLocation(userId)` and `clearRefreshTokenDiskLocation(userId)` helpers.
4. `setClientId` / `setClientSecret` — same pattern applied; extracted `clearClientIdMemoryLocation`, `clearClientIdDiskLocation`, `clearClientSecretMemoryLocation`, and `clearClientSecretDiskLocation` helpers.
5. `clearAccessToken` / `clearRefreshToken` refactored to delegate to the new private helpers instead of duplicating the inline state update calls.

**`libs/common/src/key-management/vault-timeout/services/vault-timeout-settings.service.ts`**
1. Removed the `tokenService.clearTokens(userId)` pre-eviction call that ran before `setTokens` when switching to the LogOut action — this explicit eviction is now redundant because each `setToken*` setter already clears the alternate location internally, making the prior call both unnecessary and a latent race condition source.

</details>

<details>
<summary>Tests</summary>

**`libs/common/src/auth/services/token.service.spec.ts`**
1. Added `cross-location clearing` describe blocks under `setAccessToken`, `setRefreshToken`, `setClientId`, and `setClientSecret`; each block seeds the alternate storage location and asserts it is nulled out after a write — covering the Disk→clears-Memory and Memory→clears-Disk paths for all four token types.
2. Added SecureStorage sub-blocks for `setAccessToken` and `setRefreshToken` covering the happy path, silent key-retrieval failure, and secure-storage error fallback — verifying that Memory is cleared even when the SecureStorage write falls back to plaintext disk.

**`libs/common/src/key-management/vault-timeout/services/vault-timeout-settings.service.spec.ts`**
1. Updated the "should clear the tokens when the timeout is not never and the action is log out" test to assert `tokenService.setTokens` is called (not `tokenService.clearTokens`) — reflecting that storage migration is now handled internally by the token service rather than by a separate pre-eviction step in `VaultTimeoutSettingsService`.

</details>

## 📸 Screenshots

New video just on these changes

https://github.com/user-attachments/assets/bb5560d5-f5cb-43e6-96e9-3b1cb0d9e1db

### DIFF
--- a/libs/common/src/auth/services/token.service.ts
+++ b/libs/common/src/auth/services/token.service.ts
@@ -371,10 +371,8 @@ export class TokenService implements TokenServiceAbstraction {
           // so that the caller can use it immediately.
           decryptedAccessToken = accessToken;
 
-          // TODO: PM-6408
-          // 2024-02-20: Remove access token from memory so that we migrate to encrypt the access token over time.
-          // Remove this call to remove the access token from memory after 3 months.
-          await this.singleUserStateProvider.get(userId, ACCESS_TOKEN_MEMORY).update((_) => null);
+          // Then, we can clear the memory state location to enforce that a single location holds the access token at a time.
+          await this.clearAccessTokenMemoryLocation(userId);
         } catch (error) {
           this.logService.error(
             `SetAccessToken: storing encrypted access token in secure storage failed. Falling back to disk storage.`,
@@ -387,22 +385,32 @@ export class TokenService implements TokenServiceAbstraction {
             .update((_) => accessToken, {
               shouldUpdate: (previousValue) => previousValue !== accessToken,
             });
+          // Then, we can clear the memory state location to enforce that a single location holds the access token at a time.
+          await this.clearAccessTokenMemoryLocation(userId);
         }
 
         return decryptedAccessToken;
       }
-      case TokenStorageLocation.Disk:
+      case TokenStorageLocation.Disk: {
         // Access token stored on disk unencrypted as platform does not support secure storage
-        return await this.singleUserStateProvider
+        const newAccessToken = await this.singleUserStateProvider
           .get(userId, ACCESS_TOKEN_DISK)
           .update((_) => accessToken, {
             shouldUpdate: (previousValue) => previousValue !== accessToken,
           });
-      case TokenStorageLocation.Memory:
+        // Then, we can clear the memory state location to enforce that a single location holds the access token at a time.
+        await this.clearAccessTokenMemoryLocation(userId);
+        return newAccessToken;
+      }
+      case TokenStorageLocation.Memory: {
         // Access token stored in memory due to vault timeout settings
-        return await this.singleUserStateProvider
+        const newAccessToken = await this.singleUserStateProvider
           .get(userId, ACCESS_TOKEN_MEMORY)
           .update((_) => accessToken);
+        // Then, we can clear the disk state location to enforce that a single location holds the access token at a time.
+        await this.clearAccessTokenDiskLocation(userId);
+        return newAccessToken;
+      }
     }
   }
 
@@ -453,11 +461,19 @@ export class TokenService implements TokenServiceAbstraction {
       await this.clearAccessTokenKey(userId);
     }
 
-    // Clear tokens from disk storage (all platforms)
+    // Clear from disk and memory.
+    await this.clearAccessTokenDiskLocation(userId);
+    await this.clearAccessTokenMemoryLocation(userId);
+  }
+
+  private async clearAccessTokenMemoryLocation(userId: UserId): Promise<void> {
+    await this.singleUserStateProvider.get(userId, ACCESS_TOKEN_MEMORY).update((_) => null);
+  }
+
+  private async clearAccessTokenDiskLocation(userId: UserId): Promise<void> {
     await this.singleUserStateProvider.get(userId, ACCESS_TOKEN_DISK).update((_) => null, {
       shouldUpdate: (previousValue) => previousValue !== null,
     });
-    await this.singleUserStateProvider.get(userId, ACCESS_TOKEN_MEMORY).update((_) => null);
   }
 
   async getAccessToken(userId: UserId): Promise<string | null> {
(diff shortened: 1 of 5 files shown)


enmande (approved):
Thank you for the extended discussion on this, and the clear documentation.
The extensive test cases illustrate the behavioral contract and expectations well for this update. We have to do some scooting around with memory/disk, but the well-named helpers do a nice job of keeping that expressive and clean. LGTM ✨

libs/common/src/auth/services/token.service.ts
```diff
@@ -387,22 +385,32 @@ export class TokenService implements TokenServiceAbstraction {
...
         // Access token stored on disk unencrypted as platform does not support secure storage
-        return await this.singleUserStateProvider
+        const newAccessToken = await this.singleUserStateProvider
           .get(userId, ACCESS_TOKEN_DISK)
           .update((_) => accessToken, {
             shouldUpdate: (previousValue) => previousValue !== accessToken,
           });
-      case TokenStorageLocation.Memory:
+        // Then, we can clear the memory state location to enforce that a single location holds the access token at a time.
+        await this.clearAccessTokenMemoryLocation(userId);
```
Thomas-Avery, line 402: 💭  Non-blocking but it looks like we have a repeated pattern. If setting to disk clear memory. If setting to memory clear disk. When refactoring in the future something to consider that might belong in a state service layer: 

```
  private async setAccessTokenDisk(accessToken: string, userId: UserId): Promise<void> {
    await this.singleUserStateProvider.get(userId, ACCESS_TOKEN_DISK).update((_) => accessToken, {
      shouldUpdate: (previousValue) => previousValue !== accessToken,
    });

    await this.clearAccessTokenMemoryLocation(userId);
  }

  private async setAccessTokenMemory(accessToken: string, userId: UserId): Promise<void> {
    await this.singleUserStateProvider.get(userId, ACCESS_TOKEN_MEMORY).update((_) => accessToken, {
      shouldUpdate: (previousValue) => previousValue !== accessToken,
    });

    await this.clearAccessTokenDiskLocation(userId);
  }
```

Thomas-Avery (approved):
LGTM, thank you

libs/common/src/auth/services/token.service.ts
JaredSnider-Bitwarden: Thank you! I will keep that in mind for future refactors.

PR #20336, opened 2026-04-22
[PM-34155] Enhance policy service with accepted policies retrieval and update sync response structure for new sync response

JaredScar:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-34155

## 📔 Objective

Updates the frontend to handle the new newPolicies coming back from the sync endpoint


JaredScar:
@claude review please

libs/common/src/admin-console/services/policy/default-policy.service.ts
```diff
@@ -40,7 +40,34 @@ export class DefaultPolicyService implements PolicyService {
...
+      map(([policies, organizations]) => {
+        const orgDict = Object.fromEntries(organizations.map((o) => [o.id, o]));
+        return policies.filter((policy) => {
+          const org = orgDict[policy.organizationId];
+          // No org found: include the policy (consistent with enforcedPolicyFilter behaviour)
+          return !org || org.status === OrganizationUserStatusType.Confirmed;
+        });
+      }),
+    );
+  }
```
eliykat: I didn't foresee needing to rely on `OrganizationService` here - this is more logic than I want in a simple "get from state" operation, and for a non-feature flagged change intended to maintain existing behaviour.

Let's simplify it - the server is still sending both `policies` and `policiesNew`, so let's continue using both so we can keep them separate. `policies` will keep getting saved to its current location (and used by this observable without any changes), `policiesNew` will have its own `upsert` and `replace` methods saved to its own disk state.

There is a bit of boilerplate around state management so this may benefit from a separate `NewPolicyService` which `PolicyService` can shim out to later, rather than trying to contain it all in the same class. This also helps us keep old & new code separate (e.g. when we add the sdk integration it can just be added to the new service).

Let me know what you think.

eliykat (changes requested):
This matches the ticket, but seeing it implemented I think we may need to separate old & new code a bit more.

JaredScar:
@claude please re-review

coroiu:
@JaredScar Is this ready for review? It seems @eliykat requested that the PR have changes removed but I only see additions? (I might be misunderstanding the GitHub UI though so that's why I'm asking)

JaredScar:
> @JaredScar Is this ready for review? It seems @eliykat requested that the PR have changes removed but I only see additions? (I might be misunderstanding the GitHub UI though so that's why I'm asking)

Should be, yup! He wanted to essentially have a new service made to separate these, so more code would make sense 😛

JaredScar:
@claude Can you please re-review?

libs/common/src/platform/sync/default-sync.service.ts
```diff
@@ -427,6 +430,17 @@ export class DefaultSyncService extends CoreSyncService {
     return await this.policyService.replace(policies, userId);
   }
 
+  private async syncNewPolicies(response: PolicyResponse[] | undefined, userId: UserId) {
+    if (response == null) {
+      return;
+    }
```
eliykat: This property is only sent when the feature flag is turned on, so there's a risk that there is a timing mismatch in how it's propagated. e.g. the client may switch over to the new system while the server isn't sending the new property yet. Seems unlikely, but just in case, we should fall back to `response.policies` if `policiesNew == null || policiesNew.length === 0`.

This ensures that the new service is always seeded with data.

libs/common/src/admin-console/services/policy/default-new-policy.service.ts
```diff
@@ -0,0 +1,53 @@
...
+    ]).pipe(
+      map(([policies, organizations]) => {
+        const orgDict = Object.fromEntries(organizations.map((o) => [o.id, o]));
+        return policies.filter(
+          (policy) =>
+            orgDict[policy.organizationId]?.status === OrganizationUserStatusType.Accepted,
+        );
+      }),
+    );
+  }
```
eliykat: Now that this is in a separate service with its own state, we don't need to do this dance to separate out accepted policies. That was only for maintaining backwards compatibility when it was part of the old service.

When the SDK integration is added, all `policiesNew` (i.e. both accepted and confirmed policies) will be handed off to the SDK, and the SDK logic will sort out confirmed vs. accepted policies. No need for us to handle it here.

For now, I recommend deleting this method and the public interface. This PR can just be limited to bootstrapping and seeding the new service.

libs/common/src/admin-console/abstractions/policy/new-policy.service.abstraction.ts
```diff
@@ -0,0 +1,30 @@
...
+
+import { UserId } from "../../../types/guid";
+import { PolicyData } from "../../models/data/policy.data";
+import { Policy } from "../../models/domain/policy";
+
+/**
+ * Service for managing the new policy format (`policiesNew`) received from the server.
+ * This is kept separate from {@link PolicyService} so that old and new policy state can
+ * evolve independently, and so the SDK integration can be added here without touching
+ * existing policy logic.
```
eliykat: Not wrong exactly, but could be more direct in what the new service is actually about. This can often be a problem with AI documentation - it's worth reviewing and often rewriting it with your (human) context of what we're doing.

e.g.

> Service for managing policy state and enforcement using the SDK.
> Policies can be enforced in both accepted and confirmed statuses.
> This is internal to AC Team for now and should not be used by
> outside consumers.

libs/common/src/admin-console/services/policy/default-policy.service.spec.ts
```diff
@@ -537,6 +537,34 @@ describe("PolicyService", () => {
         },
       ]);
     });
+
+    it("returns policies for organizations where the user has Accepted status", async () => {
+      // policies$ is a simple state passthrough and does not filter by org membership status.
+      // Accepted-org policies belong in acceptedPolicies$ (on NewPolicyService) but must
+      // still pass through here so downstream consumers are not silently missing data.
```
eliykat: I don't really understand what this is trying to say. The new state is not downstream of this state. I suggest changes to this file can be reverted, but please correct me if I'm missing something here.

libs/common/src/admin-console/abstractions/policy/new-policy.service.abstraction.ts
```diff
@@ -0,0 +1,30 @@
...
+/**
+ * An "internal" extension of {@link NewPolicyService} that allows updating the local
+ * `policiesNew` state. This does not update any policies on the server.
+ */
+export abstract class InternalNewPolicyService extends NewPolicyService {
+  /** Upsert a single policy into the `policiesNew` local state. */
+  abstract upsert: (policy: PolicyData, userId: UserId) => Promise<void>;
+  /** Replace all `policiesNew` local state for a user. */
+  abstract replace: (policies: { [id: string]: PolicyData }, userId: UserId) => Promise<void>;
+}
```
eliykat: Please review any existing uses of these methods on the current interface. There are a couple of places where policies are updated locally which will also need to call these new methods so that the new state is also updated. We are effectively maintaining both old and new state in parallel.

The current interface also has `syncPolicy` which needs to be accounted for as well - I would prefer not to add that here, it can just call `replace`.

eliykat (changes requested):

eliykat (dismissed):
LGTM once tests and builds are fixed.

coroiu (commented):
Re-request review when builds and tests are fixed

eliykat (dismissed):

JaredSnider-Bitwarden (approved):
Auth file change LGTM!

eliykat (approved):

coroiu (approved):

PR #20348, opened 2026-04-22
[PM-35496] - Fix WASM panic

### LINKED ISSUE #20325: Extension panics with data-encoding encode_len overflow on large organization vault (14k+ ciphers) — at-risk-cipher-badge breaks autofill end-to-end
## Steps To Reproduce
1. Log into the browser extension with an account that is a member of an organization containing ~14,000+ ciphers.
2. Sync the vault.
3. Attempt to autofill on any login page.

## Expected Result
Sync completes without error. Autofill works.

## Actual Result
A Rust panic fires in the WASM SDK on every sync and on every autofill attempt. Autofill is completely broken across all sites for this user.

The panic:

```
panicked at /home/runner/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/data-encoding-2.10.0/src/lib.rs:1278:9:
assertion failed: len <= usize::MAX / 512
```

That assertion lives in `Encoding::encode_len` and fires when encoding a byte slice larger than `usize::MAX / 512` (~8 MiB on wasm32).

### Root cause from the service worker console

```
[CipherService]: decrypt complete took 530.1ms Array(1)
panicked at data-encoding-2.10.0/src/lib.rs:1278:9: assertion failed: len <= usize::MAX / 512
BadgeService: State function "at-risk-cipher-badge" threw an error RuntimeError: unreachable
BadgeService: State function "autofill-badge-updater" threw an error RuntimeError: unreachable
Uncaught (in promise) RuntimeError: unreachable
```

The `at-risk-cipher-badge` state function is the panic origin. Because the badge state chain then collapses, `autofill-badge-updater` also throws `unreachable`, which breaks autofill end-to-end.

With ~14,137 ciphers in the organization, the at-risk-cipher-badge pipeline appears to build an aggregate byte buffer whose length exceeds `encode_len`'s wasm32 input-size limit.

## Additional Context

- **Full service worker console log (gist):** https://gist.github.com/ianmartorell/ae2a320713471c1d6b622389b92ce01c
- **Reinstalling the extension does not fix it** — panic returns on first sync after login.
- Logout/login does not fix it.
- `bw` CLI (`bw sync`, `bw list items`, `bw export`) against the same account works fine, so the underlying vault data is valid; the panic is specific to the extension's WASM badge pipeline.
- There is no user-facing toggle for the at-risk-cipher-badge feature (see active community thread: https://community.bitwarden.com/t/options-to-disable-or-dismiss-permanent-change-at-risk-password-warnings/91972), so affected users have no workaround short of downgrading the extension.

**Vault composition:**
- Personal vault: ~1,561 ciphers (all types)
- Single organization: ~14,137 ciphers (mostly type 1 logins)
- Total: ~15,700 ciphers

## Reference

Similar to resolved #15899 (large vault causing extension crash in 2025.7.0, fixed in 2025.9.0), but with a distinct code path — the panic here is specifically in the BadgeService / at-risk-cipher-badge state function, not the general cipher decryption path.

## Operating System

macOS

## Operating System Version

15.x (Darwin 25.3.0, Apple Silicon)

## Web Browser

Chrome

## Browser Version

147.0.7727.56 (Official Build) (arm64)

## Build Version

2026.3.0

jaasen-livefront:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-35496

## 📔 Objective

Fixes a WASM panic that permanently breaks autofill for users with large vaults (~14,000+ ciphers).

### Root cause

The Bitwarden WASM SDK's `symmetric_encrypt_string` calls `data-encoding`'s `encode_len` internally. On `wasm32`, `usize` is 32-bit, causing `encode_len` to overflow and panic when the input exceeds `usize::MAX / 512 ≈ 8.38 MiB`.

`LocalBackedSessionStorageService` encrypts all session state before persisting it to `chrome.storage.local`. For large vaults, the `CIPHERS_MEMORY` state (the decrypted `CipherView[]` cache) serializes to ~30 MiB of JSON — well over the limit. When `encryptString` is called with this payload, Rust panics and calls the WASM `unreachable` instruction. The WASM module instance is now permanently faulted for the rest of the service worker session. Every subsequent WASM call throws `RuntimeError: unreachable executed`, including autofill fill operations.

From the user's perspective: the vault displays normally in the popup (which runs in a separate renderer process with its own WASM instance), but clicking autofill on any site silently does nothing.

### Fix

Adds a size check in `LocalBackedSessionStorageService.updateLocalSessionValue` before calling `encryptString`. If the serialized payload exceeds 7 MiB (a conservative margin below the 8.38 MiB `wasm32` limit), local-storage persistence is skipped and a warning is logged. The in-memory cache remains populated so the extension continues to function normally for the current session. On service worker restart, the vault re-decrypts from the encrypted cipher data in `chrome.storage.local` — the same cold-start path that already occurs for these users today, since the panic was preventing the write from ever succeeding.

## 📸 Screenshots

N/A

### DIFF
--- a/apps/browser/src/autofill/services/autofill-badge-updater.service.ts
+++ b/apps/browser/src/autofill/services/autofill-badge-updater.service.ts
@@ -22,8 +22,12 @@ export class AutofillBadgeUpdaterService {
   ) {}
 
   init() {
+    // Subscribe to ciphers$ (encrypted) as a change signal only — actual decryption happens
+    // lazily in calculateCountText via getAllDecryptedForUrl, which hits the warm shareReplay
+    // cache on subsequent calls. The delay gives cipherViews$ time to re-decrypt after a change
+    // before getAllDecryptedForUrl is called.
     const ciphers$ = this.accountService.activeAccount$.pipe(
-      switchMap((account) => (account?.id ? this.cipherService.ciphers$(account?.id) : of([]))),
+      switchMap((account) => (account?.id ? this.cipherService.ciphers$(account.id) : of([]))),
     );
 
     this.badgeService.setState(StateName, (tab) => {
@@ -31,7 +35,7 @@ export class AutofillBadgeUpdaterService {
         account: this.accountService.activeAccount$,
         enableBadgeCounter:
           this.badgeSettingsService.enableBadgeCounter$.pipe(distinctUntilChanged()),
-        ciphers: ciphers$.pipe(delay(100)), // Delay to allow cipherService.getAllDecryptedForUrl to pick up changes
+        ciphers: ciphers$.pipe(delay(100)),
       }).pipe(
         mergeMap(async ({ account, enableBadgeCounter }) => {
           if (!account || !enableBadgeCounter) {
--- a/apps/browser/src/autofill/services/autofill-badge-updater.service.ts
+++ b/apps/browser/src/autofill/services/autofill-badge-updater.service.ts
@@ -22,12 +22,8 @@ export class AutofillBadgeUpdaterService {
   ) {}
 
   init() {
-    // Subscribe to ciphers$ (encrypted) as a change signal only — actual decryption happens
-    // lazily in calculateCountText via getAllDecryptedForUrl, which hits the warm shareReplay
-    // cache on subsequent calls. The delay gives cipherViews$ time to re-decrypt after a change
-    // before getAllDecryptedForUrl is called.
     const ciphers$ = this.accountService.activeAccount$.pipe(
-      switchMap((account) => (account?.id ? this.cipherService.ciphers$(account.id) : of([]))),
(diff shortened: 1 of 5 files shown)


nikwithak (dismissed):

jaasen-livefront:
@nikwithak Given how we're killing the `CipherList` feature I took a different route for dealing with this issue. If you have a spare moment to take a look that would be awesome. Ty!

nikwithak (approved):

jaasen-livefront:
@djsmith85 If you have a spare moment to take a look that would be much appreciated. Ty!

jaasen-livefront:
This has been addressed within the sdk already.

https://github.com/bitwarden/sdk-internal/pull/913

PR #20357, opened 2026-04-23
[PM-35698] filters not rendering on page load

JaredScar:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-35698

## 📔 Objective

Adds back changeDetection.default since it causes an issue utilizing OnPush where you have to click the filters twice for it to take effect


BTreston (commented):
What is the lift to actually fix this rather than revert change detection?

apps/web/src/app/vault/individual-vault/vault-filter/shared/components/vault-filter-section.component.ts
```diff
@@ -68,13 +69,13 @@ export class VaultFilterSectionComponent implements OnInit, OnDestroy {
     this.vaultFilterService.collapsedFilterNodes$
       .pipe(takeUntil(this.destroy$))
       .subscribe((nodes) => {
-        this.collapsedFilterNodes = nodes;
+        this.collapsedFilterNodes.set(nodes);
       });
```
BTreston: I think we can just wrap this in a `toSignal()` call rather than subscribing.

apps/web/src/app/vault/individual-vault/vault-filter/shared/components/vault-filter-section.component.ts
```diff
@@ -68,13 +69,13 @@ export class VaultFilterSectionComponent implements OnInit, OnDestroy {
...
   }
 
   async ngOnInit() {
     this.section?.data$?.pipe(takeUntil(this.destroy$)).subscribe((data) => {
-      this.data = data;
+      this.data.set(data);
     });
```
BTreston: Same as above, try wrapping in `toSignal()` instead of subscribing.

BTreston (changes requested):

apps/web/src/app/vault/individual-vault/vault-filter/shared/components/vault-filter-section.component.ts
```diff
@@ -55,48 +57,39 @@ export class VaultFilterSectionComponent implements OnInit, OnDestroy {
...
-  }
+  ) {}
 
-  ngOnDestroy() {
-    this.destroy$.next();
-    this.destroy$.complete();
+  ngOnInit() {
+    this.section?.data$
+      ?.pipe(takeUntilDestroyed(this.destroyRef))
+      .subscribe((data) => this.data.set(data));
```
BTreston: were we not able to wrap this in `toSignal`?

jaasen-livefront (approved):
Dig it!

BTreston (approved):

PR #20359, opened 2026-04-23
[PM-28191] Replace SaveCollectionsWithServerAdmin API calls to use SDK 

nikwithak:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-28191

## 📔 Objective

Adds a feature-flagged call to the SDK for `saveColllectionsWithServerAdmin`, replacing the legacy TS path.

Note: The Admin API calls were already migrated as part of the CRUD operations, since those were already intertwined logic. 🧇 This is using the feature flag created for that epic, but the naming is no longer accurate. We may consider changing the flag name, but that requires coordination across repos & LD.

This flag will also gate operations not yet in the SDK, once they are made available - `bulkUpdateCollectionsWithServer`, `moveManyWithServer`.


libs/common/src/vault/services/cipher.service.ts
```diff
@@ -1333,6 +1336,18 @@ export class CipherService implements CipherServiceAbstraction {
...
+    const useSdk = await firstValueFrom(this.sdkCipherAdminOpsEnabled$);
+    if (useSdk) {
+      await this.clearCache(userId);
+      const cipherView = await this.cipherSdkService.saveCollectionsWithServer(
+        cipher.id,
+        cipher.collectionIds,
+        userId,
+      );
+      const encryptResult = await this.cipherEncryptionService.encrypt(cipherView, userId);
+      return encryptResult.cipher;
```
jengstrom-bw, line 1187: test "should delete cipher locally and return undefined when SDK returns null" fails here since encryptResult is `undefined` so `.cipher` is not reachable

jengstrom-bw (changes requested):

libs/common/src/vault/services/cipher.service.ts
nikwithak: Good catch! I removed the test - I yanked the logic it was testing since the state management it was testing belongs in the SDK here.

jengstrom-bw (approved):

jengstrom-bw (approved):

PR #20361, opened 2026-04-23
[PM-28346] Use SDK for attachment delete operations

gbubemismith:
## 🎟️ Tracking
https://bitwarden.atlassian.net/browse/PM-28346

## 📸 Screenshots

https://github.com/user-attachments/assets/21ae05a2-3ae1-413f-bcc8-ed4fbfa7d1df


libs/common/src/vault/services/cipher-sdk.service.ts
```diff
@@ -343,6 +343,41 @@ export class DefaultCipherSdkService implements CipherSdkService {
...
+        switchMap(async (sdk) => {
+          using ref = sdk.take();
+
+          if (asAdmin) {
+            await ref.value
+              .vault()
+              .ciphers()
+              .admin()
+              .delete_attachment(asUuid(cipherId), attachmentId);
+            return undefined;
```
nikwithak: The Admin API endpoint returns a `CipherMiniResponseModel` - is there a reason the SDK swallows this to return `()`? I think we should return it if it's included in the response.

nikwithak (changes requested):
Looks good overall - just one observation/question

libs/common/src/vault/services/cipher-sdk.service.ts
gbubemismith: Agreed, no current TS caller consumes the admin response, which is why I left it as (), I'll update the change on the SDK and then pull that in here

nikwithak (approved):

PR #20365, opened 2026-04-24
[PM-35789] fix(desktop): repair ipcRenderer listener leak in messagingService ch…

usptact:
…annel

The removeListener implementation in apps/desktop/src/platform/preload.ts created a new anonymous function on every call instead of reusing the reference originally passed to addListener. Because ipcRenderer.removeListener requires exact reference equality to match and remove a listener, every removal attempt was silently a no-op.

As a result, every subscription to fromIpcMessaging() that was later unsubscribed permanently leaked one ipcRenderer listener bound to the "messagingService" channel. Listeners accumulated across login/logout cycles and component navigation without bound.

Fix: maintain a module-scope Map from each caller callback to its internal wrapper function. addListener stores the wrapper before registering it; removeListener looks up the stored wrapper by callback reference and passes that exact same reference to ipcRenderer.removeListener.

### DIFF
--- a/apps/desktop/src/platform/preload.ts
+++ b/apps/desktop/src/platform/preload.ts
@@ -99,6 +99,11 @@ const localhostCallbackService = {
   },
 };
 
+const _messagingListeners = new Map<
+  (message: any) => void,
+  (_event: any, message: any) => void
+>();
+
 export default {
   versions: {
     app: (): Promise<string> => ipcRenderer.invoke("appVersion"),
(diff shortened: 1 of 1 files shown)


addisonbeck:
Thanks for the contribution! We've got some duplication between this and https://github.com/bitwarden/clients/pull/20945, and an outstanding CLA on this one, so I'm going to close this PR and we'll address this on the other one.

PR #20370, opened 2026-04-24
[PM-32391] bw edit should exit with non-0 when user do not have R/W permission to the item.

### LINKED ISSUE #19034: bw edit should exit with non-0 when user do not have R/W permission to the item.
### Steps To Reproduce

 1. Retrieve the item need to be updated.
    `bwData="$(bw get item "${_BW__NOTE_NAME}")"`
 2. Update the content of `bwData`.
 3. Update the item using `bw edit`.
    `echo "${bwData}" | bw encode | bw edit item "$(echo "${bwData}" | jq -r '.id')"`

### Expected Result

If the user do not have R/W permission, the `bw edit` should return with non-0 exit status.

### Actual Result

The update is silently discarded, the `bw edit` returns with 0 exit status. The original content is printed to STDOUT.

This make it very difficult to put a guard in automation to ensure the user has R/W access. At very least, the `bw get item` should return some kind of flag to tell the permission on the item for the user, so we can programmatically test this, before doing a the actual update.

### Screenshots or Videos

_No response_

### Additional Context

Currently, we have to initiate a "dummy" update, do post-update verification to ensure the content is updated with this "dummy" update. This is cumbersome and very expensive and prolong the automation script execution time.

### Operating System

Linux

### Operating System Version

CentOS Stream 10

### Shell

Bash

### Build Version

2026.1.0

### Issue Tracking Info

gbubemismith:
## 🎟️ Tracking
https://bitwarden.atlassian.net/browse/PM-32391

## 📔 Objective
Fixes https://github.com/bitwarden/clients/issues/19034

Exit with non-zero from bw edit when the user lacks edit permission on the item, instead of silently succeeding.

## 📸 Screenshots
image


apps/cli/src/commands/edit.command.ts
```diff
@@ -103,6 +105,13 @@ export class EditCommand {
       return Response.notFound();
     }
 
+    const canEditCipher = await firstValueFrom(
+      this.cipherAuthorizationService.canEditCipher$(cipher),
+    );
```
nick-livefront, line 110: ❤️ I'm glad there's a centralized service for this!

apps/cli/src/commands/edit.command.spec.ts
nick-livefront, line 1: Thank you for the tests!

nick-livefront (approved):

shane-melton (approved):

PR #20371, opened 2026-04-24
SSH Agent v2: list keys

### LINKED ISSUE #18612: Invalid `ssh-add` flag causes ssh-agent to become unresponsive

### LINKED ISSUE #20381: Archived SSH keys continue to appear in SSH agent after sync, vault lock/unlock, and full system reboot
### Steps To Reproduce

Environment:

Bitwarden Desktop version: 2026.3.1
SDK: main (7637fdf)
Shell: 39.2.6
Renderer: 142.0.7444.226
Node: 22.1.1
Architecture: x64
OS: Windows

**Steps to reproduce:**
- Open the Bitwarden desktop app and archive one or more SSH key items
- Manually trigger a vault sync
- Lock and unlock the vault
- Run ssh-add -L -- archived keys still appear
- Fully reboot Windows, log back into Bitwarden desktop
- Run ssh-add -L again

### Expected Result

Archived SSH keys should not be offered by the SSH agent and should not appear in ssh-add -L

### Actual Result

4 out of 7 SSH keys in the vault are archived. All 4 archived keys continue to appear in ssh-add -L after a manual sync, lock/unlock cycle, and a full system reboot. Only the 3 non-archived keys should be listed.

### Screenshots or Videos

_No response_

### Additional Context
Deleting items from the vault removes them from the agent correctly. Also edits show up in the SSH agent list immediately. 
The issue is specific to the archived state, which does not appear to be respected by the SSH agent when building its key list.

_No response_

### Operating System

Windows

### Operating System Version

10 22H2

### Installation method

### Build Version

2026.3.1

### Issue Tracking Info

neuronull:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-30755

https://bitwarden.atlassian.net/browse/PM-29418

fixes: #20381

fixes: #18612

## 📔 Objective

This PR is larger than I wanted it to be. I tried to break it into two but ended up spending too much time chasing intermediary problems with doing that. So apologies in advance for the PR size. The reason it became large, is because we decided as a team to fix bug 29418 in the timeline of the v2 rewrite, and that has large implications for the implementation of 30755, so it became necessary to address both at the same time.

Note: much of the LOC additions are because I added unit test suite for the render process service, as well as coverage to the unit tests and integration tests in DN, for the list keys logic.

Objectives
1. Implement support for list keys (`ssh-add -L`) end to end.
2. Fix bug PM-29418. The bug is that the list-keys should display the keys of the most recently unlocked vault (while logged in), when locked. This was a regression from the original v1 release.

Objective 1:
- Changes to the SSH Agent server include adding support in the protocol handling for the list keys request, interfacing with our Agent's authorization API, and returning the response.
- Changes to the SSH Agent were very minimal, just decoding the keys we receive from the napi layer, and passing it forward in the call stack.
- Changes to the Electron side were more complex, as they also involved the changes for Objective 2 (more in that section below).

Objective 2:
- In the napi and SSH Agent side, the code changes are pretty cleanly subtractive. We don't need to maintain the lock/unlock state in the agent anymore. This simplifies the agent a bit.
-The Electron side sends us the keys when the vault is unlocked or when the keys need to be updated, and stops the server when there is no active account logged in that has the feature enabled.

Additional Fixes:
- Fixes feature gap support for Archived ssh key items

Design choices:
- removing the lock/unlock state management from the DN side. With the UX for vault unlocked, we don't need to store the lock/unlock state anymore. If we later change that or add other features that make it required, we can add it back. As it stands, keeping it and relying on it would only increase our surface area for errors to occur.
    - Long term, we should be using the internal SDK for the business logic around the feature enablement, account state, and vault ssh key data access.

## 📸  Testing Done

| \# | Scenario | ssh-add -L result |
| --- | --- | --- |
| 1 | Feature ON · vault unlocked | Returns all SSH key ciphers
| 2 | Feature ON · vault re-locked (was unlocked) | Returns keys (retained)
| 3 | Feature ON · vault never unlocked this session | Could not open a connection
| 4 | Feature toggled OFF (was running) | Could not open a connection
| 5 | Feature re-enabled · vault already unlocked | Returns keys
| 6 | Feature re-enabled · vault still locked | Could not open a connection
| 7 | Active account logs out, no other accounts remain | Could not open a connection
| 8 | Active account logs out, other account exists | Returns other account's keys (if that account is unlocked)
| 9 | Switch accounts: new account is unlocked | Returns new account's keys
| 10 | Switch accounts: new account is locked | Returns old account's keys (stale)
| 11 | SSH key cipher added while unlocked | Returns updated list
| 12 | SSH key cipher deleted while unlocked | Key no longer appears
| 13 | SSH key cipher name changed while unlocked | Updated name appears
| 14 | SSH key is archived | archived key no longer appears
| 15 | SSH key is un-archived | un-archived key no longer appears


apps/desktop/desktop_native/ssh_agent/src/server/connection.rs
```diff
@@ -46,25 +50,195 @@ where
...
+                result = self.connection.stream.read_exact(&mut len_buf) => {
+                    if let Err(error) = result {
+                        debug!(%error, "Connection closed");
+                        break;
+                    }
+                }
+            }
+
+            let msg_len = u32::from_be_bytes(len_buf) as usize;
+            let mut msg = vec![0u8; msg_len];
```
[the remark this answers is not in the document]
neuronull: It's good to add. But want to also note that the socket permissions are user only so the attacker would have to be in full control of the OS's user login session.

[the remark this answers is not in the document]
neuronull: https://github.com/bitwarden/clients/pull/20371/commits/23e7444b34d70baa05fc74856b4874d5507d6782

apps/desktop/src/autofill/services/ssh-agent.service.ts
```diff
@@ -285,54 +241,129 @@ export class SshAgentService implements OnDestroy {
...
+                  }),
+                  // Subscribe to live cipher data for the active account.
+                  switchMap(() => this.cipherService.cipherViews$(account.id)),
+                  // Skip emissions before cipher data is available (e.g. during initial decrypt).
+                  filter((views) => views != null),
+                  // Project to the SSH key fields needed by the agent.
+                  map((views) => this.toAgentKeys(views)),
+                  // Skip re-push when the SSH key set hasn't actually changed.
+                  distinctUntilChanged((prev, curr) => {
+                    if (prev.length !== curr.length) {
```
quexten, line 337: This condition is incorrect if the size stays the same but the content changes. Suppose you are editing an item and importing an SSH key, then the set of agent keys would have the same size, but change.

Is it that expensive to push keys to the agent that we have to diff here?

apps/desktop/src/autofill/services/ssh-agent.service.ts
```diff
@@ -285,54 +241,129 @@ export class SshAgentService implements OnDestroy {
...
+                      prev.map((k) => [k.cipherId, { privateKey: k.privateKey, name: k.name }]),
+                    );
+                    return curr.every((k) => {
+                      const p = prevMap.get(k.cipherId);
+                      return p?.privateKey === k.privateKey && p?.name === k.name;
+                    });
+                  }),
+                  // Atomically replace the agent's keystore with the current set.
+                  concatMap(async (keys) => {
+                    await ipc.autofill.sshAgent.clearKeys();
```
quexten: IMO clear + setKeys is not ideal because there is a short timeframe where the agent is empty. Ideally, setKeys would just replace the keys atomically (without crossing any await boundaries) because then we don't have any period where the agent is mistakenly empty.

quexten: Also I'm not sure why the comment says "Atomically" here? These are two awaits and in-between other javascript tasks may get scheduled, / a list or sign request may be handled.

quexten:
Looks pretty good overall, just two smaller items.

apps/desktop/src/autofill/services/ssh-agent.service.ts
neuronull: TLDR: I will improve the code comments to make it more clear what is going on.

The length check is only the first check. If the lengths are the same, the content check happens on the lines that follow. So this scenario

> Suppose you are editing an item and importing an SSH key, then the set of agent keys would have the same size, but change.

, results in the new key data being cleared and set.

> Is it that expensive to push keys to the agent that we have to diff here?

These checks in the distinctUntilChanged custom comparator are less about IPC performance cost, and more about correctness. There isn't a reason to send the keys if they haven't changed.

apps/desktop/src/autofill/services/ssh-agent.service.ts
neuronull: Agreed. I believe the clearKeys() notion was carryover from the v1 implementation.
With the changes in implementation that we have in the v2, I think it makes more sense to do what you suggested.
I'm pushing up some changes that implement that, and remove the clearKeys() from the napi API. I think it is sensible to keep that as a keystore function, but LMK what you think.

neuronull: https://github.com/bitwarden/clients/pull/20371/commits/02982e1ce8065ea05fcbd465d6b176f658f1966d

apps/desktop/src/autofill/services/ssh-agent.service.ts
neuronull: https://github.com/bitwarden/clients/pull/20371/commits/5338f20c59f897b6102426b3f340038431fb80a3

quexten (dismissed):

dani-garcia (dismissed):

apps/desktop/desktop_native/ssh_agent/src/server/protocol.rs
```diff
@@ -0,0 +1,118 @@
...
+//! <https://datatracker.ietf.org/doc/draft-ietf-sshm-ssh-agent/>
+
+use crate::crypto::PublicKey;
+
+/// SSH2_AGENTC_REQUEST_IDENTITIES
+pub(super) const REQUEST_IDENTITIES: u8 = 11;
+/// SSH2_AGENT_IDENTITIES_ANSWER
+pub(super) const IDENTITIES_ANSWER: u8 = 12;
+/// SSH_AGENT_FAILURE
+pub(super) const FAILURE: u8 = 5;
```
coltonhurst, line 13: Nit: These are used in tests in `connection.rs`, but aren't accessible with the `pub(super)`. May or may not be worth making it visible so these can still be defined here but be used from `connection.rs`.

apps/desktop/desktop_native/ssh_agent/src/server/protocol.rs
```diff
@@ -0,0 +1,118 @@
+//! SSH agent wire protocol constants and message builders.
+//!
+//! Adheres to the protocol defined in:
+//! <https://datatracker.ietf.org/doc/draft-ietf-sshm-ssh-agent/>
```
coltonhurst, line 4: Thanks for linking this! I need a "chef's kiss" emoji

apps/desktop/src/autofill/services/ssh-agent.service.ts
```diff
@@ -212,127 +207,174 @@ export class SshAgentService implements OnDestroy {
       )
       .subscribe();
```
coltonhurst, line 212: Up to you if this is relevant, and if it's worth addressing here or in another PR, as the bulk of the code leading up to the subscribe was unchanged within this PR!

Question: What happens if there are errors within these rx chains? It might be worth catching errors, depending on the behavior for the end user?

Ex:

```typescript
.subscribe({ error: (error: unknown) => { do something here }});
```

apps/desktop/src/autofill/services/ssh-agent.service.ts
```diff
@@ -212,127 +207,174 @@ export class SshAgentService implements OnDestroy {
...
+                  concatMap(async (keys) => {
+                    await ipc.autofill.sshAgent.replace(keys);
+                  }),
+                );
+              }),
+            );
+          }),
+          takeUntil(this.destroy$),
+        )
+        .subscribe();
```
coltonhurst, line 362: Question: What happens if there are errors within these rx chains? It might be worth catching errors, depending on the behavior for the end user?

Ex:

```typescript
.subscribe({ error: (error: unknown) => { do something here }});
```

apps/desktop/src/autofill/models/ipc-channels.ts
```diff
@@ -11,10 +11,10 @@ export const AUTOTYPE_IPC_CHANNELS = {
 export const SSH_AGENT_IPC_CHANNELS = {
   INIT: "sshagent.init",
   IS_LOADED: "sshagent.isloaded",
-  SET_KEYS: "sshagent.setkeys",
+  REPLACE: "sshagent.replace",
   SIGN_REQUEST_RESPONSE: "sshagent.signrequestresponse",
   LOCK: "sshagent.lock",
   CLEAR_KEYS: "sshagent.clearkeys",
```
coltonhurst: Question: Are these two only used in v1?

If so we could mark them with a comment and the ticket for when they should be removed.
If not (if they are used in both v1 and v2), please ignore this comment!

apps/desktop/src/autofill/main/main-ssh-agent.service.ts
```diff
@@ -172,39 +179,22 @@ export class MainSshAgentService {
...
-    return new Promise((resolve) => {
-      this.pendingRequests.set(id, resolve);
-      this.messagingService.send(SSH_AGENT_IPC_CHANNELS.UNLOCK_REQUEST, { requestId: id });
-    });
+    try {
+      this.agentStateV2 = await sshagent_v2.SshAgentState.serve(signCb);
+      this.logService.info("SSH agent v2 started");
+    } catch (e: unknown) {
+      this.logService.error("SSH agent v2 encountered an error: ", e);
+    }
```
coltonhurst, line 199: Nit: Really not a big deal, just an edge case - it might be worth adding a check in `initV2()` to check if the agent is running or not before the rest of the function executes.

coltonhurst (commented):
Thanks @neuronull! The PR is looking very good, just small comments.

apps/desktop/desktop_native/ssh_agent/src/server/protocol.rs
neuronull: This is actually intentional! By not using the definitions from the source code, in the unit tests, we are asserting that the source code's handling of the protocol is correct. Otherwise, if we were to use the source's definitions in the unit tests, someone (or thing) could change the values and the tests would still validate (incorrectly).

apps/desktop/src/autofill/services/ssh-agent.service.ts
neuronull: AFAICT, the place to catch the error is before the subscribe. (edit: I added these to both places you flagged)

apps/desktop/src/autofill/models/ipc-channels.ts
neuronull: correct. What I did instead is remove them from this definition, since the code that uses these already has a comment to remove with v1 + ticket number. thus the remaining definitions are the ones used in the v2 or both.

apps/desktop/src/autofill/main/main-ssh-agent.service.ts
neuronull: I see the concern. I think it isn't necessary (and IMO a step in the wrong direction) for the below reasons, and if you're in agreement, I can add a code comment if it would be helpful, but LMK what you think.

initV2() being called while the agent is already running is already prevented these ways:
 
1. In the renderer process the v2 checks isLoaded() (which intern calls is_running() to the agent) before ever calling init(), and only proceeds if it returns false.
2. If it somehow slips through that (we introduce a bug in the future), serve() would fail trying to bind to an already-held socket and that error is logged and propagated up to here, and the `agentStateV2` remains at its current running value (the assignment on line 193 never executes if the promise rejects). So the running agent is unaffected.

Moreover, in scenario 2, the failure chain from whomever called initv2() when it's already running _should_ fail because that means we have a logic bug in the electron side somewhere. If we were to add a check if it's already running and say, log the failure, that would prevent us from failing at runtime but we'd still have a logic bug, we'd just be hiding it.
(As an aside, the fact we have the render+main process shenanigans to manage is what creates this extra layer of initialization logic and the added complexity therein 😔)

apps/desktop/src/autofill/models/ipc-channels.ts
neuronull: ah, forgot that these were referenced in preload.ts which was ported from the Platform ownership. So I ended up having to add such a code comment there. Still think this approach is more clear as the v1-only channels are not using the const defines so easy to notice.

apps/desktop/src/autofill/main/main-ssh-agent.service.ts
neuronull: Another thought- this is a helper function that is only used in one place, and it's useage is just to make the code more readable. I think it's fair to have a code comment that states the pre-condition for the function. If we were hypothetically to remove the helper function and drop the code into where it's called, we'd likely also have a code comment at that place stating the same thing.

coltonhurst: Thanks for the context. With the background of points 1 and 2 you shared, I'm good with just a comment 👍

apps/desktop/src/autofill/services/ssh-agent.service.ts
coltonhurst: Changes look good, thanks!

apps/desktop/src/autofill/models/ipc-channels.ts
coltonhurst: Changes look good, thanks!

apps/desktop/src/autofill/main/main-ssh-agent.service.ts
neuronull: added in https://github.com/bitwarden/clients/pull/20371/commits/dd9bb562161700535ead761489f2b3a2cb4b1517

coltonhurst (approved):

dani-garcia (approved):

quexten (approved):

PR #20373, opened 2026-04-24
[PM-26202] Refactor vault page to subcomponents to make it more wieldy

JaredScar:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-26202

https://bitwarden.atlassian.net/browse/PM-36861

## 📔 Objective

Breaks down vault component into subcomponents to make the original component more wieldy. Original component was 1400 lines long. 
It is now 800 lines long and has 2 new components to handle collection and cipher related logic within services


apps/web/src/app/admin-console/organizations/collections/vault.component.ts
```diff
@@ -1,11 +1,12 @@
+// FIXME: Update this file to be type safe and remove this and next line
+// @ts-strict-ignore
```
BTreston: Don't do this, fix the ts-strict issues

BTreston (changes requested):
This could definitely benefit from some test coverage too, TY

apps/web/src/app/admin-console/organizations/collections/vault.component.ts
```diff
@@ -550,6 +509,25 @@ export class VaultComponent implements OnInit, OnDestroy {
...
 
   async ngOnInit() {
+    this.cipherActions.init(
+      this.organization$,
+      this.userId$,
+      this.filter$,
+      this.editableCollections$,
+      this.selectedCollection$,
+      this.routedVaultFilterBridgeService.activeFilter$,
+      () => this.refresh(),
```
BTreston: This is an odd pattern that tightly couples the service to component behavior. I think a better approach would be to:

- extract the subscription from [line 554](https://github.com/bitwarden/clients/pull/20373/changes#diff-584da0250191a90f65e8b63c1ed39f53f2757fc3d9e42433d07a750c3283fd59L568) into a service
- refactor it to address the deprecated use of `subscribe` to use an `Observable` instead
- expose the observable from the service as a public method
- subscribe to the observable in the component and then execute logic.

apps/web/src/app/admin-console/organizations/collections/services/vault-cipher-actions.service.ts
```diff
@@ -0,0 +1,604 @@
...
+    private eventCollectionService: EventCollectionService,
+    private dialogService: DialogService,
+    private toastService: ToastService,
+    private logService: LogService,
+    private accountService: AccountService,
+    private messagingService: MessagingService,
+    private platformUtilsService: PlatformUtilsService,
+    private router: Router,
+    private i18nService: I18nService,
+  ) {}
```
BTreston: Can we refactor these to use modern angular `Inject` syntax.

apps/web/src/app/admin-console/organizations/collections/services/vault-cipher-actions.service.ts
```diff
@@ -0,0 +1,604 @@
...
+  ): void {
+    this.organization$ = organization$;
+    this.userId$ = userId$;
+    this.filter$ = filter$;
+    this.editableCollections$ = editableCollections$;
+    this.selectedCollection$ = selectedCollection$;
+    this.activeFilter$ = activeFilter$;
+    this.refreshCallback = refresh;
+    this.navigateCallback = navigate;
+  }
```
BTreston: Then can we just do these in the ctor and be able to use `new`?

apps/web/src/app/admin-console/organizations/collections/vault.component.ts
```diff
@@ -550,6 +509,25 @@ export class VaultComponent implements OnInit, OnDestroy {
...
+      (queryParams, opts) => this.go(queryParams, opts),
```
BTreston: This is also unintuitive and also tightly couples component behavior with the service. We should make this more reactive. When we are doing something inside the service, emit from a public observable, then react to that emission in the component via `subscribe`

apps/web/src/app/admin-console/organizations/collections/vault.component.ts
BTreston, line 595: we're doing a lot of this `subscribe` anti-pattern where we just extract a value out of a reactive data stream like this. In your next round of changes, can we refactor all the instances where we follow this pattern to adhere to a more proper reactive pattern?

BTreston (changes requested):
This is a great start, thanks for tackling this 👏 . A few points of feedback for ya

apps/web/src/app/admin-console/organizations/collections/vault.component.ts
BTreston, line 216: We should address this. The refactor should leave us with no `TODO`s `FIXME`s nor `eslint disable`s

BTreston, line 218: We should convert all the deps in this ctor to using the modern angular `Inject` syntax as well. We can actually do that here b/c its an angular component (my previous comments about this on the service were wrong, oops)

BTreston, line 1: Looking at this some more, I think we can break this down further into another service: `vault-collection.service`. The service should be responsible for initializing and exposing observables for all of the various collections properties:
- `editableCollections$`
- `allCollections$`
- `allCollectionsWithoutUnassigned$`
- `selectedCollections$`
- `nestedCollections$`
- `showCollectionAccessRestricted$`
- move the functions from  `admin-console/utils/collection-utils into this service`
- Should follow out abstract interface pattern (`vault-collection.service.ts` for the abstract class, `default-vault-colletion.service.ts` for the concrete implementation)
If there are any observables that are not required by the component, then they can remain private.

apps/web/src/app/admin-console/organizations/collections/vault.component.ts
```diff
@@ -547,9 +518,32 @@ export class VaultComponent implements OnInit, OnDestroy {
...
+    this.cipherActions = runInInjectionContext(this._injector, () => {
+      return new VaultCipherActionsService(
+        this.organization$,
+        this.userId$,
+        this.filter$,
+        this.editableCollections$,
+        this.selectedCollection$,
+        this.routedVaultFilterBridgeService.activeFilter$,
+      );
+    });
```
BTreston: I had another look at this file as a whole. With the proposed changes to add a new service, the idea here would be to allow us to inject this like a normal service. We shouldn't actually need any params in the constructor and instead use normal angular DI for those services internally in `VaultActionsService`. We should shoot for the same pattern for the `CipherActionsService`. The only thing you should see here in the component is a regular `@Inject(VaultActionsService)` statement. That should remove our need for using manual injection context.

apps/web/src/app/admin-console/organizations/collections/vault.component.ts
```diff
@@ -225,28 +202,22 @@ export class VaultComponent implements OnInit, OnDestroy {
...
+    protected billingApiService: BillingApiServiceAbstraction,
     private organizationWarningsService: OrganizationWarningsService,
-    private collectionService: CollectionService,
     private restrictedItemTypesService: RestrictedItemTypesService,
+    private dialogService: DialogService,
+    private toastService: ToastService,
+    private collectionActions: VaultCollectionActionsService,
   ) {
     this.userId$ = this.accountService.activeAccount$.pipe(getUserId);
     this.filter$ = this.routedVaultFilterService.filter$;
```
BTreston, line 252: we should just use the property from the service directly rather than making it a class member here.

apps/web/src/app/admin-console/organizations/collections/vault.component.ts
BTreston: We should attempt something similar with ciphers too, but that might be diminishing returns here. Try it out and see how it looks, if it introduces too much complexity then we can leave it as is.

BTreston (changes requested):
Something we should be mindful of is that we are aiming for the least amount of logic as possible at the component level, it should really only be an orchestrator of reactive data into the template. Try to keep it as close to a presentational component as possible.

apps/web/src/app/admin-console/organizations/collections/vault.component.ts
BTreston: for further context, in the `CipherActionsService` we'd be initializing things like `organization$` and `userId$` by DI'ing the `OrganizationService` and `AccountService` and calling their functions in the service, similarly to how they're initialized in the component.

apps/web/src/app/admin-console/organizations/collections/services/vault-collection-actions.service.ts
```diff
@@ -0,0 +1,210 @@
...
+    selectedCollection$: Observable<TreeNode<CollectionAdminView> | undefined>,
+    editableCollections$: Observable<CollectionAdminView[]>,
+    refresh: () => void,
+  ): void {
+    this.organization$ = organization$;
+    this.userId$ = userId$;
+    this.selectedCollection$ = selectedCollection$;
+    this.editableCollections$ = editableCollections$;
+    this.refreshCallback = refresh;
+  }
```
BTreston, line 91: we should give this service the same treatment as the `default-vault-collection.servcie`. No `init()` function in this component, no dependency on an externally provided refresh subject.

BTreston (changes requested):
This is looking much better, thank you. I think this is probably the last big change we need to make here

apps/web/src/app/admin-console/organizations/collections/vault.component.ts
```diff
@@ -27,119 +35,70 @@ import {
...
-// FIXME: update to use a const object instead of a typescript enum
-// eslint-disable-next-line @bitwarden/platform/no-enums
-enum AddAccessStatusType {
-  All = 0,
-  AddAccess = 1,
-}
-
 // FIXME(https://bitwarden.atlassian.net/browse/CL-764): Migrate to OnPush
 // eslint-disable-next-line @angular-eslint/prefer-on-push-component-change-detection
 @Component({
```
BTreston, line 145: are we able to migrate to onpush?

apps/web/src/app/admin-console/organizations/collections/vault.component.ts
```diff
@@ -160,9 +119,38 @@ enum AddAccessStatusType {
...
+    VaultCollectionActionsService,
+    { provide: VaultCollectionService, useClass: DefaultVaultCollectionService },
+    VaultCipherActionsService,
   ],
 })
 export class VaultComponent implements OnInit, OnDestroy {
+  private route = inject(ActivatedRoute);
+  private organizationService = inject(OrganizationService);
+  protected vaultFilterService = inject(VaultFilterService);
+  private routedVaultFilterBridgeService = inject(RoutedVaultFilterBridgeService);
```
BTreston: unused

apps/web/src/app/admin-console/organizations/collections/services/vault-collection-actions.service.ts
BTreston: This is not addressed

BTreston (changes requested):

apps/web/src/app/admin-console/organizations/collections/vault.component.ts
```diff
@@ -583,16 +400,12 @@ export class VaultComponent implements OnInit, OnDestroy {
       });
     });
```
BTreston, line 584: I'd like to address this if possible, but I'm not sure of the best way. There are a few patterns that we use, it seems like it may be better to do something like:
```typescript
this.messageListener.allMessages$
      .pipe(
        filter((msg) => msg.command === "syncCompleted" && !!msg.successfully),
        takeUntil(this.destroy$),
      )
      .subscribe(() => this.refresh());
```
but I don't think we really want to rely on messaging like this in general (especially at the component level). This proposed change might fit better in the service layer and the component should only care about what flows out of the service's exposed observables.

cc @eliykat for a second oppinion

apps/web/src/app/admin-console/organizations/collections/vault.component.ts
BTreston, line 709: Can we also try and figure this out? We may need to sync with the vault team on this, I'm okay with not addressing it on this PR if the scope ends up being too big

BTreston, line 256: we should also investigate what it would take to fix this. Again, I'm okay with this being a follow up to try and contain the scope here.

BTreston (commented):
A few general questions, feel free to bring these up for discussion with the broader team during the sync next week or in slack

apps/web/src/app/admin-console/organizations/collections/vault.component.ts
JaredScar: I think this makes the scope too big imo -- May be best for a follow-up ticket

apps/web/src/app/admin-console/organizations/collections/vault.component.ts
```diff
@@ -160,108 +116,81 @@ enum AddAccessStatusType {
...
-      this.filter$.pipe(
-        map((filter) => filter.organizationId),
-        filter((filter) => filter !== undefined),
-        filter(
-          (value: OrganizationId | Unassigned): value is OrganizationId => value !== Unassigned,
-        ),
-        distinctUntilChanged(),
-      );
+  readonly activeFilter!: Signal<VaultFilter>;
+  protected readonly showAddAccessToggle!: Signal<boolean>;
```
jaasen-livefront: ❓ Any particular reason we're non-null asserting here?

apps/web/src/app/admin-console/organizations/collections/vault.component.ts
```diff
@@ -160,108 +116,81 @@ enum AddAccessStatusType {
...
+
+  private readonly searchText$ = new Subject<string>();
+  private readonly refresh$ = new Subject<void>();
+  protected readonly isRefreshing$ = new BehaviorSubject<boolean>(true);
+  private readonly destroy$ = new Subject<void>();
+
+  protected readonly allCollections$!: Observable<CollectionAdminView[]>;
+  protected readonly collections$!: Observable<CollectionAdminView[]>;
+  protected readonly selectedCollection$!: Observable<TreeNode<CollectionAdminView> | undefined>;
+  protected readonly showCollectionAccessRestricted$!: Observable<boolean>;
```
jaasen-livefront: ❓ Any particular reason we're non-null asserting here?

jaasen-livefront (commented):
Looks great! Just left a couple small comments.

jaasen-livefront (dismissed):

apps/web/src/app/admin-console/organizations/collections/models/ac-routed-vault-filter.model.ts
```diff
@@ -0,0 +1,26 @@
...
+  collectionId?: CollectionId | All | Unassigned;
+  type?: RoutedVaultFilterItemType;
+}
+
+/**
+ * Maps a generic `RoutedVaultFilterModel` to an `ACRoutedVaultFilterModel`, or returns
+ * `null` if the filter does not represent a valid Admin Console state (i.e. the
+ * organization ID is absent or set to `Unassigned`).
+ */
+export function toACFilter(f: RoutedVaultFilterModel): ACRoutedVaultFilterModel | null {
```
BTreston: ⛏️ prefer `undefined` over `null`

apps/web/src/app/admin-console/organizations/collections/vault.component.ts
```diff
@@ -160,108 +116,81 @@ enum AddAccessStatusType {
...
+  protected readonly noItemIcon = NoResults;
+  protected readonly loading$: Observable<boolean>;
+  protected readonly processingEvent$ = new BehaviorSubject<boolean>(false);
+  protected readonly organization$: Observable<Organization>;
+  protected readonly allGroups$: Observable<GroupView[]>;
+  protected readonly ciphers$: Observable<CipherView[]>;
+  protected readonly allCiphers$: Observable<CipherView[]>;
+
+  protected readonly isEmpty$: Observable<boolean>;
+  private readonly prevCipherId = signal<string | null>(null);
```
BTreston: ⛏️ prefer `undefined` to `null`

BTreston (changes requested):
This looks much better, thank you. Can we get a feature flag made for this? This ended up being quite a big change that touches and refactors a lot of logic. We took a similar approach with the members component refactor.

jaasen-livefront (approved):

BTreston (approved):
Looks good 👏. Thank you!

PR #20377, opened 2026-04-25
[PM-34157] Wire up SDK to NewPolicyService

eliykat:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-34157
## 📔 Objective

Wire up `NewPolicyService` to its data sources (accepted state policies and organizations) and hand them off to the SDK for policy enforcement decisions.

The existing `PolicyService` shims out to the new code when the feature flag is enabled.

**Billing and Vault:** the only changes to your code is to update `PolicyType` enum names to match the SDK. No functional changes.


libs/common/src/admin-console/models/domain/policy.ts
```diff
@@ -42,4 +45,24 @@ export class Policy extends Domain {
...
+  }
+
+  toSdkPolicyView(): SdkPolicyView {
+    return {
+      id: asUuid(this.id),
+      organization_id: asUuid(this.organizationId),
+      type: this.type as number,
+      data: this.data == null ? undefined : JSON.stringify(this.data),
+      enabled: this.enabled,
+    };
```
eliykat, line 68: Missing the revision date property.

JimmyVo16:
Reviewing now

JimmyVo16 (dismissed):

eliykat:
I'm still tweaking the SDK interface in response to mobile feedback; holding this PR until that's finished, otherwise it'll just cause breaking changes in this repo.

eliykat:
Waiting to bump sdk-internal once IT issues are resolved.

eliykat:
@JimmyVo16 Main changes since your last approval:
- use a more tightly scoped `OrganizationUserPolicyContext` model to pass org information into the function, instead of the whole organization model (a better interface, also accommodates mobile data models)
- replace our TS PolicyType enum definition with the definition that is generated by the SDK - to avoid maintaining 2 duplicate definitions (and then every time a policy is added in clients or the SDK, it would be a breaking change). Update a couple of PolicyType names to align.
- minor clean up, such as removing some `snake_case` that made it through to TS.

JimmyVo16 (approved):

amorask-bitwarden (approved):

nikwithak (approved):

PR #20382, opened 2026-04-26
[PM-35843] Fix desktop vault search tab focus

Galygious:
## Summary
- Format and clarify the SearchComponent spec constructor call setup in `apps/desktop/src/app/layout/search/search.component.spec.ts` for readability and consistency.

## Why
- This change is tied to issue [#18842](https://github.com/bitwarden/clients/issues/18842), which was closed without the underlying correction being implemented in the active code path.
- This PR applies the missing test-path fix and documents the gap for visibility.
- Context: https://github.com/bitwarden/clients/issues/18842#issuecomment-4206610691

## Details
- Updated the spec setup constructor invocation to a multi-line form to align with the expected call shape and improve maintainability.
- No production behavior changes; test intent remains unchanged.

### DIFF
--- a/apps/desktop/src/app/layout/search/search-bar.service.spec.ts
+++ b/apps/desktop/src/app/layout/search/search-bar.service.spec.ts
@@ -1,0 +1,29 @@
+import { SearchBarService } from "./search-bar.service";
+
+describe("SearchBarService", () => {
+  let service: SearchBarService;
+
+  beforeEach(() => {
+    service = new SearchBarService();
+  });
+
+  it("returns false when no search result focus target is registered", () => {
+    expect(service.focusSearchResults()).toBe(false);
+  });
+
+  it("delegates focus to the registered search result focus target", () => {
+    const focusTarget = jest.fn().mockReturnValue(true);
+
+    service.setSearchResultFocusTarget(focusTarget);
+
+    expect(service.focusSearchResults()).toBe(true);
+    expect(focusTarget).toHaveBeenCalledTimes(1);
+  });
+
+  it("clears the registered search result focus target", () => {
+    service.setSearchResultFocusTarget(jest.fn().mockReturnValue(true));
+    service.setSearchResultFocusTarget(null);
+
+    expect(service.focusSearchResults()).toBe(false);
+  });
+});
--- a/apps/desktop/src/app/layout/search/search-bar.service.ts
+++ b/apps/desktop/src/app/layout/search/search-bar.service.ts
@@ -6,10 +6,14 @@ export type SearchBarState = {
   placeholderText: string;
 };
 
+type SearchResultFocusTarget = () => boolean;
+
 @Injectable()
 export class SearchBarService {
   private searchTextSubject = new BehaviorSubject<string | null>(null);
   searchText$ = this.searchTextSubject.asObservable();
+
+  private searchResultFocusTarget: SearchResultFocusTarget | null = null;
 
   private _state = {
     enabled: false,
@@ -31,6 +35,14 @@ export class SearchBarService {
 
   setSearchText(value: string) {
     this.searchTextSubject.next(value);
+  }
+
+  setSearchResultFocusTarget(target: SearchResultFocusTarget | null) {
+    this.searchResultFocusTarget = target;
+  }
+
+  focusSearchResults() {
+    return this.searchResultFocusTarget?.() ?? false;
   }
 
   private updateState() {
--- a/apps/desktop/src/app/layout/search/search.component.html
+++ b/apps/desktop/src/app/layout/search/search.component.html
@@ -6,6 +6,7 @@
     autocomplete="off"
     [formControl]="searchText"
     appAutofocus
+    (keydown.tab)="focusSearchResults($event)"
   />
   <bit-icon name="bwi-search"></bit-icon>
 </div>
--- a/apps/desktop/src/app/layout/search/search.component.spec.ts
+++ b/apps/desktop/src/app/layout/search/search.component.spec.ts
@@ -20,9 +20,12 @@ describe("SearchComponent", () => {
       focusSearchResults: jest.fn(),
     };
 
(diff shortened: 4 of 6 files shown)


ravindUwU:
I didn't run this locally, but seeing the [`(keydown.tab)="focusSearchResults($event)"`](https://github.com/bitwarden/clients/pull/20382/changes#diff-fd00a40f4e9e9bde5dcba56070cb0322d4a3c4820cc94f4791c29aa567356076), am I correct in understanding that this changes the behaviour of a <kbd>Tab</kbd> key press from within the search box, to focus the first search result 🤔? If so,

- **With this change made, how does one focus (i.e., <kbd>Tab</kbd> into) the account switcher with the keyboard?** The account switcher is positioned adjacent to the search box so one would _expect_ to be able to press <kbd>Tab</kbd> once to switch to it (the behaviour as of v2025.12.1).

- **What are your thoughts on making this a `(keydown.enter)` instead?**. I've repeatedly suggested this in #18842, but I think the suggestion might have been misread/misunderstood in [this comment](https://github.com/bitwarden/clients/issues/18842#issuecomment-4184475576).

  Using <kbd>Enter</kbd> (which doesn't currently do anything!) to focus the first search result would allow <kbd>Tab</kbd> to function as it usually would (following the visual/DOM order more or less), without any contextual "hacks" such as `(keydown.tab)="focusSomethingElse()"`.

Galygious:
> I didn't run this locally, but seeing the [`(keydown.tab)="focusSearchResults($event)"`](https://github.com/bitwarden/clients/pull/20382/changes#diff-fd00a40f4e9e9bde5dcba56070cb0322d4a3c4820cc94f4791c29aa567356076), am I correct in understanding that this changes the behaviour of a Tab key press from within the search box, to focus the first search result 🤔? If so,
> 
> * **With this change made, how does one focus (i.e., Tab into) the account switcher with the keyboard?** The account switcher is positioned adjacent to the search box so one would _expect_ to be able to press Tab once to switch to it (the behaviour as of v2025.12.1).
> * **What are your thoughts on making this a `(keydown.enter)` instead?**. I've repeatedly suggested this in [Keyboard driven navigation with tab after search goes to sidebar menu instead of vault item list first #18842](https://github.com/bitwarden/clients/issues/18842), but I think the suggestion might have been misread/misunderstood in [this comment](https://github.com/bitwarden/clients/issues/18842#issuecomment-4184475576).
>   Using Enter (which doesn't currently do anything!) to focus the first search result would allow Tab to function as it usually would (following the visual/DOM order more or less), without any contextual "hacks" such as `(keydown.tab)="focusSomethingElse()"`.

I disagree with the assumption that Tab should move from the search box to the account switcher just because it is visually adjacent.

In this context, the user has just searched. The most useful and expected next keyboard target is the search results, not the account switcher. From a practical usage standpoint, moving from search to results is almost certainly far more common than switching accounts immediately after typing in search.

Enter can technically be used to focus the first result, but I do not think it is the right fix here. The issue is not that users need a new shortcut. The issue is that the long-standing keyboard flow changed in v2025.12.1, and that disrupted established muscle memory for keyboard-driven users.

The tab order should prioritize the normal day-to-day workflow:

search input → search results → vault/sidebar areas → secondary controls like account switcher

Or more generally:

left column → middle/content column → right/secondary column → account selector → back around

That preserves useful keyboard navigation without making the account switcher the next stop simply because it is adjacent in the UI. The account switcher is important, but it should not outrank search results in the tab flow immediately after a search.


EDIT
PS: I feel like my earlier message may have come off a bit terse, so I wanted to clarify.

I do not think the Enter solution is poor. In a scenario where tabbing to search results became a problem, Enter would probably be the ideal solution.

My issue is that tabbing to search results was not the problem. The behavior seems to have been changed as a solution looking for a problem. For a long time, keyboard-only users could do:

`Ctrl + F` → type search → `Tab` through results

That was established behavior, and this change seems to have landed out of nowhere. I genuinely cannot think of a good reason to change behavior that has been standard for so long.

The solution I proposed was meant to preserve existing, habitual user behavior rather than changing it for the sake of change.

If your point was instead that Tab should move to the account switcher first, then I disagree there too. I do not think tab order should be prioritized purely based on visual proximity over utility and day-to-day usage.

The typical Bitwarden flow is:

1. Sign in, if needed.
2. Search for a password.
3. Select the result.
4. Press `Ctrl + P` to copy it.
5. Alt-tab away and forget about Bitwarden.

I do think the account menu should remain accessible via Tab for accessibility and keyboard-only use. I just do not think it should be prioritized over what seems to be the most common workflow.

ravindUwU:
Thanks for your explanation, @Galygious.

> I do not think tab order should be prioritized purely based on visual proximity over utility and day-to-day usage.

Ehhhh, this (and the order proposed above) I still disagree with, but that's okay 😅. Either way, I'm looking forward to #18842 being _actually_ resolved. Hoping the Bitwarden team reviews your PR soon 😊. Cheers!

Piste:
This is related to https://community.bitwarden.com/t/improve-search-in-desktop-app/96094/2
I am looking forward to this being merged and https://github.com/bitwarden/clients/issues/18842 being closed for good.

Galygious:
> Thanks for your explanation, @Galygious.
> 
> > I do not think tab order should be prioritized purely based on visual proximity over utility and day-to-day usage.
> 
> Ehhhh, this (and the order proposed above) I still disagree with, but that's okay 😅. Either way, I'm looking forward to #18842 being _actually_ resolved. Hoping the Bitwarden team reviews your PR soon 😊. Cheers!

one month later....

vleague2:
Hi @Galygious, I apologize for the delayed response to your PR! Thank you for taking the time to make a contribution. 

We have been working through some large-scale changes to the desktop UI, and the files changed in this PR are no longer used in the desktop UI. With the changes we've made, the tab order follows the logical element on order on the page and tabbing after search now takes you to the vault table where the list of search results is presented, which aligns with the flow you have described above. 

Since the behavior is fixed in the current version of the desktop app, I will close this out.

PR #20392, opened 2026-04-27
[PM-32211] fix private key before key rotation

mzieniukbw:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-32211

## 📔 Objective

Moved asymmetric key regeneration (public, private key) into SDK (see https://github.com/bitwarden/sdk-internal/pull/994).

CI expected to fail until SDK PR merged.


apps/web/src/app/key-management/data-recovery/steps/private-key-step.ts
```diff
@@ -24,37 +20,15 @@ export class PrivateKeyStep implements RecoveryStep {
...
-      const publicKey = await this.cryptoFunctionService.rsaExtractPublicKey(privateKey);
-      logger.record("Public key length: " + publicKey.length);
-    } catch {
-      logger.record("Public key could not be derived; private key is corrupt");
-      workingData.isPrivateKeyCorrupt = true;
       return false;
     }
 
-    return true;
+    return !workingData.isPrivateKeyCorrupt;
```
quexten, line 31: Note (non-blocking): We're removing some logging here, which is unfortunate. The logger does not record tracing messages from the SDK, and we currently cannot hook the SDK to pull out the tracing logs.

libs/key-management/src/user-asymmetric-key-regeneration/services/default-user-asymmetric-key-regeneration.service.ts
```diff
@@ -48,175 +39,65 @@ export class DefaultUserAsymmetricKeysRegenerationService implements UserAsymmet
...
         }),
       ),
     );
 
-    try {
-      await this.userAsymmetricKeysRegenerationApiService.regenerateUserAsymmetricKeys(
-        makeKeyPairResponse.userPublicKey,
-        new EncString(makeKeyPairResponse.userKeyEncryptedPrivateKey),
+    if (newAccountCryptoState != null) {
+      await this.accountCryptographyStateService.setAccountCryptographicState(
```
quexten: Note: After the https://github.com/bitwarden/sdk-internal/pull/996 merges, we could expand it to allow setting the account cryptographic state. Then, we could remove this TS bit too.

quexten (dismissed):
Nice! I like how much we are pushing to the SDK. I've noted that we can probably move the last TS bits soon if needed.

apps/web/src/app/key-management/data-recovery/steps/private-key-step.ts
mzieniukbw: I think that's fine, since almost all of the SDK logs that was added in SDK PR are or info or warn / error, hence they would be logged. In fact, i just added instrumentation to the SDK, so it should be easier to match together. I can try to bring back some of that logging here, but it won't be much.

libs/key-management/src/user-asymmetric-key-regeneration/services/default-user-asymmetric-key-regeneration.service.ts
mzieniukbw: Fixed in https://github.com/bitwarden/clients/pull/20392/commits/77571ec8674688fcdec9811b4cb0f25c5d2b298c

quexten (dismissed):

mzieniukbw:
@quexten resolved conflict, ready to review again and merge

quexten (approved):

PR #20394, opened 2026-04-27
[PM-31059] Implement state bridge support

quexten:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-31059

## 📔 Objective

Add support for the state bridge in preparation for SDK managed PIN unlock.


libs/common/src/platform/models/domain/symmetric-crypto-key.ts
dani-garcia, line 1: I'm surprised this file is in platform's ownership, seems like it would belong more under KM

dani-garcia (dismissed):
Platform changes lgtm!

libs/common/src/platform/models/domain/symmetric-crypto-key.ts
quexten: Yeah. Left over relic from the past that we never got around to moving due to the amount of importers. I wonder whether it's worth moving, or just push to move all consumers to use the FFI-exposed SDK type + some interop functions.

KM wants to delete most `ts` code where possible.

I'll look into making a clean-up ticket, be that moving it or replacing it.

dani-garcia: Yeah that's fair, if moving it over is too complicated we can leave it as-is until we move to use the SDK types, I think. It's not like this file gets updated often.

libs/common/src/key-management/state-bridge.ts
```diff
@@ -0,0 +1,202 @@
...
+  async get_persistent_pin_envelope(): Promise<PasswordProtectedKeyEnvelope | null> {
+    try {
+      return await readAtomic(
+        this.stateProvider,
+        this.userId,
+        PIN_PROTECTED_USER_KEY_ENVELOPE_PERSISTENT,
+      );
+    } catch {
+      return null;
+    }
```
Thomas-Avery: Why the try catch here and no where else?

Thomas-Avery (commented):
Looking good, one open question from me.

libs/common/src/key-management/state-bridge.ts
```diff
@@ -0,0 +1,202 @@
...
+  }
+
+  async clear_user_key(): Promise<void> {
+    await deleteAtomic(this.stateProvider, this.userId, USER_KEY);
+  }
+
+  async set_ephemeral_pin_envelope(pinEnvelope: PasswordProtectedKeyEnvelope): Promise<void> {
+    await writeAtomic(this.stateProvider, this.userId, PIN_PROTECTED_USER_KEY_ENVELOPE_EPHEMERAL, {
+      "": { pin_envelope: pinEnvelope },
```
Thomas-Avery, line 145: for these `""` it would probably be good to export the constants like `EPHEMERAL_PIN_ENVELOPE_KEY` and keep things in sync.

libs/common/src/key-management/state-bridge.ts
quexten: I don't recall why I added it, must have debugged something. Removed now.

libs/common/src/key-management/state-bridge.ts
quexten: I would agree, but I will not do that here because the intent is to remove the repository mappings again now that we can access them via KM state instead of going via the platform state route. I'll make a separate tech debt ticket for migrating back off of the repository pattern. 

Then we don't need `""` at all.

Thomas-Avery (approved):

dani-garcia (approved):

PR #20395, opened 2026-04-27
[PM-31059] Implement SDK managed PIN unlock

quexten:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-31059

Please see the SDK changes to learn about the API we are wrapping and consuming here: https://github.com/bitwarden/sdk-internal/pull/997

Blocked by: https://github.com/bitwarden/clients/pull/20394
(Replacing PIN unlock without the state bridge present at runtime would result in runtime errors / the feature becoming unusable).

## 📔 Objective

Switch to SDK-based PIN unlock, and delete the TS implementation. The SDK is responsible for managing state and unlocking.

This switches out the term `Ephemeral` for `AfterFirstUnlock` (AFU) and `Persistent` for `BeforeFirstUnlock`, to more clearly describe these pin lock modes. These are the naming that the SDK implementation prescribes.

Please note: This entirely replaces and wipes the TS implementation! It will be QA'd before merge.


apps/desktop/src/key-management/lock/services/desktop-lock-component.service.ts
```diff
@@ -55,7 +55,12 @@ export class DesktopLockComponentService implements LockComponentService {
       // Note: defer is preferable b/c it delays the execution of the function until the observable is subscribed to
       defer(() => this.biometricsService.getBiometricsStatusForUser(userId)),
       this.userDecryptionOptionsService.userDecryptionOptionsById$(userId),
-      defer(() => this.pinService.isPinDecryptionAvailable(userId)),
+      defer(() =>
+        timer(0, 1000).pipe(
```
quexten: We don't yet have a solution for observability across SDK bounds. For now we are polling here, and we should not block this work by coming up with a general solution, but we should flag this to platform.

quexten:
Please note, this is currently broken at runtime until SDK is updated with https://github.com/bitwarden/sdk-internal/pull/1081

dani-garcia (dismissed):
Platform changes lgtm

libs/key-management-ui/src/lock/components/lock.component.ts
```diff
@@ -501,63 +501,29 @@ export class LockComponent implements OnInit, OnDestroy {
 
     const MAX_INVALID_PIN_ENTRY_ATTEMPTS = 5;
 
-    if (await this.configService.getFeatureFlag(FeatureFlag.UnlockViaSDK)) {
```
Thomas-Avery: ❓ The ticket for this describes putting this change behind a feature flag `pm-31059-sdk-pin-unlock`. It looks like we are removing a feature flag here instead?

libs/unlock/src/default-unlock.service.ts
```diff
@@ -175,30 +122,47 @@ export class DefaultUnlockService implements UnlockService {
...
-              keyConnectorUrl: {
-                url: keyConnectorUnlockData.url,
-                key_connector_key_wrapped_user_key:
-                  keyConnectorUnlockData.keyConnectorKeyWrappedUserKey,
-              },
-            },
+            method,
           });
+
+          await this.runOnUnlockSideEffects(userId, ref);
```
Thomas-Avery, line 179: `unlockWithKeyConnector` did not run `await this.runOnUnlockSideEffects(userId, ref);` previously. Refactoring all unlock methods here is increasing risk and seems out of scope.

libs/unlock/src/unlock.service.ts
```diff
@@ -45,4 +46,15 @@ export abstract class UnlockService {
...
+  /**
+   * Unlocks the user's account with a decrypted user key
+   * Note: Where possible use other unlock methods.
+   *
+   * @param userId - The user's id
+   * @param userKey - The decrypted user key to unlock with
+   * @throws If the SDK is not available
+   * @throws If decryption fails or the key is invalid
+   */
+  abstract unlockWithDecryptedKey(userId: UserId, userKey: SymmetricKey): Promise<void>;
```
Thomas-Avery:
```suggestion
  abstract unlockWithDecryptedKey(userId: UserId, userKey: UserKey): Promise<void>;
```

I would think getting a `UserKey` to whatever the SDK needs would be the `UnlockService`'s concern.

libs/common/src/key-management/utils.ts
```diff
@@ -10,3 +16,44 @@ export async function firstValueFromOrThrow<T>(
...
+          throw new Error("SDK client not available");
+        }),
+      ),
+    );
+  } catch (error) {
+    // If the error is not "SDK client not available", we re-throw it. Otherwise, try on the register client,
+    // since the vault appears to be locked.
+    if (!(error instanceof Error && error.message === "SDK client not available")) {
+      throw error;
+    }
```
Thomas-Avery: The `throw new Error("SDK client not available");` above makes this unreachable. This doesn't make sense to me.

libs/common/src/key-management/vault-timeout/services/vault-timeout-settings.service.ts
```diff
@@ -101,7 +103,15 @@ export class VaultTimeoutSettingsService implements VaultTimeoutSettingsServiceA
...
-          this.pinStateService.pinSet$(userId),
+          combineLatest([
+            this.stateProvider.getUserState$(PIN_PROTECTED_USER_KEY_ENVELOPE_EPHEMERAL, userId),
+            this.stateProvider.getUserState$(PIN_PROTECTED_USER_KEY_ENVELOPE_PERSISTENT, userId),
+          ]).pipe(
+            map(
+              ([ephemeralEnvelope, persistentEnvelope]) =>
+                !!ephemeralEnvelope || !!persistentEnvelope,
+            ),
+          ),
```
Thomas-Avery, line 114: ❓ This feels like a good reason to keep the `pinStateService` around instead of calling stateProvider keys directly.

libs/common/src/key-management/pin/pin.service.implementation.ts
```diff
@@ -1,154 +1,92 @@
...
-    const userKey: UserKey = await firstValueFromOrThrow(
-      this.keyService.userKey$(userId),
-      "userKey",
-    );
-    const userKeyEncryptedPin = await firstValueFromOrThrow(
-      this.pinStateService.userKeyEncryptedPin$(userId),
-      "userKeyEncryptedPin",
-    );
-    return this.encryptService.decryptString(userKeyEncryptedPin, userKey);
+    return this.withPinSettingsClient(userId, (client) => {
```
Thomas-Avery:
```suggestion
    return await this.withPinSettingsClient(userId, (client) => {
```

I would expect these all to be awaited.

libs/common/src/key-management/utils.ts
```diff
@@ -10,3 +16,44 @@ export async function firstValueFromOrThrow<T>(
...
+      sdkService.userClient$(userId).pipe(
+        map((sdk) => {
+          using ref = sdk.take();
+          return fn(ref.value);
+        }),
+        catchError(() => {
+          throw new Error("SDK client not available");
+        }),
+      ),
+    );
```
Thomas-Avery: I have some concerns with this utility. Specifically it looks to break the `firstValueFrom` warning described here.

https://github.com/bitwarden/clients/blob/c326b0f762af47249ce7cf1b4384b14687a3198f/libs/common/src/platform/abstractions/sdk/sdk.service.ts#L119-L121


I'm also concerned about the `using ref` going out of scope here.  These MDN docs make me think we are using this after it's been freed. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/using#using_in_a_function

Thomas-Avery (changes requested):

libs/common/src/key-management/utils.ts
quexten: What do you mean? The scope does not end until after `fn` has run.
Written differently (but equivalently), it would be:
```
{ // Scope 1: []
  using ref = sdk.take(); // Scope 1: [ref]
  var x =  fn(ref.value); // Opens Scope 2 on the scope stack
           | do stuff...  // Scope 1: [ref], Scope 2: [ref.value]
           | return x;    // Scope 1: [ref], Scope 2: [ref.value, x]; 
  return x;               // Scope 1: [ref, x], drop ref
}
```

So the lifetime of ref covers the function run of `fn` where it is used.
Specifically, the article you link also explains this:

> the resource is disposed when the function finishes executing, immediately before the function returns.

For `return fn(ref.value);`
The return statement does not run until after `fn(ref.value)` has finished executing.

Could you explain where the concern is / where you think it is dropped, and subsequently used?

quexten: Also, it does not violate the warning. All operations on ref are done within the observable, within a function call *in* the observable. We are not returning ref.value out of the observable, which *would* break it.

libs/common/src/key-management/pin/pin.service.implementation.ts
quexten: Why? There is no practical difference for us between these?

libs/common/src/key-management/utils.ts
quexten: The `throw new Error("SDK client not available");` happens inside a try/catch block. The code you quote is on the catch path. 

In the case that the vault is unlocked, the try path is the happy path. In the case the vault is locked, the creation of a regular sdk client fails, and throws the above message, and gets caught. The catch block checks for the specific error message, and tries again on the locked register client, if and only if the failure was not a failure in the actual SDK call, but a failure of creating the SDK client.

Could you explain the codepath which is unreachable? I'm not following.

libs/common/src/key-management/vault-timeout/services/vault-timeout-settings.service.ts
quexten: Both solutions are just bad, because they break the abstraction we are creating for the SDK. Once we move to SDK managed state, we cannot use state providers anyways, so I don't really want to build anything that relies on state providers...

This would be much better served by temporarily adding a polling layer around isPinUnlockAvailable, that emits an observable stream, and in the long term having platform provide a solution for observability.

At any rate, we would be maintaining three files:
- service
- spec
- abstraction

just for replacing these 9 lines, which does not seem worth it.

libs/unlock/src/unlock.service.ts
quexten: Hmm, we should find a solution that is consistent across SDK, clients, mobile here. If we want newtypes around keys for specific purposes, then we should do so in the SDK too, if we do not want it in the SDK, then I don't see a reason to keep it in clients.

I guess we can use the TS type `symmetriccryptokey` here until we get around to replacing it.

libs/unlock/src/default-unlock.service.ts
quexten: > Refactoring all unlock methods here is increasing risk and seems out of scope.

What specifically seems risky here? The change is worth it either way, so my alternative path would be to pull out the refactor into a separate PR and submit it separately, but I do think we should do it because it is getting unmaintainable.

libs/key-management-ui/src/lock/components/lock.component.ts
quexten: There are two separate questions packed into your comment:

1. Why are we removing the `UnlockViaSDK` flagged code here?

Pin service can no longer unlock. This functionality is removed and now only lives in unlock service. Therefore, the flagged code must be promoted out of the flag here. Unlocking via the SDK will be covered by QA.

2. Why are we not flagging the change?

I thought about this a fair bit, but in my opinion, in this case flagging the change would increase risk, and massively increase scope. Specifically due to the changes to the public interface of the PIN service, it is not easily flaggable. I specifically noted that we are QA'ing on the feature branch as a result, which I feel is the pragmatic and reasonable approach here.

quexten: Reading the ticket more closely, it was not saying to flag the entire change, but to flag the unlock method, which I'm ok with. However, I will note that we do still need to remove the `UnlockViaSDK` call here because of the reasoning above.

libs/common/src/key-management/utils.ts
quexten: Discussed 1:1 will update

libs/common/src/key-management/pin/pin.service.implementation.ts
quexten: Discussed 1:1 and will clarify

libs/common/src/key-management/utils.ts
quexten: @Thomas-Avery I replaced this solution with just making the sdk service properly emit a locked client in the first place. That solution feels more robust.

libs/common/src/platform/services/sdk/default-sdk.service.ts
```diff
@@ -204,8 +200,14 @@ export class DefaultSdkService implements SdkService {
               new JsTokenProvider(this.apiService, userId),
               settings,
             );
+            await this.initializeClient(userId, client);
```
quexten, line 209: @dani-garcia Just FYI, we changed this here because our helper that switched between register sdk client and regular sdk client depending on lock state was pretty brittle. I don't see an issue with returning locked instances and as far as I understand it's the intended state anyways. LMK if this is not correct.

rr-bw (dismissed):

libs/key-management-ui/src/lock/components/lock.component.ts
Thomas-Avery, line 1: ⛏️ Looks like `pinService` is no longer needed here.

Thomas-Avery (commented):
Looks good, it does look like there is a merge conflict that needs fixed.

dani-garcia (dismissed):

Thomas-Avery (dismissed):

rr-bw:
@quexten Auth changes still fine but looks like there's a handful of failing checks on the PR.

quexten:
@rr-bw looks like it was a flaky CI pipeline run not a real failure.

rr-bw (dismissed):

dani-garcia (approved):

JaredSnider-Bitwarden (approved):

Thomas-Avery (approved):

PR #20211, opened 2026-04-16
[SM-1762] Bump Jest to 30.3.0

djsmith85:
## 🎟️ Tracking
https://bitwarden.atlassian.net/browse/SM-1762

## 📔 Objective
Update Jest to the new major version 30.3.0

With https://github.com/bitwarden/clients/pull/19981 the first attempt was made, but it did include a larger change-set than needed and skipped some test cases. This aims to break it up and in case some of the additional changes are needed with another bump of jest-environment-jsdom, we can revisit that PR after this merges into main.,

This PR bumps all required dependencies around Jest and fixes any broken paths or tests due to stricter typescript rules with jest-environment-jsdom (mainly mockImplementation)

Relates to: 
https://github.com/bitwarden/clients/pull/17623
https://github.com/bitwarden/clients/pull/20202
https://github.com/bitwarden/clients/pull/19843

PR #19981, opened 2026-04-03
Major upgrade to Jest 30

cd-bitwarden:
## 🎟️ Tracking

https://bitwarden.atlassian.net/jira/software/c/projects/BW/boards/99?assignee=625cb516fd06270069beaf5d&selectedIssue=SM-1762

**Summarized Changes:**
1. Window.location unmockability: jsdom doesn't allow mocking window.location directly (non-configurable), so I created wrapper functions instead of trying to mock the property itself.                                                                                                                                                          
2. LocationService wrapper (apps/browser/src/autofill/services/location.service.ts): Created mockable service for browser autofill with methods like getHref(), getHostname(), getProtocol().                                                                                                                                             
3. Connector location wrappers (apps/web/src/connectors/common.ts): Added mockable functions getLocationHref(), getLocationOrigin(), getLocationHostname() with security validation (protocol whitelisting, hostname format checking).                                                                                                             
4. Module isolation pattern (common.spec.ts, webauthn.spec.ts): Switched from module-level jest.mocked() to jest.isolateModules() blocks so internal function calls see mocked versions.
5. XSS hardening: Added security documentation and input validation to location wrapper functions (silent failures on invalid input).
6. Type suppression: Replaced commented-out mocks with @ts-expect-error comments for jest-mock-extended v4 generic type issues.
7. Test skipping: Added .skip to tests that depend on problematic window property mocking (window.location, window.top, etc.).
  
## 📔 Objective

Upgrade Jest to V30, skip all tests that are not easily fixed (these will be assigned to respective teams and fixed at a later time)

The following files will be sorted and assigned to teams to be fixed and the .skips removed 
@bitwarden/team-autofill-dev (Autofill team)
                                                                                                                                                                            
  - apps/browser/src/autofill/fido2/content/messaging/messenger.spec.ts
  - apps/browser/src/autofill/fido2/content/fido2-content-script.spec.ts                                                                                                    
  - apps/browser/src/autofill/notification/bar.spec.ts                                                                                                                      
  - apps/browser/src/autofill/services/dom-element-visibility.service.spec.ts                                                                                               
  - apps/browser/src/autofill/services/autofill-overlay-content.service.spec.ts                                                                                             
  - apps/browser/src/autofill/fido2/content/fido2-page-script.webauthn-unsupported.spec.ts                                                                                
  - apps/browser/src/autofill/fido2/content/fido2-page-script.webauthn-supported.spec.ts

  @bitwarden/team-auth-dev (Auth team)

[...]

PR #20400, opened 2026-04-27
[PM-35903] Update clients to node 24

dani-garcia:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-35903

## 📔 Objective

Update all the clients to node 24 except the CLI, which will stay behind so we don't need to update yaopkg for now. This PR is fairly similar to the one where we had to do the same thing for node 22: https://github.com/bitwarden/clients/pull/15058

Electron is using it's own bundled node version, so it will stay on 22 until #19848 is merged.

I've also fixed a manual polyfill of `Symbol.dispose`, as they're readonly on node24 and caused some tests to fail.

Because this creates the `engines` section on the CLI:
Closes https://github.com/bitwarden/clients/pull/12519
Closes https://github.com/bitwarden/clients/pull/15856


djsmith85 (commented):
The broken tests is something that @cd-bitwarden had also found with the [initial attempt](https://github.com/bitwarden/clients/pull/19981) to bump Jest to 30.
I was under the impression that `jest` and `jest-environment-jsdom` were transient dependencies via Angular. That's why we didn't have it setup in the package.json or in renovate. With me [introducing Jest 30](https://github.com/bitwarden/clients/pull/20211) as a direct dev dependency, it seems to also require `jest-environment-jsdom` ([as of Jest 28](https://jestjs.io/blog/2022/04/25/jest-28#breaking-changes)).

Thanks for raising this and using an intermediate step towards fixing this.

I'll look into fixing the tests once this is merged

djsmith85 (changes requested):
Missing a bump of the node version in `apps/web/Dockerfile`

dani-garcia:
Thanks for the heads up @djsmith85! Seems like it wasn't updated to node 22 last time so I missed it when I searched for the version number

ike-kottlowski (dismissed):

apps/cli/package.json
```diff
@@ -90,5 +90,9 @@
     "semver": "7.7.3",
     "tldts": "7.0.28",
     "zxcvbn": "4.4.2"
+  },
+  "engines": {
+    "node": "~22",
+    "npm": "~10"
```
djsmith85, line 96: With adding the engines here, please close https://github.com/bitwarden/clients/pull/12519 and https://github.com/bitwarden/clients/pull/15856

package-lock.json
```diff
@@ -254,205 +259,6 @@
         "@napi-rs/cli": "3.5.1"
       }
     },
-    "apps/desktop/node_modules/@inquirer/confirm": {
```
djsmith85: With installing `jest-environment-jsdom` it updated the package-lock.json which cleans up the gap that was caused by a mismatch with napi-rs/cli.

Several issues and PRs have been opened lately, but mainly being tracked with https://bitwarden.atlassian.net/browse/PM-36005

djsmith85 (dismissed):
Approving for platform.

Just some comments to add context and follow-up tasks, once this merges.

AmyLGalles (dismissed):
approved for BRE

dani-garcia:
Node 24.16.0 has an fs regression (nodejs/node#63487) that breaks yauzl/extract-zip, so Electron's postinstall silently fails to extract its binary. This was causing some tests that weren't mocking electron to fail. To fix this I've done two things:
- Mocked the tests to ensure they don't require an installed electron build, which will be necessary once we move to pnpm or disable install scripts on npm.
- Updated the npm version to 24.17, which partially fixes the reported bug

Note that this only affects tests and `npm run electron`, desktop builds work through `electron-builder`, which fetches `electron` in a different way and it still works. The bug is also fixed as part of electron 41.7 on #20448

djsmith85 (approved):
Approving!

Thanks for the fix. 
❓ You mention that this is fixed with the electron bump. Does that mean we can remove the changes made to the main/os-biometrics service here on https://github.com/bitwarden/clients/pull/20448?

dani-garcia:
> Approving!
> 
> Thanks for the fix. ❓ You mention that this is fixed with the electron bump. Does that mean we can remove the changes made to the main/os-biometrics service here on #20448?

I believe so yeah, electron 41.7 contains the fixed code that should work regardless of node version. That said electron 42 has switched to lazy-downloading the binaries to avoid issues with postinstall scripts, so there's a chance we'll need the mock for that too, depending how it's implemented.

mzieniukbw (approved):
Approving for KM.

mzieniukbw:
Has this been tested locally ? I have a suspicion Aikido will block this node version, since it is just 1 day old.

dani-garcia:
> Has this been tested locally ? I have a suspicion Aikido will block this node version, since it is just 1 day old.

Yeah I tested and built the apps locally and didn't have any issues using nvm. I think Aikido is only protecting against npm packages and not the npm binaries themselves.

AmyLGalles (approved):
LGTM

ike-kottlowski (approved):

.github/workflows/build-cli.yml
```diff
@@ -72,7 +72,6 @@ jobs:
 
       - name: Get Node Version
         id: retrieve-node-version
-        working-directory: ./
         run: |
           NODE_NVMRC=$(cat .nvmrc)
```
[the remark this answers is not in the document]
dani-garcia: This doesn't seem correct. According to the logs, the CLI builds are pulling 22:
https://github.com/bitwarden/clients/actions/runs/28526276131/job/84563471654?pr=20400
> Run actions/setup-node@53b83947a5a98c8d113130e565377fae1a50d02f
Found in cache @ /opt/hostedtoolcache/node/22.23.0/x64
Environment details
/opt/hostedtoolcache/node/22.23.0/x64/bin/npm config get cache
/home/runner/.npm
npm cache is not found

PR #20401, opened 2026-04-27
[PM-34497] Handle Same-Origin iframes in queryDeepSelector

bensbits91:
## 🎟️ Tracking

[PM-34497](https://bitwarden.atlassian.net/browse/PM-34497)

## 📔 Objective

Objective 1 (this PR): Enable DeepSelector to detect inputs inside a same-origin iframe. Ensure inline menu appears (potentially in the wrong position on the page) and autofill works as expected.

Objective 2 (this PR): Handle iframe offsets so the inline menu appears in the correct position on the page.

Note: Cross-origin iframes will be handled in a separate ticket: [PM-36016](https://bitwarden.atlassian.net/browse/PM-36016).

[PM-34497]: https://bitwarden.atlassian.net/browse/PM-34497?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[PM-36016]: https://bitwarden.atlassian.net/browse/PM-36016?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


bensbits91:
### Position same-origin iframe inputs

My [last commit](https://github.com/bitwarden/clients/pull/20401/commits/cf1e2f9c24e6ef29c1ae54508790bab5f553d410) includes significant changes across multiple files. This commit is intended to improve the positioning of same-origin iframe fields detected using our newer `>>>` traversal methods. 

My previous commits enabled `>>>` traversal for same-origin iframes. After those changes, field positioning was off for those fields. The intended result of this latest commit is to make positioning of these same-origin iframe fields match the positioning of iframes detected using our older heuristics. That is, we shouldn't notice any difference between positioning of fields detected using our newer `>>>` traversal.

Here's a summary of what this commit entails:

#### 1. DOM Query Service — iframe boundary detection
dom-query.service.ts: Added findIframeCrossing(selector), which walks a >>> selector segment by segment and returns the iframe element and the remaining inner selector when it hits an iframe boundary. Used to distinguish "this crosses into an iframe" from "this is a shadow DOM path." Also updated the abstract interface.

#### 2. Background — cross-frame routing
overlay.background.ts: Added the routeTargetedFieldsToFrame message handler. When the main frame's content script detects that a targeting rule selector crosses an iframe boundary, it sends {iframeSrc, iframeTargetedFields} to the background. The background looks up all frames for the tab, finds the one matching iframeSrc, and forwards applyTargetedFields to that frame's content script. Also added updateOverlayCiphers() call in handleWebNavigationOnCommitted so the inline menu cipher cache refreshes on tab navigation. Added error handling .catch() to the tab-close updateOverlayCiphers call. New message types added to the abstraction: iframeSrc, iframeTargetedFields on OverlayBackgroundExtensionMessage; routeTargetedFieldsToFrame handler on OverlayBackgroundExtensionMessageHandlers.

#### 3. Content script init — apply targeted fields
autofill-init.ts: Added an applyTargetedFields message handler that delegates to collectAutofillContentService.applyExternalTargetedFields(iframeTargetedFields). New types added to the abstraction accordingly.

#### 4. Collect service — cross-iframe field registration and targeting fallback
collect-autofill-content.service.ts: Two changes:

(4a) applyExternalTargetedFields — new method. Called when the background routes targeted fields into this frame. Resolves each selector locally (supporting shadow DOM), creates synthetic AutofillField objects with opids like targeted_field_0_username, caches them in autofillFieldElements and autofillFieldsByOpid, then sends collectPageDetailsResponse back to the background so the background updates its per-frame page details map.

(4b) getTargetedPageDetails fallback — when this frame runs its own targeting pass but all selectors cross into an iframe (so fields is empty), and autofillFieldElements already has cached data from a prior applyExternalTargetedFields call, it returns the cached page details instead of an empty response. This prevents a race where the iframe's own secondary targeting pass would overwrite the background's good page details with an empty payload.

#### 5. Overlay content service — same-origin iframe offset
autofill-overlay-content.service.ts: Added getSameOriginIframeOffset(element), which walks up from an element's ownerDocument through frameElement to accumulate the total viewport offset of all containing iframes. Modified getMostRecentlyFocusedFieldRects to apply this offset when the focused field lives inside a same-origin iframe, so the inline menu appears at the correct absolute position on the page.

#### 6. Autofill service — filter defensive fix
autofill.service.ts: Swapped filter predicate order in collectPageDetailsFromTab$ — now checks message.sender before message.tab?.id (with optional chaining added). The old code crashed with TypeError: Cannot read properties of undefined (reading 'id') when applyExternalTargetedFields sent a collectPageDetailsResponse without a tab property. This caused the vault popup's entire cipher observable chain to error, leaving the vault blank on the iframe test page.

apps/browser/src/autofill/services/collect-autofill-content.service.ts
```diff
@@ -270,37 +278,74 @@ export class CollectAutofillContentService implements CollectAutofillContentServ
...
-        const autofillField = this.buildTargetedAutofillField(
-          formFieldElement,
-          fieldType as AutofillTargetingRuleType,
-          fields.length,
-        );
+    // Route iframe-crossing selectors to their respective frames. Fire-and-
+    // forget: the iframe content scripts apply their fields asynchronously and
+    // re-send collectPageDetailsResponse with their own frameId.
+    for (const [iframeSrc, iframeTargetedFields] of iframeTargets) {
+      void sendExtensionMessage("routeTargetedFieldsToFrame", { iframeSrc, iframeTargetedFields });
```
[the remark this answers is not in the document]
bensbits91: Addressed in commit https://github.com/bitwarden/clients/pull/20401/commits/6a7d34459457666ee3548f3672fdd05113d2b160

apps/browser/src/autofill/background/overlay.background.ts
```diff
@@ -3241,6 +3278,7 @@ export class OverlayBackground implements OverlayBackgroundInterface {
         subFrames.clear();
         delete this.subFrameOffsetsForTab[tabId];
       }
+      this.updateOverlayCiphers().catch((error) => this.logService.error(error));
```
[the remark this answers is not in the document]
bensbits91: Removed in https://github.com/bitwarden/clients/pull/20401/changes/393070c7a6426b88774e4155739b5122e510bfb7

apps/browser/src/autofill/services/collect-autofill-content.service.ts
```diff
@@ -314,6 +360,54 @@ export class CollectAutofillContentService implements CollectAutofillContentServ
...
+
+    for (let i = 0; i < targetedFields.length; i++) {
+      const { selector, fieldType } = targetedFields[i];
+      const matchedElement = this.domQueryService.queryDeepSelector(selector);
+      if (!matchedElement) {
+        continue;
+      }
+
+      const formFieldElement = matchedElement as ElementWithOpId<FormFieldElement>;
+      const fieldId = `targeted_field_0_${fieldType}`;
```
jprusik: What purpose does the `0` serve in this id?

apps/browser/src/autofill/services/dom-query.service.ts
```diff
@@ -179,6 +184,71 @@ export class DomQueryService implements DomQueryServiceInterface {
...
+    return null;
+  }
+
+  /**
+   * Returns the document inside a same-origin iframe, or null if the iframe
+   * is cross-origin or its document is otherwise inaccessible.
+   */
+  private traverseIframeBoundary(element: HTMLIFrameElement): Document | null {
+    return element.contentDocument;
+  }
```
jprusik: Given that it's a direct one-line return, why not check `contentDocument` directly instead of calling this method?

apps/browser/src/autofill/services/dom-query.service.spec.ts
```diff
@@ -272,4 +272,96 @@ describe("DomQueryService", () => {
...
+      document.body.appendChild(host);
+      const shadowRoot = host.attachShadow({ mode: "open" });
+      const input = document.createElement("input");
+      input.id = "shadow-input";
+      shadowRoot.appendChild(input);
+
+      expect(domQueryService.queryDeepSelector("#shadow-host >>> #shadow-input")).toBe(input);
+    });
+
+    it("returns null when an intermediate element has no shadow root and is not an iframe", () => {
```
jprusik, line 316: This is an especially good case to cover given we expect staleness of rules to be an issue

apps/browser/src/autofill/services/dom-query.service.spec.ts
```diff
@@ -272,4 +272,96 @@ describe("DomQueryService", () => {
...
+      const host = iframe.contentDocument!.createElement("div");
+      host.id = "shadow-host";
+      iframe.contentDocument!.body.appendChild(host);
+      const shadowRoot = host.attachShadow({ mode: "open" });
+      const input = document.createElement("input");
+      input.id = "deep-input";
+      shadowRoot.appendChild(input);
+
+      expect(
+        domQueryService.queryDeepSelector("#outer-iframe >>> #shadow-host >>> #deep-input"),
```
jprusik, line 366: A few more test cases I'd like to see:
- especially deeply nested shadow root inputs (at least 4 nested shadow roots, seen commonly on atomic composition with web components)
- input within an iframe in a shadow host
- input within an iframe in an iframe

jprusik (changes requested):
Looking good - please add a few more cases (specified in comments) to the `queryDeepSelector` unit tests, and I think we'll be good to go for this pass.

apps/browser/src/autofill/services/collect-autofill-content.service.ts
bensbits91: Oops, I meant to replace that with the loop index. Addressed in https://github.com/bitwarden/clients/pull/20401/changes/ffc217f1940f26ea1c84ee458d8d1cd6dfa80c6e

I'd hardcoded `0` during initial development and testing but forgot to go back and fix. Thanks for catching that.

apps/browser/src/autofill/services/dom-query.service.ts
bensbits91: You're right, better to inline it. The method was only used in one place, didn't add any real logic, and the JSDoc didn't add value. Addressed in https://github.com/bitwarden/clients/pull/20401/changes/1f577a438923c7ab3b200c37bec4600b1afdbd08. Thanks!

apps/browser/src/autofill/services/dom-query.service.spec.ts
bensbits91: Added tests in https://github.com/bitwarden/clients/pull/20401/changes/f529fedf45b5346f96534982d56cf7a8a27b74ba

Look okay?

apps/browser/src/autofill/services/dom-query.service.spec.ts
```diff
@@ -363,5 +366,99 @@ describe("DomQueryService", () => {
...
+        value: iframeDoc,
+        configurable: true,
+      });
+
+      const input = iframeDoc.createElement("input");
+      input.id = "iframe-input";
+      iframeDoc.body.appendChild(input);
+
+      expect(
+        domQueryService.queryDeepSelector("#shadow-host >>> #shadowed-iframe >>> #iframe-input"),
```
jprusik: Nit, in case the test changes unintentionally in the future; we care about the actual iframe tag

```suggestion
        domQueryService.queryDeepSelector("#shadow-host >>> iframe#shadowed-iframe >>> input#iframe-input"),
```

apps/browser/src/autofill/services/dom-query.service.spec.ts
```diff
@@ -363,5 +366,99 @@ describe("DomQueryService", () => {
...
+        configurable: true,
+      });
+
+      const input = innerDoc.createElement("input");
+      input.id = "nested-iframe-input";
+      innerDoc.body.appendChild(input);
+
+      expect(
+        domQueryService.queryDeepSelector(
+          "#outer-iframe >>> #inner-iframe >>> #nested-iframe-input",
```
jprusik: Similar to previous nit; let's mirror the specificity of the upstream targeting rules:

```suggestion
          "iframe#outer-iframe >>> iframe#inner-iframe >>> input#nested-iframe-input",
```

apps/browser/src/autofill/services/dom-query.service.ts
```diff
@@ -170,7 +189,7 @@ export class DomQueryService implements DomQueryServiceInterface {
       if (i < segments.length - 1) {
         const next: Document | ShadowRoot | null =
           element instanceof HTMLIFrameElement
-            ? this.traverseIframeBoundary(element)
+            ? element.contentDocument
```
jprusik, line 196: reminder that we're expecting the targeting rules to specify an `iframe` tag name in front of the frame traversal, so we should ultimately handle that more authoritatively, rather than test and fallback (if the deep selector as specified fails, we don't want to fall back to another strategy since it signals a stale selector).

Not blocking at this stage, however

jprusik (dismissed):

apps/browser/src/autofill/services/dom-query.service.spec.ts
bensbits91: Addressed in https://github.com/bitwarden/clients/pull/20401/changes/8c29ad832e3c8b0ea04dc7d4abe3dccda07f4dac

apps/browser/src/autofill/services/dom-query.service.spec.ts
bensbits91: Addressed in https://github.com/bitwarden/clients/pull/20401/changes/8c29ad832e3c8b0ea04dc7d4abe3dccda07f4dac

apps/browser/src/autofill/services/dom-query.service.ts
bensbits91: Thanks Jon. Makes sense. Will defer to future work. I added a FIXME in https://github.com/bitwarden/clients/pull/20401/changes/a37d33933a06b4ebcd80243442cad5d28cb05d2e

jprusik (approved):
Nice work 👍

PR #20406, opened 2026-04-27
[FIX] Replaced hard coded auth key name with variable from AKV

gitclonebrian:
## 🎟️ Tracking
failed [build](https://github.com/bitwarden/clients/actions/runs/25017335903/job/73277670135)

## 📔 Objective
The AKV secret was updated, but the workflow referenced the auth key name via a static string. This PR changes that so that the auth key name is pulled from the AKV


djsmith85 (approved):
Changes are looking good.

Approving for platform

PR #20407, opened 2026-04-27
[PM-30213] Add key rotation UI

Thomas-Avery:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-30213

## 📔 Objective

The objective of this PR is to add in a key rotation section in the "Keys" security section. This is behind the `SdkKeyRotation` feature flag and uses the SDK to preform a user key rotation without a master password change. 

Auth reviewers: Only thing to call out is I'm adjusting how the spacing works to fit both optional sections in `apps/web/src/app/auth/settings/security/security-keys.component.html`.

## 📸 Screenshots

image

image

### Recording 

https://github.com/user-attachments/assets/ffe404c5-e85a-4b44-8a28-1a330588df0b

### Success toast

Note because we immediately logout after the key rotation most users won't see this.
Screenshot 2026-04-28 at 9 38 48 AM


Thomas-Avery:
Going to open this up for review for any addressable feedback the teams might have while waiting on the upstream.

libs/user-crypto-management/src/user-key-rotation.service.ts
```diff
@@ -69,6 +72,40 @@ export class DefaultUserKeyRotationService implements UserKeyRotationService {
     );
   }
 
+  async rotateUserKey(currentMasterPassword: string, userId: UserId): Promise<boolean> {
```
quexten: @Thomas-Avery A question: For the other methods - TDE, Key Connector, will we switch out "currentMasterPassword" for an enum, or will we make it separate methods?

libs/user-crypto-management/src/user-key-rotation.service.abstraction.ts
```diff
@@ -30,6 +30,16 @@ export abstract class UserKeyRotationService {
...
 
+  /**
+   * Rotates the user key and associated encrypted data using the SDK without a master password change.
+   * The master password should be confirmed by proof of decryption prior to calling this method.
+   * @param currentMasterPassword The users current master password.
+   * @param userId The ID of the user.
+   * @returns True if the key rotation was successful, false if the user denied trust.
+   * @throws If the SDK call fails or the SDK is not available.
+   */
+  abstract rotateUserKey(currentMasterPassword: string, userId: UserId): Promise<boolean>;
```
quexten: @Thomas-Avery A question: For the users that do not have a master-password, but have the maindecryptionmethod TDE or KC, will we replace the currentMasterPassword param with an enum, or  will we make separate functions, with the caller having to determine the maindecryptiommethod?

The naming here currently suggests the former.

libs/key-management-ui/src/key-rotation/key-rotation-dialog.service.ts
```diff
@@ -0,0 +1,71 @@
...
+      userId,
+    );
+
+    if (!proofOfDecryption) {
+      this.toastService.showToast({
+        variant: "error",
+        message: this.i18nService.t("incorrectPassword"),
+      });
+      return false;
+    }
```
quexten: Nit for readability
```suggestion
    const isMasterPasswordValid = await this.masterPasswordUnlockService.proofOfDecryption(
      masterPassword,
      userId,
    );

    if (!isMasterPasswordValid) {
      this.toastService.showToast({
        variant: "error",
        message: this.i18nService.t("incorrectPassword"),
      });
      return false;
    }
```

libs/key-management-ui/src/key-rotation/key-rotation-dialog.component.ts
```diff
@@ -0,0 +1,110 @@
...
+      title: { key: "warning" },
+      content: { key: "oldAttachmentsNeedFixDesc" },
+      acceptButtonText: { key: "learnMore" },
+      cancelButtonText: { key: "close" },
+      type: "warning",
+    });
+
+    if (learnMore) {
+      this.platformUtilsService.launchUri(
+        "https://bitwarden.com/help/attachments/#add-storage-space",
```
quexten:
```suggestion
        "https://bitwarden.com/help/attachments/#fixing-old-attachments",
```

libs/key-management-ui/src/key-rotation/key-rotation-dialog.component.spec.ts
```diff
@@ -0,0 +1,211 @@
...
+        });
+      });
+
+      it("launches learn-more URL when user clicks 'Learn more'", async () => {
+        mockDialogService.openSimpleDialog.mockResolvedValue(true);
+
+        await callSubmit(component);
+
+        expect(mockPlatformUtilsService.launchUri).toHaveBeenCalledWith(
+          "https://bitwarden.com/help/attachments/#add-storage-space",
```
quexten:
```suggestion
          "https://bitwarden.com/help/attachments/#fixing-old-attachments",
```

quexten (changes requested):

libs/key-management-ui/src/key-rotation/key-rotation-dialog.component.ts
Thomas-Avery: Updated in https://github.com/bitwarden/clients/pull/20407/commits/73b7c0f7f171e8ef331eceddf5ec0ee197b804f9. To note the current check goes to this link https://github.com/bitwarden/clients/blob/099a8fa45e6a47bbca37c9fc0f6b87ca329e00a2/libs/auth/src/angular/input-password/input-password.component.ts#L676-L704

libs/key-management-ui/src/key-rotation/key-rotation-dialog.component.spec.ts
Thomas-Avery: Updated in https://github.com/bitwarden/clients/commit/73b7c0f7f171e8ef331eceddf5ec0ee197b804f9.

libs/key-management-ui/src/key-rotation/key-rotation-dialog.service.ts
Thomas-Avery: Updated in https://github.com/bitwarden/clients/commit/73b7c0f7f171e8ef331eceddf5ec0ee197b804f9.

libs/user-crypto-management/src/user-key-rotation.service.abstraction.ts
Thomas-Avery: I think regardless the caller will need to determine what type the user is to make determinations on what UV to do. However for this method they will all flow into the same exact code so we should expose the enum. 

Updated in https://github.com/bitwarden/clients/commit/73b7c0f7f171e8ef331eceddf5ec0ee197b804f9.

libs/user-crypto-management/src/user-key-rotation.service.ts
Thomas-Avery: I think regardless the caller will need to determine what type the user is to make determinations on what UV to do. However for this method they will all flow into the same exact code so we should expose the enum. 

Updated in https://github.com/bitwarden/clients/commit/73b7c0f7f171e8ef331eceddf5ec0ee197b804f9.

apps/web/src/app/auth/settings/security/security-keys.component.html
rr-bw, line 1: Looks like the `tw-border-0 tw-mb-0` is vestigial.

What do you think about the following approach?

**1. security-keys.component.html**

```html
<div class="tw-grid tw-gap-16 tw-mt-6">
  @if (showChangeKdf) {
    <app-change-kdf></app-change-kdf>
  }

  @if (showChangeKdf && sdkKeyRotationFlag$ | async) {
    <app-user-key-rotation></app-user-key-rotation>
  }

  <div>
    <h2 bitTypography="h2">{{ "apiKey" | i18n }}</h2>

    <p bitTypography="body1">
      {{ "userApiKeyDesc" | i18n }}
    </p>

    <div class="tw-flex tw-gap-2">
      <button type="button" bitButton buttonType="secondary" (click)="viewUserApiKey()">
        {{ "viewApiKey" | i18n }}
      </button>
      <button type="button" bitButton buttonType="secondary" (click)="rotateUserApiKey()">
        {{ "rotateApiKey" | i18n }}
      </button>
    </div>
  </div>
</div>
```

**2. change-kdf.component.html**

No longer needs `tw-mt-6`. First line becomes just:

```html
<h2 bitTypography="h2">{{ "encKeySettings" | i18n }}</h2>
...
```

**3. key-rotation.component.html**

No longer needs `tw-mt-6`. First line becomes just:

```html
<h2 bitTypography="h2">{{ "rotateAccountEncryptionKey" | i18n }}</h2>
...
```

rr-bw (changes requested):
Looks like you might need a rebase as well now that you're targeting `main`.

apps/web/src/app/auth/settings/security/security-keys.component.html
Thomas-Avery: Sounds good to me. Applied changes on https://github.com/bitwarden/clients/pull/20407/commits/f1cd3d38269ebc48d21676ea7ecd8d84e4ae46bc. Also went ahead and rebased as requested.

rr-bw (approved):
Approving Auth changes.

quexten (approved):

PR #20411, opened 2026-04-28
Non-login form/field disqualifier for signup/subscription forms

blackwood:
Pulls signals from the form, the field, sibling fields, and a new `htmlAncestorHeadings: string[]` collected from the form's nearest semantic-section ancestor and ordered closest-first by depth of common ancestor. Gated behind a strict keyword list that rules out explicit cases on form/field attrs and a broader list inside branches (which maintains desired fall back behavior to login ciphers, e.g. a single email input in the first part of multi-part form). Heading text is interpreted separately by a `classifyHeadings` first-match-wins helper that recognizes both login-positive and identity-context keywords, so a closer login heading short-circuits a more distant non-login signal.

Along the way: extracted `formContainsKeyword` to mirror `fieldContainsKeyword`, swapped the boolean `substringMatch` for `mode: "appearsWithin" | "matchesToken"`, and renamed the keyword constants to `Strict` / `Broad` with the broad list defined as a literal superset. Added a small `LoginPositiveHeadingKeywords` list used only by `classifyHeadings` for the short-circuit.

Notes:
- `htmlAncestorHeadings` is a `string[]`, not a joined string, so keyword matches can't span two headings. Order is closest-first (LCA depth from the form); document order is preserved among equidistant entries.
- Heading text is excluded from `formContainsKeyword` — the strict path scans form attrs only; the broad path consults `classifyHeadings`, which can return `"login" | "identity" | "ambiguous"`. A `"login"` result short-circuits disqualification.
- `BroadNonLoginKeywords` spreads `StrictNonLoginKeywords` so the superset relationship is structural.
- `isNonLoginFormContext` is now module-private; external callers go through `isNonLoginUsernameField` (form + field + siblings) or `isAmbiguousFieldNonLogin` (gated w/ broader keywords).
- The two `autofill.service.ts` callsites moved from form-only to form + field + siblings, which broadens the behavior.

## 🎟️ Tracking
[PM-33789](https://bitwarden.atlassian.net/browse/PM-33789)

## 📔 Objective
Consolidating the qualification methods between the inline menu and the autofill service ([PM-30478](https://bitwarden.atlassian.net/browse/PM-30478)) surfaced a pre-existing issue: non-login emails (newsletter, subscription, etc.) could be improperly qualified as logins, and that misqualification now spans both qualification domains since their methods are shared. (see [PR #19363 discussion](https://github.com/bitwarden/clients/pull/19263/changes#r2885204329))

First, `isNonLoginFormContext` was being used as a proxy for "is this a non-login username field," but its checks weren't exhaustive — surrounding callsites needed to consider sibling fields and ancestor signals as login-case disqualifiers, which is the job of a separate function (see [PR #19263 discussion](https://github.com/bitwarden/clients/pull/19263#discussion_r2965830506)). Second, the disqualifier needs both an always-on strict path and a broader path gated to structurally-ambiguous cases, to expand coverage without regressing legitimate logins.

[PM-33789]: https://bitwarden.atlassian.net/browse/PM-33789?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

Related for Test Cases:
https://github.com/bitwarden/test-the-web/pull/435


jprusik:
Manually ran BIT against the branch (both passed):
with prod feature flags: https://github.com/bitwarden/browser-interactions-testing/actions/runs/25219765393
with no feature flags: https://github.com/bitwarden/browser-interactions-testing/actions/runs/25219775025

apps/browser/src/autofill/services/autofill-constants.ts
```diff
@@ -154,19 +154,30 @@ export class AutoFillConstants {
     "kennwort ändern",
   ];
 
+  /** Strict non-login keywords disqualify a form for login unconditionally. */
+  static readonly StrictNonLoginKeywords = ["newsletter"] as const;
```
jprusik: non-blocking, but also consider: "notification", "contact", and "subscribe"/"unsubscribe" (may apply to registration, but not login)

apps/browser/src/autofill/services/inline-menu-field-qualification.service.ts
```diff
@@ -440,11 +441,7 @@ export class InlineMenuFieldQualificationService implements InlineMenuFieldQuali
     }
     const passwordFieldsInPageDetails = pageDetails.fields.filter(this.isCurrentPasswordField);
 
-    /**
-     * If a field is part of a newsletter form, or other recognized non-login type forms, it isn't a username
-     * {@link AutoFillConstants.NonLoginFormKeywords}
```
jprusik, line 445: it'd be cool to put the updated description and link back in

apps/browser/src/autofill/services/autofill-constants.ts
```diff
@@ -154,19 +154,30 @@ export class AutoFillConstants {
...
-  static readonly NonLoginFormKeywords: string[] = [
-    "newsletter",
-    // @TODO expand list thoughtfully
-    // consider possible collisions with login forms
-    // consider using a "maybe" check
-    // "subscribe",
-    // "subscription",
-    // "unsubscribe",
-  ];
+  static readonly BroadNonLoginKeywords = [
```
jprusik: non-blocking nit; "broad" feels a touch vague to me; maybe something to indicate the composed array, like `BroadlyInclusiveNonLoginKeywords`, `ExhaustiveNonLoginKeywords`; or even just:

```suggestion
  static readonly InclusiveNonLoginKeywords = [
```

🤷

apps/browser/src/autofill/services/autofill-constants.ts
```diff
@@ -154,19 +154,30 @@ export class AutoFillConstants {
...
+    ...AutoFillConstants.StrictNonLoginKeywords,
+    "mailing list",
+    "subscribe",
+    "subscription",
+    "unsubscribe",
+  ] as const;
+
+  /** Login-positive heading text used to short-circuit ambiguous-case disqualification. */
+  static readonly LoginPositiveHeadingKeywords = [
```
jprusik:
```suggestion
  static readonly StrictLoginHeadingKeywords = [
```

hrm, I kinda think the `Strong`/`Weak` metaphor might serve better to indicate the non-definitive nature of these comparisons. 🤔

apps/browser/src/autofill/services/autofill-constants.ts
jprusik: Consider also that we probably _don't_ want "registration" type keywords in here. A real example from the wild:

Image

apps/browser/src/autofill/services/collect-autofill-content.service.ts
```diff
@@ -381,15 +381,60 @@ export class CollectAutofillContentService implements CollectAutofillContentServ
...
+        // so the walk always terminates at a known ancestor.
+        let ancestor: Element = heading;
+        while (!formAncestorDepths.has(ancestor)) {
+          ancestor = ancestor.parentElement!;
+        }
+        return { text, distance: formAncestorDepths.get(ancestor)! };
+      })
+      .filter((entry): entry is { text: string; distance: number } => entry !== null)
+      .sort((a, b) => a.distance - b.distance)
+      .map((entry) => entry.text);
```
jprusik: probably somewhat expensive at scale (given how we'd be hitting this for multiple inputs) - all these chained iterations could be a single (albeit more complicated) reducer

apps/browser/src/autofill/utils/qualification.ts
```diff
@@ -72,27 +78,90 @@ function buildAutofillFieldKeywords(field: AutofillField) {
...
+ * True if any keyword matches a token from the field. With `appearsWithin` (default), the
+ * keyword may appear as a substring of any token; `matchesToken` requires an exact token.
+ * Hyphens are stripped from keywords before matching.
  */
 export function fieldContainsKeyword(
   field: AutofillField,
-  keywords: string[],
-  substringMatch = true,
+  keywords: readonly string[],
+  mode: "appearsWithin" | "matchesToken" = "appearsWithin",
```
jprusik: Can we make these const/types (here and at callsites)?

jprusik (changes requested):
A few non-blocking questions/thoughts, and one blocking need to [optimize chained array iterations per field](https://github.com/bitwarden/clients/pull/20411/changes#r3174781109) lookup.

apps/browser/src/autofill/services/collect-autofill-content.service.ts
blackwood: I think the chaining is fairly cheap compared to the `.closest` query, although I think there's still simplification we can do here. Not pictured: maybe we should consider a cache on the method itself?

```ts
return Array.from(scope.querySelectorAll("h1, h2, h3, h4, h5, h6"))
  .flatMap((heading) => {
    const f = heading.closest("form");
    if (f !== null && f !== formElement) return [];
    const text = this.getTextContentFromElement(heading);
    if (!text) return [];
    // Every retained heading lives under `scope`, and `scope` is in `formAncestorDepths`,
    // so the walk always terminates at a known ancestor.
    let ancestor: Element = heading;
    while (!formAncestorDepths.has(ancestor)) {
      ancestor = ancestor.parentElement!;
    }
    return [{ text, distance: formAncestorDepths.get(ancestor)! }];
  })
  .sort((a, b) => a.distance - b.distance)
  .map((entry) => entry.text);
```

apps/browser/src/autofill/services/autofill-constants.ts
blackwood: Let's add any we decide on in a followup PR, so they're easier to roll back without impacting the logic itself.

apps/browser/src/autofill/services/autofill-constants.ts
blackwood: I like `Exhaustive` (v. maybe `Indicative`)!

blackwood: Maybe `Unconditional`/`Conditional` is an even better hint?

apps/browser/src/autofill/services/autofill-constants.ts
blackwood: You said `Strong` but wrote `Strict`--assume typo?

apps/browser/src/autofill/utils/qualification.ts
blackwood: 0f682a8849

apps/browser/src/autofill/services/collect-autofill-content.service.ts
blackwood: Actually, worth mentioning—this is called per `form` in `getPageDetails` (not each `input`)

apps/browser/src/autofill/services/autofill-constants.ts
blackwood: (I think I'm drifting from the "composed" definition, here, sorry)

apps/browser/src/autofill/services/collect-autofill-content.service.ts
```diff
@@ -381,15 +381,60 @@ export class CollectAutofillContentService implements CollectAutofillContentServ
...
 
     return this.getFormattedAutofillFormsData();
   }
 
+  /**
+   * Headings inside the form's nearest section/article/main/aside/form ancestor,
+   * ordered by depth of common ancestor (closest first). Sibling-form headings skipped.
+   */
+  private getAncestorHeadings(formElement: HTMLFormElement): string[] {
+    const scope = formElement.parentElement?.closest("section, article, main, aside, form");
```
jprusik: `form` in `form`?

apps/browser/src/autofill/services/collect-autofill-content.service.ts
```diff
@@ -381,15 +381,60 @@ export class CollectAutofillContentService implements CollectAutofillContentServ
...
+      if (cursor === scope) {
+        break;
+      }
+      cursor = cursor.parentElement;
+    }
+
+    return Array.from(scope.querySelectorAll("h1, h2, h3, h4, h5, h6"))
+      .filter((h) => {
+        const f = h.closest("form");
+        return f === null || f === formElement;
```
jprusik: I see you handled in the presumptive code proposed, but if not:

```suggestion
        return f === formElement || [];
```

apps/browser/src/autofill/services/autofill-constants.ts
jprusik: yeah, that was a typo 🤦

jprusik (approved):

PR #20423, opened 2026-04-28
[sm-1928] Undecryptable Access Tokens

cd-bitwarden:
## 🎟️ Tracking
https://bitwarden.atlassian.net/jira/software/c/projects/SM/boards/74?assignee=625cb516fd06270069beaf5d&selectedIssue=SM-1928

## 📔 Objective

If someone updates the access token to have an unencryptable value we need to gracefully handle this and allow user to maintain access to their list of access tokens.


bitwarden_license/bit-web/src/app/secrets-manager/service-accounts/access/access.service.ts
```diff
@@ -143,7 +148,13 @@ export class AccessService {
       accessTokenResponses.map(async (s) => {
         const view = new AccessTokenView();
         view.id = s.id;
-        view.name = await this.encryptService.decryptString(new EncString(s.name), orgKey);
+        try {
+          const encString = new EncString(s.name);
```
quexten, line 152: @cd-bitwarden Just a question, did we ever encounter this for real users? If so do we know what caused the corruption?

KM so far is not aware of any of these.

cd-bitwarden: @quexten no we don't have this issue with real users, you'd have to use postman and send bad data over to the server. We do have checks on the server that check if the string they pass in encrypted, however these changes help so that if they somehow pass an encrypted, but unencryptable string it keeps their access to the other data.

quexten (dismissed):
Looks fine from KM's side, but we are not code owners on the code.

justindbaur (dismissed):

justindbaur (approved):

PR #20424, opened 2026-04-28
Desktop Native: Run integration tests in CI

neuronull:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-34750

## 📔 Objective

Recently we added a number of integration tests for the SSH Agent (v2) features.

This change enables running them in CI.

Because integration tests often require more specific environment setup, and can also inherently be more flaky than unit tests, it's better to explicitly list each crate we want to enable automated runs of them for.

Note: I attempted enabling autotype int tests but as anticipated, there were problems due to it needing DISPLAY. We can follow up in a separate task if that becomes a priority.

## Testing Done

You can see the ssh agent int tests passing here: https://github.com/bitwarden/clients/actions/runs/25077411685

(ignore the autotype failure, as that change was backed out)


coltonhurst (dismissed):
Looks great, thanks @neuronull

.github/workflows/test.yml
```diff
@@ -171,20 +171,35 @@ jobs:
         with:
           persist-credentials: false
 
+      - name: Detect changed crates
+        uses: dorny/paths-filter@de90cc6fb38fc0963ad72b210f1f284cd68cea36 # v3.0.2
```
djsmith85: ⚠️  The other workflows using this GH action are using version 4.0.1, please bump

djsmith85 (changes requested):

djsmith85 (approved):
Thanks for the update. Changes are looking good!

PR #20441, opened 2026-04-29
[PM-31942] Save Access Intelligence reports to files (pt. 3)

lastbestdev:
**Depends on server PR https://github.com/bitwarden/server/pull/7228**

## 🎟️ Tracking
Part 3 of https://bitwarden.atlassian.net/browse/PM-31942 

## 📔 Objective
The last PR for integration of saving Access Intelligence reports as files. Changes include adding a new implementation of the ReportPersistenceService abstraction, solely for the purpose of conditionally choosing to store reports using the old method (database column storage), or the new way (file storage).

The implementation is determined by the value of the newly added feature flag, for using report file storage.

## 📸 Screenshots
N/A


libs/common/src/enums/feature-flag.enum.ts
```diff
@@ -143,7 +144,8 @@ export const DefaultFeatureFlagValue = {
   [FeatureFlag.PhishingDetection]: FALSE,
   [FeatureFlag.Milestone11AppPageImprovements]: FALSE,
   [FeatureFlag.AccessIntelligenceTrendChart]: FALSE,
-  [FeatureFlag.AccessIntelligenceNewArchitecture]: FALSE,
+  [FeatureFlag.AccessIntelligenceNewArchitecture]: true,
+  [FeatureFlag.AccessIntelligenceReportFileStorage]: true,
```
Banrion: ❌ Default should be FALSE

Banrion (changes requested):

libs/common/src/enums/feature-flag.enum.ts
lastbestdev: Every once and a while I accidentally commit the file haha, good catch!

Banrion (dismissed):

Banrion (approved):

PR #20445, opened 2026-04-30
[PM-36048] - redirect back to vault after coachmark tour ends

jaasen-livefront:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-36048

## 📔 Objective

The 3-step welcome tour previously left users on the Reports page after finishing. This was confusing because it left users stranded away from the primary starting point for setting up their vault.

This change updates `completeTour()` in `CoachmarkService` to navigate the user back to `/vault` upon tour completion — whether they finish the last step or dismiss the tour early. This ensures users land on the Vault page where the checklist and empty-state UX guide them to import data or add items.

## 📸 Screenshots

https://github.com/user-attachments/assets/085ed3b5-3b4e-4c27-9067-5833ec15035f


jengstrom-bw (approved):
Looks good!

PR #20448, opened 2026-04-30
[PM-34331] Update electron to 41

dani-garcia:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-34331

## 📔 Objective

Update electron to version 41 and the rest of the electron updates to their latest patch, which should solve compilation issues with the latest Visual Studio.

I've also had to enable a new fuse from electron 41, but I just set it to the default value so everything should work the same: https://github.com/electron/fuses/blob/bbf079259e66af30858578a1e070575f11c01915/README.md?plain=1#L29

Note that this will cause the desktop app to start using the bundled node 24, the rest of the clients except CLI are updated as part of #20400 

I can't get renovate to create the PR right now so I've created it myself instead.


djsmith85 (dismissed):
Changes are looking great, but we should get https://github.com/bitwarden/clients/pull/20400 merged first. Since everyone needs to update their node-version for development.

dani-garcia:
The last commit is fixing a process reload bug reported in https://bitwarden.atlassian.net/browse/PM-36851
The fix is at: https://github.com/bitwarden/clients/pull/20448/changes/6e5f54958cbf188c451690db4ea16137e8863094?w=1

The bug appeared in electron 41 and caused process reload to crash the renderer, but then not reload the page, leaving the window empy and stuck. This issue is caused by two conditions:
- The process-reload event gets fired multiple times in close sequence, causing a race between the crash+reload. I fixed this by adding some logic to skip any duplicate events while we're in the middle of a reload.
- Calling reload immediately after crashing the process does not execute the reload correctly a lot  of times. This is a reported electron bug: https://github.com/electron/electron/issues/48661. Just putting an await setImmetiate between crash and reload is enough to fix it for me consistently over dozens of tries.

SuperSandro2000:
Electron 39 went EOL 2 weeks ago and we want to switch to 40 or 41 in nixpkgs. Will this land with the next desktop release?

dani-garcia:
Hey @SuperSandro2000, I believe today is the RC cut for the desktop release next week, so I don't think this will make it on time for next update. We had some QA issues on Linux when using electron 41, which is the cause for the delay.

SuperSandro2000:
Will this make it into the next release?

djsmith85 (approved):

SuperSandro2000:
So with the approval it will make it into the next minor release? 👀

addisonbeck:
@SuperSandro2000  Yes, this will be in the next desktop release: `2026.7.0`. That's still a ways away, our next release cycle is just server and web. Desktop `2026.7.0` will go out July 23rd. 

Thanks for the work you do on nixpkgs! Couldn't get through my life without it.

amarshall:
@addisonbeck That’s 11.5 weeks of Bitwarden Desktop running EOL Electron, and 6 months after the initial release of v40.

Beware also that Electron 41 will be EOL on Aug 25, so there will be less than 5 weeks before it’s EOL again after v2026.7.0 release (Electron 43 will be released in a few days as well). I implore Bitwarden to be more proactive in the future with regard to Electron updates, not only to ease downstream packaging for us and others, but to ensure it’s providing secure software to its users.

(Please note that any frustration perceived in my tone comes out of a want to see Bitwarden be great and safe and open, and is not directed at you specifically but at the situation. I’ve dealt with complex Electron updates in the past, and it’s not fun.)

PR #20449, opened 2026-04-30
[PM-21777]Defect Safari - Save login notification bar icon is stretched and Save login is scrollable

dan-livefront:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-21777
## 📔 Objective

The edit button was stretching vertically in Safari which caused the cipher row to extend vertically. Edited styling to avoid this case, additionally, removed typography from here for styling, it is not needed. 
## 📸 Screenshots

Screenshot 2026-04-30 at 10 14 17 AM


apps/browser/src/autofill/content/components/buttons/edit-button.ts
```diff
@@ -33,16 +37,16 @@ export function EditButton({ buttonAction, buttonText, disabled = false, theme }
 }
 
 const editButtonStyles = ({ disabled, theme }: { disabled?: boolean; theme: Theme }) => css`
-  ${typography.helperMedium}
-
   user-select: none;
   display: flex;
+  align-items: center;
+  justify-content: center;
```
blackwood, line 43: Non-blocking nit:

> `align-items` [...] sets the [align-self](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/align-self) value on all direct children as a group

Since there is only one item, is `align-self` on the child more conventional? And could the same be said RE: `justify-`?

blackwood (approved):

PR #20453, opened 2026-04-30
[PM-32743] Convert collections to folders when importing into My Items

mcamirault:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-32743

## 📔 Objective

This PR changes what happens to collections when importing into My Items. Previously other collections a cipher might be part of were ignored because My Items does not support nesting; with this change the other collections will be recreated as folders with the same names, with any collection hierarchy recreated as a folder hierarchy and with the cipher in the folder created from its collection. Since a cipher can only have one folder but be part of multiple collections, we now throw an error if you try to import into My Items with any ciphers that are part of multiple collections. Importing into the base of the org or into any other collection besides My Items is unchanged.

## 📸 Screenshots

For the following demos we are importing from an org export with three collections and two ciphers:
**Default Collection**: One cipher called "Default Collection Credential"
**Developers**: No ciphers, one sub-collection called "Tools Developers"
**Tools Developers**: One cipher called "Tools Dev Credentials"

*Importing into the base of an empty organization. This is will also happen if importing into any collection besides My Items, though the collection structure will be under the target collection*

https://github.com/user-attachments/assets/937f8129-9a76-4055-8c91-f2d45fa4f510

This results in the following collection structure:
- **Default Collection**: One cipher
- **Developers**: No ciphers, one sub-collection
  - **Tools Developers**: One cipher

and no folders are created

*Importing into My Items*

https://github.com/user-attachments/assets/7d822e22-39cd-408e-b9ab-03c935373c1d

This results in the following collection structure:
- **My Items**: Two ciphers

and the following folder structure:
- **Default collection**: One cipher
- **Developers**: No ciphers, one sub-folder
  - **Tools Developers**: One cipher


harr1424 (dismissed):
Barring the  ❌ CRITICAL finding reported during the Claude review, this looks good to me! Happy to re-review if it's not a false positive.

harr1424 (dismissed):

mcamirault:
@claude Review this again

harr1424 (approved):

PR #20459, opened 2026-04-30
[PM-29654] Hide copy button when custom field has no value

gbubemismith:
## 🎟️ Tracking
https://bitwarden.atlassian.net/browse/PM-29654

## 📸 Screenshots

https://github.com/user-attachments/assets/41e2d653-d263-4160-aa54-93b9eb5848b5


shane-melton (dismissed):

shane-melton (approved):

shane-melton (dismissed):

libs/vault/src/cipher-view/custom-fields/custom-fields-v2.component.html
```diff
@@ -17,18 +17,21 @@ <h2 bitTypography="h6">{{ "customFields" | i18n }}</h2>
...
-              bitIconButton="bwi-clone"
-              bitSuffix
-              type="button"
-              [appCopyClick]="field.value"
-              showToast
-              [valueLabel]="field.name"
-              [label]="'copyCustomField' | i18n: field.name"
-              data-testid="copy-custom-field"
-            ></button>
+            @if (field.value) {
```
nick-livefront, line 23: ❓ In the video the alignment of the copy button shifted down, is it possible to make that inline?

Image

gbubemismith: I don't think that was introduced in this PR, looks like it's an alignment issue from `form-field.component`. I can create a separate ticket in the backlog for this

nick-livefront (approved):

PR #20472, opened 2026-05-01
[PM-34918] Get rid of encrypt as any 

JaredScar:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-34918

## 📔 Objective

Gets rid of as any for encrypt function.


eliykat (changes requested):
This has a few sets of changes in it. The `encrypt` method is covered by the reversion in #20477. The actions service changes seem unrelated and maybe belong to another branch? I suggest we close this PR and you can open a new one for those changes if needed.

PR #20502, opened 2026-05-04
Small fix for ssh namespace mismatch

coltonhurst:
## 🎟️ Tracking

None

## 📔 Objective

This PR aims to fix a _potential_ case mismatch in the ssh agent v1 and v2 "namespace" implementations. It may not be accurate, but here is the concept:

Line 59 of `clients/apps/desktop/src/autofill/components/approve-ssh-request.ts` checks the namespace string, currently matching on `"git"` for v1. However, v2's `SIGNamespace.Git` enum variant from `clients/apps/desktop/desktop_native/napi/src/sshagent_v2.rs` ends up being `"Git"`, according to line 308 in `clients/apps/desktop/desktop_native/napi/index.d.ts`.

`"Git"` won't match here, so `actioni18nKey` will end up being `"sshActionSign"`, rather than `"sshActionGitSign"`.

There may be a better way to fix this, or this may be the intended behavior, and this PR is not needed. Just throwing this PR up as a starting point for conversation, and as a potential fix.

If the bug identified _is_ accurate, then we'll want to verify this doesn't break anything on the ssh v1 side.

### DIFF
--- a/apps/desktop/src/autofill/components/approve-ssh-request.ts
+++ b/apps/desktop/src/autofill/components/approve-ssh-request.ts
@@ -56,7 +56,7 @@ export class ApproveSshRequestComponent {
     namespace: string,
   ) {
     let actioni18nKey = "sshActionLogin";
-    if (namespace === "git") {
+    if (namespace === "git" || namespace === "Git") {
       actioni18nKey = "sshActionGitSign";
     } else if (namespace != null && namespace != "") {
       actioni18nKey = "sshActionSign";
--- a/apps/desktop/src/autofill/components/approve-ssh-request.ts
+++ b/apps/desktop/src/autofill/components/approve-ssh-request.ts
@@ -56,7 +56,7 @@ export class ApproveSshRequestComponent {
     namespace: string,
   ) {
     let actioni18nKey = "sshActionLogin";
-    if (namespace === "git" || namespace === "Git") {
+    if (namespace === "git") {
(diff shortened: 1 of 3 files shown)


apps/desktop/src/autofill/components/approve-ssh-request.ts
```diff
@@ -56,7 +56,7 @@ export class ApproveSshRequestComponent {
     namespace: string,
   ) {
     let actioni18nKey = "sshActionLogin";
-    if (namespace === "git") {
+    if (namespace === "git" || namespace === "Git") {
```
neuronull: nice catch, can we try using the following in the napi declaration for the enum instead? That way, we don't have to do any shenanigans electron side.

```Rust
#[napi(string_enum = "camelCase")]
```

coltonhurst: Sure, updated in 96d0b7b099e341c643b67e1d3e3c130c8c459c6d. One side benefit is that doesn't require any testing for QA, as the change is only on the V2 side!

Any concerns about the casing changing for the other variants?

neuronull: thanks! No concerns. We aren't consuming that enum yet in the v2 on the Electron side.

neuronull (approved):

dani-garcia (approved):

PR #20506, opened 2026-05-05
[PM-31061] Implement biometrics migration

quexten:
https://bitwarden.atlassian.net/browse/PM-31061

Implements migration of biometrics. This had been reviewed and approved before, but had to be reverted out due to a dependency cycle conflict. This cycle has been resolved by updating how the SDK exposes the biometric functionality.


apps/desktop/src/key-management/biometrics/main-biometrics.service.spec.ts
```diff
@@ -29,6 +29,18 @@ jest.mock("@bitwarden/desktop-napi", () => {
...
+  SdkLoadService: {
+    Ready: Promise.resolve(),
+  },
+}));
+
+jest.mock("@bitwarden/sdk-internal", () => ({
+  CryptoClient: {
+    get_key_id_for_symmetric_key: jest.fn().mockReturnValue(null),
+  },
+}));
```
Thomas-Avery: ❓ I don't see either of these in `apps/desktop/src/key-management/biometrics/main-biometrics.service.ts`?

apps/desktop/src/main.ts
```diff
@@ -233,6 +253,19 @@ export class Main {
...
+      this.environmentService,
+      this.platformUtilsService,
+      this.accountService,
+      this.kdfConfigService,
+      this.keyService,
+      this.accountCryptographicStateService,
+      this.apiService,
+      stateProvider,
+      this.configService,
+    );
```
Thomas-Avery: ❓ I'm not following the change here. All the new `this.` services are going to be undefined here?

```
  platformUtilsService: PlatformUtilsService;
  accountService: AccountService;
  kdfConfigService: KdfConfigService;
  keyService: KeyService;
  accountCryptographicStateService: AccountCryptographicStateService;
  apiService: ApiService;
  configService: ConfigService;
```

The question is why do we need to create `DefaultSdkService` in `main.ts`?

libs/common/src/key-management/encrypted-migrator/migrations/biometric-persistent-encryption-migration.ts
```diff
@@ -0,0 +1,82 @@
...
+ * since the last biometric enrollment. It detects this by comparing the stored
+ * enrolled key ID with the current user key's key ID.
+ */
+export class BiometricPersistentMigration implements EncryptedMigration {
+  constructor(
+    private readonly keyService: KeyService,
+    private readonly biometricsService: BiometricsService,
+    private readonly biometricStateService: BiometricStateService,
+    private readonly logService: LogService,
+    private readonly sdkService: SdkService,
```
Thomas-Avery:
```suggestion

```

Looks like it can be removed upstream too.

libs/common/src/key-management/encrypted-migrator/migrations/biometric-persistent-encryption-migration.ts
```diff
@@ -0,0 +1,82 @@
...
+    await this.biometricsService.enrollPersistent(userId, userKey);
+    await this.biometricsService.setBiometricProtectedUnlockKeyForUser(userId, userKey);
+
+    await SdkLoadService.Ready;
+    const keyId = CryptoClient.get_key_id_for_symmetric_key(userKey.toEncoded());
+    if (keyId != null) {
+      await this.biometricStateService.setBiometricEnrolledKeyId(
+        userId,
+        Utils.fromBufferToB64(keyId),
+      );
```
Thomas-Avery: This same logic gets ran on the RendererBiometricsService enrollPersistent invoked above. It seems odd to be doing it twice? I think there might be a unintended effect going on here as  enrollPersistent is always going to make sure the keyId has a value and this will always just wipe it out?

libs/common/src/key-management/encrypted-migrator/migrations/biometric-persistent-encryption-migration.ts
```diff
@@ -0,0 +1,82 @@
+import { firstValueFrom } from "rxjs";
+
+import { SdkLoadService } from "@bitwarden/common/platform/abstractions/sdk/sdk-load.service";
+import { SdkService } from "@bitwarden/common/platform/abstractions/sdk/sdk.service";
+// eslint-disable-next-line
```
Thomas-Avery:
```suggestion
// eslint-disable-next-line no-restricted-imports
```

libs/common/src/key-management/encrypted-migrator/default-encrypted-migrator.spec.ts
```diff
@@ -1,20 +1,30 @@
 import { mock } from "jest-mock-extended";
 
-// eslint-disable-next-line no-restricted-imports
-import { KdfConfigService } from "@bitwarden/key-management";
+import { SdkService } from "@bitwarden/common/platform/abstractions/sdk/sdk.service";
+// eslint-disable-next-line
```
Thomas-Avery:
```suggestion
// eslint-disable-next-line no-restricted-imports
```

libs/common/src/key-management/encrypted-migrator/default-encrypted-migrator.ts
```diff
@@ -1,15 +1,24 @@
-// eslint-disable-next-line no-restricted-imports
-import { KdfConfigService } from "@bitwarden/key-management";
+import { SdkService } from "@bitwarden/common/platform/abstractions/sdk/sdk.service";
+// eslint-disable-next-line
```
Thomas-Avery:
```suggestion
// eslint-disable-next-line no-restricted-imports
```

Thomas-Avery (changes requested):

apps/desktop/src/key-management/biometrics/main-biometrics.service.spec.ts
quexten: Originally they were present in the main service, but got moved into the render service, since constructing an SDK service in main was not feasible without major scope creep.

quexten: Removed.

apps/desktop/src/main.ts
quexten: Reverted out now. Left over from back and forth on sdk client, apologies.

libs/common/src/key-management/encrypted-migrator/migrations/biometric-persistent-encryption-migration.ts
quexten: You're right, this was not intentional. Moved to only happen on `enrollPersistent`. 

> I think there might be a unintended effect going on here as enrollPersistent is always going to make sure the keyId has a value and this will always just wipe it out?

I don't think this is the case. For a case where either function sets the key id, both set it. For the case where the user does not set a key id (is v1), the enrollment would have done nothing, while the migration would have wiped the state (had there been any).

Now, it will always set the key id, or wipe it depending on the user-key on enrollment, and the migration just does enrollment.

libs/common/src/key-management/encrypted-migrator/default-encrypted-migrator.spec.ts
```diff
@@ -1,20 +1,29 @@
 import { mock } from "jest-mock-extended";
 
-// eslint-disable-next-line no-restricted-imports
-import { KdfConfigService } from "@bitwarden/key-management";
+// eslint-disable-next-line
```
Thomas-Avery:
```suggestion
// eslint-disable-next-line no-restricted-imports
```

apps/browser/src/key-management/biometrics/foreground-browser-biometrics.ts
```diff
@@ -80,4 +80,9 @@ export class ForegroundBrowserBiometricsService extends BiometricsService {
     userId: UserId,
     value: SymmetricCryptoKey,
   ): Promise<void> {}
+
+  async enrollPersistent(userId: UserId): Promise<void> {}
```
Thomas-Avery:
```suggestion
  async enrollPersistent(userId: UserId, key: SymmetricCryptoKey): Promise<void> {}
```

Thomas-Avery (changes requested):
Should be good after this.

apps/browser/src/key-management/biometrics/foreground-browser-biometrics.ts
quexten: It's confusing to me that the CI / eslint don't catch this. Just typescript things i guess...

quexten:
CI failure is not related to this PR.

Thomas-Avery (approved):
LGTM

PR #20508, opened 2026-05-05
[CL-1115] accordion component

BryanCunningham:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/CL-1115

## 📔 Objective

Create new accordion component

#### Size: default
image

#### Size: sm
image

#### Size: default
image

#### Size: sm
image


libs/components/src/accordion/accordion.component.spec.ts
```diff
@@ -0,0 +1,156 @@
...
+    });
+
+    it("does not render subtitle span when not provided", () => {
+      fixture.detectChanges();
+      const spans = fixture.nativeElement.querySelectorAll("button span");
+      expect(spans.length).toBe(1);
+    });
+  });
+
+  describe("host classes", () => {
```
vleague2: This block of tests doesn't seem particularly useful

libs/components/src/accordion/accordion.component.ts
```diff
@@ -0,0 +1,155 @@
...
+  host: {
+    "[class]": "hostClassList()",
+  },
+})
+export class AccordionComponent {
+  private readonly group = inject(AccordionGroupComponent, { optional: true });
+
+  constructor() {
+    effect(() => {
+      const activeId = this.group?._activeAccordionId();
```
vleague2: The `_` in the variable name seems like it implies a private variable. Since it's being accessed outside the parent class, can we make it regularly named?

libs/components/src/accordion/accordion.component.ts
```diff
@@ -0,0 +1,155 @@
...
+      if (activeId != null && activeId !== this._baseId) {
+        untracked(() => this.open.set(false));
```
vleague2, line 47: Why the `untracked` wrapper here?

libs/components/src/accordion/accordion.component.ts
```diff
@@ -0,0 +1,155 @@
...
+      "focus-visible:tw-ring-inset",
+      "focus-visible:tw-ring-border-focus",
+      "focus-visible:tw-border-border-focus",
+      "disabled:tw-cursor-not-allowed",
+      "disabled:tw-text-fg-inactive",
+      ...this.triggerSizeClasses(),
+    ].join(" "),
+  );
+
+  private readonly triggerSizeClasses = computed((): string[] => {
```
vleague2: There are other styles inline as ternaries; does this one need to be separate?

libs/components/src/accordion/accordion.component.html
```diff
@@ -0,0 +1,36 @@
...
+  [disabled]="disabled() || null"
+  [attr.aria-expanded]="open()"
+  [attr.aria-controls]="contentId"
+  (click)="toggle()"
+>
+  @if (startIcon()) {
+    <bit-icon-tile [icon]="startIcon()" [size]="iconTileSize()" variant="contrast"></bit-icon-tile>
+  }
+  <div class="tw-flex tw-flex-col tw-flex-1 tw-min-w-0 tw-gap-1 tw-items-start">
+    <span [class]="headingClassList()">{{ heading() }}</span>
```
vleague2: Are there any typography variants that we can use for heading and subtitle?

libs/components/src/accordion/accordion.stories.ts
```diff
@@ -0,0 +1,231 @@
...
+  render: (args) => ({
+    props: args,
+    template: /*html*/ `
+      <bit-accordion
+        [heading]="heading"
+        [subtitle]="subtitle"
+        [(open)]="open"
+        [disabled]="disabled"
+        [size]="size"
+        [variant]="variant"
```
vleague2: Should we be using the `formatArgsForStorybook` (that might not be the right name) utility here?

libs/components/src/accordion/accordion.stories.ts
```diff
@@ -0,0 +1,231 @@
...
+  render: (args) => ({
+    props: args,
+    template: /*html*/ `
+      <bit-accordion
+        [heading]="heading"
+        [subtitle]="subtitle"
+        [(open)]="open"
+        [disabled]="disabled"
+        [size]="size"
+        variant="subtle"
```
vleague2: You could render the default story and pass the `subtle` variant arg here since it looks identical otherwise

libs/components/src/accordion/accordion.stories.ts
```diff
@@ -0,0 +1,231 @@
...
+    template: /*html*/ `
+      <bit-accordion
+        [heading]="heading"
+        [subtitle]="subtitle"
+        [(open)]="open"
+        [disabled]="disabled"
+        [size]="size"
+        [variant]="variant"
+      >
+        <p class="tw-m-0">
```
vleague2: Feels a little weird to be using `p` with margin-0 in all these stories since it looks like we're having to already do overrides on the usage. Span instead?

libs/components/src/accordion/accordion.component.html
```diff
@@ -0,0 +1,36 @@
...
+    @if (size() === "default" && subtitle()) {
+      <span [class]="subtitleClassList()">{{ subtitle() }}</span>
+    }
+  </div>
+  <ng-content select="[end]"></ng-content>
```
vleague2: Can we do `slot=end` like our other components?

libs/components/src/accordion/accordion.component.ts
```diff
@@ -0,0 +1,155 @@
...
+  constructor() {
+    effect(() => {
+      const activeId = this.group?._activeAccordionId();
+      if (activeId != null && activeId !== this._baseId) {
+        untracked(() => this.open.set(false));
+      }
+    });
+  }
+
+  readonly heading = input.required<string>();
```
vleague2: Thoughts on using `title` instead of `heading`? (I feel like we use that naming elsewhere)

libs/components/src/accordion/accordion.component.ts
BryanCunningham: Not necessarily. I can move it. Generally, I like the idea of all the styles being in the component vs the template. I guess I should just commit one way or the other :)

vleague2: For this one I think it is still used in the component side not the template. It's just split out into another function by itself

libs/components/src/accordion/accordion.component.ts
BryanCunningham: We are closing this accordion imperatively if the parent group component is single select and is controlling which accordion is open. We had to use `untracked` here because you can't write to a model inside of an effect AFAIK

libs/components/src/accordion/accordion.component.ts
BryanCunningham: Updated to be inline

libs/components/src/accordion/accordion.component.html
BryanCunningham: Yes and no. Because this switches text sizes and colors based on inactive state and the size variant passed in this felt easier to swap

libs/components/src/accordion/accordion.component.ts
BryanCunningham: Yeah, I'm not sure why I did it that way.

libs/components/src/accordion/accordion.component.spec.ts
BryanCunningham: Removed

vleague2 (dismissed):

vleague2 (approved):

PR #20511, opened 2026-05-05
[PM-31364] Discard edit bug fixes

JaredScar:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-36512
https://bitwarden.atlassian.net/browse/PM-36516
https://bitwarden.atlassian.net/browse/PM-36517
https://bitwarden.atlassian.net/browse/PM-36518
https://bitwarden.atlassian.net/browse/PM-36532
https://bitwarden.atlassian.net/browse/PM-36533

## 📔 Objective

Resolves all bugs brought upon by the Discard Edits dialog recently implemented.


willmartian (changes requested):
Could you describe why changes to UIF code are necessary? Based on my reading of the diff, I assume `bitDialogClose` is not calling close predicate and we want it to?

 If that is the case, I think the error is that `bitDialogClose` injects `DialogRef` from the CDK directly, instead of from the CL, which looks like a bug. I think the proper fix would be switching that import.

apps/web/src/app/admin-console/organizations/policies/policy-edit-definitions/auto-confirm-policy.component.html
willmartian, line 1: These can go back to using `bitDialogClose` now, right?

willmartian (dismissed):
One non-blocking comment:

willmartian (approved):

BTreston (approved):

PR #20519, opened 2026-05-06
[CL-1191] - spotlight and popover fixes

jaasen-livefront:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/CL-1191

## 📔 Objective

Fixes issues with the spotlight effect used in guided tour popovers, and adds customization inputs to `bit-popover`.

**Resize lag** — CDK's `reposition` scroll strategy only reacts to scroll events, so resizing the window caused the spotlight cutout to trail behind the highlighted element. Added a `window` `resize` listener in `SpotlightService` that immediately calls `updatePosition()`, cleaned up when the overlay is disposed.

**Spotlight cropping** — The spotlight cutout was sizing to the directive's host element, which could be wider than the visual target (e.g. a full-width nav row). Added a `spotlightTargetSelector` input to `PopoverAnchorForDirective` that accepts a CSS selector to find a child element to use as the spotlight target instead.

## 📸 Screenshots

Screenshot 2026-05-05 at 2 17 16 PM

https://github.com/user-attachments/assets/2ad79c20-0efd-4ec6-9f73-ecad60ad3e0e


libs/components/src/popover/popover.component.ts
```diff
@@ -25,6 +25,18 @@ export class PopoverComponent {
   /** Optional title displayed in the popover header */
   readonly title = input("");
 
+  /** Optional CSS class(es) applied to the inner content container */
```
BryanCunningham: @jaasen-livefront Why do we need to customize popover elements like this? Not sure we want to intentionally allow for inconsistent popover designs

libs/components/src/popover/spotlight.service.ts
```diff
@@ -172,11 +173,21 @@ export class SpotlightService {
       this.borderOverlayRef.updatePosition();
     });
     this.resizeObserver.observe(target);
+
+    // Reposition on window resize — CDK's reposition scroll strategy only covers scroll events
+    this.windowResizeListener = () => {
+      this.borderOverlayRef?.updatePosition();
+    };
+    window.addEventListener("resize", this.windowResizeListener);
```
BryanCunningham, line 185: I'm not sure we want to always attach this resize listener. Feels like it's not going to be used most of the time. Can we find a way to conditionally attach this?

libs/components/src/tw-theme.css
```diff
@@ -339,7 +339,7 @@
   /* Hover & Overlay */
   --color-bg-hover: rgba(var(--color-gray-950-rgb), 0.05);
   --color-bg-hover-contrast: rgba(var(--color-white-rgb), 0.05);
-  --color-bg-overlay: rgba(var(--color-gray-950-rgb), 0.3);
+  --color-bg-overlay: rgba(var(--color-gray-950-rgb), 0.55);
```
BryanCunningham: I'm not sure why we need this change. Does the sidenav item need to be showing it's hover state when we 'spotlight' it?

libs/components/src/popover/spotlight.service.ts
BryanCunningham: Or is this applicable outside of the spotlight use case? Maybe I'm not wrapping my head around it fully

libs/components/src/tw-theme.css
BryanCunningham: This is the color defined by design for the backdrop globally. If we are going to make changes here we'll need sign-off from design to do so.

libs/components/src/popover/spotlight.service.ts
BryanCunningham: Never mind... I just realized this is in the spotlight service. Not the popover component. Please disregard 😆

libs/components/src/popover/popover.component.ts
vleague2: If these are coming from the width/padding design feedback, we don't need to implement overrides here. Width and padding will be updated in #20021.

jaasen-livefront:
@BryanCunningham Updated the PR to focus only on two items that aren't covered by existing UIF work.

libs/components/src/popover/popover.component.ts
jaasen-livefront: @BryanCunningham @vleague2 Reverted.

libs/components/src/tw-theme.css
jaasen-livefront: @BryanCunningham Reverted.

libs/components/src/popover/popover-anchor-for.directive.ts
```diff
@@ -68,6 +68,12 @@ export class PopoverAnchorForDirective implements OnDestroy {
   /** Padding around the spotlight cutout in pixels */
   readonly spotlightPadding = input<number>(0);
 
+  /**
+   * CSS selector to find a child element within the anchor to use as the spotlight target.
+   * If not provided, the anchor element itself is used.
+   */
+  readonly spotlightTargetSelector = input<string>();
```
BryanCunningham: We can remove this now. This will be handled in the service moving forward

jaasen-livefront: Nice! Removed. https://github.com/bitwarden/clients/pull/20519/commits/4681307a5137742a220bd9d06e92490474c9c0fd

BryanCunningham (dismissed):

jaasen-livefront:
@BryanCunningham Just need one last approval. Had to resolve a merge conflict.

BryanCunningham (approved):

PR #20520, opened 2026-05-06
[PM-36609] Add regex manager for updating electronVersion on electron updates

trmartin4:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-36609

## 📔 Objective

We currently have to manually update the `electron-builder.json` `electronVersion` for each electron update.

This PR adds in a custom `regex` manager to look in `electron-builder.json` and update the version when `electron` updates.

This is similar to what we do for the Rust toolchain update already.


dani-garcia (approved):
Nice, thank you!

PR #20523, opened 2026-05-06
[PM-31760] Implement sign() for PrivateKey

coltonhurst:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-31760

## 📔 Objective

- Implement `sign()` for `PrivateKey` in the `crypto` module for SSH v2.
- Add tests

There were two main approaches I looked at taking:
1. (this pr) Pulling in the `signature::Signer` trait that `ssh_key` has implementations for. While this is a new dependency that will need approval, it's still part of RustCrypto; it comes from their [traits](https://github.com/RustCrypto/traits) repository.
2. (not this pr) Attempting to sign without pulling in a new dependency; this would use `ssh_key`'s `sign` function on their own private key. This returns their `SshSig` type though, which is more high level than we probably want at this implementation level.

A note on versions:
- The `signature` crate recently released v3. The latest stable version released was v2.2. However, we can't use v3 yet, as `ssh_key` doesn't yet have a release with `signature` v3 supported. This is being worked on, though! (link: https://github.com/RustCrypto/SSH/pull/501)

## 🧪 Testing

No QA needed at this stage. Only the added unit tests were run.


coltonhurst:
Opening this as ready for review. It is ready for engineering review (FYI @neuronull & @quexten). However, because this PR adds a dependency, the dependency will need to be reviewed by AppSec. If an approach using the proposed new dependency is the desired & approved one, then I'll submit the dep for AppSec's review.

apps/desktop/desktop_native/ssh_agent/Cargo.toml
```diff
@@ -12,6 +12,7 @@ base64 = { workspace = true }
 desktop_core = { path = "../core" }
 mockall = { workspace = true }
 rkyv = "=0.8.14"
+signature = { workspace = true }
```
neuronull: 👋 

It looks like we should be able to use the existing features that we're enabling for `ssh-key` crate, in order to achieve signing for the algorithms. Their [docs](https://docs.rs/ssh-key/latest/ssh_key/#supported-signature-algorithms) state

> The “Feature” column lists the name of ssh-key crate features which can be enabled to provide full support for the “Keygen”, “Sign”, and “Verify” functionality for a particular SSH key algorithm.

For example: https://docs.rs/ssh-key/latest/ssh_key/private/struct.RsaKeypair.html#method.try_sign

, which is available with `rsa`.

🤔 I might be missing something, but is there a reason we couldn't use those, and not need this new dependency? (note, this is different than the `sign()` function that is mentioned in the PR description).

apps/desktop/desktop_native/Cargo.toml
```diff
@@ -65,6 +65,7 @@ serde_json = "=1.0.127"
 serde_with = "=3.14.1"
 serial_test = "=3.4.0"
 sha2 = "=0.10.9"
+signature = "=2.2.0"
```
neuronull: In the event that we _do_ need this new dependency, is there a reason placing it here? My understanding was that we'd surface them to the workspace level when more than one workspace crate depended on it. In this case, as with others like `rkyv`, we only use it in the `ssh_agent` crate.

quexten:
@coltonhurst @neuronull `signature` does not require an AppSec review. It already is a (transitive) dependency of both `sdk-internal` but also `desktop-native`. All you are doing is making it a direct dependency. It is not a net new dependency.

From the confluence:

> This applies to any third-party library, open-source package, or external software component not already in use in the target codebase.

`signature` is not a library that is not already use in the target codebase, since it is already used as part of e.g. the eliptic curve implementations.

apps/desktop/desktop_native/Cargo.toml
quexten: `signature` is not a new dependency. It is already a dependency of all other signature schemes.

apps/desktop/desktop_native/ssh_agent/Cargo.toml
quexten: `signature` is not a new dependency. It is already a dependency of all other signature schemes.

coltonhurst:
> @coltonhurst @neuronull `signature` does not require an AppSec review. It already is a (transitive) dependency of both `sdk-internal` but also `desktop-native`. All you are doing is making it a direct dependency. It is not a net new dependency.
> 
> From the confluence:
> 
> > This applies to any third-party library, open-source package, or external software component not already in use in the target codebase.
> 
> `signature` is not a library that is not already use in the target codebase, since it is already used as part of e.g. the eliptic curve implementations.

Yeah, I was aware it was a transitive dependency, was going to submit it just to be safe. I'll seek clarity on that.

apps/desktop/desktop_native/Cargo.toml
coltonhurst: @quexten yes, thanks for calling that out. (Commented on that [here](https://github.com/bitwarden/clients/pull/20523#issuecomment-4397303175)!)

@neuronull I'm good with moving it to just the `ssh_agent` crate

apps/desktop/desktop_native/ssh_agent/Cargo.toml
coltonhurst: @quexten yes, thanks for calling that out. (Commented on that [here](https://github.com/bitwarden/clients/pull/20523#issuecomment-4397303175)!)

@neuronull let me know if _I'm_ missing something! I believe in order to use that function we have to pull the trait definition (from the `signature` crate) into scope.

neuronull: 👍  got it, yeah there seems to be zero docs that outline that dependency on the `signature` crate / the traits crates.

apps/desktop/desktop_native/Cargo.toml
neuronull: > signature is not a new dependency. It is already a dependency of all other signature schemes.

Good callout. That can be seen by looking at the lock file diff.

coltonhurst: Updated in 507a7ba118b725c264283ba1b43fd8262b831ed1

neuronull: Moving it down to the ssh_agent crate also resolved the unowned dep lint GHA failure

apps/desktop/desktop_native/ssh_agent/src/crypto/mod.rs
```diff
@@ -25,6 +29,25 @@ pub enum PrivateKey {
...
+    /// # Errors
+    ///
+    /// Returns an error if the underlying signing operation fails.
+    pub fn sign(&self, data: &[u8]) -> anyhow::Result<Signature> {
+        match self {
+            Self::Ed25519(kp) => kp.try_sign(data),
+            Self::Rsa(kp) => kp.try_sign(data),
+        }
+        .map_err(|e| anyhow!("Signing failed: {}", e))
+    }
```
neuronull, line 57: I believe we don't need to return a `Result` using try_sign here, since we are storing the private keys in memory.

ref: https://docs.rs/signature/2.2.0/signature/trait.Signer.html#tymethod.try_sign

```suggestion
        match self {
            Self::Ed25519(kp) => kp.sign(data),
            Self::Rsa(kp) => kp.sign(data),
        }
    }
```

apps/desktop/desktop_native/ssh_agent/src/crypto/mod.rs
```diff
@@ -121,4 +145,55 @@ mod tests {
...
+
+        let sig = private_key.sign(TEST_DATA).unwrap();
+
+        public_key.verify(TEST_DATA, &sig).unwrap();
+    }
+
+    #[test]
+    fn test_privatekey_sign_rsa_signature() {
+        let keypair = RsaKeypair::random(&mut OsRng, MIN_KEY_BIT_SIZE).unwrap();
+        let public_key = keypair.public.clone(); // use clone impl
```
neuronull: not a blocking question just curious- I see the two signature test cases with this comment use clone and use copy impl. Why is this relevant / using both and calling it out? 
If there is a practical need to test both paths, I think it'd be better to have separate test cases for that.
However, I'm not following why it's relevant 🤔

coltonhurst: The comments are there just to mark that the difference for that line between the tests is intentional. I'm happy to remove them if that's preferred!

neuronull: If it's intentional, as I mentioned, I think it's worth having separate cases to make that more clear (and more concretely validate specific things vs bundling validations in multiple tests.

but, if there isn't a reason to do that, then that means the clone is unecessary, and in that case we should remove the clone call, both comments, and keep it consistent!

coltonhurst: Okay, would you prefer two `test_privatekey_sign_ed25519_signature()` tests, where one uses `clone` and one uses `copy`? `Ed25519Keypair` has a `copy` and a `clone` impl; `RsaKeypair` only has a `clone` impl. Instead of using `clone` for both, I used `copy` where possible as it should be faster, and they ideally are doing the same thing.

In my opinion I don't think the goal for those tests is to test those trait implementations, just the signing, but I'm happy to alter this as needed.

apps/desktop/desktop_native/ssh_agent/src/crypto/mod.rs
coltonhurst: Good call, and as discussed internally, thanks for explaining the why there!

After looking at their docs and their implementation for `sign` vs `try_sign`, my current preference is to still use `try_sign`, because `sign` can panic. Even if we should never hit the case where it does panic (based on our usage), just because of the specific way they designed and implemented these (a unique case), I would still prefer to go the safer route.

You are more familiar with the usage of this in the whole context of ssh v2; if returning a `Result` causes problems for users of this function, then definitely let me know the implications! Worst case, while not my preference, another option is to still call `try_sign` but return a `Signature` on success, and on failure, add our own panic, so it's very visible what's happening. Then we can document it with a comment or something. (I don't see any mention of the fact `sign` _can_ panic in their docs, which is not ideal.)

apps/desktop/desktop_native/ssh_agent/src/crypto/mod.rs
neuronull: > and they ideally are doing the same thing.
> In my opinion I don't think the goal for those tests is to test those trait implementations, just the signing

That was my expectation as well, but since there were code comments I wanted to inquire about it.

Thus, can we do the below? Or, are you in disagreement with my thought process? I think removing the clone and removing the code comments would be the most clear path.

> but, if there isn't a reason to do that, then that means the clone is unecessary, and in that case we should remove the clone call, both comments, and keep it consistent!

coltonhurst: > That was my expectation as well, but since there were code comments I wanted to inquire about it.

Yeah, totally fair 👍

> Thus, can we do the below? Or, are you in disagreement with my thought process? I think removing the clone and removing the code comments would be the most clear path.
>> but, if there isn't a reason to do that, then that means the clone is unecessary, and in that case we should remove the clone call, both comments, and keep it consistent!

No disagreement on my side, just trying to communicate I can't remove `clone` on `RsaKeypair` because it doesn't implement `copy`.

apps/desktop/desktop_native/ssh_agent/src/crypto/mod.rs
neuronull: To add clarity- the reason I was able to surface this at all is because I was looking at the diff and thinking about the error testing (specifically, how we would test the error path).

That led me to reading their docs, and then to this notion: if we have already created the key means the key is cryptographically valid, so the signing of data with it can't fail.

Their docs state `"The main intended use case for signing errors is when communicating with external signers, e.g. cloud KMS, HSMs, or other hardware tokens."`

and since we are storing the private keys in memory, it would appear we are safe not to use try_sign due to our use case.

My goal is just to make sure our code simple and correct, and that the testing we do is necessary. If we are having to create tests to inject failure cases that we'll never encounter, or code paths that can't be hit, then that's wasted effort.

@quexten , maybe your expertise can help us make sense of the RustCrypto trait!

---

In the event that we _do_ keep using try_sign() , I would then ask these questions: what failures do we expect to occur here, and what will our upstream consumers of the sign() function need to take? The answers to these questions, to me, would guide our choices on what to do with the `signature::Error`

apps/desktop/desktop_native/ssh_agent/src/crypto/mod.rs
neuronull: ohhhh 😂  in that case, I think removing the `copy` code comment, and editing the `clone` code comment to say: `RsaKepair` doesn't implement `copy` , would make the most sense to viewers of the code.  Alternatively, removing both could also work. If anyone is bothering to read the code of the unit tests, and tries to remove the clone, the compiler will tell them. But I think the code comment is the best 👍

apps/desktop/desktop_native/ssh_agent/src/crypto/mod.rs
coltonhurst: > My goal is just to make sure our code simple and correct, and that the testing we do is necessary. If we are having to create tests to inject failure cases that we'll never encounter, or code paths that can't be hit, then that's wasted effort.

I definitely agree. After our internal thread, I am confident we are fine on the crypto side of things. (Happy to hear @quexten's thoughts there though.)

My concern I'm trying to highlight in my response above is just from an API standpoint. If there is any possible error / panic / etc. that can return from a function we are using, I want to make sure we have a plan for it. Even if it should never happen, if it's still _possible_ according to a lib's API or implementation, I would prefer to lean into the strength we get from Rust's design.

apps/desktop/desktop_native/ssh_agent/src/crypto/mod.rs
coltonhurst: Sure, updated in 1cf4d21edaedf4d73f8ae690c442de7de63f8b2c

apps/desktop/desktop_native/ssh_agent/src/crypto/mod.rs
neuronull: >  If there is any possible error / panic / etc. that can return from a function we are using
> Even if it should never happen

these statements are talking about using `sign()` , is that correct? (because the try_sign() returns Error, so it is expected to be able to occur, and my earlier comments about `sign()` are discussing the cases where we're not expecting it to not occur). If so,

> I am confident we are fine on the crypto side of things

The above statement suggests you are content with using `sign()`, which is different from your prior preference:

> my current preference is to still use try_sign

and in that case I can't tell what you're advocating for :( Please let me know if I'm misinterpreting something.

With

> I would prefer to lean into the strength we get from Rust's design.

Could you expand on that? It's not clear what you're referring to.

If you're referring to

> Worst case, while not my preference, another option is to still call try_sign but return a Signature on success, and on failure, add our own panic, so it's very visible what's happening

, I'm not following how that's an improvement over just using `sign()` ? I would think both panics are going to be equally visible?

---

To me the options are:

1. use `sign()` , much like we use `mutex.lock().expect("mutex not to be poisoned")` , and we can add a comment if that'd be helpful, to state why we are using sign and not try_sign.
2. use `try_sign()` and either:
  a. wrap in `anyhow!` as you have here
  b. return `signature::Error` directly
  c. use `thiserror`
  
As I indicated in my previous messages, my understanding is option 1 is most correct for us. But I'd like to get @quexten's thoughts for confidence, given the design that RustCrypto took for this trait.

And if we end up going with option 2 , my questions then become 

> what failures do we expect to occur here, and what will our upstream consumers of the sign() function need to take?

, and it may be that 2.a ends up being our approach, if there are valid error paths that can occur, and if we are only ever taking one action based on any error path that occurs. Would be great to get @quexten 's expertise to help us understand that.

coltonhurst: Sorry my message isn't clear - my main goal is to avoid using a function that can panic. Even if, in theory, we shouldn't ever hit that case, I would rather use `try_sign()` (option 2) and handle it on our side.

coltonhurst:
> > @coltonhurst @neuronull `signature` does not require an AppSec review. It already is a (transitive) dependency of both `sdk-internal` but also `desktop-native`. All you are doing is making it a direct dependency. It is not a net new dependency.
> > From the confluence:
> > > This applies to any third-party library, open-source package, or external software component not already in use in the target codebase.
> > 
> > 
> > `signature` is not a library that is not already use in the target codebase, since it is already used as part of e.g. the eliptic curve implementations.
> 
> Yeah, I was aware it was a transitive dependency, was going to submit it just to be safe. I'll seek clarity on that.

Update from AppSec; we are good to promote `signature` to a direct dependency.

apps/desktop/desktop_native/ssh_agent/src/crypto/mod.rs
neuronull: That's alright, thank you for clarifying.

I'm hoping to gather more data (by means of @quexten's insight) before a decision is made, but 

My take is essentially that based on the information present, it suggests we wouldn't _in practice_ be able to encounter it (again hoping @quexten can help clarify this) , and if it's the case, then if we use `try_sign()` just because the `sign()` function _technically_ could panic, but only _theoretically_ panics (by means of the use cases we don't have) because of unfortunate design choices by the rustcrypto crate, then by doing so we are just inheriting their unfortunate design choices, by having to have logic to handle errors (and provide injections for in testing, and explanatory code comments) that we can't hit in practice. But, that's just my opinion on how to approach that situation. Let's see if that situation is our reality.

And that if we do actually have legitimate cases where this can Error for our usage, then I'm on board for using try_sign (would then pose my aforementioned inquiries).

FWIW, I tracked down where that was introduced, and here is the [commit](https://github.com/RustCrypto/signatures/commit/c9269c1bcdad3b26c2eb804d5dff28b17c82cafb) message (no PR) , and the content supports my theory. The initial implementation of the trait only had a function that returned the `Error` but rather than refactor the underlying in order to have distinct code paths, they simply provided a function that did what all their callsites were probably doing ( e.g. `sign().expect("the signing not to fail because the key is stored in memory`)

> For many signature providers, it's not possible for an error to occur
when creating a signature. Support for handling errors which occur when
computing signatures is intended for use with cloud KMS, HSMs, or other
hardware tokens where things like I/O errors are possible.

> To simplify usage in the event that only software signers are in use,
this commit splits the `sign` API into `Result`-based `try_sign` method
and another `sign` method that always assumes success and unwraps the
result of the former method.

quexten: I concur with @neuronull. I think we should use `sign` and not `try_sign`. If we can explicitly state why we trust a function to not panic, we don't need to deal with error cases that will not realistically occur, and doing so is standard practice in other parts of the codebase.

Here, we are working with the SSH key types from the `ssh-key` crate, which generally already validate that they are correct during creation / TryFrom. Signatures should never fail.

Just because the code path contains a panic condition does not mean it can actually happen, and in this case it cannot happen. The docs specifically state this is for cases where the signing operation can fail for cases like HSMs, where the key may not be available / the signing operation may not respond.

coltonhurst: @quexten thanks for your review here! This seems to come down to what we are comfortable with in regards to panic-able code. This is especially relevant in protection against future edits where consumers of this code may not be aware of the limitations. My current personal preference is to avoid it if we can, by making it extremely clear through Rust's powerful enum types (Result in this case).

While I don't think the cons of doing it this way outweigh the pros, if people are extremely against this approach, then we can leave the panic-able code in, but we need to heavily document all appropriate locations so users of this code are very aware of what the limitations are.

coltonhurst: Approach updated with comments in 22629d5a334a7320bfd9995b1e31f80ea8773aeb

apps/desktop/desktop_native/ssh_agent/src/crypto/mod.rs
```diff
@@ -23,6 +23,12 @@ use ssh_key::{
 pub use crate::storage::keydata::{QueryableKeyData, SSHKeyData};
 
 /// Represents an SSH private key.
+///
+/// # Hardware-backed keys
+///
+/// Hardware-backed keys are not supported. The variants of this enum hold concrete key material in
+/// memory, which is incompatible with hardware-backed key workflows. If hardware-backed key support
+/// is ever added, the [`PrivateKey::sign`] function must be updated accordingly.
```
neuronull: Two thoughts:
- nit: this is largely verbatim what is in the other doc comment, I think it's fine to simply state that it only supports in-memory keys and to see the sign function for details.
- I think this doc comment could actually be moved to the module doc comment at the top of the file, where we already have notes on the limitations. e.g. after the line "//! ECDSA keys are not currently supported (PM-29894)"

apps/desktop/desktop_native/ssh_agent/src/crypto/mod.rs
```diff
@@ -36,15 +42,17 @@ impl PrivateKey {
...
     ///
-    /// # Errors
+    /// # Hardware-backed keys
     ///
-    /// Returns an error if the underlying signing operation fails.
-    pub fn sign(&self, data: &[u8]) -> anyhow::Result<Signature> {
+    /// Hardware-backed keys are not supported. This function signs directly using key material
+    /// held in memory and does not delegate to any hardware device. If hardware-backed key support
+    /// is ever added, this function must be updated. Consult the [`ssh_key`] crate documentation
+    /// before making any changes.
```
neuronull: It would be good to specify why external are not supported (it's currently somewhat ambiguous). For some reason the GH UI is not letting me do a change suggestion, so I'll note it here:

```Rust
/// We do not support external (e.g. hardware backed) keys for the SSH Agent feature. If support is expanded for them, this function must be re-worked. At present we're using the infallible function of the [`ssh_key`] crate's `Signer` trait, which is sufficient when the keys are stored in memory.
///
/// <https://docs.rs/signature/2.2.0/signature/trait.Signer.html#tymethod.try_sign>
```

apps/desktop/desktop_native/ssh_agent/src/crypto/mod.rs
```diff
@@ -36,15 +42,17 @@ impl PrivateKey {
     ///
     /// A [`Signature`] containing the algorithm identifier and raw signature bytes.
     ///
-    /// # Errors
+    /// # Hardware-backed keys
```
neuronull: nit: in the module level doc and here, we could more correctly broadly refer to these as "external signers", matching the ssh_key crate's documentation. We could say "e.g. hardware keys" though.

neuronull (commented):
Thanks for making changes to use the fallible function and documentation. Looks good overall, just some opinion takes.

apps/desktop/desktop_native/ssh_agent/src/crypto/mod.rs
coltonhurst: Simplified the comment in d080a347c673a40cc495ea1203edd2846d784a7a. Would prefer to leave it here because it only applies to the `PrivateKey` type.

apps/desktop/desktop_native/ssh_agent/src/crypto/mod.rs
coltonhurst: (As you mentioned) they aren't supported at a feature level; if we want to add that I think it would be better to do it at a higher level location in another PR.

apps/desktop/desktop_native/ssh_agent/src/crypto/mod.rs
coltonhurst: Updated in d080a347c673a40cc495ea1203edd2846d784a7a

apps/desktop/desktop_native/ssh_agent/src/crypto/mod.rs
neuronull: I think it's actually very useful to tie the connection here. As it stands, the doc comment could be interpreted to indicate that the in-memory factor is specific to this function, which it is not, it's just that we're making a conscious choice to utilize a specific flavor of our dependency crate _based_ on the feature-level support. Putting the note about that at an upstream place would be less helpful IMO. But, I'm not going to die on that hill so you can disregard if you're not seeing what I'm seeing.

I do however think it's an objective value-add to link to the Signer trait documentation. We do this for Autotype where our doc strings reference the dep crate's docs. Especially since this detail about the external keys is not surfaced in the trait implementations, which is where folks are likely to look if they just use a go-to feature in their IDE when investigating our doc comment. 

In fact, we should probably update the "Consult the ssh_key" crate note or remove it entirely like my suggestion in prior comment has, since that documentation is actually in the `signature` crate, not the `ssh_key` crate.

coltonhurst: Thoughts?

```
    /// Signs the provided data using this private key.
    ///
    /// # Returns
    ///
    /// A [`Signature`] containing the algorithm identifier and raw signature bytes.
    ///
    /// # External signers
    ///
    /// Hardware-backed keys are not supported by the SSH Agent feature. This function signs
    /// directly using key material held in memory and does not delegate to any hardware device. If
    /// hardware-backed key support is ever added, this function must be updated. Consult the following
    /// for more information.
    ///
    /// <https://docs.rs/signature/2.2.0/signature/trait.Signer.html>
```

neuronull: Sounds good to me

coltonhurst: Updated in 3971b76da625fc49a0a4e498c68852d3cc77af9c

quexten (approved):

neuronull (approved):

PR #20525, opened 2026-05-06
[PM-35196] use compile-time flag to enable performance instrumentation

audreyality:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-35196

## 📔 Objective

BIT cannot activate performance metrics within the content script's world. Swapping to an always-on compile-time flag allows BIT to run benchmarks while protecting content script timing information from host page inspection in production builds.


bensbits91 (dismissed):
One non-blocking comment/question/nit: Should we update the JSDoc comment at [performance.ts#L191](https://github.com/bitwarden/clients/blob/main/apps/browser/src/autofill/content/performance.ts#L191) to add the new namespace (i.e. `${name}:poison` >  `${name}:poison:autofill:bw`)?

jprusik (approved):

bensbits91 (approved):

PR #20526, opened 2026-05-06
PM-36610 derive users from cache after import

voommen-livefront:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-36610

## 📔 Objective

Applies to access intelligence V2
After importing data, check in the ciphers cache if there are ciphers imported/created for the org
This helps determine the next step for the user.

## 📸 Screenshots

none at this time.

### DIFF
--- a/bitwarden_license/bit-common/src/dirt/access-intelligence/services/abstractions/access-intelligence-data.service.ts
+++ b/bitwarden_license/bit-common/src/dirt/access-intelligence/services/abstractions/access-intelligence-data.service.ts
@@ -1,6 +1,6 @@
 import { Observable } from "rxjs";
 
-import { OrganizationId, UserId } from "@bitwarden/common/types/guid";
+import { OrganizationId } from "@bitwarden/common/types/guid";
 import { CipherView } from "@bitwarden/common/vault/models/view/cipher.view";
 
 import { ReportProgress } from "../../../reports/risk-insights/models/report-models";
@@ -56,6 +56,14 @@ export abstract class AccessIntelligenceDataService {
    * Null when no generation is in progress.
    */
   abstract readonly reportProgress$: Observable<ReportProgress | null>;
+
+  /**
+   * Whether the organization has any ciphers
+   *
+   * Emits true once ciphers are loaded during initialization.
+   * Emits false before initialization or when the org has no ciphers.
+   */
+  abstract readonly hasCiphers$: Observable<boolean>;
 
   /**
    * Initialize service for a specific organization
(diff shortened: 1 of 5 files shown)


Banrion (changes requested):
Thanks for digging into this and identifying the root cause!

The current proposed solution will run into a correctness problem specific to this feature. `cipherViews$` returns only ciphers the admin personally has in their loca vault. Org ciphers they access via the admin endpoint are excluded here. So an admin viewing an org that has ciphers but doesn't have personal collection membership would still see the wrong empty state. 

The correct data source for "are there any ciphers for this org?" still needs to come from the **getAllFromApiForOrganization** call. Rather than adding a new function to the abstraction, I'd suggest loading the ciphers during `initializeForOrganization$` in parallel with `loadLastReport$`. This way, `ciphers$` in the data service are accurate from the moment that page initializes and the existing `hasCiphers` in the component works as intended.

The tradeoff is the one additional api call on every page load, but it's the natural cost of getting the correct data.

bitwarden_license/bit-common/src/dirt/access-intelligence/services/abstractions/access-intelligence-data.service.ts
```diff
@@ -57,6 +57,14 @@ export abstract class AccessIntelligenceDataService {
    */
   abstract readonly reportProgress$: Observable<ReportProgress | null>;
 
+  /**
+   * Whether the organization has any ciphers
+   *
+   * Emits true once ciphers are loaded during initialization.
+   * Emits false before initialization or when the org has no ciphers.
+   */
+  abstract readonly hasCiphers$: Observable<boolean>;
```
Banrion: ♻️ DEBT: `hasCiphers$` on the abstraction is redundant, we can derive from `ciphers$` instead. Adding this to the abstract class commits every future implementation and test mocks to maintaining the second observable. Components can derive this cleanly like the following example: 

```typescript
private readonly ciphers = toSignal(this.accessIntelligenceService.ciphers$, { initialValue: [] });
protected readonly hasCiphers = computed(() => this.ciphers().length > 0);
```

I suggest removing this and restoring the computed boolean on a component need basis.

bitwarden_license/bit-common/src/dirt/access-intelligence/services/implementations/view/default-access-intelligence-data.service.ts
```diff
@@ -429,6 +435,10 @@ export class DefaultAccessIntelligenceDataService extends AccessIntelligenceData
     );
   }
 
+  private loadCiphersOnly$(orgId: OrganizationId): Observable<CipherView[]> {
+    return from(this.cipherService.getAllFromApiForOrganization(orgId));
+  }
+
```
Banrion: ❓  One liners can be in-lined where used. In this instance, I think it's good to have the helper and expand upon this to include a `catchError` to not block to last report load flow in the `forkJoin`. The generate report flow returns a more general report generation error when `getAllFromApiForOrganization` fails. 

For loading ciphers on initialization, we should consider what happens if the cipher load fails. Before changes in this pr, the cached ciphers from the report generation was used, and errors weren't communicated to the user. Since we don't have a whole page error state and an error loading the ciphers doesn't necessarily block the report loading, I think catching and logging the error is sufficient for now, but I'm happy to hear other thoughts.

```suggestion
  private loadCiphersOnly$(orgId: OrganizationId): Observable<CipherView[]> {
    return from(this.cipherService.getAllFromApiForOrganization(orgId)).pipe(
      catchError((err) => {
        this.logService.error("[DefaultAccessIntelligenceDataService] Cipher load failed", err);
        return of([] as CipherView[]);
      }),
    );
  }
```

Banrion (dismissed):
The fix looks good! I added a comment to call out some debt that the `hasCiphers$ `would add, and one comment about catching an error. May be worth addressing now, but not necessarily blocking.

Banrion (approved):

PR #20528, opened 2026-05-06
PM-36583 - Add override for 'tmp' library to 0.2.4

nikwithak:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-36583

## 📔 Objective

Adds a version override for `tmp` to `0.2.4` to fix CVE-2025-54798.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -8534,13 +8534,6 @@
         }
       }
     },
-    "node_modules/@inquirer/external-editor/node_modules/chardet": {
-      "version": "2.1.1",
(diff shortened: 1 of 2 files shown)


nick-livefront (approved):
Should we have a followup ticket to remove this when the associated package that requires it moves above the `0.24` version?

nikwithak:
@nick-livefront I created https://bitwarden.atlassian.net/browse/PM-37187 to track it - note that we using `>=` in the override doesn't lock in the version, so we will get the newer version when the associated package upgrades it.

PR #20533, opened 2026-05-07
[PM-36534] - fixes to coachmarks

jaasen-livefront:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-36534

## 📔 Objective

**Collapsed nav fallback** — When the side nav is collapsed, child nav items inside a `bit-nav-group` are not rendered. Added a second `[bitPopoverAnchorFor]` binding on the tools `bit-nav-group` itself, gated by `!sideNavService.open()`, so the spotlight fires on the group's container pill when the import coachmark is active and the nav is collapsed.

**Step indicator copy** — Renamed `coachmarkStepIndicator` → `coachmarkStepsIndicator` and shortened the message from `"Step $CURRENT$ of $TOTAL$"` to `"$CURRENT$ of $TOTAL$"`.

**Description text color** — Added `tw-text-fg-body` to the coachmark description paragraph for correct color token usage.

## 📸 Screenshots
Screenshot 2026-05-26 at 6 02 49 PM
Screenshot 2026-05-26 at 6 02 31 PM


apps/web/src/app/layouts/user-layout.component.html
```diff
@@ -34,6 +41,7 @@
       [position]="coachmarkService.getStepPosition('monitorSecurity')"
       [closeOnBackdropClick]="false"
       [spotlight]="true"
+      spotlightTargetSelector="[data-testid='nav-item-container']"
```
BryanCunningham: @jaasen-livefront I wonder if we should use a different/new data attr here. So that changing the test id doesn't silently break this functionality. Thoughts?

jaasen-livefront: @BryanCunningham Totally agree.

BryanCunningham: @jaasen-livefront With my upcoming changes, you can remove this entirely. This will be handled by the service now!

nikwithak (dismissed):

BryanCunningham:
@jaasen-livefront I thought this use case was already handled by the service? I wouldn't think you would need to add a 2nd coachmark to the 'group' to handle the closed case. Am I mistaken?

nikwithak (approved):

jaasen-livefront:
@BryanCunningham Currently the `bitPopoverAnchorFor` doesn't go up the chain looking for a nav-group anchor if it can't find one in the nav-item  so we need to explicitly attach it in this case.

BryanCunningham:
> @BryanCunningham Currently the `bitPopoverAnchorFor` doesn't go up the chain looking for a nav-group anchor if it can't find one in the nav-item so we need to explicitly attach it in this case.

@jaasen-livefront Yeah, I tested this out after I sent that message :( I could have sworn this use case was taken care of in my recent update to the service but, I'll have to take another look. This is totally fine for now. Once I fix it, we can revisit!

BryanCunningham (approved):

PR #20537, opened 2026-05-07
[PM-34172] Fix badge position for TOTP on stripe.com on Chrome

bensbits91:
## 🎟️ Tracking

[PM-34172](https://bitwarden.atlassian.net/browse/PM-34172)

## 📔 Objective

On Google Chrome, fix the inline menu badge appearing on top of the first digit in Stripe's multi-field TOTP verification code form, obscuring it from the user.

### Issue and Root Cause

Stripe's TOTP input uses a "cursor-follower" pattern (CodePuncher): a single hidden <input> (opacity: 0.01) that starts at container/form width, but shrinks to a single digit-box width (~40px) at focus time and repositions to align with whichever digit is currently active. Because the hidden field fallback mechanism fires on the focus event, we measure the input's bounding rect after Stripe has already repositioned it. With a measured width of ~40px and height of ~52px, the badge positioning formula places the badge at the right edge of that narrow element — which coincides with the first visible digit box.

### Fix

In `updateMostRecentlyFocusedField`, after measuring the focused field's rect, we detect the cursor-follower pattern by checking two conditions: the field is a TOTP field (`isTotpField`), and its measured width <= height (a normal text input is typically much wider than tall; a single digit-box is roughly square). When both are true, we walk up the DOM to find the nearest ancestor whose width > inputHeight * 2 — the container spanning all digit boxes — and substitute its left and width for badge positioning. The input's own top and height are preserved for correct vertical placement. This ensures the badge is positioned at the right edge of the full digit row regardless of where the cursor-follower input currently sits.

Unit tests cover the container-walking logic and the detection conditions.

### Before fix

image

---

### After fix

image

Figma design used for reference: https://www.figma.com/design/PObYTVXo0re9GwyKJxC8CG/Inline-auto-fill-menu?node-id=5255-3092&t=NpFNPOlJnOwBZ37e-4

[PM-34172]: https://bitwarden.atlassian.net/browse/PM-34172?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/browser/src/autofill/background/overlay.background.ts
```diff
@@ -1802,6 +1802,20 @@ export class OverlayBackground implements OverlayBackgroundInterface {
         if (bounds) {
           ({ left, top } = bounds);
         }
+      } else if (
+        this.focusedFieldData.isCursorFollowerTotp &&
+        left !== undefined &&
+        width !== undefined &&
+        top !== undefined &&
+        height !== undefined
```
blackwood, line 1810: Non-blocking: entire if statement could be collapsed to two predicates
`[left, top, width, height].every(property => property !== undefined)`

apps/browser/src/autofill/services/autofill-overlay-content.service.ts
```diff
@@ -1014,10 +1014,31 @@ export class AutofillOverlayContentService implements AutofillOverlayContentServ
...
+    // to single-digit-box width at focus time and repositions to follow the active digit.
+    // Detect this by checking width <= height (normal inputs are always much wider than
+    // tall), then walk up to find the container spanning all digit boxes so the inline
+    // menu list is aligned to the full form width and the badge is positioned above the
+    // right edge of the form rather than overlapping the first digit.
+    let isCursorFollowerTotp = false;
+    if (
+      autofillFieldData &&
+      this.inlineMenuFieldQualificationService.isTotpField(autofillFieldData) &&
+      width <= height
```
blackwood, line 1032: > normal inputs area always much wider than tall

The individual digit inputs here don't follow that rule, right? (in your attached screenshot, they are more tall than wide) 

Are those mask inputs naturally excluded from this logic because there's always a singular input bound to the TOTP value that is the one being identified in qualification here?

blackwood (approved):
One nit, one question, nothing blocking!

apps/browser/src/autofill/services/autofill-overlay-content.service.ts
bensbits91: Sorry, I think the screenshots are misleading. In stripe's TOTP forms, the six visible boxes aren't actually `<input>` fields. They're decorative `<div data-testid="code-puncher-field">` elements with no form semantics. The actual form field is a single hidden `<input data-testid="code-puncher-input" autocomplete="one-time-code">` positioned over them (it shrinks and follows the cursor at focus time). So our field collection only ever sees that one input, and `isTotpField` is true exactly once. The decorative divs are never qualified.

blackwood: It sounds like that's the more typical pattern (`<div>`s intstead of `<input>s) for the mask among these types of fields across sites? the decisive language around input shapes is more what threw me off, but I think this converges on a pretty narrow condition, at any rate--thanks for the clarification.

PR #20543, opened 2026-05-07
Add filter-as-you-type to autofill inline menu

Gamerhund:
# Add filter-as-you-type to the autofill inline menu

## Summary

When typing in a login field, the inline menu now filters the cipher list in real time based on what you type. Previously, the list showed all matching ciphers for the current domain regardless of the field value — this PR changes it so the list updates as the user types, matching on cipher name and username.

This is a frequently requested UX improvement. Several threads in the community forums have asked for this, and it brings the inline menu behavior closer to what users expect from native browser password managers.

## Problem

`triggerFormFieldInput` was closing the inline menu list every time a key was pressed, and only reopening it when the field was empty again. This meant users saw the full list on focus but had no way to narrow it down by typing.

```
// Before: list closes on input, only reopens for empty fields
await this.sendExtensionMessage("closeAutofillInlineMenu", { ... });
if (!formFieldElement?.value) {
  await this.sendExtensionMessage("openAutofillInlineMenu");
}
```

## Solution

Added a new `filterInlineMenuCiphers` message that updates the cipher list in place without closing and reopening the menu. The filter runs against the already-loaded `inlineMenuCiphers` Map (populated on focus), so there are no extra vault decryption calls on each keystroke.

The filter matches on:
- Cipher name
- Login username
- Card number
- Identity full name

If the field is cleared, the full list is shown again. If no ciphers match, the list closes cleanly instead of showing an empty dropdown.

## Changes

**`autofill-overlay-content.service.ts`**
- `triggerFormFieldInput` now sends `filterInlineMenuCiphers` with the current field value instead of the close/reopen pattern

**`overlay.background.ts`**
- New `filterInlineMenuCiphers` private method — updates the list in place if visible, opens with filter if not
- New `openInlineMenuWithFilter` private method — positions the menu only when it isn't already visible (avoids repositioning on every keystroke)
- Filter logic in `getInlineMenuCipherData` — applies case-insensitive substring match when a `filterValue` is provided
- `OverlayBackgroundExtensionMessage` interface extended with `filterValue?: string`
- Message handler map updated to route `filterInlineMenuCiphers`

## Testing

Tested manually with multiple accounts on the same domain:
- Typing narrows the list correctly
- Backspacing to empty shows the full list again
- Typing with no matches closes the list
- Switching between fields resets the filter correctly
- No visible flickering while typing

## Related

- https://community.bitwarden.com/t/inline-autofill-menu-resizing-filtering-and-customization/
- https://community.bitwarden.com/t/inline-autofill-menu-ability-to-filter-items-by-typing-searching/


audreyality:
Hey Gamerhund!

👍🏻 Looks like this is your first contribution to Bitwarden! Welcome!

❌ It appears that your code fails to compile and breaks several unit tests. I'm going to switch this PR to draft so that you can address those issues. 

🎨 Also be sure to run `npm run prettier` in the root of the repository so that your code passes our formatting and lint checks!

Gamerhund:
Hi @audreyality, thanks for the feedback! I've addressed both issues:

Fixed the failing unit tests to reflect the new filterInlineMenuCiphers behavior
Removed the unused AutofillOverlayElement import so ESLint passes

Ready for another look when you have time!

audreyality:
⚠️ You still have failing builds because your code fails typechecking.
⛏️ Lints are failing because you need to run prettier to format your code.

Gamerhund:
Hi @audreyality, I've just pushed fixes for both issues:

✅ TypeScript errors fixed - added `filterValue` to [OverlayBackgroundExtensionMessage]
✅ Prettier formatting applied

The build should pass now. Could you please re-trigger the checks or take another look? Thanks!

Gamerhund:
Hi @audreyality, also resolved the merge conflict with main. The PR should be fully ready now!

package.json
```diff
@@ -127,6 +127,7 @@
     "husky": "9.1.7",
     "jest": "30.3.0",
     "jest-diff": "30.3.0",
+    "jest-environment-jsdom": "30.4.0",
```
audreyality: ⚠️ We can't accept a new package in this PR. You should be able to build without this dependency by using Node v22. If you have a more recent version of node installed, consider using a [node version manager](https://docs.npmjs.com/downloading-and-installing-node-js-and-npm#using-a-node-version-manager-to-install-nodejs-and-npm) to switch back to Node v22.

Gamerhund: Thanks for the feedback! I've removed the jest-environment-jsdom dependency by reverting the package.json and package-lock.json changes. I'm now using Node v22 as suggested.

audreyality:
Thank you @Gamerhund!

I've kicked off the next batch of builds! Once they complete successfully, I'll forward this to our internal teams for design and product review. There's some excitement among the team for this work! 🚀

Gamerhund:
Thank you so much!

audreyality:
🤔 Looks like you're still failing typechecks. You can observe them locally using `npm run test:types`.

Gamerhund:
Fixed the TypeScript errors! Added proper null checks for sender.tab. The tests are now passing. Thanks for the feedback! 🙏

chiefbrain:
Is it possible to help here? Or should the internal ticket https://bitwarden.atlassian.net/browse/PM-36868 be awaited?

Gamerhund:
Hi @chiefbrain! Thanks for offering to help. There's also a parallel implementation in #20836 that references this PR and has adopted the same wire-protocol naming. I'm happy to defer to whichever approach the Bitwarden team prefers to move forward with. If you or anyone wants to pick up the conflict resolution here, feel free!

chiefbrain:
I also started implementing it before any pull request was there and finished it yesterday just to see that you have been faster :smiley:

The main difference in my PR:
 - it avoids rebuilding on every keystroke
 - it normalizes the search value further to match more items
 
 I would be happy with your implementation too, leaving this for reference.

Gamerhund:
Ha, great minds think alike! 😄 Really impressive that you built it independently too. The Bitwarden team now has three implementations to choose from — yours, #20836, and this one. Whatever gets merged, the feature will finally be there. Thanks for the kind words!

audreyality:
Hi @Gamerhund,

Gamerhund:
Thanks for the update! Totally understood — appreciate the transparency. Happy to have contributed to the discussion around this feature regardless of outcome. Feel free to close. 👍

PR #20544, opened 2026-05-07
[PM-35357] Update Trial Length Parameter

sbrown-livefront:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-35357

## 📔 Objective

Adds trial length and trailInitiationId throughout the billing components. It ensures that the trial duration is consistently calculated and displayed, and that relevant identifiers are passed through service layers and API requests. The changes also update UI messaging to reflect the specific trial length and add comprehensive tests for the new logic.

- Introduced a `DEFAULT_TRIAL_LENGTH_DAYS` constant

**UI and messaging improvements:**

- Updated UI templates to use new i18n keys that include the specific trial length in their messaging, ensuring users see accurate trial information. 

**Testing:**

- Added a new unit test suite for `ConfirmationDetailsComponent` to verify correct rendering and argument passing for trial length-dependent messages.
## 📸 Screenshots

https://github.com/user-attachments/assets/37ad988b-105c-48e8-882b-9192bdb9389b


amorask-bitwarden (dismissed):
Looks good. Only ❓ is are the following translation keys used anywhere else? If not, we should clean them up:
- `subscribeEnterpriseSubtitle`
- `trialPaidInfoMessage`

amorask-bitwarden (approved):

PR #20545, opened 2026-05-07
[PM-31945] Fix missing group access to collections in Access Intelligence V2

lastbestdev:
## 🎟️ Tracking
https://bitwarden.atlassian.net/browse/PM-31945

## 📔 Objective
Access Intelligence V2's client side report generation had a bug where group access to collections was not fetched from anywhere, resulting in the collection access used during report generation to only account for direct user access to collections. This caused incorrect member counts in the Access Intelligence report, since group access to ciphers was missing/not used in the calculations.

These changes update the client to fetch all collections with details for the Org (includes both user and group access to the collections), which allows the report to be properly generated using lookups for both direct user access and group access to collections. 

## 📸 Screenshots
**Before changes - groups set in collection access objects always empty**
image
 

**After changes - groups set in collection access objects populated properly**
image


Banrion:
Just commenting that type checking is failing

Banrion:
@claude Please add a review

bitwarden_license/bit-common/src/dirt/access-intelligence/services/implementations/view/default-access-intelligence-data.service.ts
Banrion: 🎨 Since we are manually fetching collections for access info, we can remove the `includeCollections: true` from the user fetch

Banrion (dismissed):

Banrion (approved):
✅ Code looks good, but there are failing checks

PR #20549, opened 2026-05-08
[PM-35378]Remove stale iconServerUrl

dan-livefront:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-35378
## 📔 Objective

Removed the stale iconsServerUrl field. Instead, `getInlineMenuCipherData()` now resolves the active account’s environment fresh on each call using `firstValueFrom(environmentService.environment$)`, which is already scoped to the active user. Applied the same fix in NotificationBackground as well.


apps/browser/src/autofill/background/overlay.background.ts
```diff
@@ -640,13 +646,14 @@ export class OverlayBackground implements OverlayBackgroundInterface {
...
     showFavicons: boolean,
+    iconsServerUrl: string | null,
   ) {
     const inlineMenuCipherData: InlineMenuCipherData[] = [];
     const passkeyCipherData: InlineMenuCipherData[] = [];
     const domainExclusions = await this.getExcludedDomains();
     let domainExclusionsSet: Set<string> | null = null;
     if (domainExclusions) {
-      domainExclusionsSet = new Set(Object.keys(await this.getExcludedDomains()));
+      domainExclusionsSet = new Set(Object.keys(domainExclusions));
```
dan-livefront, line 656: we're getting domainExclusions already, not sure why the double call, removed.

apps/browser/src/autofill/background/notification.background.ts
```diff
@@ -247,7 +247,7 @@ export default class NotificationBackground {
       firstValueFrom(this.organizationService.organizations$(activeUserId)),
     ]);
 
-    const iconsServerUrl = env.getIconsUrl();
+    const iconsServerUrl: string | null = env.getIconsUrl() ?? null;
```
dan-livefront, line 250: I originally considered putting `const iconsServerUrl: string | null = env.getIconsUrl()?.replace(/\/+$/, "") ?? null;` in order to drop the trailing slash in icon urls. This is being normalized downstream but I would be open to bringing this back for hygiene.

blackwood (approved):

PR #20550, opened 2026-05-08
[PM-36617] Enhance PoliciesComponent: Add focus restoration for active element

JaredScar:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-36617

## 📔 Objective

Adds keyboard support better. Focus restoration for active element

### DIFF
--- a/apps/web/src/app/admin-console/organizations/policies/policies.component.ts
+++ b/apps/web/src/app/admin-console/organizations/policies/policies.component.ts
@@ -1,7 +1,7 @@
 import { ChangeDetectionStrategy, Component, DestroyRef, signal } from "@angular/core";
 import { takeUntilDestroyed } from "@angular/core/rxjs-interop";
 import { ActivatedRoute } from "@angular/router";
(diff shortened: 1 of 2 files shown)


apps/web/src/app/admin-console/organizations/policies/policies.component.ts
```diff
@@ -181,6 +184,11 @@ export class PoliciesComponent {
       });
       if (ref !== undefined) {
         this.drawerRef.set(ref);
+        ref.closed.pipe(take(1), takeUntilDestroyed(this.destroyRef)).subscribe(() => {
+          if (triggerEl?.isConnected) {
+            triggerEl.focus();
+          }
+        });
```
BTreston: `await lastValueFrom()` is probably a better option here.

BTreston (commented):
There are failing tests. I think this solution is fine, but this feels like a design smell for our CL components. I'd potentially raise this with UIF if drawers mess with focus. We may want to build focus handling directly into the component rather than attempting to store and refocus in each caller that may open a drawer in the future.

BTreston (approved):

PR #20551, opened 2026-05-08
[PM-36616] Fix fido2 script injection not respecting blocked domains

jprusik:
## 🎟️ Tracking

[PM-36616](https://bitwarden.atlassian.net/browse/PM-36616)

## 📔 Objective

Fix fido2 script injection not respecting blocked domains

[PM-36616]: https://bitwarden.atlassian.net/browse/PM-36616?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/common/src/platform/services/fido2/fido2-client.service.ts
+++ b/libs/common/src/platform/services/fido2/fido2-client.service.ts
@@ -93,11 +93,7 @@ export class Fido2ClientService<
     const blockedInteractionsUris = await firstValueFrom(
       this.domainSettingsService.blockedInteractionsUris$,
     );
-    const isBlockedDomain =
-      blockedInteractionsUris != null &&
-      Object.keys(blockedInteractionsUris).some((blockedHostname) =>
-        hostname.endsWith(blockedHostname),
-      );
+    const isBlockedDomain = blockedInteractionsUris != null && hostname in blockedInteractionsUris;
     if (isBlockedDomain) {
       return false;
     }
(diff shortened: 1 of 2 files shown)


libs/common/src/platform/services/fido2/fido2-client.service.ts
djsmith85, line 106: ❓ Is there are reason why the check against the hostname is different for `isBlockedDomain` and  `neverDomains`?

If not:
🎨  As `blockedInteractionsUris$` and `neverDomains$` are both of type `NeverDomain`, couldn't these checks be combined?

jprusik: `NeverDomain` is the type/data shape; it's expressed in excluded domains (notification blocking) and blocked domains (injection blocking). We have plans to combine the concepts, but it's lower priority tech debt at the moment.

The checks here are separate in order to potentially exit early and avoid an extra evaluation 🤷

jprusik: There's additional work coming to further reinforce injection blocking; this is just a scoped fix to bring blocked domains up to parity with excluded domains.

djsmith85: Thanks for the updated changes, using the same comparison for the hostnames and calling out future work to combine these blocked/excluded domains.

djsmith85 (approved):

PR #20556, opened 2026-05-08
Add autofill check for autocomplete="current-password"

differsthecat:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-36852 

## 📔 Objective

Add a method to check for the `autocomplete="current-password"` identifier.

### DIFF
--- a/apps/browser/src/autofill/services/autofill-overlay-content.service.ts
+++ b/apps/browser/src/autofill/services/autofill-overlay-content.service.ts
@@ -1213,6 +1213,12 @@ export class AutofillOverlayContentService implements AutofillOverlayContentServ
       autofillFieldData.autoCompleteType || ([] as string[])
     ).includes("webauthn");
 
+    if (
+      this.inlineMenuFieldQualificationService.hasCurrentPasswordAutocomplete(autofillFieldData)
+    ) {
+      return;
+    }
+
     this.qualifyAccountCreationFieldType(autofillFieldData);
   }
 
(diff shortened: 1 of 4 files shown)


apps/browser/src/autofill/services/inline-menu-field-qualification.service.spec.ts
```diff
@@ -1037,4 +1037,33 @@ describe("InlineMenuFieldQualificationService", () => {
...
+    });
+
+    it("returns false when the autocomplete attribute is missing", () => {
+      const field = mock<AutofillField>({
+        type: "password",
+        autoCompleteType: "",
+      });
+
+      expect(inlineMenuFieldQualificationService.hasCurrentPasswordAutocomplete(field)).toBe(false);
+    });
```
jprusik, line 1067: Non-blocking nit; probably wouldn't be a bad idea to ensure we handle non-standard cases like "password" and standard but irrelevant values like "[one-time-code](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Attributes/autocomplete#one-time-code)" and "[off](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Attributes/autocomplete#off)"

If we want to be really thorough, space-delimited [token lists](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Attributes/autocomplete#token-list), but that's probably overkill.

apps/browser/src/autofill/services/autofill-overlay-content.service.ts
```diff
@@ -1213,6 +1213,12 @@ export class AutofillOverlayContentService implements AutofillOverlayContentServ
       autofillFieldData.autoCompleteType || ([] as string[])
     ).includes("webauthn");
 
+    if (
+      this.inlineMenuFieldQualificationService.hasCurrentPasswordAutocomplete(autofillFieldData)
+    ) {
+      return;
+    }
```
jprusik, line 1220: nit - given that we're depending on this happening after the login type assignment to `inlineMenuFillType` and before any other assignment, it might be a good idea to drop in a comment with the intent of the check and early return.

jprusik (approved):
Some non blocking nits, but a really helpful add to keep the experience consistent; thank you!

PR #20560, opened 2026-05-08
[PM-36565] creating new item from new dialog is not working

JaredScar:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-36565

## 📔 Objective

Fixes up codebase to support Rust changes incoming


libs/common/src/admin-console/models/collections/collection-sdk-mapping.spec.ts
```diff
@@ -32,7 +32,7 @@ function makeSdkCollection(overrides: Partial<SdkCollection> = {}): SdkCollectio
     readOnly: false,
     manage: true,
     defaultUserCollectionEmail: undefined,
-    type: "SharedCollection",
+    type: 0, // SdkCollectionType.SharedCollection
```
eliykat: You should be able to use the enum directly throughout this file - TS treats a numeric enum and a number as equivalent.

eliykat (changes requested):
Looks good, please also include the SDK version bump in this PR after your SDK PR has merged.

JaredScar:
> Looks good, please also include the SDK version bump in this PR after your SDK PR has merged.

Version updated!

eliykat (approved):

PR #20498, opened 2026-05-04
[PM-35228] Add Premium Status Changed Push Notification

sbrown-livefront:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-35228

## 📔 Objective

This pull request introduces support for handling premium status change notifications across the application. The main changes add a new `PremiumStatusChanged` notification type, update the notification processing logic to handle this new type, and ensure that the premium status is correctly updated in the user's billing profile state. 

**Notification Type and Model Additions:**

* Added a new `NotificationType.PremiumStatusChanged` enum value and corresponding `PremiumStatusChangedNotification` model to represent premium status change events. [[1]](diffhunk://#diff-75b3c01a1a01675b8b11957902ab8732c4f6654577d85bda78daa7d9d4a6a189R39-R40) [[2]](diffhunk://#diff-0bc93892be83ed14fde331c43695c9e63c936772ae8a700422746dd917566981R234-R244)
* Updated the `NotificationResponse` class to handle deserialization of the new notification payload.

**Notification Handling Logic:**

* Updated `DefaultServerNotificationsService` to inject and use `BillingAccountProfileStateService` for processing `PremiumStatusChanged` notifications, updating the user's premium status accordingly. 

## Trigger a `PremiumStatusChanged` in Browser Extension
https://github.com/user-attachments/assets/535fbfc5-6e54-497f-b030-07f8f4bf388d

Issue #7550, opened 2024-01-15
Cancel does nothing in pop out window when Adding a Send


corey851:
### Steps To Reproduce

In the browser extension Click on the 'Send' Tab.  
Click Add Send.
Click on the top banner to open a new pop out window. 
Now suppose you have changed your mind and click on Cancel in the upper left.
Nothing happens.

### Expected Result

the New Send window should close returning you to previous view

### Actual Result

Nothing happens.

Workaround: you can simply close the pop out completely by clicking the window close button ("X") 

### Operating System

Linux

### Operating System Version

Debian 12 

### Web Browser

Chrome, Firefox

### Browser Version

Firefox 115.5.0esr;   Chromium 120.0.6099.109

### Build Version

vers: 2023.12.1     server vers: 2024.1.0

### Issue Tracking Info

- [X] I understand that work is tracked outside of Github. A PR will be linked to this issue should one be opened to address it, but Bitwarden doesn't use fields like "assigned", "milestone", or "project" to track progress.

Krychaz:
Hello there, 

Unfortunately we were not able to replicate this behaviour. Will you try uninstalling the application, removing any leftover data?

Guide to the leftover data: https://bitwarden.com/help/data-storage/#on-your-local-machine

corey851:
I uninstalled the browser extension, after which any leftover data (browser extension directory) is gone anyway.  I reinstalled, re-logged in and the behavior is the same.  Thinking it might be Linux related I even tested on a Windows machine and got the same behavior.  

Sometimes I'm not clear so let me try to explain better:

1. Click on the Bitwarden browser extension icon.
2. Click the Send tab with the paper airplane icon.
3. Click Add Send.
(If you click Cancel here, in the upper left corner, the 'New Send' window disappears and you are returned to the previous window demonstrating what I assume is the expected behavior. 
But to see what I'm doing don't cancel, continue with next step.)
4. Click the banner at the top (where it says "before you start") to open up a Bitwarden pop out window. 
Now, the Cancel button in the pop out window is inoperable.  

Not a big deal, you can just close the pop out window itself.  But you might want the Cancel button to just work.

djsmith85:
Hey @corey851,

thanks for providing more details around your issue with the cancel button. I can't say what caused the initial issue, but since  your report the UI has been overhauled and the cancel-button is no longer present within the UI.

As this is not reproducible and likely no longer and issue, I'm going to go ahead and close this.

Thanks again for your report.

[...]

PR #20592, opened 2026-05-11
[PM 35229](clients) [Browser/Desktop] Stripe Checkout from upgrade dialog

cyprain-okeke:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-35229

## 📔 Objective

When a non-premium user clicks Upgrade Now in the PremiumUpgradeDialogComponent on the browser extension or desktop
  app, the dialog now calls the new POST /account/billing/vnext/premium/checkout endpoint to obtain a Stripe Checkout
  session URL and opens it directly via platformUtilsService.launchUri(), rather than redirecting to the web vault
  subscription page.

  This puts the entire premium-upgrade flow in the user's current client context — Stripe Checkout opens in a new
  browser tab (extension) or in the system's default browser (desktop) — and removes the round-trip through the web
  vault for the new dialog path introduced under PM23713_PremiumBadgeOpensNewPremiumUpgradeDialog.

  The change is gated behind a new feature flag, pm-34515-browser-desktop-checkout, and only takes effect when all of
  the following are true:

  - the user is in a cloud environment (environment.isCloud()),
  - the client is the browser extension or desktop app (ClientType.Browser or ClientType.Desktop),
  - the new flag is enabled.

  When any of those preconditions are not met — flag off, self-host, or unsupported client type — the dialog continues
  to open the existing web vault subscription URL, preserving today's behavior for self-host customers and any code path
   that does not yet target this flow.

  What's added on the client:

  - createPremiumCheckoutSession method on BillingApiServiceAbstraction and BillingApiService, posting { platform: 
  "browser" | "desktop" } and returning a PremiumCheckoutSessionResponse containing the Stripe Checkout URL.
  - PremiumCheckoutSessionRequest / PremiumCheckoutSessionResponse models under libs/common/src/billing/models/.
  - PM34515_BrowserDesktopCheckout entry in the FeatureFlag enum and DefaultFeatureFlagValue map under the Billing team
  section.
  - Loading-state handling on the upgrade button ([loading]="upgrading()") and a re-entrant click guard so the API call
  cannot be triggered twice.

  Dependencies (call out in your PR description):

  - Server PR bitwarden/server#7550 — [PM-35227] Extend checkout endpoint for browser/desktop platforms. Adds browser
  and desktop to the allowed Platform values and resolves new
  GlobalSettings.Stripe.{Browser,Desktop}PremiumCheckout{Success,Cancel}Url settings per platform. This clients PR
  cannot ship to production until #7550 is merged and deployed.
  - Sibling clients PR bitwarden/clients#20498 — [PM-35228] Add Premium Status Changed Push Notification. Required for
  the acceptance criterion that the app's premium state updates reactively after the user completes checkout (handles
  NotificationType.PremiumStatusChanged = 27 by calling BillingAccountProfileStateService.setHasPremium(...), no full
  sync needed). Without #20498 the user must manually sync or reopen the app to see premium-gated UI unlock; this PR's
  flow itself works without it.
## 📸 Screenshots

https://github.com/user-attachments/assets/ca46f78a-555b-4525-9a29-c41e2d0cfd2b

https://github.com/user-attachments/assets/a70c7153-7ec8-4a6e-a442-c29f086e5cca

[PM-35227]: https://bitwarden.atlassian.net/browse/PM-35227?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


amorask-bitwarden (dismissed):

libs/angular/src/billing/components/premium-upgrade-dialog/premium-upgrade-dialog.component.ts
```diff
@@ -66,15 +73,56 @@ export class PremiumUpgradeDialogComponent {
...
+
+    try {
+      const environment = await firstValueFrom(this.environmentService.environment$);
+      const checkoutFlagEnabled = await this.configService.getFeatureFlag(
+        FeatureFlag.PM34515_BrowserDesktopCheckout,
+      );
+      const platform = this.resolveCheckoutPlatform();
+
+      if (checkoutFlagEnabled && environment.isCloud() && platform != null) {
+        const { checkoutSessionUrl } = await this.billingApiService.createPremiumCheckoutSession({
```
sbrown-livefront, line 103: ❓ Do we have a matching server PR to add BrowserExtension as a valid platform? https://github.com/bitwarden/server/blob/3c2cc45215a128f35a0328938c072cdaf2d39886/src/Api/Billing/Models/Requests/Premium/CreatePremiumCheckoutSessionRequest.cs#L6

sbrown-livefront: NVM just saw it

amorask-bitwarden (approved):

PR #19216, opened 2026-02-25
[PM-32698] feat(vault): add copy button for cardholder name in card detail view

Sebastianhayashi:
## 🎟️ Tracking

Resolves https://github.com/bitwarden/clients/issues/12439

## 📔 Objective

*Note: This contribution was developed with the assistance of the Antigravity AI agent and has been thoroughly reviewed, locally tested, and manually verified by Sebastian Lim.*

**What Was Done**
This PR fixes a bug in the Bitwarden web vault where the **Cardholder Name** field in the Card detail view was missing a copy-to-clipboard button, while the **Number** and **Security Code** fields both had one.

**The Fix**
The pattern exactly matches the existing copy buttons on the **Number** and **Security Code** fields to maintain UI consistency. Added a `<button>` with `bitIconButton="bwi-clone"` and `[appCopyClick]` directive to the cardholder name `<bit-form-field>` in [card-details-view.component.html](file:///Users/microseyuyu/repo_rank/clients/libs/vault/src/cipher-view/card-details/card-details-view.component.html).

**Verification**
1. Cloned the `bitwarden/clients` repo and pointed [development.json](file:///Users/microseyuyu/workspace/clients/apps/web/config/development.json) to the production cloud APIs.
2. Reproduced the issue locally: no copy button on the cardholder name field.
3. Applied the fix and confirmed through Webpack hot-reloading that the copy button correctly renders and functions.

### Before

card_detail_view_no_copy_button_1771988253640

### After

card_fix_verified_1771988414975

PR #20602, opened 2026-05-11
[PM-32698] Add copy button for cardholder name in card detail view

gbubemismith:
## 🎟️ Tracking
https://bitwarden.atlassian.net/browse/PM-32698

## 📔 Objective
QA build for https://github.com/bitwarden/clients/pull/19216

### DIFF
--- a/apps/desktop/src/locales/en/messages.json
+++ b/apps/desktop/src/locales/en/messages.json
@@ -1755,6 +1755,10 @@
     "message": "Copy number",
     "description": "Copy credit card number"
   },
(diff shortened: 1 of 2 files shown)


nick-livefront:
@gbubemismith I'm assuming you don't need a review here because it is only for a QA build. Let me know if that is not the case!

gbubemismith:
> @gbubemismith I'm assuming you don't need a review here because it is only for a QA build. Let me know if that is not the case!

Yeah no review needed here. Thanks

PR #20603, opened 2026-05-11
[PM-30365] Account recovery salt fallback

ike-kottlowski:
## 🎟️ Tracking

[PM-30365](https://bitwarden.atlassian.net/browse/PM-30365)

## 📔 Objective

When an admin initiates account recovery and resets a user's password we want to ensure the salt being used for the key derivation uses the `user.MasterPasswordSalt` that exists in the user entity on the server, saved from user registration. If the salt is `null`, which is the case for legacy users, then we fall back to the `user.email` for use as the key derivation salt.

### Admin recovers account

https://github.com/user-attachments/assets/ca870ad9-112a-40b4-b226-4d1880758285

NOTE: If the master password salt is null in the database, then the server will fall back to email using the extension method on the User object. In the video there is manual modification of the `OrganizationUserResetPasswordDetailsResponseModel` to forcibly set the `MasterPasswordSalt` to `null`. The remainder of the video shows a successful master password recovery by the user that had their account recovered.

[PM-30365]: https://bitwarden.atlassian.net/browse/PM-30365?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/web/src/app/admin-console/organizations/members/services/organization-user-reset-password/organization-user-reset-password.service.ts
```diff
@@ -233,9 +245,21 @@ export class OrganizationUserResetPasswordService implements UserKeyRotationKeyR
     kdfMemory?: number;
     kdfParallelism?: number;
   }): KdfConfig {
-    return response.kdf === KdfType.PBKDF2_SHA256
```
ike-kottlowski, line 236: This change was done to satisfy the strict requirements.

This allows us to have better type safety by validating the `kdfMemory` and `kdfParallelism` are not null

apps/web/src/app/admin-console/organizations/members/services/organization-user-reset-password/organization-user-reset-password.service.ts
```diff
@@ -268,17 +292,10 @@ export class OrganizationUserResetPasswordService implements UserKeyRotationKeyR
 
   private async buildResetPasswordRequest(
     newMasterPassword: string,
-    email: string,
+    salt: MasterPasswordSalt,
     kdfConfig: KdfConfig,
     existingUserKey: UserKey,
   ): Promise<Pick<OrganizationUserResetPasswordRequest, "newMasterPasswordHash" | "key">> {
-    // In the Account Recovery flow, the target user's UserId is not available (only orgUserId),
```
ike-kottlowski, line 275: this change is abstracted out above on ln 136. we prefer the server provided salt, if null then we fall back to email.

libs/admin-console/src/common/organization-user/models/requests/organization-user-reset-password-enrollment.request.ts
```diff
@@ -1,11 +1,11 @@
...
-// @ts-strict-ignore
 import { SecretVerificationRequest } from "@bitwarden/common/auth/models/request/secret-verification.request";
 
 export class OrganizationUserResetPasswordEnrollmentRequest extends SecretVerificationRequest {
-  resetPasswordKey: string;
+  resetPasswordKey?: string;
 }
 
 export class OrganizationUserResetPasswordWithIdRequest extends OrganizationUserResetPasswordEnrollmentRequest {
-  organizationId: string;
```
ike-kottlowski, line 10: Also done to support strict typing.

Making the consuming class strict meant we needed to ensure that the `organizationId` was not null.

ike-kottlowski:
@claude please review this PR. We have established for this flow that an empty `MasterPasswordSalt`, meaning `user.MasterPasswordSalt = ""` is not probable even if it is possible. There are protections on the server side to ensure this doesn't happen. Please be sure the strict-type enablement is correct.

apps/web/src/app/admin-console/organizations/members/services/organization-user-reset-password/organization-user-reset-password.service.ts
enmande: 🎨  This looks good. Per [PM-32059](https://bitwarden.atlassian.net/browse/PM-32059) (1.a.4), we will need to unwind email fallback at that time. Should we leave issue number in comments at :136 so it's an easier find?

apps/web/src/app/admin-console/organizations/members/services/organization-user-reset-password/organization-user-reset-password.service.ts
```diff
@@ -125,10 +133,15 @@ export class OrganizationUserResetPasswordService implements UserKeyRotationKeyR
         resetPasswordDetails,
         request.organizationId,
       );
+      // Prefer server-provided salt; fallback to normalized email for legacy servers
+      // that don't yet return MasterPasswordSalt in the account-recovery details response.
+      const salt: MasterPasswordSalt =
+        (resetPasswordDetails.masterPasswordSalt as MasterPasswordSalt | undefined) ??
```
enmande: 🎨 ⛏️  The `EmergencyAccessService` ([main](https://github.com/bitwarden/clients/blob/6f3d3239c29966574b86003cb01817051103139a/apps/web/src/app/auth/emergency-access/services/emergency-access.service.ts#L326)) implements this with a `typeof` guard.

Realistically, these do the same thing unless the server were to return a non-string truthy value (e.g., `0`) -- and it shouldn't under _any_ conceivable circumstance. `typeof` was just chosen to serve as a runtime catch-out because the `getResponseProperty` will hand out an `any`. So there's a strong argument that it's not necessary, I'm just identifying it as a question of "should we ensure these are consistent where we can?"

enmande (dismissed):
Thank you ✨ 
Excellent, expressive tests, and appreciate scrict-ifying where we could.
Two non-blocking nitpicks below.

apps/web/src/app/admin-console/organizations/members/services/organization-user-reset-password/organization-user-reset-password.service.ts
ike-kottlowski: https://github.com/bitwarden/clients/pull/20603/commits/06ec7660ef1788d7eb08766d9d72f4c595b72132

Added in this commit. Good catch I would prefer consistency between implementaitons.

apps/web/src/app/admin-console/organizations/members/services/organization-user-reset-password/organization-user-reset-password.service.ts
ike-kottlowski: https://github.com/bitwarden/clients/pull/20603/commits/05f6442be9ac78980c7f4de80636b2453236a260

added comment in this commit calling out the cleanup ticket for the future generations.

apps/web/src/app/admin-console/organizations/members/services/organization-user-reset-password/organization-user-reset-password.service.ts
enmande: Thank you! I will wear my own cone of shame for the _details_ of my claim above though -- `0` wouldn't pass that truthy sniff-check because zero (`!!0 == false`), but e.g., `1` would (`!!1 == true`). Point stands, shabby example given, thank you for the update ✨

apps/web/src/app/admin-console/organizations/members/services/organization-user-reset-password/organization-user-reset-password.service.ts
```diff
@@ -125,10 +133,17 @@ export class OrganizationUserResetPasswordService implements UserKeyRotationKeyR
         resetPasswordDetails,
         request.organizationId,
       );
+      // Prefer server-provided salt; fallback to normalized email for legacy servers
+      // that don't yet return MasterPasswordSalt in the account-recovery details response.
+      // TODO: PM-32059 — When salt is disconnected from email (Stage 3) we will no longer fall back to email.
+      const salt: MasterPasswordSalt =
+        typeof resetPasswordDetails.masterPasswordSalt === "string"
+          ? (resetPasswordDetails.masterPasswordSalt as MasterPasswordSalt)
+          : this.masterPasswordService.emailToSalt(email);
```
eliykat, line 142: This works but you'll probably be repeating yourself everywhere - what makes a valid salt, and the conversion to the opaque type, is not well encapsulated.

There are multiple ways of doing this, but the simplest is probably using a type guard:

```ts
      // This would live somewhere else for re-use
      const isValidSalt = (maybeSalt: string | undefined): maybeSalt is MasterPasswordSalt =>
        typeof maybeSalt === "string" && !Utils.isNullOrWhitespace(maybeSalt);

      const salt: MasterPasswordSalt = isValidSalt(resetPasswordDetails.masterPasswordSalt)
        ? resetPasswordDetails.masterPasswordSalt
        : this.masterPasswordService.emailToSalt(email);
```

eliykat (dismissed):
Comment is non-blocking.

apps/web/src/app/admin-console/organizations/members/services/organization-user-reset-password/organization-user-reset-password.service.ts
ike-kottlowski: This is a great idea but to keep the code owners slim I'm going to opt to leave it out. 

Also, this is temporary with work planned to remove the fallback in the near future.

Thanks for the feedback though!

eliykat (approved):

JaredSnider-Bitwarden (approved):

PR #20606, opened 2026-05-11
[PM-31084] - migrate badge to chip-action in vault components

jaasen-livefront:
## 🎟️ Tracking

[PM-31084](https://bitwarden.atlassian.net/browse/PM-31084)

## 📔 Objective

Migrates usages of the deprecated `Badge` component to the `bit-chip-action` component across the browser extension, desktop app, and vault library.

No behavioral changes — this is a like-for-like component swap. Unused `Badge` imports are removed where applicable.

**Files updated:**
- `apps/browser` — `add-edit`, `open-attachments`, and `view` components in the vault popup
- `apps/desktop` — `vault-orig` component
- `libs/vault` — `item-details-section`, `item-details-v2`, `login-credentials-view`, and `vault-item-dialog` components

## 📸 Screenshots

N/A — no visual changes intended; `Chip` renders identically to `Badge` for these usages.

[PM-31084]: https://bitwarden.atlassian.net/browse/PM-31084?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/browser/src/vault/popup/components/vault/add-edit/add-edit.component.html
```diff
@@ -9,9 +9,7 @@
   >
     @if (config?.originalCipher?.archivedDate) {
       <ng-container slot="end">
-        <span bitBadge variant="secondary" [appA11yTitle]="'archived' | i18n">
-          {{ "archived" | i18n }}
-        </span>
+        <button bit-chip-action [label]="'archived' | i18n"></button>
```
jengstrom-bw: I'm not sure if these lint errors are legit, since the component instructions say to do exactly this. Since none of these need to be a button, `<bit-chip>` could be appropriate instead.

jaasen-livefront: @jengstrom-bw The linter is in fact correct. Since we're using the `bit-chip-action` (not to be confused with `bit-chip`) it requires either an `a` or `button` tag. I went with `button` so the requisite `type="button"` is required. I've updated the PR description to be more clear on this as well.

jengstrom-bw (dismissed):

apps/desktop/src/vault/app/vault-v3/vault-orig.component.ts
jengstrom-bw, line 1: was there a rename of this file and if so is this one still needed?

jaasen-livefront: @jengstrom-bw Ah that was removed on main and snuck in. I've removed it now.

jengstrom-bw (approved):

PR #20608, opened 2026-05-11
Remove unused strings

mKoonrad:
## 📔 Objective

While reviewing the translations in the project, I found some unused strings.


djsmith85 (approved):
@mKoonrad Thanks again for your contributions !!

I have verified that these strings are no longer used anywhere and can be removed.

PR #20610, opened 2026-05-11
[PM-36878] Update Existing User Trail Flow

sbrown-livefront:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-36878

## 📔 Objective

Standardizes how trial lengths are handled and displayed across the billing and organization creation flows. The changes update user-facing messages and tests to reflect the specific trial length in use.

## 📸 Screenshots

https://github.com/user-attachments/assets/09c0d163-5baf-446d-8501-bcb0aa6bd88e

### NOTE: Free no longer shows the trial period


apps/web/src/app/billing/organizations/organization-plans.component.html
```diff
@@ -47,6 +47,7 @@ <h2 bitTypography="h2">{{ "chooseYourPlan" | i18n }}</h2>
           @let hasDirectory = selectableProduct.hasDirectory;
           @let usersGetPremium = selectableProduct.usersGetPremium;
           @let trialPeriodDays = selectableProduct.trialPeriodDays;
+          @let effectiveTrial = trialLength() ?? trialPeriodDays;
```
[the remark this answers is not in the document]
sbrown-livefront: Addressed in https://github.com/bitwarden/clients/pull/20610/commits/2e0d543fc81ece566d663d5409e958b3a11cea21

apps/web/src/app/admin-console/settings/create-organization.component.ts
```diff
@@ -49,6 +50,8 @@ export class CreateOrganizationComponent implements OnInit, OnDestroy {
       }
 
       this.secretsManager = qParams.product == ProductType.SecretsManager;
+
+      this.trialLength = qParams.trialLength ? parseInt(qParams.trialLength) : null;
```
jrmccannon: We should probably switch this to be `undefined`.

https://contributing.bitwarden.com/architecture/adr/typescript-strict/#avoid-null-prefer-undefined

apps/web/src/app/billing/organizations/organization-plans.component.ts
```diff
@@ -126,6 +127,9 @@ export class OrganizationPlansComponent implements OnInit, OnDestroy {
    */
   readonly initialPlan = input<PlanType>(PlanType.Free);
 
+  /** Custom trial length from the URL, overrides the plan's default trialPeriodDays for display and API calls. */
+  readonly trialLength = input<number | null>(null);
```
jrmccannon: Update here to avoid null
https://contributing.bitwarden.com/architecture/adr/typescript-strict/#avoid-null-prefer-undefined

apps/web/src/app/billing/organizations/organization-plans.component.ts
```diff
@@ -1073,6 +1080,10 @@ export class OrganizationPlansComponent implements OnInit, OnDestroy {
       request.coupons = this.eligibleCouponIds();
     }
 
+    if (this.trialLength() != null) {
```
jrmccannon: Update here to avoid null

jrmccannon (changes requested):

apps/web/src/app/admin-console/settings/create-organization.component.ts
sbrown-livefront: Great callout. Thanks for referencing the ADR! Addressed in https://github.com/bitwarden/clients/pull/20610/commits/0a8cc1bb84f822d84c1c9e1468811511611c2e22

apps/web/src/app/billing/organizations/organization-plans.component.ts
sbrown-livefront: Addressed in https://github.com/bitwarden/clients/pull/20610/commits/0a8cc1bb84f822d84c1c9e1468811511611c2e22

apps/web/src/app/billing/organizations/organization-plans.component.ts
sbrown-livefront: Addressed in https://github.com/bitwarden/clients/pull/20610/commits/730f5d6cb56f639349c413886c6a2328eb80e86d. Checks against undefined to assure 0 can still get through.

apps/web/src/app/billing/organizations/organization-plans.component.ts
```diff
@@ -659,7 +665,10 @@ export class OrganizationPlansComponent implements OnInit, OnDestroy {
     if (this.acceptingSponsorship()) {
       return this.i18nService.t("paymentSponsored");
     } else if (this.freeTrial() && this.createOrganization() && !this.canUpgradeFromPremium()) {
-      return this.i18nService.t("paymentChargedWithTrial");
+      return this.i18nService.t(
+        "paymentChargedWithTrialSpecificLength",
+        this.trialLength() ?? DEFAULT_TRIAL_LENGTH_DAYS,
```
kdenney: I think this should be using the equivalent of this: `this.trialLength() ?? trialPeriodDays` instead of `?? DEFAULT_TRIAL_LENGTH_DAYS`. The rest of this component is taking the trialLength from the URL first and falling back to the selected plan's trialPeriodDays and not using the default at all so I believe this would show an incorrect value if any of the plans have a trial length !== 7.

You could pull the `effectiveTrial` from the template into this component as a computed and then use it in both places.

apps/web/src/app/billing/organizations/organization-plans.component.ts
```diff
@@ -1073,6 +1082,11 @@ export class OrganizationPlansComponent implements OnInit, OnDestroy {
       request.coupons = this.eligibleCouponIds();
     }
 
+    const trialLength = this.trialLength();
+    if (trialLength !== undefined) {
+      request.trialLength = trialLength;
+    }
```
kdenney, line 1088: I believe this also needs to change to use something like this.effectiveTrial, or does the server automatically use the plan's trial length if the request trialLength is not sent?

kdenney (changes requested):
Just a couple small changes needed.

apps/web/src/app/billing/organizations/organization-plans.component.ts
sbrown-livefront: Yeah, backend defaults to the plan's trialLength.

apps/web/src/app/billing/organizations/organization-plans.component.ts
sbrown-livefront: Good call. We have a default length if sent through the email(which is the same as most of the plan lengths) so I'm not sure if we're moving towards one default length or continuing with the plan lengths. I'll update this for now. 

EDIT: That trial length is calculated per plan in the html. I'll have the last fallback be the default if the custom trial is null or the selected plan trial length can't be found.

Thanks 🎉 !

jrmccannon (dismissed):

apps/web/src/app/billing/organizations/organization-plans.component.ts
sbrown-livefront: Addressed in https://github.com/bitwarden/clients/pull/20610/commits/a6de0b6703cb2e2c269c81f6e2a93af26f08a3ef

kdenney (dismissed):

kdenney (approved):

jrmccannon (approved):

PR #20625, opened 2026-05-12
[PM-36359] Include splunk integration

voommen-livefront:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-36359

## 📔 Objective

To setup Splunk as a possible HEC integration in the admin console

## 📸 Screenshots

Screenshot 2026-05-12 at 3 37 29 PM


bitwarden_license/bit-common/src/dirt/organization-integrations/models/integration-configuration-config/configuration-template/hec-template.ts
lastbestdev, line 60: is this a shared template? meaning these enrichment details would be used for other existing HEC event integrations as well?

lastbestdev (dismissed):
Assuming my only comment doesn't cause issues with adding data to events for other HEC integrations, this looks good. 

Also, kudos on how these integrations are set up! This PR shows how easily extendable integration setups are, which is a sign of good design IMO. 👏🏻

Banrion (dismissed):
✅ Approved, but also had the same question as Brad. 

❓ Does changing the template make those fields available to all hec integrations? Are there any situations where we would want a different template per integration?

bitwarden_license/bit-common/src/dirt/organization-integrations/models/integration-configuration-config/configuration-template/hec-template.ts
voommen-livefront: yes, all HEC integrations get the same template. My thinking goes like this - it wouldn't seem right for Splunk to get the enriched tags but Blumira or Huntress not get them. In the public API, all consumers get the same data shape, therefore, in this case too, it should be the same.

Banrion (dismissed):

Banrion (approved):

PR #20627, opened 2026-05-12
[PM-34156] Add accepted-state organization sync

eliykat:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-34156

## 📔 Objective

Wire the client to consume the new accepted-state organization sync data from the server.

- adds an `acceptedOrganizations$` stream for Accepted-state members
- keeps `organizations$` scoped to Confirmed-state members, so existing consumers see no behavior change.
- not feature-flagged at this layer — flagging happens at the consumer in follow-up tickets

Depends on https://github.com/bitwarden/server/pull/7627 which moves the response model property to the expected location.

## 📸 Screenshots

N/A — no UI changes.


coroiu (approved):

JaredScar (approved):
LGTM

PR #20629, opened 2026-05-13
[PM-37234] - fix launch button

jaasen-livefront:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-37234

## 📔 Objective

This PR fixes an issue where the launch button in the cipher row wasn't properly rendering because of a conflict with the `bitMenuItem` and `bitIconButton` directives.  Removing `bitIconButton` fixed the issue.

## 📸 Screenshots

Screenshot 2026-05-12 at 4 59 47 PM


apps/web/src/app/vault/components/vault-items/vault-cipher-row.component.html
```diff
@@ -167,13 +167,8 @@
...
             @if (canLaunch) {
-              <button
-                type="button"
-                bitMenuItem
-                (click)="openUri(launchUri)"
-                bitIconButton="bwi-external-link"
-                [label]="'launchWebsiteName' | i18n: cipher.name"
-              >
+              <button type="button" bitMenuItem (click)="openUri(launchUri)">
+                <i class="bwi bwi-fw bwi-external-link" aria-hidden="true"></i>
```
nick-livefront, line 171: nit: Can this be updated to use the `bit-icon` component rather than `i`?

nick-livefront (approved):

apps/web/src/app/vault/components/vault-items/vault-cipher-row.component.html
jaasen-livefront: @nick-livefront Normally I would totally do this but using `bit-icon` causes it to be misaligned with all the other items in the menu. It would be worth revisiting and refactoring all the `<i>`  elements in this component to `bit-icon`.

Image

nick-livefront: Interesting.. Thanks for looking into it!

PR #20636, opened 2026-05-13
[PM-31054] Add state service for V2UpgradeToken

Thomas-Avery:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-31054

## 📔 Objective

Adds a state service for the `V2UpgradeToken`. Use of this state service on the syncService will come in a following PR that will require multiple reviewers so trying to keep that one light.

See sync PR here for reference https://github.com/bitwarden/clients/pull/20641


libs/common/src/key-management/upgrade-token/services/default-v2-upgrade-token-state.service.ts
```diff
@@ -0,0 +1,34 @@
...
+  constructor(private readonly stateProvider: StateProvider) {}
+
+  v2UpgradeToken$(userId: UserId): Observable<V2UpgradeToken | null> {
+    assertNonNullish(userId, "userId");
+
+    return this.stateProvider.getUser(userId, V2_UPGRADE_TOKEN).state$;
+  }
+
+  async setV2UpgradeToken(token: V2UpgradeToken, userId: UserId): Promise<void> {
+    assertNonNullish(token, "token");
```
quexten, line 21: Note: Would be nice if we could use some macro magic at some point to replace these / not have to write them for every parameter and potentially be able to forget them.

quexten (approved):

PR #20651, opened 2026-05-14
[CL-667] refactor(components): extract reusable overflow-list directive

willmartian:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/CL-667

## 📔 Objective

Extract the measurement / packing / hiding logic from `bit-tab-group` and `bit-tab-nav-bar` into a reusable `[bitOverflowList]` + `[bitOverflowItem]` directive pair under `libs/components/src/overflow-list/`. The tabs components become thin consumers that own only tab-specific concerns (selection, keyboard navigation, ARIA, content vs nav variants).

### Why

- The overflow logic in #20010 is a general "priority+" pattern that applies beyond tabs (toolbars, action bars, breadcrumbs). Extracting it makes that reuse explicit.
- `bit-tab-group` and `bit-tab-nav-bar` shrink to focus on their domain. The data pipeline (resize → pack → hide) lives in one place.
- The pack decision is a pure function of measured widths, container width, gap, and the pinned index — no feedback loop from render state back into measurement.

### What's in the directive

- `[bitOverflowList]` — smart container. Measures items, owns the `ResizeObserver`, exposes `displayed` / `overflow` / `ready` signals, and applies `[hidden]` to overflowed items. Accepts a `[gap]` input (applied as `style.gap` on the host) and an `[items]` input as an alternative to `contentChildren` for projection cases.
- `[bitOverflowItem]` — marker on each item with a `[pinned]` model. When pinned, the item stays visible at its ordinal position regardless of fit.

## 📸 Screenshots

N/A — visual behavior matches #20010. Verified in Storybook: Overflow Tabs, Overflow Navigation Tabs, and the resize cycle (1 → many → 1).


BryanCunningham (approved):
Claude comment feels like it needs addressed but great overall!

PR #20656, opened 2026-05-14
Expired Card POC

coltonhurst:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-37606


nick-livefront:
Hey @coltonhurst! This work floated my way from an internal vault discussion around showing License's and Passports as expired. Those currently aren't out in the wild yet but making `isCardExpired` more generic would make that integration much more seamless. Showing the new item types as expired isn't currently tracked by any means, I only wanted to mention it here! We can definitely solve for that if/when we get into that work.

coltonhurst:
> Hey @coltonhurst! This work floated my way from an internal vault discussion around showing License's and Passports as expired. Those currently aren't out in the wild yet but making `isCardExpired` more generic would make that integration much more seamless. Showing the new item types as expired isn't currently tracked by any means, I only wanted to mention it here! We can definitely solve for that if/when we get into that work.

@nick-livefront sorry for the late reply, that's great! So is Vault looking for something like this to be pushed through first, or are you saying that when you all do that work, this will be automatically included (ie - I should close this PR)?

nick-livefront:
> @nick-livefront sorry for the late reply, that's great! So is Vault looking for something like this to be pushed through first, or are you saying that when you all do that work, this will be automatically included (ie - I should close this PR)?

@coltonhurst - Whichever comes first more than likely. We don't have plans at this point in time to do the associated work. I only wanted to raise it if you wanted to move this through.

coltonhurst:
> > @nick-livefront sorry for the late reply, that's great! So is Vault looking for something like this to be pushed through first, or are you saying that when you all do that work, this will be automatically included (ie - I should close this PR)?
> 
> @coltonhurst - Whichever comes first more than likely. We don't have plans at this point in time to do the associated work. I only wanted to raise it if you wanted to move this through.

Excellent, yes, thank you for the context, that's helpful!

PR #20657, opened 2026-05-14
SSH Agent v2: Sign Request

### LINKED ISSUE #16681: SSH Agent signs all RSA Keys with sha512, regardless of what was asked
### Steps To Reproduce

1. Import an RSA SSH Key into Bitwarden
2. Enable the SSH-Agent
3. Try to use the key with openssh, when it is requesting a sign using rsa-sha2-256 (I don't know how to force this for now, but if you do ssh -vvv it should say "debug3: sign_and_send_pubkey: signing using rsa-sha2-256")

### Expected Result

A successful connection

### Actual Result

The connection fails with 
"agent key RSA SHA256:<fingerprint> returned incorrect signature type
sign_and_send_pubkey: no mutual signature supported"

### Screenshots or Videos

_No response_

### Additional Context

So I actually tracked down the issue exactly, here is the relevant information : 

I retraced the whole key signing code flow and it becomes clear what happens : 

First, the key is constructed here : 
https://github.com/bitwarden/clients/blob/147b511d64fe6d13d91e57a9ec19a30602f7f623/apps/desktop/desktop_native/core/src/ssh_agent/mod.rs#L282
and put in the `BitwardenSshKey` struct, the actual private key is a ssh_key::private::PrivateKey struct (important because in the agent library its refered as a dyn ssh_key::SigningKey)

Then, in the agent library, the function call to sign data with the key is here : 

https://github.com/bitwarden/bitwarden-russh/blob/a641316227227f8777fdf56ac9fa2d6b5f7fe662/src/ssh_agent.rs#L240

It's a trait method, so we can follow to the implementation for ssh_key::private::PrivateKey (since we know the type by looking at the bitwarden/clients code

Here is the implementation : 

https://github.com/RustCrypto/SSH/blob/bea6fe42a47089c0b35879ef69e632ad256f145f/ssh-key/src/signature.rs#L273-L294

Nothing, special, it just forwards based on the enabled features and the key type. (it extracts the KeyPairData and calls try_sign on the keypair value inside of KeyPairData)

For RSA Keys, the keypair value of KeyPairData is of type RsaKeypair, and look at it's signing implementation : 

https://github.com/RustCrypto/SSH/blob/bea6fe42a47089c0b35879ef69e632ad256f145f/ssh-key/src/signature.rs#L664-L678

It always makes a signature with sha512 !

So actually the code changed here
https://github.com/RustCrypto/SSH/commit/26d482b1ddc133344f780ca825f718fd87a729da
 and now allows to make sha-1 or sha-256 signatures, so you would need to upgrade the ssh-key version, and also special case RSA keys.

However, the ssh-agent specification has a "flags" field right after the data to be signed, that specifies extra properties for RSA keys :

You can see the "flags" field here and the special RSA flags in 4.5.1 right below
https://www.ietf.org/archive/id/draft-miller-ssh-agent-07.html#name-private-key-operations

And the corresponding values here (bit flags) : 
https://www.ietf.org/archive/id/draft-miller-ssh-agent-07.html#name-signature-flags-2

And so there is also an issue with the code in bitwarden-russh because it doesn't read this field at all.

So, Finally, a summary of what to do : 

1. Upgrade ssh-key in bitwarden-russh (I don't know if you're okay depending on release candidates or want to wait for the full 0.7.0, but in any case you need at least 0.7.0-rc.0)
2. In bitwarden-russh, change the type of the private_key() method in the SshKey trait to return a full PrivateKey struct (from ssh-key)
3. In bitwarden-russh also depend on the "signature" crate to be able to use the "Signer" trait
4. Parse the flags field
5. if the key is RSA, and there are flags, call the correct signing method.

I have a small PR for bitwarden-russh, but some other code broke after upgrading ssh-key and I don't know how to fix it for now.

### Operating System

Linux

### Operating System Version

Fedora 42

### Installation method

Flatpak

### Build Version

tested on latest commit as well (7b94d6ab2b35ba65f294454daa6de7901be81015 at time of writing)

### Issue Tracking Info

neuronull:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-31761

fixes: #16681 

## 📔 Objective

Following up on List Request , this PR adds support for Sign Request in our v2 of the SSH Agent. The changes are largely isolated to the server and crypto modules.

Reviewer notes:

- Roughly 2/3 of the LOC are unit/int tests.

- The deps added to Cargo.toml, are not new deps. They are already used by `ssh-key` thereby a transient dep for us , we are now using them explicitly ourselves (as per the changes to the lock file)

- The change to the napi callback is because original introduction of that callback wasn't being used until now, and I had a misunderstanding about the call_async. The change is a simplification to our napi macro definition so it's a win win.

- The removal of the `sign_data()` and addition of `get_private_key` to the KeyStore is because IMO the keystore should just store and manage the access to the keys. The signing should be part of the crypto module only. So the caller now gets the private key and then the private key knows how to sign. Otherwise we'd have to add a dependency on the server, to the keystore, which would be fugly.

- This PR adds support for the RSA flags, as was first astutely pointed out and a fix for proposed in the v1 by @LelouBil  . 
    - This solution does so by still leveraging the latest stable version of `ssh-key`.

- Took the opportunity to do minor touchups on the parsing of the requests in the server, just added a couple helper functions and data structures to clean it up. Note that I considered a full fledged structure system for serializing and deserializing the requests, but all that will do is hide the logic we need to do anyway, and there are a very small number of requests we are dealing with. The `ConnectionHandler` code is fairly high level over the protocol so I think it's reasonable to do without.

- Improved the logging a bit by changing log levels, adding more log statements, based on manual testing.

## Testing Done

- manually validated signing with the following methods, with all permutations of the User Settings for Authorization:
    - "ssh -T git@github.com"
    - creation of a text file, signing of the file and validating the signature


neuronull:
FYI the CI failure on Testing / Rust Coverage ; is not unique to this PR. Other PRs are hitting this as well, not related to changes here. It's been reported to be fixed by retries; I will attempt that.

apps/desktop/desktop_native/ssh_agent/Cargo.toml
```diff
@@ -12,6 +12,8 @@ base64 = { workspace = true }
 desktop_core = { path = "../core" }
 mockall = { workspace = true }
 rkyv = "=0.8.16"
+rsa = "0.9"
```
quexten, line 15: Can we consider using `[v0.10.0-rc.18](https://github.com/RustCrypto/RSA/releases/tag/v0.10.0-rc.18)`? The reason why I'm asknig that here is because it fixes a vulnerability around timing, and also is a breaking change in terms of API. If we are adding it as a net new dependency anyways, we may as well do that breaking change here so we don't have to re-QA it, and so it fits in well into the way we consume it in our crate.

apps/desktop/desktop_native/ssh_agent/src/crypto/mod.rs
```diff
@@ -49,11 +54,32 @@ impl PrivateKey {
...
+
+            Signature::new(
+                ssh_key::Algorithm::Rsa {
+                    hash: Some(flag.hash_alg()),
+                },
+                rsa_sig.to_vec(),
+            )
+            .expect("RSA-SHA256 signature construction should succeed")
         }
+        Some(SignFlags::RsaSha512) | None => kp.sign(data),
```
quexten: Following the RFC's, `None` should actually be an RSA signature with SHA-1 as the hashing scheme. (SHA1 is insecure and should not be used). I think in this case we should just reject the signature request? What do you think?

apps/desktop/desktop_native/ssh_agent/src/server/connection.rs
```diff
@@ -4,13 +4,21 @@ use std::sync::Arc;
...
+    protocol::{
+        build_identities_answer, build_sign_response, detect_namespace, failure, frame,
+        parse_message, AgentMessage, SignFlags,
+    },
+    KeyStore,
 };
+use crate::crypto::PublicKey;
+
+// Guards against oversized allocations from untrusted length prefixes on the socket.
+const MAX_MESSAGE_LEN: usize = 256 * 1024;
```
quexten, line 21: Interesting. Does this mean on the current SSH agent, if you forward your agent to a server, that server can crash your local machine via crafted message?

apps/desktop/desktop_native/ssh_agent/src/server/connection.rs
```diff
@@ -140,21 +170,70 @@ async fn handle_list_request<K: KeyStore, A: AuthPolicy>(
...
+    peer_info: Option<&PeerInfo>,
+    keystore: &Arc<K>,
+    auth_policy: &Arc<A>,
+) -> Vec<u8> {
+    debug!("handling sign request");
+
+    let sign_request = SignRequest {
+        public_key: public_key.clone(),
+        process_name: peer_info.map(|p| p.process_name().to_string()),
+        is_forwarding: peer_info.is_none(),
```
quexten, line 210: Don't address here;

This feels weird. We should probably have another `Option<ForwardingInfo>` which is an enum, one variant of which can contain:
- Who are we forwarding to?
  - This can be determined via a few protocol extensions, we can even match a hostname to it

quexten (commented):
Mostly looking good I think. Love the tests.

I have some comments w.r.t. spec adherence, and one around connection forwarding info.

apps/desktop/desktop_native/ssh_agent/Cargo.toml
neuronull: I had intentionally used the same version that's already being used because it's the latest stable. 
But I wasn't aware of the newer one fixing a timing vulnerability.

That said, I do think it's helpful to have a separation of concerns. IIUC, the newer version fixes a security vulnerability, but isn't necessary for us to complete the implementation. Thus, it's better to do them separately in my opinion. If there's any problem with the newer version, it doesn't impact the rest of this implementation. 

I have some other questions, I'll ask out of band.

neuronull: Summarizing our decision: we do want the vuln fix, however since it's not a new vuln and the practical exploitability is low, and to separate concerns out of the v2 effort, we'll track and work on that in a separate task.
Thank you for raising!

apps/desktop/desktop_native/ssh_agent/src/crypto/mod.rs
neuronull: Good point. I agree that even though the action taken here is safe, it silently dis-obeys the protocol's definition.

Agreed to reject the request. I think the cleanest way to do that is not allow the rsa requests that don't have the flags. And do that early rather than here. I'll push up a change for that.

neuronull: Addressed in https://github.com/bitwarden/clients/pull/20657/commits/9c25ae80972092b473cfe77274f1eaab0304f307

apps/desktop/desktop_native/ssh_agent/src/server/connection.rs
neuronull: I haven't inspected the current agent for this guard, but yes, that is the worst case that this protects against. In theory, either a buggy client or malicious one, without such a guard, could theoretically set the len to `0xFFFFFFFF` , which would result in a heap allocation of ~4 GiB. That could OOM the agent or stall it with swaps. Kind of DoS-esque.

apps/desktop/desktop_native/ssh_agent/src/server/connection.rs
neuronull: Good callout. I believe I raised a similar note in one of the earlier interface PRs.

IIRC the issue I ran into is that when we cross the boundary to TS, we don't have as many language features to take advantage of. So even though `is_forwarding` and `process_name` are conceptually tied together, the napi translation of an Option where its an enum variant that contains a string, is probably going to result in us separating out that into two distinct fields. I believe that (in combination with not having the implementation done end-to-end yet), led me to not pursue it at the time,
That's not to say we can't or shouldn't still do something about it though. I'd prefer us to not make sacrifices in DN code just because of the eventual translation across napi. 

If you're on the same page, I'll create a ticket to track this as a follow-up to the v2.

neuronull: https://bitwarden.atlassian.net/browse/PM-37774

dereknance (approved):
Platform changes look good.

apps/desktop/desktop_native/ssh_agent/src/server/connection.rs
quexten: Thanks for clarifying :)

quexten (approved):

PR #20670, opened 2026-05-15
[PM-36172]Update notification bar logic when passwords match. 

dan-livefront:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-36172
## 📔 Objective

Currently, when the form username + password matched a vault password but not the vault username, we still queued `change password` for every password only match. The notification bar also displayed stored usernames instead of the submitted form username, which made the UX feel misleading.

With the changes in this PR, when the form contains a `username` and the vault classification is `passwordOnlyMatches`:

- Prefer `update` only for ciphers with no stored username
- Otherwise prefer `save new login` using the submitted form values
- If `save new login` is disabled and all matches already contain usernames, show no notification

The `password plus new password` flow without a username field remains unchanged.

| Scenario | Behavior vs `main` |
|---|---|
| `usernamePasswordNewPassword` | Changed when `passwordOnlyMatches` and the form contains a username: update only vault logins without usernames, otherwise prefer add login from the form, otherwise suppress the misleading change password list when save new is disabled |
| `usernamePassword` | Changed with the same `passwordOnlyMatches` plus username on form behavior as `usernamePasswordNewPassword` |
| `passwordNewPassword` | Unchanged |
| `usernameNewPassword` | Unchanged |
| `username` | Unchanged |
| `password` | Unchanged |
| `newPassword` | Unchanged |

> [!NOTE]  
> Existing TODO was hoisted to one place and will be tracked here https://bitwarden.atlassian.net/browse/PM-37609


apps/browser/src/autofill/background/notification.background.ts
```diff
@@ -1028,6 +1028,12 @@ export default class NotificationBackground {
       usernameOnlyMatches,
       usernamePasswordMatches,
     } = ciphersByInputMatchCategory;
+
+    // @TODO handle empty strings / incomplete data structure
+    const submittedPassword = data.newPassword || data.password;
```
dan-livefront, line 1083: No optional chaining needed, data is never undefined at the type level, `handleInputMatchScenario` takes `ModifyLoginCipherFormData`, and that type requires all four fields as string.

```
export type ModifyLoginCipherFormData = {
  uri: string;
  username: string;
  password: string;
  newPassword: string;
};

```

in overlay notification background

jprusik (approved):

PR #20676, opened 2026-05-15
[WIP] Qualification engine adapter and confidence scoring refactor

blackwood:
## Notes

TODO
- Performance benchmark comparison
- Pruning
- Readability/diagramming/examples
- Run against BIT
- Distribute in smaller PRs


apps/browser/src/autofill/qualification/abstractions/qualification-engine.ts
```diff
@@ -0,0 +1,76 @@
...
+export interface PageQualification {
+  fieldFor(opid: string): FieldClassification | null;
+  formFor(opid: string): FormClassification | null;
+  scenario(): PageScenario | null;
+  /**
+   * Resolves when the engine has finished classifying. Engines that produce
+   * eager, synchronous results omit this. Remote engines may expose it so
+   * consumers that want to await full classification can.
+   */
+  ready?: Promise<void>;
```
audreyality, line 25: 🤔 Is it necessary for this to be undefined? An awaited resolved promise automatically continues. You can simplify the consuming path by having this value always defined.

🎨 If there are some paths that cannot use `await`, consider instead creating two `PageQualification` types and distinguishing between them using `Promise<T>` for those methods that may emit asynchronous results.

For a small increase in type and utility complexity, you now get full typechecking support and better callsite ergonomics (less branching, better JIT optimization).

```ts
export interface SyncPageQualification {
  type: "synchronous";
  fieldFor(opid: string): FieldClassification;
  formFor(opid: string): FormClassification;
  scenario(): PageScenario;
}

export interface AsyncPageQualification = {
  type: "asynchronous";
  fieldFor(opid: string): Promise<FieldClassification>;
  formFor(opid: string): Promise<FormClassification>;
  scenario(): Promise<PageScenario>;
}

export type PageQualification = SyncPageQualification | AsyncPageQualification;

/** type guard identifying synchronous quals */
export function isSynchronous(qualification: PageQualification) 
  : qualification is SyncPageQualification {
  return qualification.type === "synchronous";
}

/** normalize a qual to synchronous */
export function toAsynchronous(qualification: PageQualification) : AsyncPageQualification {
  if(isSynchronous(qualification)) {
    return new Proxy(qualification, {
      get(target, prop, receiver) {
        if (prop in ["fieldFor", "formFor", "scenario"]) {
          return (...args: any[]) => Promise.resolve(target[prop](...args)));
        }
        return Reflect.get(...arguments);
      },
    });
  }
}

// usage
const allQuals : PageQualification[]; // = ...

// callsites that cannot be async need can skip the uncallable quals
const qualifications = allQuals.filter(isSynchronous);

// this can also work the other way, in case you need to to something
// fancy like schedule async quals in sync code
const qualifications = allQuals.filter((q) => !isSynchronous(q));

// If you need to run all of them, map them!
const qualifications = allQuals.map(toAsynchronous);
```

💭 Another possibility is that you adjust the design to have the async interface return the sync interface once it's ready. This is better if there's no need to distinguish between the runtime characteristics at your callsite.

```ts
export interface AsyncPageQualification {
  ready: Promise<PageQualification>;
}
```

🎨 An observable solution would be much more involved, so I won't document it here. The cool part about it, however, is that you'd be able to keep the backend running across page qualifications. Say, for example, that you have a SPA with a multi-step login process or a form that swaps address templates by region. The qual could emit each time a new field becomes available.

apps/browser/src/autofill/qualification/abstractions/qualification-engine.ts
```diff
@@ -0,0 +1,76 @@
...
+   * qualification source to avoid silent regression.
+   */
+  readonly coveredRoles?: ReadonlySet<FieldRole>;
+
+  /**
+   * FormCategories this engine knows how to emit on `FormClassification.matchedCategories`
+   * (and on each field's `matchedFormContexts`). Same routing principle as
+   * {@link coveredRoles}.
+   */
+  readonly coveredCategories?: ReadonlySet<FormCategory>;
```
audreyality, line 75: 🤔 I think it's worthwhile to eliminate optionality on these fields and require callers always opt-in to their covered roles and categories. Each QualificationEngine will have a fixed set, and it's unlikely that we'll want a qualification engine to opt-in to as-yet-unknown roles or categories.

🎨 If there's a specific use-case in which all roles would be useful, we could instead use a builder pattern or pass it into the engine's constructor. The engine could then create a set containing all available roles or categories.

apps/browser/src/autofill/qualification/types/classification.ts
```diff
@@ -0,0 +1,56 @@
...
+   * The role the engine selected as this field's primary classification, if
+   * any. Engines that do mutual-exclusion dispatch (e.g. a future scoring
+   * engine) populate this with their dispatch winner. Engines without
+   * dispatch (the legacy bridge) populate it with the first role inserted
+   * into `matchedRoles` — which reflects predicate iteration order, not any
+   * notion of "best." Consumers that need the legacy-faithful boolean should
+   * read `matchedRoles.has(role)`, not `topRole === role`.
+   */
+  topRole: FieldRole | null;
+  confidence: ConfidenceBand;
```
audreyality, line 43: ⚠️ Confidence should be communicated in terms of a floating point number [0..1]. This ensures that the classification value of a field can be [hypothesis tested](https://en.wikipedia.org/wiki/Statistical_hypothesis_test). If there are certain places where a qualitative bucket is needed, a reduction function can be used.

audreyality (commented):
🤔 This rapidly got too big for me to comprehend. There's a lot of code and very little documentation. I'd consider outlining a few things.

1. Design Philosophy - what are the key decisions and constraints driving the design? Here, focus on outlining the problems to be solved and explain how the design addresses those problems on an architectural level. I'd include a high-level overview of how data flows through the system, and provide references to resources to learn more.
2. Usage - How is this expected to be used? Here, focus on how it's extended (how do I add a new kind of qualification). I'd include a detailed explanation of the core concept(s) needed for each kind of extension in the usage documentation.
3. Compatibility - There's a substantial part of this logic focused on retaining compatibility with the existing qualification system. I'd call this out separately from the design philosophy, and focus on those things necessitated by the old system that are out of scope of the new system. That is, focus on things that you had to do, but wish to eliminate once compatibility is no longer required.

blackwood:
@audreyality Good feedback. I'd clarify that this is a exploratory draft--a true WIP that is subject to change.

PR #13324, opened 2025-02-09
[PM-17784/PM-17588] Flatpak & Mac Store SSH Agent

quexten:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-17784
https://bitwarden.atlassian.net/browse/PM-17588

## 📔 Objective

Both mac app store and flatpak sandbox the process list, so the information we get back when inspecting the process for the pid obtained by peercred fails. This should not crash, or stop the ssh agent, but instead return "unknown process".

Further, flatpak does not have access to the home directory and requires a separate path (`.var/app/com.bitwarden.desktop/data/.bitwarden-ssh-agent.sock`) configured that other apps can access without changing any permissions.

## ⏰ Reminders before review

- Contributor guidelines followed
- All formatters and local linters executed and passed
- Written new unit and / or integration tests where applicable
- Protected functional changes with optionality (feature flags)
- Used internationalization (i18n) for all UI strings
- CI builds passed
- Communicated to DevOps any deployment requirements
- Updated any necessary documentation (Confluence, contributing docs) or informed the documentation team

## 🦮 Reviewer guidelines

- 👍 (`:+1:`) or similar for great changes
- 📝 (`:memo:`) or ℹ️ (`:information_source:`) for notes or general info
- ❓ (`:question:`) for questions
- 🤔 (`:thinking:`) or 💭 (`:thought_balloon:`) for more open inquiry that's not quite a confirmed issue and could potentially benefit from discussion
- 🎨 (`:art:`) for suggestions / improvements
- ❌ (`:x:`) or ⚠️ (`:warning:`) for more significant problems or concerns needing attention
- 🌱 (`:seedling:`) or ♻️ (`:recycle:`) for future improvements or indications of technical debt
- ⛏ (`:pick:`) for minor or nitpick changes

PR #20680, opened 2026-05-17
[PM-37656] Resolve SSH agent caller name via libproc on macOS

### LINKED ISSUE #20679: [Desktop] SSH agent prompt shows "Unknown application" on macOS — process_name lookup via sysinfo silently fails under hardened runtime
> **Edit 2026-05-21**: Corrected installation method to **Mac App Store** (the original report incorrectly said Direct Download). The direct-download build is **not** affected — see [@neuronull's verification](https://github.com/bitwarden/clients/issues/20679#issuecomment-4489851230) and [@quexten's clarification](https://github.com/bitwarden/clients/issues/20679#issuecomment-4505442154). The bug is specific to the App Store sandboxed build.

### Steps To Reproduce

1. Install Bitwarden Desktop 2026.3.1 on macOS via the **Mac App Store**.
2. Add at least one SSH key to the vault and enable the built-in SSH agent (Settings → SSH Agent).
3. From Terminal, point any SSH client at Bitwarden's socket and trigger a signing request — e.g.
   ```
   export SSH_AUTH_SOCK="$HOME/Library/Containers/com.bitwarden.desktop/Data/.bitwarden-ssh-agent.sock"
   ssh -T git@github.com
   ```
   (Any caller works: `ssh-add -l`, `git push`, VS Code Remote-SSH, etc.)
4. Observe the approval prompt that Bitwarden displays.

### Expected Result

The approval prompt should identify the calling process — e.g. `ssh`, `Terminal`, `git` — similar to how 1Password's SSH agent shows the caller's name on macOS. The SSH agent protocol itself has no standard way to pass caller identity, but the OS does: on macOS this is typically resolved via `SO_PEERPID` (or equivalent) plus `proc_pidpath()`.

### Actual Result

The approval prompt always displays `Unknown application`, regardless of which process initiated the connection. The user cannot distinguish a legitimate `ssh` call from any other tool on the same machine, which materially weakens the security value of the prompt: an attacker process can request signatures and the prompt looks identical to a benign `git push`.

### Screenshots or Videos

N/A — prompt text is described in *Actual Result* above.

### Additional Context

**Root cause (from source reading):**

`PeerInfo` already has a `process_name: String` field, and `peerinfo/gather.rs` uses `sysinfo`'s `process.name()` to populate it:

https://github.com/bitwarden/clients/blob/main/apps/desktop/desktop_native/core/src/ssh_agent/peerinfo/gather.rs

Inside the Mac App Store sandbox the `sysinfo` crate cannot retrieve the name of another user-space process without the `task_for_pid` entitlement (which the App Store build does not — and cannot — ship with), so `process.name()` returns empty (or `get_peer_info` returns `Err`). The Rust side then falls back to `PeerInfo::unknown()` whose `process_name` is the hardcoded literal `"Unknown application"`, which is what the UI renders. This is the same root cause as #13075; the existing workaround in #13324 falls back to `"unknown process"` instead, but does not resolve the caller's actual name. Direct-download builds are not sandboxed and `sysinfo` works fine for them.

**Proposed fix (sketch — happy to send a PR if maintainers are open to the approach):**

Use the BSD libc `proc_pidpath(pid, ...)` via the [`libproc`](https://crates.io/crates/libproc) crate on macOS. `proc_pidpath` does **not** require the `task_for_pid` entitlement and is reachable from inside the App Store sandbox.

1. `peercred_unix_listener_stream.rs` already captures peer PID via tokio's `peer_cred()`.
2. On macOS, call `libproc::proc_pid::pidpath(pid)`, take the basename, and fill `PeerInfo.process_name`.
3. Keep the existing `sysinfo` path as the generic fallback for Linux and as a secondary fallback on macOS.

Roughly a 60–80 line, macOS-only change in 2–3 files under `desktop_native/core/src/ssh_agent/peerinfo/`. No IPC schema changes — the front-end automatically stops showing `Unknown application` once a non-empty `process_name` flows through. `libproc` would be added as a `[target.'cfg(target_os = "macos")'.dependencies]` entry, so no impact on Linux/Windows builds.

1Password's SSH agent uses this same `SO_PEERPID` + `proc_pidpath` mechanism on macOS — it's the established pattern.

**Cross-references:**

- #14138 (closed) — addressed the "remember authorization" pain point via #13995 but did not touch the caller-name display.
- #13075 / #13324 — handled the same App Store sandbox case by falling back to `"unknown process"`. The `libproc` path proposed here resolves the actual caller name instead.
- SSH Agent v2 refactor in flight (the #18880 → #20657 chain). Guidance appreciated on whether this fix should target the v1 path (current production), the new `PeerInfo::from_pid()` in v2, or both.

Happy to sign the CLA and submit a PR once a maintainer confirms the preferred direction (v1 vs v2 vs both) and whether `libproc` is acceptable as a new macOS-only dependency.

### Operating System

macOS

### Operating System Version

macOS 26.3 (Tahoe), build 25D125

### Installation method

Mac App Store

### Build Version

2026.3.1

### Issue Tracking Info

lnky102:
> **Edit 2026-05-21**: Issue #20679 originally said direct-download builds were also affected — that was incorrect (corrected in the issue). This PR addresses the **Mac App Store sandboxed build specifically**, complementing #13324's `"unknown process"` fallback with a real caller-name lookup. Direct-download / Linux / Windows behavior is unchanged.

## 🎟️ Tracking

Resolves #20679

## 📔 Objective

On macOS the SSH agent approval prompt always shows "Unknown application" instead of the caller's process name (`ssh`, `git`, `Terminal`, …) inside the **Mac App Store sandboxed build**. The root cause is the same in both code paths that currently live in `main`:

- **v1** — `apps/desktop/desktop_native/core/src/ssh_agent/peerinfo/gather.rs`. On lookup failure the upstream listener falls back to `PeerInfo::unknown()` whose `process_name` is the literal string `"Unknown application"`, which is what the UI renders.
- **v2** — `apps/desktop/desktop_native/ssh_agent/src/server/peer_info.rs`, part of the #18880 → #20657 refactor chain that has already begun landing on `main`. `PeerInfo::from_pid` returns `None` with the same downstream effect.

Both paths use `sysinfo::Process::name()` to populate `process_name`, which depends on APIs that need the `task_for_pid` entitlement. The Mac App Store sandboxed build does not — and cannot — ship with that entitlement, so the lookup returns empty for App Store users. Direct-download builds are not sandboxed and work correctly today (verified by maintainers in #20679).

This PR adds a macOS-only fast path in **both** code paths that resolves the caller's executable via `proc_pidpath(2)` — the BSD syscall documented in the macOS SDK man pages as the supported way to map a PID to its executable path. `proc_pidpath` does **not** require `task_for_pid` and is reachable from inside the App Store sandbox. On any failure the existing `sysinfo` path is reached, so direct-download macOS / Linux / Windows behavior is unchanged.

### Changes

- `apps/desktop/desktop_native/Cargo.toml` — add `libproc = "=0.14.11"` to `[workspace.dependencies]`.
- `apps/desktop/desktop_native/core/Cargo.toml` (v1) — depend on `libproc` only under `[target.'cfg(target_os = "macos")'.dependencies]`.
- `apps/desktop/desktop_native/core/src/ssh_agent/peerinfo/gather.rs` (v1) — macOS fast path via `libproc::proc_pid::pidpath`, basename extraction, sysinfo fallback unchanged. Unit tests for self-pid resolution.
- `apps/desktop/desktop_native/ssh_agent/Cargo.toml` (v2) — add `[target.'cfg(target_os = "macos")'.dependencies]` with `libproc`.
- `apps/desktop/desktop_native/ssh_agent/src/server/peer_info.rs` (v2) — identical macOS fast path in `PeerInfo::from_pid`, sysinfo fallback unchanged. Existing v2 tests preserved; one mac-specific basename test added.

No IPC schema changes and no front-end changes — the existing "Unknown application" string in `PeerInfo::unknown()` and the v2 `None` branch are only reached when peer-info lookup errors, which the libproc path makes far less likely inside the App Store sandbox.

### Test plan

> **Important**: the unit tests below exercise the new code paths and validate basename extraction / error handling on an unsandboxed `cargo` harness. **They do not validate the fix inside the actual App Store sandbox** — sandbox validation requires manual end-to-end testing against a signed App Store build, which only Bitwarden's internal build pipeline can produce.

- `cargo test -p desktop_core --lib` on macOS 26.3 — 27 / 27 pass.
- `cargo test -p ssh_agent --lib` on macOS 26.3 — passes including a new mac-specific basename test.

### Dependency choice

`libproc` is a thin wrapper over `proc_pidpath(2)`. An alternative is to skip the new crate and call `proc_pidpath` directly through the existing `libc` workspace dependency (already used by the Linux target): roughly 10 lines of `unsafe extern` in exchange for one fewer external dependency. Happy to switch to that variant on request.

## 📸 Screenshots

N/A — no UI change. The localized prompt string is unchanged; only the value flowing into the existing `process_name` field changes from `"Unknown application"` (today's App Store behavior) to the caller's executable basename (e.g. `ssh`, `git`, `Code Helper`).


quexten (dismissed):
Approving for @bitwarden/wg-ssh-keys, but note that this also needs approval from @bitwarden/team-desktop-native-dev.

Pretty sweet work, thank you!

neuronull:
👋  hey @lnky102 thanks for opening a PR for this.
We had a small miscommunication in the opened GH issue that this targets, but now it's clarified.
This PR is in my queue and I hope to get to it tomorrow, thanks for your patience.

neuronull:
My main focus on the changes proposed is to ensure functionality.
The added unit tests don't actually cover the scenario that the source changes are attempting to address, because that requires sandboxing the process the agent's server is running within.

To that end I'm still looking into this PR, but I do have some questions/concerns about the PR description.

> 1Password's SSH agent uses the same mechanism on macOS — it's the established pattern.

Do you have a source or reference for this?

> libproc as a new mac-only dep. 

It's first release was actually in 2018.

Also, it claims to be supported on Linux. If that's true, we should in theory be able to utilize it for the linux code path as well.

> it's actively maintained (5.7M+ downloads, latest 0.14.11), pure-Rust wrapper over BSD libc, no transitive C deps. If you'd rather inline the FFI call against libc directly to avoid the dependency, happy to switch — let me know.

It seems to be stable ( and it's last release was 8 months ago), but the verbiage in this statement is rather embellishing.

Did you look into any alternative crates or solutions?

lnky102:
Thanks for the careful review @neuronull — these are all fair. Going through:

**Test coverage.** You're right; the self-pid tests run in the (unsandboxed) cargo harness and exercise the code paths but don't reproduce the actual App Store sandbox failure. An automated test of the sandboxed scenario would need to spawn a child via `sandbox-exec` with an App-Store-like profile — happy to do that if you think the assertion is worth the machinery, but end-to-end validation against an internal signed App Store build is probably more reliable.

**1Password reference.** That was sloppy of me — I don't have primary-source access to 1Password's implementation; the claim was inferred from comparing prompts. I'll drop that sentence. The real justification is the `proc_pidpath(2)` man page (shipped in the macOS SDK): it's the documented Apple-supported way to map a PID to its executable path without `task_for_pid`.

**libproc on Linux.** Correct, the crate is cross-platform (reads `/proc/<pid>/exe` on Linux). I scoped the PR to macOS because no Linux bug has been reported and `sysinfo` works there today. Happy to extend to Linux for consistency if you'd prefer, but it would be no-op churn behaviorally.

**Alternative solutions / dep wording.** Fair on the wording — I'll drop the download-count framing. On alternatives: the strongest one is direct FFI against the existing `libc` workspace dep. `libc::proc_pidpath` is already declared there, and `desktop_native` already depends on `libc` for Linux. Adding `libc` to the macOS target and calling `proc_pidpath` through ~10 lines of `unsafe extern` removes the `libproc` dep entirely. Would you prefer that variant? I'd actually lean toward it for the smaller surface. I'll push it as a follow-up commit if you say go.

Other alternatives I weighed and dismissed: `mach2` (heavier, full Mach API), low-level `sysctl` (`proc_pidpath` is the documented public path), and the existing `sysinfo` (the very thing that fails in sandbox).

neuronull:
Thanks for being flexible and ready to adapt the solution! 

TLDR: I'll do some local manual tests.

> **Test coverage.** You're right; the self-pid tests run in the (unsandboxed) cargo harness and exercise the code paths but don't reproduce the actual App Store sandbox failure. An automated test of the sandboxed scenario would need to spawn a child via `sandbox-exec` with an App-Store-like profile — happy to do that if you think the assertion is worth the machinery, but end-to-end validation against an internal signed App Store build is probably more reliable.

I mention this mostly as a note since the PR description makes it sound like the "test plan" is sufficient to cover it, which isn't entirely correct.

My plan is to do manual validations for this, which will increase the time a bit but I think is the way to go here.

> **libproc on Linux.** Correct, the crate is cross-platform (reads `/proc/<pid>/exe` on Linux). I scoped the PR to macOS because no Linux bug has been reported and `sysinfo` works there today. Happy to extend to Linux for consistency if you'd prefer, but it would be no-op churn behaviorally.

I'll see if I can repro the unknown application thing on a sandboxed linux build; I'm pretty sure I experienced that some months ago but I'll double check and report back.

> **Alternative solutions / dep wording.** Fair on the wording — I'll drop the download-count framing. On alternatives: the strongest one is direct FFI against the existing `libc` workspace dep. `libc::proc_pidpath` is already declared there, and `desktop_native` already depends on `libc` for Linux. Adding `libc` to the macOS target and calling `proc_pidpath` through ~10 lines of `unsafe extern` removes the `libproc` dep entirely. Would you prefer that variant? I'd actually lean toward it for the smaller surface. I'll push it as a follow-up commit if you say go.
> 
> Other alternatives I weighed and dismissed: `mach2` (heavier, full Mach API), low-level `sysctl` (`proc_pidpath` is the documented public path), and the existing `sysinfo` (the very thing that fails in sandbox).

Thanks for detailing your alternative considerations, I agree that if this crate accomplishes the need, it is probably the way to go. It is helpful when introducing a new dependency to describe the research done and why the given crate chosen over the alternatives (I do recognize you did this to an extent, with mention of the useage of unsafe and libc). Thanks for providing extra context.

quexten:
> libproc on Linux. Correct, the crate is cross-platform (reads /proc/<pid>/exe on Linux). I scoped the PR to macOS because no Linux bug has been reported and sysinfo works there today. Happy to extend to Linux for consistency if you'd prefer, but it would be no-op churn behaviorally.

> I'll see if I can repro the unknown application thing on a sandboxed linux build; I'm pretty sure I experienced that some months ago but I'll double check and report back.

@neuronull This will not work at least on flatpak (and most likely on snap, it will not work either). `/proc/pid/exe` itself is sandboxed by the pid namespace that the flatpak bubblewrap sandbox creates. There are a few ways to solve this, one of them would be granting `--filesystem=host-root:ro` and then reading from `/run/host/root/proc/<pid>/exe`.

apps/desktop/desktop_native/core/src/ssh_agent/peerinfo/gather.rs
```diff
@@ -3,6 +3,18 @@ use sysinfo::{Pid, System};
 use super::models::PeerInfo;
 
 pub fn get_peer_info(peer_pid: u32) -> Result<PeerInfo, String> {
+    // On macOS, sysinfo's `process.name()` often fails or returns empty under
+    // hardened runtime / sandbox because the underlying APIs require the
+    // `task_for_pid` entitlement which the desktop app does not ship with.
+    // `libproc::proc_pid::pidpath` wraps the BSD `proc_pidpath(2)` syscall
+    // which does not require that entitlement and works for arbitrary peer
+    // PIDs in both sandboxed and non-sandboxed builds. We try it first on
+    // macOS and fall back to the cross-platform sysinfo path on any failure.
```
neuronull: This comment is rather verbose and there is another one on the function itself. Please collapse them into one comment on the function which describes what it does.

apps/desktop/desktop_native/core/src/ssh_agent/peerinfo/gather.rs
```diff
@@ -3,6 +3,18 @@ use sysinfo::{Pid, System};
...
+    #[cfg(target_os = "macos")]
+    if let Some(name) = macos_process_name(peer_pid) {
+        return Ok(PeerInfo::new(peer_pid, peer_pid, name));
+    }
```
neuronull, line 17: 🎨  minor organizational suggestion- call the function `try_peer_info_for_platform()` , and have an implementation per OS target. The windows and linux will return None. 

Please also extract the sysinfo logic below into a helper function `try_peer_info_sysinfo()`.

Then, this get_peer_info() function just calls the two and have a comment explaining the logical chain.

```Rust

try_peer_info_for_platform().or(try_peer_info_sysinfo())

```

apps/desktop/desktop_native/core/src/ssh_agent/peerinfo/gather.rs
```diff
@@ -25,3 +37,49 @@ pub fn get_peer_info(peer_pid: u32) -> Result<PeerInfo, String> {
...
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    /// `get_peer_info` must resolve the *current* process to a non-empty name.
+    /// Otherwise the SSH agent UI would render "Unknown application" even for
+    /// the most easily-resolvable PID — exactly the regression this change
+    /// guards against on macOS.
```
neuronull: I added this unit test in the v2 so no harm in adding it here, but if adding a code comment, please write it in a way that is digestible. I would opt to just remove the comment and rename the function to match the name I used in the v2 unit test. But if you want to take a swing at revising the comment you're welcome to. At a minimum, don't reference "this change guards against" , because once the commit is merged, there is no "change" to reference.

apps/desktop/desktop_native/core/src/ssh_agent/peerinfo/gather.rs
```diff
@@ -25,3 +37,49 @@ pub fn get_peer_info(peer_pid: u32) -> Result<PeerInfo, String> {
 
     Err("Failed to get process".to_string())
 }
+
+/// Resolve the executable basename of a peer process on macOS via
+/// `proc_pidpath(2)`. Returns `None` on any failure so the caller can fall
+/// back to the generic sysinfo path without surfacing the error.
+#[cfg(target_os = "macos")]
+fn macos_process_name(peer_pid: u32) -> Option<String> {
+    let path = libproc::proc_pid::pidpath(peer_pid as i32).ok()?;
```
neuronull: did you try using `name()` ? any reason not to use that? It seems to use the same library.

https://docs.rs/libproc/0.14.11/libproc/proc_pid/fn.name.html

apps/desktop/desktop_native/core/src/ssh_agent/peerinfo/gather.rs
```diff
@@ -25,3 +37,49 @@ pub fn get_peer_info(peer_pid: u32) -> Result<PeerInfo, String> {
...
+            !info.process_name().is_empty(),
+            "process_name must not be empty for current process"
+        );
+    }
+
+    /// On macOS the libproc fast path should return a non-empty basename for
+    /// the current process without requiring any extra entitlement.
+    #[cfg(target_os = "macos")]
+    #[test]
+    fn macos_process_name_returns_basename_for_self() {
```
neuronull: what are the scenarios where we get the path instead of the basename and do we need to test for that as well?

apps/desktop/desktop_native/core/src/ssh_agent/peerinfo/gather.rs
```diff
@@ -25,3 +37,49 @@ pub fn get_peer_info(peer_pid: u32) -> Result<PeerInfo, String> {
...
+        let info = get_peer_info(std::process::id())
+            .expect("self pid should resolve to a PeerInfo");
+        assert!(
+            !info.process_name().is_empty(),
+            "process_name must not be empty for current process"
+        );
+    }
+
+    /// On macOS the libproc fast path should return a non-empty basename for
+    /// the current process without requiring any extra entitlement.
```
neuronull: the comment about the entitlement, I'm not sure is valid to make here. Because this unit test doesn't do anything with the app packaging.

apps/desktop/desktop_native/ssh_agent/src/server/peer_info.rs
```diff
@@ -10,9 +10,22 @@ pub(crate) struct PeerInfo {
...
+    /// entitlement (which the desktop app does not ship with) and therefore
+    /// works under hardened runtime and the App Store sandbox, whereas
+    /// `sysinfo::Process::name()` typically returns empty in those builds.
+    /// We fall back to the cross-platform sysinfo path on any failure.
     pub(crate) fn from_pid(pid: u32) -> Option<Self> {
+        #[cfg(target_os = "macos")]
+        if let Some(process_name) = macos_process_name(pid) {
+            return Some(Self { pid, process_name });
+        }
+
```
neuronull, line 26: Please mirror the requested changes for the v1, here in the v2.

apps/desktop/desktop_native/Cargo.toml
```diff
@@ -44,6 +44,7 @@ homedir = "=0.3.6"
 interprocess = "=2.2.1"
 itertools = "=0.14.0"
 libc = "=0.2.178"
+libproc = "=0.14.11"
```
neuronull: Please move the dep to the core and ssh_agent crates directly. That way when re remove the v1 agent from core, we don't have to make any Cargo.toml changes in relation to this dependency.

neuronull (changes requested):
👋  thanks for your patience, it's been a busy week.

I was able to validate this manually for MAS builds.

A few suggestions:

lnky102:
Thanks for the careful pass @neuronull. Pushed e901c88 addressing all 7 inline points + a few extras from a pre-push review.

**Inline comments addressed**

1. **Comment consolidation** (`gather.rs:6-12`): collapsed both comments into a single function-level doc on `get_peer_info`.

2. **Refactor to `try_for_platform` / `try_sysinfo` helpers** (`gather.rs:13-16`): done in both v1 and v2 with the `try_peer_info_for_platform / try_peer_info_sysinfo` naming. Main path is `try_peer_info_for_platform().or_else(|| try_peer_info_sysinfo())` per your sketch — using `or_else` so the sysinfo work is skipped on the macOS happy path.

3. **Misleading test comments** (`gather.rs:58-61`): removed; v1 test renamed to mirror v2 style.

4. **`name()` vs `pidpath()`** (`gather.rs:46`): I kept `pidpath()`. `proc_pid::name()` is backed by `proc_pidinfo(PROC_PIDT_BSDINFO).pbi_name`, which is `MAXCOMLEN+1 = 16 bytes` — long names like `"Code Helper (Renderer)"` would truncate to `"Code Helper (Re"`. References: [libproc-rs `name` impl](https://github.com/andrewdavidmackenzie/libproc-rs/blob/master/src/libproc/proc_pid.rs) and `pbi_name` in [Apple's `sys/proc_info.h`](https://github.com/apple-oss-distributions/xnu/blob/main/bsd/sys/proc_info.h). Happy to switch if you prefer the truncation trade-off, but I assumed the full basename was more user-friendly.

5. **path-vs-basename coverage** (`gather.rs:76`): expanded `basename_or_path()` tests to cover normal path, no-separator input, root `/`, trailing slash `/foo/`, empty string, `.`, `..`, and NUL byte trimming.

6. **Entitlement comment in test** (`gather.rs:73`): removed — the test runs in an unsandboxed cargo harness, the entitlement claim was misplaced there.

7. **v2 mirror** (`peer_info.rs`): everything above applied symmetrically.

8. **`libproc` to crate-level** (`Cargo.toml`): moved out of `[workspace.dependencies]` into `core/Cargo.toml` and `ssh_agent/Cargo.toml` directly, matching the pattern from #20733. Added a `# Keep ... in sync with ...` comment on both sides.

**Extras from a pre-push review pass**

- `pidpath(peer_pid as i32)` could silently wrap for `peer_pid > i32::MAX`; switched to `i32::try_from(peer_pid).ok()?`.
- If `libproc` ever returns an empty path, the macOS fast path now treats it as failure so `sysinfo` retries.
- `basename_or_path()` defensively strips C-string NUL terminators so a stray NUL never reaches the UI.

**One pre-existing behavior worth flagging**

`PeerInfo::new(uid, pid, name)` on the v1 macOS branch passes `peer_pid` twice (uid == pid). The sysinfo branch calls `process.pid().as_u32()` for the second argument, but since sysinfo was queried for that exact PID it is identical to `peer_pid`. Both branches behaved this way before this PR — I left them as-is to keep the diff minimal, but happy to either (a) make the sysinfo branch pass `peer_pid` literally for symmetry, or (b) actually resolve the peer UID via `peer_cred()` if that's worth the scope. Let me know.

**Verification after this revision** (macOS 26.3)

- `cargo test -p desktop_core --lib` — 35 / 35 pass
- `cargo test -p ssh_agent --lib` — 70 / 70 pass
- `cargo clippy -p desktop_core -p ssh_agent --all-targets` — 0 warnings on the touched code

apps/desktop/desktop_native/core/src/ssh_agent/peerinfo/gather.rs
lnky102: Done in e901c88 — collapsed into the function-level doc on `get_peer_info`.

apps/desktop/desktop_native/core/src/ssh_agent/peerinfo/gather.rs
lnky102: Done in e901c88 — extracted `try_peer_info_for_platform()` (macOS impl + Windows/Linux no-op stub) and `try_peer_info_sysinfo()` helpers; `get_peer_info()` is now `try_peer_info_for_platform(peer_pid).or_else(|| try_peer_info_sysinfo(peer_pid)).ok_or_else(...)`. Used `or_else` (lazy) so sysinfo doesn't run on the macOS happy path. Mirrored in v2 with the same naming.

apps/desktop/desktop_native/core/src/ssh_agent/peerinfo/gather.rs
lnky102: Done in e901c88 — removed the "this change guards against" comment and renamed the v1 test to `macos_fast_path_returns_basename_for_self` so it matches the v2 naming.

apps/desktop/desktop_native/core/src/ssh_agent/peerinfo/gather.rs
lnky102: Kept `pidpath()` — `proc_pid::name()` is backed by Darwin's `proc_name()`, which fills `proc_pidinfo(PROC_PIDT_BSDINFO).pbi_name`. That field is `char[MAXCOMLEN+1]` = 17 bytes incl. NUL, so long names like `"Code Helper (Renderer)"` would truncate to `"Code Helper (Re"`. See [libproc-rs `name` impl](https://github.com/andrewdavidmackenzie/libproc-rs/blob/master/src/libproc/proc_pid.rs) and the `pbi_name` declaration in [Apple's `sys/proc_info.h`](https://github.com/apple-oss-distributions/xnu/blob/main/bsd/sys/proc_info.h). Happy to switch if you'd rather have the shorter API and accept the 16-byte limit, but `pidpath` + basename gives the full executable name.

apps/desktop/desktop_native/core/src/ssh_agent/peerinfo/gather.rs
lnky102: Expanded `basename_or_path()` test coverage in e901c88: normal path, no-separator input, root `/`, trailing slash `/foo/`, empty string, `.`, `..`, and NUL byte trimming. Covers both the `Path::file_name()` → `None` cases (root, `.`, `foo/..`) and the input-as-fallback case where no separator exists.

apps/desktop/desktop_native/core/src/ssh_agent/peerinfo/gather.rs
lnky102: Removed in e901c88 — agreed, the test runs in an unsandboxed cargo harness so the entitlement claim was misplaced here.

apps/desktop/desktop_native/ssh_agent/src/server/peer_info.rs
lnky102: Done in e901c88 — `peer_info.rs` now uses the same `try_peer_info_for_platform / try_peer_info_sysinfo` naming as v1, the same `or_else` chain in `from_pid`, a parallel `basename_or_path` helper with the same defensive NUL-trim, and the same expanded test suite.

apps/desktop/desktop_native/Cargo.toml
lnky102: Done in e901c88 — removed `libproc` from `[workspace.dependencies]` and inlined `libproc = "=0.14.11"` in both `core/Cargo.toml` and `ssh_agent/Cargo.toml`. Added a `# Keep ... in sync with ...` comment on both sides. When v1 is removed from core, only the `core/Cargo.toml` line needs to go — workspace stays untouched.

apps/desktop/desktop_native/core/src/ssh_agent/peerinfo/gather.rs
```diff
@@ -2,84 +2,144 @@ use sysinfo::{Pid, System};
...
+        .ok_or_else(|| "Failed to resolve peer process".to_string())
+}
+
+/// macOS-specific peer-info resolution via `libproc::proc_pid::pidpath`
+/// (the BSD `proc_pidpath(2)` syscall). This avoids the `task_for_pid`
+/// entitlement that `sysinfo::Process::name()` ends up needing on macOS,
+/// so it can still resolve the caller when `sysinfo` cannot (notably
+/// inside the Mac App Store sandbox). Returns `None` on any failure or
+/// when the resolved basename is empty, so the caller can fall through
+/// to the `sysinfo` path.
```
neuronull: (no idea where GH lost the "add suggestion" button but they lost it somewhere)

nit: remove ", so the caller can fall through
/// to the `sysinfo` path." , because that couples the function to the behavior of the caller unnecessarily.

apps/desktop/desktop_native/core/src/ssh_agent/peerinfo/gather.rs
```diff
@@ -2,84 +2,144 @@ use sysinfo::{Pid, System};
...
+    let pid_i32 = i32::try_from(peer_pid).ok()?;
+    let path = libproc::proc_pid::pidpath(pid_i32).ok()?;
+    let process_name = basename_or_path(&path);
+    if process_name.is_empty() {
+        return None;
     }
+    Some(PeerInfo::new(peer_pid, peer_pid, process_name))
+}
 
+/// Non-macOS platforms have no fast path; they fall through to sysinfo.
```
neuronull: nit: remove

apps/desktop/desktop_native/core/src/ssh_agent/peerinfo/gather.rs
```diff
@@ -2,84 +2,144 @@ use sysinfo::{Pid, System};
...
+#[cfg(not(target_os = "macos"))]
+fn try_peer_info_for_platform(_peer_pid: u32) -> Option<PeerInfo> {
+    None
+}
+
+/// Cross-platform peer-info resolution via the `sysinfo` crate.
```
neuronull: nit: remove

apps/desktop/desktop_native/core/src/ssh_agent/peerinfo/gather.rs
```diff
@@ -2,84 +2,144 @@ use sysinfo::{Pid, System};
...
+
+/// macOS-specific peer-info resolution via `libproc::proc_pid::pidpath`
+/// (the BSD `proc_pidpath(2)` syscall). This avoids the `task_for_pid`
+/// entitlement that `sysinfo::Process::name()` ends up needing on macOS,
+/// so it can still resolve the caller when `sysinfo` cannot (notably
+/// inside the Mac App Store sandbox). Returns `None` on any failure or
+/// when the resolved basename is empty, so the caller can fall through
+/// to the `sysinfo` path.
+#[cfg(target_os = "macos")]
+fn try_peer_info_for_platform(peer_pid: u32) -> Option<PeerInfo> {
```
neuronull: With the change to extract these out to helpers and returning Option, we're now losing error handling context (which can make debugging challenging). 
Let's have the helpers return `Result` with String as the error type, and on those cases where we're currently returning `None`, return `Err("<informative message">` , the same error type then allows the `get_peer_info()` to utilize `?`

apps/desktop/desktop_native/core/src/ssh_agent/peerinfo/gather.rs
```diff
@@ -2,84 +2,144 @@ use sysinfo::{Pid, System};
...
+    let pid_i32 = i32::try_from(peer_pid).ok()?;
+    let path = libproc::proc_pid::pidpath(pid_i32).ok()?;
+    let process_name = basename_or_path(&path);
+    if process_name.is_empty() {
+        return None;
```
neuronull: For example, `Err("Process name is empty")`

apps/desktop/desktop_native/core/src/ssh_agent/peerinfo/gather.rs
```diff
@@ -2,84 +2,144 @@ use sysinfo::{Pid, System};
...
+}
 
+/// Non-macOS platforms have no fast path; they fall through to sysinfo.
+#[cfg(not(target_os = "macos"))]
+fn try_peer_info_for_platform(_peer_pid: u32) -> Option<PeerInfo> {
+    None
+}
+
+/// Cross-platform peer-info resolution via the `sysinfo` crate.
+fn try_peer_info_sysinfo(peer_pid: u32) -> Option<PeerInfo> {
```
neuronull: Same suggestion here (return Result instead)

neuronull (commented):
Thanks for iterating on the solution! Overall looking good.

apps/desktop/desktop_native/core/src/ssh_agent/peerinfo/gather.rs
lnky102: Done in 7f7d97f — dropped the trailing sentence; helper doc no longer references caller behavior.

apps/desktop/desktop_native/core/src/ssh_agent/peerinfo/gather.rs
lnky102: Done in 7f7d97f — removed the doc comment from the non-macOS stub.

apps/desktop/desktop_native/core/src/ssh_agent/peerinfo/gather.rs
lnky102: Done in 7f7d97f — removed the doc comment from `try_peer_info_sysinfo`.

apps/desktop/desktop_native/core/src/ssh_agent/peerinfo/gather.rs
lnky102: Done in 7f7d97f. Both `try_peer_info_for_platform` and `try_peer_info_sysinfo` now return `Result<PeerInfo, String>` with informative error messages:

- `peer pid {n} exceeds i32 range`
- `proc_pidpath({n}) failed: {e}`
- `Process name is empty`
- `sysinfo: process {n} not found`
- `sysinfo: process {n} name is not valid UTF-8`
- `platform fast path not implemented` (non-macOS stub)

`get_peer_info()` is now `try_peer_info_for_platform(peer_pid).or_else(|_| try_peer_info_sysinfo(peer_pid))` so the sysinfo error surfaces when both paths fail. Added `macos_fast_path_propagates_error_for_dead_pid` that locks in the error-propagation guarantee for an unresolvable PID. Same change mirrored in v2 (`peer_info.rs`); `from_pid` still returns `Option<Self>` at the public boundary (with `.ok()`), but the internal helpers carry the error context.

apps/desktop/desktop_native/core/src/ssh_agent/peerinfo/gather.rs
lnky102: Done in 7f7d97f — that branch now returns `Err("Process name is empty".to_string())`.

apps/desktop/desktop_native/core/src/ssh_agent/peerinfo/gather.rs
lnky102: Done in 7f7d97f — `try_peer_info_sysinfo` returns `Result<PeerInfo, String>` with two distinct error strings: `sysinfo: process {n} not found` (missing PID lookup) and `sysinfo: process {n} name is not valid UTF-8` (name conversion failure).

apps/desktop/desktop_native/ssh_agent/src/server/peer_info.rs
```diff
@@ -10,20 +10,54 @@ pub(crate) struct PeerInfo {
...
-    /// Looks up the process name for `pid` via sysinfo and constructs a [`PeerInfo`].
-    /// Returns `None` if the process cannot be found or its name is not valid UTF-8.
+    /// Looks up the process name for `pid` by trying the platform-specific
+    /// fast path first and falling back to the cross-platform `sysinfo`
+    /// path. Returns `None` only when neither path can resolve the peer
+    /// process.
     pub(crate) fn from_pid(pid: u32) -> Option<Self> {
+        Self::try_peer_info_for_platform(pid)
+            .or_else(|_| Self::try_peer_info_sysinfo(pid))
+            .ok()
```
neuronull:
```Rust
        #[cfg(target_os = "macos")]
        match peer_info_from_libproc(pid) {
            Ok(info) => return Some(info),
            Err(e) => debug!("libproc peer-info resolution failed for pid {pid}: {e}; falling back to sysinfo"),
        }

        match peer_info_from_sysinfo(pid) {
            Ok(info) => Some(info),
            Err(e) => {
                warn!("failed to resolve peer process for pid {pid}: {e}");
                None
            }
        }
```

apps/desktop/desktop_native/ssh_agent/src/server/peer_info.rs
```diff
@@ -10,20 +10,54 @@ pub(crate) struct PeerInfo {
...
+    }
+
+    /// macOS-specific peer-info resolution via `libproc::proc_pid::pidpath`
+    /// (the BSD `proc_pidpath(2)` syscall). This avoids the `task_for_pid`
+    /// entitlement that `sysinfo::Process::name()` ends up needing on
+    /// macOS, so it can still resolve the caller when `sysinfo` cannot
+    /// (notably inside the Mac App Store sandbox).
+    #[cfg(target_os = "macos")]
+    fn try_peer_info_for_platform(pid: u32) -> Result<Self, String> {
```
neuronull: change back to `peer_info_from_libproc(pid: u32) -> Result<PeerInfo, String> ` , and make as a non-member helper function

apps/desktop/desktop_native/ssh_agent/src/server/peer_info.rs
```diff
@@ -10,20 +10,54 @@ pub(crate) struct PeerInfo {
...
+            .map_err(|e| format!("proc_pidpath({pid_i32}) failed: {e}"))?;
+        let process_name = basename_or_path(&path);
+        if process_name.is_empty() {
+            return Err("Process name is empty".to_string());
+        }
+        Ok(Self { pid, process_name })
+    }
+
+    #[cfg(not(target_os = "macos"))]
+    fn try_peer_info_for_platform(_pid: u32) -> Result<Self, String> {
```
neuronull: (delete this one)

apps/desktop/desktop_native/ssh_agent/src/server/peer_info.rs
```diff
@@ -10,20 +10,54 @@ pub(crate) struct PeerInfo {
...
+        Err("platform fast path not implemented".to_string())
+    }
+
+    fn try_peer_info_sysinfo(pid: u32) -> Result<Self, String> {
```
neuronull: move to non-member helper

neuronull (commented):
Looking good! I have one final request. And this is on me. In hindsight the suggestion to create a platform function and chain was a bad one. I didn't realize it until I looked at the handling for "error" path of your most recent commit for the case where we don't support. It ended up creating 3 logical branches to handle in the from_pid , where we really only have two that matter.

The latest commit also made me realize that the upstream callers of the from_pid and get_peer_info aren't doing anything with any Error returned ... I think there is _some_ value in logging for both cases.

My below suggestion is to return to just having the fast path in the base function, but to improve the logging. 
Note that this does introduce additional logging in the v1 where we aren't currently (when the sysinfo gives us nothing). I think it's a reasonable thing to log since generally we'll expect that to succeed.

For the v2 implementation I also suggest making the helper functions non-member functions since they don't rely on any data fields from the type itself.

Note that my below suggestions are only on the v2 files, but please make the corresponding changes for the v1 as well.

Thanks for your patience and sorry for the 180 on that aspect. I do think it was a worthwhile avenue to explore, as it also led to the discovery of the upstream handling of the return types.

Even in absence of the added value your PR brings for the macos sandboxed case, we're now also improving this area of the code for all platforms 🎉

apps/desktop/desktop_native/ssh_agent/src/server/peer_info.rs
lnky102: Applied verbatim in d053d18 for v2 `from_pid`, and mirrored for v1 `get_peer_info` — same `match peer_info_from_libproc(...)` with `debug!` on err / fall through, then `peer_info_from_sysinfo(...)` with `warn!` on err. v1 still returns `Result<PeerInfo, String>` for API stability, v2 still returns `Option<Self>`.

apps/desktop/desktop_native/ssh_agent/src/server/peer_info.rs
lnky102: Done in d053d18 — both `peer_info_from_libproc` and `peer_info_from_sysinfo` are now module-level non-member functions (no `impl PeerInfo` wrapper). Same shape in v1 `gather.rs`.

lnky102:
No worries on the 180 — the helper-chain detour was worth taking, the discovery that upstream callers ignore the `Error` and that the real value here is structured logging is a better outcome than the structural change alone. Pushed d053d18 with your suggested inline pattern in both v1 (`get_peer_info`) and v2 (`from_pid`): `peer_info_from_libproc` → `debug!` on err / fall through → `peer_info_from_sysinfo` → `warn!` on err. Helpers are non-member, return `Result<PeerInfo, String>`, error messages unchanged.

Tests (macOS 26.3): v1 36/36, v2 71/71. Clippy clean on the touched files.

quexten:
@lnky102 Are you using an automated AI agent for the interactions in this PR? If so I wonder about the dynamics of approving the CLA, since only humans can agree to CLAs.

lnky102:
@quexten Fair question to ask, and worth being explicit about.

I do use an AI coding assistant (Claude) to help draft replies and code in this thread, but it's not running unattended — every push, every comment, and the CLA signature itself are mine after manual review at each step. The assistant doesn't have my GitHub credentials, doesn't open browser sessions, and doesn't sign anything on my behalf. The CLA was signed by me personally via cla-assistant when the PR was first opened.

If Bitwarden has a policy that requires AI-assisted contributions to be flagged differently (in commits, PR title, or anywhere else), please point me to it and I'll comply.

apps/desktop/desktop_native/ssh_agent/src/server/peer_info.rs
lnky102: Done in d053d18 — the non-macOS stub is gone. The macOS fast-path call site is now `#[cfg(target_os = "macos")] match peer_info_from_libproc(...) { ... }` inside `from_pid`, so non-macOS targets skip the block entirely and no stub function is needed.

apps/desktop/desktop_native/ssh_agent/src/server/peer_info.rs
lnky102: Done in d053d18 — both `peer_info_from_libproc` and `peer_info_from_sysinfo` are now module-level non-member functions; neither lives in `impl PeerInfo`. Same shape in v1 `gather.rs`.

quexten:
No, we don't currently have such policies in place. Thank you for confirming!

neuronull:
👋  the changes look good to me.

There is a lint failure for `cargo +nightly fmt --check` , that affects both v1 and v2 , please address that.

I'll queue this change up for our internal QA team who will test it. Note that they have a lot on their plate with the testing of the v2 and other work, so this may take some time but we will get to it.

lnky102:
Done in 960b33f — pure formatting (`cargo +nightly fmt` collapsed a few short line-wraps in both `gather.rs` and `peer_info.rs`). v1 36/36 + v2 71/71 still pass, clippy clean.

Sounds good on the QA queue — no rush, and thanks for the patience through the iterations.

apps/desktop/desktop_native/core/src/ssh_agent/peerinfo/gather.rs
```diff
@@ -1,27 +1,171 @@
 use sysinfo::{Pid, System};
+use tracing::{debug, warn};
```
neuronull: there is another Rust lint failure (linux)

`debug` isn't used outside of the macos code path. I'd suggest referencing it directly as `tracing::debug!()` 

this applies to both v1 an v2

neuronull: for future reference: you can use `cargo cross` to do build checks etc on various cross platform targets

lnky102: Done in b554a80 — dropped `debug` from the import; the single macOS-only call site uses `tracing::debug!()` fully-qualified now. Same change in v1 and v2.

lnky102: Thanks for the tip on `cargo cross` — I'll wire that into my pre-push checks. Not having Linux/Windows build checks locally is what landed me here twice in a row.

neuronull:
Sorry for the churn, some more CI failures. This time while running `cargo sort` check.

While our contributing docs discuss how to install the required tools for the checks we don't explicitly state all the commands, that's something we can improve.
https://contributing.bitwarden.com/getting-started/clients/desktop/desktop-native/

Here is what you need to run to see the failure locally: `cargo sort --workspace --check` , removing the check option should fix the issue by writing to the file.

neuronull (dismissed):

neuronull:
All the remaining CI checks are expected. (some would clear by syncing your branch with `main` , but it's not required)

What remains is :
- platform team approval
- pass QA

neuronull:
👋  hi @lnky102  , I apologize for the runaround here ; we had another iteration of review by internal team whose preference it is to go with the `libc` choice (not adding the libproc crate dep). Would you be open to modifying your solution to make this adaptation?

lnky102:
Yes, happy to — that was actually the alternative I'd floated early on, so no problem adapting. Pushed in 62a2016.

**What changed**
- Dropped the `libproc` crate from both `core` and `ssh_agent` (and 73 lines of transitive deps from `Cargo.lock`).
- macOS resolution now calls `proc_pidpath(2)` directly through the existing `libc` dependency, wrapped in a small `pid_executable_path` helper that isolates the single `unsafe` block (zero-initialized `4 * PATH_MAX` buffer, errno-based error via `std::io::Error::last_os_error()`, UTF-8 validation).
- `core/Cargo.toml`: swapped `libproc` → `libc` under the macOS target. `ssh_agent/Cargo.toml`: dropped the macOS `libproc` block entirely since `libc` is already available via the existing `cfg(unix)` block.
- Kept the `peer_info_from_libproc` name — `proc_pidpath` is part of the macOS system `libproc` library (`<libproc.h>`), so it still describes the mechanism after dropping the crate.

**One robustness note from a pre-push review pass:** I had a second model audit the `unsafe` FFI, and it flagged that some XNU versions historically counted the trailing NUL in `proc_pidpath`'s returned length. To stay version-agnostic I trim at the first NUL after truncating to the returned length, plus a `debug_assert!` that the length never exceeds the buffer. Without that, a NUL-inclusive return would embed a `\0` in the path and reintroduce the exact mismatch this PR fixes.

**Two things for your call:**
1. The `unsafe` FFI body (`pid_executable_path`) is duplicated byte-for-byte across v1 and v2, same as `basename_or_path`. I kept them mirrored per your earlier guidance, but `ssh_agent` already depends on `core`, so I'm glad to hoist it into a single shared definition if you'd prefer one `unsafe` block to audit instead of two. Your call.
2. I can only run the macOS leg locally. Local checks are green — `cargo +nightly fmt --check`, `cargo sort --workspace --check`, `cargo clippy`, and tests (36 v1 / 71 v2) — but the Linux/Windows build legs only run under your CI, so I'd lean on that to confirm the dependency re-wiring across all targets.

apps/desktop/desktop_native/core/src/ssh_agent/peerinfo/gather.rs
```diff
@@ -25,24 +25,64 @@ pub fn get_peer_info(peer_pid: u32) -> Result<PeerInfo, String> {
...
+            "proc_pidpath({pid}) failed: {}",
+            std::io::Error::last_os_error()
+        ));
+    }
+    let written = written as usize;
+    debug_assert!(
+        written <= buf.len(),
+        "proc_pidpath wrote {written} bytes into a {}-byte buffer",
+        buf.len()
+    );
```
neuronull: This signals a buffer overrun, and I think we'd rather want to return an error and log this appropriately in release builds, than to let it go silently.

apps/desktop/desktop_native/core/src/ssh_agent/peerinfo/gather.rs
```diff
@@ -25,24 +25,64 @@ pub fn get_peer_info(peer_pid: u32) -> Result<PeerInfo, String> {
...
+/// string describing the failure. Requires no entitlement.
+#[cfg(target_os = "macos")]
+fn pid_executable_path(pid: i32) -> Result<String, String> {
+    // PROC_PIDPATHINFO_MAXSIZE from <sys/proc_info.h> is 4 * PATH_MAX.
+    let mut buf = vec![0u8; 4 * (libc::PATH_MAX as usize)];
+    // SAFETY: `proc_pidpath` writes at most `buf.len()` bytes into `buf` and
+    // returns the number of bytes written (<= 0 on failure). `buf` is a
+    // uniquely-owned, correctly-sized allocation that outlives the call, and
+    // its true length is passed as the buffer size, so the call cannot write
+    // out of bounds.
```
neuronull: nit: the text starting with description of `buf` seems unnecessary to me

apps/desktop/desktop_native/core/src/ssh_agent/peerinfo/gather.rs
```diff
@@ -25,24 +25,64 @@ pub fn get_peer_info(peer_pid: u32) -> Result<PeerInfo, String> {
...
-        .map_err(|e| format!("proc_pidpath({pid_i32}) failed: {e}"))?;
+    let path = pid_executable_path(pid_i32)?;
     let process_name = basename_or_path(&path);
     if process_name.is_empty() {
         return Err("Process name is empty".to_string());
     }
     Ok(PeerInfo::new(peer_pid, peer_pid, process_name))
 }
 
+/// Safe wrapper over `proc_pidpath(2)` (declared in the macOS `<libproc.h>`
```
neuronull: would be helpful to link to the header file itself here

apps/desktop/desktop_native/core/src/ssh_agent/peerinfo/gather.rs
```diff
@@ -25,24 +25,64 @@ pub fn get_peer_info(peer_pid: u32) -> Result<PeerInfo, String> {
...
+/// system header). Returns the absolute executable path of `pid`, or an error
+/// string describing the failure. Requires no entitlement.
```
neuronull: nit: remove the note about entitlement, as it is covered above.

apps/desktop/desktop_native/core/src/ssh_agent/peerinfo/gather.rs
```diff
@@ -25,24 +25,64 @@ pub fn get_peer_info(peer_pid: u32) -> Result<PeerInfo, String> {
...
-/// macOS-specific peer-info resolution via `libproc::proc_pid::pidpath`
-/// (the BSD `proc_pidpath(2)` syscall). This avoids the `task_for_pid`
-/// entitlement that `sysinfo::Process::name()` ends up needing on macOS,
-/// so it can still resolve the caller when `sysinfo` cannot (notably
-/// inside the Mac App Store sandbox).
+/// macOS-specific peer-info resolution via the `proc_pidpath(2)` call from
+/// the system `libproc` library. This avoids the `task_for_pid` entitlement
+/// that `sysinfo::Process::name()` ends up needing on macOS, so it can still
+/// resolve the caller when `sysinfo` cannot (notably inside the Mac App Store
+/// sandbox).
```
neuronull: nit- rephrase to something simpler like "This alternative to the sysinfo() call is permissive within sandboxed runtimes such as MAS builds.

neuronull:
> Two things for your call

1. let's keep them separate.
2. I'll validate manually before approving 👍

apps/desktop/desktop_native/core/src/ssh_agent/peerinfo/gather.rs
lnky102: Done in bf3d200 — adopted your phrasing, helper doc is now two lines.

apps/desktop/desktop_native/core/src/ssh_agent/peerinfo/gather.rs
lnky102: Done in bf3d200 — linked `libproc.h` in the Apple xnu source tree from the wrapper's doc.

apps/desktop/desktop_native/core/src/ssh_agent/peerinfo/gather.rs
lnky102: Done in bf3d200 — removed; the helper doc above already conveys the same point implicitly.

apps/desktop/desktop_native/core/src/ssh_agent/peerinfo/gather.rs
lnky102: Done in bf3d200 — trimmed the SAFETY comment to the two lines that justify the `unsafe` call (kept what `proc_pidpath` writes + that 0 signals failure).

apps/desktop/desktop_native/core/src/ssh_agent/peerinfo/gather.rs
lnky102: Done in bf3d200 — replaced the `debug_assert!` with a fallible check. An overrun now returns `Err("proc_pidpath({pid}) overran buffer: …")` in both debug and release builds; the existing fallback at the call site (`from_pid` / `get_peer_info`) logs it via `warn!` when both paths fail.

lnky102:
Pushed bf3d200 addressing the five inline points. Acknowledged on (1) — keeping v1/v2 separate, and (2) — appreciate the manual validation pass.

neuronull:
👋  hi @lnky102 , thanks for making those changes. They look good to me. I'm going through some other work now, but this PR is still on my radar- I need to do some internal builds to make it accessible as a MacOS App Store build in order to do the manual tests.

lnky102:
> 👋 hi @lnky102 , thanks for making those changes. They look good to me. I'm going through some other work now, but this PR is still on my radar- I need to do some internal builds to make it accessible as a MacOS App Store build in order to do the manual tests.

Great to know, thank you neuronull.

neuronull:
👋  hey @lnky102 , wanted to let you know that this is still very much on the radar, but we've had an internal issue that is inhibiting forward progress on this one in the short term. Appreciate your patience.

neuronull:
I have pulled main into the branch so that we can do testing against the v2. Prior to that merge, the branch didn't contain the v2's sign request implementation, so it was only possible to test the caller name changes on the v1 codepath.

neuronull:
Validated on the MAS and DMG installs.

neuronull (approved):

dani-garcia (approved):

quexten (approved):

neuronull:
Thanks for this contribution @lnky102  and for your patience getting in across the line !

PR #20684, opened 2026-05-18
[PM-2588] resolved sends root url conditionally

bmbitwarden:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-2588

## 📔 Objective

The goal of this task is to adjust this logic to serve up send.bitwarden.eu instead of vault.bitwarden.eu/#/send for sends created in the EU.  

## 📸 Screenshots

https://github.com/user-attachments/assets/aef82434-51b7-42b7-89c8-997cdce55a13


harr1424 (dismissed):

mcamirault (dismissed):

libs/common/src/platform/services/default-environment.service.ts
```diff
@@ -420,9 +422,11 @@ abstract class UrlEnvironment implements Environment {
   }
 
   getSendUrl() {
-    return this.getWebVaultUrl() === "https://vault.bitwarden.com"
-      ? "https://send.bitwarden.com/#"
-      : this.getWebVaultUrl() + "/#/send/";
+    if (this.urls.send != null) {
```
trmartin4, line 429: ❓Are there scenarios where the `getSendUrl()` method will return differently now, versus previously, for customers who have overridden their URLs?

libs/common/src/platform/services/default-environment.service.ts
```diff
@@ -95,6 +95,7 @@ export const PRODUCTION_REGIONS: RegionConfig[] = [
       notifications: "https://notifications.bitwarden.com",
       events: "https://events.bitwarden.com",
       scim: "https://scim.bitwarden.com",
+      send: "https://send.bitwarden.com/#",
```
trmartin4: 📓 We'll probably want to add this to the documentation of overrideable URLs: https://bitwarden.com/help/configure-clients-selfhost/

libs/common/src/platform/services/default-environment.service.ts
```diff
@@ -109,6 +110,7 @@ export const PRODUCTION_REGIONS: RegionConfig[] = [
       notifications: "https://notifications.bitwarden.eu",
       events: "https://events.bitwarden.eu",
       scim: "https://scim.bitwarden.eu",
+      send: "https://send.bitwarden.eu/#",
```
trmartin4: ⚠️ I think this may also need to be added as an option in the self-hosted environment selector, so customers can override it as well, right?

trmartin4 (dismissed):

libs/common/src/platform/services/default-environment.service.ts
bmbitwarden: No — self-hosted customers with overridden URLs are unaffected. The urls.send field is never saved for self-hosted environments (only base, api, identity, webVault, etc. are stored), so that branch never fires and they continue to get getWebVaultUrl() + "/#/send/" as before.

For cloud regions, the send URL is now driven by the region config (PRODUCTION_REGIONS) rather than a hardcoded vault.bitwarden.com string comparison — which is actually a fix, since the old code would have produced an incorrect fallback URL for EU region customers.

libs/common/src/platform/services/default-environment.service.ts
bmbitwarden: Self-hosted customers run everything on a single domain, so getWebVaultUrl() + "/#/send/" is always the correct derivation — there's no separate send service to point to. Happy

trmartin4: 🤔 The other URLs in this list represent distinct deployed entities.  Is there a reason why we're adding `send` here, when it is really a URL based on a path from the web vault URL?  Are there plans to host Sends independent of the web vault?

trmartin4:
@bmbitwarden I added @coroiu back as I don't want to singlehandedly approve Platform code without a review from a member of the team.

libs/common/src/platform/services/default-environment.service.ts
bmbitwarden: i talked with Matt McCabe and made this ticket https://bitwarden.atlassian.net/browse/PM-38124

coroiu: 🎨 **SUGGESTED**: Consider storing `send` as a clean base URL, with `/#` added by the getter

<details>

The other URL constants in `PRODUCTION_REGIONS` (`api`, `events`, `identity`, `icons`, `notifications`, `scim`) are clean base URLs — the routing suffix is owned by the getter (`getEventsUrl()` appends `/events`, `getScimUrl()` appends `/v2`, etc.).

`send` breaks that pattern by embedding `/#` in the data:

```typescript
send: "https://send.bitwarden.com/#",
```

A more consistent shape would be:

```typescript
// data: clean base URL
send: "https://send.bitwarden.com",

// getter owns the routing convention
getSendUrl() {
  if (this.urls.send != null) {
    return this.urls.send + "/#/";
  }
  return this.getWebVaultUrl() + "/#/send/";
}
```

Callers that do `getSendUrl() + accessId + "/" + key` would be unaffected, and downstream code that wants the origin (see Comment 3 on `receive.command.ts`) could compare to `urls.send` directly instead of parsing it through `new URL(...).origin` to strip the `/#`.

</details>

libs/common/src/platform/services/default-environment.service.ts
```diff
@@ -420,9 +422,11 @@ abstract class UrlEnvironment implements Environment {
   }
 
   getSendUrl() {
```
coroiu, line 428: 🎨 **SUGGESTED**: `getSendUrl()` doesn't follow the `getUrl()` helper pattern used by other getters

<details>

Almost every other getter (`getApiUrl`, `getEventsUrl`, `getIconsUrl`, `getIdentityUrl`, `getNotificationsUrl`) delegates to the shared `getUrl(key, baseSuffix)` helper at line 446, which provides a three-rung fallback: region config → `base + suffix` → default region config.

`getSendUrl()` (and historically `getScimUrl()`) reimplements only two of those rungs and silently drops the `base + suffix` rung that self-hosted users rely on for every other URL.

If Comment 1 is applied (clean base URL), this getter could likely fold into the helper pattern, leaving `scim` as the only outlier. Keeps the count of "URLs that don't follow the pattern" from growing.

</details>

apps/cli/src/tools/send/commands/receive.command.ts
```diff
@@ -88,9 +89,16 @@ export class SendReceiveCommand extends DownloadCommand {
   private async getApiUrl(url: URL) {
     const env = await firstValueFrom(this.environmentService.environment$);
     const urls = env.getUrls();
-    if (url.origin === "https://send.bitwarden.com") {
-      return "https://api.bitwarden.com";
-    } else if (url.origin === urls.api) {
+
+    // Check if the URL origin matches any known production region's send domain
+    const matchingRegion = PRODUCTION_REGIONS.find(
```
coroiu: ❓ **QUESTION**: Why reach into `PRODUCTION_REGIONS` directly instead of `availableRegions()`?

<details>
<summary>Details</summary>

```typescript
const matchingRegion = PRODUCTION_REGIONS.find(
  (r) => r.urls.send != null && new URL(r.urls.send).origin === url.origin,
);
```

Two questions about this:

1. `PRODUCTION_REGIONS` is imported directly from `default-environment.service.ts`, bypassing `environmentService.availableRegions()`. That means any build-time-injected regions (QA, dev) are excluded from the match — `receive.command.spec.ts:460-470` only exercises EU because it's the only other entry in `PRODUCTION_REGIONS`. Is that intentional, or should we go through `availableRegions()` so dev/QA builds can also receive sends pointed at their own send domain?

2. `new URL(r.urls.send).origin` only exists because `/#` is baked into the constant (see Comment 1). With a clean base URL, this would simplify to `r.urls.send === url.origin`.

</details>

apps/web/src/app/platform/web-environment.service.ts
```diff
@@ -116,7 +116,8 @@ export class WebEnvironmentService extends DefaultEnvironmentService {
 export class WebCloudEnvironment extends CloudEnvironment {
   constructor(config: RegionConfig, urls: Urls) {
     super(config);
-    // We override the urls to avoid CORS issues
-    this.urls = urls;
+    // We override the urls to avoid CORS issues, but preserve the send URL from the region config
+    // since the build-time envUrls don't include a send-specific domain.
+    this.urls = { send: config.urls.send, ...urls };
```
coroiu: ❓ **QUESTION**: Could `send` be plumbed through `envUrls` instead of patched in here?

<details>
<summary>Details</summary>

```typescript
this.urls = { send: config.urls.send, ...urls };
```

The comment above this explains the workaround — build-time `envUrls` doesn't include `send`, so we manually pull it from the region config to avoid losing it during the CORS-safe URL override.

If `envUrls` were extended to plumb `send` through the build pipeline naturally (the same way `api`, `identity`, etc. are wired), this special-case spread could go away. Is that infeasible for some reason, or just a follow-up?

Also worth confirming: if `urls.send` is ever set in the future, the spread order means it would override `config.urls.send` (because `urls` comes second). That's probably desirable, but worth a unit test to lock it in.

</details>

libs/common/src/platform/services/default-environment.service.ts
bmbitwarden: i just added it as an option per  your first comment

If we remove send here, this.urls.send is null for cloud environments, and getSendUrl() falls back to getWebVaultUrl() + "/#/send/"

im not sure about the plans to host Sends independent of the web vault

apps/cli/src/tools/send/commands/receive.command.ts
bmbitwarden: just made commit to switch to availableRegions()

libs/common/src/platform/services/default-environment.service.ts
trmartin4: Ah, ok.  I see.  That makes sense.  `Urls` here is not only capturing the URLs that can be configured by the user, but any URLs that we have to configure as subdomained URLs for our cloud instances.  Thank you for clarifying.

trmartin4 (commented):
My questions have been answered, thank you!  However, I'm not going to approve because that will approve on behalf of Platform, and I want to make sure that @coroiu's questions get answered and addressed as well.

coroiu (dismissed):

mcamirault (dismissed):

libs/auth/src/angular/self-hosted-env-config-dialog/self-hosted-env-config-dialog.component.ts
```diff
@@ -200,6 +206,7 @@ export class SelfHostedEnvConfigDialogComponent implements OnInit, OnDestroy {
       webVault: this.webVaultUrl.value,
       icons: this.iconsUrl.value,
       notifications: this.notificationsUrl.value,
+      send: this.sendUrl.value || null,
```
JaredSnider-Bitwarden: ⚠️ : the inconsistency for the `|| null` pattern vs the other values made me look into this. 

Per Claude:
The `|| null` here is masking a missing `formatUrl(urls.send)` call in `DefaultEnvironmentService.setEnvironment()` ([L237-244](https://github.com/bitwarden/clients/blob/main/libs/common/src/platform/services/default-environment.service.ts#L237-L244)). Every other URL field gets normalized there — `""` → `null`, trailing slashes stripped, `https://` auto-prepended — which is why none of them need a `|| null` at the call site. Without that normalization, a blank Send field would persist as `""`, and since `getUrl()` only falls through when the value `!= null`, `getSendUrl()` would return `"/#/"` instead of the expected fallback.

Can we add `urls.send = formatUrl(urls.send);` alongside the others in `setEnvironment` and drop the `|| null` here? This dialog shouldn't own URL normalization — it should be symmetric with how the other fields are handled. We also should see if the default env service setEnv test can be updated for this scenario.

libs/auth/src/angular/self-hosted-env-config-dialog/self-hosted-env-config-dialog.component.ts
JaredSnider-Bitwarden, line 39: ❓ : do we need to update the `selfHostedEnvSettingsFormValidator` logic to consider the send url?

libs/auth/src/angular/self-hosted-env-config-dialog/self-hosted-env-config-dialog.component.ts
```diff
@@ -180,6 +185,7 @@ export class SelfHostedEnvConfigDialogComponent implements OnInit, OnDestroy {
             identityUrl: urls.identity || "",
             iconsUrl: urls.icons || "",
             notificationsUrl: urls.notifications || "",
+            sendUrl: urls.send || "",
```
JaredSnider-Bitwarden, line 197: ⚠️ : Per Claude:

Hydration of the saved Send URL is broken because [`Environment.getUrls()`](https://github.com/bitwarden/clients/blob/main/libs/common/src/platform/services/default-environment.service.ts#L369-L381) wasn't updated to include `send`. `ngOnInit` reads `env.getUrls()` and patches `sendUrl: urls.send || ""`, but `urls.send` is always `undefined` here — so the field opens blank even after a successful save. Worse, if the user reopens and submits without re-typing, the `|| null` will wipe the previously stored value.

Can we add `send: this.urls.send` to `getUrls()`? Both the write side (`setEnvironment`) and the read side (`getUrls`) need to know about the new field — right now only the write side does. The existing `getUrls()` assertions in `default-environment.service.spec.ts` (lines 123, 204, 235, 272, 297, 327) should also be updated to assert the new `send` value so we don't regress on this in the future.

libs/auth/src/angular/self-hosted-env-config-dialog/self-hosted-env-config-dialog.component.html
```diff
@@ -81,6 +81,17 @@
           appInputVerbatim
         />
       </bit-form-field>
+
+      <bit-form-field>
+        <bit-label>{{ "sendUrl" | i18n }}</bit-label>
```
JaredSnider-Bitwarden, line 86: ⚠️ : this translation key was only added to web's `messages.json`, but we use this env selector on desktop and extension as well.

libs/common/src/platform/services/default-environment.service.ts
```diff
@@ -420,9 +423,12 @@ abstract class UrlEnvironment implements Environment {
...
-    return this.getWebVaultUrl() === "https://vault.bitwarden.com"
-      ? "https://send.bitwarden.com/#"
-      : this.getWebVaultUrl() + "/#/send/";
+    const url = this.getUrl("send", "/#/send/");
```
JaredSnider-Bitwarden: Claude ⚠️ **Self-hosted regression when only `webVault` is configured**

If a self-hosted user sets only `urls.webVault` (no `urls.base`, no `urls.send`) — which the self-hosted env dialog and `bw config server --webvault ...` both allow — `getSendUrl()` now returns `https://send.bitwarden.com/#/` (Bitwarden cloud) instead of their own server.

Walk-through:
- `getUrl("send", "/#/send/")` falls through to `DEFAULT_REGION_CONFIG.urls.send` = `https://send.bitwarden.com` (because both `urls.send` and `urls.base` are null).
- `!this.urls.base` is true → returns `https://send.bitwarden.com/#/`.

Previously this case returned `urls.webVault + "/#/send/"`, i.e. the user's own server.

Suggested fix: consult `urls.webVault` before falling back to the cloud default, mirroring `getScimUrl()` at line 415.

JaredSnider-Bitwarden (changes requested):
Thank you for this addition! Unfortunately, I identified several structural issues 👇

mcamirault (approved):

JaredSnider-Bitwarden (approved):
LGTM! Thank you!

addisonbeck (approved):
approving for andreas, who is out of office

bmbitwarden:
hey @coroiu - can you please take one more look?

coroiu (approved):

PR #20689, opened 2026-05-18
[Shared Unlock] [PM-35083] Implement biometrics over sdk IPC

quexten:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-35083

## 📔 Objective

Adds biometric unlock over SDK IPC. Note: This is behind a feature flag, the feature flag will only be enabled when the browser and desktop have been out for a sufficient time. Also note, this is currently not testable, since the IPC transport layer for desktop - browser is not merged yet.

Please note there are currently some dep cycles which are resolved by lazy injection. Specifically, there is a cycle of Key Service -> Biometrics (Via Electron Key Service) -> SDK -> Key Service

This will eventually be resolved, in separate work.

This also temporarily adds a noop ipc service, that is in place until platform merges the transport layer support work for browser - desktop.


apps/browser/src/key-management/biometrics/background-browser-biometrics.service.ts
```diff
@@ -67,6 +87,15 @@ export class BackgroundBrowserBiometricsService extends BiometricsService {
...
 
   async getBiometricsStatus(): Promise<BiometricsStatus> {
+    if (await this.configService().getFeatureFlag(FeatureFlag.BiometricsSDKIPC)) {
+      if (!this.nativeMessagingBackground().connected) {
+        return BiometricsStatus.DesktopDisconnected;
+      } else {
+        // Handle SDK-based biometrics status check
+        return BiometricsStatus.Available;
+      }
+    }
```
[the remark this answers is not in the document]
quexten: Yes. For the transition period, you can enable biometrics *always*. We will replace it with shared unlock with biometric integration being enabled always exactly if it is enabled in the desktop app.

libs/common/src/platform/ipc/noop-ipc.service.ts
```diff
@@ -0,0 +1,40 @@
...
+    super();
+  }
+
+  override async init() {
+    try {
+      await SdkLoadService.Ready;
+
+      const communicationBackend = new IpcCommunicationBackend({
+        send: async (_message: OutgoingMessage): Promise<void> => {
+          // Drop the message. No transport is available in this environment.
```
coroiu, line 30: nit: I would consider logging this as an error just in case

coroiu (dismissed):
Approving platform code

apps/desktop/src/key-management/biometrics/renderer-biometrics.service.ts
```diff
@@ -2,16 +2,65 @@ import { Injectable } from "@angular/core";
...
+    return toSdkBiometricsStatus(status);
+  }
+
+  async unlock_biometrics(user_id: SdkUserId): Promise<SymmetricKey | undefined> {
+    const key = await this.biometricsService.unlockWithBiometricsForUser(fromSdkUserId(user_id));
+    if (key == null) {
+      return undefined;
+    }
+
+    if (SET_USERKEY_UNLOCK) {
```
mzieniukbw, line 53: ❓ Is this related to `UnlockViaSDK` and shared unlock feature flags combined ? If it is, could this be a feature flagged code instead ? It is easier to track and fix later.

mzieniukbw (dismissed):

apps/desktop/src/key-management/biometrics/renderer-biometrics.service.ts
quexten: No, it's not part of the unlockviasdk featureflag, instead this is part of shared unlock. Biometric unlock currently does:
- Extension asks for key
  - Desktop gets key (while locked)
  - Desktop sends key to extension
  - Extension unlocks

With shared unlock, we want:
- Extension asks to unlock
  - Desktop gets key (while locked)
  - Desktop sets key
  - ...
  - Shared unlock happens to unlock extension

For this PR, we still implement the "send key back to extension path". This will be dropped and replaced with only set on the desktop, and let shared unlock handle the extension unlock once that is rolled out. The code above, is the "set the key on the desktop" part, which will be enabled after shared unlock is enabled.

I have it in the code here both to make it easier to add in the right place later, since i already had the context, i just added it in a dormant state, but also since it makes testing the future approach locally possible.

quexten:
Had to resolve conflicts

mzieniukbw (approved):

coroiu (approved):

PR #20695, opened 2026-05-18
[PM-37251] Add getStatus API method for public invite link status endpoint

r-tome:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-37251

## 📔 Objective

Add a method that wraps a call to an unauthenticated endpoint that returns the organization name, seat availability, and SSO details.

This will be used in the invite link landing page on load, before login/signup, so the client can render the right error state (link not found, no seats) or route the joining user through SSO.

[Server PR](https://github.com/bitwarden/server/pull/7656)


eliykat (changes requested):
I left some comments on the linked server PR which may require some changes here - the HTTP method at a minimum.

libs/organization-invite-link/src/services/default-organization-invite-link-api.service.ts
```diff
@@ -75,9 +75,9 @@ export class DefaultOrganizationInviteLinkApiService implements OrganizationInvi
 
   async getStatus(code: string): Promise<OrganizationInviteLinkStatusResponseModel> {
     const r = await this.apiService.send(
-      "GET",
-      `/organizations/invite-link/${code}/status`,
-      null,
+      "POST",
+      `/organizations/invite-link/status`,
```
eliykat, line 79: Do we have the orgId available at this stage? If so, we could include it in this URL so that you need the correct orgId as well as the code - much harder to guess, and is more RESTful. If not, it doesn't matter because the code is random anyway. I just can't remember whether we ended up including the orgId in the link or not.

eliykat (approved):
One suggestion but it's non-blocking and may not be applicable.

libs/organization-invite-link/src/services/default-organization-invite-link-api.service.ts
r-tome: We don't, this is the first request that is made. The client will only get the `organizationId` after the accept step

PR #20709, opened 2026-05-18
[CL-1203] Place tab overflow More button next to last visible tab

willmartian:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/CL-1203

## 📔 Objective

In both `bit-tab-group` and `bit-tab-nav-bar`, the "More" overflow button was rendered at the far-right of the tab header, separated from the tabs by a wide gap. This PR places the More button immediately after the last visible tab and tightens a few related behaviors that came out of the design change.

### `bitOverflowTrigger`

A new directive that marks an element as the trailing affordance of a `[bitOverflowList]`. The list discovers it via `contentChild`, measures its width once at first render, and subtracts that (plus a gap) from the space available to items when packing. The directive also exposes its `elementRef` so the list can toggle its `hidden` property — same mechanism it already uses for overflowed items, which means we drop the previous `visibility: hidden` + `aria-hidden` host bindings and the trigger participates in focus and the accessibility tree on identical terms.

### Trigger only consumes space when needed

The pack decision checks total item width against the *full* container width first. If everything fits without the trigger, we return all-displayed — otherwise we'd be hiding an item just to reserve room for the affordance that would summon it back.

### Focus management on overflow menu close

When a user picks an item from the overflow menu, selection logic moves the focus key manager's active item to the newly selected tab. CDK's overlay simultaneously restores DOM focus to the trigger. The next arrow key would jump from the trigger past the just-reintroduced item. Both consumers now realign the key manager to the trigger on menu close, so navigation continues from where focus actually is.

### Tab item width

Tab item host classes pick up `tw-max-w-fit` so each tab caps at its content width rather than stretching.

## 📸 Screenshots

- [ ] Tab group with all tabs visible (no overflow) — More button hidden, tabs left-aligned
- [ ] Tab group with overflow — More button immediately after the last visible tab
- [ ] Tab nav bar equivalent of both states
- [ ] Lone-tab truncation state (single tab visible with many in overflow)
- [ ] Keyboard navigation: arrow keys after selecting an overflow item land on adjacent visible tabs

### DIFF
--- a/libs/components/src/overflow-list/overflow-trigger.directive.ts
+++ b/libs/components/src/overflow-list/overflow-trigger.directive.ts
@@ -1,16 +1,13 @@
-import { Directive, ElementRef, inject, signal } from "@angular/core";
+import { Directive, ElementRef, inject } from "@angular/core";
 
 /**
  * Marks an element as the trailing affordance of a parent `[bitOverflowList]` —
  * typically a "More" button that surfaces overflowed items in a menu. The list
  * measures this element and reserves its width from the space available to items,
- * so packing accounts for the affordance instead of overlapping it.
- *
- * The directive hides itself (via `visibility: hidden` + `aria-hidden`) when the
- * list has no overflowed items, while keeping a stable layout width so the list's
- * one-shot width measurement stays accurate. Using `visibility` rather than the
- * `hidden` attribute is what avoids a feedback loop between the trigger's width
- * and the pack decision.
+ * so packing accounts for the affordance instead of overlapping it. The list also
+ * toggles the trigger's `hidden` property to mirror whether anything has actually
+ * overflowed — the trigger is rendered (and measured) on the first pass, then
+ * hidden if no items overflow.
  *
  * Place the trigger as a child of the `[bitOverflowList]` host so the list can
  * find it via `contentChild`.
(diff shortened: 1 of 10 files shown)


libs/components/src/overflow-list/overflow-trigger.directive.ts
```diff
@@ -0,0 +1,31 @@
...
+ *
+ * Place the trigger as a child of the `[bitOverflowList]` host so the list can
+ * find it via `contentChild`.
+ */
+@Directive({
+  selector: "[bitOverflowTrigger]",
+  exportAs: "bitOverflowTrigger",
+  host: {
+    "[style.visibility]": "hasOverflow() ? null : 'hidden'",
+    "[attr.aria-hidden]": "hasOverflow() ? null : 'true'",
```
BryanCunningham: I think you'd still be able to focus this button right? Do we need `inert` here or adjust the tab-index when hasOverflow is false

willmartian: I updated this to work more like the other list items, controlled by the overflow list directive

libs/components/src/tabs/tab-group/tab-group.component.html
```diff
@@ -47,29 +47,28 @@
...
+      `hidden` property when there's nothing to surface, which also takes it out
+      of focus and the accessibility tree.
+    -->
+    <button
+      #moreButton
+      bitTabListItem
+      bitOverflowTrigger
+      type="button"
+      class="tw-shrink-0"
+      tabindex="-1"
```
BryanCunningham, line 62: Do we need this tab index to be conditional? Or, are we able to get to this via left/right arrows?

willmartian: Left/right arrows. The active tab should always receive the focus upon entering the group, which is never the menu trigger.

BryanCunningham (approved):

PR #20642, opened 2026-05-13
[PM-35399] autofill chrome performance tuning (#20622)

audreyality:
## 🎟️ Tracking

N/A

## 📔 Objective
Cherry pick 6f3d3239c29966574 to `hotfix-rc-browser`

This code was also reviewed in #20622

PR #20713, opened 2026-05-18
[PM-35196] pass addedElements to checkForNewShadowRoots

addisonbeck:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-37687
https://bitwarden.atlassian.net/browse/PM-35196

## 📔 Objective

`checkForNewShadowRoots()` was scanning the entire DOM every 500ms on pages with shadow DOM, calling `chrome.dom.openOrClosedShadowRoot()` on every candidate element (div, span, section, etc.). This is a cross-process extension API call that consumed 63% of main thread time per invocation. This fix passes only the nodes added in the triggering mutation to the scan, reducing the check from O(all DOM elements) to O(added nodes).

New shadow roots attached to pre-existing elements without   accompanying DOM insertions may not be detected until the next full-scan fallback fires. I am not sure if this is a case we are considering.

### DIFF
--- a/apps/browser/src/autofill/services/dom-query.service.ts
+++ b/apps/browser/src/autofill/services/dom-query.service.ts
@@ -121,7 +121,7 @@ export class DomQueryService implements DomQueryServiceInterface {
    * This is an expensive operation that should be debounced.
    * @returns True if any new shadow roots are found that aren't being observed
    */
-  checkForNewShadowRoots = (): boolean => {
+  checkForNewShadowRoots = (addedElements?: Element[]): boolean => {
     // Short-circuit: if we have already confirmed the page has no shadow DOM,
     // skip the expensive querySelectorAll(":defined") + getShadowRoot scan entirely.
     // FIXME: this disables all checks after the page initializes; introduce a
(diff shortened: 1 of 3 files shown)


apps/browser/src/autofill/services/dom-query.service.ts
```diff
@@ -130,6 +130,27 @@ export class DomQueryService implements DomQueryServiceInterface {
...
+        const root = this.getShadowRoot(el);
+        if (root && !this.observedShadowRoots.has(root)) {
+          return true;
+        }
+        for (const child of el.querySelectorAll("*")) {
+          const childRoot = this.getShadowRoot(child);
+          if (childRoot && !this.observedShadowRoots.has(childRoot)) {
+            return true;
+          }
+        }
```
audreyality, line 148: @jprusik, @blackwood - As noted offline, it's likely that we'll want to fold the `child` iteration into the root branch and fire off `this.recursivelyQueryShadowRoots(root)` inside the if branch here. In the case where we're only processing additions, we'd be able to inspect nested closed shadow roots.

🤔 Something we might want to consider in addition is updating `observedShadowRoots` when we perform recursive queries. This could alleviate the FIXME's that we accrued in #20642. Since `observedShadowRoots` is a full set, however, it could also cause memory issues on the highly dynamic pages we're seeking to optimize.

📝 @addisonbeck - I'll be running benchmarks on these edits; you do not need to address this feedback. :)

audreyality:
@addisonbeck - Thanks for offering this patch! It went went on to inform #20871, which merged earlier this week and was released in [2026.5.1](https://github.com/bitwarden/clients/releases/tag/browser-v2026.5.1). I'm going to close this one. 🙂

PR #20718, opened 2026-05-19
[PM-36229] Route bulk-collections, move, and assigned-cipher ops to SDK

gbubemismith:
## 🎟️ Tracking
https://bitwarden.atlassian.net/browse/PM-36229

## 📔 Objective
Wire bulkUpdateCollectionsWithServer, moveManyWithServer, and getManyFromApiForOrganization through CipherSdkService behind the existing `pm-27632-cipher-crud-operations-to-sdk` feature flag. The legacy apiService paths remain unchanged when the flag is off


libs/common/src/vault/services/cipher.service.ts
```diff
@@ -674,6 +674,11 @@ export class CipherService implements CipherServiceAbstraction {
   }
 
   async getManyFromApiForOrganization(organizationId: string): Promise<CipherView[]> {
+    const useSdk = await firstValueFrom(this.sdkCipherCrudEnabled$);
```
nikwithak: Should we gate this behind the `pm-28191-cipher-admin-ops-to-sdk` flag isntead of the `crud-ops-to-sdk` flag? Reasoning is that the CRUD flag will already be enabled when this goes in, whereas the admin ops flag is not yet turned on and can gate these remaining operations.

nikwithak (changes requested):
Changes look good - just one comment about which feature flag to use to gate these. Great work!

libs/common/src/vault/services/cipher.service.ts
gbubemismith: I agree that makes sense

nikwithak (dismissed):

nikwithak (approved):

PR #20719, opened 2026-05-19
[PM-377730] Refactor unlock service & add decrypted user key method

quexten:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-37730

## 📔 Objective

Mostly a refactor of the unlock service, consolidating the implementations to make them more readable and maintainable.

Non refactor changes:
- Add the decrypted user key unlock method
  - This is temporary until we stop disposing password manager clients and will be used by shared unlock, biometric unlock
- KC now runs the unlock side effects, there is a comment as to why this is okay

### DIFF
--- a/libs/unlock/src/default-unlock.service.ts
+++ b/libs/unlock/src/default-unlock.service.ts
@@ -141,7 +141,12 @@ export class DefaultUnlockService implements UnlockService {
         decrypted_user_key: userKey.toSdk(),
       },
     });
-    this.logService.measure(startTime, "Unlock", "DefaultUnlockService", "unlockWithDecryptedUserKey");
+    this.logService.measure(
+      startTime,
+      "Unlock",
+      "DefaultUnlockService",
+      "unlockWithDecryptedUserKey",
+    );
   }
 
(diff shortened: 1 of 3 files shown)


libs/unlock/src/default-unlock.service.ts
```diff
@@ -175,30 +126,47 @@ export class DefaultUnlockService implements UnlockService {
...
-              keyConnectorUrl: {
-                url: keyConnectorUnlockData.url,
-                key_connector_key_wrapped_user_key:
-                  keyConnectorUnlockData.keyConnectorKeyWrappedUserKey,
-              },
-            },
+            method,
           });
+
+          await this.runOnUnlockSideEffects(userId, ref);
```
quexten, line 170: Please note, this makes key connector unlock run side effects. This is okay because:
- Biometrics does not apply to a new KC login
- EverHadUserKey feels unintuitive and weird not to set, but also is irrelevant for KC users. Nothing about the name or docs suggests that it is not being set on those logins, so we should set it while we have to support it
- The never lock key / auto key *should* be set for KC users if needed

Thomas-Avery (approved):

PR #20732, opened 2026-05-19
[PM-37720] Fixes org switch issue on AC collections page

JaredScar:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-37720

## 📔 Objective
Fixes issue where a refresh won't happen when switching between organizations on the admin console

### DIFF
--- a/apps/web/src/app/admin-console/organizations/collections/vault.component.spec.ts
+++ b/apps/web/src/app/admin-console/organizations/collections/vault.component.spec.ts
@@ -1,0 +1,238 @@
+import { NO_ERRORS_SCHEMA } from "@angular/core";
+import { ComponentFixture, TestBed } from "@angular/core/testing";
+import { ActivatedRoute } from "@angular/router";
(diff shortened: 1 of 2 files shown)


apps/web/src/app/admin-console/organizations/collections/vault.component.spec.ts
eliykat, line 1: **No changes required** because we're about to replace this with a v2 - but I would love to see tests on the v2 component, so here's some general feedback.

---

It's a good instinct to add tests here, although I think using TestBed to simulate the page render would allow more robust test. Otherwise we are testing very specific internals of the component which may or may not result in the actual behavior we want to see.

This would also allow for better TDD, i.e. the test could test the behaviour of having 2 orgs and switching between them, and assert what gets rendered to the page.

eliykat (approved):

eliykat:
I have spoken to QA, they will test from the feature branch.

PR #20735, opened 2026-05-19
[CL-1204] bulk actions bar

BryanCunningham:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/CL-1204

## 📔 Objective
- Adds bulk action bar component
- supreses tooltip showing when a menu is showing
- Allow for custom position in menu component

## 📸 Screenshots
image

image

image


libs/components/src/bulk-actions-bar/bulk-actions-bar.component.ts
```diff
@@ -0,0 +1,242 @@
...
+  protected handleShortcut(event: KeyboardEvent): void {
+    if (!this.visible()) {
+      return;
+    }
+
+    // Cmd+B (Mac) or Ctrl+B (Windows/Linux) — exactly one of metaKey/ctrlKey.
+    if (event.key.toLowerCase() !== "b" || event.metaKey === event.ctrlKey) {
+      return;
+    }
+    event.preventDefault();
```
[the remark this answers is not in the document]
BryanCunningham: We are skipping this fix for now. I do not believe this will be an issue

apps/browser/src/_locales/en/messages.json
```diff
@@ -2489,6 +2489,31 @@
...
     "description": "To clear something out. example: To clear browser history."
   },
+  "selected": {
+    "message": "Selected"
+  },
+  "bulkActionsBar": {
+    "message": "Bulk actions"
+  },
+  "bulkActionsBarAnnouncement": {
+    "message": "$COUNT$ item(s) selected. The bulk actions bar is now available at the bottom of the screen. Press $SHORTCUTKEY$+B to toggle focus to the bulk action bar.",
```
willmartian: I don't think we should localize part of the key combo--I would just extract all of it:

`$SHORTCUTKEY$+B` --> `$SHORTCUTKEY$+B`

libs/components/src/bulk-actions-bar/bulk-action-button.component.spec.ts
willmartian, line 1: I don't find this test suite to be very useful and would recommend removing it. What do you think? I think Angular spec tests are useful when we are testing the integration of business logic with UI. If we are just testing UI, I prefer relying on Storybook (and interaction tests when necessary).

libs/components/src/bulk-actions-bar/bulk-action-button.component.ts
```diff
@@ -0,0 +1,110 @@
...
+
+  ngAfterViewInit(): void {
+    this.labelText.set(this.label()?.nativeElement.textContent?.trim() ?? "");
+  }
+
+  focus(): void {
+    // focusVisible is not yet in TypeScript's FocusOptions but is supported in all modern browsers
+    this.elementRef.nativeElement.focus({ focusVisible: true } as FocusOptions & {
+      focusVisible?: boolean;
+    });
```
willmartian, line 70: I notice we don't use this anywhere else in the codebase--why is it needed here?

libs/components/src/bulk-actions-bar/bulk-action-button.component.ts
```diff
@@ -0,0 +1,110 @@
...
+  hostDirectives: [TooltipDirective],
+  host: {
+    "[class]": "actionClasses()",
+    "[attr.tabindex]": "tabIndex()",
+  },
+})
+export class BulkActionButtonComponent implements FocusableOption, AfterViewInit {
+  private readonly elementRef = inject<ElementRef<HTMLElement>>(ElementRef);
+  private readonly parent = inject(BULK_ACTIONS_BAR_CONTEXT, { optional: true });
+  private readonly tooltip = inject(TooltipDirective, { self: true });
```
willmartian, line 33: Rather than defining the tooltip as a host directive, injecting, and imperatively setting its value, we could just put it on a wrapper div in the template.

libs/components/src/bulk-actions-bar/bulk-action-button.component.ts
```diff
@@ -0,0 +1,110 @@
...
+        setA11yTitleAndAriaLabel({ element: el, label: this.labelText() });
+      } else {
+        el.removeAttribute("aria-label");
+      }
+    });
+  }
+
+  ngAfterViewInit(): void {
+    this.labelText.set(this.label()?.nativeElement.textContent?.trim() ?? "");
+  }
```
willmartian: Does this not work as a `computed` signal?

```ts
private labelText = computed(() => this.label()?.nativeElement.textContent?.trim() ?? "");
```

libs/components/src/bulk-actions-bar/bulk-action-button.component.ts
```diff
@@ -0,0 +1,110 @@
...
+    // content is guarded inside the tooltip component template, so nothing
+    // renders while compact is false.
+    effect(() => {
+      this.tooltip.tooltipContent.set(this.compact() ? this.labelText() : "");
+    });
+
+    effect(() => {
+      const el = this.elementRef.nativeElement;
+      if (this.compact() && this.labelText()) {
+        setA11yTitleAndAriaLabel({ element: el, label: this.labelText() });
```
willmartian, line 59: This helper fn doesn't really do much. Perhaps these attributes should just be bound in the host object?

libs/components/src/bulk-actions-bar/bulk-action-button.component.ts
```diff
@@ -0,0 +1,110 @@
...
+} from "@angular/core";
+
+import { setA11yTitleAndAriaLabel } from "../a11y/set-a11y-title-and-aria-label";
+import { IconComponent } from "../icon/icon.component";
+import { BitwardenIcon } from "../shared/icon";
+import { TooltipDirective } from "../tooltip/tooltip.directive";
+
+import { BULK_ACTIONS_BAR_CONTEXT } from "./bulk-actions-bar-context";
+
+@Component({
```
willmartian, line 20: Could you add a jsdoc that marks this as an `@internal` helper component

libs/components/src/bulk-actions-bar/bulk-action-button.component.ts
```diff
@@ -0,0 +1,110 @@
...
+  // roving-tabindex pattern: only the active item is part of the document
+  // tab order; the rest are reachable only via arrow keys.
+  readonly tabIndex = signal(-1);
+
+  private readonly label = viewChild<ElementRef<HTMLSpanElement>>("label");
+  private readonly labelText = signal("");
+
+  protected readonly compact = computed(
+    () => this.parent?.additionalActionsTrigger() === this || (this.parent?.compact() ?? false),
+  );
```
willmartian: If this state is fully derived by the parent, it should be set as an input, so that the data flow is unidirectional--that obviates the workaround with the context object

libs/components/src/bulk-actions-bar/bulk-actions-bar.component.ts
```diff
@@ -0,0 +1,254 @@
...
+
+  /** True when the wrapper is narrower than the bar's intrinsic width. */
+  readonly compact = signal(false);
+
+  private readonly shortcutKey = computed(() => {
+    const nav = this.document.defaultView?.navigator;
+    return nav?.platform?.startsWith("Mac") || /Macintosh/.test(nav?.userAgent ?? "")
+      ? "Command"
+      : "Ctrl";
+  });
```
willmartian: 1. This doesn't contain a signal, so it should be a regular variable instead of a computed signal
2. Alternative to this is to read+store KeyboardEvent.metaKey in a document level keydown listener. Do you think that is more foolproof?

libs/components/src/bulk-actions-bar/bulk-actions-bar.component.ts
```diff
@@ -0,0 +1,254 @@
...
+    return this.i18nService.t(
+      "bulkActionsBarAnnouncement",
+      this.selectedCount(),
+      this.shortcutKey(),
+    );
+  });
+
+  protected readonly barStateClasses = computed(() =>
+    this.visible() ? "tw-pointer-events-auto" : "tw-translate-y-[110%] tw-opacity-0",
+  );
```
willmartian, line 113: Maybe just put this directly in the host object instead of a property?

libs/components/src/bulk-actions-bar/bulk-actions-bar.component.ts
```diff
@@ -0,0 +1,254 @@
...
+
+  // Stashes whatever was focused on the page before the bar took focus, so
+  // a second shortcut press can restore it (the same pattern CDK Overlay
+  // uses internally).
+  private readonly previousFocus = signal<HTMLElement | null>(null);
```
willmartian, line 118: Did you look into using CDK dialog for this component? It has focus restoration built in: https://material.angular.dev/cdk/dialog/overview#focus-restoration

Worth using?

libs/components/src/bulk-actions-bar/bulk-actions-bar.component.ts
```diff
@@ -0,0 +1,254 @@
...
+    this.keyManager()?.onKeydown(event);
+  }
+
+  /** Invoke the consumer-provided callback on a primary or additional action. */
+  protected handleClick(item: BulkActionComponent | BulkAdditionalActionComponent): void {
+    if (item.disabled()) {
+      return;
+    }
+    item.action()();
+  }
```
willmartian: Do we need this method? We can't just do `(click)="item.action"` in the template?

One way or the other, we shouldn't need to check item.disabled, because that is already being bound to the disabled attribute of the button and will prevent the click event from firing

libs/components/src/bulk-actions-bar/bulk-actions-bar.component.ts
```diff
@@ -0,0 +1,254 @@
...
+    afterNextRender(() => {
+      const barEl = this.bar()?.nativeElement;
+      const wrapperEl = this.wrapper().nativeElement;
+      if (!barEl) {
+        return;
+      }
+
+      // Pin `min-width: max-content` for the read so the flex parent can't
+      // shrink the bar below content size when mounted in a constrained
+      // context. `COMPACT_THRESHOLD_BUFFER_PX` absorbs any imprecision.
```
willmartian: We just jump right into resizing logic in the constructor and I got lost for a second, maybe add a comment above the afterNextRender? Or put in a helper fn `initResizeObserver`?

libs/components/src/bulk-actions-bar/bulk-action-button.component.ts
```diff
@@ -0,0 +1,110 @@
...
+
+import { setA11yTitleAndAriaLabel } from "../a11y/set-a11y-title-and-aria-label";
+import { IconComponent } from "../icon/icon.component";
+import { BitwardenIcon } from "../shared/icon";
+import { TooltipDirective } from "../tooltip/tooltip.directive";
+
+import { BULK_ACTIONS_BAR_CONTEXT } from "./bulk-actions-bar-context";
+
+@Component({
+  selector: "button[bitBulkAction], a[bitBulkAction]",
```
willmartian: `bitBulkAction` --> `bitBulkActionButton` to be consistent with how we name other things

libs/components/src/bulk-actions-bar/bulk-additional-action.component.ts
```diff
@@ -0,0 +1,25 @@
...
+ * Declares an action that lives only in the bar's "Additional actions" menu (never in the
+ * toolbar row). Like `BulkActionComponent`, this renders no DOM of its own — the bar reads its
+ * inputs via `contentChildren` and renders the corresponding `<button bitMenuItem>` inside its
+ * own `<bit-menu>`.
+ *
+ * `icon` is **optional** here (unlike on `bit-bulk-action`) because menu items can be label-only
+ * — there's no compact icon-fallback mode for the menu.
+ */
+@Component({
+  selector: "bit-bulk-additional-action",
```
willmartian, line 15: definitely a nit pick, but I find "additional" to be vague/weird. Not sure that I have a better idea though

libs/components/src/menu/default-positions.ts
```diff
@@ -0,0 +1,19 @@
...
+export interface DefaultMenuPosition extends ConnectedPosition {
+  id: MenuPositionIdentifier;
+}
+
+/**
+ * Order matters: CDK uses the first position that fits within the viewport.
+ * Today's default is below-start; consumers can opt into any of the others
+ * via `[position]` on `bitMenuTriggerFor`.
+ */
+export const defaultPositions: DefaultMenuPosition[] = [
```
willmartian, line 14: Does this not duplicate `CONTEXT_MENU_POSITIONS` in menu trigger? What is the relationship between the two?

libs/components/src/menu/menu.component.spec.ts
```diff
@@ -99,3 +140,62 @@ describe("Menu", () => {
   imports: [MenuModule],
 })
 class TestAppComponent {}
+
+// FIXME(https://bitwarden.atlassian.net/browse/CL-764): Migrate to OnPush
+// eslint-disable-next-line @angular-eslint/prefer-on-push-component-change-detection
```
willmartian, line 153: Could you make this OnPush?

libs/components/src/tooltip/tooltip.directive.ts
```diff
@@ -65,6 +73,14 @@ export class TooltipDirective implements OnInit, OnDestroy {
   private currentDescribedByIds =
     this.elementRef.nativeElement.getAttribute("aria-describedby") || null;
 
+  constructor() {
+    effect(() => {
+      if (this.suppressed()) {
+        this.destroyTooltip();
```
willmartian, line 79: We could just make `destroyTooltip` public, right? Or do we need to do something when `suppressed` flips back to false?

libs/components/src/bulk-actions-bar/bulk-actions-bar.component.ts
willmartian, line 1: Do we need to focus trap tab when inside the bulk action bar? In the current implementation, you can tab in and out of it.

libs/components/src/bulk-actions-bar/bulk-actions-bar.component.html
```diff
@@ -0,0 +1,82 @@
...
+    {{ announcement() }}
+  </div>
+
+  <div
+    #bar
+    class="tw-mb-4 tw-flex tw-items-center tw-gap-4 tw-p-2 tw-rounded-xl tw-bg-bg-brand-stronger tw-transition-[transform,opacity] tw-duration-200 tw-ease-out"
+    [class]="barStateClasses()"
+    [attr.inert]="!visible() ? '' : null"
+    [attr.aria-hidden]="!visible() ? 'true' : null"
+    role="toolbar"
```
willmartian, line 13: Role toolbar makes sense here, but should the outer host element be role dialog?

willmartian (dismissed):
Great work thus far!

willmartian:
@claude What do you think about my above comments?

libs/components/src/bulk-actions-bar/bulk-actions-bar.component.ts
BryanCunningham: Because this isn't a dialog, no, I did not. But, it could be worth looking into in the future maybe?

libs/components/src/bulk-actions-bar/bulk-actions-bar.component.html
BryanCunningham: It should not :)

libs/components/src/bulk-actions-bar/bulk-actions-bar.component.ts
BryanCunningham: I don't think so. This is the normal pattern for toolbars

libs/components/src/tooltip/tooltip.directive.ts
BryanCunningham: We need to keep suppressed. It allows us to continually deny showing the tooltip while a menu is open. Without it, while certainly a bit of an edgecase, if you hover back over the button with the menu open, the tooltip would reappear

libs/components/src/menu/menu.component.spec.ts
BryanCunningham: _Probably_... but, do we want to just handle that in a follow up? I'm just concerned it turns into a can of worms somehow and bulk actions is delayed because of it.

libs/components/src/bulk-actions-bar/bulk-additional-action.component.ts
BryanCunningham: I choose additional here because they aren't _really_ overflow items. They could be, once we implement the more fully featured responsive solution. But, they are more like... "things we want the user to be able to do but, should not be as easily accessible". From my understanding, this space is meant for things like destructive actions. Where we actually want friction for the user to be able to accomplish those things. I'm open to changing the name but, 'additional' feels about right to me.

libs/components/src/bulk-actions-bar/bulk-actions-bar.component.html
```diff
@@ -0,0 +1,84 @@
...
+    aria-orientation="horizontal"
+    [attr.aria-label]="'bulkActionsBar' | i18n"
+    (keydown)="onToolbarKeydown($event)"
+  >
+    <!-- <div class=""></div> -->
+    <span class="tw-px-2 tw-text-sm" [class]="elementWithDividerClasses">
+      <span class="tw-font-semibold tw-text-fg-contrast">
+        {{ selectedCount() }}
```
vleague2, line 20: In the screenshots, this number looks really close to the following text. Does it need some spacing?

libs/components/src/bulk-actions-bar/bulk-action-button.component.ts
vleague2: @BryanCunningham I agree with this, considering the bar has full control over the rendering of the children buttons

libs/components/src/bulk-actions-bar/bulk-action-button.component.ts
vleague2: Pasting Claude's opinion here for easier reviewing:

> The reason for the host-directive approach: tooltips need to be aria-describedby-related to a focusable element, and the focusable element here is the `<button>` host. Wrapping the content in a tooltip-host `<div>` puts aria-describedby on the wrapper, not the button — screen readers reading the button focus won't pick up the tooltip.

libs/components/src/bulk-actions-bar/bulk-action-button.component.ts
vleague2: @BryanCunningham I don't see a Claude response for this one, can you answer?

libs/components/src/bulk-actions-bar/bulk-action-button.component.ts
vleague2: Pasting Claude's opinion here for easier reviewing:

>  It's needed precisely because FocusKeyManager programmatically calls focus() on arrow-key navigation, and without the hint, browsers may skip the focus ring on programmatic focus calls — which would break the keyboard-only nav UX completely.

I don't know if this is true.... Pretty sure the focus ring would still appear since we are doing programmatic focus elsewhere and focus rings appear. @BryanCunningham do we need it to be `focusVisible` or would the regular `.focus()` call work?

libs/components/src/bulk-actions-bar/bulk-actions-bar.component.html
```diff
@@ -0,0 +1,84 @@
...
+    class="tw-mb-4 tw-flex tw-items-center tw-gap-[17px] tw-p-2 tw-rounded-xl tw-bg-bg-brand-stronger tw-transition-[transform,opacity] tw-duration-200 tw-ease-out"
+    [class]="barStateClasses()"
+    [attr.inert]="!visible() ? '' : null"
+    [attr.aria-hidden]="!visible() ? 'true' : null"
+    role="toolbar"
+    aria-orientation="horizontal"
+    [attr.aria-label]="'bulkActionsBar' | i18n"
+    (keydown)="onToolbarKeydown($event)"
+  >
+    <!-- <div class=""></div> -->
```
vleague2: ?

libs/components/src/bulk-actions-bar/bulk-actions-bar.component.ts
vleague2: Pasting Claude's opinion here for easier reviewing:

> The host object is applied to the component element, but the classes target the inner #bar element, not the host — so it can't go on the host.

libs/components/src/menu/default-positions.ts
vleague2: Pasting Claude's opinion for easier reviewing:

> They're not duplicates of each other — defaultPositions anchors to the trigger element edges (origin Y = bottom/top of the button), CONTEXT_MENU_POSITIONS anchors to a click point (origin Y = top/bottom of a zero-size point). Different originY semantics. But the naming divergence (below-start/below-end vs below-right/below-left) is real. Worth a follow-up to align names; not blocking.

@BryanCunningham I think it is worth fixing the naming and adding some comments explaining the difference because they do look duplicative otherwise

libs/components/src/menu/menu.component.spec.ts
vleague2: I will add this to the ticket for updating the menu component.

libs/components/src/bulk-actions-bar/bulk-actions-bar.component.html
BryanCunningham: I'll double check with Emily on this

libs/components/src/bulk-actions-bar/bulk-action-button.component.ts
BryanCunningham: I was not able to get the focus ring to appear while just calling `.focus()` on the element.

libs/components/src/menu/default-positions.ts
BryanCunningham: I went ahead and merged the 2 position options arrays

libs/components/src/bulk-actions-bar/bulk-action-button.component.ts
BryanCunningham: Correct, it doesn't do anything that we can't do by just binding to the host object. But, this is how we set these in the icon button so I just went ahead and used the same pattern that we use there

libs/components/src/menu/default-positions.ts
BryanCunningham: Merged here: https://github.com/bitwarden/clients/pull/20735/commits/cf31c379dd2a5cf7ac426fd4b2310ec60fb503e8

libs/components/src/bulk-actions-bar/bulk-actions-bar.component.ts
BryanCunningham: I cleaned this up previously but I went ahead and made the logic more verbose. Previously we ONLY relied on UA sniffing. Now we make our best guess at the navigator and then set it once a key is pressed https://github.com/bitwarden/clients/pull/20735/commits/336f45ed894dbc46c6e77737b75e7da7ea11e90c

libs/components/src/bulk-actions-bar/bulk-action-button.component.ts
BryanCunningham: Makes sense. Went ahead and just drilled the input down into the children components here: https://github.com/bitwarden/clients/pull/20735/changes/212f4c4ef253ee07e2fb40efa1fbdbb37f4ca77f

libs/components/src/bulk-actions-bar/bulk-actions-bar.component.html
BryanCunningham: Rechecked the design. I did need to add a little space here. It's been updated to `tw-gap-1.5`. Thanks for the heads up!

vleague2 (approved):

apps/browser/src/_locales/en/messages.json
```diff
@@ -2489,6 +2489,31 @@
...
     "description": "To clear something out. example: To clear browser history."
   },
+  "selected": {
+    "message": "Selected"
+  },
+  "bulkActionsBar": {
+    "message": "Bulk actions"
+  },
+  "bulkActionsBarAnnouncement": {
+    "message": "$COUNT$ item(s) selected. The bulk actions bar is now available at the bottom of the screen. Press $SHORTCUTBINDING$ to toggle focus to the bulk action bar.",
```
mKoonrad, line 2499: When will you add native support for plural forms? These strings look awful and translators can't translate them in any accepted way.

mKoonrad:
Translators needs plural versions here. See comments: https://crowdin.com/editor/bitwarden-browser/26/en-pl?view=comfortable#63411

PR #20738, opened 2026-05-19
[PM-37759] Unable to use passkey that is stored in my vault to log in

jengstrom-bw:
## 🎟️ Tracking

[Jira](https://bitwarden.atlassian.net/browse/PM-37759)

## 📔 Objective

This should fix the ability to see and use passkeys to authenticate using the browser with feature flag `pm-27632-cipher-crud-operations-to-sdk` turned on.


nikwithak (dismissed):
Nice work! Hopefully we don't have any more stragglers that don't pass the sdkClient

nikwithak (approved):

PR #20752, opened 2026-05-20
[PM-37779] - Add notification bar confirmation mode for inline menu save flow

jaasen-livefront:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-37779

## 📔 Objective

Adds the infrastructure needed to open the notification bar directly in
the "Login saved" confirmation state, bypassing the normal "Save login?"
prompt. This is a prerequisite for the "Save and fill" inline menu flow
(PR 2), but is independently useful — it decouples the confirmation
notification from the save-prompt lifecycle.

**Changes:**
- `NotificationBarIframeInitData`: add `isConfirmation` and
  `confirmationData` fields
- `bar.ts`: when `isConfirmation` is set, short-circuit to the
  post-save confirmation state rather than rendering the "Save login?"
  prompt
- `overlay-notifications-content.service.ts` (abstractions +
  implementation): thread `isConfirmation` / `confirmationData` through
  from the `openNotificationBar` message to the bar iframe init data
- `notification.background.ts`: add `triggerLoginSavedNotification` —
  sends `openNotificationBar` with `isConfirmation: true` to a given
  tab; add `showLoginSavedNotification` message handler — looks up the
  tab by `senderTabId` and calls `triggerLoginSavedNotification`


jprusik (approved):
This approach works, but I kinda think we should consider decoupling the confirmation from the notification prompts down the road, especially if we see ourselves pursuing in-page toasts.

jaasen-livefront:
> This approach works, but I kinda think we should consider decoupling the confirmation from the notification prompts down the road, especially if we see ourselves pursuing in-page toasts.

Totally.

PR #20755, opened 2026-05-20
[PM-37816] Remove Getting Started tab on extension install

trmartin4:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-37816

## 📔 Objective

Removes the "Getting Started" browser tab from opening on extension install.  This was already gated to prevent opening in a development environment, and that flag was removed as well.

## 📸 Screenshots

https://github.com/user-attachments/assets/36212c89-f8a9-421f-9164-0b3b07dabc99

### DIFF
--- a/apps/browser/src/background/runtime.background.ts
+++ b/apps/browser/src/background/runtime.background.ts
@@ -12,7 +12,6 @@ import { ConfigService } from "@bitwarden/common/platform/abstractions/config/co
 import { LogService } from "@bitwarden/common/platform/abstractions/log.service";
 import { MessagingService } from "@bitwarden/common/platform/abstractions/messaging.service";
 import { MessageListener, isExternalMessage } from "@bitwarden/common/platform/messaging";
-import { devFlagEnabled } from "@bitwarden/common/platform/misc/flags";
 import { Utils } from "@bitwarden/common/platform/misc/utils";
 import { CipherType } from "@bitwarden/common/vault/enums";
 import { VaultMessages } from "@bitwarden/common/vault/enums/vault-messages.enum";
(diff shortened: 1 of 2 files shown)


trmartin4:
Note that the BIT test failures are addressed in https://github.com/bitwarden/browser-interactions-testing/pull/492, which will merge after this, to avoid tests breaking due to the tab remaining open.

addisonbeck (approved):

apps/browser/src/background/runtime.background.ts
```diff
@@ -494,10 +493,6 @@ export default class RuntimeBackground {
           this.onInstalledReason === "install" &&
           !(await firstValueFrom(this.browserInitialInstallService.extensionInstalled$))
         ) {
-          if (!devFlagEnabled("skipWelcomeOnInstall")) {
-            void BrowserApi.createNewTab("https://bitwarden.com/browser-start/");
```
addisonbeck, line 498: Not sure where to follow up and remove this from the website. Maybe we just leave it dead idk.

trmartin4: yep, I have a note in the ticket for that 👍 .  It'll need to wait until we roll this out.

RowanDudas96: Why isn't this in the latest release?

trmartin4: Hi @RowanDudas96.  The https://github.com/bitwarden/clients/releases/tag/browser-v2026.5.1 release was a patch release on top of `2026.5.0`, so it didn't include this change.

It will be included in the next full extension release.

RowanDudas96: Thanks for the info :+1:

RowanDudas96: When is the next full extension release due?

PR #20765, opened 2026-05-20
[PM-32696] Add data field to Cipher object to support blob encryption

nikwithak:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-32696

## 📔 Objective

The bulk of the work to support blob encryption is done in the SDK - however since the encrypted data is stored in `Cipher.data` instead of the CipherType fields, decrypting a blob cipher results in an empty data set. This adds the `data` field as optional to `Cipher`, so that the client can pass that data to the SDK when decrypting a blob cipher.


libs/common/src/vault/models/domain/cipher.ts
shane-melton: 🎨 I had to wrap this in a `if (obj.login != null)` check. Otherwise, `this.login` would be populated with an invalid `Login` object with a bunch of undefined properties that would cause the SDK to complain.

libs/common/src/vault/models/data/cipher.data.ts
shane-melton: 🎨 Had a similar issue here. `response.login` will be empty/undefined for blob ciphers and cause unexpected undefined properties in `LoginData` which can bubble up to the SDK and cause exceptions.

libs/common/src/vault/models/request/cipher.request.ts
```diff
@@ -57,6 +58,7 @@ export class CipherRequest {
     this.archivedDate = cipher.archivedDate;
     this.reprompt = cipher.reprompt;
     this.key = cipher.key?.encryptedString;
+    this.data = cipher.data;
 
     switch (this.type) {
```
shane-melton, line 63: 🎨 We should avoid populating the legacy fields (`login`, `identity`, etc.) for blob ciphers; otherwise the API may complain that the fields are missing required values internally.

We should be able to just check if `this.data == null` and continue with these assignments. Though we will need to make sure we still take care of attachments.

shane-melton (changes requested):
Just a few comments on things I had in my local tweaks to get blob working.

libs/common/src/vault/models/request/cipher.request.ts
nikwithak: @shane-melton Will `this.data` only ever be populated for blob ciphers? I think it will always be populated with the `data` contents of the cipher from the server, even for field-encrypted ciphers.

I went with the approach of just breaking early in the switch, but if we can guarantee that the `data` field will only ever populate for blob ciphers then your approach is certainly simpler.

shane-melton (approved):

PR #20766, opened 2026-05-21
[PM-37842] Add ProxiedMail as forwarded email alias integration

yatsenkolesh:
### 🎟️ Tracking
https://bitwarden.atlassian.net/browse/PM-37842

## Type of change
- [x] New feature development

## Objective
Add ProxiedMail as a new forwarded email alias service option in Bitwarden clients. ProxiedMail recently added an addy.io-compatible API endpoint.

### New Files:
- **libs/common/src/tools/extension/vendor/proxiedmail.ts** - Vendor metadata and extension configuration
- **libs/tools/generator/core/src/integration/proxiedmail.ts** - Integration configuration with API endpoints
- **libs/tools/generator/core/src/integration/proxiedmail.spec.ts** - Unit tests

### Modified Files:
- **libs/common/src/tools/extension/vendor/data.ts** - Added proxiedmail to Vendor enum
- **libs/common/src/tools/extension/vendor/index.ts** - Registered ProxiedMail in vendor registry
- **libs/tools/generator/core/src/data/integrations.ts** - Added to Integrations registry
- **libs/tools/generator/extensions/legacy/** - Added ProxiedMail support to legacy service

## Implementation Details
The implementation follows the exact same pattern as the addy.io integration with these specifications:
- **Service Name**: ProxiedMail
- **Base URL**: https://proxiedmail.com
- **API Endpoint**: POST `/api/v1/aliases`
- **Authentication**: Bearer token
- **Required Fields**: API token and email domain
- **Self-host Support**: No
- **Response Format**: Compatible with addy.io (`{ data: { email: string } }`)

## Testing
- Added comprehensive unit tests mirroring addy.io test suite
- TypeScript compilation passes without errors
- All changes follow existing code patterns and conventions


mcamirault:
Hey @yatsenkolesh thank you for your contribution! The first step in our contribution pipeline is the [Product Assessment](https://contributing.bitwarden.com/contributing/pull-requests/community-pr-process/#product-assessment), and only after that is complete can we move on to engineering review. I'll keep you posted here when Product has finished its assessment.

yatsenkolesh:
Hi @mcamirault,

Thanks for keeping me informed. 

Should we bring our users here in order to show the demand of Bitwarden users for that - https://community.bitwarden.com/t/proxiedmail-bitwarden-integration/95993/2 or is there anything else we can do in order for that integration to be more convincing ?

mcamirault:
Hey @yatsenkolesh , since your service follows Addy.io's API contract you should be able to use it by selecting "Addy.io" from the `Service` dropdown and setting the `Self-host server URL` to "https://proxiedmail.com" (see screenshot from the browser extension)

Screenshot 2026-05-26 at 14 01 45

Can you test this on your end to confirm that it works?

yatsenkolesh:
Hi @mcamirault

It does work this way. 
Meanwhile, we do want to simplify the user journey by giving the same experience as Addy has of the selecting the service name in a drop-down.
Otherwise we would need to instruct our users about that. 

If we can’t select the service — why we would need the drop-down at all ? 

let me know if we can give any arguments for that from our side.

thank you.

yatsenkolesh:
Also, self-hosted doesn’t work on web. So it should be merged anyway.

mcamirault:
Hi @yatsenkolesh, I appreciate the desire for simplicity, but after talking with Product we've decided to not move forward with the PR since ProxiedMail is compatible with the existing [Addy.io](https://addy.io) integration via the method from my previous comment.

As an aside, all the "Self-host server URL" field does is change the base URL used for making service calls so it doesn't necessarily mean you're self hosting. I'd agree that the label is a bit confusing in this case, and I'll start a conversation with our designers internally about improving the wording.

Thank you again for your contribution and time! Closing the PR now

retorquere:
As a user of both Bitwarden and ProxiedMail, I’d like to strongly urge the product team to reconsider this decision.

There is a strange paradox in this outcome: by adhering to the Addy.io API contract, ProxiedMail chose to minimize the changes required to bitwarden, but in doing so it has effectively lowered its chances of being officially recognized in the UI? By the reasoning given above, if proxiedmail had built a completely custom, requiring more extensive changes to bitwarden for integration, a place in the dropdown would be granted. Rewarding this by hiding the service behind a competitor's brand name feels like the wrong incentive for contributors.

Forcing ProxiedMail users to select 'Addy.io' and input a 'Self-host' URL creates a highly counterintuitive and confusing user experience for several reasons:

- **Discoverability & UX:** New or less-technical users who use ProxiedMail will look at the Bitwarden generator, see that ProxiedMail isn't listed, and assume it isn’t supported. They are highly unlikely to guess that they need to select a competitor's name and pretend it is 'self-hosted' to make it work.
- **Brand Identity:** ProxiedMail is an independent service, not a self-hosted instance of Addy.io. Labeling it as such misrepresents the service.
- **Minimal Maintenance Overhead:** Since the PR creator has already done the heavy lifting to conform to the existing API contract, the code overhead to simply add 'ProxiedMail' as a named alias pointing to that same logic in the UI is negligible, but the reduction in user friction is massive.

sodalimes:
I recently attempted the suggested workaround by selecting "addy.io" as the provider while manually setting the self-hosting URL to https://proxiedmail.com along with a custom API key. While this functional workaround allows aliases to generate, it creates a cumbersome user experience.

Currently, users must manually toggle integration values depending on whether they are using Addy.io or ProxiedMail. Since these are two independent and distinct services, forcing them to share a single configuration slot is inefficient. A dedicated, separate integration for ProxiedMail is the correct architectural approach.

@mcamirault , I strongly echo @retorquere’s suggestion: this should be implemented as a dropdown menu allowing users to explicitly select their provider. This would not only streamline the setup process but also properly pay homage to the specific work and infrastructure completed by the ProxiedMail team, rather than treating it as a generic "self-hosted" add-on.

yatsenkolesh:
Hey @mcamirault
Can you please bring the products above to the product managers attention ?

mcamirault:
I've shared this PR and the wider context with them and will let you know when I hear anything more

yatsenkolesh:
Hi @mcamirault,

How are you?

There was no feedback ?

retorquere:
What I find weird - DDG got an entry in the username generator, and from the looks of it they don't even want to be there, given how user-hostile it is to get the token.

yatsenkolesh:
Hi @mcamirault,

How are you?

There was no feedback ?

PR #20780, opened 2026-05-21
[PM-36960] fix: TOTP not copied to clipboard on stripe settings in Chrome

bensbits91:
## 🎟️ Tracking

[PM-36960](https://bitwarden.atlassian.net/browse/PM-36960)

## 📔 Objective

Fix issue preventing TOTP from being copied to the clipboard when you click the code suggested in our inline menu when using Google Chrome.

## 📸 Screenshots

N/A - This PR only addresses the copying to the clipboard. No UI changes.

[PM-36960]: https://bitwarden.atlassian.net/browse/PM-36960?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/browser/src/autofill/services/autofill.service.ts
```diff
@@ -1234,6 +1230,12 @@ export default class AutofillService implements AutofillServiceInterface {
           (fieldContainsKeyword(field, AutoFillConstants.TotpFieldNames) ||
             field.autoCompleteType === "one-time-code");
 
+        // Allow explicit TOTP fields through even when hidden — some sites (e.g. Stripe's
+        // CodePuncher) use a single low-opacity input that Bitwarden marks non-viewable.
+        if (!field.viewable && !isTotpField) {
```
jprusik, line 1235: I worry about making visibility exceptions like this. The user ostensibly interacts with something visible; so should we.

Any other approaches we might take?

jprusik (changes requested):

apps/browser/src/autofill/services/autofill.service.ts
bensbits91: Makes sense. I agree the visibility bypass was the wrong approach.

Instead of loosening the "don't fill hidden inputs" rule, I decoupled the TOTP clipboard copy from a successful fill. Now the copy runs independently of the fill, so they still get the code on their clipboard even when there's no fillable target on the page. (Similar to how passkeys work.)

Specifics:

- Reverted the `!field.viewable && !isTotpField` bypass.
- Added `getTotpCopyCode(cipher)` on `AutofillService` — resolves the code under the same premium/org gate and auto-copy setting `doAutoFill` uses.
- In `fillInlineMenuCipher`, fall back to that helper when the fill produced no TOTP.

See https://github.com/bitwarden/clients/pull/20780/commits/d4c4d99bddd25fc9f6e070f62c88701d158a2964

apps/browser/src/autofill/background/overlay.background.ts
```diff
@@ -1502,13 +1502,19 @@ export class OverlayBackground implements OverlayBackgroundInterface {
...
     });
 
-    // A no-fill leaves nothing to copy and no use to record; the prior throw aborted here.
-    if (!result.didAutofill) {
-      return;
+    // A no-fill (or a fill with no TOTP target) doesn't imply no TOTP: some sites hide the TOTP
+    // input so the fill script can't target it. Still resolve + copy when the user explicitly
+    // chose a TOTP-bearing cipher.
+    const totpCode =
+      result.didAutofill && result.totp
```
jprusik, line 1509: We've established that we don't care if the autofill occurred as a gate to copying to clipboard; so why check `didAutofill` here?

```suggestion
      !!result.totp
```

or even something like

```
      result.totp ?? await this.autofillService.getTotpCopyCode(cipher);
```

apps/browser/src/autofill/services/autofill.service.ts
```diff
@@ -441,6 +441,34 @@ export default class AutofillService implements AutofillServiceInterface {
     return await firstValueFrom(this.domainSettingsService.resolvedDefaultUriMatchStrategy$);
   }
 
+  /**
+   * Resolves a cipher's TOTP code for clipboard copy when the caller can't rely on a successful
+   * fill to produce one (e.g., a hidden TOTP input that the fill script can't target). Applies
+   * the same premium/organization gate and auto-copy setting as {@link doAutoFill}.
+   */
+  async getTotpCopyCode(cipher: CipherView): Promise<string | undefined> {
```
jprusik, line 449: I'm surprised we don't already have this somewhere 🤔

jprusik (commented):
One question, otherwise LGTM

apps/browser/src/autofill/background/overlay.background.ts
bensbits91: `AutoFillResult` is a discriminated union:

```
export type AutoFillResult =
  | { didAutofill: false }
  | { didAutofill: true; totp?: string };
```

`.totp` only exists on the `didAutofill: true` branch, so accessing `result.totp` without narrowing first gives:

```
TS2339: Property 'totp' does not exist on type 'AutoFillResult'.
  Property 'totp' does not exist on type '{ didAutofill: false; }'.
```

`result.didAutofill && result.totp` is type narrowing needed to reach `.totp`.

We could possibly:
a. Restructure `AutoFillResult` to allow `totp` on both branches, but `doAutoFill` never produces a TOTP without also filling.
b. Extract to two lines, like:
```
const filledTotp = result.didAutofill ? result.totp : undefined;
const totpCode = filledTotp ?? (await this.autofillService.getTotpCopyCode(cipher));
```

(a) seems funky. Would (b) be better/clearer than what we have currently?

apps/browser/src/autofill/services/autofill.service.ts
bensbits91: We do have `CopyCipherFieldService.copy` in `/libs/vault/src/services/copy-cipher-field.service.ts` that handles TOTP copy, but it's built for a different situation -- the user deliberately clicking the "Copy verification code" icon in a cipher in their vault. This doesn't involve the auto-copy setting because that setting doesn't apply here — the user just told us to copy.

The new `getTotpCopyCode` handles the user clicking a cipher in the inline menu. No master password prompt, no toast... must honor the "Copy TOTP automatically" setting. 

Reusing `CopyCipherFieldService.copy` would require adding flags like `silent`, `skipReprompt`, `showToast` and `respectTOTPAutoCopySetting`. This seems messy, but I'm happy to discuss.

An alternative might be to create a new shared primitive function (e.g. `resolveTOTPCode(cipher)`) that handles the shared middle (gate + fetch), and update each caller wrap it with the surrounding behavior needed for the different situations (either in this PR or as a followup). Would this be better?

jprusik (approved):
Thanks for answering my questions! 👍

PR #20782, opened 2026-05-21
[PM-24223] Add feature flagged support for password v2 registration

eligrubb:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-24223

## 📔 Objective

This PR adds support for registering new master password users using the SDK to generate new v2 encryption users. The new flow is gated behind the `pm-27278-v2-password-registration` flag. Testing related to the new flow was included. Some of the existing tests were duplicated, the idea is once the v1 flow has been deprecated, the test block can be deleted completely.

## 📸 Screenshots

https://github.com/user-attachments/assets/fd793f39-6eca-4066-9ab3-9b7bd50c4875

At the end, the demo shows that the newly created user has the new v2 encryption state and KDF settings saved to the server.


dani-garcia (dismissed):
Platform changes LGTM, but there are some file conflicts to resolve

eligrubb:
Fixed the conflicts and merged changes from main into the branch.

dani-garcia (dismissed):

libs/auth/src/angular/registration/registration-finish/default-registration-finish.service.ts
```diff
@@ -44,6 +56,45 @@ export class DefaultRegistrationFinishService implements RegistrationFinishServi
...
+      );
+
+      // The SDK call returns the
+      // - account_cryptographic_state
+      // - master_password_unlock
+      // - user_key
+      // we currently discard this as all finishRegistration flows immediately
+      // log in after a successful return. However, there is potential for
+      // setting state directly upon successful registration in the future,
+      // similar to how Key Connector and TDE handle registration.
```
JaredSnider-Bitwarden: ❓ : I'm not sure about this part of the comment:
" However, there is potential for setting state directly upon successful registration in the future, similar to how Key Connector and TDE handle registration."

Those flows are post-SSO with an authenticated userId in `activeAccount$`. Here we're pre-login with no `userId` in the response, so writing user-scoped state isn't mechanically possible — and even if we plumbed one through, persisting key material before `logIn` succeeds risks orphaned per-user state that never gets cleaned up (logout only runs for accounts that were logged in).

libs/auth/src/angular/registration/registration-finish/default-registration-finish.service.ts
```diff
@@ -44,6 +56,45 @@ export class DefaultRegistrationFinishService implements RegistrationFinishServi
...
+        passwordInputResult.salt,
+        passwordInputResult.newPassword, // String,
+        passwordInputResult.newPasswordHint, // Option<String>,
+        emailVerificationToken, // Option<String>,
+        orgSponsoredFreeFamilyPlanToken,
+        acceptEmergencyAccessInviteToken,
+        emergencyAccessId,
+        providerInviteToken,
+        providerUserId,
+      );
```
JaredSnider-Bitwarden, line 80: 🎨: Base method wires "web only" args directly into the SDK request — breaks the paired-token invariant the web override enforces.

<details>

`buildSdkRegisterRequest` accepts `orgSponsoredFreeFamilyPlanToken`, `acceptEmergencyAccessInviteToken`, `emergencyAccessId`, `providerInviteToken`, and `providerUserId` and assigns each one straight into the SDK payload (lines 147–151). The legacy `buildRegisterRequest` accepts the same args but ignores them — only the web override applies them, and only with paired-token gating (`if (acceptEmergencyAccessInviteToken && emergencyAccessId) { ... }`).

It's dormant today because the web override calls `super` with only the first 5 args, so the unconditional assignments resolve to `undefined` and get overwritten by the override's gated logic. But the base now silently violates the pairing rule if any future non-web caller passes half a pair (e.g., `providerInviteToken` without `providerUserId`) — it ships a half-pair to the SDK with no client-side guard.

We can't strip the web-only args from the base signature because `finishRegistration` (defined on the base, not overridden in web) calls `this.buildSdkRegisterRequest(...)` polymorphically and needs to forward those args to the override. The signature is required for dispatch.

**Suggested fix:** mirror the legacy base pattern — keep the args in the signature for dispatch, but ignore them in the body and let the web override own the paired gating.

```typescript
protected async buildSdkRegisterRequest(
  email: string,
  salt: string,
  masterPassword: string,
  masterPasswordHint?: string,
  emailVerificationToken?: string,
  orgSponsoredFreeFamilyPlanToken?: string, // web only
  acceptEmergencyAccessInviteToken?: string, // web only
  emergencyAccessId?: string, // web only
  providerInviteToken?: string, // web only
  providerUserId?: string, // web only
): Promise<UserMasterPasswordRegistrationRequest> {
  // Base intentionally ignores web-only args; paired-token gating lives in the web override.
  return {
    email,
    salt,
    master_password: masterPassword,
    master_password_hint: masterPasswordHint,
    email_verification_token: emailVerificationToken,
    organization_user_id: undefined,
    org_invite_token: undefined,
    org_sponsored_free_family_plan_token: undefined,
    accept_emergency_access_invite_token: undefined,
    accept_emergency_access_id: undefined,
    provider_invite_token: undefined,
    provider_user_id: undefined,
  };
}
```

This matches how the legacy `buildRegisterRequest` works today and keeps the web override as the single source of truth for paired-token rules.

</details>

apps/web/src/app/auth/core/services/registration/web-registration-finish.service.ts
```diff
@@ -76,6 +81,55 @@ export class WebRegistrationFinishService
     return masterPasswordPolicyOpts;
   }
 
+  override async buildSdkRegisterRequest(
```
JaredSnider-Bitwarden, line 76: 🎨: The mutual-exclusivity invariant between alternative invite tokens and email verification could be restated more accurately here.

<details>

The legacy override carries `// Note: the org invite token and email verification are mutually exclusive. Only one will be present.` ([web-registration-finish.service.ts:133](apps/web/src/app/auth/core/services/registration/web-registration-finish.service.ts#L133)), but the comment is misleadingly narrow — the same rule applies to all four alternative token flows, not just org invite. Each alternative invite/acceptance token replaces email verification because the user already proved control of the email by clicking the invite link.

The new SDK override is a chance to fix the framing rather than just port the old comment forward. Suggested:

```typescript
// Each alternative invite/acceptance token below (org invite, org-sponsored family plan,
// emergency access, provider) is mutually exclusive with emailVerificationToken — presence
// of any one of them proves email ownership via the server-issued invite link, so the
// standalone email verification token is not required.
```

</details>

apps/web/src/app/auth/core/services/registration/web-registration-finish.service.spec.ts
JaredSnider-Bitwarden, line 1: Thank you for adding the new tests! 

Claude and I identified the following scenarios that are untested but important:
⚠️: A handful of behavior-locking tests are missing — most importantly, the paired-token gating that drives the bulk of this service's branching logic.

<details>
<summary>Details and missing cases</summary>

The happy-path coverage for each of the five token flows is solid and snapshot-guarded, but the following invariants aren't exercised:

**1. Paired-token gating (highest priority)**

Both `buildRegisterRequest` and `buildSdkRegisterRequest` gate emergency-access and provider fields on both halves of each pair:

```typescript
if (acceptEmergencyAccessInviteToken && emergencyAccessId) { ... }
if (providerInviteToken && providerUserId) { ... }
```

Every existing test passes both halves, so the gate itself is never exercised. Please add the four half-pair cases for both the legacy and SDK blocks:

- emergency access token without `emergencyAccessId` → request fields stay `undefined`
- `emergencyAccessId` without emergency access token → request fields stay `undefined`
- provider invite token without `providerUserId` → request fields stay `undefined`
- `providerUserId` without provider invite token → request fields stay `undefined`

Example (SDK block):

```typescript
it("does not set emergency access fields when only the token is provided", async () => {
  await service.finishRegistration(
    email,
    passwordInputResult,
    undefined,
    undefined,
    acceptEmergencyAccessInviteToken, // token only, no id
    undefined,
  );

  const sdkRequest = postKeysForUserPasswordRegistration.mock.calls[0][0];
  expect(sdkRequest.accept_emergency_access_invite_token).toBeUndefined();
  expect(sdkRequest.accept_emergency_access_id).toBeUndefined();
});
```

**2. Feature flag OFF negative cross-check**

The top-level `beforeEach` ([line 70](apps/web/src/app/auth/core/services/registration/web-registration-finish.service.spec.ts#L70)) sets the flag to `false` so the legacy block runs with the flag off, but no test asserts the inverse: that the SDK path is NOT invoked when the flag is off. The default-service spec does the mirror check ([default-registration-finish.service.spec.ts:295-298](libs/auth/src/angular/registration/registration-finish/default-registration-finish.service.spec.ts#L295-L298) — asserts `makeMasterKey` is not called when the flag is on). Add the inverse here:

```typescript
it("does not invoke the SDK flow when the feature flag is off", async () => {
  // sdkService stub from SDK block, or just spy on sdkService.client$
  await service.finishRegistration(email, passwordInputResult, emailVerificationToken);
  expect(accountApiService.registerFinish).toHaveBeenCalled();
  // SDK call should never be reached
});
```

**3. Invalid UUID for `orgInvite.organizationUserId`**

The default-service spec covers invalid-UUID rejection for `emergencyAccessId` and `providerUserId` ([lines 342-372](libs/auth/src/angular/registration/registration-finish/default-registration-finish.service.spec.ts#L342-L372)), but the web override has its own `toOptionalSdkOrganizationId(orgInvite.organizationUserId)` call that also flows through `asUuid`. Add a defense-in-depth test that sets `orgInvite.organizationUserId = "not-a-uuid"` and asserts `finishRegistration` rejects — protects against schema drift from older client versions where invites might have been persisted with a non-UUID format.

**4. SDK rejection propagation**

Assert that if `post_keys_for_user_password_registration` rejects, the error surfaces from `finishRegistration`. Cheap insurance against a future try/catch silently swallowing it.

```typescript
it("propagates SDK errors", async () => {
  postKeysForUserPasswordRegistration.mockRejectedValue(new Error("sdk boom"));
  await expect(
    service.finishRegistration(email, passwordInputResult, emailVerificationToken),
  ).rejects.toThrow("sdk boom");
});
```


</details>

apps/web/src/app/auth/core/services/registration/web-registration-finish.service.ts
JaredSnider-Bitwarden: 🎨 : I also wonder if we should actually enforce this invariant with code in either the base or here or both + add tests.  Alternative invite tokens (org invite, org-sponsored family plan, emergency access, provider) are mutually exclusive with the `emailVerificationToken` — meaning a registration request should never contain both, because the invite already proves email ownership, but we don't actually enforce this in this logic. Also, apparently it's not enforced on the server like it should be (oof, that's on me) either - it's just a [linear decision tree](https://github.com/bitwarden/server/blob/main/src/Core/Auth/Models/Api/Request/Accounts/RegisterFinishRequestModel.cs#L135) keyed on existence.

JaredSnider-Bitwarden (changes requested):
Thank you for tackling this for Auth and KM. I've got a few notes below but looking mostly great so far:

apps/web/src/app/auth/core/services/registration/web-registration-finish.service.spec.ts
eligrubb: I added tests for:
- [x] Paired token gating to `web-registration-finish.service.spec.ts`, both the legacy and sdk code paths.
- [x] Feature flag OFF negative cross check test added to `default-registration-finish.service.spec.ts` legacy code path.
- [x] Invalid UUID checking test added for `orgInvite.organizationUserId` in sdk `web-registration-finish` path.
- [x] SDK rejection propagation test added to sdk path.

apps/web/src/app/auth/core/services/registration/web-registration-finish.service.ts
eligrubb: I added a in code check that throws an error if this invariant is violated for both the legacy and SDK paths, along with full test coverage. I also included your suggested comment to better explain the full context.

libs/auth/src/angular/registration/registration-finish/default-registration-finish.service.ts
eligrubb: Good call, I matched the legacy functionality and now the sdk flow also sets all web-only parameters to undefined and ignores them in the `default-registration-finish` case.

libs/auth/src/angular/registration/registration-finish/default-registration-finish.service.ts
eligrubb: Good callout, I removed the incorrect pieces from the comment to clarify.

dani-garcia (dismissed):
Platform changes lgtm!

JaredSnider-Bitwarden (dismissed):
LGTM! Thank you!

JaredSnider-Bitwarden (approved):

djsmith85 (approved):

PR #20785, opened 2026-05-21
[bre-1942] fix package lock issue

aj-bw:
## 🎟️ Tracking
[BRE-1942](https://bitwarden.atlassian.net/browse/BRE-1942)

## 📔 Objective

Fixes a broken `package-lock.json` state introduced by the Angular 21 upgrade (#19725 ) that causes the desktop native module build to fail whenever the build cache is cold.

## Root Cause

The Angular 21 upgrade stripped `resolved` and `integrity` fields from two desktop-workspace-scoped lockfile entries during a "prune desktop packages" step:

- `apps/desktop/node_modules/@napi-rs/cli` — downgraded from `3.5.1` → `3.2.0`, missing
  `resolved`/`integrity`
- `apps/desktop/node_modules/find-up` — `7.0.0`, missing `resolved`/`integrity`

The `unicorn-magic` package (a transitive dependency of `find-up@7` → `@napi-rs/cli@3.2.0`) was deleted from the lockfile entirely.

When `npm ci` runs on a cold cache, it installs `@napi-rs/cli@3.2.0` (from the lockfile). At runtime, `napi build` triggers `find-up` which ESM-imports `unicorn-magic` — which was never installed because it had no lockfile entry. Result:

```
Error [ERR_MODULE_NOT_FOUND]: Cannot find package 'unicorn-magic'
    imported from .../find-up/index.js
```

**Why this was invisible at merge time:** warm-cache runs skip the build step entirely when `steps.cache.outputs.cache-hit == 'true'`. The Angular 21 PR's build jobs all received cache hits and went green without ever exercising the build path — confirmed in [this job from that PR's CI run](https://github.com/bitwarden/clients/actions/runs/26037979841/job/76541377741?pr=19725#step:20:1)
where you can see the native module cache was restored and the build step was skipped.

## ✅ Fix

Removed the stale desktop-workspace-scoped lockfile entries and ran `npm install`, which resolved `@napi-rs/cli@3.5.1` (matching `napi/package.json`) at the root `node_modules` level with correct `resolved` and `integrity` fields. The broken `3.2.0` hoisting and its broken transitive dependency chain are gone.

**Only `package-lock.json` changed** — no source, build scripts, or application code modified.

## ⚠️ Cache note

The native module cache key is `rust-<OS>-${{ hashFiles('apps/desktop/desktop_native/**/*') }}`.
Since this PR only touches `package-lock.json`, a dummy `Cargo.toml` comment was added to bust the
cache and force a build on CI. That commit has been dropped before final review. I will post a second pr in a moment that will improve cache busting to catch this in the future

## 🧪 Verified

- Local build: `Finished dev profile [unoptimized + debuginfo] target(s) in 3.27s` on
  Node `v22.22.3` / npm `10.9.8`
- CI native build succeeded on both macOS and Windows:
  [run 26235956822](https://github.com/bitwarden/clients/actions/runs/26235956822)
- macOS DMG built and launched successfully
- Remaining CI failures are pre-existing infrastructure issues (Snapcraft outage, transient GitHub artifact upload blip) unrelated to this change

[BRE-1942]: https://bitwarden.atlassian.net/browse/BRE-1942?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/desktop/desktop_native/Cargo.toml
+++ b/apps/desktop/desktop_native/Cargo.toml
@@ -1,4 +1,5 @@
 [workspace]
+# ci-cache-bust: force native module rebuild to validate lockfile fix
 resolver = "2"
 members = [
     "autofill_provider",
--- a/apps/desktop/desktop_native/Cargo.toml
+++ b/apps/desktop/desktop_native/Cargo.toml
@@ -1,5 +1,4 @@
 [workspace]
-# ci-cache-bust: force native module rebuild to validate lockfile fix
 resolver = "2"
 members = [
     "autofill_provider",
--- a/apps/desktop/test-file.md
+++ b/apps/desktop/test-file.md
@@ -1,1 +1,0 @@
-// dummy file edit
--- a/apps/desktop/test-file.md
+++ b/apps/desktop/test-file.md
@@ -1,0 +1,1 @@
+// dummy file edit
--- a/package-lock.json
+++ b/package-lock.json
@@ -255,137 +255,6 @@
         "@napi-rs/cli": "3.5.1"
       }
     },
-    "apps/desktop/node_modules/@napi-rs/cli": {
-      "version": "3.2.0",
-      "dev": true,
-      "license": "MIT",
-      "dependencies": {
-        "@inquirer/prompts": "^7.8.4",
-        "@napi-rs/cross-toolchain": "^1.0.3",
-        "@napi-rs/wasm-tools": "^1.0.1",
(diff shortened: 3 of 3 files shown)


iinuwa (approved):
I should not be considered authoritative, but this seems legitimate and fixes the build issues for me locally.

dani-garcia (approved):

PR #20789, opened 2026-05-21
SSH Agent v2: testing coverage for unsupported key types

neuronull:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-37824

## 📔 Objective

To ensure we are properly handling errors for unsupported key types from both the BW server and ssh clients interfacing with the agent:

- Add unit tests to cover the keystore side (receiving request to add ssh key type that we dont support)
- Add integration tests to cover the server side (receive client request that specifies a key which we dont support)


quexten (dismissed):

neuronull:
I believe platform team review was requested as part of the automatic rebase to main, I'll remove

quexten (approved):

PR #20795, opened 2026-05-21
 Managed install check for displaying getting started tab

trmartin4:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-37912

## 📔 Objective

Adds detection so the Getting Started tab is suppressed for installs the user did not actively initiate. Uses `management.getSelf().installType` as the gate; the welcome tab opens for normal user-installed and developer installs and is suppressed when `installType` is `admin` (enterprise force-install), `sideload` (installed by other software on the machine), or `other` (unclassified). We show it when the install type is `normal`, `development`, or `unknown` (the browser doesn't expose the install type for us).

Changes:
- New `BrowserApi.getInstallType()` wrapping `chrome.management.getSelf()` and `browser.management.getSelf()` for cross-browser support. Safari falls back to `Unknown` since it has no managed-install concept, preserving the default welcome behavior.
- New `ExtensionInstallType` const + type alias, so call sites do not depend on the upstream `chrome.management.ExtensionInstallType` type.
- Moved all welcome page logic to the `BrowserInitialInstallService` so that it could be tested.
- `runtime.background.checkOnInstalled` now calls `displayWelcomePage()`, which opens the welcome tab only when the install is user-initiated (`Normal`, `Development`, or `Unknown`) and the `skipWelcomeOnInstall` dev flag is not set.
- Kept the existing `skipWelcomeOnInstall` dev flag so devs can opt out of the welcome tab on every unpacked reload, with the option to flip it in `local.json` when they want to exercise the welcome path.
- No new manifest permissions; `getSelf()` is the one `management` API that does not require the `"management"` permission.

### Behavior for each scenario

| Build | Install type | Dev flag (`skipWelcomeOnInstall`) | Old behavior | New behavior | Change? |
|---|---|---|---|---|---|
| Development | `Development` (unpacked) | `true` (default in `development.json`) | Suppressed | Suppressed | No |
| Development | `Development` (unpacked) | `false` (opt-in via `local.json`) | Opens | Opens | No |
| Production | `Normal` | Forced off in prod | Opens | Opens | No |
| Production | `Development` (unpacked prod build, used by CI/QA) | Forced off in prod | Opens | Opens | No |
| Production | `Unknown` (Safari, unrecognized) | Forced off in prod | Opens | Opens | No |
| Production | `Admin` (enterprise force-install) | Forced off in prod | Opens | **Suppressed** | **Yes (intended)** |
| Production | `Sideload` (installed by other software) | Forced off in prod | Opens | **Suppressed** | **Yes (intended)** |
| Production | `Other` | Forced off in prod | Opens | **Suppressed** | **Yes (intended)** |

### Why not `chrome.storage.managed` for this

`chrome.storage.managed` is the right place to expose admin-tunable settings, but it is not a reliable signal for "is this install managed" at extension install time. Why? Because there is a race at install. `onInstalled` fires the first time the service worker boots. Managed policy values are populated by Chrome's policy layer, and for cloud-delivered policy such as Chrome Browser Cloud Management, Workspace, or MDM-pushed Chrome policy there is no guarantee the values have arrived yet. `chrome.storage.managed.get()` returns `{}` rather than waiting, so an install-time read can race with policy delivery and produce a false negative.

## 📸 Screenshots

⚠️ I was not able to test either of the scenarios under which the tab is disabled, because I do not have an MDM or third-party managed environment. This demonstrates that the tab opens as expected with the dev override off:

https://github.com/user-attachments/assets/d1cba458-4b6b-4f2a-b707-e1cad4017260


dereknance (approved):
This is awesome!

PR #20805, opened 2026-05-22
[PM-30848] Key connector for user key rotation

mzieniukbw:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-30848

## 📔 Objective

Key-connector users should support key-rotation.

## 📸 Screenshots

https://github.com/user-attachments/assets/b5c8510c-8ff1-4045-86e6-25b9488a58d9


libs/key-management-ui/src/key-rotation/key-rotation-dialog.component.ts
```diff
@@ -35,30 +38,54 @@ import { KeyRotationDialogService } from "./key-rotation-dialog.service";
...
+  private readonly userDecryptionOptionsService = inject(UserDecryptionOptionsServiceAbstraction);
 
-  protected readonly submit = async () => {
-    this.form.markAllAsTouched();
-    if (this.form.invalid || !this.form.value.masterPassword) {
+  async ngOnInit() {
+    const userId = await firstValueFrom(this.accountService.activeAccount$.pipe(getUserId));
+    const usesKeyConnector = await this.keyConnectorService.getUsesKeyConnector(userId);
+
+    if (usesKeyConnector) {
```
Thomas-Avery: 💭  In the upstream we check for `hasManagingOrganization` for key connector, should we be checking here as well incase the upstream changes?

libs/key-management-ui/src/key-rotation/key-rotation-dialog.component.ts
```diff
@@ -35,30 +38,54 @@ import { KeyRotationDialogService } from "./key-rotation-dialog.service";
...
       return;
     }
+
+    const hasMasterPassword = await firstValueFrom(
+      this.userDecryptionOptionsService.hasMasterPasswordById$(userId),
+    );
+    this.isMasterPasswordEncryptionUser.set(hasMasterPassword);
+  }
+
+  protected readonly submit = async () => {
```
Thomas-Avery: Should we put in a `loading` check that makes sure the `ngOnInit` is done prior to a `submit` executing? Right now I think this can execute with `isMasterPasswordEncryptionUser` and `isKeyConnectorEncryptionUser` both being false.

Swapping to the enum described below would probably make this easier.

libs/key-management-ui/src/key-rotation/key-rotation-dialog.component.ts
```diff
@@ -69,10 +96,16 @@ export class KeyRotationDialogComponent {
         return;
       }
 
-      const closeDialog = await this.keyRotationDialogService.rotateKeys(
-        this.form.value.masterPassword,
-        userId,
-      );
+      let closeDialog = false;
```
Thomas-Avery: Optional ⛏️  would be nice to turn this into a private helper function and have the booleans be an enum so we can keep expanding to TDE.

```
  protected readonly userPrimaryEncryptionType = signal<
    "masterPassword" | "keyConnector" | "TDE" | undefined
  >(undefined);
```


```


 let closeDialog = performKeyRotation(userId, this.userPrimaryEncryptionType())
```

Thomas-Avery (changes requested):
Looking good, some things to check out.

libs/key-management-ui/src/key-rotation/key-rotation-dialog.component.ts
mzieniukbw: Fixed in https://github.com/bitwarden/clients/pull/20805/commits/805c999f5fa6c3ec921247eec862c6a958047215

libs/key-management-ui/src/key-rotation/key-rotation-dialog.component.ts
mzieniukbw: Fixed in https://github.com/bitwarden/clients/pull/20805/commits/805c999f5fa6c3ec921247eec862c6a958047215

libs/key-management-ui/src/key-rotation/key-rotation-dialog.component.ts
mzieniukbw: Fixed in https://github.com/bitwarden/clients/pull/20805/commits/805c999f5fa6c3ec921247eec862c6a958047215

libs/key-management-ui/src/key-rotation/key-rotation-dialog.component.html
```diff
@@ -3,26 +3,39 @@
...
-    </bit-form-field>
+    @if (loading()) {
+      <div class="tw-text-center">
+        <i
+          class="bwi bwi-spinner bwi-spin bwi-2x tw-text-muted"
+          title="{{ 'loading' | i18n }}"
+          aria-hidden="true"
+        ></i>
+        <span class="tw-sr-only">{{ "loading" | i18n }}</span>
+      </div>
```
Thomas-Avery, line 9: ⛏️  non blocking.

There looks to be a component for this now.  https://components.bitwarden.com/?path=/docs/component-library-spinner--docs

```
<bit-spinner></bit-spinner>
```

Thomas-Avery (dismissed):
LGTM, one non blocking thing I saw.

libs/key-management-ui/src/key-rotation/key-rotation-dialog.component.html
mzieniukbw: Fixed in https://github.com/bitwarden/clients/pull/20805/commits/55bb191c3ef7f341d68ccf167614a03b20ce68e4

Thomas-Avery (approved):
LGTM

apps/web/src/app/auth/settings/security/security-keys.component.ts
```diff
@@ -31,13 +33,21 @@ export class SecurityKeysComponent implements OnInit {
...
   ) {}
 
   async ngOnInit() {
     const userId = await firstValueFrom(this.accountService.activeAccount$.pipe(getUserId));
-    this.showChangeKdf = await firstValueFrom(
+    const hasMasterPassword = await firstValueFrom(
       this.userDecryptionOptionsService.hasMasterPasswordById$(userId),
     );
+    const usesKeyConnector = await this.keyConnectorService.getUsesKeyConnector(userId);
+    const hasManagingOrganization = usesKeyConnector
```
JaredSnider-Bitwarden, line 45: Per a question to @mzieniukbw : The hasManagingOrganization is due to:
"it's to make sure the user is not owner or admin of the organization, which should use master password for decryption, instead of key connector. Technically speaking such user should have master password, because it won't be using the KC"

JaredSnider-Bitwarden (approved):
Auth file changes make sense.

apps/web/src/app/auth/settings/security/security-keys.component.ts
JaredSnider-Bitwarden: Also:
"so while this const usesKeyConnector = await this.keyConnectorService.getUsesKeyConnector(userId);  might return true, if the getManagingOrganization returns null it might also mean user is not confirmed yet, hence can't do rotation yet"

PR #20808, opened 2026-05-22
[PM-35059] Coachmark tour for Access Intelligence

voommen-livefront:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-35059

## 📔 Objective

Modeled on the coach marks tour in Vault
Configure the dev menu to do the following:
- to begin the coach marks tour
- to view the user state of the coach marks tour
- to unset the user state of the coach marks tour
Setup a coach marks tour for access Intelligence
- Create a list of steps and and verbage for the tour
- Create a coach marks component (popover, that renders with the appropriate style)
- Create a coach marks service that begins, navigates and retrieves data for the tour
- Setup the coachmarks within the relevant components
Modify the Tab component to render the tab label directive

## 📸 Screenshots

Screenshot 2026-05-22 at 9 00 34 AM
Screenshot 2026-05-22 at 9 08 46 AM
Screenshot 2026-05-22 at 9 11 36 AM
Screenshot 2026-05-22 at 9 12 10 AM
Screenshot 2026-05-22 at 9 12 59 AM
Screenshot 2026-05-22 at 9 13 32 AM


bitwarden_license/bit-web/src/app/dirt/access-intelligence/onboarding/access-intelligence-coachmark.service.ts
jaasen-livefront, line 1: ❓ I wonder if there's a way to use the existing [coachmark service](https://github.com/bitwarden/clients/blob/main/apps/web/src/app/vault/components/coachmark/coachmark.service.ts) as a base class here and extend/modify the methods as required. Same with the coachmark component.

libs/components/src/tabs/tab-group/tab-label.directive.ts
```diff
@@ -0,0 +1,6 @@
+import { Directive } from "@angular/core";
```
BryanCunningham: Just trying to wrap my head around why we need this? Can we not just use the bit-tab component/api here? This doesn't seem necessary to me but, maybe I'm missing something?

bitwarden_license/bit-web/src/app/dirt/access-intelligence/onboarding/access-intelligence-coachmark.service.ts
voommen-livefront: @jaasen-livefront I think this is a great idea. This is surely something that can be simplified - can I request that this can be a new ticket that affects vault and access intelligence please? This PR is already 31 files.
I can create a ticket in our backlog.

libs/components/src/tabs/tab-group/tab-label.directive.ts
voommen-livefront: Hey @BryanCunningham - I know you know this better than me - so pardon my explanation if it seems scattered.
Ideally a popover should be within my component (like a div or card or button) - that's because the popover applies the `bitPopoverAnchorFor` directive on the div and can then control that element.
The tab-group actually renders a button and at this time, we cannot use it to set the popover elements.
That's why the directive came in handy. It allows us to setup a element that can render a popover and using the directive, we can render the div.

bitwarden_license/bit-web/src/app/dirt/access-intelligence/onboarding/access-intelligence-coachmark.service.spec.ts
```diff
@@ -0,0 +1,286 @@
...
+      await service.startTour(mockOrgId);
+      tick();
+      await service.nextStep();
+      tick(); // step 2
+      await service.nextStep();
+      tick(); // step 3 = criticalApplications
+      expect(service.requiredTabIndex()).toBe(2);
+    }));
+
+    it("returns null for step 5 (runReport)", fakeAsync(async () => {
```
voommen-livefront: ❓ **QUESTION**: Test name says "returns null" but the assertion expects `0`.

<details>
<summary>Details</summary>

The test is named `"returns null for step 5 (runReport)"` but asserts `expect(service.requiredTabIndex()).toBe(0)`. The `runReport` step has `tabIndex: 0` in the step config, so the assertion matches the implementation. However:

- If the intent was for the final step to have **no required tab** (`null`), then the step config should set `tabIndex: undefined` (or omit it) and the assertion should be `toBeNull()`. Currently, finishing the tour on `runReport` will force the user back to tab 0 (Activity) via the effect in `risk-insights.component.ts` and `access-intelligence-page.component.ts`, which may not be desired since `runReport` anchors to the banner above the tab group.
- If the intent was always `0`, the test name should be updated to match (e.g., `"returns 0 for step 5 (runReport)"`).

Future readers may use the test name as documentation of intent and inadvertently "fix" the implementation back to `null`, breaking the tab-switch behavior elsewhere.
</details>

voommen-livefront: corrected

bitwarden_license/bit-web/src/app/dirt/access-intelligence/onboarding/access-intelligence-coachmark.service.ts
jaasen-livefront: @voommen-livefront I understand the hesitation but it would actually reduce the number of files required. It may be worth putting claude to work on this since we've already established the requirements. If you'd rather do this in a follow-up PR I'm fine with that though.

bitwarden_license/bit-web/src/app/dirt/access-intelligence/onboarding/access-intelligence-coachmark-step.ts
```diff
@@ -0,0 +1,60 @@
...
+  position: PositionIdentifier;
+  learnMoreUrl?: string;
+  tabIndex?: number;
+}
+
+export const ACCESS_INTELLIGENCE_COACHMARK_STEPS: AccessIntelligenceCoachmarkStep[] = [
+  {
+    id: "monitorActivity",
+    titleKey: "aiCoachmarkMonitorActivityTitle",
+    descriptionKey: "aiCoachmarkMonitorActivityDescription",
```
voommen-livefront, line 26: ⚠️ **IMPORTANT**: `criticalApplications` step (`tabIndex: 2`) stalls the tour when `Milestone11AppPageImprovements` is enabled.

<details>

This step config assumes the milestone11-disabled tab layout (`[0] activity, [1] allApplications, [2] criticalApplications`). When `milestone11Enabled` is `true`, both `risk-insights.component.html` and `access-intelligence-page.component.html` render only two tabs (`[0] activity, [1] applications`), and the `#criticalAppsCoachmark` component plus the anchored critical-applications tab live exclusively in the `@else (milestone11 disabled)` branch.

When the tour advances to this step under milestone11-enabled:
- `criticalApplicationsOpen()` becomes `true`, but there is no `#criticalAppsCoachmark` popover to display, so the user sees no popover and no Next/Close button — the tour cannot be advanced or dismissed normally.
- The `requiredTabIndex` effect calls `tabIndex.set(2)`, but `bit-tab-group` clamps to index 1 in `ngAfterContentChecked` and writes `1` back through the `selectedIndex` model. The effect re-reads `requiredTabIndex() === 2 !== tabIndex() === 1` and sets `2` again, producing repeated clamp/write churn.

`AccessIntelligenceAdoptionUxImprovements` (which enables the tour) and `Milestone11AppPageImprovements` are independent flags, so this combination is reachable.

Suggested fix: derive the step list (or at least the `criticalApplications` step's presence and `tabIndex`) from `milestone11Enabled`, or skip/guard the `criticalApplications` step when its coachmark anchor is not rendered.
</details>

voommen-livefront: yes, this is intentional. Milestone 11 changes will be executed later.

bitwarden_license/bit-web/src/app/dirt/access-intelligence/onboarding/access-intelligence-coachmark-step.ts
```diff
@@ -0,0 +1,60 @@
...
+    learnMoreUrl: "https://bitwarden.com/help/access-intelligence/",
+    tabIndex: 0,
+  },
+  {
+    id: "criticalApplications",
+    titleKey: "aiCoachmarkCriticalApplicationsTitle",
+    descriptionKey: "aiCoachmarkCriticalApplicationsDescription",
+    position: "below-start",
+    learnMoreUrl: "https://bitwarden.com/help/access-intelligence/",
+    tabIndex: 2,
```
voommen-livefront, line 45: ⚠️ **IMPORTANT**: `criticalApplications` and `helpMembers` steps target a tab that does not exist when `Milestone11AppPageImprovements` is enabled.

<details>

Both steps use `tabIndex: 2`, but the critical-applications tab (and its `#criticalAppsCoachmark` / `#helpMembersCoachmark` anchors) only exist inside the `@else (milestone11Enabled)` branch in `risk-insights.component.html` and `access-intelligence-page.component.html`. When `Milestone11AppPageImprovements` is on, the tab group only renders Activity (0) and Applications (1):

- `requiredTabIndex()` returns `2`, which the tab group clamps to `1` — so the tour silently lands on the wrong tab.
- There is no popover anchor for these two steps, so the spotlight dims the page with no visible coachmark, leaving the user unable to read the step (they can only blindly click Next).
- In `risk-insights.component.ts` (v1) the two-way `[(selectedIndex)]="tabIndex"` writes the clamped value back, which the `requiredTabIndex` effect resets to `2`, risking a change-detection feedback loop.

The step config is static and not conditioned on the flag, so nothing guards this. `AccessIntelligenceAdoptionUxImprovements` (tour gate) currently defaults to `true` while `Milestone11AppPageImprovements` defaults to `false`, so this is latent today but will surface during the milestone-11 rollout.

Consider deriving the tour steps (or their `tabIndex` / inclusion) from the active feature flag, or skipping the `criticalApplications`/`helpMembers` steps when milestone 11 is enabled.
</details>

jaasen-livefront (dismissed):

bitwarden_license/bit-web/src/app/dirt/access-intelligence/onboarding/access-intelligence-coachmark-step.ts
```diff
@@ -0,0 +1,60 @@
+import { PositionIdentifier } from "@bitwarden/components";
+
+export type AccessIntelligenceCoachmarkStepId =
+  | "monitorActivity"
+  | "prioritizeRisks"
+  | "criticalApplications"
+  | "helpMembers"
+  | "runReport";
```
Banrion: 🎨 ♻️  You could use an [enum-like](https://contributing.bitwarden.com/architecture/adr/ts-deprecate-enums#the-enum-like-pattern) here to reduce tech debt later if the coachmark steps need updating.

bitwarden_license/bit-web/src/app/dirt/access-intelligence/onboarding/access-intelligence-coachmark-step.ts
```diff
@@ -0,0 +1,60 @@
...
+  tabIndex?: number;
+}
+
+export const ACCESS_INTELLIGENCE_COACHMARK_STEPS: AccessIntelligenceCoachmarkStep[] = [
+  {
+    id: "monitorActivity",
+    titleKey: "aiCoachmarkMonitorActivityTitle",
+    descriptionKey: "aiCoachmarkMonitorActivityDescription",
+    position: "below-start",
+    learnMoreUrl: "https://bitwarden.com/help/access-intelligence/",
```
Banrion: ⚠️ Learn more urls are defined in the figma. 

```suggestion
    learnMoreUrl: "https://bitwarden.com/help/access-intelligence/#activity",
```

bitwarden_license/bit-web/src/app/dirt/access-intelligence/onboarding/access-intelligence-coachmark-step.ts
```diff
@@ -0,0 +1,60 @@
...
+    tabIndex: 0,
+  },
+  {
+    id: "prioritizeRisks",
+    titleKey: "aiCoachmarkPrioritizeRisksTitle",
+    descriptionKey: "aiCoachmarkPrioritizeRisksDescription",
+    position: "above-center",
+    learnMoreUrl: "https://bitwarden.com/help/access-intelligence/",
```
Banrion: ⚠️ Learn more urls are defined in the figma. 

❓ Should the help doc have a section that covers the Review widget? I think a link to that section from here would be beneficial. I think this is a worthy addition to add, but out of scope of this pull request to update the link to that section. 

```suggestion
    learnMoreUrl: "https://bitwarden.com/help/access-intelligence/#all-applications",
```

bitwarden_license/bit-web/src/app/dirt/access-intelligence/onboarding/access-intelligence-coachmark-step.ts
```diff
@@ -0,0 +1,60 @@
...
+    tabIndex: 0,
+  },
+  {
+    id: "criticalApplications",
+    titleKey: "aiCoachmarkCriticalApplicationsTitle",
+    descriptionKey: "aiCoachmarkCriticalApplicationsDescription",
+    position: "below-start",
+    learnMoreUrl: "https://bitwarden.com/help/access-intelligence/",
```
Banrion: ⚠️ Learn more urls are defined in the figma. 

```suggestion
    learnMoreUrl: "https://bitwarden.com/help/access-intelligence/#all-applications",
```

bitwarden_license/bit-web/src/app/dirt/access-intelligence/onboarding/access-intelligence-coachmark-step.ts
```diff
@@ -0,0 +1,60 @@
...
+    tabIndex: 2,
+  },
+  {
+    id: "helpMembers",
+    titleKey: "aiCoachmarkHelpMembersTitle",
+    descriptionKey: "aiCoachmarkHelpMembersDescription",
+    position: "left-center",
+    learnMoreUrl: "https://bitwarden.com/help/access-intelligence/",
```
Banrion: ⚠️ Learn more urls are defined in the figma. 

```suggestion
    learnMoreUrl: "https://bitwarden.com/help/access-intelligence/#requesting-password-changes",
```

bitwarden_license/bit-web/src/app/dirt/access-intelligence/onboarding/access-intelligence-coachmark-step.ts
```diff
@@ -0,0 +1,60 @@
...
+    tabIndex: 2,
+  },
+  {
+    id: "runReport",
+    titleKey: "aiCoachmarkRunReportTitle",
+    descriptionKey: "aiCoachmarkRunReportDescription",
+    position: "below-center",
+    learnMoreUrl: "https://bitwarden.com/help/access-intelligence/",
```
Banrion: ⚠️ Learn more urls are defined in the figma. 

```suggestion
    learnMoreUrl: "https://bitwarden.com/help/access-intelligence/#run-the-report",
```

bitwarden_license/bit-web/src/app/dirt/access-intelligence/onboarding/post-import-modal-dialog.component.ts
```diff
@@ -13,10 +13,12 @@ import {
...
   TypographyModule,
+  DIALOG_DATA,
 } from "@bitwarden/components";
 import { LogService } from "@bitwarden/logging";
 import { I18nPipe } from "@bitwarden/ui-common";
 
+import { AccessIntelligenceCoachmarkService } from "./access-intelligence-coachmark.service";
 import { OnboardingService } from "./services/onboarding.service";
 
 export type WelcomeModalDialogData = {
```
Banrion: ⚠️ Name might be confused with the other DialogData. I suggest updating to match the current component name.

```suggestion
export type PostImportModalDialogData = {
```

bitwarden_license/bit-web/src/app/dirt/access-intelligence/onboarding/access-intelligence-coachmark.service.ts
```diff
@@ -0,0 +1,115 @@
...
+
+  async startTour(_organizationId: OrganizationId): Promise<void> {
+    const completed = await this.onboardingService.isAICoachmarkTourCompleted();
+    if (completed) {
+      return;
+    }
+    this.activeStepId.set(ACCESS_INTELLIGENCE_COACHMARK_STEPS[0].id);
+  }
+
+  async nextStep(): Promise<void> {
```
Banrion: 🎨 Not blocking. Prefer verbs for methods. Examples with matches for the "previousStep" method:

- goToNextStep | goToPreviousStep
- stepForward | stepBackward

bitwarden_license/bit-web/src/app/dirt/access-intelligence/onboarding/access-intelligence-coachmark.service.ts
```diff
@@ -0,0 +1,115 @@
...
+      (s) => s.id === this.activeStepId(),
+    );
+    if (currentIndex >= ACCESS_INTELLIGENCE_COACHMARK_STEPS.length - 1) {
+      await this.completeTour();
+    } else {
+      this.activeStepId.set(ACCESS_INTELLIGENCE_COACHMARK_STEPS[currentIndex + 1].id);
+    }
+  }
+
+  async previousStep(): Promise<void> {
```
Banrion: 🎨 Prefer verbs for methods. Examples with matches for the "nextStep" method:

- goToNextStep | goToPreviousStep
- stepForward | stepBackward

bitwarden_license/bit-web/src/app/dirt/access-intelligence/onboarding/access-intelligence-coachmark.service.ts
```diff
@@ -0,0 +1,115 @@
...
+  getStepDescription(id: AccessIntelligenceCoachmarkStepId | null): string {
+    const step = this.getStepConfig(id);
+    return step ? this.i18nService.t(step.descriptionKey) : "";
+  }
+
+  getStepLearnMoreUrl(id: AccessIntelligenceCoachmarkStepId | null): string | undefined {
+    return this.getStepConfig(id)?.learnMoreUrl;
+  }
+
+  getStepPosition(
```
Banrion: ♻️ Unused code. It can be removed

bitwarden_license/bit-web/src/app/dirt/access-intelligence/v2/access-intelligence-page/access-intelligence-page.component.spec.ts
```diff
@@ -135,6 +137,14 @@ describe("AccessIntelligencePageComponent", () => {
       queryParams: new BehaviorSubject({}),
     };
 
+    mockCoachmarkService = {
+      activeStepId: jest.fn().mockReturnValue(null),
+      tourCompleted$: new BehaviorSubject<void>(undefined),
```
Banrion: ⚠️ The t`ourCompleted$` mock is a `BehaviorSubject<void>(undefined)`, while the production service uses a plain `Subject<boolean>`. The `BehaviorSubject` emits on subscription, so the `tourCompleted$` subscriber fires once during test setup. The handler short-circuits on `!report`, so no assertion fails today, but if a future test seeds a report into the mock before `ngOnInit` runs, the dialog open path will fire unexpectedly. Consider switching the mock to `Subject` (or `ReplaySubject(0)`) to match production behavior.

bitwarden_license/bit-web/src/app/dirt/access-intelligence/v2/access-intelligence-page/access-intelligence-page.component.ts
```diff
@@ -501,4 +541,8 @@ export class AccessIntelligencePageComponent implements OnInit, OnDestroy {
       );
     }
   }
+
+  protected async beginCoachmarksTour(): Promise<void> {
+    await this.coachmarkService.startTour(this.organizationId());
```
Banrion: ⚠️ Missing `adoptionUxImprovementsEnabled` flag check for v2 flow and only reachable via the dev menu. With v2 in active testing, we want to make sure that the coachmarks tour kicks off for v2 as well

bitwarden_license/bit-web/src/app/dirt/access-intelligence/onboarding/access-intelligence-coachmark.service.ts
```diff
@@ -0,0 +1,115 @@
...
+  readonly tourCompleted$ = this.tourCompleted.asObservable();
+
+  async startTour(_organizationId: OrganizationId): Promise<void> {
+    const completed = await this.onboardingService.isAICoachmarkTourCompleted();
+    if (completed) {
+      return;
+    }
+    this.activeStepId.set(ACCESS_INTELLIGENCE_COACHMARK_STEPS[0].id);
+  }
+
```
Banrion, line 56: ⚠️ `startTour` does not guard against already-running tour while a test case calls out "does not start if already running" which passes due to the restart of the tour. The test case could be improved by stepping once after starting the tour the first time.


```suggestion
  async startTour(_organizationId: OrganizationId): Promise<void> {
    if(this.isRunning()) {
        return;
    }
    const completed = await this.onboardingService.isAICoachmarkTourCompleted();
    if (completed) {
      return;
    }
    this.activeStepId.set(ACCESS_INTELLIGENCE_COACHMARK_STEPS[0].id);
  }

```

bitwarden_license/bit-web/src/app/dirt/access-intelligence/onboarding/access-intelligence-coachmark.service.ts
Banrion: 🎨  Unnecessary `async` usage

bitwarden_license/bit-web/src/app/dirt/access-intelligence/onboarding/access-intelligence-coachmark.service.ts
```diff
@@ -0,0 +1,115 @@
...
+  private readonly i18nService = inject(I18nService);
+
+  readonly activeStepId = signal<AccessIntelligenceCoachmarkStepId | null>(null);
+
+  readonly currentStepNumber = computed(() => {
+    const activeId = this.activeStepId();
+    if (!activeId) {
+      return 0;
+    }
+    const index = ACCESS_INTELLIGENCE_COACHMARK_STEPS.findIndex((s) => s.id === activeId);
```
Banrion: ♻️ Minor debt: `currentStepNumber`, `nextStep`, and `previousStep` each independently call `findIndex` against the same array for the same active step. A private `currentStepIndex` computed signal would de-duplicate and document the relationship.

```typescript
private readonly currentStepIndex = computed(() => {
  const activeId = this.activeStepId();
  if (!activeId) return -1;
  return ACCESS_INTELLIGENCE_COACHMARK_STEPS.findIndex((s) => s.id === activeId);
});
```

bitwarden_license/bit-web/src/app/dirt/access-intelligence/risk-insights.component.ts
```diff
@@ -148,16 +172,33 @@ export class RiskInsightsComponent implements OnInit, OnDestroy {
...
+    effect(() => {
+      // coachmarks are running, so set tab index to the coachmark's required tab
+      const tabIndex = this.coachmarkService.requiredTabIndex();
+      if (tabIndex !== null && tabIndex !== this.tabIndex()) {
+        this.tabIndex.set(tabIndex);
+      }
+    });
+
+    this.coachmarkService.tourCompleted$.pipe(takeUntilDestroyed(this.destroyRef)).subscribe(() => {
+      NewApplicationsDialogComponent.open(this.dialogService, {
```
Banrion, line 200: ⚠️ Missing the no report guard while v2 has it.

bitwarden_license/bit-web/src/app/dirt/access-intelligence/onboarding/access-intelligence-coachmark.component.ts
```diff
@@ -0,0 +1,44 @@
...
+ * @example
+ * ```html
+ * <div [bitPopoverAnchorFor]="myCoachmark.popover()" [popoverOpen]="isOpen()">
+ *   Highlighted element
+ * </div>
+ * <app-ai-coachmark #myCoachmark stepId="atRiskMembers" />
+ * ```
+ */
+@Component({
+  selector: "app-ai-coachmark",
```
Banrion: ⚠️ Prefer `dirt` prefix instead of `app`

```suggestion
  selector: "dirt-ai-coachmark",
```

bitwarden_license/bit-web/src/app/dirt/access-intelligence/risk-insights.component.ts
```diff
@@ -107,8 +118,11 @@ export class RiskInsightsComponent implements OnInit, OnDestroy {
   protected milestone11Enabled: boolean = false;
   protected adoptionUxImprovementsEnabled: boolean = false;
   protected isDevMode = isDevMode();
+  protected newApplicationsCount = 0;
+  protected newApplications: ApplicationHealthReportDetail[] = [];
+  protected totalCriticalAppsCount = 0;
 
-  tabIndex: RiskInsightsTabType = RiskInsightsTabType.AllActivity;
+  protected readonly tabIndex = signal<RiskInsightsTabType>(RiskInsightsTabType.AllActivity);
```
Banrion, line 126: 👍 Good update!

Banrion (changes requested):
Thanks for the work on this! 

I think we could have separated this pull request into one to create the coachmark component and service with storybooks and a second one to utilize it. I'll keep this in mind when we have Jira stories that require us to make new components and services.

⚠️ Missing storybooks for the new coachmark component. Since this pull request is already on the larger side, we should create a follow up Jira ticket to add the storybooks for this and the dialog components if they were missed as well.

bitwarden_license/bit-web/src/app/dirt/access-intelligence/onboarding/access-intelligence-coachmark.service.ts
voommen-livefront: 👍 implemented a guard to check if the tour has started.

bitwarden_license/bit-web/src/app/dirt/access-intelligence/risk-insights.component.ts
voommen-livefront: Guard has been added (lines 197–199). The remaining difference from v2's reactive Signal approach is tracked via the TODO at line 164, which links back to this discussion for future cleanup.

Banrion (approved):
✅ LGTM and thanks for addressing the feedback, even non-blocking ones!

PR #20820, opened 2026-05-22
Shadow dom unified fix

blackwood:
## 📔 Objective

Fix `checkForNewShadowRoots` latch and nested-shadow traversal

Two bugs in `checkForNewShadowRoots`, addressed together because they
share the function and the test surface:

1. `pageContainsShadowDom` resolved at `init()` and never re-evaluated.
   On sites that attach shadow DOM after init (i.e. SPAs
   that boot their custom-element renderers post-LOAD), the short-circuit
   blocked all detection for the rest of the session.
2. The narrow scan used `el.querySelectorAll("*")`, which does not pierce
   shadow boundaries. A shadow host nested inside another shadow root was
   invisible to the scan even when the latch was on.

Changes:
- Narrow the latch short-circuit: fires only when both
  `pageContainsShadowDom` is `false` AND `addedElements` is empty.
- `markShadowDomPresent`: a named, one-way sticky-ratchet transition that
  flips the latch when the narrow scan finds a root post-init.
- `scanForNewShadowRootInSubtree`: a depth-bounded recursive helper that
  pierces nested shadow boundaries
  (`MAX_DEEP_QUERY_RECURSION_DEPTH = 4`).
- `pendingMutationAddedElements`: now `Set<Element>`; `isConnected`
  filter at the debounced flush skips dead trees.

Inherits from #20713 (pass addedElements to checkForNewShadowRoots).

### DIFF
--- a/apps/browser/src/autofill/services/abstractions/dom-query.service.ts
+++ b/apps/browser/src/autofill/services/abstractions/dom-query.service.ts
@@ -8,7 +8,7 @@ export interface DomQueryService {
   ): T[];
   updatePageContainsShadowDom(): boolean;
   checkMutationsInShadowRoots(mutations: MutationRecord[]): boolean;
-  checkForNewShadowRoots(): boolean;
+  checkForNewShadowRoots(addedElements?: Element[]): boolean;
   resetObservedShadowRoots(): void;
   queryDeepSelector(selector: string): Element | null;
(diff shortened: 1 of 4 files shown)


blackwood:
Obviated by https://github.com/bitwarden/clients/pull/20956

PR #20823, opened 2026-05-22
[PM-37228] - Add price increase callout to organization subscription page

jaasen-livefront:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-37228

## 📔 Objective

Show a price increase callout on the Organization > Subscription (cloud) page when the API returns a scheduled subscription change.

- Added `ScheduledSubscriptionResponse` model class and `scheduledSubscription` field to `BillingSubscriptionResponse` to map the new API property
- Added `priceIncreaseCalloutMessage` getter on `OrganizationSubscriptionCloudComponent` that formats the effective date and per-seat price using `DatePipe` and `formatCurrency`, with separate i18n strings for monthly vs. annually billed plans
- Rendered the message in a `bit-callout type="info"` block in the template, visible only for active subscriptions with a scheduled price change
- Added two new i18n keys: `subscriptionPriceIncreaseMonthly` and `subscriptionPriceIncreaseAnnually`

## Notes

- Assumes the shape of the response will have the following form so this will need adjustment if it differs from the implementation:

```
  ScheduledSubscription: {
    EffectiveDate: 2026-05-22T22:06:50.131Z
    Price:  10.00
}
```

## 📸 Screenshots

Screenshot 2026-05-28 at 11 41 18 AM

### DIFF
--- a/apps/web/src/app/billing/organizations/organization-subscription-cloud.component.html
+++ b/apps/web/src/app/billing/organizations/organization-subscription-cloud.component.html
@@ -81,13 +81,13 @@
           </ng-template>
         </bit-table>
       </div>
-      <ng-container *ngIf="priceIncreaseCalloutMessage">
-        <div class="tw-mt-6">
+      @if (priceIncreaseCalloutMessage) {
+        <div class="tw-mt-4">
           <bit-callout type="info">
             <span>{{ priceIncreaseCalloutMessage }}</span>
           </bit-callout>
         </div>
-      </ng-container>
+      }
     </ng-container>
 
     <ng-container *ngIf="userOrg.canEditSubscription">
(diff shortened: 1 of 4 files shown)


amorask-bitwarden:
Hi @jaasen-livefront, this PR is still in Draft with the ticket in code review and looks like there are server-side changes needed as well?

jaasen-livefront:
@amorask-bitwarden Yes I believe so, do you know what the new api request/response shape will be?

amorask-bitwarden:
> @amorask-bitwarden Yes I believe so, do you know what the new api request/response shape will be?

This ticket encompasses the work of both `server` and `clients` changes, so that would be determined by the implementation of this ticket. I think you just need to put up a matching `server` PR to get the data you need through and then you can test this change fully.

PR #20824, opened 2026-05-22
[PM-4439] Replace oidc-client-ts with oauth4webapi

harr1424:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-4439

## 📔 Objective

Much of the functionality provided by `oidc-client-ts` was not being used, and has been replaced with `oauth4webapi` which is better scoped to the functionality we depend upon. 

[AppSec dependency review](https://bitwarden.atlassian.net/browse/VULN-591)


dereknance (approved):
✅ Platform

itsadrago (approved):

PR #20831, opened 2026-05-24
[PM-37875] Automatic V2 Upgrade Migration

quexten:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-37875

## 📔 Objective

Implements automatic upgrades to v2 encryption. This excludes the cases of:
- User's primary unlock method is not master password
- User is enrolled in emergency access
- User is enrolled in account recovery
- User has corrupt ciphers
- User has corrupt private key
- User has legacy attachments

## 📸 Screenshots

https://github.com/user-attachments/assets/c038a420-3330-4d19-bd29-a6d6e5bbd90d


libs/common/src/key-management/encrypted-migrator/migrations/v2-key-rotation-migration.ts
```diff
@@ -0,0 +1,181 @@
...
+    return CryptoClient.get_key_id_for_symmetric_key(userKey.toEncoded()) == null;
+  }
+
+  private async userEnrolledInAccountRecovery(userId: UserId): Promise<boolean> {
+    const orgs = await firstValueFrom(this.organizationService.organizations$(userId));
+    return orgs.some((o) => o.resetPasswordEnrolled);
+  }
+
+  private async userHasGrantedEmergencyAccess(userId: UserId): Promise<boolean> {
+    const response = await this.apiService.send(
```
quexten: Note: It seemed out of scope to port over emergency access to be in common. We could use an SDK call here if we really want to.

libs/common/src/platform/services/sdk/default-sdk.service.ts
```diff
@@ -191,55 +199,69 @@ export class DefaultSdkService implements SdkService {
       accountCryptographicState$,
       userKey$,
       orgKeys$,
+      v2UpgradeToken$,
       SdkLoadService.Ready, // Makes sure we wait (once) for the SDK to be loaded
     ]).pipe(
+      // Do not emit when multiple state values are written in quick succession
+      debounceTime(100),
       // switchMap is required to allow the clean-up logic to be executed when `combineLatest` emits a new value.
-      switchMap(([env, account, kdfParams, accountCryptographicState, userKey, orgKeys]) => {
```
quexten, line 196: Huge git diff but all that changed here is passing in the v2 token

libs/common/src/key-management/encrypted-migrator/migrations/v2-key-rotation-migration.ts
```diff
@@ -0,0 +1,169 @@
+import { firstValueFrom } from "rxjs";
+
+import { assertNonNullish } from "@bitwarden/common/auth/utils";
+import { SdkService } from "@bitwarden/common/platform/abstractions/sdk/sdk.service";
```
quexten: Relative import

libs/common/src/key-management/encrypted-migrator/migrations/v2-key-rotation-migration.ts
```diff
@@ -0,0 +1,169 @@
...
+    assertNonNullish(userId, "userId");
+    if (!(await this.configService.getFeatureFlag(FeatureFlag.ForceUpgradeV2Encryption))) {
+      return "noMigrationNeeded";
+    }
+
+    if (!(await this.userKeyIsV1(userId))) {
+      return "noMigrationNeeded";
+    }
+
+    this.logService.info("checking for v1 attachments that would be lost in rotation");
```
quexten: remove

libs/common/src/key-management/encrypted-migrator/migrations/v2-key-rotation-migration.ts
```diff
@@ -0,0 +1,169 @@
...
+    await this.syncService.fullSync(false);
+
+    if (!(await this.userKeyIsV1(userId))) {
+      this.logService.info(
+        `[V2KeyRotationMigration] After syncing, user ${userId} is already on v2. Skipping.`,
+      );
+      return "noMigrationNeeded";
+    }
+
+    return "needsMigrationWithMasterPassword";
```
quexten, line 100: Add check for has master password.

libs/common/src/key-management/encrypted-migrator/migrations/v2-key-rotation-migration.ts
```diff
@@ -0,0 +1,169 @@
...
+    assertNonNullish(userId, "userId");
+
+    this.logService.info(
+      `[V2KeyRotationMigration] Performing full sync before v2 upgrade for user ${userId}`,
+    );
+    await this.syncService.fullSync(true);
+
+    this.logService.info(`[V2KeyRotationMigration] Rotating user key for user ${userId}`);
+    const success = await this.userKeyRotationService.rotateUserKey(
+      { Password: { password: masterPassword! } },
```
quexten: add code for kc / tde.

libs/common/src/key-management/encrypted-migrator/migrations/v2-key-rotation-migration.spec.ts
```diff
@@ -0,0 +1,288 @@
+import { mock } from "jest-mock-extended";
+import { of } from "rxjs";
+
+import { SdkService } from "@bitwarden/common/platform/abstractions/sdk/sdk.service";
```
quexten: relative import

libs/common/src/key-management/encrypted-migrator/migrations/v2-key-rotation-migration.spec.ts
```diff
@@ -0,0 +1,288 @@
...
+  const makeCipherWithAttachments = (attachments: AttachmentView[]): CipherView => {
+    const cipher = new CipherView();
+    cipher.attachments = attachments;
+    return cipher;
+  };
+
+  const makeAttachment = (hasEncryptedKey: boolean): AttachmentView => {
+    const a = new AttachmentView();
+    a.encryptedKey = hasEncryptedKey
+      ? new EncString("0.abc|def|ghi")
```
quexten: 2.

libs/user-crypto-management/src/user-key-rotation.service.ts
```diff
@@ -42,30 +41,22 @@ export class DefaultUserKeyRotationService implements UserKeyRotationService {
...
+          password: newMasterPassword,
+          hint: hint,
+          trusted_emergency_access_public_keys: trustedEmergencyAccessUserPublicKeys,
+          trusted_organization_public_keys: trustedOrganizationPublicKeys,
+        } as PasswordChangeAndRotateUserKeysRequest);
+        return true;
+      } catch (error) {
+        this.logService.error(`Failed to rotate user keys: ${error}`);
+        return false;
+      }
```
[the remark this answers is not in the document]
quexten: It's not silent, it's logging it to the console.

libs/common/src/platform/services/sdk/default-sdk.service.ts
```diff
@@ -190,55 +197,69 @@ export class DefaultSdkService implements SdkService {
       accountCryptographicState$,
       userKey$,
       orgKeys$,
+      v2UpgradeToken$,
       SdkLoadService.Ready, // Makes sure we wait (once) for the SDK to be loaded
     ]).pipe(
+      // Do not emit when multiple state values are written in quick succession
```
quexten, line 203: Noting: The purpose here is that we are currently setting a lot of values to state during sync, and each would trigger an emission here. We want to deduplicate this to just one emission.

nick-livefront (dismissed):

libs/common/src/key-management/encrypted-migrator/migrations/v2-key-rotation-migration.ts
```diff
@@ -0,0 +1,169 @@
...
+    if (await this.userHasCorruptCiphers(userId)) {
+      this.logService.info(
+        `[V2KeyRotationMigration] User ${userId} has corrupt ciphers that cannot be decrypted. Skipping.`,
+      );
+      return "noMigrationNeeded";
+    }
+
+    // Sync first so a rotation already performed on another client is reflected
+    // here before we prompt the user.
+    await this.syncService.fullSync(false);
```
mzieniukbw: (nit) 🤔  I think it would be more correct to sync just after first "userKeyIsV1" optimistic check (before `userHasV1Attachments`, `userHasCorruptedPrivateKey` or `userHasCorruptCiphers`)

libs/common/src/key-management/encrypted-migrator/migrations/v2-key-rotation-migration.ts
```diff
@@ -0,0 +1,169 @@
...
+    // Currently not supported for users that have emergency access enabled
+    if (await this.userHasGrantedEmergencyAccess(userId)) {
+      this.logService.info(
+        `[V2KeyRotationMigration] User ${userId} has granted emergency access. Skipping.`,
+      );
+      return "noMigrationNeeded";
+    }
+
+    if (!(await this.masterPasswordService.userHasMasterPassword(userId))) {
+      // KC / TDE not supported in initial PR
```
mzieniukbw: ❓ The TDE + MP is valid combination.
Is the intention to exclude the TDE users or to allow MP decryption only users ?

libs/common/src/key-management/encrypted-migrator/migrations/v2-key-rotation-migration.ts
```diff
@@ -0,0 +1,169 @@
...
+    }
+  }
+
+  async runMigrations(userId: UserId, masterPassword: string | null): Promise<void> {
+    assertNonNullish(userId, "userId");
+
+    this.logService.info(
+      `[V2KeyRotationMigration] Performing full sync before v2 upgrade for user ${userId}`,
+    );
+    await this.syncService.fullSync(true);
```
mzieniukbw: ❓ Why do we need it ? From what i know the user key rotation does not use local state data, but syncs on demand for the key rotation. This one feels completely unnecessary.

libs/common/src/key-management/encrypted-migrator/migrations/v2-key-rotation-migration.ts
```diff
@@ -0,0 +1,169 @@
...
+
+    this.logService.info(`[V2KeyRotationMigration] Rotating user key for user ${userId}`);
+    const success = await this.userKeyRotationService.rotateUserKey(
+      { Password: { password: masterPassword! } },
```
mzieniukbw: Password key rotation method with null master password is invalid. We should assert the nullability of master password. (or abort)

libs/common/src/key-management/encrypted-migrator/migrations/v2-key-rotation-migration.ts
```diff
@@ -0,0 +1,169 @@
...
+
+    if (!success) {
+      throw new Error("[V2KeyRotationMigration] Rotation aborted by user trust prompt.");
+    }
+  }
+
+  private async userKeyIsV1(userId: UserId): Promise<boolean> {
+    const userKey = await firstValueFrom(this.keyService.userKey$(userId));
+    if (userKey == null) {
+      return false;
```
mzieniukbw: This should not be valid, ever. Throw exception instead ?

libs/common/src/key-management/encrypted-migrator/migrations/v2-key-rotation-migration.ts
```diff
@@ -0,0 +1,169 @@
...
+    }
+    return userKey.inner().type !== EncryptionType.CoseEncrypt0;
```
mzieniukbw: I would revert this statement for safety: check for `AesCbc256_HmacSha256_B64` equality instead.

libs/common/src/key-management/encrypted-migrator/migrations/v2-key-rotation-migration.ts
```diff
@@ -0,0 +1,169 @@
...
+
+    // Currently not supported for users that have emergency access enabled
+    if (await this.userHasGrantedEmergencyAccess(userId)) {
+      this.logService.info(
+        `[V2KeyRotationMigration] User ${userId} has granted emergency access. Skipping.`,
+      );
+      return "noMigrationNeeded";
+    }
+
+    if (!(await this.masterPasswordService.userHasMasterPassword(userId))) {
```
mzieniukbw: Since `userEnrolledInAccountRecovery` and `userHasGrantedEmergencyAccess` are expensive (Api request), how about we check for master password first ?

libs/common/src/key-management/encrypted-migrator/migrations/v2-key-rotation-migration.ts
```diff
@@ -0,0 +1,169 @@
...
+    });
+  }
+
+  private async userHasCorruptedPrivateKey(userId: UserId): Promise<boolean> {
+    return await withPasswordManagerSdk(userId, this.sdkService, async (sdk) => {
+      return await sdk.user_crypto_management().should_regenerate_public_key_encryption_key_pair();
+    });
+  }
+
+  private async userHasV1Attachments(userId: UserId): Promise<boolean> {
```
mzieniukbw, line 162: (nit) This is already part of SDK key rotation. I understand that we want to fail early for those users ? (if so, i don't mind this)

libs/common/src/key-management/encrypted-migrator/migrations/v2-key-rotation-migration.ts
```diff
@@ -0,0 +1,169 @@
...
+    const ciphers = await firstValueFrom(this.cipherService.cipherViews$(userId));
+    return (
+      ciphers != null && ciphers.some((c) => c.attachments?.some((a) => a.isLegacyAttachment()))
```
mzieniukbw: We need to exclude organization ciphers. (pseudo code, check if correct)
```suggestion
      ciphers != null && ciphers.filter(c => c.organization_id == null).some((c) => c.attachments?.some((a) => a.isLegacyAttachment()))
```

libs/common/src/key-management/encrypted-migrator/migrations/v2-key-rotation-migration.ts
```diff
@@ -0,0 +1,169 @@
...
+    );
+  }
+
+  private async userHasCorruptCiphers(userId: UserId): Promise<boolean> {
+    const ciphers = await firstValueFrom(this.cipherService.failedToDecryptCiphers$(userId));
```
mzieniukbw, line 176: (nit) Note, the docs says the ciphers have to be decrypted at least once:
> Does not emit until decryption has completed.

We do check for `userHasV1Attachments` first, which calls `cipherViews$`, then this is fine.
I just wonder that it might be safer to pipe the `cipherViews$(userId)` and concatMap on `failedToDecryptCiphers$`, just in case the order get's reversed - unless we say this is fine, since SDK will error during key rotation anyway ?

mzieniukbw (changes requested):

libs/common/src/key-management/encrypted-migrator/migrations/v2-key-rotation-migration.ts
quexten: I cleaned this up a bit and added a "isUserOwnedCipher" get accessor to the cipherview, which makes this a little better to read.

libs/common/src/key-management/encrypted-migrator/migrations/v2-key-rotation-migration.ts
quexten: The *mainDecryptionMethod* being TDE, or KC is out of scope for the ticket. We'll add these in subsequent tickets. TDE as a secondary decryption method is in scope for this PR.

libs/common/src/key-management/encrypted-migrator/migrations/v2-key-rotation-migration.ts
quexten: Yeah, you are right. I incorrectly assumed it would read from state.

libs/common/src/key-management/encrypted-migrator/migrations/v2-key-rotation-migration.ts
quexten: This does mean that every unlock for a user with a corrupt private key, or corrupt ciphers would always sync, so as long as we don't support fixing up these users, they would have slower unlocks.

I'm ok with that, but something to consider.

libs/common/src/key-management/encrypted-migrator/migrations/v2-key-rotation-migration.ts
quexten: Updated the function to better untangle the flows and assert non null for the MP flow.

libs/common/src/key-management/encrypted-migrator/migrations/v2-key-rotation-migration.ts
quexten: Sure, though before real production, we will remove the checks anyways, since we have to implement all flows.

libs/common/src/key-management/encrypted-migrator/migrations/v2-key-rotation-migration.ts
quexten: Yes, the migrations should correctly state when they cannot migrate, not just fail during the migration.

libs/common/src/key-management/encrypted-migrator/migrations/v2-key-rotation-migration.ts
quexten: Fair. Added a forced decryption.

libs/common/src/key-management/encrypted-migrator/migrations/v2-key-rotation-migration.ts
mzieniukbw: Fair, let's leave it then.

mzieniukbw (dismissed):

mzieniukbw (approved):

nick-livefront (approved):

djsmith85 (approved):

PR #20838, opened 2026-05-24
[PM-38027] Fix local IP autofill URI matching by port

xrvk:
## 🎟️ Tracking

Community contribution, addresses:
- https://community.bitwarden.com/t/uri-match-detection-for-ports/54661
- https://community.bitwarden.com/t/client-uri-not-recognizing-ports-for-locally-hosted-applications-qol/49975

ID: PM-38027  
Link: https://bitwarden.atlassian.net/browse/PM-38027

## 📔 Objective

Fix incorrect autofill suggestions for local and self-hosted services that share the same host but run on different ports.

Issue:
- With default URI match strategy (`Domain`), local URIs like `localhost`, private IPv4 addresses, and IPv6 loopback effectively matched by host identity only.
- Multiple credentials on the same host but different ports could all match the current page, causing the wrong login to be suggested first.

Root cause:
- For local hosts, domain resolution collapses to the host itself (`localhost` or IP), so port-specific services were not distinguished during domain matching.

What changed:
- `LoginUriView.matchesDomain` now requires matching ports when both saved and target URIs are IP-based or `localhost`.
- Local helper logic for local-host detection and explicit port extraction is colocated in `login-uri.view.ts` (per review feedback), rather than in shared `Utils`.
- Added test coverage for:
  - IPv4 private ranges
  - IPv6 loopback
  - localhost
  - matching and non-matching ports
  - explicit vs implicit ports
  - unchanged behavior for regular domains

Behavior impact:
- Local services on the same host are now correctly disambiguated by port during domain matching.
- Existing domain matching behavior for regular web domains remains unchanged.

## 📸 Screenshots

N/A - no UI changes.


libs/common/src/platform/misc/utils.ts
coroiu, line 1: I completely understand the reasoning behind putting these in `Utils` but I want to suggest putting these next to `login-uri.view.ts` where they are used instead. They can be moved back in the future if they end up being needed across teams :)

coroiu (commented):
Thank you for your contribution :) Just one comment that should remove me and Platform as codeowners

libs/common/src/platform/misc/utils.ts
xrvk: Done :)

jaasen-livefront (dismissed):

xrvk:
Hi team, a friendly bump here. 

The vault helpers are now colocated in login-uri.view.ts per [@coroiu](https://github.com/coroiu)'s suggestion, and [@jaasen-livefront](https://github.com/jaasen-livefront) approved. 

The PR is currently blocked only because we're waiting for maintainer approval to run full CI. Whenever someone has a moment to kick off the workflow, I'd appreciate it. Let me know if there's any further adjustments if needed. Thanks! <3

coroiu:
Hey @xrvk, @jaasen-livefront will be assigning this to QA and make sure you get feedback throughout the process! Thank you for the changes!

xrvk:
> ❌ morganzellers-bw

Anything I can do to assist here @morganzellers-bw ?

jaasen-livefront:
@xrvk I initiated CI. Can you look at the TS issue?

xrvk:
@jaasen-livefront It's ready for CI to kick off again. Let me know if there's anything I can assist with on my end!

Pushed a fix in 4de08e5.

I addressed a few more bugs on further review. The TS error was a strict-null narrowing issue in `isIpOrLocalhost`: `Utils.isNullOrWhitespace` returns `boolean` rather than a type predicate, so `uriString.trim()` wasn't safe under strict mode. Added an explicit `uriString == null` guard.

While reviewing the work submitted, I also happened to catch a related bug in `getPort`. `Utils.getUrl` returns `null` for schemeless inputs that don't contain a dot (`localhost:3000`, `[::1]:9001`), so saved entries without a scheme would have stopped matching their targets after this change. `getPort` now prepends `http://` when no scheme is present.

Some more notes, added regression tests for:
- schemeless `localhost:3000` and `[::1]:9001`
- default port equivalence (`:80` ↔ omitted, `:443` ↔ omitted)

Also moved the regular-domain test out of the IP/localhost `describe` block and tightened the inline comment.

Final pass: typecheck, strict mode, eslint, and all 43 tests in the spec pass locally :)

jaasen-livefront:
@xrvk Nice catch! That would have been a nasty one to make it in. I'll pass that onto the team, perhaps we need to fix `Utils.getUrl` or at least explicitly make this use-case known in the TS doc.

jaasen-livefront:
@xrvk Looks like lint is failing due to an indentation issue. Running prettier should fix it up.

xrvk:
@jaasen-livefront done :)

jaasen-livefront:
@xrvk Looks like the CLA just needs signing by you and I'll pass this onto product for approval.

xrvk:
@jaasen-livefront cla has been signed, lmk otherwise!

per latest comment the CLA needs signing by someone from the bw team

>❌ @morganzellers-bw

jaasen-livefront:
@xrvk This is in QA. I cannot guarantee on the timeliness of it getting done but it's now in their capable hands!

jaasen-livefront (approved):

jaasen-livefront:
@xrvk QA has passed. Awaiting final approval before merging. Thank you for your work on this!

PR #20848, opened 2026-05-25
[PM-37932] Recover browser IPC after process reload

coroiu:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-37932

## 📔 Objective

Re-inject the IPC content script into already-open tabs when the service worker starts up. After a process reload (`chrome.runtime.reload()`), tabs opened before the reload retain a content script bound to the previous extension context, leaving the IPC bridge dead until a full page reload. The old instance now tears itself down via a long-lived port's onDisconnect so the freshly injected script can register its listeners without duplicating handlers.


djsmith85 (approved):
Changes are looking good

Issue #20743, opened 2026-05-20
macOS Chrome extension does not intercept passkey requests; Chrome's native dialog appears instead


sandrodw3:
### Steps To Reproduce

1. Use Google Chrome 148.0.7778.179 (arm64) on macOS 15.7.3 (Apple Silicon).
2. Install Bitwarden browser extension version 2026.4.1.
3. Confirm "Ask to save and use passkeys" is enabled in the extension's Notification settings.
4. Have a passkey stored in Bitwarden for a site that supports passkeys (e.g., github.com).
5. Trigger a passkey flow on that site. For example, on GitHub:
   - Go to Settings → Emails → "Add email address". GitHub will request passkey re-authentication ("sudo mode").

### Expected Result

The Bitwarden extension intercepts the WebAuthn request and shows its own popup offering the passkey stored in the vault for the site.

### Actual Result

The native Chrome passkey dialog appears instead. Bitwarden never intercepts the request, so the passkey stored in the Bitwarden vault is not offered. The user is forced to use Chrome's own passkey provider (e.g., iCloud Keychain on macOS) rather than the Bitwarden-stored passkey.

### Additional Context

The problem started a few days ago. I cannot fully confirm the extension update was the trigger, but the timing aligns closely with the release of extension v2026.4.1 (May 14, 2026), which included autofill / DOM-observer performance changes for Chrome (PM-35399). Those changes touch the same machinery used to detect WebAuthn requests, which may explain a regression in passkey interception.

This may be related to prior macOS Apple Silicon passkey extension issues (e.g., #17428). The symptom pattern is also similar to #17598, where a refactor in v2025.11.1 broke passkey detection on Firefox.

### Operating System

macOS

### Operating System Version

macOS 15.7.3 (24G419)

### Web Browser

Chrome

### Browser Version

oogle Chrome 148.0.7778.179 (Official Build, arm64)

### Environment Versions

2026.4.1

### Issue Tracking Info

- [x] I understand that work is tracked outside of Github. A PR will be linked to this issue should one be opened to address it, but Bitwarden doesn't use fields like "assigned", "milestone", or "project" to track progress.

pamperer562580892423:
Did you check your "Excluded domains" in the extension (Settings --> Notifications)?

sandrodw3:
> Did you check your "Excluded domains" in the extension (Settings --> Notifications)?

Yep! I don't have any excluded domain, it was working properly last week

[...]

PR #20849, opened 2026-05-25
[PM-38098] [PM-37790] Intercept Chrome passkeys via chrome.webAuthenticationProxy

### LINKED ISSUE #14720: Bitwarden fails to prompt for passkey if 1Password extension is installed
### Steps To Reproduce

1. Install both the Bitwarden and 1password browser extensions.
2. Enter 1password settings and disable all passkey-related features.
3. Enter Bitwarden settings and enable "Ask to save and use passkeys".
4. Log out of websites and restart the browser.
5. Attempt to sign into accounts.google.com or other sites that accept passkeys using a passkey.

### Expected Result

Bitwarden opens a prompt to use a passkey for the site.

### Actual Result

Bitwarden does not open a prompt to use a passkey for the site.

### Screenshots or Videos

Image

### Additional Context

I have tested this both on Fedora 40 and MacOS (Darwin Kernel Version 24.3.0). I am using the LibreWolf fork of Firefox.

**A temporary workaround for this is to disable and re-enable the Bitwarden setting which prompts passkeys, after which Bitwarden's passkey feature will temporarily work.**

I have not extensively tested sites. This *is* an issue with Google accounts, but I have *not* had this issue when using passkeys to log into GitHub.

### Operating System

Linux

### Operating System Version

Fedora Linux 40 (KDE Plasma) x86_64, kernel version 6.13.8-100.fc40.x86_64

### Web Browser

Firefox

### Browser Version

LibreWolf version 136.0.2-1

### Environment Versions

Version: 2025.4.0

SDK: 'main (50f5a17)'

Server version: 2025.4.3

### Issue Tracking Info

xrvk:
## 🎟️ Tracking

[PM-38098](https://bitwarden.atlassian.net/browse/PM-38098), [PM-37790](https://bitwarden.atlassian.net/browse/PM-37790)

Fixes #14720, #20743, #20973, #21128

## 📔 Objective

The existing `fido2-page-script.ts` interception of `navigator.credentials.{create,get}` has two limitations on Chrome:

1. **Chrome 146+ native picker (#20743).** Chrome 146 ships a native passkey selector that runs before the content script can intercept, so Bitwarden-stored passkeys are not offered. Reproducible on macOS, Linux, and Windows per the issue thread.

2. **Coexistence with proxy-API password managers (#20973, originally #14720).** The page-script cannot coexist with extensions that already register through `chrome.webAuthenticationProxy`. 1Password is the common case.
   - **Mechanism.** 1Password's `webauthn-listeners.js` installs a property setter on `navigator.credentials` that rejects Bitwarden's overwrite, and passkey flows fail with `NotAllowedError`.
   - **User impact.** In enterprise environments where IT policy mandates 1Password, affected users cannot use Bitwarden passkeys today.
   - **Bitwarden-side confirmation.** #20973 was reproduced by Bitwarden engineering ([MarikaBW](https://github.com/bitwarden/clients/issues/20973#issuecomment-4599033533)). The original #14720 was triaged by [@abergs](https://github.com/bitwarden/clients/issues/14720#issuecomment-2955194575) as a "last-intercept-wins" limitation of the content-script approach.
   - **Repro evidence.** Full devtools capture on macOS Chrome 149: https://github.com/bitwarden/clients/pull/21128#issuecomment-4674830298.

Brave (no new selector, no proxy-API conflict) and Firefox / Safari are unaffected.

Adopts Chrome's officially supported MV3 extension-as-WebAuthn-provider API (`chrome.webAuthenticationProxy`, available since Chrome 115) as a second interception path alongside the page-script. The proxy forwards to the existing `Fido2ClientService.createCredential` / `assertCredential`. A shared `Fido2PageScriptFallbackTracker` keyed by tab id prevents the proxy from re-prompting when the page-script has just fallen back to the native API. Coexistence with other proxy-registered providers works through the browser's IPC, the same way Chrome arbitrates between any two proxy providers today.

The proxy event does not carry frame origin, so requests whose `rpId` is not a registrable suffix of the focused tab's hostname are declined and Chrome's native picker handles them. Same-registrable-domain iframe requests still pass that check and are signed with the top-level origin; see follow-up #21185.

Behind feature flag `UseWebAuthenticationProxy` (`pm-37790-use-webauthentication-proxy`), default off. On Firefox / Safari / Chrome <115 the new background `init()` is a no-op.

Supersedes #21128, the previous attempt at the page-script layer that was closed in favor of this structural change. Follow-ups: #21184, #21185, #21186.

## 📸 Screenshots

N/A. Extension UI is unchanged. With the flag on, Chrome 146+ users see the Bitwarden picker where they currently see Chrome's native dialog.

[PM-38098]: https://bitwarden.atlassian.net/browse/PM-38098?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


holow29:
It is hard for me to find information to support Brave not shipping the new selector. https://github.com/brave/brave-browser/issues/37762 seems to indicate they do ship the new selector and since the flag is gone since Chromium 146+, this is an issue there as well. (However, I have not had many issues with this on Brave versions of Chromium 146+, so perhaps the reports in that issue are inaccurate.)

It seems to me that adopting the native extension-as-WebAuthn-provider API is the best design decision regardless, and hopefully would be the primary mechanism instead of a fallback. There are certainly performance and potential privacy concerns with existing script-level overrides. I'm having trouble seeing a benefit to preferring to use these overrides vs. the native API, especially as the native API will integrate better with OS-native APIs as development occurs.

lucletoffe:
Some Linux data points for this PR, plus a heads-up on a unit test that fails on the branch.

**Context:** Arch Linux, Chromium 148.0.7778.215, self-hosted Vaultwarden, and a self-hosted Pocket ID instance (standard WebAuthn, resident keys) as the RP. We hit exactly #20973 there (Bitwarden 2026.6.1 + 1Password 8.12.26.40 co-installed → only the native dialog ever appears; removing the 1Password extension restores Bitwarden interception). Most repro reports in these threads are macOS, so happy to contribute the Linux angle.

**Failing unit test on the branch** — CI hasn't run on this PR (the workflow-approval gate short-circuits before any job), so this is probably not visible to anyone yet:

`fido2-webauthn-proxy.background.spec.ts › attach/detach lifecycle › recovers when attach() throws and continues reacting to state changes` fails deterministically here — expected 2 `attach()` calls, received 3 (Node 25, Linux; the other 81 tests across the four touched suites pass).

Cause, as far as I can tell: when `attach()` *throws* (as opposed to resolving with an error string), `runAttach()` catches and logs but does not set `attachBlocked`. The post-async re-check in `reconcileAttachment()` then sees `shouldBeAttachedAfter !== this.isAttached` and retries within the same pass — so a throwing attach yields two calls for that emission, and this test's sequence produces 3 total. The same-pass retry itself looks harmless, but the test documents the opposite intent. If the intended behavior is "no same-pass retry on throw", setting `attachBlocked = true` in the catch block of `runAttach()` aligns implementation and test; otherwise the assertion presumably wants to be 3.

I have this branch built and loaded in a throwaway Chromium profile against the setup above, and can run a with/without-1Password matrix (create + get, plus the conditional-mediation login page case, plus attach handover when 1Password is disabled live) if that's useful for QA — the fork-CI gate means the thread has little third-party test data so far.

FWIW the branch currently conflicts with main in 4 files only (`fido2.background.ts` + its spec, `main.background.ts`, `feature-flag.enum.ts`) — looks like a quick rebase.

nexthop0:
Is this ever going to get any traction? I can't use passkeys at all on Brave. And no, I don't also have 1Password installed like an idiot.

pamperer562580892423:
> I can't use passkeys at all on Brave. And no, I don't also have 1Password installed

@nexthop0 I would guess, then you have another (unrelated) issue. - I don't have any issues with passkeys on Brave at all.

audreyality:
Hi @xrvk! We've finally been able to prioritize this work! It's going to take me a little while to familiarize myself with all of the technologies involved. I just wanted to reach out as I begin my review to let you know that I'm checking it out, and am excited to improve our Chrome passkey integration.

lxgr:
At least one Chrome developer explicitly recommends against this approach: https://groups.google.com/a/chromium.org/g/chromium-extensions/c/psbSwTkbdH8

> The webAuthenticationProxy extension API targets remote desktop software that needs to fully take over all WebAuthn request processing on a remote desktop host. We don't think webAuthenticationProxy is a good fit for password managers because an extension using this API would not just need to emulate a platform authenticator, but also be responsible for dealing with all other kinds of authenticators, like USB security keys and phones via CTAP2 hybrid.

It doesn't even seem to be necessary anymore, as 1Password has recently fixed the regression where they hook the API even if "autofill passkeys" and "save passkeys" in its settings is disabled and plays nice with Bitwarden once again.

Two WebAuthN hooks can still not be active at the same time, and the WebAuthenticationProxy API would not change this, as far as I understand. (The old way of hooking the native implementation did, to some extent, but with 1Password doing what they do now, it's unfortunately no longer possible in any way.)

PR #20855, opened 2026-05-26
[BEEEP] [PM-38226] Add disable auto-start option for snap

### LINKED ISSUE #20825: Desktop App Autostarts on Ubuntu
### Steps To Reproduce

1. Turn on computer or restart.
2. Desktop client opens up automatically without any interaction.

### Expected Result

Not to start on system startup automatically.

### Actual Result

Starts on system startup automatically, without any interaction.

### Screenshots or Videos

Image

### Additional Context

I've unchecked all the tray icon settings, but the app still starts automatically on system startup.

If this is the intended behavior of the desktop app, please add an option that allows us to uncheck `Autostart on System Startup`! I do not need the app to be open all the time in the background, and it can be a security risk by doing so!

### Operating System

Linux

### Operating System Version

Ubuntu 24.04.4 LTS x86_64

### Installation method

Snap

### Build Version

Version 2026.4.0

### Issue Tracking Info

### LINKED ISSUE #20857: [SNAP] Bitwarden always startup on first boot even if it's disabled
### Steps To Reproduce

install bitwarden and check that autostart is disabled -> reboot.

### Expected Result

if the autostart is disabled it must stay closed

### Actual Result

it opens at every reboot

### Screenshots or Videos

_No response_

### Additional Context

_No response_

### Operating System

Linux

### Operating System Version

ubuntu 26.04

### Installation method

Snap

### Build Version

2026.04

### Issue Tracking Info

quexten:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-38226

## 📔 Objective

Adds the UI option to disable autostart on snap. Electron builder registers autostart always, by placing the desktop file in the autostart directory. Autostart is detected in the desktop app, since we modify the autostart desktop entry to launch bitwarden with the --autostart argument.

There are two paths we can take here:
- Delete / recreate the autostart desktop entry dynamically based on the autostart setting
  - Since electron builder creates this file, this seemed complex, though workable if really needed
- Detect the --autostart parameter, and exit the application if the autostart setting is not set
  - This makes the app launch for a split second on boot, and then quit. However, it is the far simpler approach to implement.


JeffZwolle:
"Detect the --autostart parameter, and exit the application if the autostart setting is not set"
I fear that this path will still have Ubuntu ask for the password to unlock keyring, which is annoying.

This path seems better to me: 
"Delete / recreate the autostart desktop entry dynamically based on the autostart setting"

quexten:
@JeffZwolle I agree that that option would be preferable, but it would be a fairly considerable scope increase here. However, the check is early enough in the start-up that there is no access to the key ring.

Thomas-Avery (dismissed):
LGTM

quexten:
After getting returned from QA, i rewrote this PR to just do the job properly instead of hackily. Linux now writes the .desktop file entirely from within desktop native, depending on whether it's snap / flatpak / unsandboxed.

dereknance (dismissed):

apps/desktop/src/main/messaging.main.ts
```diff
@@ -128,96 +120,22 @@ export class MessagingMain {
...
-          fs.mkdirSync(dir);
-        }
-        fs.writeFileSync(this.linuxStartupFile(), data);
-      }
+      autostart
+        .setAutostart(enabled, {
+          execPath: app.getPath("exe"),
+          autostartFlag: AUTOSTART_FLAG,
+        })
+        .catch((e) => {});
```
Thomas-Avery: Can we log this error? I see that the rust layer has good logging so my concern here is any of the framework errors that could occur while going through the NAPI layer and would just be silently dropped right now.

apps/desktop/desktop_native/core/src/autostart/impl_direct_write.rs
```diff
@@ -0,0 +1,236 @@
+//! Direct `.desktop` autostart-file management for plain Linux and Snap.
+//!
+//! On Snap the autostart Exec uses the bare `bitwarden` command (the snap package name from
+//! `electron-builder.json`), not the unpacked `exec_path`, because the exec path is different
+//! since the snap sandbox has a filsystem namespace.
```
Thomas-Avery:
```suggestion
//! since the snap sandbox has a filesystem namespace.
```

apps/desktop/desktop_native/core/src/autostart/linux.rs
```diff
@@ -1,26 +1,41 @@
...
-        Err(err) => {
-            error!(
-                error = %err,
-                "[ASHPD] Error enabling autostart");
-            Err(anyhow::anyhow!("error enabling autostart {}", err))
-        }
+/// Values supplied by the Electron main process that the autostart implementations need.
+pub struct AutostartConfig {
+    /// Absolute path to the app executable (`app.getPath("exe")`).
+    pub exec_path: String,
```
Thomas-Avery, line 18: 💭  I don't see any validation on this. It would be good to ensure the claim and looks like rust has some easy utilities to verify a path is absolute. If a future upstream changes the input we have nothing enforcing the assumed contract on the Rust side.

Thomas-Avery (changes requested):
Some minor feedback from me. Overall looks good.

apps/desktop/desktop_native/core/src/autostart/linux.rs
quexten: Added a check here that just errors out if a non-relative path is passed in.

Thomas-Avery (dismissed):
LGTM

Thomas-Avery (approved):

dereknance (approved):

Issue #19292, opened 2026-02-27
Clipboard Copy Not Working in Chrome Extension (Works in Edge)


akashsaha58c:
### Steps To Reproduce

1. Open Google Chrome (latest version) on Windows 11.
2. Click the Bitwarden browser extension.
3. Open any saved login entry.
4. Click "Copy Username" or "Copy Password".
5. Press Win + V to open Windows Clipboard History.

### Expected Result

The copied username or password should appear in Windows Clipboard History (Win + V), similar to when copying text normally (Ctrl + C).

### Actual Result

The copied value can be pasted normally using Ctrl + V in any application.
However, it does NOT appear in Windows Clipboard History (Win + V).

The same behavior works correctly in Microsoft Edge, where the copied value appears in Windows Clipboard History.

### Screenshots or Videos

<img width="773" height="305" alt="Image" src="https://github.com/user-attachments/assets/33d770ab-a91a-4367-a6ba-b421e570be8c" />

<img width="767" height="301" alt="Image" src="https://github.com/user-attachments/assets/e2508499-7f7f-4f56-bd80-ee90ce77c2b9" />

<img width="765" height="326" alt="Image" src="https://github.com/user-attachments/assets/b7bb029d-87bd-49a8-b121-da188ea2f44d" />

<img width="767" height="310" alt="Image" src="https://github.com/user-attachments/assets/b2e08101-5008-4f3c-a84d-62a4ad4d486c" />

<img width="765" height="299" alt="Image" src="https://github.com/user-attachments/assets/4260400d-3f35-450e-9974-3951a3a15f5d" />

<img width="768" height="304" alt="Image" src="https://github.com/user-attachments/assets/4a9631d6-5c1b-4818-a0e2-d3685c6b8f21" />

### Additional Context

Clipboard history is enabled in Windows settings.
Other copied text (Ctrl + C) appears normally in clipboard history.

### Operating System

Windows

### Operating System Version

Windows 11 Home Single Language. Version 25H2 OS. build 26200.7840. Experience Windows Feature Experience Pack 1000.26100.291.0

### Web Browser

Chrome

### Browser Version

145.0.7632.117 (Official Build) (64-bit)

### Environment Versions

Version 2026.1.1

### Issue Tracking Info

- [ ] I understand that work is tracked outside of Github. A PR will be linked to this issue should one be opened to address it, but Bitwarden doesn't use fields like "assigned", "milestone", or "project" to track progress.

jtodddd:
[...]

PR #20859, opened 2026-05-26
[PM-38109] fix(browser): use direct clipboard API when popup has DOM access

RobinAngele:
## Problem

When clicking the "Copy" button on a login field in the Bitwarden browser extension popup on Brave (and potentially other Chromium-based browsers with strict clipboard policies), the toast notification says "Website is copied" but the value never actually reaches the system clipboard. Users must manually select text and Ctrl+C.

This affects **all** copy operations from the popup: usernames, passwords, TOTP codes, URIs, and custom fields.

## Root Cause

The `copyToClipboard()` method in `BrowserPlatformUtilsService` unconditionally routes **all** clipboard operations through the Chrome offscreen document API when running on Manifest V3, regardless of whether the calling context has direct DOM access.

The offscreen document is only necessary for contexts that lack DOM access (background service worker). The popup has a focused `Window` with a user gesture.

Additionally, the `useLegacyCopyMethod()` fallback never checked the return value of `execCommand("copy")`, silently swallowing failures.

### 1. Guard offscreen routing by DOM capability

Added `typeof windowContext.document === "undefined"` guard to both `copyToClipboard()` and `readFromClipboard()`. The offscreen document is now only used when the calling context truly lacks DOM access (service worker).

### 2. Check execCommand return value

`useLegacyCopyMethod()` now checks the boolean return value of `execCommand("copy")`, logging a warning on `false`. Matches the existing pattern in `useLegacyReadMethod()`.

## Changes

| File | Change |
|------|--------|
| `browser-platform-utils.service.ts` | DOM-capability guard in `copyToClipboard()` and `readFromClipboard()` |
| `browser-clipboard.service.ts` | Check `execCommand("copy")` return value |
| `browser-platform-utils.service.spec.ts` | Updated 3 tests + added 2 new tests for MV3-with-DOM path |
| `browser-clipboard.service.spec.ts` | Added test for `execCommand` returning `false` |

## How to Reproduce

1. Install the Bitwarden extension on **Brave**
2. Open any saved login in the extension popup
3. Click the "Copy" icon next to the username, password, or URI field
4. Observe: toast shows "copied" but pasting (Ctrl+V) reveals nothing was copied

## Testing

```bash
npx jest --testPathPatterns="browser-clipboard.service.spec"
npx jest --testPathPatterns="browser-platform-utils.service.spec"
cd apps/browser && npm run build:chrome
# Load unpacked from apps/browser/build/ in brave://extensions
```

## Related

- Issue #19292 (similar symptom in Chrome with Windows Clipboard History)

## Out of scope / follow-up

`useLegacyReadMethod` has the same pattern — it calls `execCommand("paste")` but silently returns `""` on `false` without logging a warning, unlike `useLegacyCopyMethod` which now warns. Left out of this PR to keep the change focused; happy to add it here if preferred.


djsmith85 (approved):
Thank you for this contribution @RobinAngele. The changes are looking good and I'm passing these on for QA to verify.

You mention that useLegacyReadMethod, lacks a similar issue and that would be a nice follow-up PR to this. Although practically users will notice when paste doesn't succeed contrary to the copy not working properly.

djsmith85:
Hi @RobinAngele, thanks again for your contribution 🎉 

QA has verified your changes and this is ready to get merged. Your changes will be included in an upcoming browser release.

PR #20864, opened 2026-05-26
[PM-36823] ensure browser uses Inter font

BryanCunningham:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-35823

## 📔 Objective
Update browser extension to use correct font

Why we needed this
Two distinct problems, both fixed here:

The Inter font wasn't loading at all in some browsers. The old format("woff2 supports variations") string was an early draft of the variable-font syntax that has since been replaced. Modern browsers expect format("woff2") tech(variations). Until that line was updated, the variable font silently failed to register and the app fell back to system fonts.

The browser extension's autofill surfaces weren't applying Inter even when it loaded. The notification bar, inline-menu list, and inline-menu container are injected into host pages as iframes. They don't inherit styles from the main extension UI, and none of them were importing the theme — so the autofill UI showed up in whatever default font the iframe document picked up. Importing theme.css into each surface (plus the new :host/html/body !important rule to defeat any inherited styles) forces them onto the Inter design-system font, matching the rest of the product.

The !important is doing real work here: autofill iframes sit inside arbitrary third-party pages, and :host covers the case where the surface is rendered through shadow DOM. Without !important, host-page CSS could override the font-family.


libs/components/src/theme.css
```diff
@@ -25,3 +25,9 @@
   --font-serif: ui-serif, Georgia, Cambria, "Times New Roman", Times, serif;
   --font-mono: Menlo, SFMono-Regular, Monaco, Consolas, "Liberation Mono", "Courier New", monospace;
 }
+
+html,
+body,
+:host {
+  font-family: var(--font-sans) !important;
+}
```
vleague2: I think we are doing the global body font in each app's css instead of globally here. Not sure if there are unintended side effects of doing it here?

BryanCunningham: Removed this from theme and added to popup css

vleague2:
Looks good except there are failing tests

BryanCunningham:
> Looks good except there are failing tests

@vleague2 Yeah, this _should_ fix that https://github.com/bitwarden/clients/pull/20864/changes/27c7e33eeb97d4e94d53ecb8d76d1e535ee690d5

vleague2 (approved):
I'm unsure about the autofill changes, so I'll defer to the team for that

coroiu (approved):
Approving platform tsconfig change

jprusik (approved):

PR #20696, opened 2026-05-18
[PM-37675] fix: import archived 1Password items as soft-deleted

algojogacor:
## 🎟️ Tracking
https://bitwarden.atlassian.net/browse/PM-37675

This is an independent contribution made as part of an open-source contribution initiative.

## Summary

When importing 1pux files, items with `state: "archived"` were silently skipped instead of being imported. This is inconsistent with the documented behavior that all 1Password export items should be imported.

## Root Cause

The `OnePassword1PuxImporter.parse()` method had an early return for archived items:

```typescript
if (item.state === "archived") {
  return;
}
```

This caused archived items to be completely omitted from the import result.

## Fix

Replace the early return with a flag and set `deletedDate` on the cipher after processing, which places the item in Bitwarden's Trash:

```typescript
const isArchived = item.state === "archived";
// ... processing ...
if (isArchived) {
  cipher.deletedDate = new Date();
}
```

## Testing

- Verified that the 1pux parser test suite structure supports this change
- Archived items now flow through the normal parsing pipeline and are marked as soft-deleted
- Non-archived items are unaffected

## References

Closes #20694

PR #20867, opened 2026-05-26
[PM-38123] [bug] Preserve 1Password archive state on 1pux import (closes #20694)

### LINKED ISSUE #20694: 1password import does not archive "state: archived" entries
### Steps To Reproduce

1. Export a 1pux and do any necessary cleanup (large entries, etc)
2. Import to bitwarden

### Expected Result

Items tagged `"state": "archived"` are archived in Bitwarden.

### Actual Result

All items are active and the archive is empty.

### Screenshots or Videos

_No response_

### Additional Context

First paragraph of https://bitwarden.com/help/import-from-1password/ strongly indicates that everything from the [1pass export docs](https://support.1password.com/export/?mac#get-help) will be imported.

### Operating System

Linux

### Operating System Version

_No response_

### Web Browser

Firefox

### Browser Version

_No response_

### Build Version

web vault doesn't show any version?

### Issue Tracking Info

999purple999:
Closes #20694.

cc @djsmith85 since you flagged this as `good first issue` with the file pointer to `onepassword-1pux-importer.ts:L44-46`.

## Bitwarden internal tracking
https://bitwarden.atlassian.net/browse/PM-38123

## The bug

`OnePassword1PuxImporter` silently dropped every entry that 1Password exported with `state: "archived"`. The 1pux→Bitwarden [help-center docs](https://bitwarden.com/help/import-from-1password/) state that everything from the 1Password export is imported, so the silent drop is a contract violation: users who migrated their vault and trusted the docs ended up with their archive empty.

The drop happened in `libs/importer/src/importers/onepassword/onepassword-1pux-importer.ts` lines 43-46:

```ts
vault.items.forEach((item: Item) => {
  if (item.state === "archived") {
    return;
  }
  // ...
});
```

The existing unit test `should not import items with state 'archived'` explicitly enforced the buggy behaviour, so part of the fix was correcting that test to assert the correct contract.

## The fix

1. Remove the early `return`, letting archived items flow through the existing import pipeline.
2. After `cleanupCipher(cipher)` and before the push to `result.ciphers`, stamp `cipher.archivedDate = new Date()` when the source item was archived. `CipherView.archivedDate` is what powers the `isArchived` getter (`cipher.view.ts:178-179`), so this is the canonical way to mark a cipher as archived programmatically.
3. The 1pux schema only encodes the archive state, not the original archive timestamp (see `Item.state: "active" | "archived"` in `onepassword-1pux-importer-types.ts:65`), so we stamp the import time — the same behaviour as a user archiving a cipher manually in Bitwarden after the import.

## Tests

- **Renamed + corrected** `should import items with state 'archived' as archived ciphers`. Asserts `result.ciphers.length === 1` (not 0), `archivedDate` is a `Date`, and `isArchived` getter returns `true`. Also switched the fixture mutation to `JSON.parse(JSON.stringify(LoginData))` to avoid the shared-state mutation that the previous test had.
- **Added regression guard** `should leave active items unarchived (regression guard)` — parses the unmodified `LoginDataJson` and asserts `archivedDate === undefined` + `isArchived === false`. Locks in that active items are NOT accidentally archive-stamped by the new code.

## Verification

- `npx jest libs/importer/src/importers/onepassword` → **43/43 tests pass** in 3s on Node 22 / npm.
- `npx tsc --noEmit --project libs/importer/tsconfig.json` → no errors.
- `npx eslint libs/importer/src/importers/onepassword/...` → no errors.
- Husky pre-commit hooks → all green.

## Impact

Fix lands in `libs/importer/` which is shared by all four clients (`browser`, `desktop`, `cli`, `web` — exactly the labels on the issue). One change, four clients fixed.

## Honest scope notes

- The archive timestamp is `new Date()` (import time). If 1pux ever exports `archivedAt` in the future, the stamp is one line to update.
- I used `JSON.parse(JSON.stringify(LoginData))` for deep-cloning the test fixture. This was the minimum-diff fix for the shared-mutation latent issue; if you'd rather standardise on `structuredClone` or a Jest fixture helper, happy to follow up.

---

Francesco Pernice Botta · @999purple999


999purple999:
I have read the CLA Document and I hereby sign the CLA

pamperer562580892423:
Someone else opened a similar PR a week ago: https://github.com/bitwarden/clients/pull/20696

999purple999:
Thanks @pamperer562580892423 for the link, hadn't seen #20696. Quick heads up though, the two PRs actually do different things, so it might be worth keeping them separate until the team picks an approach.

#20696 sets `cipher.deletedDate` on archived items, which puts them in Bitwarden's Trash  they get auto-purged after the retention window. This PR sets `cipher.archivedDate` instead, which puts them in the Archive (separate feature, kept indefinitely, hidden from the default view). The issue body says archived items should be archived, so I went with `archivedDate`, but it's a real product call between the two states.

Also worth noting: the existing unit test `should not import items with state 'archived'` was actually asserting the buggy behaviour (`ciphers.length === 0`). This PR renames it and flips the assertion; #20696 leaves it as-is, which would break their CI when their fix lands. I also added a small regression guard for active items so we don't accidentally archive-stamp them.

cc @djsmith85 since you flagged the issue  happy to defer to #20696 if the team prefers the Trash mapping, or fold the test fix + regression guard into that PR as a follow-up. Just wanted to flag both before either gets closed as duplicate.

djsmith85:
Thank you for your contribution @999purple999 and thanks for the ping. I've informed the owning team about the two approaches and they should be able to make a decision soon, which they prefer.

harr1424 (commented):
Thanks @999purple999 for your contribution! 

While it's still awaiting [assessment by our product team](https://contributing.bitwarden.com/contributing/pull-requests/community-pr-process/#product-assessment) I think this is a great approach. The code is easy to understand and works as I would expect. 

One question: have you tested this using a non-premium account? Since archived ciphers are a premium feature in Bitwarden, I am curious if a Bitwarden user does not have premium what will happen to the items imported as archived? 

What **should** happen is likely going to be decided by our internal product teams, but I am curious what the current behavior looks like.

999purple999:
Thanks @harr1424, good question. I traced the path end-to-end. Here's the current behavior plus what I think the product team would want to weigh in on.

The 1pux importer (this PR) sets `cipher.archivedDate = new Date()` on the cipher object client-side. The import payload then goes to `POST /ciphers/import`, not `PUT /ciphers/archive`, so `ArchiveCiphersCommand` (the premium-gated path) isn't involved.

Server-side, `ImportCiphersCommand.ImportIntoIndividualVaultAsync` explicitly accepts the archived state on import without a premium check:

```csharp
if (cipher.UserId.HasValue && cipher.ArchivedDate.HasValue)
{
    cipher.Archives = $"{{\"{cipher.UserId.Value.ToString().ToUpperInvariant()}\":\"" +
                      $"{cipher.ArchivedDate.Value:yyyy-MM-ddTHH:mm:ss.fffffffZ}\"}}";
}
```
(server, `src/Core/Tools/ImportFeatures/ImportCiphersCommand.cs`)

`CipherRequestModel.ArchivedDate` is a first-class field and `ImportCiphersRequestModel` is just a flat container of those, so the API contract already accepts the archive state on import for any user.

After persistence, `DefaultCipherArchiveService.archivedCiphers$` (clients, `libs/common/src/vault/services/default-cipher-archive.service.ts`) displays archived ciphers regardless of premium. The premium check on the client lives on `userCanArchive$`, which gates the explicit archive action, not the display. The service even has a `showSubscriptionEndedMessaging$` flag specifically for the case "user has archived items but is not currently premium", so the UX is designed for this scenario.

**Net effect** on a non-premium user importing 1pux archive items: items are imported with their archive state preserved, appear in the archived view alongside any prior archived items, remain readable, but the user cannot perform new archive or unarchive actions while non-premium. The `showSubscriptionEndedMessaging$` banner nudges an upgrade.

**One product question worth surfacing** to your internal team: with the current import path, a never-premium user who imports 1pux archive items gets them as read-only on entry. The product team can pick:

1. Current (graceful): import succeeds, items visible read-only, upgrade nudge.
2. Strict: importer strips the archive flag for non-premium users, items land active.

This PR ships behavior (1). If your team picks (2), the cleanest implementation is a client-side strip in the importer gated on `userCanArchive$`. I can have a follow-up commit ready within a few hours of a green light.

**On verification depth**: the analytical proof above is the source-of-truth chain (importer to upload to server-side persistence to client display). If you'd like belt-and-suspenders empirical verification, two quick follow-ups I can do here:

- Add a more targeted unit test to this PR that asserts the `CipherRequestModel` emitted from an imported archived item preserves `ArchivedDate`, as a regression guard against future drift in the import payload shape.
- Spin up a self-hosted Bitwarden instance with a fresh non-premium account, import a synthetic .1pux with one archived item, and post the observed behavior here.

Let me know which (if either) is most useful before merge.

Re: CI, the `EnforceLabel` failures come from the `needs-qa` label the bot auto-added (`BANNED_LABELS: "hold,needs-qa"` in `.github/workflows/enforce-labels.yml`). They'll go green once QA passes.

harr1424:
Thanks, @999purple999 , for your response! I've relayed this to our product team, and will reach out here with any updates.

harr1424:
Hi @999purple999 I'm happy to share that we will be proceeding with your contribution! In the coming days I will review the code changes, which appear minimal and focused, and perform some functional testing before sending this off to QA.  I'll reach out here in case any questions arise. Thanks again for your contribution!

harr1424 (approved):

harr1424:
Hi again @999purple999 I just wanted to let you know that your contribution is now ready for QA. I'll reach out here with any next steps needed to have your contribution merged.

harr1424:
Thanks @999purple999 for contributing a fix! This merged PR closes #20694

PR #20872, opened 2026-05-26
[PM-37555] disable autofill when logged out

audreyality:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-37686
https://bitwarden.atlassian.net/browse/PM-37555

## 📔 Objective

Reduce the impact of autofill scripts when a user is logged out.

A full overview of the design is available in [lifecycle.design.md](https://github.com/bitwarden/clients/blob/autofill/pm-37686/disable-autofill-when-logged-out/apps/browser/src/autofill/content/lifecycle.design.md)


apps/browser/src/autofill/services/autofill.service.ts
```diff
@@ -3143,4 +3169,67 @@ export default class AutofillService implements AutofillServiceInterface {
...
+    // here are all idempotent on the receivers, so the duplicate is benign.
+    const targets = new Map<string, { tab: chrome.tabs.Tab; frameId: number | undefined }>();
+    this.autofillScriptPortsSet.forEach((port) => {
+      const tab = port.sender?.tab;
+      const frameId = port.sender?.frameId;
+      if (!tab?.id) {
+        return;
+      }
+      targets.set(`${tab.id}:${frameId ?? -1}`, { tab, frameId });
+    });
```
audreyality, line 3226: 🤔 As written, this is broadcasting to every frame on the page, and scanning is happening on all of them. This matches the current functionality, which runs unconditionally. Is this desirable? It seems to me like frames containing, say, ads, we wouldn't want to include in the broadcast.

audreyality:
📝 There's something in `main` that causes abject chaos to happen when I merge and try to resolve conflicts, so I'm closing this PR and continuing one that's rebased on the latest main instead.

PR #20875, opened 2026-05-27
[PM-38159] feat(desktop): rich-context SSH agent approval prompt

passabilities:
**Why this PR:** today the SSH agent approval prompt only says "ssh is requesting access to {key name} to authenticate to a server." That is not enough to make a confident approve/deny decision — the user can't tell which app/process triggered the request, which server is being contacted, or whether the request is legitimate vs a malicious local process posing as a normal SSH client. This PR adds the locally-derivable context the agent already has (process tree, executable path, argv, host-key fingerprint, known_hosts match) and surfaces it in a trust-labeled prompt.

## Summary

Adds a rich, trust-labeled SSH agent approval prompt that surfaces the requesting application, its parent process chain, the derived hostname, the SSH server's host-key fingerprint, and any `~/.ssh/known_hosts` match status. Gated behind `FeatureFlag.SSHAgentRichContext` (default OFF); legacy prompt path is byte-for-byte preserved.

### What changed

- **Rust agent (`desktop_native/core`)** — new `ssh_agent::context` module: OpenSSH argv parser, SHA256 host-key fingerprint, `~/.ssh/known_hosts` matcher (plain + hashed entries), process-tree walker, and a `build()` orchestrator that produces a trust-labeled `HostContext`. Hashed `known_hosts` entries (Linux default with `HashKnownHosts yes`) are forward-verified against the argv hostname so the user sees real names, not `<hashed>`. SSH-config aliases that resolve to a different canonical name on disk fall back to the typed alias for display — the cryptographic identity (host key) is what's actually verified.
- **NAPI bridge** — new `RequestContextNapi`, `AppContextNapi`, `HostContextNapi`, `ProcessFrameNapi` types and a `setRichContextEnabled` function.
- **Main process** — `SIGN_REQUEST` IPC payload carries the new `context` field; new `SET_RICH_CONTEXT_ENABLED` channel.
- **Renderer** — `ApproveSshRequestComponent` builds prose from the `RequestContext`: friendly-name lookup for known apps (`Code` → "Visual Studio Code", etc.), noise-process filter (`bash`/`tmux`/etc.) so the parent chain "(run by …)" picks the meaningful ancestor, inline Verified/Unverified trust badges, and a collapsible Advanced section with PID, executable path, full parent chain, host key fingerprint, `known_hosts` match status, and the raw command line (clearly labeled as untrusted).
- **i18n** — new strings under `apps/desktop/src/locales/en/messages.json`. Legacy `sshkeyApproval*` / `sshAction*` keys retained for the flag-off path.

### Trust model

The prompt distinguishes three trust levels per hostname source:

| Source | Trust | When it fires |
|---|---|---|
| `known-hosts` | **Verified** (green badge) | Live host key matches an entry in `~/.ssh/known_hosts` (plain entry, or hashed entry whose hash matches the argv hostname, or hashed entry where argv is the user-typed alias) |
| `argv` | **Unverified** (amber badge) | Hostname parsed from the requesting process's argv only — process can claim any value |
| `host-key` | no badge | Host key bytes available but no `known_hosts` entry; fingerprint shown for the user to verify out-of-band |

### Screenshots

**1. Collapsed prompt — verified GitHub host via gitkraken**

Selection_607

**2. Same request, Advanced expanded — full process chain `systemd → gitkraken → gitkraken → git → sh → ssh`, host key fingerprint, raw command line**

Selection_612

**3. Collapsed prompt — SSH connection to a host defined in `~/.ssh/config` (`yotta-server-data` is a custom alias). The agent recognizes the server's host key from `known_hosts` and displays the user-typed alias with the Verified badge.**

Selection_610

**4. Same alias request, Advanced expanded — shows the `sftp` invocation through the alias plus the verified fingerprint**

Selection_611

### Backwards compatibility

- Feature flag `SSHAgentRichContext` defaults to FALSE. With the flag off, every field beyond `process_name` / `is_forwarding` / `namespace` is `None` and the renderer falls through to the existing template — byte-identical behavior to current main.
- Pipeline is fail-soft: any single derivation step (process tree, argv, fingerprint, known_hosts) that errors yields `None` for that field and the prompt still renders.
- All derivation is local: no DNS, no reverse lookups, no network calls. argv values and hostnames are never logged.

### Scope

- This PR implements rich context on the **V1** SSH agent path (`desktop_native/core/src/ssh_agent/`). V2 (`desktop_native/ssh_agent/`) does not yet plumb `host_key` / `pid` through its `SignRequest`; the derivation utilities (`desktop_core::ssh_agent::context::*`) are deliberately public so V2 can adopt them in a follow-up PR once that plumbing lands.

## Test plan

- [x] Rust: full `cargo test` green — 31 new tests covering argv parser (10), fingerprint (3), known_hosts (13), process-tree gather (5), and `resolve_host` orchestration (5).
- [x] TS: full jest autofill suite green (138 tests) — 8 new prose-builder cases + 2 service-plumbing cases.
- [x] Manual: real `ssh git@github.com` via gitkraken on Linux → Verified badge, real fingerprint shown (screenshots 1, 2).
- [x] Manual: `sftp` to an `~/.ssh/config` alias on Linux → Verified badge displays the alias, not `<hashed>` (screenshots 3, 4).
- [ ] Manual cross-platform verification (macOS Full Disk Access fallback, Windows named-pipe peer pid) — follow-up.


dwbit:
Hi there, and thanks for the contribution! Closing this one out for now as it references the older code rather than the newer V2 build. 

As per the [contribution guidelines](https://contributing.bitwarden.com/contributing/), since this modifies the UX, please open a [Github discussion post](https://github.com/orgs/bitwarden/discussions/categories/password-manager) for Bitwarden team feedback on the design.

PR #20879, opened 2026-05-27
SSH Agent v2: agent forwarding

neuronull:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-31762

## 📔 Objective

Building off of Sign Request functionality, this change implements the Session Bind extension for the v2 agent.

Reviewer notes:
- Over half the code changes are new unit tests / integration tests
- The primary notable architectural decision is the storage of the is_forwarding only, as opposed to the storage of additional fields that are provided (e.g. host_key). The v1 on the electron side only utilizes is_forwarding, so for now we will YAGNI to simplify the code a bit. If we elect to utilize other fields in the future (active discussions happening on this), we'll just introduce a struct for that.
- The Claude review discussing the "latching" of the session bind - I implemented it this way because I discovered a bug in the v1 implementation (ticket pending), which is that we detected the session bind but on subsequent packet from the client, cleared our cache of the value, so when the request was surfaced to electron we didn't have it noted as an agent forwarding request.

## 📸 Testing Done

- created two containers with ssh support. ssh'd into the first container with forwarding, then to the second. At the second, confirmed that the authorization prompt in the UI displays the agent forwarding note, and that the hop succeeds.


apps/desktop/desktop_native/ssh_agent/tests/common.rs
```diff
@@ -189,3 +189,67 @@ mockall::mock! {
...
 }
+
+fn write_ssh_string(buf: &mut Vec<u8>, data: &[u8]) {
+    buf.extend_from_slice(&(data.len() as u32).to_be_bytes());
+    buf.extend_from_slice(data);
+}
+
+fn make_session_bind_payload(
+    keypair: &ssh_key::private::Ed25519Keypair,
+    session_id: &[u8],
```
quexten, line 211: Just a thought: Is it worth to introduce a newtype around session id here? I am really never a fan of working with byte slices directly unless needed. 
(Non blocking)

quexten (dismissed):

apps/desktop/desktop_native/ssh_agent/tests/common.rs
neuronull: I'm open to it but I'm curious the primary motivation, because in this unit test helper this function argument is the only byte slice, so the newtype wouldn't help us out much from a type safety standpoint. In the function itself we only ever use the bytes as-is / in-place, so the usage in the function would call `.0` in two places. 

If the motivation is mostly for aesthetics,  we might consider a type alias instead.

However, for both the newtype and the type alias cases, since it's a reference that we need, we'd have to introduce a lifetime specifier. e.g.

```Rust
type SessionId<'a> = &'a [u8];

// or

struct SessionId<'a>(&'a [u8]);
```

, the combination of which seems like it might add more mental overhead than it's removing...

This also isn't the only place we're working with bytes slices in the server module, so if we're going to make a decision for this helper fn , we should zoom out to consider the other areas they're used. 
(on that note- I left a comment on one of the recent previous PRs to the effect of: I'd considered creating a more structured type hierarchy for the decoding/parsing of requests in the server, and decided against it based on how small the processing we're doing is, and weighing the costs/benefits of doing that. It seemed not worth it to me but I'm happy to revisit that idea. If it's of interest I can track down the comment and/or we can sidebar discussion this internally).

Again I'm open to the discussion and to making changes, let me know if I'm missing anything or making assumptions that don't match what you had in mind.

neuronull:
platform team was added automatically when the base changed to main, removing

apps/desktop/desktop_native/ssh_agent/tests/server_windows.rs
```diff
@@ -5,11 +5,34 @@ use tokio::{io::AsyncWriteExt, net::windows::named_pipe::ClientOptions};
...
+    unsupported_dsa_key_blob, // ||||||| 2abb233e60
+    //     agent_with_keys, always_approving_agent, always_denying_agent,
+    // framed_request_identities,     framed_sign_request, init_tracing, parse_first_key_name,
+    // parse_sign_response_algorithm,     read_framed_response, test_ed25519_key,
+    // test_ed25519_key_blob, test_rsa_key, test_rsa_key_blob, =======
+    //     agent_with_keys, always_approving_agent, always_denying_agent,
+    // framed_request_identities,     framed_sign_request, init_tracing, parse_first_key_name,
+    // parse_sign_response_algorithm,     read_framed_response, test_ed25519_key,
+    // test_ed25519_key_blob, test_rsa_key, test_rsa_key_blob,     unsupported_dsa_key_blob,
+    // >>>>>>> origin
```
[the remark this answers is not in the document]
neuronull: fixed in https://github.com/bitwarden/clients/pull/20879/commits/d5a4ca862ce3f00f317fc48e99d6bd035c1429cb

apps/desktop/desktop_native/ssh_agent/tests/common.rs
quexten: Thank you for putting this much thought in the response. I think you are right and the abstraction is probably not worth the effort here.

I think in general adding newtypes in a few places made things easier to reason about / harder to mess up, so I thought I'd suggest it, but if you think it's not worth it, let's leave it.

However, I'll say that I didn't want to suggest putting a reference, but actually to own the data. Though, after conferring with the docs, it seems the size depends on the hash of the handshake transcript, so this would be more pain than it would be worth.

quexten (dismissed):
Approval is still valid.

neuronull:
FYI @quexten  , in https://github.com/bitwarden/clients/pull/20879/commits/f00a5317266f4d9f4bbe7103a93a2b7ad60088ee , I had to introduce the host key fingerprint across the electron blood brain barrier , in order to support the "remember until locked" for the agent forwarding case.

quexten (approved):

dereknance (approved):

PR #20881, opened 2026-05-27
fix(cli): harden bw serve origin-protection against DNS rebinding [PM-36600]

addisonbeck:
## Tracking

[PM-36600](https://bitwarden.atlassian.net/browse/PM-36600)

## Objective

`bw serve` runs an unauthenticated local HTTP API and was protected only by an `Origin`-header presence check. Browsers omit `Origin` on same-origin GETs, so any DNS-rebinding attack, where a malicious page rebinds its own domain to `127.0.0.1`, sends `Host: evil.example:8087` with no `Origin`, and the original middleware passed those requests through. 

This is fixed over two layers

1. Reject any request whose `Host` header isn't in `{localhost:PORT, 127.0.0.1:PORT, [::1]:PORT, <configured-hostname>:PORT}`. DNS-rebound requests carry the attacker's hostname in `Host`, so this catches them. Bare-host forms are also accepted when PORT is 80 or 443 (clients omit the default port per RFC 7230 §5.4); the allowlist is disabled when `--hostname all` is passed and Layer 2 alone applies.
2. Reject any request that carries an `Origin` header. We are already doing this today.

The allowlist construction (`buildServeAllowedHosts`) and the middleware factory (`buildOriginProtectionMiddleware`) are extracted into pure, exported functions for unit testing.

[PM-36600]: https://bitwarden.atlassian.net/browse/PM-36600?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/cli/src/commands/serve.command.ts
+++ b/apps/cli/src/commands/serve.command.ts
@@ -83,7 +83,7 @@ export class ServeCommand {
 
   async run(options: OptionValues) {
     const protectOrigin = !options.disableOriginProtection;
-    const port = options.port || 8087;
+    const port = Number(options.port) || 8087;
     const hostname = options.hostname || "localhost";
     this.serviceContainer.logService.info(
       `Starting server on ${hostname}:${port} with ${
--- a/apps/cli/src/commands/serve.command.ts
+++ b/apps/cli/src/commands/serve.command.ts
@@ -15,11 +15,16 @@ import { OssServeConfigurator } from "../oss-serve-configurator";
 import { ServiceContainer } from "../service-container/service-container";
 
 export function buildServeAllowedHosts(hostname: string, port: number): Set<string> {
-  return new Set(
-    [`localhost:${port}`, `127.0.0.1:${port}`, `[::1]:${port}`, `${hostname}:${port}`].map(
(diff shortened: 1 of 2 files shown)


apps/cli/src/commands/serve.command.ts
```diff
@@ -4,14 +4,77 @@ import net from "node:net";
...
+
+function rejectIfOriginPresent(ctx: Context, logService: LogService): boolean {
+  if (ctx.headers.origin === undefined) {
+    return false;
+  }
+  ctx.status = 403;
+  logService.warning(
+    `Blocking request from "${
+      Utils.isNullOrEmpty(ctx.headers.origin) ? "(Origin header value missing)" : ctx.headers.origin
+    }"`,
```
dani-garcia, line 51: Non-blocking suggestion: We're already returning early if `ctx.headers.origin` is undefined, so I'd say this check is overly defensive:
```suggestion
    `Blocking request from "${ctx.headers.origin}"`,
```

dani-garcia (approved):

PR #20882, opened 2026-05-27
[CL-1213] allow tab label to be anchor for spotlight

BryanCunningham:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/CL-1213

## 📔 Objective

Allow consumers to attach spotlights to tabs

### DIFF
--- a/libs/components/src/popover/popover-anchor-for.directive.ts
+++ b/libs/components/src/popover/popover-anchor-for.directive.ts
@@ -20,6 +20,11 @@ import { PositionIdentifier, defaultPositions } from "./default-positions";
 import { PopoverComponent } from "./popover.component";
 import { SpotlightService } from "./spotlight.service";
 
+/** Implement and provide as `useExisting` to redirect `[bitPopoverAnchorFor]` from the host to another element. */
+export abstract class PopoverElementProvider {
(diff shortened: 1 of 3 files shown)


libs/components/src/popover/popover-anchor-for.directive.ts
```diff
@@ -20,6 +20,11 @@ import { PositionIdentifier, defaultPositions } from "./default-positions";
 import { PopoverComponent } from "./popover.component";
 import { SpotlightService } from "./spotlight.service";
 
+/** Implement and provide as `useExisting` to redirect `[bitPopoverAnchorFor]` from the host to another element. */
+export abstract class PopoverElementProvider {
+  abstract readonly elementRef: ElementRef<HTMLElement>;
```
willmartian: maybe rename `popoverAnchorElementRef` in case an element already has a element ref defined for some other reason?

willmartian (dismissed):
One nit:

libs/components/src/popover/popover-anchor-for.directive.ts
BryanCunningham: Makes sense. Also easier to wrap your head around what it actually is. Will make that change

willmartian (approved):
Good to merge if workflows all pass

PR #20883, opened 2026-05-27
[PM-37175] Preserve individual vault export folders when importing to My Items

mcamirault:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-37175

## 📔 Objective

When importing into the `My Items` collection the importer will convert any collections present in the export file into folders, but would ignore folders if they were present (as is the case with an individual vault export file). This PR makes the importer preserve folders in those cases.

So importing either of the following two files into the `My Items` collection will have the same result:
- An individual vault export with one cipher in a folder called `Engineering`
- An organization vault export with one cipher in a collection called `Engineering`

The result is one new cipher in the `My Items` collection and in a new folder called `Engineering`

### DIFF
--- a/libs/importer/src/services/import.service.spec.ts
+++ b/libs/importer/src/services/import.service.spec.ts
@@ -297,7 +297,31 @@ describe("ImportService", () => {
       await expect(setImportTargetMethod).rejects.toThrow();
     });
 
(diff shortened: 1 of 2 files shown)


mcamirault:
@claude Give this a review

harr1424 (approved):

PR #20885, opened 2026-05-27
[PM-37800] Add VaultBatchBarService to libs/vault

nick-livefront:
## 🎟️ Tracking

[PM-37800](https://bitwarden.atlassian.net/browse/PM-37800)

## 📔 Objective

Adds `VaultBatchBarService` to `libs/vault` — a component-scoped service that manages selection state and bulk vault actions (delete, restore, move, archive, unarchive, assign to collections).

- **`ASSIGN_COLLECTIONS_DIALOG` token** — decouples the service from the web-only `AssignCollectionsWebComponent`, keeping `libs/vault` free of app-layer dependencies while still allowing web to provide its own dialog implementation via `AssignCollectionsWebDialogAdapter`.
- **`BULK_DELETE_DIALOG` token** — same pattern for the bulk delete dialog; allows `VaultBatchBarService` to open a delete confirmation without importing from `apps/web`.
  - This was needed because the delete dialog depends on `CollectionService` which cannot be imported into `libs/vault`
- `BulkMoveDialogComponent` is moved from `apps/web` into `libs/vault` so it can be shared across vault contexts.
- `BulkDeleteDialogParams` / `BulkDeleteDialogResult` are relocated to the new token file.
- Both vault components wire up the service behind the `PM37785_VaultBatchBar` feature flag.

UI integration and testing is tracked in [PM-37806](https://bitwarden.atlassian.net/browse/PM-37806) (individual vault) and [PM-37807](https://bitwarden.atlassian.net/browse/PM-37807) (admin console).

## 📸 Screenshots

N/A — service-only change, no UI.

[PM-37800]: https://bitwarden.atlassian.net/browse/PM-37800?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[PM-37806]: https://bitwarden.atlassian.net/browse/PM-37806?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/vault/src/services/vault-batch-bar.service.ts
```diff
@@ -0,0 +1,667 @@
...
+      return false;
+    }
+
+    if (uniqueOrgIds.size === 0 && hasEditableCollections) {
+      return hasPersonalItems;
+    }
+
+    const [orgId] = uniqueOrgIds;
+    const org = allOrganizations.find((o) => o.id === orgId);
+    const canEditOrManageAll = org?.canEditAllCiphers === true && config.isOrgVault;
```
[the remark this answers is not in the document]
nick-livefront: [d4d6a44](https://github.com/bitwarden/clients/pull/20885/commits/d4d6a44dc2b4c525bcc62db1173d9f36413268b2)

libs/vault/src/services/vault-batch-bar.service.ts
```diff
@@ -0,0 +1,667 @@
...
+          message: this.i18nService.t("nothingSelected"),
+        });
+        return;
+      }
+
+      await this.cipherService.restoreManyWithServer(
+        [...unassignedCiphers, ...editAccessCiphers],
+        userId,
+        org.id,
+      );
```
jrmccannon: Should this be wrapped in a try/catch and show a toast on failure?

jrmccannon (changes requested):
One small comment

libs/vault/src/services/vault-batch-bar.service.ts
nick-livefront: Nice catch: [a11c429](https://github.com/bitwarden/clients/pull/20885/commits/a11c4298578f19bf0bf57ba6170b0fd3e2468414)

jrmccannon (dismissed):

jaasen-livefront (commented):
@nick-livefront

`canDelete`/`canRestore` and the corresponding action methods appear to use different authorization rules. The `can*` signals delegate to `CipherAuthorizationService`, which allows admin-console delete/restore for unassigned org ciphers when `organization.canEditUnassignedCiphers` is true. However, `bulkDelete()` only allows `c.edit` or `org.canEditAllCiphers`, and `bulkRestore()` also omits the `canEditUnassignedCiphers` case. This means the UI can enable Delete/Restore for an unassigned org cipher, but clicking the action then fails with `missingPermissions`. Can the action methods reuse the same authorization helper, or otherwise mirror the unassigned-cipher permission case?

libs/vault/src/services/vault-batch-bar.service.ts
```diff
@@ -0,0 +1,675 @@
...
+    const selectedCiphers = selected.filter((i) => i.cipher).map((i) => i.cipher as C);
+    const anyArchived = selectedCiphers.some((c) => CipherViewLikeUtils.isArchived(c));
+
+    const bulkAssignAllowed = config.hasCiphers && !anyArchived;
+
+    if (!bulkAssignAllowed) {
+      return false;
+    }
+
+    if (config.isOrgVault) {
```
shane-melton: ❓ Won't this always short circuit and skip the rest of the logic when in the Admin Console? I know this is to be a common/expected action in the AC, but if the user has no cipher items selected, this would still be returning true.

shane-melton: Oh, I see that was the existing behavior in AC, and it would show a "nothing selected toast". Should we still check that they've at least selected one cipher?

libs/vault/src/services/vault-batch-bar.service.ts
```diff
@@ -0,0 +1,675 @@
...
+  };
+
+  private readonly config = signal<VaultBatchBarConfig>(this.defaultConfig);
+
+  private readonly showBulkTrashOptions = toSignal(
+    this.routedVaultFilterService.filter$.pipe(map((f) => f.type === "trash")),
+    { initialValue: false },
+  );
+
+  private readonly showBulkMove = computed(() => !this.showBulkTrashOptions());
```
shane-melton: 🎨 The bulk move to folder option should not be available in the organization vault (folders are a individual vault concept).

shane-melton: ❓ 💭 Should we add "folder" somewhere to these names to make it more clear they're related to folder vs collections?

libs/vault/src/services/vault-batch-bar.service.ts
```diff
@@ -0,0 +1,675 @@
...
+        return combineLatest(canDeleteCiphers$).pipe(
+          map((results) => results.every((r) => r) && canDeleteCollections),
+        );
+      }),
+    ),
+    { initialValue: true },
+  );
+
+  /** True when the selected ciphers can be assigned to collections.*/
+  readonly canAssignToCollections = computed(() => {
```
shane-melton, line 277: 💭 Wow, there is a lot of complex logic for this check and quite a few early returns. Its nice to have it all in one place now, but perhaps we could add some clarify documentation on some of these early returns so its easier to understand when this action should be available.

shane-melton (changes requested):
Looks pretty good! Just a few comments/questions.

nick-livefront:
> `canDelete`/`canRestore` and the corresponding action methods appear to use different authorization rules. The `can*` signals delegate to `CipherAuthorizationService`, which allows admin-console delete/restore for unassigned org ciphers when `organization.canEditUnassignedCiphers` is true. However, `bulkDelete()` only allows `c.edit` or `org.canEditAllCiphers`, and `bulkRestore()` also omits the `canEditUnassignedCiphers` case. This means the UI can enable Delete/Restore for an unassigned org cipher, but clicking the action then fails with `missingPermissions`. Can the action methods reuse the same authorization helper, or otherwise mirror the unassigned-cipher permission case?

@jaasen-livefront nice catch! [0ff5f9b](https://github.com/bitwarden/clients/pull/20885/commits/0ff5f9b428386f9cecec6455255cbcc8f00a6ebc)

libs/vault/src/services/vault-batch-bar.service.ts
nick-livefront: [ec1d427](https://github.com/bitwarden/clients/pull/20885/commits/ec1d4276f10ffc5d4c9f10c1fbc3624ecc518c2d)

nick-livefront: Agreed, good idea. I changed this to `Add to folder` naming which aligns with the UI and the mental model: [ec1d427](https://github.com/bitwarden/clients/pull/20885/commits/ec1d4276f10ffc5d4c9f10c1fbc3624ecc518c2d)

libs/vault/src/services/vault-batch-bar.service.ts
nick-livefront: [fd89969](https://github.com/bitwarden/clients/pull/20885/commits/fd89969b24a24b8fc82ddc2f4580c8ee77d7deb7) added! Let me know if anything else specific should be called out

libs/vault/src/services/vault-batch-bar.service.ts
nick-livefront: Bingo! Good call, I didn't catch this because the UI hides when nothing is selected but it's still worth adding in there. I snuck that change into this commit: [fd89969](https://github.com/bitwarden/clients/pull/20885/commits/fd89969b24a24b8fc82ddc2f4580c8ee77d7deb7)

libs/vault/src/services/vault-batch-bar.service.ts
```diff
@@ -0,0 +1,691 @@
...
+      .filter((i) => i.collection !== undefined)
+      .map((i) => i.collection as CollectionView),
+  );
+
+  /** True when bulk add-to-folder is allowed. */
+  readonly canAddToFolder = computed(() => {
+    return (
+      this.showBulkAddToFolder() && this.selected().filter((item) => item.collection).length === 0
+    );
+  });
```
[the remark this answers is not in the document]
nick-livefront: [378f2fe](https://github.com/bitwarden/clients/pull/20885/commits/378f2fe2b37620c9b663d1ce51d8b16aab7aec4f)

jaasen-livefront (approved):
Nice work @nick-livefront !

jrmccannon (approved):
Looks good from AC

shane-melton (approved):
🚀

PR #20886, opened 2026-05-27
[PM-37079] Fix applicationName empty save and fetch in Access Intelligence

Banrion:
## 🎟️ Tracking

[PM-37079](https://bitwarden.atlassian.net/browse/PM-37079)

## 📔 Objective

Fixes critical app selections being lost on every login. A single bad `applicationName` entry was breaking the entire report on decryption, causing the UI to fall back to "Run Report" and wipe the user's selections.

**Changes:**

1. Prevent empty `applicationName` at generation by filtering null/empty domains in `getTrimmedCipherUris`.
2. Make report validators recover gracefully: drop invalid entries and log a warning instead of throwing. Structural failures still throw.
3. Add `ReportStatus.LoadError` and a matching toast for fetch/decryption failures, distinct from generation failures.
4. Update encryption service, versioning service, and type-guard tests to match the new behavior.

[PM-37079]: https://bitwarden.atlassian.net/browse/PM-37079?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Banrion:
@claude Please review

AlexRubik (dismissed):
Approving. `validateAccessReportPayload` in `risk-insights-type-guards.ts` wasn't converted to the tolerant pattern, so the V2 read path still throws strictly. Not a current-prod issue (`AccessIntelligenceNewArchitecture` is off) but it would reappear at V2 rollout. Tracked in [PM-38262](https://bitwarden.atlassian.net/browse/PM-38262).

AlexRubik (approved):

PR #20891, opened 2026-05-27
[PM-14890] Localize import instruction text 

harr1424:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-14890

## 📔 Objective

This PR moves hard-coded English instructions for each import into locale files so that they may be translated.

### DIFF
--- a/apps/browser/src/_locales/en/messages.json
+++ b/apps/browser/src/_locales/en/messages.json
@@ -4862,6 +4862,186 @@
     "message": "See detailed instructions on our help site at",
     "description": "This is followed a by a hyperlink to the help website."
   },
+  "importInstructionsKeepassxcsv": {
+    "message": "Using the KeePassX desktop application, navigate to \"Database\" → \"Export to CSV file\" and save the CSV file."
+  },
+  "importInstructionsAviracsv": {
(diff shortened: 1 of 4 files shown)


djsmith85 (commented):
@harr1424: Thanks for tackling this. Something that has been on my list for a long time.

At the time I wasn't sure if it would be easier to just include the instructions on a separate help site (translated by us) or have them in product and translated. My main concern was keeping these instructions up to date.

🎨  To further clean up the messy markup by essentially removing all the conditionals (`<ng-container *ngIf="format === 'importerXXcsv'">`) and make it more service-data-driven:

Define the importers as [ImportersMetadata]( https://github.com/bitwarden/clients/blob/main/libs/importer/src/metadata/importers.ts#L14) which we started introducing for the chromium importers.

Maybe instead of defining `Instructions.unique` it could be the i18n key? Or it's `unique` and we have an additional metadata field for that key 🤷 

While this metadata is now exposed via the service layer, the component can simply read the value from `metadata.instructions` or `key`. This reduces the markup to a single line. For the importers that have an external link in the instructions, that could also just be added to the ImportersMetadata

PR #20892, opened 2026-05-27
[PM-37780] - Add "Save and fill" button to inline menu cipher creation

jaasen-livefront:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-37780
https://bitwarden.atlassian.net/browse/PM-37588

## 📔 Objective

Adds a "Save and fill" button to the add/edit vault item form when it is opened from the browser inline menu. When a user creates a new login via the inline menu and clicks "Save and fill", the cipher is saved and immediately autofilled into the originating page, followed by the "Login saved" confirmation bar.

This is gated behind the `PM36385FillAfterSave` feature flag (default: off).

### Changes

- **`feature-flag.enum.ts`**: Add `PM36385FillAfterSave` feature flag (default `false`)
- **`vault-popout-window.ts`**: Pass `fillAfterSave=true` as a query param when opening the add/edit popout from the inline menu; extend `cipherOptions` type to include `fillAfterSave`
- **`overlay.background.ts`**: Pass `fillAfterSave: true` when calling `openAddEditVaultItemPopout` from `buildCipherAndOpenAddEditVaultItemPopout`
- **`abstractions/overlay.background.ts`**: Add optional `cipherId` field to `OverlayBackgroundExtensionMessage`
- **`add-edit.component.ts`**: Read `fillAfterSave` query param, check feature flag, inject `VaultPopupAutofillService` and `ConfigService`; on submit when `fillAfterSave` is active, autofill the new cipher and send `showLoginSavedNotification` to the originating tab; add `submitAndFill` action
- **`add-edit.component.html`**: Render a conditional footer when `fillAfterSave` is active — "Save and fill" (primary), "Save" (secondary), "Cancel"; falls back to the existing footer otherwise
- **`messages.json`**: Add `saveAndFill` i18n string ("Save and fill")

## 📸 Screenshots

https://github.com/user-attachments/assets/1f35b1b1-e42f-43bb-9d62-90aff3aee1d5


apps/browser/src/autofill/background/abstractions/overlay.background.ts
```diff
@@ -157,6 +157,7 @@ export type OverlayBackgroundExtensionMessage = {
   isOpeningFullInlineMenu?: boolean;
   styles?: Partial<CSSStyleDeclaration>;
   data?: LockedVaultPendingNotificationsData;
+  cipherId?: string;
```
nick-livefront: ❓ Is this used anywhere? Should it be?

libs/common/src/enums/feature-flag.enum.ts
```diff
@@ -88,7 +88,7 @@ export enum FeatureFlag {
   PM32180PremiumUpsellAccountAge = "pm-32180-premium-upsell-account-age",
   PM28091_AddCopyAndQuickLaunchActions = "pm-28091-add-copy-and-quick-launch-actions",
   PM34500_StrictCipherDecryption = "pm-34500-strict-cipher-decryption",
-
+  PM29968_FillAfterSave = "pm-29968-fill-after-save",
```
nick-livefront, line 98: Only verifying the ticket numbers here, your PR description has `PM36385FillAfterSave`

apps/browser/src/vault/popup/components/vault/add-edit/add-edit.component.ts
```diff
@@ -183,10 +193,13 @@ export type AddEditQueryParams = Partial<Record<keyof QueryParams, string>>;
 })
 export class AddEditComponent implements OnInit, OnDestroy {
   readonly cipherFormComponent = viewChild(CipherFormComponent);
+
   headerText: string;
   config: CipherFormConfig;
   canDeleteCipher$: Observable<boolean>;
   routeAfterDeletion: ROUTES_AFTER_EDIT_DELETION = "/tabs/vault";
+  protected fillAfterSave = false;
```
nick-livefront: ⛏️ `fillAfterSave` and `shouldFillOnSave` are pretty close naming wise. Can the former have more context around the feature flag? `fillAfterSaveFlagEnabled` or something alike!

nick-livefront (changes requested):

apps/browser/src/autofill/background/abstractions/overlay.background.ts
jaasen-livefront: @nick-livefront It isn't, nice catch! Removed. https://github.com/bitwarden/clients/pull/20892/commits/35dee9d1bd7b3fadda9cbfb71240e5e7da7e2670

libs/common/src/enums/feature-flag.enum.ts
jaasen-livefront: Good 👁️ . I updated the description to match the ticket number here. ;)

apps/browser/src/vault/popup/components/vault/add-edit/add-edit.component.ts
jaasen-livefront: Agreed. I've updated them to better reflect their use https://github.com/bitwarden/clients/pull/20892/commits/35dee9d1bd7b3fadda9cbfb71240e5e7da7e2670

nick-livefront (dismissed):
Thanks for updating those!

apps/browser/src/vault/popup/components/vault/add-edit/add-edit.component.ts
```diff
@@ -336,6 +351,20 @@ export class AddEditComponent implements OnInit, OnDestroy {
       return;
     }
 
+    if (this.inSingleActionPopout && this.fillOnSuccessfulSave) {
+      const senderTab = await firstValueFrom(this.vaultPopupAutofillService.currentAutofillTab$);
+      await this.vaultPopupAutofillService.doAutofill(cipher, false, true);
```
bensbits91: Looks good overall. Approving with one thought:

We might have an edge case where the cipher is saved but autofill didn't work, and the user would only see a "Login saved" success notification with no indication that autofill didn't work.

The new `inSingleActionPopout && fillOnSuccessfulSave` branch in `onCipherSaved` discards the boolean returned by `vaultPopupAutofillService.doAutofill(...)`, so the success-path messaging fires regardless of whether autofill actually succeeded. `doAutofill` returns false if the master-password reprompt is declined, page details haven't been collected on the sender tab, or the underlying autofill call throws. The cipher is genuinely saved in all of these cases, but the user might be confused as to why autofill didn't work.

If we want to address this edge case, we could gate the `showLoginSavedNotification` dispatch on the `didAutofill` return value and give `closeSingleActionPopout` a delay on failure so the existing error toast is actually visible.

I'm not sure how common or disruptive this scenario will be, so I'll leave it to your discretion on whether to address it. Let me know if you'd like more details.

bensbits91 (dismissed):
Approved with one comment about a potential edge case. Feel free to disregard my comment or let me know if you want more info. Looks good :)

apps/browser/src/vault/popup/components/vault/add-edit/add-edit.component.ts
jaasen-livefront: @bensbits91 Thanks for the comment! While I agree it would be nice to acknoledge to the user autofill somehow failed, we currently don't handle this in any of our current autofill flows. There's no error toast on autofill failure so we can't rely on it either. I think this would be best brought up to @bitwarden/team-autofill-dev to explore a potential solution as a separate discussion/PR.

jaasen-livefront:
@nick-livefront @bensbits91 I merged the target branch so I had to rebase and need re-approval when you have a moment. Ty!

nick-livefront (dismissed):

bensbits91 (dismissed):

jaasen-livefront:
@nick-livefront @bensbits91 Had to resolve a merge conflict if you don't mind re-approving. Ty!

nick-livefront (approved):

dan-livefront (approved):
Approving due to merge conflict resolution

Issue #15746, opened 2025-07-23
Missing plural forms for localization process


mKoonrad:
### Steps To Reproduce

Missing support for plurals forms in required strings.
In English there is just two plural forms: one form and other (most often additional "s").

Most of languages have more that 2 plural forms.
https://www.unicode.org/cldr/charts/47/supplemental/language_plural_rules.html

For example Polish:
1 character (one form): 1 znak
2 characters (two form): 2 znak**i**
5 characters (many form): 5 znak**ów**

This issue is related to most European languages and was reported on Crowdin by many translators (many languages).

### Expected Result

Please add support for plurals. Duplicate strings with (s) variants will not be required with plural strings. Each string will be translated in Crowdin to proper format for translator's language.

You need to add support for forms:
- one
- two
- many
- others

### Actual Result

Incorrect strings which are correct only in some plural cases.

### Additional Context

Strings to fix (identifiers):

- loginUpdateTaskSuccessAdditional
- nCollections
- passwordExposed
- reviewAndChangeAtRiskPasswordsPlural
- limitSendViewsCount
- hours
- minutes
- couldNotDecryptVaultItemsBelow
- characterMinimum
- inputMinLength
- inputMaxLength
- fieldsNeedAttention
- multipleFieldsNeedAttention
- importSuccessNumberOfItems
- filterAppliedPlural
- bulkCollectionAssignmentWarning
- personalItemsTransferWarningPlural
- personalItemsWithOrgTransferWarningPlural

### Operating System

Windows

### Web Browser

Brave

### Environment Versions

Wersja: 2025.6.1
SDK: 'main (1fa0f7a)'
Wersja serwera: 2025.7.1

### Issue Tracking Info

- [x] I understand that work is tracked outside of Github. A PR will be linked to this issue should one be opened to address it, but Bitwarden doesn't use fields like "assigned", "milestone", or "project" to track progress.

NovaSilentium:
Hi there,

Thank you for your report! We use [Crowdin](https://crowdin.com/profile/kspearrin) to manage our translations, and these are commonly done by our amazing translation community!

Please take a look at the localization section of our contributing ​documentation for further information.

https://contributing.bitwarden.com/contributing/#localization-l10n 

New translations can be added, or existing ones can be up-/down-voted.

mKoonrad:
[...]

PR #20899, opened 2026-05-28
[CL-51] create file upload component

BryanCunningham:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/CL-51

## 📔 Objective
- Adds new file upload component and dropzone upload

## 📸 Screenshots
image
image

image
image

image
image

image
image

image
image

image
image

image
image


libs/components/src/file-upload/dropzone.component.ts
```diff
@@ -0,0 +1,158 @@
...
+      "tw-items-center",
+      "tw-gap-4",
+      "tw-py-10",
+      "tw-border",
+      "tw-border-dashed",
+      "tw-rounded-xl",
+
+      "tw-transition-colors",
+      "tw-bg-bg-secondary",
+      // "peer-focus-visible/dropzone-input:tw-border-solid",
```
vleague2: Commented out

libs/components/src/file-upload/file-list.component.html
```diff
@@ -0,0 +1,23 @@
+@if (files().length > 0) {
```
vleague2, line 1: I thought we weren't going to componentize the "uploaded files" list because teams are already using other existing components to do this?

libs/components/src/file-upload/file-upload.component.html
```diff
@@ -0,0 +1,105 @@
...
+
+@if (useDropzoneVariant()) {
+  <div class="tw-flex-col tw-flex tw-gap-2 [&_bit-hint]:tw-m-0">
+    <label
+      [id]="labelId"
+      class="tw-flex tw-items-center tw-gap-0.5 tw-text-sm/5 tw-font-medium tw-max-w-full"
+      [class.tw-text-fg-inactive]="disabled()"
+    >
+      <ng-container *ngTemplateOutlet="labelContent"></ng-container>
+    </label>
```
vleague2: The sonar warning for the label being unassociated seems legit?

libs/components/src/file-upload/file-upload.component.html
```diff
@@ -0,0 +1,105 @@
...
+        class="tw-relative tw-flex tw-shrink-0 tw-items-center tw-border-e tw-border-border-base tw-px-3 tw-py-2 tw-rounded-s-[11px]"
+        [class.tw-bg-bg-quaternary]="!disabled()"
+        [class.tw-bg-bg-inactive]="disabled()"
+        [class.tw-text-fg-inactive]="disabled()"
+      >
+        <span class="tw-text-sm tw-pointer-events-none">{{ "chooseFile" | i18n }}</span>
+      </div>
+      <!-- Button overlay: accessible control with no native file-input value to announce -->
+      <button
+        [id]="inputId"
```
vleague2: Shouldn't this `id` go on the `input`?

libs/components/src/file-upload/file-upload.component.ts
```diff
@@ -0,0 +1,182 @@
...
+   * @see https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Attributes/accept#unique_file_type_specifiers
+   *
+   * NOTE: This is only a user hint. Not a validation
+   **/
+  readonly accept = input("");
+
+  /**
+   * Maximum file size in MB
+   *
+   * NOTE: This is only a user hint. Not a validation
```
vleague2: Should we validate?

libs/components/src/file-upload/file-upload.component.html
```diff
@@ -0,0 +1,105 @@
+<ng-template #labelContent>
+  <ng-content select="bit-label"></ng-content>
+</ng-template>
```
vleague2: Can we do something to require a `bit-label`? I think there's a way to do it with a child query and error if it is not provided

libs/tools/send/send-ui/src/send-form/components/send-details/send-file-details.component.ts
```diff
@@ -34,37 +32,23 @@ export class SendFileDetailsComponent implements OnInit {
...
     this.sendFormService.registerChildForm("sendFileDetailsForm", this.sendFileDetailsForm);
 
     this.sendFileDetailsForm.valueChanges.pipe(takeUntilDestroyed()).subscribe((value) => {
-      this.sendFormService.patchSend((send) => {
-        return Object.assign(send, {
-          file: value.file,
-        });
-      });
+      if (value.file) {
+        this.sendFormService.setFile(value.file);
```
vleague2: Seems like there is a decent amount of refactoring in these team-owned files. Do we need to do all of this just to drop in the new component usage?

libs/components/src/file-upload/file-list.component.html
BryanCunningham: I think this definitely needs to be part of the component. These should always render the same way across experiences. We'll need to swap their implementations out to use this. The thing that remains unclear is which truncation method we want to use. Ours is simple (mostly)CSS only solution. Theirs is a more complex measurement and resize observing. Which... I feel like is overkill for this personally

libs/components/src/file-upload/file-upload.component.ts
BryanCunningham: I don't think so. That is too much logic for this component to handle. We'll never know what types of validations the consumer will need. We _could_ create some validation utils for them to use to reduce boiler plate though. I just don't think building that logic into the component is the right move. Should stay as unopinionated as possible IMHO

libs/components/src/file-upload/file-upload.component.html
BryanCunningham: I can check into that!

libs/tools/send/send-ui/src/send-form/components/send-details/send-file-details.component.ts
BryanCunningham: Seemingly, yes. This was mostly due to the file type changing/being enforced by the new component

libs/components/src/file-upload/file-upload.component.html
BryanCunningham: This _should_ be taken care of now

libs/components/src/file-upload/file-upload.component.html
```diff
@@ -0,0 +1,91 @@
...
+        [class.tw-text-fg-inactive]="disabled()"
+      >
+        <span class="tw-text-sm tw-pointer-events-none">{{ "chooseFile" | i18n }}</span>
+      </div>
+      <!-- Button overlay: accessible control with no native file-input value to announce -->
+      <button
+        type="button"
+        [attr.aria-labelledby]="ariaLabelledBy"
+        [attr.aria-describedby]="ariaDescribedBy()"
+        [disabled]="disabled() || null"
```
vleague2: I believe this is causing it to be inaccessible when `disabled`

libs/components/src/file-upload/file-upload.component.html
```diff
@@ -0,0 +1,91 @@
...
+    <div bitFieldContainer class="tw-relative" [hasError]="!!errorMessage()">
+      <div
+        class="tw-relative tw-flex tw-shrink-0 tw-items-center tw-border-e tw-border-border-base tw-px-3 tw-py-2 tw-rounded-s-[11px]"
+        [class.tw-bg-bg-quaternary]="!disabled()"
+        [class.tw-bg-bg-inactive]="disabled()"
+        [class.tw-text-fg-inactive]="disabled()"
+      >
+        <span class="tw-text-sm tw-pointer-events-none">{{ "chooseFile" | i18n }}</span>
+      </div>
+      <!-- Button overlay: accessible control with no native file-input value to announce -->
```
vleague2: I'm curious why we decided to go this route of having the button/live region instead of using the native input field. From an accessibility perspective, focusing on the file input tells me it's a button. Do we feel like users are going to be expecting/more familiar with the native file input type?

libs/components/src/file-upload/dropzone.component.ts
```diff
@@ -0,0 +1,178 @@
...
+    const input = event.target as HTMLInputElement;
+    if (!input.files?.length) {
+      return;
+    }
+
+    const files = Array.from(input.files);
+    this.emitFiles(files);
+
+    // Clear value so re-selecting the same file still triggers a change event
+    input.value = "";
```
vleague2: Related to the native vs not native question for `file-upload`, the dropzone does expose its native input. And we are clearing the `value`, so the screenreader reads "no file selected" when focusing on the dropzone, even when files are selected. The file list below is invisible to a screenreader until they navigate to it. Could we do an `aria-describedby` and keep the `value` set here? Do we need to fire a change event when the same file is re-selected?

(Perhaps we're worried about how the native file input only reads out the first file as its `value` when multiple files are selected? I think associating the file list with the input would address that concern.)

libs/components/src/file-upload/file-upload.component.html
```diff
@@ -0,0 +1,91 @@
...
+      <span
+        class="tw-flex tw-overflow-hidden tw-flex-1 tw-min-w-0 tw-px-3 tw-py-2 tw-text-sm"
+        [class.tw-text-fg-body-subtle]="!fileLabel() && !disabled()"
+        [class.tw-text-fg-heading]="!!fileLabel() && !disabled()"
+        [class.tw-text-fg-inactive]="disabled()"
+      >
+        @if (fileLabel(); as label) {
+          <bit-file-name [name]="label" />
+        } @else {
+          {{ "noFileChosen" | i18n }}
```
vleague2: Design spec is different copy, "No file selected"

libs/components/src/file-upload/dropzone.component.html
```diff
@@ -0,0 +1,52 @@
...
+    >
+      {{ "clickToUploadOrDragAndDrop" | i18n }}
+    </span>
+    @if (maxFileSize() != null) {
+      <span
+        class="tw-text-xs"
+        [class.tw-text-fg-body-subtle]="!disabled()"
+        [class.tw-text-fg-inactive]="disabled()"
+      >
+        {{ "maxFileSizeParam" | i18n: maxFileSize() }}
```
vleague2: None of the text in here is being read by the screenreader

libs/components/src/file-upload/dropzone.component.ts
vleague2: It's still saying "no file selected" and then says "2 files uploaded" lol

libs/components/src/file-upload/file-upload.component.html
```diff
@@ -0,0 +1,95 @@
...
+      [accept]="accept()"
+      [maxFileSize]="maxFileSize()"
+      [multiple]="multiple()"
+      [errorMessage]="errorMessage()"
+      [disabled]="disabled()"
+      [ariaLabelledBy]="labelId"
+      [ariaDescribedBy]="ariaDescribedBy()"
+      (filesSelected)="onFilesSelected($event)"
+    />
+    <span class="tw-sr-only" role="status" aria-live="polite" aria-atomic="true">{{
```
vleague2: I might be wrong here, but polite feels a little too delayed as I'm testing it out with the screenreader on. It's reading out the full file upload field info before announcing the status update, which feels maybe like the status should interrupt?

libs/components/src/file-upload/index.ts
```diff
@@ -0,0 +1,4 @@
+export { DropzoneComponent } from "./dropzone.component";
+export { FileListComponent } from "./file-list.component";
+export { FileUploadComponent } from "./file-upload.component";
+export { FileNameComponent } from "./file-name.component";
```
willmartian: `FileUploadComponent` should be the only public export, as the rest are internal helper components, right?

libs/components/src/file-upload/file-list.stories.ts
willmartian, line 1: Since file-list and dropzone are just internal helper components, I find it atypical that they have their own spec and story files. I think there would be less boilerplate and the tests would better reflect actual usage if we tested against the public component instead of its internals.

libs/components/src/file-upload/file-upload.component.ts
```diff
@@ -0,0 +1,292 @@
...
+   **/
+  readonly maxFileSize = input<number>();
+
+  /**
+   * Allow multiple file selection
+   *
+   * NOTE: If `multiple="true"` it will render the 'dropzone' version of the component
+   */
+  readonly multiple = input(false, { transform: booleanAttribute });
```
willmartian: > If `multiple="true"` it will render the 'dropzone' version of the component

Then why is there also a `variant` input? 🤔  

Because we can have a single file dropzone?

libs/components/src/file-upload/file-upload.component.ts
```diff
@@ -0,0 +1,292 @@
...
+
+  /** Error state — shows danger border and message */
+  readonly errorMessage = input<string>();
+
+  /**
+   * Two-way bound file list — use [(files)] for two-way binding
```
willmartian: Avoid explaining how Angular works within comments unless it is unexpected

libs/components/src/file-upload/file-upload.component.ts
```diff
@@ -0,0 +1,292 @@
...
+  private readonly cvaOnChange = signal<(value: File | File[] | null) => void>(() => {});
+  private readonly cvaOnTouched = signal<() => void>(() => {});
+
+  /**
+   * Required for NG_VALUE_ACCESSOR.
+   *
+   * The form value is `File[]` when `multiple` is set, and `File | null` otherwise.
+   * `writeValue` accepts either shape so consumers don't have to pre-normalize.
+   */
+  writeValue(value: File | File[] | null | undefined): void {
```
willmartian: 1. If `File | File[] | null` is the CVA type, we should create a helper type for that instead of repeating it.
2. It is confusing that the CVA has a different type than `files` (`File[]`)
3. I would suggest simplifying this type to just `File[]` (if we don't split dropzone and non-dropzone into two components). In the current form, we are forcing users to do either a runtime typecheck (is Array?) or unsafe assertion (as File). `File[]` alone can express all the same variant states as `File | File[] | null`

libs/components/src/file-upload/file-upload.component.ts
```diff
@@ -0,0 +1,292 @@
...
+
+  /** Error state — shows danger border and message */
+  readonly errorMessage = input<string>();
+
+  /**
+   * Two-way bound file list — use [(files)] for two-way binding
+   *
+   * NOTE: File list only renders in the dropzone variant usage
+   */
+  readonly files = model<File[]>([]);
```
willmartian: In other CVA components, I don't think we also expose a `model` like we are here. Consumers can already bind to `ngModel` directly if they are not using a reactive form (which would be atypical). I would remove this.

libs/components/src/file-upload/file-upload.component.html
willmartian, line 1: I find it surprising that we are not using the form-field component as the wrapping element here. Did you explore that path?

libs/components/src/file-upload/file-upload.component.ts
```diff
@@ -0,0 +1,292 @@
...
+
+  /**
+   * Allow multiple file selection
+   *
+   * NOTE: If `multiple="true"` it will render the 'dropzone' version of the component
+   */
+  readonly multiple = input(false, { transform: booleanAttribute });
+
+  /** Error state — shows danger border and message */
+  readonly errorMessage = input<string>();
```
willmartian: We need to rework this so that errors flow from Angular form validation, the same way that we do with form field and form control. This is a strong reason to consider a design that utilizes bit-form-field (or other code sharing), if possible.

libs/components/src/file-upload/file-upload.component.ts
```diff
@@ -0,0 +1,292 @@
...
+  private readonly fileInput = viewChild<ElementRef<HTMLInputElement>>("fileInput");
+  private readonly dropzoneRef = viewChild(DropzoneComponent);
+  private readonly fileListRef = viewChild(FileListComponent);
+
+  /**
+   * Drives the dropzone-variant aria-live announcement and the post-removal
+   * focus effect. `{ type: "none" }` keeps the live region empty on first
+   * render so screen readers don't announce anything on load.
+   */
+  protected readonly lastAction = signal<
```
willmartian: Would any of this be simplified by using CDK's live announcer? It allows for imperative announcements: https://material.angular.dev/cdk/a11y/overview#liveannouncer

libs/components/src/file-upload/file-upload.component.ts
```diff
@@ -0,0 +1,292 @@
...
+      const remaining = this.files();
+      if (remaining.length === 0) {
+        this.dropzoneRef()?.focus();
+      } else {
+        this.fileListRef()?.focusDeleteAt(Math.min(idx, remaining.length - 1));
+      }
+      this.pendingFocusIndex.set(null);
+    });
+  }
+
```
willmartian: Could you describe our goal around focus management? I wasn't expecting it to get so complex in these components.

willmartian (changes requested):

willmartian (commented):
Do you think file input and dropzone should maybe be two separate components? We do a lot of branching in the template and the TS file. It feels like we have two components in a single file for not much benefit.

libs/components/src/file-upload/file-upload.component.ts
BryanCunningham: @willmartian This is to handle when a file is deleted from the list. It will focus the next item when an item is deleted. Otherwise we lose focus

BryanCunningham:
> Do you think file input and dropzone should maybe be two separate components? We do a lot of branching in the template and the TS file. It feels like we have two components in a single file for not much benefit.

@willmartian Maybe. I did it this way so that we could better control which variant is rendered ie.. when it's multiple, render the dropzone version. If we do not wish to programmatically enforce that rule, we could break these into 2 different components.

willmartian (changes requested):
(from our DMs)

I propose an alternative approach here: https://github.com/bitwarden/clients/pull/21613

Could you check that out and merge it in if you like it?

libs/components/src/file-upload/dropzone.component.ts
BryanCunningham: I think this is working as best it can now.

libs/components/src/file-upload/dropzone.component.html
BryanCunningham: This should be being read now after a recent change

libs/components/src/file-upload/file-upload.component.html
BryanCunningham: This is fixed now with the addition of the announcer strategy vs the sr only text

libs/components/src/file-upload/file-dropzone.component.html
```diff
@@ -0,0 +1,69 @@
+<bit-form-field [control]="formFieldControl">
+  <ng-content select="bit-label" ngProjectAs="bit-label" />
+
+  <div bitCustomInput>
+    <!-- Drag & drop area. Wrapping <label> (no [for]) triggers the descendant input on click;
+       bit-form-field renders the canonical <label [for]> targeting the input via [id]. -->
```
vleague2: It's not an a11y failure but Axe does warn against multiple labels for a form field based on screenreader support https://dequeuniversity.com/rules/axe/4.11/form-field-multiple-labels?application=axeAPI

libs/components/src/file-upload/file-name.component.ts
```diff
@@ -0,0 +1,30 @@
...
+    firstFourth: name.slice(0, firstEnd),
+    middleHalf: name.slice(firstEnd, lastStart),
+    lastFourth: name.slice(lastStart),
+  };
+};
+
+@Component({
+  selector: "bit-file-name",
+  template: `
+    <span class="tw-contents" [attr.title]="name()">
```
vleague2: Any typography styles for this?

vleague2 (commented):
Looking good! One observation about a11y (let me know if you think this is even worth addressing at this point) and one question!

vleague2:
Okay Storybook findings:

Issue with file name truncation
Screenshot 2026-07-13 at 4 06 22 PM

Very random tiny square on focus when the screenreader is on. I think this is not a major concern, I don't see it when the screenreader is off.
Screenshot 2026-07-13 at 4 07 28 PM

Disabled file dropzone is not accessible by keyboard (this might be intended?)

willmartian (commented):
Vicki's above comments still need addressing, but the code generally looks good to me. One change that we _could_ make now that the file input and dropzone are separate components is reintroduce the bifurcated types:

bit-file-upload → File | null
bit-file-dropzone → File[] 

That removes the [0] casting noise that was introduced. Seems worth doing to me, if bit-file-upload will only support a single file.

libs/components/src/file-upload/file-name.component.ts
BryanCunningham: Added a font size here and removed it from the consuming component

BryanCunningham:
@vleague2
> Very random tiny square on focus when the screenreader is on. I think this is not a major concern, I don't see it when the screenreader is off.

This is the screen reader highlighting the actual input. I think it's probably fine

libs/components/src/file-upload/file-upload.component.html
BryanCunningham: @vleague2 I've made the label required in both components

libs/components/src/file-upload/file-dropzone.component.html
BryanCunningham: @vleague2 I changed the label inside the dropzone to just be a plain div with a click handler

willmartian (dismissed):

mcamirault (dismissed):
Tools files look good, and well done not being confused by Send's confusing "edit mode" terminology (we have an open ticket to improve this)

mcamirault (approved):

vleague2 (approved):

vincentsalucci (approved):
SM changes approved.

sbrown-livefront (approved):
Billing looks good! ✅  Loving the simplification.

jengstrom-bw (approved):
vault items look good!

apps/browser/src/_locales/en/messages.json
```diff
@@ -5002,6 +5002,79 @@
...
+      "file_name": {
+        "content": "$1",
+        "example": "image.png"
+      }
+    }
+  },
+  "oneFileUploaded": {
+    "message": "1 file uploaded"
+  },
+  "filesUploaded": {
```
mKoonrad, line 5051: This isn't compatible with plural versions. You need to handle all plural versions (one, few, many, others).
https://github.com/bitwarden/clients/issues/15746

PR #20900, opened 2026-05-28
[CL-998] bit-table-v2 and related components

willmartian:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/CL-998

## ⛓️ Stacked on / blocked by

This PR is being split up. It is stacked on and blocked by the following extracted PRs, which should merge first:

- [x] #21788 — [CL-1272] Support visually-hidden labels in bit-form-field
- [x] #21789 — [CL-1273] Add bitMenuClose directive to decouple menu dismissal from ARIA roles
- [x] #21790 — [CL-1274] Add bit-page layout region for bit-layout main content

Once those merge, this branch will be rebased on `main` and the duplicated changes will drop out of the diff.

> **Note:** `bit-badge-group` (#21070 — [CL-1176]) is no longer required by this PR. The kitchen-sink demo was switched back to the existing `bit-badge-list`, and all badge-group work was removed from this branch so #21070 owns it exclusively — eliminating the duplication and rebase-conflict risk.

## 📔 Objective

Adds `<bit-table-v2>`, a column-driven evolution of `<bit-table>`. It lays out a CSS grid over ARIA-role elements (`role="table"`/`"row"`/`"columnheader"`/`"cell"`) rather than the CDK table; OnPush throughout, with columns registering on the table via DI.

A table is bound to a slim **`TableDef`** via `[tableDef]`; everything else is configured with inputs on the table and its helper components.

### `TableDef`

`TableDef<T, S>`, built with `defineTable(data)`, is the **typed contract** bound via `[tableDef]`. It carries just the row `data` (a `Signal<T[]>`) and the typed `columns` references. It exists for the one thing inputs can't do: thread the row type `T` (and declared synthetic column names `S`) across the `<bit-table-v2>` ↔ `*bitCellDef` boundary so `let row` and the column keys are strictly typed.

```ts
type User = { id: number; name: string; email: string };

const table = defineTable<User>(this.users); // this.users: Signal<User[]>
table.columns.email; // typed ColumnRef — drives `*bitCellDef`'s `let row`; `columns.emial` fails to compile
```

- **Typed refs** — `table.columns.name` carries `T`; the row type flows to `*bitCellDef` so `let row` is typed via `ngTemplateContextGuard` (`row.emial` fails to compile). Synthetic columns (actions, etc.) are declared via the second type param `S` (`defineTable<User, "actions">(...)`) and referenced the same way.
- **Reactive data** — `data` is a `Signal<T[]>`; the table re-renders as it changes. Mutate by updating your own writable signal (no `setData`).
- **Composable** — columns register via a DI lookup of the ancestor `<bit-table-v2>`, so a `<bit-column>` can live inside another component's view; pass the `TableDef` in for its typed refs.

Everything else is an input on the table or a helper component:

| Input / helper          | Purpose                                                                 |
| ----------------------- | ----------------------------------------------------------------------- |
| `[displayedColumns]`    | Columns to show, in order. Omit to show all in declaration order.       |
| `[(sort)]`              | Active sort (`{ column, direction }`); header clicks cycle it.          |
| `[loading]`             | Show skeleton rows in place of data.                                    |
| `[filter]`              | Client-side row test `(row, filterValues) => boolean`.                  |
| `[selection]`           | Selection config (`{ multiple, canSelect }`); `(selectedChange)` emits. |
| `<bit-table-paginator>` | Projected; owns page state and slices the rows.                         |

`filterValues()` and `filtered()` are signals on the table, readable off a `#t="bitTableV2"` template ref for server-side queries.

### Authoring modes

- **Column-def** — `<bit-column>` wrappers; the cell template and key come from `*bitCellDef="table.columns.x"`, the header from `<bit-header-cell>`. Column metadata is inputs (`sortable`, `defaultSort`, `sortFn`, `width`). Columns register via DI, so a `<bit-column>` can live inside another component's view.
- **Manual** — project `<bit-header-row>` / `<bit-row>` directly (no `TableDef`) for simple presentational tables.

```html
<bit-table-v2 [tableDef]="table">
  <bit-column sortable defaultSort="asc">
    <bit-header-cell>Name</bit-header-cell>
    <bit-cell *bitCellDef="table.columns.name; let row">{{ row.name }}</bit-cell>
  </bit-column>
</bit-table-v2>
```

### Filtering, search & toolbar

Filtering works like a **reactive form**:

- Filter controls (`bit-filter-menu` / `bit-filter-toggle`) projected into the toolbar each declare a `key` and own their selection. Projected into the table, each resolves the table's `FILTER_HOST` token by DI and self-registers (so a chip never depends on the table type and works outside a table), and the table collects their values into **`table.filterValues()`** — a `{ key: value }` object, like a `FormGroup`'s `.value`.
- You provide one **`[filter]="(row, values) => boolean"`**; the table applies it to derive `filtered()`, which drives the rendered rows, select-all scope, the paginator count, and a "no matches" empty state. Annotate the fn's `values` parameter with your filter shape and the table types `filterValues()` to match. Seed initial values with `[filters]`. Omit `[filter]` for **server-side** filtering and read `filterValues()` to build a query.
- A `<bit-search>` projected into the table is **auto-adopted** as a filter under the `search` key (no `key`, no `ngModel`) — its term joins `filterValues().search`.
- A **`filter-menu`** family composes the chips: `bit-filter-option`, `bit-filter-section` (collapsible, with a selected-count berry), `bit-filter-divider`, in-menu search past 10 options, single vs. multi (`[multiple]`) select, and a selected-count berry on the chip. On small screens the filter row collapses into a single trigger that opens a dialog of the filters, with active filters shown as dismissible chips.

`<bit-table-toolbar>` renders in the chrome above the header row, in two rows: the `<bit-search>` (its own slot, capped at ~400px) and `slot="end"` controls share the top row; the filter chips get their own row beneath with a live "N items" count at the end. Group related chips with a `<bit-filter-divider>`.

```html
<bit-table-v2 [tableDef]="table" [filter]="filter" [filters]="{ type: 'login' }">
  <bit-table-toolbar>
    <bit-search class="tw-flex-1" placeholder="Search"></bit-search>
    <button slot="end" bitButton buttonType="primary">New</button>
    <bit-filter-menu key="type" placeholderText="Type" unsetLabel="All">
      <bit-filter-option [value]="'login'">Login</bit-filter-option>
    </bit-filter-menu>
    <bit-filter-divider></bit-filter-divider>
    <bit-filter-toggle key="favorites" label="Favorites" icon="bwi-star"></bit-filter-toggle>
  </bit-table-toolbar>
  <bit-column>…</bit-column>
</bit-table-v2>
```

### Selection & bulk actions

- `[selection]="{ multiple, canSelect }"` prepends an internal checkbox column with select-all over the filtered rows; `canSelect` scopes which rows are selectable and is enforced on programmatic selection too. `(selectedChange)` emits the selected rows.
- `<bit-bulk-actions-bar>` projected inside the table reads `selectedCount` from the table's selection via DI; its clear button clears the selection. No explicit wiring.

### Pagination

- Project a `<bit-table-paginator [pageSize]="10" [pageSizeOptions]="[10, 25, 50, 100]" />` and the table slices its filtered + sorted rows to the page. The paginator owns the page state (`[(pageIndex)]`, `[pageSize]`) and reads the total back from the table; its footer shows the row range, a page-size select, prev/next, and a page input. **Client-side** by default.
- Supply an explicit `[length]` for **server-side** mode — the table doesn't slice; `data()` is expected to hold just the current page, fetched off the paginator's `pageIndex()` / `pageSize()`.
- Mutually exclusive with virtualization.

### URL sync

- Bind **`[queryParam]="ns"`** and the table mirrors its filters, sort, and pagination to the URL under that namespace — `?ns.type=login&ns.favorite=true&ns.sort=name&ns.direction=asc&ns.page=2` — so a filtered/sorted/paged view is shareable and survives a reload. On load the params restore state (winning over `[filters]` and `defaultSort`); on change the table writes back with `replaceUrl` (flat history) and `merge` (params it doesn't own are untouched). Inactive/default facets leave no param behind, so the URL self-cleans.
- Each filter chip syncs under its own `key`; `sort`/`direction`/`page`/`pageSize` use those keys within the namespace. Values are read back by shape (`true`/`false` → boolean, integers → number, repeated key → array). A distinct prefix per table lets several sync on one page; needs a router in context (a no-op without one).
- Built on **`queryParamStore<T>`** (in `@bitwarden/components`), a generic writable signal mirrored to a URL namespace — usable directly to sync any component-owned state (a tab, a search term) to the query string, table or not.

### Sizing, scrolling & the sticky header

- **`[height]`** (`"fill" | number`) — `"fill"` grows the table to fill a bounded, full-height container (idiomatically the body of a `bit-page`); a number caps the body at that many rows (minimum 4), then scrolls with the header pinned. The row-count cap needs a known row height, so it applies only to virtualized tables and is otherwise a no-op. Omit to grow to content.
- **`[virtualRowHeight]`** — opts into a CDK virtual-scroll viewport for large datasets (needs a bounded height via `[height]`; `[trackBy]` recommended).
- **`<bit-page>`** — a layout primitive for the main content area: a full-height, scrolling body that a `fill` table grows into. (Header/footer slots are deferred until a consumer needs them.) Migrate a view to it incrementally — `bit-layout` drops its own padding via `:has(bit-page)`, no shell-wide flag.

### Data states

The table covers the full lifecycle — **loading → data → empty**:

- **Empty** — when no rows render (empty data or a filter that excluded everything), it shows a default `<bit-no-items>` below the header; override by projecting `slot="empty"`.
- **Loading** — bind `[loading]` (e.g. a resource's `isLoading()`) to show skeleton rows (count via `[loadingRows]`, default 3). Each column can declare a `<bit-cell *bitCellLoading>` skeleton; columns without one get a default. Loading takes precedence over empty, and the header stays put.

### List presentation & row grouping

A **`[presentation]`** input (`"table"` | `"list"`) switches the same engine — data, filtering, sort, selection — between the bordered column grid and a `bit-item`-style card list. In `list` mode the column header is visually hidden but kept in the accessibility tree (and the sort affordance is suppressed, since there's no visible header to click — a list-mode sort control is a follow-up). The goal is a future responsive table ⇄ list switch with no state to re-thread.

**`<bit-row-group>`** is declarative, predicate-based grouping that renders in `list` presentation:

```html
<bit-table-v2 [tableDef]="table" presentation="list">
  <bit-row-group [match]="isFavorite" collapsible>
    Favorites
    <!-- rows no subgroup claims render flat first, then the "Cards" subgroup -->
    <bit-row-group [match]="isCard">Cards</bit-row-group>
  </bit-row-group>
  <bit-row-group [match]="all" collapsible>All items</bit-row-group>
  <bit-column listRole="title">…</bit-column>
</bit-table-v2>
```

- **Predicate, first-match-wins.** A row joins the first group whose `[match]` claims it, in declaration order. Empty groups are skipped; rows no group claims trail unheadered. A `() => true` group is a natural catch-all.
- **Auto count** on top-level headers; **`collapsible`** headers styled after the extension's section headers (chevron + hover/focus affordance).
- **One level of subgroups** — nest `<bit-row-group>`s; a parent's unclaimed rows render flat above its subgroups (the extension's "flat section then Cards subgroup" shape).
- **Guardrails** — grouping no-ops with a warning in `table` presentation, and errors when combined with a paginator. Virtualized grouping is not yet supported (see `libs/components/src/table/v2/TODO.md`).

### Other

- **Cell slots** — `<bit-cell>` slot vocabulary (`start` / default / `secondary` / `end`) for rich cells, with a `truncate` toggle.
- **Refreshed chrome** — rounded container, subtle shadow, secondary-bg header per the updated design spec.

No existing `<bit-table>` consumers are migrated in this PR. `bit-table` is now marked deprecated in favor of `bit-table-v2`; see the migration guide in the Storybook docs.

## 📸 Screenshots

Stories under `Component Library/Table V2`:

[CL-1272]: https://bitwarden.atlassian.net/browse/CL-1272?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[CL-1273]: https://bitwarden.atlassian.net/browse/CL-1273?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[CL-1274]: https://bitwarden.atlassian.net/browse/CL-1274?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[CL-1176]: https://bitwarden.atlassian.net/browse/CL-1176?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


willmartian:
@claude please review. Pay extra attention to API design, not just implementation. Nothing is set in stone at this time.

libs/components/src/table/v2/table-v2.component.ts
```diff
@@ -0,0 +1,362 @@
...
+   * so manual-mode tables need not bind it.
+   */
+  readonly table = input(new TableModel<T>());
+
+  /**
+   * Defaults to `"auto"`. Forced to `"fixed"` when virtualization is on
+   * ({@link rowSize} is set) — virtual scrolling requires predictable row
+   * geometry, which a fixed layout provides.
+   */
+  readonly layout = input<"auto" | "fixed">("auto");
```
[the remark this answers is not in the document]
willmartian: Removed the `layout` input along with its JSDoc and the `[layout]` reference in the mdx — it was never read. Done in 0e884e6.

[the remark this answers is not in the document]
willmartian: Removed the `layout` input along with its JSDoc and the `[layout]` reference in the mdx — it was never read. Done in 0e884e6.

libs/components/src/table/v2/table-v2.component.html
```diff
@@ -0,0 +1,118 @@
...
+      }
+    } @else {
+      <!--
+        Manual mode. The consumer projects `<bit-header-row>` and `<bit-row>`
+        directly; rows fall back to `grid-auto-columns: 1fr` because no
+        `<bit-column>` registry exists to drive `grid-template-columns`.
+      -->
+      <ng-content select="bit-header-row"></ng-content>
+      <ng-content select="bit-row"></ng-content>
+    }
```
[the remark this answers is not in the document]
willmartian: Added a `LogService.warning` in `ngAfterContentInit` when `selection` is configured but no `<bit-column>` is projected (manual mode), so the silent no-op now surfaces in dev. 0e884e6.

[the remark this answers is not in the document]
willmartian: Added a `LogService.warning` in `ngAfterContentInit` when `selection` is configured but no `<bit-column>` is projected (manual mode), so the silent no-op now surfaces in dev. 0e884e6.

libs/components/src/table/table-data-source.ts
```diff
@@ -119,98 +212,25 @@ export class TableDataSource<T> extends DataSource<T> {
...
-      } else if (valueA != null) {
-        comparatorResult = 1;
-      } else if (valueB != null) {
-        comparatorResult = -1;
-      }
-
-      return comparatorResult * directionModifier;
-    });
+    return sortRows(data, sort, (d, id) => this.sortingDataAccessor(d, id));
   }
```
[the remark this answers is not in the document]
willmartian: Good catch — confirmed the regression (`getCheckedUsersInVisibleOrder` reads `filteredData` expecting sorted order, and the v1 in-place sort was what kept it aligned). Rather than patch v1, I reverted `table-data-source.ts` to its original in-place `sortData` and gave bit-table-v2 its own self-contained `sortRows`. A v2-introduction PR shouldn't be changing v1's runtime semantics for a dedup, and since v1 is slated for removal the minor duplication is acceptable. `people-table-data-source.spec` passes again. Fixed in 2a6cd71.

apps/browser/src/platform/popup/layout/popup-layout.stories.ts
```diff
@@ -326,6 +344,357 @@ class MockSettingsPageComponent {}
...
+      <!-- fill needs a bounded flex parent; popup-page's content slot isn't flex, so wrap it. -->
+      <div class="tw-flex tw-h-full tw-min-h-0 tw-flex-col">
+        <bit-table-v2
+          [tableDef]="table"
+          [filter]="filter"
+          [presentation]="presentation()"
+          [displayedColumns]="displayedColumns"
+          fill
+        >
+          <bit-table-toolbar>
```
shane-melton, line 514: ❓ This may be too early, but I'm curious what the API surface would look like here for the browser where:

- Filters are collapsed into a bottom sheet dialog
- The bit-search is visible at all times
- Only dismissible filter chips are shown when active

libs/components/src/table/v2/table-v2.mdx
```diff
@@ -0,0 +1,580 @@
...
+filtering, omit `[filter]` and read `filterValues()` off a `#t="bitTableV2"` template reference to
+build your query (feed pre-filtered rows to `data`). See the
+[Filter Menu](?path=/docs/component-library-filter-menu--docs) for composing the chips (sections,
+in-menu search, and a radio + checkbox group in one chip, which nests under the chip's key).
+
+<Canvas of={stories.Filterable} />
+
+### Search
+
+A `<bit-search>` projected into the table (typically the toolbar) is adopted automatically as a
```
shane-melton: 💭 This is pretty cool and seems like magic haha. My only (very small) hesitation is that its not obvious how a bit-search value gets piped to the `filterValues` as the `search` key by just looking at the template. 

Would it make sense to allow `bit-search` to specify its own filter `key`? It could still default to "search".

libs/components/src/filter-menu/filter-menu.mdx
shane-melton, line 1: 👍 Just wanted to say this is a really clean component and API! 🎉

shane-melton (commented):
Just a few comments on the pieces I was looking through. Really great work!

apps/browser/src/platform/popup/layout/popup-layout.stories.ts
willmartian: The API is the same. This is currently controlled by viewport size--it is treated as a responsive view instead of a web vs extension difference (see in Storybook on a small screen). I believe design and I are aligned on this approach (pending further review of the prototype). However, it could be changed to an explicit input on the table toolbar as needed: `[overlayFilters]="true"`, etc.

libs/components/src/menu/menu.component.ts
```diff
@@ -39,4 +39,16 @@ export class MenuComponent {
...
   }
+
+  /**
+   * Closes the menu only when a menu item is activated. Clicks on other projected
+   * content — e.g. a filter chip's checkboxes, search field, or section headers
+   * (`role="menuitemcheckbox"` / `"menuitemradio"` / plain controls) — leave the
+   * menu open. `bitMenuItem` always carries `role="menuitem"`.
+   */
+  protected onContentClick(event: MouseEvent): void {
+    if ((event.target as HTMLElement | null)?.closest('[role="menuitem"]')) {
```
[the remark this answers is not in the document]
willmartian: [bf6ca7e](https://github.com/bitwarden/clients/pull/20900/commits/bf6ca7ed5f7c761fe7d269840644a07e2185692f)

apps/web/src/app/layouts/product-switcher/product-switcher-content.component.html
```diff
@@ -11,6 +11,7 @@
     >
       <a
         *ngFor="let product of products.bento"
+        bitMenuClose
```
willmartian, line 14: These were added because the menu was updated to not auto-close by default. Consumers passing custom content into menus can use this or `bitMenuItem` to auto-close on click.

libs/components/src/filter-menu/filter-menu.stories.ts
```diff
@@ -0,0 +1,81 @@
...
+        <bit-filter-section label="Engineering" collapsible>
+          <bit-filter-option [value]="'cicd'" [count]="2">CI/CD</bit-filter-option>
+          <bit-filter-option [value]="'devtools'" [count]="1">Dev tools</bit-filter-option>
+        </bit-filter-section>
+        <bit-filter-section label="Operations" collapsible>
+          <bit-filter-option [value]="'support'" [count]="4">Support</bit-filter-option>
+        </bit-filter-section>
+      </bit-filter-chip>
+
+      <bit-filter-toggle key="favorites" label="Favorites" icon="bwi-star"></bit-filter-toggle>
```
willmartian: @BryanCunningham thoughts on this usage? and the naming? 

`bit-filter-chip` would replace `bit-chip-filter`

`filter-` felt like the appropriate namespace when I was doing this (vs `chip-`), because we also have a `filter-toggle`. 

But I'm not sure what is best

willmartian: Maybe resolved? renamed `bit-filter-chip` → **`bit-filter-menu`** (`FilterChipComponent` → `FilterMenuComponent`) in a25e2c5811.

The old `bit-chip-filter` is now marked `@deprecated` (successor: `bit-filter-menu` — same chip primitives, but composable + table-integrated), and its Storybook entry is labeled *(Deprecated)*.

libs/components/src/filter-menu/filter-dialog.component.html
```diff
@@ -0,0 +1,59 @@
+<bit-dialog [title]="activeFilter()?.label() ?? ('filter' | i18n)" dialogSize="small">
```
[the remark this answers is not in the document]
willmartian: Fixed in 1b4da197ac — added `"filter": { "message": "Filter" }` to `apps/browser/src/_locales/en/messages.json` and `apps/desktop/src/locales/en/messages.json` (en only; Crowdin handles the rest), placed next to the existing `filters` key. So the small-screen filter dialog title no longer renders the raw key on browser/desktop.

libs/components/src/table/v2/bit-row.component.ts
```diff
@@ -0,0 +1,54 @@
...
+  protected readonly gridTemplateColumns = computed(() => this.table?.gridTemplateColumns());
+
+  /**
+   * Row chrome. The grid classes lay the cells out in both presentations; the
+   * rest is presentation-specific: `table` connects rows with a top divider,
+   * `list` renders each row as a standalone `bit-item`-style card (background,
+   * rounded corners, spacing, hover).
+   */
+  protected readonly hostClasses = computed(() => {
+    const layout = "tw-grid tw-grid-flow-col tw-auto-cols-fr";
```
shane-melton: 🎨 @willmartian Can we add `tw-group/row` to the host here so that cell definitions can add their own hover effects via `group-hover/row:*`? Vault quick copy / launch actions are only to be rendered on row hover.

libs/components/src/filter-menu/filter-menu.component.ts
```diff
@@ -0,0 +1,310 @@
...
+  readonly placeholderText = input.required<string>();
+
+  /**
+   * Label shown after the prefix while inactive, e.g. "All" → "Type: All". Omit
+   * to show just the prefix when nothing is selected.
+   */
+  readonly nullLabel = input<string>();
+
+  /** Multi-select (checkbox) when `true`; single-select (radio) when omitted. */
+  readonly multiple = input(false, { transform: booleanAttribute });
```
vleague2, line 116: I was under the impression at one point that a filter menu could contain multiple types of inputs -- like some sections were checkboxes and some radios, within the same menu. Did I dream that??

libs/components/src/filter-menu/filter-menu.component.ts
```diff
@@ -0,0 +1,310 @@
...
+
+  protected readonly baseChip = inject(BaseChipDirective, { host: true });
+
+  /** The selection: a single value (single-select) or an array (multi-select). */
+  private readonly _value = signal<unknown>(undefined);
```
vleague2, line 125: Feels like we could type this more strongly (and others typed as unknown in this file)

vleague2 (commented):
I did not get very far in my review

libs/components/src/filter-menu/filter-menu.component.ts
vleague2: Discussed with Emily today -- I did not dream this but it is also no longer the plan

libs/components/src/table/v2/bit-row.component.ts
willmartian: Yes, can do! Ty for the suggestion.

libs/components/src/filter-menu/filter-menu.stories.ts
```diff
@@ -0,0 +1,81 @@
...
+ * `table.filterValues()`; here they simply display their selection.
+ */
+@Component({
+  selector: "filter-menu-demo",
+  changeDetection: ChangeDetectionStrategy.OnPush,
+  imports: [FilterMenuModule],
+  template: `
+    <div class="tw-flex tw-flex-wrap tw-items-start tw-gap-2 tw-p-4">
+      <bit-filter-menu key="type" placeholderText="Type" nullLabel="All">
+        <bit-filter-option [value]="'login'" [count]="12">Login</bit-filter-option>
```
vleague2, line 22: So the number of matching table rows (what I am presuming `count` is) is provided by the consumer, not calculated by the table? I guess we wouldn't have all that info in a paginated scenario for example?

libs/components/src/filter-menu/filter-toggle.component.ts
```diff
@@ -0,0 +1,99 @@
...
+  /** @see FilterPresenter.optionsTemplate — a toggle has no drill-in; it flips in place. */
+  readonly optionsTemplate = computed<TemplateRef<unknown> | undefined>(() => undefined);
+
+  /** The displayed icon — the filled variant while active, when one is conventionally available. */
+  protected readonly displayIcon = computed<BitwardenIcon | undefined>(() => {
+    const icon = this.icon();
+    if (!icon) {
+      return undefined;
+    }
+    return this._value() ? ((icon + "-f") as BitwardenIcon) : icon;
```
vleague2: This seems like it would break if a filled isn't available

libs/components/src/form-field/form-field.stories.ts
```diff
@@ -259,6 +259,34 @@ export const Required: Story = {
...
+  render: (args) => ({
+    props: {
+      formObj: formObj,
+      ...args,
+    },
+    template: /*html*/ `
+      <!--
+        A visually-hidden (\`sr-only\`) label keeps the field accessible while
+        collapsing the label/field gap, so the field stays vertically centered
+        alongside sibling controls in a row.
```
vleague2, line 272: Unnecessary Claude comment

libs/components/src/table/v2/bit-cell.component.ts
```diff
@@ -0,0 +1,50 @@
...
+  readonly truncate = input(true);
+
+  protected readonly secondarySlot = viewChild<ElementRef<HTMLDivElement>>("secondarySlot");
+  protected readonly endSlot = viewChild<ElementRef<HTMLDivElement>>("endSlot");
+
+  protected readonly hasSecondary = signal(false);
+  protected readonly hasEnd = signal(false);
+
+  ngAfterContentChecked(): void {
+    this.hasSecondary.set((this.secondarySlot()?.nativeElement.childElementCount ?? 0) > 0);
```
vleague2: Does the `empty` css selector not work instead of these?

libs/components/src/table/v2/bit-header-cell.component.html
```diff
@@ -0,0 +1,29 @@
...
+  >
+    @if (sortable()) {
+      <button
+        type="button"
+        [class]="sortButtonClasses"
+        [attr.aria-pressed]="active()"
+        (click)="onSortClick()"
+      >
+        <ng-container *ngTemplateOutlet="content"></ng-container>
+        <i class="bwi tw-ms-2" [class]="sortIcon()"></i>
```
vleague2: For follow-up, use `bit-icon`?

libs/components/src/table/v2/bit-row.component.ts
vleague2: Each row might need a unique tw group name 🤔 You'll have to see if you get group naming collisions in testing

libs/components/src/table/v2/bit-table-toolbar.component.html
```diff
@@ -0,0 +1,61 @@
...
+          ></bit-berry>
+        }
+      </div>
+    }
+    <div class="tw-flex tw-items-center tw-gap-2 tw-ms-auto">
+      <ng-content select="[slot=end]"></ng-content>
+    </div>
+  </div>
+  <!--
+    Desktop filter row: the projected chips and the item count. Hidden (not removed)
```
vleague2: I am interpreting this as a desktop-client-specific feature but I feel like that is not what's happening?

libs/components/src/table/v2/table-v2.mdx
```diff
@@ -0,0 +1,610 @@
...
+Project one `<bit-header-row>` and any number of `<bit-row>` components, iterating your own data.
+Columns default to an equal share of the width.
+
+The table picks its mode automatically: any `<bit-column>` you project switches it to the full
+column-driven table; otherwise your `<bit-header-row>`/`<bit-row>` render directly. Don't mix the
+two in one table — if you need sort, selection, or any other built-in feature, use columns.
+
+## Accessibility
+
+- Always include a header — even an empty one — for each column; screen readers use it to give cell
```
vleague2: Perhaps should specify visually empty but with screenreader-accessible text

libs/components/src/table/table-data-source.ts
```diff
@@ -15,8 +15,8 @@ export type FilterFn<T> = (data: T) => boolean;
 // Loosely based on CDK TableDataSource
 //  https://github.com/angular/components/blob/main/src/material/table/table-data-source.ts
 export class TableDataSource<T> extends DataSource<T> {
-  private readonly _data: BehaviorSubject<T[]>;
-  private readonly _sort: BehaviorSubject<Sort>;
+  private readonly _data = new BehaviorSubject<T[]>([]);
+  private readonly _sort = new BehaviorSubject<Sort>({ direction: "asc" } as Sort);
```
vleague2: Do these changes have to happen to the existing table in this PR? If we can avoid adding any regression testing scope to this ticket, that would probably help us

vleague2 (commented):
Overall shape of this makes sense to me. I have some non-blocking thoughts as I was reading through, you can decide what should be a follow-up

libs/components/src/form-field/form-field.stories.ts
willmartian: Resolved in 8a24f7cc3d

libs/components/src/table/v2/table-v2.mdx
willmartian: Resolved in 8a24f7cc3d

libs/components/src/table/v2/bit-row.component.ts
willmartian: Resolved in 8a24f7cc3d

willmartian: Well, partially resolved. This will not work if/when we add nested rows--hover on a parent would affect all children. Not sure if we should solve that now or later--I lean later when I have more time to think about it.

libs/components/src/table/v2/bit-header-cell.component.html
willmartian: Resolved in f09fa20643

libs/components/src/table/table-data-source.ts
willmartian: Resolved in 4e24e4aeff — reverted `table-data-source.ts` to main. The changes (a cosmetic field-init move and a new `sort$` observable) weren't needed: bit-table-v2 has its own model and never consumes `TableDataSource`, and `sort$` had no consumers. This keeps v1's runtime out of the PR's scope. `people-table-data-source.spec` still passes.

libs/components/src/filter-menu/filter-menu.stories.ts
willmartian: This one I am leaning towards deferring. The filters are technically independent of the table right now, they operate over a `FilterHost` abstraction. While I could expand that abstraction to have a full reference to the host's (table's) data, that would only work for client side filtering and not server side. I think it is solvable I just need to think about it.

libs/components/src/filter-menu/filter-menu.component.ts
willmartian: I actually think `unknown` is correct here. The type comes from the `value` passed to the child option component, and Angular cannot infer a component's generic type parameter from its projected content. 

And furthermore, this `unknown` actually doesn't bite, since it is never operating on in code. It is passed back to the filter function defined on the table/host, which is strongly typed. Generally, an unknown that is never operated on and a generic `T` are similar.

libs/components/src/table/v2/bit-cell.component.ts
willmartian: Resolved in abec5598f8 — good call. Switched both slots to `empty:tw-hidden` (the same projection-slot pattern used across the lib: tab-group, header, banner, etc.) and removed the `ngAfterContentChecked` + viewChild + signal machinery entirely. `:empty` handles the projection case correctly (comment anchors are ignored per spec, and Angular strips the template whitespace), so this drops a hook that ran every change-detection cycle.

libs/components/src/table/v2/bit-table-toolbar.component.html
willmartian: 🤖: Good catch — this is a wording collision, not a client-specific feature. The row is gated on `isLargeScreen()`, i.e. a viewport breakpoint (≥ md), and the whole toolbar lives in `libs/components`, so it's shared across web, browser, and desktop. "Desktop" here was loose shorthand for "wide viewport," which reads as the Desktop *client* in this repo.

Reworded the comment to "Wide-viewport filter row," and swept the same loose "desktop" wording in `filter-tokens.ts` and `filter-menu.component.ts` for consistency (now "wide viewport" / "wide-viewport chip row"). No behavior change.

libs/components/src/filter-menu/filter-toggle.component.ts
willmartian: 🤖: Good catch — you were right. The old `(icon + "-f") as BitwardenIcon` guessed at a filled variant that mostly doesn't exist (only `bwi-cog-f`, `bwi-lock-f`, and `bwi-star-f` do), so any other icon would render a nonexistent `bwi-*-f` class and blank out while active — and the `as` cast hid it from the type checker.

Replaced the guess with an explicit, type-checked `iconActive` input. If it's omitted, `icon` is used in both states, so nothing silently breaks. Matches the existing `iconActive` convention over in `bottom-navigation`.

```html
<bit-filter-toggle key="favorites" label="Favorites" icon="bwi-star" iconActive="bwi-star-f"></bit-filter-toggle>
```

Fixed in 53b7c4e4bf.

libs/components/src/filter-menu/filter-menu.stories.ts
willmartian: We talked sync about this. Decided to auto-compute unless `count` is passed explicitly. 2d52bdf29b 

This was actually much simpler after the `bitTableFilter` directive removal in [98c760e](https://github.com/bitwarden/clients/pull/20900/commits/98c760e977b19a801266c963517f1fde9f40ff5c)

libs/components/src/table/v2/bit-cell.component.ts
```diff
@@ -33,18 +25,7 @@ import { TypographyModule } from "../../typography";
...
-  protected readonly secondarySlot = viewChild<ElementRef<HTMLDivElement>>("secondarySlot");
-  protected readonly endSlot = viewChild<ElementRef<HTMLDivElement>>("endSlot");
-
-  protected readonly hasSecondary = signal(false);
-  protected readonly hasEnd = signal(false);
-
-  ngAfterContentChecked(): void {
-    this.hasSecondary.set((this.secondarySlot()?.nativeElement.childElementCount ?? 0) > 0);
-    this.hasEnd.set((this.endSlot()?.nativeElement.childElementCount ?? 0) > 0);
-  }
```
vleague2, line 49: :cat-clap:

vleague2 (approved):
No API-related concerns for the dev preview merge!

PR #20912, opened 2026-05-29
[PM-38330] Fix per-request memory leak in `bw serve` (authStatusFor$)

### LINKED ISSUE #20548: bw serve leaks ~10 KB per HTTP request regardless of endpoint
## Summary

`bw serve` retains roughly 10 KB of process memory per HTTP request handled, **independent of which endpoint is called** (`/status`, `/sync`, `/object/item/...` all behave the same). Over hours/days of normal operation this accumulates to tens of MB to multiple GB, eventually triggering V8 heap exhaustion (`Ineffective mark-compacts near heap limit ... Allocation failed - JavaScript heap out of memory`) and process death.

The leak is in the request-handling path, not in vault decryption or any sync-specific machinery — `/status`, which doesn't touch vault items, leaks at the same per-request rate as `/object/item/...`.

## Observed in production

Four `bw serve` deployments in our cluster all exhibit the same pattern. Memory growth measured via `container_memory_working_set_bytes` (cAdvisor).

A canonical 30-hour cycle on one deployment — smooth, monotonic, 110 MB → 330 MB before OOM-induced restart:

Image

The same pattern, scaled up: a deployment that ran for **9.5 days without a memory limit** so the only ceiling was V8's internal heap. 130 MB → 1.1 GB, perfectly linear, no plateau:

Image

A deployment with a wrapper script that detects bw process death and restarts in-process — same per-request leak, capped at V8's default ~256 MB old-gen ceiling, ~22 hour cycles repeating for 10 days:

Image

## Diagnostic experiment

To isolate the cause we set `HEALTH_CHECK_INTERVAL=600` (sync once every 10 minutes instead of every 30 seconds) on one deployment. Sync rate dropped 20×; total leak rate only dropped ~35%.

| Config | API call rate (status + sync) | Leak rate | MB per request |
|---|---|---|---|
| 30 s sync wrapper interval (probes + wrapper) | ~720 req/h | 7.73 MB/h | 0.0107 |
| 600 s sync wrapper interval (probes + wrapper) | ~492 req/h | 5.05 MB/h | 0.0103 |

Per-request leak is essentially identical at very different sync frequencies. **The leak is per-HTTP-request, not specific to `/sync`.** If sync was the cause, dropping sync 20× would have reduced leak rate ~95%, not ~35%.

Liveness/readiness probes (which only call `/status`) account for most of the residual API rate — and the residual leak rate scales with them, confirming `/status` leaks at the same rate as endpoints that touch vault data.

Image

## Predicted minimal reproduction

We have not directly run this minimal docker-based repro — the per-request leak rate is inferred from production cAdvisor measurements correlated with API call rate. Based on those measurements, an upstream maintainer should be able to reproduce in minutes:

```bash
# Auth + serve
bw login --apikey   # (BW_CLIENTID, BW_CLIENTSECRET in env)
bw unlock --raw > /tmp/session
export BW_SESSION=$(cat /tmp/session)
bw serve --port 8087 &
SERVE_PID=$!

# Hit /status in a tight loop — no vault items needed
while true; do curl -s http://localhost:8087/status > /dev/null; done &

# Watch RSS climb
while true; do echo "$(date +%T) $(ps -o rss= -p $SERVE_PID) KB"; sleep 60; done
```

Predicted: at ~1 req/s, RSS grows ~36 MB/h, hitting V8's default ~256 MB old-gen ceiling within a few hours.

## Versions tested

- `bw 2025.12.1` (nexe-built single binary in our deployments), confirmed leak.
- Latest npm release `2026.4.1` not directly tested — would be useful to confirm whether the leak still reproduces on current.

## Environment

- Linux x86_64, single-binary `bw` (nexe-bundled Node).
- No proxy, no TLS — request handling is direct HTTP.
- Vault size is independent of leak rate in our observations: deployments serving small vaults (~150 items) leak at the same per-request rate as ones serving thousands of items. Consistent with `/status` leaking at the same rate as `/object/item/...`.

## Suggested investigation

Since `/status` leaks at the same rate as endpoints that touch vault data, the leak is unlikely to be in:

- Vault decryption / cache management
- Sync handlers
- Item-specific endpoint logic

More likely candidates:

- Per-request middleware or logging that retains references
- HTTP response buffer / framework state not being released
- An accumulating in-memory log or telemetry buffer
- Closure-capture of request/response objects in promise chains

A heap profile (`--heap-prof`) taken before and after a few thousand `/status` requests should pinpoint the retainer.

## Existing related issues

- bitwarden/cli#261 (now in the archived `bitwarden/cli` repo) — closed for repository migration. Symptoms match: heap OOM under sustained `bw serve` API traffic from Ansible.
- #8762 — "CLI \`bw serve\` is killed after few minutes on arm64". Closed; symptoms align (OOM under load).

Filing fresh against this repo since both prior reports are closed and neither investigation appears to have continued.

## Workarounds (documented for downstream consumers, not blocking on this issue)

- `NODE_OPTIONS=--max-old-space-size=512` (or higher) defers the OOM but doesn't prevent it
- A wrapper script that detects bw process death and restarts in-process keeps the listening socket present (~7 s outage per restart)
- Client-side retry on `ConnectionError` / `Timeout` covers the restart window transparently

ggiesen:
## 🎟️ Tracking

Fixes #20548

## 📔 Objective

`bw serve` grows memory linearly per HTTP request and never plateaus, eventually exhausting the V8 heap. Reported against 2025.12.1; the defect is present on `main`.

### Root cause

`AuthService.authStatusFor$` (`libs/common/src/auth/services/auth.service.ts`) was a *method*, so it built a fresh `combineLatest([...]).pipe(..., shareReplay({ bufferSize: 1, refCount: false }))` on every call. With `refCount: false` the internal `ReplaySubject` subscribes to the upstreams and never unsubscribes, even after every downstream subscriber is gone. Those upstreams (`keyService.getInMemoryUserKeyFor$`, `tokenService.hasAccessToken$`) are plain pipes off long-lived `stateProvider` subjects that live for the whole process, so each call permanently retains another observer graph that is never garbage-collected.

Every request that resolves an auth status hits this:
- `GET /status` -> `StatusCommand` -> `authService.getAuthStatus()` -> `firstValueFrom(authStatusFor$(userId))`
- `POST /sync` -> `SyncService.fullSync()` -> `firstValueFrom(authService.authStatusFor$(userId))` (`default-sync.service.ts`)

In Kubernetes deployments, liveness/readiness probes hammering `/status` make this present as endpoint-independent growth.

### Fix

Memoize `authStatusFor$` per `userId` with the existing `perUserCache$` helper (`libs/common/src/vault/utils/observable-utilities.ts`), which applies `shareReplay({ bufferSize: 1, refCount: false })` once per `userId` and caches the result. The per-call factory is gone, so a single shared upstream subscription exists per user instead of one per request. This is the same pattern `CipherService` already uses. `getAuthStatus()`, `activeAccountStatus$`, and `authStatuses$` call sites are unchanged (`authStatusFor$` is now a writable property of the same call signature).

### Reproduction and measurement (bundled CLI)

Built the OSS CLI (`apps/cli`, `npm run build:oss`), seeded local state with one active account (status `locked`), and drove `GET /status` over keep-alive. Heap snapshots taken with `node --heapsnapshot-signal=SIGUSR2` (V8 performs a full GC before writing, so counts are of retained/live objects).

| /status requests | RSS before | RSS after |
| --- | --- | --- |
| 2,000 | 212 MB | 255 MB |
| 10,000 | 812 MB | 616 MB |
| 20,000 | 998 MB | 651 MB |
| 40,000 | 1,154 MB | 654 MB |

Before: RSS climbs linearly with no plateau. The heap snapshot at 2,000 requests retains 2,011 `ReplaySubject` (about one per request), 2,012 `SafeSubscriber`, and 22,036 `OperatorSubscriber`, all surviving GC. The snapshot file grows from 85 MB (2k) to 809 MB (40k).

After: RSS plateaus (the 20k -> 40k delta is +3 MB, i.e. noise). The heap snapshot retains 13 `ReplaySubject` at 2k and 11 at 40k (a small constant, not per-request), and the total heap node count decreases over the 2k -> 40k interval. The snapshot file is 48 MB (2k) and 47 MB (40k).

### Tests

New regression coverage in `auth.service.spec.ts` (four of the five new tests fail on the pre-fix code; the fifth is behavior-preservation):
- same cached observable instance returned for repeated calls with the same `userId`
- exactly one shared upstream subscription across 50 reads (the pre-fix code produced 50)
- independent, single-subscription stream per distinct `userId` (no cache-key collision or cross-contamination)
- invalid `userId` (`null`/`undefined`) is short-circuited to `LoggedOut` and cached
- live lock/unlock/logout transitions still propagate through the cached stream

`npm run test:types`, eslint, and prettier all pass.

### Notes / out of scope

- The same `shareReplay({ refCount: false })`-inside-a-per-call-factory pattern also exists in `cipher-authorization.service.ts` (`canCloneCipher$`), but it currently has no callers and is not reachable from `bw serve`; it is better addressed separately. Other `refCount: false` usages reviewed are either assigned once per service instance or memoized per key (e.g. `account.service` `activeAccount$`, `key.service` `activeUserOrgKeys$`, `folder.service` `folderViews$`, `cipher.service` via `perUserCache$`) and do not leak per request.
- As with the existing `perUserCache$` usages, the per-user cache `Map` is not evicted on account removal; it is bounded by the number of distinct users seen and is not introduced by this change.

## 📸 Screenshots

N/A (no UI changes).


ggiesen:
**As a general comment on the copyright-assignment requirement here:**

I spent a fair amount of time diagnosing and isolating this leak (the analysis is in #20548), and this PR is the resulting fix. I've signed the CLA so it can be merged, and given how many people are hitting the same heap exhaustion in `bw serve`, I wanted it landed quickly and properly.

Even so, the agreement gave me real pause, and it's something I'll weigh carefully before contributing again. It isn't a licence grant. It's a copyright assignment, and several of its terms go well beyond what a contribution agreement needs:

- **It assigns ownership, not a licence.** You hand over all right, title, and interest in the contribution's copyright and waive your moral and database rights. You don't license the code; you stop owning it.
- **The patent grant is one-way, with no defensive termination.** The Apache ICLA carries a patent-retaliation clause: if anyone brings patent litigation claiming the contribution or the project infringes a patent, that party's patent licence terminates, which discourages patent aggression against the project. Bitwarden's grant is irrevocable with no equivalent. A broad, sublicensable patent licence flows to Bitwarden, backed by a covenant not to sue, and nothing flows back.
- **The non-infringement warranty is uncapped and unqualified.** You warrant that your contribution infringes no one's IP, with no knowledge qualifier, no liability cap, and no reciprocal indemnity. That's a hard thing to guarantee honestly for any non-trivial code.
- **It reaches backward.** It applies to contributions you "have made," not just future ones, retroactively sweeping in work submitted before the agreement existed or before you'd read it.
- **Bitwarden can transfer the whole agreement freely.** It can assign the agreement, and the copyright you handed over with it, "without notice or restriction." You get no equivalent right.

@kspearrin,

Your justification, that Bitwarden needs to own the copyright in order to relicense commercially, doesn't hold up. Relicensing only requires a broad inbound licence: perpetual, irrevocable, sublicensable, with the right to relicense under any terms. That's exactly what the Apache ICLA and the Project Harmony "licence" templates grant while leaving copyright with the author. You don't need to own code to relicense it; you need permission to. Full assignment is well beyond what the business case requires.

There is one legitimate reason a copyleft project might hold copyright, and it isn't relicensing: standing to enforce the licence against violators. It's why the FSF takes assignment for GNU code, and why MariaDB uses one too. But the projects that assign for that reason build in protections this CLA drops. MariaDB is GPLv2 and commercially dual-licensed, started by MySQL's original author after Oracle acquired it, so it had every reason to get this right. Its agreement takes only joint ownership, so you keep full rights to your own code; it binds the project to always release contributions under an FSF- or OSI-approved free licence; its warranty is knowledge-qualified; and anyone who can't sign can contribute under a plain BSD licence instead. The FSF, for its part, assigns to a non-profit chartered to keep software free. Bitwarden's CLA has none of this: exclusive ownership, no promise to keep contributions free, a flat uncapped warranty, no opt-out, and the right to transfer the lot, your copyright included, to an acquirer without notice.

Enforcement isn't even the reason on the table here, though; the stated reason is commercial relicensing, which a licence grant covers. On that point, scale cuts against you. Broadcom sells VMware commercially and relicenses aggressively, yet Salt, an Apache 2.0 project Broadcom now owns, runs on a licence-grant CLA where contributors keep their copyright outright. So both ends of the range are covered: a huge permissive project that takes no ownership, and a commercially dual-licensed copyleft one that takes only joint ownership and guarantees the code stays free. An exclusive, transferable assignment with no such guarantee is neither.

The transferability risk isn't hypothetical, and it traces to the same acquisition behind MariaDB. Oracle bought Sun, which owned both MySQL and Solaris, then took Solaris proprietary, and the community's only recourse was to fork the last open release as illumos. Copyleft means existing code can always be forked, so nobody can lock up what's already public, and AGPL even closes the hosted-service loophole, so a fork holds up better than most. But an exclusive, transferable assignment is what hands a future owner the keys to close future development and relicense the whole codebase commercially. Assignment hasn't died; it has split into the copyleft-defending kind that keeps ownership joint and guarantees the code stays free, and the proprietary-capture kind. This is the second kind.

A licence-grant CLA (Apache ICLA or a Harmony variant), or a simple DCO sign-off, would let contributions like this flow upstream without asking anyone to surrender ownership of their work. As it stands, the assignment requirement is the main thing that would make me think twice about contributing again, and going by the existing CLA threads, I doubt I'm the only one.

JaredSnider-Bitwarden:
Hello, @ggiesen , thank you for this contribution and thorough investigation! I will be picking this up to review at the start of next week as work prioritization allows.

libs/common/src/auth/services/auth.service.spec.ts
```diff
@@ -194,4 +194,103 @@ describe("AuthService", () => {
...
+      }
+
+      // perUserCache$ builds the stream once and shares it via
+      // shareReplay({ refCount: false }), so exactly one upstream subscription exists
+      // regardless of how many reads occur. The pre-fix method created a new shareReplay
+      // per call, which would make both of these equal 50 (the leak).
+      expect(totalSubscriptions).toBe(1);
+      expect(activeSubscriptions).toBe(1);
+    });
+
```
JaredSnider-Bitwarden, line 275: Thank you for adding the new test coverage! I confirmed it failed on main and passes on this branch. I've identified a few additional tests that I think would be worth adding below (confirmed passing in your branch). 

```suggestion
     // Mirrors the actual bw serve failure surface: GET /status -> StatusCommand ->
    // authService.getAuthStatus(userId) -> firstValueFrom(authStatusFor$(userId)). Pre-fix
    // code would produce 50 upstream subscriptions here (one per HTTP request); the fix
    // must hold when invoked through getAuthStatus(), not just authStatusFor$ directly.
    it("does not leak upstream subscriptions when invoked through getAuthStatus()", async () => {
      let totalSubscriptions = 0;
      const hasAccessToken$ = new Observable<boolean>((subscriber) => {
        totalSubscriptions++;
        subscriber.next(true);
      });
      tokenService.hasAccessToken$.mockReturnValue(hasAccessToken$);

      for (let i = 0; i < 50; i++) {
        await sut.getAuthStatus(userId);
      }

      expect(totalSubscriptions).toBe(1);
    });

    // activeAccountStatus$ is built once in the constructor but switchMap-rebinds to
    // authStatusFor$(userId) every time the active account changes. Pre-fix code
    // produced a brand-new shareReplay pipeline per (re)bind, so switching back to a
    // previously-seen user added another permanent upstream subscription. With
    // memoization, each distinct userId has exactly one cached stream regardless of how
    // many times the active user switches.
    it("does not multiply upstream subscriptions when activeAccount switches back and forth", async () => {
      const userId2 = Utils.newGuid() as UserId;
      const accountInfo1 = {
        status: AuthenticationStatus.Unlocked,
        id: userId,
        ...mockAccountInfoWith({ email: "email", name: "name" }),
      };
      const accountInfo2 = {
        status: AuthenticationStatus.Unlocked,
        id: userId2,
        ...mockAccountInfoWith({ email: "email2", name: "name2" }),
      };

      const subscriptionsPerUser: Record<string, number> = {};
      tokenService.hasAccessToken$.mockImplementation(
        (id: UserId) =>
          new Observable<boolean>((subscriber) => {
            subscriptionsPerUser[id] = (subscriptionsPerUser[id] ?? 0) + 1;
            subscriber.next(true);
          }),
      );

      // Subscribe once; switchMap inside activeAccountStatus$ rebinds per emission.
      trackEmissions(sut.activeAccountStatus$);

      for (let i = 0; i < 10; i++) {
        accountService.activeAccountSubject.next(accountInfo1);
        accountService.activeAccountSubject.next(accountInfo2);
      }

      expect(subscriptionsPerUser[userId]).toBe(1);
      expect(subscriptionsPerUser[userId2]).toBe(1);
    });
```

libs/common/src/auth/services/auth.service.spec.ts
```diff
@@ -194,4 +194,103 @@ describe("AuthService", () => {
...
+          activeSubscriptions--;
+        };
+      });
+      tokenService.hasAccessToken$.mockReturnValue(hasAccessToken$);
+      // getInMemoryUserKeyFor$ is of(userKey) (beforeEach), which completes immediately;
+      // within combineLatest only hasAccessToken$ stays open, so the counters below track
+      // that single long-lived upstream.
+
+      for (let i = 0; i < 50; i++) {
+        await firstValueFrom(sut.authStatusFor$(userId));
```
JaredSnider-Bitwarden: Could we enhance this assertion to be:

```suggestion
        expect(await firstValueFrom(sut.authStatusFor$(userId))).toEqual(
          AuthenticationStatus.Unlocked,
        );
```
The existing test only checks how many times upstream is subscribed — never what value the 50 reads return. Asserting the value inside the loop adds the missing half: it proves the cache actually replays Unlocked each time, not just that it doesn't re-subscribe. I confirmed this passes locally.

JaredSnider-Bitwarden (changes requested):
Thank you again for the thoroughly tested fix! I've confirmed your tests show the multiple subscriptions issue is resolved. I've requested two additional test cases + one assertion tweak in a test below 👇 ,but then we should be good to go!

ggiesen:
Thanks Jared. Addressed in f843a23:

- Added the `getAuthStatus()` subscription-count test and the activeAccount back-and-forth test (used your suggested code).
- Strengthened the repeated-reads loop to assert the replayed value (`Unlocked`) each iteration, not just the subscription count.

All 20 tests in the file pass locally; branch is up to date with `main`.

JaredSnider-Bitwarden (approved):
Looks great to me! Thank you again for the submission. I will run our CI workflows on your branch and then mark it ready for QA so our QA team can test it next week.

ggiesen:
## Empirical validation in production: 4 deployments, 66h post-patch, leak gone

Built the patch onto bw 2025.12.1 source (the change ports cleanly — `perUserCache$` exists at the same path in 2025.12.1) and deployed across the four `bw serve` workloads from #20548 on 2026-05-29 ~18:40Z. Same workloads, diverse call rates from probe-driven (~2 req/min) to 30s sync (~720 req/h).

### Before — pre-patch baseline on stock bw 2025.12.1, 8.8 days

Most active deployment, 16 V8-ceiling-hit cycles, ~13h cycle period each:


Screenshot 2026-05-29 115046

### After — patched binary, 66 hours

All four deployments overlaid. Memory bounces in a 105-140 MiB band; no cycles, no ramp:

Screenshot 2026-06-01 083846

### Per-deployment slope, 66h post-patch

| Deployment | Call pattern | Pre-patch | Post-patch |
|---|---|---|---|
| 30s sync, A | ~720 req/h | 16 V8 cycles in 211h, ~22 MB/h between cycles | **0 cycles, -0.06 MB/h** |
| 30s sync, B | ~480 req/h | 3 V8 cycles in 211h | **0 cycles, +0.13 MB/h** |
| ESO refresh | low-rate continuous | +6.4 MB/h monotonic to V8 ceiling | **+0.21 MB/h** |
| 600s sync | ~5 req/h | 1 V8 cycle in 127h | **0 cycles, +0.10 MB/h** |

All four slopes within ±0.21 MB/h — V8 GC oscillation, not leak growth. ESO-driven deployment (which previously climbed to the V8 ceiling in ~50h) is now flat at 125 MiB after 66h. None of the four pods has had a single container restart since deploy.

Confirms the per-request leak attributed in this PR's description was the entire observable leak in our workload — no second source.

JaredSnider-Bitwarden:
@ggiesen thank you again for this contribution! This has just passed QA testing across all clients and I will merge shortly!

PR #20919, opened 2026-05-29
[PM-38362] fix(powermonitor): Improve connection handling and remove unwraps

### LINKED ISSUE #20909: Linux Flatpak desktop app crashes entire KDE desktop session by leaking D-Bus connections
# Debug by Claude

## Summary

On Linux, the Bitwarden desktop app continuously opens new D-Bus connections to the user session bus at a steady rate (~1 every ~5 seconds) and **never closes them**. Each leaked connection consumes one file descriptor in the session message-bus daemon. Over time the bus daemon's open-fd count climbs without bound until it hits its `RLIMIT_NOFILE` ceiling, fails an `accept()`/`pidfd_open()` with `EMFILE` ("Too many open files"), and exits.

Because the session bus is a single point of failure for the whole desktop, when it dies **every** app on the session loses the bus simultaneously and the desktop session collapses. On my machine this has produced two full, unrecoverable session crashes requiring a hard reboot. The leak is slow and linear, so the time-to-crash depends only on the bus daemon's fd limit — roughly **one hour** on my system.

I was able to attribute the leak to Bitwarden directly (not to other apps and not to the flatpak proxy layer itself) — see **Attribution** below.

## Environment

| | |
|---|---|
| Bitwarden desktop | **2026.4.0** (flatpak `com.bitwarden.desktop`, origin flathub, commit `db57031d34b6…`) |
| Flatpak runtime | `org.freedesktop.Platform/x86_64/25.08` |
| OS | Fedora 44 (Workstation) |
| Kernel | 7.0.10-201.fc44 |
| Desktop | KDE Plasma 6.6.5, **Wayland** |
| Session bus | `dbus-broker` 37 |
| Flatpak / proxy | flatpak 1.17.7, xdg-dbus-proxy 0.1.7 |
| Hardware | Framework 13, Intel Meteor Lake / Arc Graphics |

SSH Agent feature was enabled. Bitwarden was set to start on login.

## Steps to reproduce

1. Install the Bitwarden desktop app on a Linux system whose session bus is `dbus-broker` (default on Fedora, Arch, and others).
2. Log in and unlock Bitwarden; leave it running normally.
3. Monitor the session bus daemon's open file-descriptor count over time:
   ```bash
   # PID of your user session dbus-broker
   watch -n 15 'for b in $(pgrep -u "$UID" -x dbus-broker); do \
     echo "broker $b: $(ls /proc/$b/fd | wc -l) fds \
     (limit $(awk "/Max open files/{print \$4}" /proc/$b/limits))"; done'
   ```
4. Observe the fd count rise monotonically and never fall back.

## Expected behavior

The app's D-Bus connection count should be bounded — connections should be reused or closed. The session bus daemon's fd count should stay roughly flat during steady-state operation.

## Actual behavior

The session bus daemon's fd count climbs linearly and without bound for as long as Bitwarden runs, eventually reaching the fd ceiling and crashing the bus (and with it, the entire desktop session).

### Measured leak (30-minute capture, sampled every ~15s)

| Elapsed | Session `dbus-broker` fds | Bitwarden's `xdg-dbus-proxy` fds |
|--------:|--------------------------:|---------------------------------:|
| 0 min   | 285 | 128 |
| ~6 min  | 369 | — |
| ~13 min | 449 | 244 |
| ~19 min | 527 | 302 |
| ~26 min | 603 | 360 |
| ~30 min | **677** | **476** |

- Rate: **~12.5 fds/min ≈ 1 new connection every ~5 seconds**, perfectly linear, never released.
- On this system the broker's soft `RLIMIT_NOFILE` was **1024**, so it would have crashed ~30 minutes after the last sample (≈1h after login). On a system with a higher fd limit it takes proportionally longer, but the unbounded climb is the same.

## Attribution (why this is the app, not other software)

The leak flows through the flatpak `xdg-dbus-proxy`, but the proxy only holds a connection because the app opened one. Three facts pin it to Bitwarden specifically:

1. **Lockstep growth.** The bus daemon and the `xdg-dbus-proxy` fd counts rise at the *same* rate (see table). Each leaked client connection appears as one fd on the proxy and one on the broker — the signature of an app opening connections it never closes.
2. **cgroup ownership.** The leaking `xdg-dbus-proxy` process is in cgroup
   `…/app-flatpak-com.bitwarden.desktop-<id>.scope` — i.e. it belongs to **Bitwarden's** sandbox, the same cgroup as `bitwarden-app`. It was the only `xdg-dbus-proxy` running, and Bitwarden was the only flatpak app running. Quitting Bitwarden releases the connections and the broker's fd count drops back to normal.
3. **The connections are live, not orphaned by the proxy.** This rules out an `xdg-dbus-proxy` bug. At ~328 leaked connections, the proxy held **~328 ESTABLISHED client-side sockets and 0 in `CLOSE-WAIT`** (`ss -x`). If the proxy were failing to tear down the bus side after the app disconnected, those client-side sockets would be in `CLOSE-WAIT`. They aren't — every connection is alive end-to-end, i.e. the app is actively holding all of them open. (`busctl --user --list` correspondingly attributes ~328 of the bus's connections to the proxy PID, versus 5–12 for each normal session service.)

## Suspected source (lead, not confirmed)

The ~5-second cadence looks like a polling timer that opens a fresh D-Bus connection per tick instead of reusing one. A likely candidate is the **power/lock monitor** in `desktop_core`. During the crashes, this module logged:

```
thread 'tokio-runtime-worker' panicked at core/src/powermonitor/linux.rs:49:50:
Error occurred in handler for 'powermonitor.isLockMonitorAvailable'
```

The SSH-agent feature is another periodic D-Bus/socket consumer worth auditing. These are leads for where to look, not a confirmed root cause — the maintainers can confirm by counting the app's own open D-Bus sockets (`ls /proc/<bitwarden-pid>/fd | grep socket | wc -l`) over time.

## Impact

- **Severity: high.** A password manager running idle in the background eventually crashes the user's entire desktop session (all apps, unsaved work lost) on any Linux setup using `dbus-broker`. The failure is silent until the cascade and is easy to misattribute to the OS/desktop rather than to Bitwarden.

## Is this flatpak-specific?

The flatpak proxy is **not** the culprit — see Attribution point 3 (zero `CLOSE-WAIT` sockets means the proxy is faithfully holding live connections the app opened, not leaking orphaned ones). The flatpak layer only **doubles the fd cost** (one fd each on the client and bus side per connection) and makes the leak easy to observe. The mechanism — the app opening D-Bus connections without closing them — is app-level and would accumulate fds in the session bus regardless of packaging; under a native build the connections would land directly on the bus daemon.

Definitive cross-check for maintainers: run a **non-flatpak build** (`.deb`/`.rpm`/AppImage) and watch the session bus daemon's fd count (repro command above). If it still climbs — which is expected — the leak is squarely in the app, independent of flatpak.

Note: I could not read the `bitwarden-app` process's own `/proc/<pid>/fd` from outside the sandbox (owned `root:root`), so a direct count of the app's sockets undercounts; the proxy's ESTABLISHED client-side count is the reliable signal. I also can't tell from outside whether the connections originate in Bitwarden's own code or the bundled Electron/Chromium layer — both ship as part of this client.

## Related issues

Not a duplicate of any open issue, but possibly related:
- #19054, #17932 — resource leaks over time, but reported as **memory**; this is the file-descriptor / D-Bus variant of the same "leaks while running" theme.
- #19883 ("Desktop crashes on lock") — may share the `desktop_core` power/lock code path suspected above.
- #13099, #16288 — SSH-agent socket handling (the SSH agent is one of the periodic D-Bus consumers worth auditing as the leak source).

## Workaround

Quit Bitwarden (or disable "start on login") to halt and reset the leak.

ChihweiLHBird:
**Potentially** fix the crashing issue of KDE caused by leaking D-Bus connections by Bitwarden #20909

Improve connection handling and remove unwraps of `is_lock_monitor_available`. It can now timeout when the API is unresponsive (speculated: KDE doesn't respond GNOME API).


quexten:
@ChihweiLHBird Have you verified that this fixes the issue for your machine?

ChihweiLHBird:
@quexten Sorry, not yet... But I plan to test it maybe later this week.

quexten:
Moved to draft until then. Thank you!

ChihweiLHBird:
Hi @quexten @djsmith85, I tested this fix on my machine and it works. Since it's crashing my desktop environment every hour, may I get this in quicker for the next release? Thanks!

apps/desktop/desktop_native/core/src/powermonitor/linux.rs
```diff
@@ -43,10 +43,18 @@ pub async fn on_lock(tx: tokio::sync::mpsc::Sender<()>) -> Result<(), Box<dyn st
     Ok(())
 }
 
-// FIXME: Remove unwraps! They panic and terminate the whole application.
-#[allow(clippy::unwrap_used)]
+async fn lock_monitor_connection() -> zbus::Result<Connection> {
+    Builder::session()?
+        .method_timeout(Duration::from_secs(25))
```
quexten: I don't know why the duration is so high here, one second should be enough, I think?

quexten (dismissed):
Still needs @bitwarden/team-platform-dev  approval but from my perspective the changes look reasonable.

apps/desktop/desktop_native/core/src/powermonitor/linux.rs
ChihweiLHBird: Yes, lower values are fine too. I increased it  when worrying about some systems may be slow to respond to this in extreme cases. What's the ideal value in you mind? How about 2 seconds (before my increase)?

quexten: Seems reasonable.

quexten (approved):

Fak3:
Should there be guards against it in the desktop-portal?

ChihweiLHBird:
> Should there be guards against it in the desktop-portal?

Sorry, I am not familiar with it. What's that?

addisonbeck (approved):

quexten:
@Fak3 This uses raw dbus interfaces. There is no portal for lock monitoring.

Fak3:
> @Fak3 This uses raw dbus interfaces. There is no portal for lock monitoring.

It goes through xdg-dbus-proxy. Perhaps it can be limited there somehow? Via cgroups?

ChihweiLHBird:
I think `xdg-dbus-proxy` is in another layer, right? And even if it catch the seemingly leaking sessions, it's hard for it to tell whether it's legitimate heavy usage or leaking. I think the root cause fix should still happen in the application code.

PR #20943, opened 2026-06-01
Add bench instrumentation to autofill content collection service

blackwood:
## 🎟️ Tracking

[PM-35196](https://bitwarden.atlassian.net/browse/PM-35196)

## 📔 Objective

[PM-35196]: https://bitwarden.atlassian.net/browse/PM-35196?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/browser/src/autofill/services/collect-autofill-content.service.ts
```diff
@@ -100,6 +117,31 @@ export class CollectAutofillContentService implements CollectAutofillContentServ
       inputQuery += `:not([type="${type}"])`;
     }
     this.formFieldQueryString = `${inputQuery}, textarea:not([data-bwignore]), select:not([data-bwignore]), span[data-bwautofill]`;
+
+    this.handleMutationObserverMutation = stopwatch(
+      "handleMutationObserverMutation",
+      this.handleMutationObserverMutation,
+    );
+    this.processMutations = stopwatch("processMutations", this.processMutations);
```
audreyality: ⚠️ `stopwatch` should only be used with synchronous functions. `processMutations` has a synchronous prototype, but its work happens within `requestIdleCallbackPolyfill`. You're only measuring the setup time.

📝 This is also true of `updateAutofillElementsAfterMutation`, which called by a few of your stopwatches. This method performs exponential backoff of another `requestIdCallbackPolyfill`.

apps/browser/src/autofill/services/collect-autofill-content.service.ts
```diff
@@ -100,6 +117,31 @@ export class CollectAutofillContentService implements CollectAutofillContentServ
...
+    this.handleNewShadowRoots = stopwatch("handleNewShadowRoots", this.handleNewShadowRoots);
+    this.setupMutationObserver = stopwatch("setupMutationObserver", this.setupMutationObserver);
```
audreyality: ⚠️ `stopwatch` and `measure` are intended for use where you're trying to capture the runtime of a large function. `setupMutationObserver` is performing very little work, and so you're paying the cost of running the stopwatch that won't track meaningful timing data. If you're trying to measure the number of times this method is invoked, consider using a `meter` instead.

apps/browser/src/autofill/services/collect-autofill-content.service.ts
```diff
@@ -100,6 +117,31 @@ export class CollectAutofillContentService implements CollectAutofillContentServ
...
+    this.handleWindowLocationMutation = stopwatch(
+      "handleWindowLocationMutation",
+      this.handleWindowLocationMutation,
+    );
```
audreyality: ❓ `handleWindowLocationMutation` appears to be a very sparse method, is the timing information useful? 

🎨 Since the value passed to `recordRequireUpdate` is a fixed string, you could instead create one meter for each with no arguments and measure the invocation count for `handleWindowLocationMutation` by counting its specific marks. You can then add together each of the three meters to get your combined `recordRequireUpdate` metric.

apps/browser/src/autofill/services/collect-autofill-content.service.ts
```diff
@@ -114,6 +156,21 @@ export class CollectAutofillContentService implements CollectAutofillContentServ
    * @public
    */
   async getPageDetails(): Promise<AutofillPageDetails> {
+    // Split shell wraps all five return paths in getPageDetailsInner without a try/finally dance.
+    this.recordGetPageDetails("start", -1);
+    const pageDetails = await this.getPageDetailsInner();
+    this.recordGetPageDetails("end", pageDetails.fields.length);
```
audreyality: ⚠️ It appears that here you're using `meter` to construct a span manually. As a result, you have to put in a marker for an invalid meter and hard-code two meter names. This information could instead be gathered by adding a stopwatch over `getPageDetails` and a single meter for `pageDetails.fields.length`.

⛏️ Don't introduce a separate `getPageDetailsInner` method. Your metrics should be captured inline with the logic so that all invocations are tracked. If measurement is impacting the runtime speed or legibility of the method, you are likely measuring too much at once. Instead, determine which measurements are key indicators--which indicate whether you've meaningfully addressed page performance--and focus on those.

apps/browser/src/autofill/services/collect-autofill-content.service.ts
```diff
@@ -569,6 +626,7 @@ export class CollectAutofillContentService implements CollectAutofillContentServ
   private async buildAutofillFieldsData(
     formFieldElements: FormFieldElement[],
   ): Promise<AutofillField[]> {
+    this.recordBuildFields("start", formFieldElements.length);
```
audreyality: ⚠️ Here's another case where you're using meters to track a span. In this case, it appears it is to circumvent the synchronous limitation of measures. Async is not presently handled because tracking it presents a number of reliability issues for the gathered statistics. Don't measure it.

apps/browser/src/autofill/services/collect-autofill-content.service.ts
```diff
@@ -1321,24 +1380,31 @@ export class CollectAutofillContentService implements CollectAutofillContentServ
...
-        this.handleNewShadowRoots();
-        this.pendingShadowDomCheck = false;
-      }, this.shadowDomCheckTimeoutMs);
-    }
+        this.shadowDomCheckTimeout = setTimeout(() => {
+          this.handleNewShadowRoots();
+          this.pendingShadowDomCheck = false;
+        }, this.shadowDomCheckTimeoutMs);
+      }
+    });
```
audreyality: ⚠️ This is measuring how long it takes to call `clearTimeout` and `setTimeout`.

audreyality (commented):
💭 Something to note is that these performance APIs are tuned to be as lightweight as possible. Their ideal usage is somewhat counter to your intuition.

* Only use `stopwatch` and `measure` where you need to measure the processing time of synchronous code.
* If code is asynchronous, including callback-based asynchrony, it cannot be measured suing `stopwatch` and `measure`.
* Only pass variables into meters. Fixed inputs can be inferred by the number of times the related `record` function is called. Prefer multiple `record` methods to a fixed-variable input.
* Rely on post-processing to compound meters.

apps/browser/src/autofill/services/collect-autofill-content.service.ts
```diff
@@ -71,6 +72,22 @@ export class CollectAutofillContentService implements CollectAutofillContentServ
...
+    "fieldCount",
+  );
+  private readonly recordMapSizes = createMeter("mapSizes", "fields", "byOpid", "forms", "heap");
+  private readonly recordRequireUpdate = createMeter("requireUpdate", "trigger");
+  // Larger ring than default: fires per MO callback; churn-heavy pages can outrun 256 slots before flush.
+  private readonly recordMutationBatch = createMeter(
+    { name: "mutationBatch", bits: 10 },
+    "records",
+    "inShadow",
+  );
```
audreyality: ⛏️ I'd not use the word "record" here, since it's misleading. It brings up the idea of activity monitoring or playback, neither of which are what we're doing. "monitor", "observe", or "measure" are more accurate.

apps/browser/src/autofill/services/collect-autofill-content.service.ts
audreyality: One thing I'm wondering is if you still want to capture `formFieldElements.length` and `autofillFields.length`? If those are meaningful metrics, you could still capture them. If not, then this comment is resolved.

apps/browser/src/autofill/services/collect-autofill-content.service.ts
audreyality: Is `pageDetails.fields.length` a meaningful metric? If so, you might want to capture that. 

My comments re: the artifical span are resolved.

apps/browser/src/autofill/services/collect-autofill-content.service.ts
blackwood: They're a little superfluous to this particular case. Leaving them out!

apps/browser/src/autofill/services/collect-autofill-content.service.ts
blackwood: Omitted since `monitorMapSizes` should already contain `fields` during sample time

audreyality (dismissed):

audreyality (approved):

PR #20944, opened 2026-06-01
Fix: SSH Agent v1 no agent forwarding warning

neuronull:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-38133

## 📔 Objective

The v1 agent seems to have introduced a regression at some point in the v1 timeline. The request for authorization UI prompt is supposed to give a warning when agent forwarding is being utilized. This is not happening.

The root cause is that the is_forwarding flag is not being toggled in the request sent to Electron services. We detect it correctly in session bind decoding, but it does not persist across the next data buffer decoded due to the `set_forwarding` function's logic, which is always keeping the latest value.

## 📸 Manual testing

A/B test of agent forwarding confirmed the warning appears. See ticket for recordings.


quexten (approved):

quexten:
Thought: [not blocking for this PR] we should investigate the security of ssh agent forwarding. Can the remote machine do anything (craft any packets) to make the client believe it is not forwarded (anymore) in order to bypass approval prompts?

Issue #19054, opened 2026-02-18
Memory leak on windows desktop version


tony8077616:
### Steps To Reproduce

Let Application in the background just around more than two days.

### Expected Result

Memory leak more than normal (bigger than 500 MB)

### Actual Result

I just leave my laptop around two days with screen lock and memory leak when I login again

### Screenshots or Videos

<img width="1920" height="380" alt="Image" src="https://github.com/user-attachments/assets/65a64591-46c5-40fb-8897-7e60da0767b2" />

### Additional Context

No only this version have memory leak, sometime old version also have same problem.

### Operating System

Windows

### Operating System Version

Windows 11 26H1 (28020.1611)

### Installation method

Direct Download (from bitwarden.com)

### Build Version

2026.1.0

### Issue Tracking Info

- [x] I understand that work is tracked outside of GitHub. A PR will be linked to this issue should one be opened to address it, but Bitwarden doesn't use fields like "assigned", "milestone", or "project" to track progress.

kotovasia5120:
Similar problem.
<img width="483" height="237" alt="Image" src="https://github.com/user-attachments/assets/a446e70e-a5c4-4f38-aaf8-60ef6b821935" />

Gallium70:
This issue still exists with Bitwarden 2026.2.1 on Windows 11 25H2. I have to restart Bitwarden from time to time to release the memory.

Edit: A lock and unlock release the memory too.

<img width="1035" height="80" alt="Image" src="https://github.com/user-attachments/assets/b94656e7-2656-4eb9-8af9-88dbd3175049" />

brontide:
Same on linux/flatpack. Regularly consumes multiple GB of RAM and causes OOM killer.

tony8077616:
Hasn't anyone from the contributors or developers come forward to provide an explanation?

idan-h:
I have the same bug

brontide:
What do I win?

<img width="661" height="37" alt="Image" src="https://github.com/user-attachments/assets/2c8a2077-f1ec-4f65-b737-df461b3b9045" />

AwaisKing:
maybe instead of using node.js like sheep, they should've consider good old C or C++
p.s. that's 5 minutes after restarting Bitwarden

<img width="100%" alt="Image" src="https://github.com/user-attachments/assets/c98ed77f-618e-45eb-a6e7-2c476189ae7c" />

idan-h:
[...]

PR #20945, opened 2026-06-01
[PM-38402] fix(desktop): fix IPC listeners that were never getting removed

fabiand93:
## 🎟️ Tracking

Relates to #19054

> I noticed the discussion in that issue spent more time debating what language the app should be rewritten in than actually fixing the leak. With all due respect to everyone involved, here is a practical fix in the language it's already written in.

## 📔 Objective

The desktop renderer process grows to 400–500 MB+ after several days of uptime. After digging into the codebase I found three concrete bugs that cause IPC event listeners to accumulate in memory and never get cleaned up.

---

### Bug 1 — `onMessage.removeListener` was silently doing nothing

In `apps/desktop/src/platform/preload.ts`, the `removeListener` implementation was creating a brand new arrow function every time it was called:

```ts
// Before: creates a NEW function on every call
// ipcRenderer.removeListener never finds a matching reference
removeListener: (callback) => {
  ipcRenderer.removeListener("messagingService", (_event, message: any) => {
    if (message.command) { callback(message); }
  });
},
```

`ipcRenderer.removeListener` requires the **exact same function reference** that was passed to `addListener`. Since both methods were creating separate arrow functions, the call to `removeListener` always silently failed. Every `addListener` call registered a permanent listener that was impossible to clean up.

The listener registration now returns a cleanup function created in preload. That function closes over the exact Electron wrapper and removes it directly. `fromIpcMessaging` uses the returned function as its RxJS teardown, so cleanup does not depend on callback identity across `contextBridge`.

---

### Bug 2 — `onSystemThemeUpdated` had no cleanup at all

Same file — `onSystemThemeUpdated` registered an `ipcRenderer.on` listener but had no way to remove it. `fromIpcSystemTheme` was using `fromEventPattern` without a `removeHandler`, so the listener leaked on every unsubscribe.

The fix changes `onSystemThemeUpdated` to return a cleanup function. `fromIpcSystemTheme` was rewritten to use a plain `Observable` that calls that cleanup function as its teardown logic.

---

### Bug 3 — `openModal` subscription was explicitly bypassing the cleanup lint rule

In `app.component.ts`, the `openModal` method had a `// eslint-disable-next-line rxjs-angular/prefer-takeuntil` comment to skip the rule that exists exactly to prevent this kind of leak. Replaced with `.pipe(takeUntil(this.destroy$))`.

---

## 📸 Screenshots

N/A — no UI changes.


fabiand93:
I also opened a second PR (#20946 — https://github.com/bitwarden/clients/pull/20946) that includes everything from this one plus a migration of `AppComponent` to `ChangeDetectionStrategy.OnPush`.

The reason I split them is that the IPC listener fixes in this PR are low-risk, self-contained corrections with basically no chance of introducing regressions. The OnPush migration in the other one is also correct, but it touches change detection behavior at the root component level, which is a bigger surface area to review and test.

The idea was to give the team the option to merge the safe fix quickly and evaluate the OnPush change separately, instead of having to block everything on the more complex one.

apps/desktop/src/platform/preload.ts
```diff
@@ -151,22 +153,27 @@ export default {
 
   sendMessage: (message: { command: string } & any) =>
     ipcRenderer.send("messagingService", message),
-  onMessage: {
-    addListener: (callback: (message: { command: string } & any) => void) => {
-      ipcRenderer.addListener("messagingService", (_event, message: any) => {
-        if (message.command) {
-          callback(message);
+  onMessage: (() => {
+    const listenerMap = new Map<Function, (...args: any[]) => void>();
```
coroiu, line 157: issue: `new Map<Function, ...>()` uses the bare Function type, which violates @typescript-eslint/no-unsafe-function-type rule and blocks CI. Replace `Function` with the concrete callback signature already used by `addListener`/`removeListener`.

coroiu (commented):
Thank you for your contribution, just one small issue

apps/desktop/src/platform/preload.ts
fabiand93: Thanks for catching this. I removed the `Map<Function, ...>` entirely rather than only changing the key type.

While testing the listener lifecycle through Electron's real `contextBridge`, I found that the same renderer callback is exposed as a different proxy on separate `addListener` and `removeListener` calls. That means a callback-keyed map can still miss the original listener even with a concrete type.

`addListener` now returns a typed cleanup function created in preload. It closes over the exact wrapper registered with `ipcRenderer`, and `fromIpcMessaging` calls that function when the observable is torn down. This removes the unsafe `Function` type and avoids relying on callback identity across the bridge.

I added focused tests for registration, delivery, filtering, independent listeners, and teardown, and also verified the cleanup through a real Electron `contextBridge` boundary.

If you spot anything else that could be improved, please let me know. I'm happy to take another look. 🙂

apps/desktop/src/platform/preload.ts
```diff
@@ -149,31 +153,27 @@ export default {
     ipcRenderer.invoke("ipc.log", { level, message, optionalParams }),
 
   getSystemTheme: (): Promise<ThemeType> => ipcRenderer.invoke("systemTheme"),
-  onSystemThemeUpdated: (callback: (theme: ThemeType) => void) => {
-    ipcRenderer.on("systemThemeUpdated", (_event, theme: ThemeType) => callback(theme));
+  onSystemThemeUpdated: (callback: (theme: ThemeType) => void): (() => void) => {
+    const wrapper = (_event: IpcRendererEvent, theme: ThemeType) => callback(theme);
+    ipcRenderer.on("systemThemeUpdated", wrapper);
+    return () => ipcRenderer.removeListener("systemThemeUpdated", wrapper);
```
coroiu, line 160: suggestion(non-blocking): honestly, I think we could've just returned the observable directly, instead of having to use a wrapper, that way this can't be called directly and not cleaned up, but I'll leave this as a non blocking comment because the pattern is pre-existing

coroiu (dismissed):
I'll approve this and mark it as ready for QA. Thank you for your patience! In the meanwhile, please take a look at the breaking change! Also, please don't force-push, it breaks GitHub's "changed since last reviewed" UI :)

coroiu:
@fabiand93 I've removed the dialog fix since that function has been removed on main

coroiu (approved):

dani-garcia (approved):

PR #20946, opened 2026-06-01
[PM-38404] fix(desktop): fix memory leak - IPC listeners + OnPush change detection

fabiand93:
## 🎟️ Tracking

Relates to #19054

> Same as the other PR I opened — the issue had a long debate about what language this should be rewritten in but the actual causes of the leak are fixable right now without rewriting anything. This PR goes a bit deeper than the other one.

## 📔 Objective

This PR includes the same three IPC listener leak fixes from my other PR ([fix/ipc-listener-leaks](https://github.com/bitwarden/clients/pull/20945)), plus a migration of `AppComponent` to `ChangeDetectionStrategy.OnPush` that was already tracked as a FIXME in the code.

---

### Bugs 1–3 — IPC listener leaks (same as the other PR)

Quick summary:

- `onMessage.removeListener` was creating a new arrow function instead of reusing the original wrapper, so `ipcRenderer.removeListener` never found a match and listeners accumulated forever
- `onSystemThemeUpdated` registered an `ipcRenderer.on` listener with no way to remove it, and `fromIpcSystemTheme` was using `fromEventPattern` without a `removeHandler`
- `openModal` had an explicit eslint-disable to skip `takeUntil`, leaving the subscription uncleaned

See the other PR for the full explanation of each one.

---

### Bug 4 — `AppComponent` was running change detection on every single async event

There was already a `// FIXME(CL-764): Migrate to OnPush` comment in the code. This is the most impactful change for the memory growth over time.

With `ChangeDetectionStrategy.Default`, every IPC message, every SignalR heartbeat and every timer callback triggers a full change detection cycle across the whole component tree. Over several days running in the background this creates a constant stream of short-lived object allocations and deallocations. Even when the GC collects those objects, V8 does not aggressively release pages back to the OS — the heap keeps growing to accommodate the peak usage, which is exactly the behavior people are reporting in the issue.

Migrating to `OnPush` means Angular only runs change detection when:
- An `async` pipe observable emits a new value
- `markForCheck()` is called explicitly
- An event originates from the component or its children

The only place where the template reads mutable state directly is the `loading` boolean, which is set in two cases inside the broadcaster callback. Added `this.cdr.markForCheck()` right after each assignment so the template still updates correctly when those transitions happen.

---

## 📸 Screenshots

N/A — no UI changes.

### DIFF
--- a/apps/desktop/src/app/app.component.ts
+++ b/apps/desktop/src/app/app.component.ts
@@ -1,6 +1,8 @@
 // FIXME: Update this file to be type safe and remove this and next line
 // @ts-strict-ignore
 import {
+  ChangeDetectionStrategy,
+  ChangeDetectorRef,
   Component,
   DestroyRef,
   NgZone,
@@ -85,10 +87,9 @@ const BroadcasterSubscriptionId = "AppComponent";
 const IdleTimeout = 60000 * 10; // 10 minutes
 const SyncInterval = 6 * 60 * 60 * 1000; // 6 hours
 
-// FIXME(https://bitwarden.atlassian.net/browse/CL-764): Migrate to OnPush
-// eslint-disable-next-line @angular-eslint/prefer-on-push-component-change-detection
 @Component({
   selector: "app-root",
+  changeDetection: ChangeDetectionStrategy.OnPush,
   styles: [],
   template: `
     <ng-template #settings></ng-template>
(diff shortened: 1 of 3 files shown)


fabiand93:
This PR is a superset of #20945 (https://github.com/bitwarden/clients/pull/20945), which contains only the three IPC listener fixes without the OnPush change.

I split them because the IPC fixes are small and low-risk — they correct obvious bugs where listeners were never being removed. The OnPush migration is also a valid fix and the code already had a FIXME asking for it, but it changes how change detection works at the root component level and may need more careful review and QA.

If merging both at the same time feels like too much, #20945 is the safer option to start with. This one can wait or be reviewed separately.

addisonbeck:
Since we're moving forward with https://github.com/bitwarden/clients/pull/20945 would you mind slimming this PR to the app component changes? I think those are okay too, but we'll want to track the changes independently.

djsmith85 (changes requested):
Marking as Request changes: Please see feedback given with https://github.com/bitwarden/clients/pull/20946#issuecomment-4858726176

PR #20952, opened 2026-06-01
added secret value generator

maxkpower:
## 🎟️ Tracking
[https://bitwarden.atlassian.net/browse/SM-1955](https://bitwarden.atlassian.net/browse/SM-1955)

## 📔 Objective
Adds an inline value generator to the Secrets Manager secret dialog, so users can generate a strong secret value directly when creating or editing a secret instead of copying one from the Password Manager generator.

## 📸 Screenshots
https://github.com/user-attachments/assets/9668e8c4-c0e2-44b5-b4c3-621410e2518c

### DIFF
--- a/libs/tools/generator/core/src/metadata/password/random-password.ts
+++ b/libs/tools/generator/core/src/metadata/password/random-password.ts
@@ -9,6 +9,85 @@ import { GeneratorDependencyProvider } from "../../providers";
 import { CredentialGenerator, PasswordGeneratorSettings } from "../../types";
 import { Algorithm, Profile, Type } from "../data";
 import { GeneratorMetadata } from "../generator-metadata";
+import { CoreProfileMetadata } from "../profile-metadata";
+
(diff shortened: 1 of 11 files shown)


libs/tools/generator/core/src/metadata/password/random-password.ts
harr1424, line 1: Tools Team is avoiding adding any new business logic to the `clients` repository, favoring instead the SDK. @itsadrago or myself would be happy to discuss this PR via Slack to determine if this is a high priority or deadlined feature, and if we would make an exception. Currently we don't have generator logic ready to consume in the SDK, so I don't think this is a hard no, but would like to discuss it as a team. 

I would also like to get the opinion of our generator SME @audreyality on the changes proposed to this file.

harr1424 (approved):
Looks good! Thanks for shifting the generation logic to use the SDK.

PR #20954, opened 2026-06-01
[PM-37806] Add batch action bar to individual vault

nick-livefront:
## 🎟️ Tracking

[PM-37806](https://bitwarden.atlassian.net/browse/PM-37806)
~~Stacked on top of: https://github.com/bitwarden/clients/pull/20885~~

## 📔 Objective

Adds a batch action bar to the individual (password manager) vault, enabling users to select multiple vault items and perform bulk operations including delete, move to folder, restore, archive, and assign to collections.

The batch bar is gated behind feature flags (`PM37785_VaultBatchBar` / `PM37785_DesktopVaultBatchBar`) and uses injection tokens for dialog implementations to keep the shared vault library platform-agnostic.

## 📸 Screenshots

<video src="https://github.com/user-attachments/assets/c7e8dbd2-fa4b-40e5-971d-e8023371e055" />

[PM-37806]: https://bitwarden.atlassian.net/browse/PM-37806?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/vault/src/components/vault-batch-bar/vault-batch-action.component.ts
+++ b/libs/vault/src/components/vault-batch-bar/vault-batch-action.component.ts
@@ -14,6 +14,7 @@ type ActionDescriptor = {
   action: () => void;
   icon: BitwardenIcon;
   label: string;
+  color?: string;
 };
 
 /** When the total number of available actions exceeds this, split into primary + overflow. */
@@ -75,7 +76,7 @@ export class VaultBatchActionComponent {
       actions.push({
         action: this.service.bulkDelete.bind(this.service),
         icon: "bwi-trash",
-        label: this.i18nService.t("delete"),
+        label: this.i18nService.t(this.service.inTrash() ? "permanentlyDelete" : "delete"),
       });
     }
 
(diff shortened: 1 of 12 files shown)


libs/vault/src/components/vault-batch-bar/vault-batch-action.component.ts
```diff
@@ -0,0 +1,96 @@
...
+        icon: "bwi-undo",
+        label: this.i18nService.t("restore"),
+      });
+    }
+    if (this.service.canDelete()) {
+      actions.push({
+        action: this.service.bulkDelete.bind(this.service),
+        icon: "bwi-trash",
+        label: this.i18nService.t("delete"),
+      });
```
jaasen-livefront, line 80: ⛏️ We used to check if we were in the trash and used the `permanentlyDeleteSelected` key for the bulk delete menu item. Not a big deal since we're still serving the delete dialog with `permanent: true` but just thought I'd point it out in case QA notices the difference.

nick-livefront: Fixed! [79d2aac](https://github.com/bitwarden/clients/pull/20954/commits/79d2aacffdcec80449bd28650ff7788c022c92ef)

nick-livefront: Designs had "Permanently Delete" rather than "Permanently Delete Selected"

jaasen-livefront (approved):

PR #20871, opened 2026-05-26
[PM-35196] Mutation memory management

blackwood:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-35196

## 📔 Objective

Address performance regression affecting Chromium browsers.

PR #20956, opened 2026-06-01
Gate childList rebuild on autofill-relevant mutations

blackwood:
## 🎟️ Tracking

Follow-on to [PM-35196](https://bitwarden.atlassian.net/browse/PM-35196)

## 📔 Objective

Extracted from [shadow-dom-unified-fix-mem-mgmt](https://github.com/bitwarden/clients/pull/20871) so the mutation-filtering optimization can be measured and reviewed independently of the observer lifecycle work it was originally bundled with.

[PM-35196]: https://bitwarden.atlassian.net/browse/PM-35196?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/browser/src/autofill/services/collect-autofill-content.service.ts
+++ b/apps/browser/src/autofill/services/collect-autofill-content.service.ts
@@ -1355,8 +1355,8 @@ export class CollectAutofillContentService implements CollectAutofillContentServ
 
     for (const mutation of mutations) {
       if (mutation.type === "attributes") {
-        // nodeType === 1 instead of `instanceof Element` — works across realms (adopted-from-iframe).
-        if (mutation.target.nodeType !== 1) {
+        // nodeType === 1 (Node.ELEMENT_NODE) instead of `instanceof Element`; faster & includes iframes
+        if (mutation.target.nodeType !== Node.ELEMENT_NODE) {
           continue;
         }
         const attributeName = mutation.attributeName?.toLowerCase();
@@ -1374,9 +1374,12 @@ export class CollectAutofillContentService implements CollectAutofillContentServ
           this.pendingTopLayerTargets.add(target);
         }
       } else if (mutation.type === "childList") {
-        this.pendingChildListUpdate = true;
+        // Without this gate, cosmetic mutations invalidate the noFieldsFound cache.
+        if (this.mutationAffectsAutofillTarget(mutation)) {
+          this.pendingChildListUpdate = true;
+        }
         for (const node of mutation.addedNodes ?? []) {
-          if (node.nodeType !== 1) {
+          if (node.nodeType !== Node.ELEMENT_NODE) {
             continue;
           }
           const element = node as Element;
@@ -1555,6 +1558,32 @@ export class CollectAutofillContentService implements CollectAutofillContentServ
         }
       }
     }
+  }
+
+  private mutationAffectsAutofillTarget(mutation: MutationRecord): boolean {
+    return (
+      this.nodeListIsAutofillRelevant(mutation.addedNodes) ||
+      this.nodeListIsAutofillRelevant(mutation.removedNodes)
+    );
+  }
+
+  private nodeListIsAutofillRelevant(nodes: NodeList | undefined): boolean {
+    if (!nodes) {
+      return false;
+    }
+    for (const node of nodes) {
+      if (node.nodeType !== Node.ELEMENT_NODE) {
+        continue;
+      }
+      const el = node as Element;
+      if (
+        el.matches(this.formFieldQueryString) ||
+        el.querySelector(this.formFieldQueryString) !== null
+      ) {
+        return true;
+      }
+    }
+    return false;
   }
 
   private isShadowRootCandidate(node: Node): node is Element {
--- a/apps/browser/src/autofill/services/collect-autofill-content.service.spec.ts
+++ b/apps/browser/src/autofill/services/collect-autofill-content.service.spec.ts
@@ -3474,4 +3474,291 @@ describe("CollectAutofillContentService", () => {
       expect(Array.from(pending.get(target)!).sort()).toEqual(["id", "value"]);
     });
   });
+
+  describe("childList gate is wired into mutation handling", () => {
+    const childListMutationFor = (...added: Node[]): MutationRecord => {
+      const container = document.createElement("div");
+      document.body.appendChild(container);
+      added.forEach((n) => container.appendChild(n));
+      return {
+        type: "childList",
+        addedNodes: container.childNodes,
+        removedNodes: document.querySelectorAll("nothing-matches-this-selector"),
+        attributeName: null,
+        attributeNamespace: null,
+        nextSibling: null,
+        oldValue: null,
+        previousSibling: null,
+        target: document.body,
+      };
+    };
+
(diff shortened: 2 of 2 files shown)


apps/browser/src/autofill/services/collect-autofill-content.service.ts
```diff
@@ -104,6 +117,11 @@ export class CollectAutofillContentService implements CollectAutofillContentServ
       inputQuery += `:not([type="${type}"])`;
     }
     this.formFieldQueryString = `${inputQuery}, textarea:not([data-bwignore]), select:not([data-bwignore]), span[data-bwautofill]`;
+
+    const inputExclusions = Array.from(this.autofillIrrelevantInputTypes)
+      .map((t) => `:not([type=${t}])`)
+      .join("");
```
jprusik: https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Selectors/:not#description

> You can negate several selectors at the same time. Example: :not(.foo, .bar) is equivalent to :not(.foo):not(.bar).

apps/browser/src/autofill/services/collect-autofill-content.service.ts
```diff
@@ -1460,6 +1481,42 @@ export class CollectAutofillContentService implements CollectAutofillContentServ
...
   }
 
+  private mutationAffectsAutofillTarget(mutation: MutationRecord): boolean {
+    return (
+      this.nodeListContainsAutofillRelevant(mutation.addedNodes) ||
+      this.nodeListContainsAutofillRelevant(mutation.removedNodes)
+    );
+  }
+
+  private nodeListContainsAutofillRelevant(nodes: NodeList | undefined): boolean {
```
jprusik:
```suggestion
  private nodeListIsAutofillRelevant(nodes: NodeList | undefined): boolean {
```

Or: `nodeListIsRelevantToAutofill` 🤷

apps/browser/src/autofill/services/collect-autofill-content.service.ts
```diff
@@ -1460,6 +1481,42 @@ export class CollectAutofillContentService implements CollectAutofillContentServ
...
+      if (node.nodeType !== 1) {
+        continue;
+      }
+      const el = node as Element;
+      if (this.autofillRelevantTags.has(el.tagName)) {
+        return true;
+      }
+      if (el.tagName === "INPUT") {
+        // .type property resolves to "text" for type-less inputs; getAttribute("type") doesn't.
+        const type = (el as HTMLInputElement).type;
```
jprusik: nit

```suggestion
        const inputType = (el as HTMLInputElement).type;
```

apps/browser/src/autofill/services/collect-autofill-content.service.ts
```diff
@@ -1460,6 +1481,42 @@ export class CollectAutofillContentService implements CollectAutofillContentServ
...
+        if (!this.autofillIrrelevantInputTypes.has(type)) {
+          return true;
+        }
```
jprusik: ⛏️ Slightly more verbose, but easier to parse, imo:

```suggestion
        if (this.autofillIrrelevantInputTypes.has(type)) {
          return false;
        }
        return true;
```

apps/browser/src/autofill/services/collect-autofill-content.service.ts
```diff
@@ -104,6 +117,11 @@ export class CollectAutofillContentService implements CollectAutofillContentServ
       inputQuery += `:not([type="${type}"])`;
     }
     this.formFieldQueryString = `${inputQuery}, textarea:not([data-bwignore]), select:not([data-bwignore]), span[data-bwautofill]`;
+
+    const inputExclusions = Array.from(this.autofillIrrelevantInputTypes)
+      .map((t) => `:not([type=${t}])`)
+      .join("");
+    this.autofillRelevantDescendantSelector = `form, input${inputExclusions}, select, textarea, span[data-bwautofill]`;
```
jprusik: Any reason we wouldn't use the existing and more comprehensive `inputQuery`?

```suggestion
    this.autofillRelevantDescendantSelector = `form, ${inputQuery}, select, textarea, span[data-bwautofill]`;
```

jprusik: it strikes me that this query selector has significant overlap with `this.formFieldQueryString` - perhaps use that?

apps/browser/src/autofill/services/collect-autofill-content.service.ts
```diff
@@ -76,6 +76,19 @@ export class CollectAutofillContentService implements CollectAutofillContentServ
   private readonly maxMutationWaitMs = 5000;
   private readonly formFieldQueryString;
   private readonly nonInputFormFieldTags = new Set(["textarea", "select"]);
+  // INPUT intentionally omitted; type-aware check below admits text-likes only.
+  private readonly autofillRelevantTags = new Set(["FORM", "SELECT", "TEXTAREA"]);
+  // Distinct from ignoredInputTypes — format-restricted text types still admit here.
```
jprusik: What types are you looking to capture that were being dropped by `ignoredInputTypes`?

jprusik (commented):
I've a few questions before approving

apps/browser/src/autofill/services/collect-autofill-content.service.ts
blackwood: Is that more performant? The `map` still ends up necessary to decorate the selectors with `[type=*]`

apps/browser/src/autofill/services/collect-autofill-content.service.ts
blackwood: I don't disagree--I was imagining a world in which we may have preferred a wider net in the gate, but let's close the gap since in a perfect world those fields will still be ignored later.

apps/browser/src/autofill/services/collect-autofill-content.service.ts
jprusik: probably not, just more concise; not an ask, just a note.

Like, we could avoid a join with a reducer instead of a map, but that feels very much like only a minor stylistic opinion on the far end of diminishing returns 🤷

apps/browser/src/autofill/services/collect-autofill-content.service.ts
```diff
@@ -1557,6 +1560,32 @@ export class CollectAutofillContentService implements CollectAutofillContentServ
     }
   }
 
+  private mutationAffectsAutofillTarget(mutation: MutationRecord): boolean {
```
jprusik, line 1563: I kinda wish there was a name for this to better signal that this doesn't handle attribute mutations 🤔

jprusik (approved):

apps/browser/src/autofill/services/collect-autofill-content.service.ts
blackwood: `nonAttributeMutationAffectsAutofillTarget`?

blackwood:
Superseded by https://github.com/bitwarden/clients/pull/21084

PR #20961, opened 2026-06-01
[CL-1200] remove visible label from table header checkboxes

BryanCunningham:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/CL-1200

## 📔 Objective
To improve the bulk actions experience, design has requested that we move to a checkbox-only table header for the select all checkbox functionality. To keep the header accessible, this PR includes replacing the existing visible text with an `aria-label` if one did not already exist. (As long as the table header has accessible text, it actually does not need to be visible.)

Scope:
- Removes visible label from all table header 'select all' checkboxes
- Adds aria-label for the checkbox that says 'Select all'

## 📸 Screenshots
image


eliykat (dismissed):
Thanks for handling this!

Banrion (dismissed):

apps/web/src/app/admin-console/organizations/members/members.component.html
```diff
@@ -90,13 +90,11 @@
                   <input
                     type="checkbox"
                     bitCheckbox
-                    class="tw-mr-1"
                     (change)="dataSource.checkAllFilteredUsers($any($event.target).checked)"
-                    id="selectAll"
+                    [attr.aria-label]="
+                      (dataSource.isAllFilteredChecked() ? 'deselectAll' : 'selectAll') | i18n
+                    "
```
willmartian: Talked sync on this, but for posterity: 

We shouldn't dynamically flip checkbox text because it can present a confusing double negative to screenreader users. "Deselect all unchecked" etc

willmartian (changes requested):

apps/web/src/app/admin-console/organizations/members/members.component.html
BryanCunningham: @willmartian Updated [here](https://github.com/bitwarden/clients/pull/20961/changes/5dbd27d6d910a4df9e3c726fe198ea9fde056cd7)

willmartian (approved):

apps/web/src/app/admin-console/organizations/members/members.component.html
willmartian: thanks!

eliykat (approved):

vincentsalucci (approved):
💭 Non-blocking; just wanted to confirm the desire to move from "all" to "select all" for the aria labels?

Banrion (approved):

BryanCunningham:
> 💭 Non-blocking; just wanted to confirm the desire to move from "all" to "select all" for the aria labels?

@vincentsalucci TBH, this just felt 'better' to me personally. More explicit/action oriented. Not really sure who I should really even check with to confirm this :) 

I can ask around design and sanity check for good measure. And, if they want something different, I suppose it's easy enough to swap back as a fast follow

shane-melton (approved):

PR #20966, opened 2026-06-01
[PM-38400] Fix avoid ambiguous character bug

adudek-bw:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-38400

## 📔 Objective

Fixes bug where the avoid ambiguous logic is backwards. Seen here: https://bitwarden.slack.com/archives/C04PZA13FQD/p1780329220086709?thread_ts=1780328285.918579&cid=C04PZA13FQD

### DIFF
--- a/libs/tools/generator/core/src/engine/sdk-password-randomizer.spec.ts
+++ b/libs/tools/generator/core/src/engine/sdk-password-randomizer.spec.ts
@@ -46,17 +46,91 @@ describe("SdkPasswordRandomizer", () => {
 
       await sdk.generate({ algorithm: Algorithm.password }, settings);
 
+      // `ambiguous: true` means ambiguous characters MAY be included, so the
+      // SDK's `avoidAmbiguous` (the exclude flag) must be `false`.
       expect(generator.password).toHaveBeenCalledWith({
         lowercase: true,
         uppercase: true,
         numbers: true,
         special: true,
         length: 16,
-        avoidAmbiguous: true,
+        avoidAmbiguous: false,
         minLowercase: 2,
         minUppercase: 1,
         minNumber: 3,
         minSpecial: 4,
+      });
+    });
+
+    it("passes ambiguous: false through to the SDK as avoidAmbiguous: true", async () => {
+      // Regression test for the semantic-inversion bug where `avoidAmbiguous` was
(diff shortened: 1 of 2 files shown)


libs/tools/generator/core/src/engine/sdk-password-randomizer.spec.ts
```diff
@@ -46,20 +46,94 @@ describe("SdkPasswordRandomizer", () => {
 
       await sdk.generate({ algorithm: Algorithm.password }, settings);
 
+      // `ambiguous: true` means ambiguous characters MAY be included, so the
+      // SDK's `avoidAmbiguous` (the exclude flag) must be `false`.
```
harr1424, line 50: ❓ The whole wording around this seems unnecessarily confusing.  Why not standardize the language here? I realize it will affect code outside of just the spec file, but I think the SDK did this the right way: `avoidAmbiguous` is very clear in terms of the expected behavior. `ambiguous` is by itself ambiguous.

libs/tools/generator/core/src/engine/sdk-password-randomizer.spec.ts
harr1424, line 32: 😢  This probably should have been named `avoidAmbiguous` ?

harr1424 (changes requested):
I think we would benefit long term from re-working the naming used here, which in my understanding is responsible for both the regression noticed today and the test failures in the last hour.

libs/tools/generator/core/src/engine/sdk-password-randomizer.spec.ts
adudek-bw: Yeah, that would def lead to less confusion!

adudek-bw: So I think that might be a good followup ticket, since this is supposed to be a quick fix. Would that be cool with you?

adudek-bw: But I for sure def agree 😂

harr1424 (dismissed):

itsadrago (approved):

harr1424 (approved):

PR #20967, opened 2026-06-01
PM-38423 - Fix unsavedSendEditsGuard blocking logout

JaredSnider-Bitwarden:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-38423

## 📔 Objective

To resolve an issue w/ the new `unsavedSendEditsGuard` on web where, when the app was logging out, angular would emit null and the guard would attempt to call `null.saveUnsavedSendEdits()` and get a `Cannot read properties of null (reading 'saveUnsavedSendEdits')` error.

Note: I adjusted the logic in the desktop guard, but it doesn't actually seem to work? See below. Logging out sometimes flashes the discard edits dialog but it is quickly and automatically closed without interaction. Regardless, it still seemed safer to add the logic to the desktop guard since Angular itself can emit null for the component in some scenarios. 

## 📸 Screenshots

Web:

https://github.com/user-attachments/assets/e7213920-bcac-488e-b023-73b43ea914e8

Desktop:

https://github.com/user-attachments/assets/cb6ca5d8-069c-490c-9f9f-edb7ccec231a

### DIFF
--- a/apps/desktop/src/app/tools/send/guards/unsaved-send-edits.guard.ts
+++ b/apps/desktop/src/app/tools/send/guards/unsaved-send-edits.guard.ts
@@ -3,5 +3,10 @@ import { CanDeactivateFn } from "@angular/router";
 import { SendComponent } from "../send.component";
 
 export const unsavedSendEditsGuard: CanDeactivateFn<SendComponent> = async (component) => {
+  // Angular passes null when the component has already been destroyed mid-navigation
+  // (e.g. during logout teardown). No edits to save in that case — allow the navigation.
+  if (component == null) {
+    return true;
+  }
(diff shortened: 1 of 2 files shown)


harr1424 (commented):
I am not familiar with this issue (yet) but wanted to ask if something similar should be implemented for the browser (extension) client?

JaredSnider-Bitwarden:
> I am not familiar with this issue (yet) but wanted to ask if something similar should be implemented for the browser (extension) client?

@harr1424 , I checked, but I only saw the `unsavedSendEditsGuard` on web and desktop - not extension?

harr1424 (approved):

PR #20970, opened 2026-06-01
[PM-34491] Fill Targeting Rules implementation gaps

jprusik:
## 🎟️ Tracking

[PM-34491](https://bitwarden.atlassian.net/browse/PM-34491)

## 📔 Objective

Fixed:
- `newPassword` selector arrays match up to two elements (instead of one); instead of stopping on the first new password field, fill will handle up to two new password fields if specified in the targeting rules.

Deferred:
- ~`newPassword` targeted fields no longer fill with the cipher's current password during a normal Login fill.~ 
  (split to https://github.com/bitwarden/clients/pull/21238)
- ~`newPassword` targeted fields now map to `InlineMenuFillTypes.PasswordGeneration` (not `CipherType.Login`), so the inline menu offers the generator instead of the cipher list.~ 
  (split to https://github.com/bitwarden/clients/pull/21238)
- ~`generateTargetedFillScript` honors `PasswordGeneration` by routing `cipher.login.password` into targeted `newPassword` fields, otherwise the menu would appear but fail to fill.~ 
  (split to https://github.com/bitwarden/clients/pull/21239)
- ~Filling a generated password no longer prunes targeted `newPassword` fields through heuristic filter before fill~ 
  (split to https://github.com/bitwarden/clients/pull/21245)
- ~the inline menu save prompt no longer fires on "new password" fields just because the current-password field was filled on an update form.~
  (split to https://github.com/bitwarden/clients/pull/21244)
- ~autofill logic now leverages provided selectors for form "submit" actions (other actions and auto submit support not implemented)~ 
  (split to [PM-38476](https://bitwarden.atlassian.net/browse/PM-38476))

[PM-34491]: https://bitwarden.atlassian.net/browse/PM-34491?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[PM-38476]: https://bitwarden.atlassian.net/browse/PM-38476?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/browser/src/autofill/background/overlay.background.ts
+++ b/apps/browser/src/autofill/background/overlay.background.ts
@@ -2230,10 +2230,16 @@ export class OverlayBackground implements OverlayBackgroundInterface {
       return false;
     }
 
+    // On password-generation fields the save prompt should only fire once a
+    // new password has actually been entered. Gating on `loginData.password`
+    // here would surface the prompt as soon as the *current* password field
+    // on an update form is filled, which isn't a "save new login" signal.
+    const hasAnyPassword = !!(loginData.password || loginData.newPassword);
+    const hasNewPassword = !!loginData.newPassword;
+
     return (
-      (this.shouldShowInlineMenuAccountCreation() ||
-        this.focusedFieldMatchesFillType(InlineMenuFillTypes.PasswordGeneration)) &&
-      !!(loginData.password || loginData.newPassword)
+      (this.shouldShowInlineMenuAccountCreation() && hasAnyPassword) ||
+      (this.focusedFieldMatchesFillType(InlineMenuFillTypes.PasswordGeneration) && hasNewPassword)
     );
   }
 
--- a/apps/browser/src/autofill/background/overlay.background.ts
+++ b/apps/browser/src/autofill/background/overlay.background.ts
@@ -2245,8 +2245,8 @@ export class OverlayBackground implements OverlayBackgroundInterface {
     const hasNewPassword = !!loginData.newPassword;
 
     return (
-      (this.shouldShowInlineMenuAccountCreation() && hasAnyPassword) ||
-      (this.focusedFieldMatchesFillType(InlineMenuFillTypes.PasswordGeneration) && hasNewPassword)
+      (hasAnyPassword && this.shouldShowInlineMenuAccountCreation()) ||
+      (hasNewPassword && this.focusedFieldMatchesFillType(InlineMenuFillTypes.PasswordGeneration))
     );
   }
 
--- a/apps/browser/src/autofill/background/overlay.background.ts
+++ b/apps/browser/src/autofill/background/overlay.background.ts
@@ -23,6 +23,7 @@ import { AuthenticationStatus } from "@bitwarden/common/auth/enums/authenticatio
 import { getOptionalUserId, getUserId } from "@bitwarden/common/auth/services/account.service";
 import {
   AutofillOverlayVisibility,
+  AutofillTargetingRuleTypes,
   SHOW_AUTOFILL_BUTTON,
 } from "@bitwarden/common/autofill/constants";
 import { AutofillSettingsServiceAbstraction } from "@bitwarden/common/autofill/services/autofill-settings.service";
@@ -2172,10 +2173,16 @@ export class OverlayBackground implements OverlayBackgroundInterface {
 
       // If our currently focused field is for a login form, we want to fill the current password field.
       // Otherwise, map over all page details and filter out fields that are not new password fields.
+      // Targeted fields qualified as `newPassword` by targeting rules bypass the heuristic
+      // check; the rules represent an explicit assertion that the field is a new-password
+      // field, and heuristic keyword tokenization can miss compact names like `confirmPassword`.
       if (!this.focusedFieldMatchesFillType(CipherType.Login)) {
         pageDetails = this.getFilteredPageDetails(
           pageDetails,
-          this.inlineMenuFieldQualificationService.isNewPasswordField,
+          (field) =>
+            (field.targeted === true &&
+              field.fieldQualifier === AutofillTargetingRuleTypes.newPassword) ||
+            this.inlineMenuFieldQualificationService.isNewPasswordField(field),
         );
       }
 
--- a/apps/browser/src/autofill/services/autofill-overlay-content.service.ts
+++ b/apps/browser/src/autofill/services/autofill-overlay-content.service.ts
@@ -9,7 +9,6 @@ import {
   AUTOFILL_TRIGGER_FORM_FIELD_SUBMIT,
   EVENTS,
 } from "@bitwarden/common/autofill/constants";
-import { AutofillTargetingRuleType } from "@bitwarden/common/autofill/types";
 import { CipherType } from "@bitwarden/common/vault/enums";
 
 import { ModifyLoginCipherFormData } from "../background/abstractions/overlay-notifications.background";
@@ -61,6 +60,7 @@ import {
   loginQualifiers,
   cardQualifiers,
   identityQualifiers,
+  isAutofillTargetingRuleType,
 } from "./autofill-constants";
 
 export class AutofillOverlayContentService implements AutofillOverlayContentServiceInterface {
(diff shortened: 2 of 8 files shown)


apps/browser/src/autofill/background/overlay.background.ts
```diff
@@ -2230,10 +2237,16 @@ export class OverlayBackground implements OverlayBackgroundInterface {
...
+    // on an update form is filled, which isn't a "save new login" signal.
+    const hasAnyPassword = !!(loginData.password || loginData.newPassword);
+    const hasNewPassword = !!loginData.newPassword;
+
     return (
-      (this.shouldShowInlineMenuAccountCreation() ||
-        this.focusedFieldMatchesFillType(InlineMenuFillTypes.PasswordGeneration)) &&
-      !!(loginData.password || loginData.newPassword)
+      (this.shouldShowInlineMenuAccountCreation() && hasAnyPassword) ||
+      (this.focusedFieldMatchesFillType(InlineMenuFillTypes.PasswordGeneration) && hasNewPassword)
```
audreyality: ⛏️ 

```suggestion
      (hasAnyPassword && this.shouldShowInlineMenuAccountCreation()) ||
      (hasNewPassword && this.focusedFieldMatchesFillType(InlineMenuFillTypes.PasswordGeneration))
```

apps/browser/src/autofill/models/autofill-form.ts
```diff
@@ -39,4 +39,14 @@ export default class AutofillForm {
...
   htmlAncestorHeadings!: string[];
+
+  /**
+   * Opid assigned to the form's submit button when a targeting rule's
+   * `actions.submit` selector resolved during collection. Absent when no rule
+   * applied, when the rule did not provide an `actions.submit` selector, or
+   * when none of the provided selectors matched. The targeted-submit-button
+   * registry on the overlay service distinguishes the unresolved case (which
+   * suppresses heuristic submit-button discovery) from the no-rule case.
+   */
```
audreyality: 🤔 Consider restating this to be independent of targeting rules. We'll probably want to heuristically determine the submit button at some point.

apps/browser/src/autofill/services/autofill-overlay-content.service.ts
```diff
@@ -74,6 +75,19 @@ export class AutofillOverlayContentService implements AutofillOverlayContentServ
...
+   *
+   * - Entry present, element non-null: rule resolved a button; use it as the
+   *   authoritative click target for submit-listener wiring.
+   * - Entry present, element null: rule declared `actions.submit` but no
+   *   selector matched; suppress heuristic submit-button discovery for the
+   *   form so a stale/wrong listener isn't attached.
+   * - No entry: no targeting rule applied to the form, or the rule did not
+   *   declare `actions.submit`; heuristic discovery runs as usual.
+   */
+  private targetedSubmitButtonsByFormOpid: Map<string, HTMLElement | null> = new Map();
```
audreyality: ❓ Is `null` necessary here? We're supposed to prefer `undefined` per [ADR-14](https://contributing.bitwarden.com/architecture/adr/typescript-strict#avoid-null-prefer-undefined).

apps/browser/src/autofill/services/autofill-overlay-content.service.ts
```diff
@@ -451,6 +474,34 @@ export class AutofillOverlayContentService implements AutofillOverlayContentServ
     return;
```
audreyality, line 452: 🎨 Consider using `if/else if/else` instead of early returns. This makes it more clear that this function is performing dispatch.

```ts
    if (
      !elementIsFillableFormField(formFieldElement) ||
      autofillFieldData.inlineMenuFillType === CipherType.Card
    ) {
      return;
    } else if (
      autofillFieldData.targeted &&
      autofillFieldData.form &&
      this.targetedSubmitButtonsByFormOpid.has(autofillFieldData.form)
    ) {
      this.setupSubmitListenerOnTargetedField(formFieldElement, autofillFieldData.form);
    } else if (autofillFieldData.form) {
      await this.setupSubmitListenerOnFieldWithForms(formFieldElement);
    } else {
      await this.setupSubmitListenerOnFormlessField(formFieldElement);
    }
```

🤔 It also reveals a possible bug. The `autofillFieldData.form` branch should probably be:

```ts
await this.setupSubmitListenerOnFieldWithForms(autofillFieldData.form);
```

apps/browser/src/autofill/services/autofill-overlay-content.service.ts
```diff
@@ -1176,6 +1227,11 @@ export class AutofillOverlayContentService implements AutofillOverlayContentServ
     // which are distinct from the heuristic AutofillFieldQualifierType values.
     const qualifier = autofillFieldData.fieldQualifier as AutofillTargetingRuleType;
 
+    if (qualifier === AutofillTargetingRuleTypes.newPassword) {
+      autofillFieldData.inlineMenuFillType = InlineMenuFillTypes.PasswordGeneration;
+      return;
+    }
```
audreyality: 🎨 This is another location where I'd consider using `if/else` blocks instead of early returns. Alternatively, making the function calculate an `InlineMenuFillType` instead of assigning it would ensure that a `return` without a value would fail to typecheck.

♻️ The `as AutofillTargetingRuleType` type assertion on line 1228 does not guarantee that `fieldQualifier` is typesafe. If you decide to refactor the method, consider a [type guard](https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates) instead of an assertion.

apps/browser/src/autofill/services/collect-autofill-content.service.ts
```diff
@@ -1,6 +1,21 @@
...
+ * Per-field-type cap on how many DOM matches the targeted-collection pass
+ * accepts from a selector array. Selector arrays are alternatives by default
+ * (first match wins), but certain field types appear in pairs on the same
+ * form — most notably `newPassword`, which is commonly mirrored by a
+ * "confirm new password" input on registration and password-update flows.
+ */
+const MAX_MATCHES_BY_FIELD_TYPE: Partial<Record<AutofillTargetingRuleType, number>> = {
+  [AutofillTargetingRuleTypes.newPassword]: 2,
+};
+const DEFAULT_MAX_MATCHES = 1;
```
audreyality: ⛏️ These declarations shouldn't be in the middle of the import block.

🎨 Consider freezing `MAX_MATCHES_BY_FIELD_TYPE` so that it cannot be overwritten at runtime.


```ts
const MAX_MATCHES_BY_FIELD_TYPE: Partial<Record<AutofillTargetingRuleType, number>> = Object.freeze({
  [AutofillTargetingRuleTypes.newPassword]: 2,
});
```

apps/browser/src/autofill/services/collect-autofill-content.service.ts
```diff
@@ -295,28 +351,35 @@ export class CollectAutofillContentService implements CollectAutofillContentServ
               }
               iframeTargets.get(iframeSrc)!.push({
                 selector: innerSelector,
-                fieldType: fieldType as AutofillTargetingRuleType,
+                fieldType: typedFieldType,
               });
             }
-            break;
+            matchCount++;
+            continue;
```
audreyality: ⚠️ This `continue` continues the for loop on line 333. The line it replaces broke out of the for loop. It's unlikely that this `continue` is correct, since its loop ends immediately afterward.

♻️ This function's cyclomatic complexity is too high. Consider refactoring it to use array processing (`map`, `filter`, and `reduce`) instead of highly nested `if` statements to reduce the complexity.

apps/browser/src/autofill/services/collect-autofill-content.service.ts
```diff
@@ -354,12 +417,103 @@ export class CollectAutofillContentService implements CollectAutofillContentServ
...
+      }
+      if (this.domQueryService.findIframeCrossing(selector)) {
+        continue;
+      }
+      const match = this.domQueryService.queryDeepSelector(selector);
+      if (match instanceof HTMLElement) {
+        return match;
+      }
+    }
+    return null;
```
audreyality, line 477: 🎨 Consider using an array method. Control flow is harder to follow than comparison logic. If you'd like to preserve the `return match` breakpoint, split the `find` and `return` onto different lines.

```suggestion
    const found = container.find(
      (s) => 
        typeof selector s === "string" &&
        !this.domQueryService.findIframeCrossing(selector) &&
        this.domQueryService.queryDeepSelector(selector) instanceof HTMLElement
    );
    return found ?? null;
```

audreyality (changes requested):
♻️ I'm concerned about how verbose this code is. I've pointed out a few places where this impedes readability and in one case it appears to be related to a bug.

apps/browser/src/autofill/models/autofill-form.ts
jprusik: I decided to split the submit button implementation work to it's own task, since it deserves some more thought (and remaining implementation of other action types). I'm referencing this feedback so we don't lose track of it as well.

Moved to [PM-38476](https://bitwarden.atlassian.net/browse/PM-38476)

apps/browser/src/autofill/services/autofill-overlay-content.service.ts
jprusik: Moved to [PM-38476](https://bitwarden.atlassian.net/browse/PM-38476)

apps/browser/src/autofill/services/autofill-overlay-content.service.ts
jprusik: > The autofillFieldData.form branch should probably be:

fwiw, my understanding of `setupSubmitListenerOnFieldWithForms`: this is meant to take a field and any form associated with it that isn't already in `formElements` is added to it and given submit event listeners

apps/browser/src/autofill/services/collect-autofill-content.service.ts
jprusik: Addressed, i believe, in the refactor

jprusik:
Apologies for the rebase - I had to pull in some important related changes that have been merged to `main`.

apps/browser/src/autofill/services/collect-autofill-content.service.ts
```diff
@@ -278,93 +302,204 @@ export class CollectAutofillContentService implements CollectAutofillContentServ
...
+    let matchCount = 0;
+
+    for (const selector of selectorAlternatives ?? []) {
+      if (matchCount >= maxMatches) {
+        break;
+      }
+
+      if (typeof selector !== "string") {
+        continue;
+      }
```
audreyality, line 373: 🎨 The number of `continue` branches in this loop suggest that pre-filtering `selectorAlternatives` would be worthwhile.

apps/browser/src/autofill/services/collect-autofill-content.service.ts
```diff
@@ -278,93 +302,204 @@ export class CollectAutofillContentService implements CollectAutofillContentServ
...
+    const isFormElement = element instanceof HTMLFormElement;
+    const form = new AutofillForm();
+    form.opid = opid;
+    form.htmlID = this.getPropertyOrAttribute(element, AUTOFILL_ATTRIBUTES.ID) ?? "";
+    form.htmlName = this.getPropertyOrAttribute(element, AUTOFILL_ATTRIBUTES.NAME) ?? "";
+    form.htmlClass = this.getPropertyOrAttribute(element, "class") ?? "";
+    form.htmlAction = isFormElement
+      ? (this.getFormActionAttribute(element as ElementWithOpId<HTMLFormElement>) ?? "")
+      : "";
+    form.htmlMethod = isFormElement
```
audreyality, line 496: ⛏️ The type assertion is incorrect. `getFormActionAttribute()` never uses `opid`, however, so you can fix this by narrowing the argument's type to `HtmlFormElement`.

audreyality (approved):

jprusik:
Remaining work has been carried over to https://github.com/bitwarden/clients/pull/21364 - linking this pull request to preserve existing feedback.

PR #20999, opened 2026-06-02
[PM-38405] Fix SDK ref leaking out of scope and add eslint rule

quexten:
## 🎟️ Tracking

Jira Bug: https://bitwarden.atlassian.net/browse/PM-38405

## 📔 Objective

```ts
        await firstValueFrom( // Promise on ref.value gets awaited here
          this.sdkService.userClient$(userId).pipe(
            map((sdk) => {
              if (!sdk) {
                throw new Error("SDK not available");
              }

              using ref = sdk.take();

              return ref.value.user_crypto_management().migrate_to_key_connector(keyConnectorUrl); // but created here. also, the return disposes the ref
            }),
          ),
```

the promise is created inside the observable, but passed out and holds an invalid reference and tries to access it. It should be awaited.

This PR fixes existing cases and adds an eslint rule that enforces correct behavior. It is no longer allowed to return unawaited promised if a using lives in the scope. This is not possible with existing eslint rules.

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -83,6 +83,7 @@ export default tseslint.config(
       "@bitwarden/platform/required-using": "error",
       "@bitwarden/platform/no-enums": "error",
       "@bitwarden/platform/no-page-script-url-leakage": "error",
+      "@bitwarden/platform/no-unawaited-using-return": "error",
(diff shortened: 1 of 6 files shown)


mzieniukbw (approved):
Nice find! Approved KM code change side.

coroiu (approved):
Thank you for this!

PR #21004, opened 2026-06-02
[PM-35796] Use SDK bindings for invite link generation

BTreston:
## 🎟️ Tracking
https://bitwarden.atlassian.net/browse/PM-35796

## 📔 Objective
Wires up and unit tests the SDK implementation of `generateOrganizationInviteCryptoBundle` in `OrganizationInviteLinkService` via the newly generated WASM bindings from https://github.com/bitwarden/sdk-internal/pull/1157


BTreston:
@claude review

BTreston:
@claude re-review

eliykat:
Let's wait until https://github.com/bitwarden/sdk-internal/pull/1196 is merged to make sure that your PR is compatible with those changes.

eliykat (commented):
Confirming my comment above - this will change (albeit only slightly) when we adopt the `Invite` blob structure. I recommend reverting to draft until then. The SDK PR will have merged by then as well.

BTreston:
@r-tome Adding you as a reviewer for visibility purposes

libs/organization-invite-link/src/services/default-organization-invite-link.service.ts
```diff
@@ -44,47 +42,59 @@ export class DefaultOrganizationInviteLinkService implements OrganizationInviteL
...
-
-    await this.upsert(userId, inviteLink);
+    await this.upsert(userId, new OrganizationInviteLink(response));
   }
 
   async updateInviteLink(
     userId: UserId,
     orgId: OrganizationId,
     allowedDomains: string[],
+    supportsConfirmation: boolean,
```
eliykat: I'm not sure `supportsConfirmation` should be added here. If it's being updated, then we'd need to call the SDK to actually update the `Invite` to support confirmation. But that's not done here, and it doesn't seem to be the purpose of the method - should this just be called `updateAllowedDomains` instead?

libs/organization-invite-link/src/services/default-organization-invite-link.service.ts
```diff
@@ -127,29 +137,13 @@ export class DefaultOrganizationInviteLinkService implements OrganizationInviteL
...
-  private getOrgKey(userId: UserId, orgId: OrganizationId) {
-    return this.keyService.orgKeys$(userId).pipe(
-      map((orgKeys) => {
-        const orgKey = orgKeys?.[orgId] ?? undefined;
-        if (orgKey == null) {
-          throw new Error(`Organization key not found for org ${orgId}`);
-        }
-
-        return orgKey;
+  private generateEncryptedKey(userId: UserId, orgId: OrganizationId): Observable<InviteBundle> {
```
eliykat: `generateEncryptedKey` is too specific, I suggest this mirrors the SDK interface given that it's just a convenience wrapper: `makeInvite`.

libs/organization-invite-link/src/services/default-organization-invite-link.service.ts
```diff
@@ -44,47 +42,59 @@ export class DefaultOrganizationInviteLinkService implements OrganizationInviteL
...
+    await this.upsert(userId, new OrganizationInviteLink(response));
   }
 
-  async refreshInviteLink(userId: UserId, orgId: OrganizationId) {
-    const encryptedInviteKey = await this.generateEncryptedKey(userId, orgId);
-    const request = new OrganizationInviteLinkRefreshRequest({ encryptedInviteKey });
+  async refreshInviteLink(
+    userId: UserId,
+    orgId: OrganizationId,
+    supportsConfirmation: boolean = false,
```
eliykat: Same concern here with `supportsConfirmation`.

libs/organization-invite-link/src/services/default-organization-invite-link.service.ts
```diff
@@ -1,14 +1,14 @@
...
-import { EncString } from "@bitwarden/common/key-management/crypto/models/enc-string";
 import { ErrorResponse } from "@bitwarden/common/models/response/error.response";
 import { EnvironmentService } from "@bitwarden/common/platform/abstractions/environment.service";
+import { asUuid, SdkService } from "@bitwarden/common/platform/abstractions/sdk/sdk.service";
 import { OrganizationId, UserId } from "@bitwarden/common/types/guid";
-import { KeyService } from "@bitwarden/key-management";
+import { Invite, OrganizationId as SdkOrganizationId } from "@bitwarden/sdk-internal";
 import { StateProvider } from "@bitwarden/state";
 
+type InviteBundle = { invite: Invite };
```
eliykat: ⛏️(non-blocking) not sure why this is necessary - why not just use `Invite` directly?

eliykat (changes requested):

libs/organization-invite-link/src/services/default-organization-invite-link.service.ts
BTreston: The server models expect these for the crud operations. As its implemented in this PR, its hard coded to `false` by the callers and I thought it would be up to the work in M3 to pass in the real value. I left a TODO comment at the call site noting this.

r-tome: The server update endpoint does not expect the `supportsConfirmation` property. https://github.com/bitwarden/server/blob/main/src/Api/AdminConsole/Models/Request/Organizations/UpdateOrganizationInviteLinkRequestModel.cs
`Invite` and `SupportsConfirmation` should always be set together, so I think we should remove it from here.
I named the endpoint "Update" so it could have more purpose in the future and not be tied to a single property but I'm not against renaming it.

libs/organization-invite-link/src/services/default-organization-invite-link.service.ts
r-tome: For refresh I think we should keep it because it will regenerate the link via SDK. Clients will send the `Invite` blob so it should as well send `supportsConfirmation`

libs/organization-invite-link/src/services/default-organization-invite-link.service.ts
BTreston: That's my bad then. I could have sworn I saw it on this endpoint too, but I must have mistaken seeing it in the sproc. I'll make the fix for this function

eliykat (approved):
Thanks!

PR #21013, opened 2026-06-03
[PM-34926] Allow single-character CJK vault searches

### LINKED ISSUE #20110: Desktop: Single CJK character search fails when UI language is English
### Steps To Reproduce

1. Set the macOS Bitwarden desktop client language to English.
2. Create a vault entry with the name "密碼".
3. Type "密" into the search bar.
4. Observe that no results are returned.
5. Type "密碼" into the search bar.
6. Observe that the entry is successfully found.
7. (Optional) Change the UI language to Traditional Chinese, search for "密" again, and observe it successfully finds the entry.

### Expected Result

Searching for a single CJK character (like "密") should return vault entries containing that character, regardless of the application's UI language setting. Since CJK characters often represent whole words or concepts, a 1-character search is perfectly valid.

### Actual Result

The search returns no results. The minimum search character limit is tied to the UI language setting rather than the input itself. When the app is in English, the standard 2-character minimum is enforced, completely blocking valid 1-character CJK searches.

### Additional Context

Multilingual users who run their OS or apps in English but store entries in Chinese, Japanese, or Korean are heavily penalized by this limitation. A more robust approach would be to evaluate the search string itself: if the query contains a CJK ideograph, the minimum search length requirement should dynamically drop to 1. 

### Operating System

macOS

### Operating System Version

macOS Tahoe 26.3 (25D125) arm64

### Installation method

Mac App Store

### Build Version

Version 2026.3.1 (58867)

### Issue Tracking Info

jaasen-livefront:
## 🎟️ Tracking

[PM-34926](https://bitwarden.atlassian.net/browse/PM-34926)

## 📔 Objective

Allow single-character CJK vault search queries to return matching vault entries regardless of the active UI language.

Previously, the minimum searchable query length was based only on the active locale. This meant users running the app in English needed at least two characters before search would run, even when the query was a valid one-character CJK term such as `密`.

This updates `SearchService` to detect CJK characters in the query itself and allow immediate search for those queries, while preserving the existing minimum length behavior for non-CJK one-character searches.

This also adds regression coverage for:

- Single-character CJK queries being searchable under the default English locale
- Single-character non-CJK queries remaining non-searchable
- Searching ciphers by a single CJK character, such as finding `密碼` with `密`

## 📸 Screenshots

https://github.com/user-attachments/assets/fa13fdb0-e527-489f-acae-dfbc9544a54e

[PM-34926]: https://bitwarden.atlassian.net/browse/PM-34926?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/common/src/vault/services/search.service.ts
+++ b/libs/common/src/vault/services/search.service.ts
@@ -15,6 +15,9 @@ export const SearchTextDebounceInterval = 100; // milliseconds
 
 export class SearchService implements SearchServiceAbstraction {
   private readonly immediateSearchLocales: string[] = ["zh-CN", "zh-TW", "ja", "ko", "vi"];
+  // Immediately search for CJK characters as they can represent complete search terms, regardless of the active locale.
+  private readonly immediateSearchQueryRegex =
+    /[\u3040-\u30ff\u3400-\u4dbf\u4e00-\u9fff\uf900-\ufaff\uac00-\ud7af\u{20000}-\u{2ebef}]/u;
   private readonly defaultSearchableMinLength: number = 2;
   private searchableMinLength: number = this.defaultSearchableMinLength;
 
@@ -46,6 +49,10 @@ export class SearchService implements SearchServiceAbstraction {
     }
 
     query = normalizeSearchQuery(query);
+
+    if (this.immediateSearchQueryRegex.test(query)) {
+      return true;
+    }
 
     // Regular queries only require a minimum length
     return query.length >= this.searchableMinLength;
--- a/libs/common/src/vault/services/search.service.ts
+++ b/libs/common/src/vault/services/search.service.ts
@@ -17,7 +17,7 @@ export class SearchService implements SearchServiceAbstraction {
   private readonly immediateSearchLocales: string[] = ["zh-CN", "zh-TW", "ja", "ko", "vi"];
   // Immediately search for CJK characters as they can represent complete search terms, regardless of the active locale.
   private readonly immediateSearchQueryRegex =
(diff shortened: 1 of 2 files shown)


libs/common/src/vault/services/search.service.ts
```diff
@@ -15,6 +15,9 @@ export const SearchTextDebounceInterval = 100; // milliseconds
 
 export class SearchService implements SearchServiceAbstraction {
   private readonly immediateSearchLocales: string[] = ["zh-CN", "zh-TW", "ja", "ko", "vi"];
+  // Immediately search for CJK characters as they can represent complete search terms, regardless of the active locale.
+  private readonly immediateSearchQueryRegex =
```
shane-melton, line 19: 🎨 Could we look at using Unicode script property escapes instead? They're more self-documenting and look less like magic 🪄 

```ts 
private readonly immediateSearchQueryRegex =
  /[\p{Script=Han}\p{Script=Hiragana}\p{Script=Katakana}\p{Script=Hangul}]/u;
```

shane-melton (commented):
One small suggestion, otherwise looks good.

libs/common/src/vault/services/search.service.ts
jaasen-livefront: @shane-melton TIL! https://github.com/bitwarden/clients/pull/21013/commits/69ea5ada1f3248d5ccf413b8fe181804aced135a

shane-melton (approved):

alexianos11:
Desktop v2026.6.1 was released 17 hours ago. CJK search does not work.
I understand this possible fix had been merged 2 weeks ago, so, something else is causing the problem?

PR #21023, opened 2026-06-03
[PM-38421] Remove duplicate copy actions

nick-livefront:
## 🎟️ Tracking

[PM-38421](https://bitwarden.atlassian.net/browse/PM-38421)

## 📔 Objective

Removes the duplicate Launch and Copy actions from the Desktop vault item ellipsis/right-click context menu, bringing it into parity with the Web app which already omits these once dedicated row buttons are present.

## 📸 Screenshots

<video src="https://github.com/user-attachments/assets/9d59937e-e82f-4a47-9b15-c16f48d8e933" />

[PM-38421]: https://bitwarden.atlassian.net/browse/PM-38421?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/desktop/src/locales/en/messages.json
+++ b/apps/desktop/src/locales/en/messages.json
@@ -1824,6 +1824,15 @@
     "message": "Copy security code",
     "description": "Copy credit card security code (CVV)"
   },
+  "copyPrivateKey": {
+    "message": "Copy private key"
+  },
+  "copyPublicKey": {
+    "message": "Copy public key"
+  },
+  "copyFingerprint": {
+    "message": "Copy fingerprint"
+  },
   "cardNumber": {
     "message": "card number"
   },
--- a/apps/desktop/src/vault/app/vault-v3/vault-items/vault-cipher-row.component.html
+++ b/apps/desktop/src/vault/app/vault-v3/vault-items/vault-cipher-row.component.html
@@ -101,21 +101,6 @@
       label="{{ 'options' | i18n }}"
     ></button>
     <bit-menu #cipherOptions>
-      @if (canLaunch()) {
-        <button type="button" bitMenuItem (click)="handleLaunch()">
-          <i class="bwi bwi-fw bwi-external-link" aria-hidden="true"></i>
-          {{ "launch" | i18n }}
-        </button>
-      }
-      @for (copyField of copyFields(); track copyField.field) {
-        <button type="button" bitMenuItem [appCopyField]="copyField.field" [cipher]="cipher()">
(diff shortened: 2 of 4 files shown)


nikwithak (dismissed):

pamperer562580892423:
@nick-livefront @nikwithak Sorry, I saw this by chance. And my comment probably doesn't change anything here.

But could you please get that feedback to the design team: I understand you want to "debloat" the UI, but seriously: removing that redundancy is a really bad idea. https://uxdesign.cc/designing-redundancies-is-healthy-57dda9944e61

- Users can use different ways to achieve a goal with "redundancy"
- Users don't have to memorize which function is located where with "redundancy" (--> it doesn't make any sense that there are only some functions in the context menu, and you have to remember that other functions are located somewhere else)
- my Prof in ergonomics used to say "everything that requires working memory is bad" --> users should not have to learn "to memorize the UI", but use the copy buttons in whatever way they like... either by clicking the context menu or by using the buttons the right
- another reason: users with large screens have to reach far to the right in the item list for the copy buttons, while they probably click on the left for everything else (items, checkboxes, navigation sidebar) --> that is impractical for many users I guess (and it would be far more convenient to right-click the item row with the mouse on the left and use a copy button in the context menu instead)
- and again: users should have the choice! --> that is also part of a good "user experience"

Instead of removing those buttons in the web vault, they should rather be added back to the desktop app.

nick-livefront:
> @nick-livefront @nikwithak Sorry, I saw this by chance. And my comment probably doesn't change anything here.
> 
> But could you please get that feedback to the design team: I understand you want to "debloat" the UI, but seriously: removing that redundancy is a really bad idea. https://uxdesign.cc/designing-redundancies-is-healthy-57dda9944e61

@pamperer562580892423 I appreciate the feedback and your perspective! I will pass it along to the design team for future consideration. This change had design input before the work began, they are aligned with the consistent experience across the applications without duplicated functionality.

nick-livefront:
@nikwithak QA noticed that the SSH Key copy actions weren't available inline: [3ff3a5a](https://github.com/bitwarden/clients/pull/21023/commits/3ff3a5ad9ed71a802b573b57ac7759a9f9e88edf) + [518a91e](https://github.com/bitwarden/clients/pull/21023/commits/518a91ef90c312f67b33ba995b6282afd821ee2e)

nikwithak (approved):

PR #21024, opened 2026-06-03
[PM-36016] Handle cross-origin iframes in queryDeepSelector

bensbits91:
## 🎟️ Tracking

[PM-36016](https://bitwarden.atlassian.net/browse/PM-36016)

## 📔 Objective

Extend the iframe-aware `>>>` deep-selector traversal from [PM-34497](https://github.com/bitwarden/clients/pull/20401) to support cross-origin iframes, achieving feature parity with the existing heuristic-based autofill.

The same-origin PR introduced the routing infrastructure (routeTargetedFieldsToFrame in the background, applyExternalTargetedFields in content scripts), but cross-origin coverage was incomplete in two ways:

1. Multi-hop chains weren't supported. A selector that crossed two iframe boundaries (e.g., iframe#a >>> iframe#b >>> input#c) succeeded on the first hop and silently dropped on the second, because the destination frame's applyExternalTargetedFields resolved selectors locally without re-routing onward.
2. Frame URL matching was brittle. Routing used strict equality between the iframe's src and the URL reported by chrome.webNavigation. Common normalization differences (trailing slashes, missing components) caused legitimate matches to fail.

This PR addresses both:

- A new shared resolveTargetedFields resolver is used by both getTargetedPageDetails and applyExternalTargetedFields. When a routed selector itself crosses another iframe, the destination frame routes onward via the background — making multi-hop chains terminate correctly regardless of how the boundaries mix same-origin and cross-origin frames.
- The getSubFrameUrlVariations helper (previously a private method on AutofillOverlayContentService, used for sub-frame offset resolution) is extracted into a shared utility and now drives the routing match in routeTargetedFieldsToFrame as well. Routing fails soft and logs when zero or ambiguous matches are encountered, and send failures (e.g. sandboxed iframes that refuse messages) are caught and logged.

Sandboxed iframes are handled implicitly: the destination frame's content script already self-gates via currentlyInSandboxedIframe(), so the routing layer needs no special case — it dispatches, and the receiver decides.

Out of scope (mirrors current heuristic behavior):

- srcdoc iframes — chrome.webNavigation reports all of them as about:srcdoc, which makes URL-keyed routing ambiguous. The existing heuristic sub-frame coordination has the same limitation; deferred to a follow-up.
- about:blank iframes with no src — same disambiguation problem, same deferral.

[PM-36016]: https://bitwarden.atlassian.net/browse/PM-36016?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[PM-34497]: https://bitwarden.atlassian.net/browse/PM-34497?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/browser/src/autofill/background/overlay.background.ts
```diff
@@ -1109,6 +1110,15 @@ export class OverlayBackground implements OverlayBackgroundInterface {
...
    *
+   * Frame matching uses the URL variation set to tolerate normalization
+   * differences (trailing slashes, missing components, etc.) between the
+   * iframe's reported `src` and the URL `webNavigation` reports. Ambiguous
+   * matches (multiple frames matching the same variation) are dropped to
+   * avoid routing to the wrong frame; this mirrors the sub-frame offset
+   * resolution policy. Send failures (e.g. sandboxed iframes that refuse
+   * messages) are logged and otherwise ignored — the destination frame's
+   * content script self-gates on sandbox restrictions.
+   *
```
audreyality: ⛏️ This comment is overly verbose. Some of the parentheticals outline cases that need not be enumerated. Others contain text that is more clear than the phrase they apply to.

apps/browser/src/autofill/utils/url-variations.ts
```diff
@@ -0,0 +1,65 @@
...
+ *   Defaults to the current global location when available, mirroring the
+ *   behavior used by content-script callers.
+ */
+export function getSubFrameUrlVariations(
+  subFrameUrl: string,
+  baseUrl: string | undefined = typeof globalThis !== "undefined"
+    ? globalThis.location?.href
+    : undefined,
+): Set<string> | null {
+  try {
```
audreyality, line 25: 🤔 This `try`/`catch` block applies to the entire function. This leads me to believe it's concealing a problem that might be, instead, fixed. 

🎨 If there's a specific problem it's attempting to capture, limit the catch block to only those lines where you expect an error.

apps/browser/src/autofill/utils/url-variations.ts
```diff
@@ -0,0 +1,65 @@
...
+      url.hostname + pathWithoutTrailingSlashSearchAndHash,
+      url.origin + url.pathname,
+      url.origin + pathAndHash,
+      url.origin + pathAndSearch,
+      url.origin + pathSearchAndHash,
+      url.origin + pathNameWithoutTrailingSlash,
+      url.origin + pathWithoutTrailingSlashAndHash,
+      url.origin + pathWithoutTrailingSlashAndSearch,
+      url.origin + pathWithoutTrailingSlashSearchAndHash,
+    ]);
```
audreyality: 🤔 You're building this set every time that the function is called, only to discard it and rebuilt it the next time that you're matching iframes. Caching the set would avoid recomputing it.

🎨 It is likely that this set is performing character-by-character string comparisons on each of these entries when `Set.has()` is called. It may be more efficient to construct a regular expression containing the variations.

apps/browser/src/autofill/background/overlay.background.ts
```diff
@@ -1126,15 +1136,33 @@ export class OverlayBackground implements OverlayBackgroundInterface {
...
-    if (!targetFrame) {
+    const variations = getSubFrameUrlVariations(iframeSrc);
+    const candidates = variations
+      ? frames.filter((f) => variations.has(f.url))
+      : frames.filter((f) => f.url === iframeSrc);
+
+    if (candidates.length === 0) {
+      this.logService.debug(
+        `[OverlayBackground] No frame matched iframeSrc for targeted field routing: ${iframeSrc}`,
+      );
```
audreyality, line 1142: 👍🏻 Excellent use of logging! These branches would otherwise be hidden, which would make this a pain to debug.

apps/browser/src/autofill/services/collect-autofill-content.service.ts
```diff
@@ -265,77 +270,21 @@ export class CollectAutofillContentService implements CollectAutofillContentServ
...
-    const iframeTargets = new Map<
-      string,
-      { selector: string; fieldType: AutofillTargetingRuleType }[]
-    >();
-
-    for (let formIndex = 0; formIndex < forms.length; formIndex++) {
-      const form = forms[formIndex];
-
+    const targets: ResolveFieldTarget[] = [];
+    for (const form of forms) {
```
audreyality: 📝 Your changes to `getTargettedPageDetails` are going to conflict with #20970. 

👍🏻 I'm encouraged to see that you've isolated some of the complexity I mentioned on the other PR.

apps/browser/src/autofill/services/collect-autofill-content.service.ts
```diff
@@ -265,77 +270,21 @@ export class CollectAutofillContentService implements CollectAutofillContentServ
...
-
-            fields.push(autofillField);
-            this.cacheAutofillFieldElement(fields.length - 1, formFieldElement, autofillField);
-            break;
-          }
-        }
+        targets.push({
+          selectorAlternatives: selectorAlternatives as string[],
+          fieldType: fieldType as AutofillTargetingRuleType,
+        });
```
audreyality: 🎨 Consider reducing this loop to array functions for clarity.

```ts
// type assertion narrows object keys
const entries = Object.entries(form.fields) as Array<[AutofillTargetingRuleType, string[]]>;
const fields = targets
  .filter(([, alternatives]) => !alternatives.length)
  .map(([fieldType, selectorAlternatives]) => ({ fieldType, selectorAlternatives }));
targets.push(...fields);
```

apps/browser/src/autofill/services/collect-autofill-content.service.ts
```diff
@@ -413,6 +366,100 @@ export class CollectAutofillContentService implements CollectAutofillContentServ
...
+  } {
+    const localFields: AutofillField[] = [];
+    // Accumulates targets that live inside iframes, keyed by the iframe's URL.
+    // These are routed to the iframe's own content script instead of being
+    // collected here, so the existing sub-frame offset infrastructure handles
+    // their positioning correctly.
+    const iframeTargets = new Map<
+      string,
+      { selector: string; fieldType: AutofillTargetingRuleType }[]
+    >();
```
audreyality, line 385: ⚠️ This variable hides the `iframeTargets` argument. That has consequences for the test-and-set operations on line 413, which will always be true. It also means that the return value will always have a single entry.

❓ Did you intend `const iframeTargets = new Map(iframeTargets)`?

apps/browser/src/autofill/services/collect-autofill-content.service.ts
```diff
@@ -413,6 +366,100 @@ export class CollectAutofillContentService implements CollectAutofillContentServ
...
+   * or re-route onward, enabling multi-hop chains.
+   */
+  private routeIframeTargets(
+    iframeTargets: Map<string, { selector: string; fieldType: AutofillTargetingRuleType }[]>,
+  ): void {
+    for (const [iframeSrc, iframeTargetedFields] of iframeTargets) {
+      void this.sendExtensionMessage("routeTargetedFieldsToFrame", {
+        iframeSrc,
+        iframeTargetedFields,
+      });
```
audreyality, line 458: ⚠️ This silently drops errors.


```ts
      void this.sendExtensionMessage("routeTargetedFieldsToFrame", {
        iframeSrc,
        iframeTargetedFields,
      }).catch((e: unknown) => ...);
```

audreyality (changes requested):
📝 The requested changes have ⚠️ symbols.

👍🏻 I'm a fan of the refactoring being done here!

apps/browser/src/autofill/services/collect-autofill-content.service.ts
bensbits91: Hmmm... The `iframeTargets` on line 382 is a property of the return type annotation, not a function argument. Line 389 is the local declaration, not shadowing. The `has`/`set`/`get` pattern at line 413 accumulates entries across iterations ("ensure-key" pattern), and multi-iframe scenarios produce multi-entry returns. Our existing tests for "multiple iframe targets coalesce" and "routes via routeTargetedFieldsToFrame" both exercise paths that depend on this accumulation. 

I don't think `const iframeTargets = new Map(iframeTargets)` would work here because there's no `iframeTargets` in scope to copy from at that point.

Happy to discuss further if I'm missing something.

apps/browser/src/autofill/services/collect-autofill-content.service.ts
bensbits91: Good `catch`, thanks 😃 

Added a `console.warn` in https://github.com/bitwarden/clients/pull/21024/commits/e4160713346a013bc452eabf976935d422ac62e5. Do you think that is sufficient?

apps/browser/src/autofill/utils/url-variations.ts
bensbits91: Thanks for raising these. I think we should leave this as-is but please push back if I'm missing anything.

Caching: the function is called a bounded number of times per page (once per iframe), and each call should take less than a millisecond. I think caching would add complexity (cache key, lifetime, invalidation on dynamic-URL iframes) for an unmeasurable speedup, and inflates the service worker's memory on long-lived sessions.

Set vs regex: V8's `Set<string>` lookups are hash-based, not character-by-character — `Set.has()` is O(1) amortized for fixed-string sets. A regex alternation like `/^(a|b|c|...)$/` runs a state-machine traversal, which is typically slower than a hash lookup for membership testing on a fixed set.

apps/browser/src/autofill/background/overlay.background.ts
bensbits91: You're right, yuck. I cleaned up this comment in https://github.com/bitwarden/clients/pull/21024/changes/505fa9cf81a459f217056f8963366b748be287d3. Look ok?

apps/browser/src/autofill/utils/url-variations.ts
bensbits91: Addressed in https://github.com/bitwarden/clients/pull/21024/changes/9e2af01f9f05b81fcc1da6924fbb67f6b3f85dbc. Better?

apps/browser/src/autofill/services/collect-autofill-content.service.ts
bensbits91: 🫶 Addressed in https://github.com/bitwarden/clients/pull/21024/changes/b1d93969bffbfa857f2affbfa0adc76ba4d36535. Look ok?

apps/browser/src/autofill/services/collect-autofill-content.service.ts
bensbits91: You're right, significant conflict.

My PR refactors `getTargetedPageDetails` to extract a unified resolver (`resolveTargetedFields`) that's also called from `applyExternalTargetedFields`, which enables recursive routing across multi-hop iframe chains.

PR #20970 reworks the same method for different purposes — container resolution, form-OPID generation, per-type match caps, and dedupe-by-element. The other files we both touch (overlay.background.ts, autofill-overlay-content.service.ts) don't conflict, right?

@jprusik how should we resolve this? Merge your changes first and I'll refactor mine? Vice-versa? Or some kind of unified change that incorporates both before merging either?

apps/browser/src/autofill/utils/url-variations.ts
audreyality: 💭 This isn't blocking, but I'd like to offer food for thought.

1. The spec does not guarantee O(1) performance, and we're not only concerned with V8 optimizations. There are 2 other browsers we also need to perform well on. 
2. You have to iterate 25 strings to compute their hashcodes each time you call this method, and you also have to construct those strings. 
3. The amortized cost of construction doesn't mean as much when you discard it, unless we're comparing against a very large number of iframes at its callsites.

Since this logic is not net-new, I'm going to move on. My main concern is that the fixed costs to construct the set don't pay for the marginal gains of the set operation.

apps/browser/src/autofill/services/collect-autofill-content.service.ts
audreyality: 🤦🏻‍♀️ You're right; I misread the code.

⛏️ The return type is difficult to read. Consider reducing its complexity using a type alias.

```ts
type IframeTarget = { selector: string; fieldType: AutofillTargetingRuleType };

// ...
const iframeTargets = new Map<string, IframeTarget[]>();
```

audreyality (commented):
👍🏻 Significantly improved. Just one nit; I'll approve once @jprusik comments on how to resolve the upcoming conflict.

apps/browser/src/autofill/services/collect-autofill-content.service.ts
jprusik: I'm inclined to have your changes go in first; https://github.com/bitwarden/clients/pull/20970 is mostly a collection of small* gap fixes which should have been broken out anyway.

jprusik:
> 👍🏻 Significantly improved. Just one nit; I'll approve once @jprusik comments on how to resolve the upcoming conflict.

I haven't fully reviewed this PR, but if you're otherwise ready to approve and merge, let's get this one in and I'll resolve my conflicts after.

audreyality (approved):

PR #21028, opened 2026-06-03
SSH Agent support of ECDSA key type

neuronull:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-37917

## 📔 Objective

We don't have support yet across the entire product for using ECDSA keys in the Desktop app. This change only introduces support to the SSH Agent itself for handling this key type.

Reviewer notes:
- most of the changes are unit and integration tests
- there was a small change to the `from_private_key_pem` , where I noticed we were hard coding strings for the algorithm which we could just take from the key itself, so I addressed that and also added unit tests for that.

## 📸 Manual testing

None. We can't utilize the new key support yet until later functions. The new integration tests do add coverage though.


neuronull:
The failure in the windows beta job step looks unrelated to these change

apps/desktop/desktop_native/ssh_agent/src/crypto/mod.rs
```diff
@@ -297,4 +306,103 @@ mod tests {
...
+        let ssh_key =
+            ssh_key::PrivateKey::new(ssh_key::private::KeypairData::Ecdsa(keypair), "").unwrap();
+
+        let private_key = PrivateKey::try_from(ssh_key).unwrap();
+        assert!(matches!(private_key, PrivateKey::Ecdsa(_)));
+    }
+
+    #[test]
+    fn test_privatekey_from_ecdsa_p384() {
+        let keypair = EcdsaKeypair::random(&mut OsRng, EcdsaCurve::NistP384).unwrap();
```
quexten: non-blocking thought: Would be nice to use something like https://docs.rs/test-case/latest/test_case/ to reduce duplicate tests

quexten (dismissed):

apps/desktop/desktop_native/ssh_agent/src/crypto/mod.rs
neuronull: Good callout about the duplicate code.

I haven't seen that crate before, it's neat.

Historically I've solved this by just creating helper functions in the test module (this has the added benefit of being able to put such helper functions in a parent module so that they can be re-used, which I think would not be possible with `test_case`).

I will push up a commit to reduce the duplicate logic in the tests using the helper fn method.

If we do want to propose shifting to `test_case` I don't think it's a bad idea but I'd suggest it could be done in a separate task; and tackle it for existing places that use helper functions as well.

neuronull: https://github.com/bitwarden/clients/pull/21028/commits/3642c574a311543679f16604add5c294fc726663

quexten (dismissed):

apps/desktop/desktop_native/ssh_agent/src/crypto/mod.rs
```diff
@@ -297,4 +306,92 @@ mod tests {
...
+    }
+
+    fn ecdsa_private_key_from_ssh_key(curve: EcdsaCurve) -> PrivateKey {
+        let ssh_key = ssh_key::PrivateKey::new(
+            ssh_key::private::KeypairData::Ecdsa(ecdsa_keypair(curve)),
+            "",
+        )
+        .unwrap();
+        PrivateKey::try_from(ssh_key).unwrap()
+    }
```
coltonhurst: Non-blocking nit: Any reason these are in the middle of the tests, instead of at the top of the test block?

coltonhurst (dismissed):
Did a quick review, approved. Only one non-blocking question/nit.

apps/desktop/desktop_native/ssh_agent/src/crypto/mod.rs
neuronull: oversight. fixed in https://github.com/bitwarden/clients/pull/21028/commits/84e84d8403550b71999395f46420a850c6534df2

coltonhurst (approved):

PR #21029, opened 2026-06-03
[PM-32089] Sync after collection deletion

nick-livefront:
## 🎟️ Tracking

[PM-32089](https://bitwarden.atlassian.net/browse/PM-32089)

## 📔 Objective

When a collection is deleted the underlying ciphers are disassociated with that collection, this can cause unexpected behavior and stale data for those ciphers until a sync as occurred. In the context of this bug that presented itself in an item still appearing in the users trash when it should be in the organization's trash.

Notes: 
- A full sync is a heavy call to make. I didn't see a smoother way to ensuring that the client had fresh data without it because:
  -  There currently isn't any dependency between the `CollectionService` and `CipherService`.
  - Updating only the cipher state client side would open up for differences to creep in between the server state and client state if not maintained. 
- I double checked the server side logic and the ciphers are properly getting unassigned when a collection is deleted regardless of their soft delete status.

## 📸 Screenshots

|Cipher only in a single collection|Cipher in multiple collections|
|-|-|
|<video src="https://github.com/user-attachments/assets/3474764d-0a70-4b0b-aef3-409f22d583b4" />|<video src="https://github.com/user-attachments/assets/ce9c5539-cd24-4c8d-bb27-c562062d225a" />|

[PM-32089]: https://bitwarden.atlassian.net/browse/PM-32089?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/web/src/app/vault/individual-vault/bulk-action-dialogs/bulk-delete-dialog/bulk-delete-dialog.component.ts
+++ b/apps/web/src/app/vault/individual-vault/bulk-action-dialogs/bulk-delete-dialog/bulk-delete-dialog.component.ts
@@ -10,6 +10,7 @@ import { Organization } from "@bitwarden/common/admin-console/models/domain/orga
 import { AccountService } from "@bitwarden/common/auth/abstractions/account.service";
 import { getUserId } from "@bitwarden/common/auth/services/account.service";
 import { I18nService } from "@bitwarden/common/platform/abstractions/i18n.service";
+import { SyncService } from "@bitwarden/common/platform/sync";
 import { CollectionId } from "@bitwarden/common/types/guid";
 import { CipherService } from "@bitwarden/common/vault/abstractions/cipher.service";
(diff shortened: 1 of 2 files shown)


apps/web/src/app/vault/individual-vault/bulk-action-dialogs/bulk-delete-dialog/bulk-delete-dialog.component.ts
```diff
@@ -160,7 +162,13 @@ export class BulkDeleteDialogComponent {
     }
   }
 
-  private async deleteCollections(): Promise<any> {
+  private async deleteCollections(): Promise<void> {
```
nick-livefront, line 165: The switch from `any` to `void` here was intentional. There is no response from these endpoints and if there was it wasn't used.

jengstrom-bw (approved):

PR #21040, opened 2026-06-04
[PM-38547] Fix missing Flatpak desktop icon

quexten:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-38547

## 📔 Objective

The Flatpak build showed no app icon because the manifest never installed the icon into the hicolor theme path. The desktop entry references `Icon=com.bitwarden.desktop`, but that name could not resolve. This installs the bundled PNG icons into `/app/share/icons/hicolor/<size>/apps/` so the icon resolves and appears in the app grid and taskbar.

Additionally, on X11 the icon is not showing because of the `StartupWMClass` which was wrong. This fixes that too.


dani-garcia (dismissed):

quexten:
I tested this to work, but note that this is still broken with the default X11 setting. On wayland (with USE_X11=false) it works.

quexten:
^ Added a fix for X11

quexten:
Flathub is also experiencing this because the startupwmclass is wrong..

dani-garcia (approved):

PR #21044, opened 2026-06-04
[PM-38565] Replace tray settings with "Keep Bitwarden running in the background"

### LINKED ISSUE #20923: [FLATPAK] Start Minimized option is gone
### Steps To Reproduce

1. Go to 'File'
2. Click on 'Settings'
3. Scroll down to 'App Settings'
The option for start application minimized is missing

### Expected Result

The option to start minimized would be here and function.

### Actual Result

The option is now missing and the application launches at startup everytime taking up half of the desktop

### Screenshots or Videos

Image

### Additional Context

It was functioning as intended before This afternoons update

### Operating System

Linux

### Operating System Version

Fedora 44

### Installation method

Flatpak

### Build Version

2026.5.0

### Issue Tracking Info

### LINKED ISSUE #21074: Missing option "Start to menu bar" in Mac App Store Edition 2026.5.0
### Steps To Reproduce

Since the new version 2026.5.0 was installed (App Store version), the Bitwarden App starts in foreground with the main windows opened. That's not the behavior I want, especially on system startup. So I checked the settings and found (or should I say did not find) the option "Start to menu bar" missing.

### Expected Result

I expect an option in the settings like "Start to menu bar" to prevent the main window from opening on (system) startup.

### Actual Result

There is no option named like this and I have no way to prevent the main window from opening on startup

### Screenshots or Videos

Image

### Additional Context

_No response_

### Operating System

macOS

### Operating System Version

26.5.1

### Installation method

Mac App Store

### Build Version

2026.5.0 (62263)

### Issue Tracking Info

- [ ] I understand that work is tracked outside of GitHub. A PR will be linked to this issue should one be opened to address it, but Bitwarden doesn't use fields like "assigned", "milestone", or "project" to track progress.

quexten:
## 🎟️ Tracking

https://bitwarden.atlassian.net/wiki/spaces/~712020c1551f20895b498884b767b55a92b68d/pages/edit-v2/3029368834
https://bitwarden.atlassian.net/browse/PM-38565

## 📔 Objective
Consolidate the settings around minimization and tray behavior to the "Keep Bitwarden running in the background" setting.

### Ubuntu-Gnome [Linux, Tray]

https://github.com/user-attachments/assets/b8d85f77-c031-4105-972f-a6f878073c7b

### Fedora-Gnome [Linux, Gnome Background]

https://github.com/user-attachments/assets/08392c77-4a41-4488-b4f3-42c0c2de2dac

### MacOS

https://github.com/user-attachments/assets/e3ab3c83-bf21-4d12-be87-40a6c9822240

### Windows

https://github.com/user-attachments/assets/73251ce1-e7f4-43a1-b05e-bd197ac45357

## 📄 Copy of Confluence Document
(Since the confluence document is internal, I've made a copy here so that the decisions made here are public).

We want our app to behave in the way that is standard to the operating system, in a way that users expect. At the same time we want to keep cognitive and configuration complexity low, to simplify the application for our users, to not make it possible to configure your application in a way that is “broken”, and to make reasoning about bugs easier, because there is fewer states the application can be in.

### The App Running State

We want the app to be running in the background in most cases, to enable SSH agent, shared unlock, biometric unlock, autotype and magnify. It should be always clear to the user when the app is running, and it should be possible for the user to close the app.

There are a few states the app can be in:

- Foreground
  - The window is visible, the tray icon is present
- Minimized
  - The window is not visible, the window is in the taskbar, the tray icon is present
- Tray (closed, on macOS)
  - The window is not visible, the tray icon is present
- Closed (quit, on macOS)
  - The tray icon is not present, the application is not running

We should treat these as discrete states. Previously (currently), there were more combinations possible than the above, which we will remove.

### Autostart

To ensure the app is running in the background in most cases, it needs to be auto-started with the system. Autostart always happens to the “Tray” state.

### The Tray

The tray (menubar (mac), background list (gnome)) indicates that the app is running. We should always display the tray icon if the app is running, and a tray is available. The tray is available on all environments except Gnome. On gnome, the background app list is used instead. We do not have control over the background app list, and it will show automatically when the window is hidden.

Clicking the tray icon should always foreground the application in the same way that clicking the desktop icon does, a minimized window gets unminimized, a tray application gets un-trayed. It should not have a behavior to toggle foreground state.

Q: Why not make tray configurable

Given that the application is running in the background, there must be a way to access it (bring it to the foreground, quit it). Doing so consistently via the tray creates a predictable experience and is what other apps that need to run in the background do.

The tray becomes the central indicator that the app is running.

### Closing Windows

Other apps treat closing the window as closing the window, not closing the application. For a user, it is easy to otherwise accidentally completely close the app when they do not intend to. Further, on some systems (gnome), you only have the “close window” option, which in general apps treat as close to tray.

Closing the app is instead done via the tray or via the menu option “quit Bitwarden”.

This is consistent with how other applications behave.

Q: Why is closing the window not mapped to closing the application? 
In general, apps that users expect to run in the background do not behave in this way. The close action closes the window, not the application. We have countless app examples of this, and closing the application would lead to users accidentally closing the application, not being able to use the features that require the desktop app running.

### Minimizing Windows

On systems that support minimizing, the app may currently hide to tray and remove the minimized window. This is unexpected behavior that should be removed. Minimizing always should just minimize on systems that support minimizing.

After minimizing we should always be in the “Minimized” state.

### Clicking The Desktop Icon

Clicking the desktop icon should always bring the app to the foreground, never start the app minimized or hidden. When doing so, for the minimized state, the window should be unminimized. For the tray state, the app window should be recreated. For the closed state, the app should start.

After clicking the desktop icon, we should always be in the Foreground state.

### Running The App From The Terminal

Same as clicking the desktop icon.

### New Settings List

With the above behavior, there is exactly two settings, available on all apps:
- Enable auto-start
- Keep Bitwarden open in the background
  - "When exiting the Bitwarden window, Bitwarden remains accessible via the [menubar / tray]. This option is required for unlock via TouchID and SSH agent."


dereknance (dismissed):
This is a lovely change! Since I took so long to review this (sorry about that), there are now conflicts to resolve 😬

dereknance (approved):

Kommynct:
I can't start bitwarden from the tray, how do I ensure it starts from the tray and not with a window after this change?

PR #21050, opened 2026-06-04
Add form validation documentation to component library

willmartian:
## 🎟️ Tracking

N/A — documentation for the component library's form validation patterns.

## 📔 Objective

Adds a new **Component Library → Form → Validation** Storybook page (`validation.mdx`) documenting how form validation works across our Angular clients, with a runnable example per section:

- **Rendering errors** — how a `ValidationErrors` object reaches the user via `bit-error` (inline) and `bit-error-summary` (count), the `{ key: { message } }` fallback contract, and where toasts fit.
- **Synchronous validators** — built-in/shared validators, custom `ValidatorFn` factories, and cross-field validators.
- **Asynchronous validators** — `AsyncValidatorFn` for any awaited check (server *or* client-side Promise/Observable services), Observable cancellation, and the `pending` state.
- **Validation timing** — `updateOn` and submit-time validation.
- **Server reconciliation at submit** — mapping attributable failures inline via `setErrors`, with the danger toast as the fallback for unattributable errors.

Supporting changes:

- Adds `validation.stories.ts` with the examples backing the page.
- Moves the shared `Validations` story out of `form.stories.ts` into `validation.stories.ts` (single definition) and repoints `forms.mdx` at its new location.

No production code changes — docs and Storybook stories only.

### DIFF
--- a/libs/components/src/form-field/validation.mdx
+++ b/libs/components/src/form-field/validation.mdx
@@ -1,0 +1,222 @@
+import { Meta, Canvas } from "@storybook/addon-docs/blocks";
+
+import * as stories from "./validation.stories";
+
+<Meta title="Component Library/Form/Validation" />
+
+# Form Validation
+
+This page covers how form validation works across our Angular clients: how an error becomes visible,
+how to express a rule, when it runs, and what happens when the server has the final say.
+
+## Principles
+
+**Validation happens on both the client and the server.** Some checks can only run on one side: the
+server can't inspect end-to-end-encrypted Vault Data, while a check like "is this value already
+taken?" can only be answered by the server. Expect both — and expect
+[server errors at submit](#server-reconciliation-at-submit) as a normal case, not a gap.
+
+**Most errors share one path.** Built-in validators, custom async checks, and field-attributable
+server errors all end up as a [`ValidationErrors`][validation-errors] object on a control and render
+inline the same way. Only errors that can't be tied to a field fall back to a toast.
+
+## Rendering errors
+
+Errors are shown two ways. **Inline, contextual errors are preferred** — they sit next to the field
+they describe. **Toasts** are the fallback for errors that can't be tied to a specific field.
+
+Inline errors come from a control's `ValidationErrors` — a map of error key to value. `bit-error`
+renders the first error on the control:
+
+- known keys (`required`, `email`, `minlength`, `forbiddenCharacters`, `trim`, …) get a translated,
+  interpolated message;
+- anything else falls back to the error value's `message` property:
+
+```ts
+return { emailTaken: { message: "This email address is already in use." } };
+```
+
+Two components render control errors:
+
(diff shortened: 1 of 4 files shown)


libs/components/src/form-field/validation.mdx
```diff
@@ -0,0 +1,222 @@
...
+    validators: [Validators.required, Validators.email],
+    asyncValidators: [emailNotTakenValidator()],
+  },
+],
+```
+
+- `"change"` (default) — every keystroke. Fine for cheap sync checks; wrong for async or expensive
+  ones.
+- `"blur"` — when the field loses focus. The right default for async/expensive validation.
+- `"submit"` — only on submit. Used by `trimValidator`, which mutates the value.
```
eliykat, line 164: This could be more opinionated. e.g. I don't think we use `change` anywhere. As far as I know, `blur` is our default choice, and we could recommend `submit` for complex server-side validation that we don't want to expose separately (as you recommended).

libs/components/src/form-field/validation.mdx
```diff
@@ -0,0 +1,222 @@
...
+
+Validation must also run on submit, because the user can submit without ever blurring a field. Call
+`formGroup.markAllAsTouched()` in your submit handler so untouched-but-invalid fields reveal their
```
eliykat, line 167: (non-blocking, future suggestion only) Could be handled automatically at the CL level? It's easy to forget and every form needs it.

libs/components/src/form-field/validation.mdx
```diff
@@ -0,0 +1,222 @@
...
+```ts
+submit = async () => {
+  if (this.formGroup.invalid) {
+    return;
+  }
+  try {
+    await this.api.createMember(this.formGroup.value);
+  } catch (e) {
+    // Attributable to a field → render inline like any other error.
+    if (isEmailTakenError(e)) {
```
eliykat, line 191: (non-blocking) Once we get a bit of consensus on a solution here (e.g. error codes) this can be updated with a bit more detail.

libs/components/src/form-field/validation.mdx
```diff
@@ -0,0 +1,222 @@
...
+};
+```
+
+**Unattributable failures** — infrastructure errors, bugs, anything you can't pin to a field. There
+is nothing to render inline, so the danger toast is the honest fallback. By default
+[`bitSubmit`][async-forms] routes any uncaught error to `ValidationService.showError`, which shows
+that toast — so rethrowing (or simply not catching) already gives you the right behavior.
+
+This is the answer to "no toasts ever": attributable failures move inline, and the toast shrinks to
+the genuinely-unexpected tail it was always meant for.
```
eliykat, line 209: AI getting caught up on my Slack message wording - this should be cleaned up as it looks odd out of context.

eliykat (commented):
Very helpful guidance, thank you. Minor comments only, overall LGTM.

PR #21052, opened 2026-06-04
[PM-38387] Introduce KeePass KDBX importer 

harr1424:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-38387

## 📔 Objective

This PR introduces a KeePass KDBX importer. 

[Companion SDK PR](https://github.com/bitwarden/sdk-internal/pull/1168)

## 📸 Screenshots

Screenshot 2026-06-10 at 16 43 03

Screenshot 2026-06-10 at 16 43 11

Screenshot 2026-06-10 at 16 43 46

Screenshot 2026-06-10 at 16 43 59

Screenshot 2026-06-10 at 16 44 28

Screenshot 2026-06-10 at 16 53 15


package.json
```diff
@@ -198,6 +199,7 @@
     "inquirer": "8.2.6",
     "jsdom": "26.1.0",
     "jszip": "3.10.1",
+    "kdbxweb": "2.1.1",
```
quexten: We need a dependency review from appsec

package.json
```diff
@@ -184,6 +184,7 @@
     "@microsoft/signalr": "10.0.0",
     "@microsoft/signalr-protocol-msgpack": "10.0.0",
     "@ng-select/ng-select": "21.8.2",
+    "@noble/hashes": "1.4.0",
```
quexten: We need a dependency review from appsec

libs/importer/src/importers/keepass-kdbx-importer.spec.ts
```diff
@@ -0,0 +1,199 @@
...
+
+import { I18nService } from "@bitwarden/common/platform/abstractions/i18n.service";
+import { Utils } from "@bitwarden/common/platform/misc/utils";
+import { FieldType } from "@bitwarden/common/vault/enums";
+
+import { KdbxCredentials, KeePassKdbxImporter } from "./keepass-kdbx-importer";
+
+// Register an Argon2 implementation so kdbxweb can save KDBX4 fixtures in this test (the importer
+// registers its own when it loads). Mirrors the implementation in the importer.
+kdbxweb.CryptoEngine.setArgon2Impl(
```
quexten: For new importers, a question. Why not just write it in the SDK? We have better tooling available, all crypto algorithms, and better parsers. Example: https://docs.rs/keepass/latest/keepass/

This PR currently is adding a new argon2 implementation entirely.

quexten (commented):
@harr1424 I think tools should invest some time to investigate their SDK onboarding strategy, and consider writing new importers in rust. There is a better ecosystem there around crypto, and also it is scalable to moblie.

package.json
harr1424: Yes absolutely, I have opened this as a draft PR, it still needs a lot of work.

package.json
harr1424: Yes absolutely, I have opened this as a draft PR, it still needs a lot of work.

libs/importer/src/importers/keepass-kdbx-importer.spec.ts
harr1424: I don't disagree, but we have been asked to deliver this as soon as possible, in the current sprint ideally. 

In hindsight, I should have explored what options were available in the SDK. The team discussed this during sync this morning, I've only done some cursory exploration, but all import functionality seems CXF centric, which is a direction I understand we are moving towards. 

I've reached out to ask if it would be possible to spend more time investigating the approach you are suggesting.

harr1424:
@claude please give this a review. Disregard the failing typecheck job for now, the proper version of the SDK has not yet merged and cannot be pinned.

harr1424:
@claude please re-review, taking into consideration many CI checks will fail until the companion SDK PR is merged

apps/cli/src/tools/import.command.ts
```diff
@@ -23,6 +29,8 @@ export class ImportCommand {
     private organizationService: OrganizationService,
     private syncService: SyncService,
     private accountService: AccountService,
+    private logService: LogService,
+    private i18nService: I18nService,
```
bmbitwarden, line 33: could use new inject mechanism for dependency inject

private readonly i18nService = inject(I18nService);

apps/cli/src/utils.ts
```diff
@@ -219,6 +232,7 @@ export class CliUtils {
     password: string,
     options: { passwordFile?: string; passwordEnv?: string },
     logService?: LogService,
+    promptMessage = "Master password:",
```
bmbitwarden, line 235: goes in translation file?

libs/importer/src/sdk/importers/kdbx-sdk-importer.ts
```diff
@@ -0,0 +1,59 @@
...
+      case "KdbxWrongCredentials":
+        return "invalidFilePassword";
+      case "KdbxInvalidFormat":
+        return "kdbxWrongFileType";
+      case "KdbxCorruptOrUnsupported":
+        return "kdbxCorruptOrOutdated";
+      case "KdbxFileTooLarge":
+        return "kdbxFileTooLarge";
+      default:
+        return undefined;
```
bmbitwarden, line 55: maybe throw here to enunciate we have an untrcked error type

bmbitwarden (dismissed):

apps/cli/src/tools/import.command.ts
harr1424: This file, and other files inside `apps/cli/...` contain code for the CLI, which isn't an Angular app. I don't think the proposed injection is possible? Since this would be managed by Angular DI.

apps/cli/src/utils.ts
harr1424: I agree we should probably stop doing this, but right now every interactive command I have seen in the CLI is using hardcoded English string: 

https://github.com/bitwarden/clients/blob/4a196a533bf51739280c3d50e95e40d4b2e8be9a/apps/cli/src/auth/commands/login.command.ts#L149

https://github.com/bitwarden/clients/blob/afc45ee0c8fc823301bb361b0dcac581eb0aff0c/apps/cli/src/tools/export.command.ts#L127

https://github.com/bitwarden/clients/blob/ccf047fc63a8c6cc499801d495402126fbacccd0/apps/cli/src/tools/send/commands/receive.command.ts#L301

If we didn't have changes planned for the CLI, I would propose a tech-debt ticket to shift our approach to using `i18n`, but I think that effort would be invalidated in the coming months.

libs/importer/src/sdk/importers/kdbx-sdk-importer.ts
harr1424: I don't think this is a good idea because `default` here covers the non-KDBX `ImportError` variants (NotAuthenticated, Api, BitwardenCrypto, Export) which are all expected errors. `undefined` means "fallback to the generic message" per `sdkErrorMessageKey`'s contract. 

I think throwing on those routine errors would mask the original error. 

Please see https://github.com/bitwarden/clients/blob/0add389b36ca8446381aacae847444b7ebf2fbdc/libs/importer/src/services/import.service.abstraction.ts#L38

and https://github.com/bitwarden/sdk-internal/pull/1168

and if you still think we should throw, let me know.

itsadrago (dismissed):

mcamirault (dismissed):

itsadrago (approved):

PR #21054, opened 2026-06-04
[PM-37768] VULN - webauthn honor permissions policy

bensbits91:
## 🎟️ Tracking

[PM-37768](https://bitwarden.atlassian.net/browse/PM-37768) related to [VULN-582](https://bitwarden.atlassian.net/browse/VULN-582)

[PM-37769](https://bitwarden.atlassian.net/browse/PM-37769) related to [VULN-398](https://bitwarden.atlassian.net/browse/VULN-398)

## 📔 Objective

The Bitwarden browser extension overrides `navigator.credentials.{create,get}` via an injected page script (`apps/browser/src/autofill/fido2/content/fido2-page-script.ts`). Until now, those overrides routed every `publicKey` request through the messenger without consulting the document's Permissions Policy. This let in-page script summon a Bitwarden passkey ceremony in contexts the browser's own WebAuthn API would have rejected with `NotAllowedError`:

- **VULN-582** — pages that set `Permissions-Policy: publickey-credentials-get=()` (or `publickey-credentials-create=()`) intend to forbid WebAuthn in that context, often as defense-in-depth against XSS or supply-chain compromise. The extension ignored the header.
- **VULN-398** — cross-origin iframes are denied WebAuthn by default; the spec requires the parent to explicitly delegate via `allow="publickey-credentials-get"` (or `…-create`). The extension routed regardless of delegation.

### Fix

The gate runs inside the content script's **isolated world** (in `respondToCredentialRequest`, the chokepoint for both `create` and `get`) — not in the page world where attacker JS can shadow `document.featurePolicy` / `top` / `self` to defeat the check. The isolated world has its own DOM-wrapper copies that page script cannot tamper with, so reads of `document.permissionsPolicy` / `featurePolicy` and `self === top` are authoritative.

The check prefers `document.permissionsPolicy` (per the standardized WICG spec) and falls back to `document.featurePolicy`. When neither API is available, a defense-in-depth fallback denies WebAuthn in cross-origin iframes — matching the spec default allowlist (`self`) for `publickey-credentials-*`. On denial, the gate rejects with a `NotAllowedError` `DOMException`. The page-script side rehydrates the DOMException after it crosses the messenger boundary (which strips the prototype during JSON serialization), so RPs see `instanceof DOMException === true` as they would with a native browser denial.

Both VULNs collapse to the same gate because the default allowlist is `self` — the policy returns `false` for cross-origin iframes that lack the delegating `allow` attribute (VULN-398) and for any document where the header explicitly denies (VULN-582).

### Cross-browser API surface

Verified directly from each engine's main branch:

| Browser | `document.permissionsPolicy` | `document.featurePolicy` |
|---|---|---|
| Chromium (Chrome, Edge) | not exposed | exposed |
| Firefox (default) | not exposed | gated behind `dom.security.featurePolicy.webidl.enabled`, default `false` |
| Safari | not exposed | not exposed |

Sources: Firefox [`Document.webidl`](https://hg.mozilla.org/mozilla-central/raw-file/tip/dom/webidl/Document.webidl) and [`StaticPrefList.yaml`](https://hg.mozilla.org/mozilla-central/raw-file/tip/modules/libpref/init/StaticPrefList.yaml); Chromium [`document.idl`](https://raw.githubusercontent.com/chromium/chromium/main/third_party/blink/renderer/core/dom/document.idl).

No shipping browser exposes `document.permissionsPolicy` today; the gate checks it first as forward-compat for the standardized name. In default-config Firefox and in Safari, neither API is available, so the cross-origin iframe fallback engages. Firefox does enforce these directives natively at the WebAuthn entry point ([`CredentialsContainer.cpp`](https://hg.mozilla.org/mozilla-central/raw-file/tip/dom/credentialmanagement/CredentialsContainer.cpp), via `FeaturePolicyUtils::IsFeatureAllowed`), but our injected override bypasses the native path and has no equivalent JS API to query.

### Residual limitations

- **VULN-582 on default-config Firefox and Safari**: a top-level page emitting `Permissions-Policy: publickey-credentials-{get,create}=()` still surfaces the Bitwarden ceremony, because we have no JS API to read the header on those browsers. The attack still requires in-page script execution (XSS or supply-chain), so the residual surface is the same as before this PR for that attacker model — but the residual is real.
- **Over-rejection trade-off in the fallback**: on Firefox/Safari, cross-origin iframes that legitimately received `allow=publickey-credentials-*` from their parent are refused, because we can't introspect the `allow` attribute without the policy API. Chrome is unaffected — the `featurePolicy` API answers authoritatively.

Both will be closed by a follow-up that reads the `Permissions-Policy` response header server-side; see [Follow-up](#follow-up) below.

### Implementation notes

- The gate runs in the content script's isolated world, **not** in the page script's main world. The page world is where attacker JS executes (XSS, supply chain) — the precondition for these VULNs — so a gate there would be trivially defeated by shadowing `document.featurePolicy` / `top` / `self`. The isolated world is the only place this check is meaningful.
- The check happens at request-handling time, not at script-injection time, so the timing matches what the spec expects.
- When the gate denies, no `sendExtensionMessage` is dispatched — denied origins generate no background traffic and no inline UI, mirroring native browser behavior.
- Error text matches Chromium's verbatim so callers that parse the message see no difference. The DOMException is rehydrated on the page side after messenger serialization so `instanceof DOMException` and `.name` both match what the native API would throw.

### Unit tests

In `apps/browser/src/autofill/fido2/content/fido2-content-script.spec.ts`:

- `Permissions Policy enforcement` — `create` denied, `get` denied, allowed path, `permissionsPolicy` precedence over `featurePolicy`, `permissionsPolicy` standalone.
- `Permissions Policy enforcement — defense-in-depth fallback` — denies in cross-origin iframe with API absent, denies when `top.location` throws `SecurityError`, allows at top level when API absent.

In `apps/browser/src/autofill/fido2/content/fido2-page-script.webauthn-supported.spec.ts`:

- `DOMException rehydration across the messenger boundary` — rehydrates on `create`, rehydrates on `get`, passes through rejections with other names.

Full `apps/browser/src/autofill/fido2` suite: **77 tests pass** (11 new).

### Manual verification

Verified against seven test pages in the test-the-web repo. Matrix below — "before" = `main`, "after" = this branch. NAE = `NotAllowedError` returned to page, no Bitwarden UI. sOWA = `Invalid 'sameOriginWithAncestors' value` (see [Known related limitation](#known-related-limitation-not-in-scope) below).

| Test page | Action | main (Chrome) | this branch (Chrome) | this branch (Firefox) |
|---|---|---|---|---|
| `permissions-policy-unrestricted` | Register | prompt → OK | prompt → OK ✓ | prompt → OK ✓ |
| `permissions-policy-unrestricted` | Authenticate | prompt → OK | prompt → OK ✓ | prompt → OK ✓ |
| `permissions-policy-blocked` | Register | prompt (BUG) | NAE ✓ | **prompt (residual)** |
| `permissions-policy-blocked` | Authenticate | prompt (BUG) | NAE ✓ | **prompt (residual)** |
| `permissions-policy-get-blocked` | Register | prompt → OK | prompt → OK ✓ | prompt → OK ✓ |
| `permissions-policy-get-blocked` | Authenticate | prompt (BUG) | NAE ✓ | **prompt (residual)** |
| `permissions-policy-create-blocked` | Register | prompt (BUG) | NAE ✓ | **prompt (residual)** |
| `permissions-policy-create-blocked` | Authenticate | prompt → OK | prompt → OK ✓ | prompt → OK ✓ |
| `cross-origin-iframe-no-allow` | Register | prompt (BUG) | NAE ✓ | NAE ✓ (defense-in-depth) |
| `cross-origin-iframe-no-allow` | Authenticate | prompt (BUG) | NAE ✓ | NAE ✓ (defense-in-depth) |
| `cross-origin-iframe-with-allow` | Register | sOWA | sOWA | **NAE (over-rejection)** |
| `cross-origin-iframe-with-allow` | Authenticate | prompt → OK | prompt → OK ✓ | **NAE (over-rejection)** |
| `cross-origin-iframe-policy-overrides-allow` | Register | sOWA | sOWA | **NAE (over-rejection)** |
| `cross-origin-iframe-policy-overrides-allow` | Authenticate | prompt (BUG) | NAE ✓ | NAE ✓ |

**Bold** entries flag the residual VULN-582 gap and the cross-origin iframe over-rejection on Firefox documented above.

### Known related limitation (not in scope)

During manual verification, registering a passkey from inside any iframe — including cross-origin iframes with `allow="publickey-credentials-create"` — fails with `Invalid 'sameOriginWithAncestors' value`. This is a pre-existing limitation in `Fido2ClientService.createCredential` (`libs/common/src/platform/services/fido2/fido2-client.service.ts`, lines 134–138) tracked separately in [PM-16480]. The behavior is identical on `main` and this branch; the `cross-origin-iframe-with-allow` test page therefore serves as a regression guard (parity between branches) rather than a functional verification.

### Follow-up

A comprehensive fix for the Firefox/Safari residual VULN-582 gap and the cross-origin iframe over-rejection requires reading `Permissions-Policy` response headers from the background via `chrome.webRequest.onHeadersReceived` and chaining through `chrome.webNavigation.getAllFrames` for `allow=` delegation. Tracked in [PM-38940].

**Per AppSec, on merge of this PR:**

- **VULN-398** moves to "Remediated" and closes after release verification — fully fixed by this PR.
- **VULN-582** stays open until PM-38940 also lands, because this PR only partially remediates it (Chrome fully, Firefox/Safari top-level header-deny is the residual).

[PM-37768]: https://bitwarden.atlassian.net/browse/PM-37768
[PM-37769]: https://bitwarden.atlassian.net/browse/PM-37769
[PM-16480]: https://bitwarden.atlassian.net/browse/PM-16480
[PM-38940]: https://bitwarden.atlassian.net/browse/PM-38940
[VULN-582]: https://bitwarden.atlassian.net/browse/VULN-582
[VULN-398]: https://bitwarden.atlassian.net/browse/VULN-398


apps/browser/src/autofill/fido2/content/fido2-page-script.ts
```diff
@@ -219,6 +227,42 @@ import { Messenger } from "./messaging/messenger";
...
   }
 
+  /**
+   * Checks whether the document's Permissions Policy allows the requested WebAuthn feature.
+   * Defaults to allowing the request when the FeaturePolicy API is unavailable (e.g. Safari),
+   * preserving existing functionality on browsers that do not expose the API.
+   *
+   * @param featureName Permissions Policy feature name, e.g. `publickey-credentials-get`.
+   */
+  function isFeaturePolicyAllowed(featureName: string): boolean {
```
dan-livefront: I believe Firefox exposes `document.permissionsPolicy`. If only `permissionsPolicy` exists, this function returns true and the vuln may remain on Firefox. Can we verify?

bensbits91: Good catch! The investigation flipped a couple of my assumptions, so summarizing what I found and what I changed:

**The actual browser API state** (verified against each engine's main branch):

| Browser | `document.permissionsPolicy` | `document.featurePolicy` |
|---|---|---|
| Chromium | not exposed | exposed |
| Firefox (default) | not exposed | gated behind `dom.security.featurePolicy.webidl.enabled`, default `false` |
| Safari | not exposed | not exposed |

Sources: Firefox [`Document.webidl`](https://hg.mozilla.org/mozilla-central/raw-file/tip/dom/webidl/Document.webidl), [`StaticPrefList.yaml`](https://hg.mozilla.org/mozilla-central/raw-file/tip/modules/libpref/init/StaticPrefList.yaml); Chromium [`document.idl`](https://raw.githubusercontent.com/chromium/chromium/main/third_party/blink/renderer/core/dom/document.idl).

No shipping browser exposes `document.permissionsPolicy` today, but your underlying concern was directionally right: on default-config Firefox the gate falls through to "allow," leaving the vuln exposed. Firefox does enforce the policy natively at the WebAuthn entry point ([`CredentialsContainer.cpp`](https://hg.mozilla.org/mozilla-central/raw-file/tip/dom/credentialmanagement/CredentialsContainer.cpp), via `FeaturePolicyUtils::IsFeatureAllowed`), but our injected override bypasses that path and can't query the same internal API from JS.

**What I changed in [6bcc45c](https://github.com/bitwarden/clients/pull/21054/commits/6bcc45ca1d1a55a44d077433b0f7092ccc07cebb):**

1. Renamed the helper to `isWebAuthnFeatureAllowed` and check `document.permissionsPolicy` first (forward-compat for the standardized name) before falling back to `document.featurePolicy`. Costs nothing today, future-proofs if browsers ship the spec name.
2. Added a defense-in-depth fallback for when neither API is available: deny in cross-origin iframes (`self !== top` and `top.origin !== self.origin`, or accessing `top.location` throws). Matches the spec default allowlist (`self`) for `publickey-credentials-*`.
3. Five new unit tests covering `permissionsPolicy` precedence, `permissionsPolicy` standalone, cross-origin iframe deny (different origin), cross-origin iframe deny (`top.location` throws), and top-level allow.
4. Manually verified on Firefox and Chrome against the seven test pages from the test-the-web repo — matrix in the updated PR description.

**Residual gaps that this PR knowingly accepts and documents:**

- VULN-582 on default-config Firefox and Safari: a top-level page emitting `Permissions-Policy: publickey-credentials-{get,create}=()` still surfaces the ceremony, because we have no way to read the header on those browsers.
- Over-rejection trade-off: cross-origin iframes that legitimately received `allow=publickey-credentials-*` are refused on Firefox/Safari, since we can't introspect the `allow` attribute without the policy API.

Both of these require reading the response header server-side and walking the frame chain for `allow=` delegation — meaningful additional work that I think belongs in its own PR. I filed [PM-38940](https://bitwarden.atlassian.net/browse/PM-38940) to track a background-side `Permissions-Policy` header reader (via `chrome.webRequest.onHeadersReceived` + `chrome.webNavigation.getAllFrames`) which would close both residuals and remove the over-rejection.

apps/browser/src/autofill/fido2/content/fido2-content-script.ts
```diff
@@ -126,6 +126,11 @@ import { MessageWithMetadata, Messenger } from "./messaging/messenger";
     requestId: string,
     messageData: InsecureCreateCredentialParams | InsecureAssertCredentialParams,
   ): Promise<Message | undefined> {
+    const featureName = permissionsPolicyFeatureForCommand(command);
+    if (featureName != null && !isWebAuthnFeatureAllowed(featureName)) {
+      return Promise.reject(buildPermissionsPolicyError(featureName));
```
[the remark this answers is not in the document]
bensbits91: Addressed in https://github.com/bitwarden/clients/pull/21054/commits/e63cc32cd4bdad50969e370ca15016ff4dab11e8: The serialization round-trip stripped instanceof DOMException and .code even though .name and .message survived. Added a small rehydrateDOMException helper in the page script that catches NotAllowedError-named rejections coming back from the messenger and reconstructs a real DOMException before re-throwing. Scoped narrowly to that one name (the only DOMException our gate produces), so other rejections pass through untouched. Added unit tests covering create, get, and the pass-through case.

dan-livefront (approved):
Discussed various items offline that were addressed and others that were identified and tracked.

PR #21058, opened 2026-06-04
[PM-37913] - fix cloning archived cipher

jaasen-livefront:
## 🎟️ Tracking

[PM-37913](https://bitwarden.atlassian.net/browse/PM-37913)

## 📔 Objective

When cloning a cipher through SDK create operations, ensure the new cipher is not created as archived.

This updates the SDK create request mapping to explicitly send `archivedDate: null`, even when the source `CipherView` has an archived date. It also adds test coverage in `DefaultCipherSdkService` to verify archived source ciphers are sent to the SDK create API with `archivedDate` cleared.

Testing:
- `npx jest libs/common/src/vault/services/cipher-sdk.service.spec.ts -t "should create cipher using SDK when orgAdmin is false" --runInBand`

## 📸 Screenshots

https://github.com/user-attachments/assets/45c9a128-7921-48ef-a5dd-519aa084ba84

[PM-37913]: https://bitwarden.atlassian.net/browse/PM-37913?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/common/src/vault/models/view/cipher.view.spec.ts
+++ b/libs/common/src/vault/models/view/cipher.view.spec.ts
@@ -411,6 +411,7 @@ describe("CipherView", () => {
         username: "testuser",
         password: "testpass",
       });
+      expect(result.archivedDate).toBeNull();
     });
 
     it("handles undefined organizationId and folderId", () => {
--- a/libs/common/src/vault/models/view/cipher.view.spec.ts
+++ b/libs/common/src/vault/models/view/cipher.view.spec.ts
@@ -411,7 +411,7 @@ describe("CipherView", () => {
         username: "testuser",
         password: "testpass",
       });
-      expect(result.archivedDate).toBeNull();
+      expect(result.archivedDate).toBeUndefined();
     });
 
     it("handles undefined organizationId and folderId", () => {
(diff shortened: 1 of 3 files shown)


nikwithak (dismissed):

libs/common/src/vault/models/view/cipher.view.spec.ts
```diff
@@ -411,6 +411,7 @@ describe("CipherView", () => {
         username: "testuser",
         password: "testpass",
       });
+      expect(result.archivedDate).toBeNull();
```
jengstrom-bw: this should be `toBeUndefined()` now that `archivedDate` is undefined in `cipher.view.ts`

libs/common/src/vault/services/cipher-sdk.service.spec.ts
```diff
@@ -136,6 +137,7 @@ describe("DefaultCipherSdkService", () => {
         expect.objectContaining({
           name: cipherView.name,
           organizationId: expect.anything(),
+          archivedDate: null,
```
jengstrom-bw: this should be changed to 
`archivedDate: undefined,`

jengstrom-bw (changes requested):

jengstrom-bw (approved):

PR #21061, opened 2026-06-04
[PM-38583] chore: remove incompatible special character for passwords

Lof-31:
This special character "^" is often not accepted (I would say in 40% of cases) by websites as a valid character for a password.  As a user, I always click "generate again" when it appears in my new generated passwords.

## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-38583

### DIFF
--- a/libs/tools/generator/core/src/engine/data.ts
+++ b/libs/tools/generator/core/src/engine/data.ts
@@ -6,7 +6,7 @@ function toCharacterSet(characters: string) {
   return Object.freeze(set as CharacterSet);
 }
 
-const SpecialCharacters = toCharacterSet("!@#$%^&*");
+const SpecialCharacters = toCharacterSet("!@#$%&*");
 
 /** Sets of Ascii characters used for password generation */
 export const Ascii = Object.freeze({
--- a/libs/tools/generator/core/src/engine/data.ts
+++ b/libs/tools/generator/core/src/engine/data.ts
@@ -6,7 +6,7 @@ function toCharacterSet(characters: string) {
   return Object.freeze(set as CharacterSet);
 }
 
-const SpecialCharacters = toCharacterSet("!@#$%&*");
+const SpecialCharacters = toCharacterSet("!@#$%?&*");
 
 /** Sets of Ascii characters used for password generation */
 export const Ascii = Object.freeze({


harr1424:
Hi @Lof-31  👋  and thanks for your contribution! 

We appreciate the time you took to improve the password generator, but I want to emphasize that any change impacting generated credential entropy will be considered very carefully. The first phase of this assessment is [review by our internal product team](https://contributing.bitwarden.com/contributing/pull-requests/community-pr-process/#product-assessment). 

I'll reach out with any questions that may come up, and if you would ever like a status update please feel free to tag me in a comment here. 

Thanks again and congratulations on your first contribution to Bitwarden!

Lof-31:
Sooo any updates? It's the smallest change possible (removed one character)

pamperer562580892423:
> It's the smallest change possible (removed one character)

That change is not completely trivial. It reduces the entropy of generated passwords, although only to a relatively small degree.

I'm only another user, but I would rather like to see the opposite (the option to include more special characters).

harr1424:
@Lof-31 I apologize but our product team hasn't yet had an opportunity to review this. In general, we try to avoid any reduction in entropy. For context, the EFF maintains [a list of 66,666 words](https://www.eff.org/files/2016/07/18/eff_large_wordlist.txt) that we use in order to generate passphrases. Four of these words contain hyphens, which led to generated passphrases seeming to contain more than the specified number of words. We ultimately decided to [avoid using the hyphenated words](https://github.com/bitwarden/sdk-internal/blob/21488a3776675afd9525905d3081f854692fcd6f/crates/bitwarden-generators/src/passphrase.rs#L110) but this was a decision that spanned multiple teams. 

Your PR proposes reducing the set of special characters from 8 to 7. This is a much more significant reduction than 66,666 to 66,662, which multiple developers were hesitant to approve. 

I don't want to speak for our product team, but I expect they will want the `^` character replaced with some other special character. 

>This special character "^" is often not accepted (I would say in 40% of cases) by websites as a valid character for a password

Do you happen to have any ideas for a replacement character that could be expected to work well across sites?

pamperer562580892423:
> For context, the EFF maintains [a list of 66,666 words](https://www.eff.org/files/2016/07/18/eff_large_wordlist.txt) that we use in order to generate passphrases.

@harr1424 Sorry, but that list only contains 7776 words.

harr1424:
@pamperer562580892423 thank you for correcting me, I should have used `wc` instead of assuming the digits adjacent to the word in the file represented the word count. 

Notwithstanding, 7776 -> 7772  is a very different impact on generated credential entropy than 8 -> 7.

Lof-31:
> Do you happen to have any ideas for a replacement character that could be expected to work well across sites?

Ok, thanks for the precision. That's a good idea to avoid loosing antropy. 

We could replace '^' by one (or even all) of these: '-', '_', '?'.

These are almost always accepted as valid chars in passwords. 

I just updated my PR, adding the question mark. 

Thank you

harr1424:
@Lof-31 thanks for adding in `?` to the special characters list! I've reached out once more to our product team to ask if we can move this to Engineering Review. I'll keep you updated.

Lof-31:
> I'll keep you updated.

Hey, so do you have any news? :) 

Thank you

harr1424:
@Lof-31 I've received approval to move your contribution to "Engineering Review" as of this morning, and I will begin evaluating your proposed changes at some point during the week. The code changes look good, my main concern is ensuring the change is in fact a net improvement in terms of our generated credentials. I'd like to empirically test that the `?` character is more agreeable with password validators on some common sites compared to `^`, and thereafter run the proposed change by some colleagues who have more history with our credential generators.

harr1424:
Hi @Lof-31 today I began creating accounts on a subset of the [most-visited websites](https://en.wikipedia.org/wiki/List_of_most-visited_websites). I've skipped any NSFW sites and likewise skipped over sites owned by the same parent company, since Meta/Alphapbet/Microsoft have likely standardized their password validation. I've not yet encountered any problems using a `^` character in my account passwords. 

Would you be willing to share examples of sites where this character was not allowed?

djsmith85 (commented):
@harr1424 Regardless if adjusting the special character set is accepted, I thought we had migrated the generator code into the [SDK](https://github.com/bitwarden/sdk-internal/blob/4963a281c5a4c0de9454f01282f814a30c57470b/crates/bitwarden-generators/src/password.rs#L132) and have also switched over to using it in the clients. I remember this happening a while back and that would mean most/all of the typescript code is deprecated and should be removed.

@adudek-bw Thoughts?

harr1424:
@Lof-31 I would like to share some findings and a few suggestions, as well as provide an explanation for why I am closing this PR: 
- I haven't been able to find a website disallowing the `^` character during new account creation; I am certain such sites exist, but I can't support a 40% occurrence rate and suspect this may be more infrequent. 
- Workarounds do exist: you may edit the generated password manually, or else regenerate a password until the offending character is not present.
- Instead of forcing a character set change on all users, a more comprehensive solution would be to allow flexibility to users to add/remove specific characters from the set. [This is a long-running feature request](https://community.bitwarden.com/t/password-generator-ability-to-filter-manually-exclude-certain-special-characters/141). Top voted feature requests are routinely reviewed by our internal teams for prioritization. 

The Web Hypertext Application Technology Working Group has proposed an HTML attribute `passwordrules` which allows websites to expose their password requirements to credential generators like Bitwarden ([See proposed standard](https://github.com/whatwg/html/issues/3518)). Bitwarden [already respects](https://github.com/bitwarden/sdk-internal/pull/1102/changes) this attribute. It may be worth trying to contact these websites that have been problematic and request they begin using this attribute on their password form fields. 

The time you took to propose these code changes is appreciated, but I can't proceed with introducing this change to all Bitwarden users based upon my analysis of the issue so far.

PR #21067, opened 2026-06-05
[PM-38357] Surface more descriptive error messages from server on Send edit failure

### LINKED ISSUE #20897: Send text not working, gives empty error message
### Steps To Reproduce

Best explained with a screenshot:

Image

Tried many times - it doesn't work, and the error message is always blank.

### Expected Result

Not that.

### Actual Result

See above

### Screenshots or Videos

_No response_

### Additional Context

_No response_

### Operating System

Windows

### Operating System Version

_No response_

### Web Browser

Brave

### Browser Version

 Brave 1.90.124 (Official Build) (64-bit) Chromium: 148.0.7778.179

### Environment Versions

Version: 2026.4.1

SDK: 'main (aaccd8d)'

Server version: 2026.4.2

### Issue Tracking Info

mcamirault:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-38357

## 📔 Objective

Fixes https://github.com/bitwarden/clients/issues/20897

When a Send edit fails we used to catch the error and show the user a generic "Saving Send edits failed" message. This is especially unhelpful in cases where the edit failed because the Send violates an organization policy (the Send UI will usually hide policy-violating options for its fields, but if the policy is a new one and the user is using an old version of the clients they could still save a Send in an invalid state). Instead we'll now surface error messages from the server if they occur, which give the user much more useful feedback.

## 📸 Screenshots

*In the following video an org policy requires email verification on Sends, but the user is using an old version of the web client and can thus attempt to create a Send with no auth*

https://github.com/user-attachments/assets/86a66400-084b-4b22-a7f8-072ac3345dbb


mcamirault:
@claude Give this a review

harr1424 (dismissed):
Please see findings from Claude, if any are valid I am happy to re-review any changes. Otherwise looks good! Down the road we may want to map the various errors in `SendValidationService.cs` to enum variants, and provide `i18n` translations for them, to avoid relying on hardcoded English messages.

libs/tools/send/send-ui/src/send-form/components/send-form.component.ts
```diff
@@ -135,11 +135,6 @@ export class SendFormComponent implements AfterViewInit {
   submit = async () => {
     const sendView = await this.sendFormService.submitSendForm();
 
-    // Send form had errors or otherwise failed to submit
```
mcamirault, line 138: ~~`submitSendForm` will now throw on errors so this branch is no longer needed~~
Nope, see Claude's comment below

mcamirault:
@harr1424 Adding the label took long enough to trigger the review that I thought it wasn't working, hence my explicit callout to Claude. What's interesting is that the latter's review is much more detailed and helpful than the former. Wonder if they're using different skills?

Anyway, I've addressed all the comments that the more-useful Claude gave me

harr1424 (dismissed):
Looks good!

harr1424 (approved):

PR #21068, opened 2026-06-05
[PM-18344] Implement Clipboard Support on Wayland/Gnome via RemoteDesktop Portal

### LINKED ISSUE #13431: Flatpak + Wayland - Clipboard / Copy-Paste not working
### Steps To Reproduce

1. go to any entry
2. click on copy username (or password)

### Expected Result

user or pass copied to system clipboard

### Actual Result

clipboard unchanged

### Screenshots or Videos

_No response_

### Additional Context

log output:

```
Error occurred in handler for 'clipboard.write': [Error: Unknown error while interacting with the clipboard: X11 server connection timed out because it was unreachable] {
  code: 'GenericFailure'
}
```

### Operating System

Linux

### Operating System Version

Pop!_OS 24.04 LTS

### Installation method

Other

### Build Version

Version 2025.2.0 SDK 'main (28c7e29)' Shell 34.0.0 Renderer 132.0.6834.83 Node 20.18.1 Architecture x64

### Issue Tracking Info

quexten:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-18344

## 📔 Objective

Implements gnome wayland clipboard support by using the remote desktop portal. Gnome does not support `zwlr_data_control_manager_v1` so we need a different way to access the clipboard. One way to do this reliably is using the remote desktop (or input capture on very new protals) portal. The remote desktop session can be used to write to the clipboard.

This shows the user a permission prompt once, and subsequent copies do not show a permissions prompt. The remote desktop session is resumed using a token saved to disk, and used to write to the clipboard.

## 📸 Screenshots

https://github.com/user-attachments/assets/61256a6a-bb69-46e7-a95d-599c7332d6c7


apps/desktop/desktop_native/core/src/clipboard/portal_backend.rs
```diff
@@ -0,0 +1,333 @@
...
+    PersistMode, Session,
+};
+use futures::StreamExt;
+use tracing::{error, info};
+
+/// MIME type advertised and served for the clipboard selection.
+const MIME_TEXT: &str = "text/plain;charset=utf-8";
+
+/// File name (under the config dir) used to persist the RemoteDesktop session restore token.
+const TOKEN_FILE: &str = "remote_desktop_portal_token";
```
quexten, line 24: Note: Later we will refactor this to re-use the portal for auto-type. We can re-use the same permissions and the same remote desktop session.

We will not do that here because it will increase complexity.

apps/desktop/desktop_native/core/src/clipboard_portal_linux.rs
```diff
@@ -0,0 +1,333 @@
+//! Alternative clipboard implementation for GNOME, using XDG Desktop Portals.
```
quexten: Note: Works on DE's other than gnome, but there is no point for the clipboard. We may use it for auto-type on KDE though.

apps/desktop/desktop_native/core/src/clipboard/portal_backend.rs
```diff
@@ -0,0 +1,307 @@
...
+}
+
+/// Persist the session token to a plain file under the config dir.
+fn save_session_token(token: &str) -> Result<()> {
+    let mut path = token_path()?;
+    std::fs::create_dir_all(&path)?;
+    path.push(TOKEN_FILE);
+    std::fs::write(path, token)?;
+    Ok(())
+}
```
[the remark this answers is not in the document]
quexten: Yes this is accepted and common within flatpak apps.

apps/desktop/desktop_native/core/src/clipboard/portal_backend.rs
```diff
@@ -0,0 +1,307 @@
...
+/// Whether the portal-based clipboard fallback should be used when the direct `arboard` write
+/// fails.
+///
+/// True on a GNOME desktop, where `arboard` cannot reliably set the clipboard on Wayland. Reads
+/// `XDG_CURRENT_DESKTOP` for the desktop environment.
+pub(crate) fn should_use_portal() -> bool {
+    std::env::var("XDG_CURRENT_DESKTOP")
+        .map(|desktop| desktop.to_ascii_uppercase().contains("GNOME"))
+        .unwrap_or(false)
+}
```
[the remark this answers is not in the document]
quexten: Gnome has dropped x11 support in mutter. There is no concern with x11 compatibility.

apps/desktop/resources/com.bitwarden.desktop.devel.yaml
```diff
@@ -9,8 +9,10 @@ finish-args:
   - --share=ipc
   - --share=network
   - --socket=wayland
-  # This should be fallback-x11, but on gnome/mutter, zwlr_data_control_manager_v1 is not implemented
-  # so we need to use x11 as a fallback to make copy paste work
+  # A bug in Electron 39 (which now enables Wayland by default) causes a crash on
```
quexten, line 12: Note to plaform: Once the bug with electron is resolved we can remove the double socket and become wayland only.

coroiu (approved):

PR #21070, opened 2026-06-05
[CL-1176] Add bit-badge-group and deprecate bit-badge-list

willmartian:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/CL-1176

## 📔 Objective

Adds `<bit-badge-group>`, a newer version of `<bit-badge-list>` that renders badges responsively instead of by user-defined max number, split out from #20900 so it can land independently of the `bit-table-v2` work.

The group composes `bitOverflowList` for measurement-driven sizing: badges that don't fit the container width are hidden, and a default-variant "+N" badge is rendered at the end with a `bit-chip-action` triggering a `bit-popover` that renders the remaining badges. Resize the container and more or fewer badges become visible — no `maxItems` knob.

To enable the popover approach without all the required inputs for a normal popover, we split out common UI to a base component and used that to compose the existing popover and new popover-panel.

Additionally, we updated the badge to use the tooltip component internally (leftover from a previous approach to this design, but still a valuable change). This better supports existing usages of tooltips on badges -- a couple of small template changes were required in **AC and SM** for this.

**Deprecation**

`bit-badge-list` gets a `@deprecated` JSDoc tag pointing at `bit-badge-group`, and its storybook title becomes "Component Library/Badge/List (Deprecated)".

## 📸 Screenshots

See the Badge/Group story in Storybook.


libs/components/src/badge-group/badge-group.component.ts
```diff
@@ -0,0 +1,56 @@
...
+    return overflow
+      .map((i) => items[i]?.elementRef.nativeElement.textContent?.trim() ?? "")
+      .filter(Boolean)
+      .join(", ");
+  });
+
+  constructor() {
+    effect(() => {
+      const items = this.items();
+      items.forEach((item, i) => item.pinned.set(i === 0));
```
vleague2: Is there a way we could do this with a `computed` instead of an `effect`?

libs/components/src/badge-group/badge-group.stories.ts
```diff
@@ -0,0 +1,73 @@
...
+          <span bitBadge variant="success">Shared</span>
+          <span bitBadge variant="warning">Archived</span>
+          <span bitBadge variant="primary">Favorite</span>
+        </bit-badge-group>
+      </div>
+    `,
+  }),
+};
+
+export const Mixed: Story = {
```
vleague2: `Mixed` implies to me that it is multiple widths, which is not actually true. I'm also not sure what this story shows that `Narrow` doesn't?

vleague2 (commented):
Storybook testing reveals some flashing and that the title on the `+N` badge is incorrect:

https://github.com/user-attachments/assets/2e1da8da-3b93-43ac-8815-a3311047068c

Image

libs/components/src/badge-group/badge-group.component.ts
willmartian: Yes but no, because we are setting an input. An input cannot be designated as a `linkedSignal` (new signal construct that is both writeable and based on an existed signal).

We _could_ invert this (which is generally the way to solve these things), such that badge injects the badge group, and does the _am I first check_ itself. But then we would have a circular dependency. That itself could be solved with a separate injectable interface, but at that point it feels like a lot of boilerplate. Instead, I think this is an okay use of `effect`. Good question.

libs/components/src/badge-group/badge-group.stories.ts
willmartian: Renamed to `WithIcons`

willmartian:
> title on the `+N` badge is incorrect:

@vleague2 Updated the badge component to use `bitTooltip` internally instead of the title attribute. Increases scope, but fixes this and remedies some design debt. I can change the fix to use `title` if we don't want the increased testing surface right now--just lmk.

> Storybook testing reveals some flashing

Fixed this by 1.) measuring earlier in the overflow util and 2.) binding to `tw-invisible`

libs/components/src/badge-group/badge-group.component.ts
vleague2: Sounds good!

vleague2 (dismissed):

libs/components/src/badge-group/badge-group.component.html
```diff
@@ -0,0 +1,13 @@
+<div
+  bitOverflowList
+  [items]="items()"
+  [gap]="8"
+  #ovf="bitOverflowList"
+  [class.tw-invisible]="!ovf.ready()"
+  class="tw-flex [&>*]:tw-shrink-0"
+>
+  <ng-content></ng-content>
+  <span bitBadge bitOverflowTrigger variant="subtle" [title]="hiddenLabels()">
```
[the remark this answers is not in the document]
vleague2: This might be legit @willmartian. I just tried with Voiceover to navigate the badge group and it's only letting me read it out as one big group, like all the badges at once with the "+N" at the end. No content from the tooltip is coming through.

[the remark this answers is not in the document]
vleague2: No longer applicable with new chip / popover approach

libs/components/src/chips/shared/base-chip.directive.ts
```diff
@@ -58,6 +65,18 @@ const variantStyles: Record<ChipVariant, string[]> = {
...
+    "tw-text-fg-body",
+    "[&:is(button,a)]:hover:tw-bg-bg-quaternary",
+    "[&:is(button,a)]:focus-visible:tw-bg-bg-quaternary",
+    "has-[button:hover:not([bit-chip-dismiss-button])]:tw-bg-bg-quaternary",
+    "has-[a:hover]:tw-bg-bg-quaternary",
+    "has-[button:focus-visible:not([bit-chip-dismiss-button])]:tw-bg-bg-quaternary",
+    "has-[a:focus-visible]:tw-bg-bg-quaternary",
+  ],
+  // deprecated -- remove with filter chip variant
+  filter: [
```
vleague2, line 79: ℹ️  I noticed the subtle styles didn't match the Figma. The filter chip has a different background than the subtle style. Since filter is going to get deprecated in the table PR, this was a quick and easy way to update the styles for the real subtle variant while not breaking the filter chip in the meantime

libs/components/src/popover/popover.component.html
```diff
@@ -5,14 +5,14 @@
...
-    [attr.aria-label]="title()"
+    [attr.aria-label]="dialogLabel()"
   >
     <div class="tw-overflow-hidden tw-rounded-xl tw-shadow-md tw-border tw-border-border-base">
       <div
         class="tw-relative tw-z-20 tw-max-w-[22.5rem] tw-break-words tw-bg-bg-primary tw-text-fg-body"
       >
         <ng-content select="bit-popover-header"></ng-content>
-        <div class="tw-p-6">
+        @if (showCloseButton()) {
```
vleague2: ℹ️ Design requested the title and close button to become optional

libs/components/src/popover/popover.component.html
```diff
@@ -23,17 +23,24 @@
...
+              bitTypography="body2"
+              [class]="contentClasses()"
+              [tabindex]="maxHeightClass() ? '0' : null"
+              (scroll)="onContentScroll($event)"
+            >
+              <ng-content></ng-content>
             </div>
-            <ng-content select="bit-popover-footer"></ng-content>
           </div>
+          <ng-content select="bit-popover-footer"></ng-content>
```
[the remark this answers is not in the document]
vleague2: Fix here [7f6a2cf](https://github.com/bitwarden/clients/pull/21070/commits/7f6a2cf8dd3ec340dfc4793033088ee82e86dde5)

libs/components/src/popover/popover.component.ts
```diff
@@ -32,19 +35,92 @@ export class PopoverComponent {
...
+      "tw-duration-200",
+      this.bodyScrolled() ? "tw-border-border-base" : "tw-border-transparent",
+    ].join(" "),
   );
+
+  protected readonly contentClasses = computed(() =>
+    [
+      "tw-ps-6",
+      !this.footer() && "tw-pb-6",
+      !this.title() && "tw-pt-6",
```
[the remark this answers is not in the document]
vleague2: [3df2c8f](https://github.com/bitwarden/clients/pull/21070/commits/3df2c8f788d53a4c4815768fe202b1445dea21e8)

libs/components/src/badge/badge.component.ts
```diff
@@ -94,15 +97,35 @@ const getDefaultIconForVariant = (variant: BadgeVariant) => defaultIconMap[varia
...
   imports: [IconComponent],
+  // OverflowItemDirective is applied to every badge so wrappers like
+  // `bit-badge-group` can let `bitOverflowList` measure and hide them. None
+  // of its inputs/outputs are exposed — `pinned` is internal-only, set
+  // programmatically by the wrapper.
+  //
+  // TooltipDirective surfaces the badge's text (or the `title` override) as a
+  // styled tooltip in place of the native `title` attribute; the content is
+  // wired up below from `titleContent()`.
+  hostDirectives: [OverflowItemDirective, TooltipDirective],
```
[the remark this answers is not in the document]
vleague2: addressed here [f8cf28a](https://github.com/bitwarden/clients/pull/21070/commits/f8cf28af0045dbe10b4ba9bec8ba22f90d3a2c73)

libs/components/src/popover/popover.component.ts
```diff
@@ -32,19 +35,92 @@ export class PopoverComponent {
...
+
+  protected readonly contentClasses = computed(() =>
+    [
+      "tw-ps-6",
+      !this.footer() ? "tw-pb-6" : "",
+      !this.title() ? "tw-pt-6" : "",
+      // When there's no title (and no header) to carry it, the content sits at the
+      // top and must reserve room for the absolutely-positioned close button so its
+      // first line doesn't run underneath it. Only needed while that button is shown.
+      this.showCloseButton() && !this.title() && !this.header() ? "tw-pe-13" : "tw-pe-6",
```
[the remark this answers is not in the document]
vleague2: fixed [a7bce28](https://github.com/bitwarden/clients/pull/21070/commits/a7bce28a8cd8c17cd1107c14f33ed2501bca65e1)

libs/components/src/popover/popover.component.html
BryanCunningham: Close button can't be optional. We trap focus in the popover. A keyboard user could never get out out of it right?

libs/components/src/popover/popover.component.ts
```diff
@@ -32,19 +35,92 @@ export class PopoverComponent {
   /** Reference to the popover content template */
   readonly templateRef = viewChild.required(TemplateRef);
 
+  /**
+   * Whether the scrollable content area has been scrolled away from the top
+   */
+  protected readonly bodyScrolled = signal(false);
```
BryanCunningham: Is there really a use case for a popover to be scrollable? They aren't really suppose to have that much content in them are they?

libs/components/src/popover/popover.stories.ts
```diff
@@ -117,6 +117,109 @@ export const Default: Story = {
...
+    const canvasEl = context.canvasElement;
+    const button = getByRole(canvasEl, "button");
+    await userEvent.click(button);
+  },
+};
+
+/**
+ * Set `[showCloseButton]="false"` to omit the built-in close button when the
+ * popover is dismissed another way — a backdrop click, Escape, or re-triggering
+ * its anchor. The content reclaims the space the button would have reserved.
```
BryanCunningham: Are we just assuming that a SR user will automatically know to use the escape key to close the popover?

libs/components/src/popover/popover.component.html
BryanCunningham: Disregard my comment

vleague2: For posterity, we discussed this off GH -- yes, focus is trapped in the popover (it goes to the popover body, which is scrollable in this implementation). Keyboard users would use the escape key to close the popover

libs/components/src/popover/popover.component.ts
vleague2: For posterity, we discussed this off GH -- for 99% of cases I don't think we will need the scrollability, but for some use cases like this badge group overflow, there is the possibility that there could be tens of badges inside the overflow. Scrolling seems like a better outcome than the popover being so tall it runs off the page

libs/components/src/popover/popover.stories.ts
vleague2: For posterity, we discussed this off GH -- I think the escape key would be a normal expected keyboard functionality for closing a popover. I can do more research to be sure

BryanCunningham (dismissed):

libs/components/src/badge-group/badge-group.component.ts
```diff
@@ -0,0 +1,64 @@
...
+    PopoverPanelComponent,
+    OverflowItemDirective,
+    OverflowListDirective,
+    OverflowTriggerDirective,
+    I18nPipe,
+  ],
+  changeDetection: ChangeDetectionStrategy.OnPush,
+})
+export class BadgeGroupComponent {
+  readonly badges = input<BadgeGroupItem[]>([]);
```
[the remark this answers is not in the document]
vleague2: I believe this is covered in https://github.com/bitwarden/clients/pull/21742

libs/components/src/badge-group/badge-group.component.html
```diff
@@ -0,0 +1,43 @@
...
+      bitOverflowItem
+      #ovfItem="bitOverflowItem"
+      [pinned]="$first"
+      [variant]="badge.variant ?? 'primary'"
+      [startIcon]="badge.startIcon"
+      [class]="
+        ovfItem.shouldShrink() ? 'tw-flex-initial tw-min-w-0 tw-overflow-hidden' : 'tw-shrink-0'
+      "
+    >
+      {{ badge.label }}<span class="tw-sr-only">,</span>
```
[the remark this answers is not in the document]
vleague2: Fixed [36a56aa](https://github.com/bitwarden/clients/pull/21070/commits/36a56aab83ed2c08fd02ed05c5a824cbcd90513d)

libs/components/src/badge-group/badge-group.component.html
```diff
@@ -0,0 +1,39 @@
...
+  #ovf="bitOverflowList"
+  [class.tw-invisible]="!ovf.ready()"
+  class="tw-flex [&>*]:tw-shrink-0"
+>
+  <!-- The first badge is pinned so at least one is always visible. -->
+  @for (badge of badges(); track $index) {
+    <span
+      bitBadge
+      bitOverflowItem
+      [pinned]="$first"
```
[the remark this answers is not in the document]
vleague2: Fixed [181f7d2](https://github.com/bitwarden/clients/pull/21070/commits/181f7d2a69d8ff999e991c716fc8201f119c9e56)

libs/components/src/badge-group/badge-group.component.html
```diff
@@ -0,0 +1,43 @@
...
+    <span
+      bitBadge
+      bitOverflowItem
+      #ovfItem="bitOverflowItem"
+      [pinned]="$first"
+      [variant]="badge.variant ?? 'primary'"
+      [startIcon]="badge.startIcon"
+      [class]="
+        ovfItem.shouldShrink() ? 'tw-flex-initial tw-min-w-0 tw-overflow-hidden' : 'tw-shrink-0'
+      "
```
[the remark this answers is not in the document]
vleague2: Fixed [11ea2a4](https://github.com/bitwarden/clients/pull/21070/commits/11ea2a43b4693991127892f095ce34ab6cd8a97c)

libs/components/src/badge-group/badge-group.component.html
```diff
@@ -0,0 +1,43 @@
...
+  }
+  <button
+    type="button"
+    bit-chip-action
+    bitOverflowTrigger
+    variant="subtle"
+    [label]="'+' + overflow().length"
+    [attr.aria-label]="'showMoreCount' | i18n: overflow().length"
+    [bitPopoverTriggerFor]="overflowPopover"
+  ></button>
```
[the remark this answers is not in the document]
vleague2: This looks to be working as expected already

apps/browser/src/_locales/en/messages.json
```diff
@@ -7287,6 +7287,15 @@
   "showMore": {
     "message": "Show more"
   },
+  "showMoreCount": {
+    "message": "Show $COUNT$ more",
```
BryanCunningham, line 7292: @vleague2 Is this the correct message for this? I thought it was just `+$COUNT$`

libs/components/src/badge-group/badge-group.component.html
```diff
@@ -0,0 +1,48 @@
...
+    that it does not show up in the tooltip -->
+    <span class="tw-sr-only" [hidden]="overflow().includes($index)">,</span>
+  }
+  <button
+    type="button"
+    bit-chip-action
+    bitOverflowTrigger
+    variant="subtle"
+    [label]="'+' + overflow().length"
+    [attr.aria-label]="'showMoreCount' | i18n: overflow().length"
```
BryanCunningham, line 33: @vleague2 Disregard message about copy. I see that is merely the aria-label and not the displayed text

BryanCunningham (dismissed):

jrmccannon (dismissed):

BryanCunningham (approved):

jrmccannon (approved):

cd-bitwarden (approved):
This looks good 👍

PR #21077, opened 2026-06-05
PM-36835 [Client] Add logging and error handling to Reports

prograhamming:
## 🎟️ Tracking

[PM-36835](https://bitwarden.atlassian.net/browse/PM-36835)

## 📔 Objective

Reports lack consistent error handling and logging. When a report fails to load, the failure is silent — no diagnostic output is produced, making it difficult to identify what went wrong. This has become pressing as clients are seeing the weak passwords report fail to load with no explanation.

## 📸 Screenshots

[PM-36835]: https://bitwarden.atlassian.net/browse/PM-36835?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/web/src/app/dirt/reports/pages/cipher-report.component.ts
```diff
@@ -132,22 +138,44 @@ export abstract class CipherReportComponent implements OnDestroy {
...
   async load() {
     this.loading = true;
-    await this.syncService.fullSync(false);
-    // when a user fixes an item in a report we want to persist the filter they had
-    // if they fix the last item of that filter we will go back to the "All" filter
-    if (this.currentFilterStatus) {
-      if (this.ciphers.length > 2) {
-        this.filterOrgStatus$.next(this.currentFilterStatus);
-        await this.filterOrgToggle(this.currentFilterStatus);
+    try {
```
lastbestdev, line 143: Are we able to remove this outer try/finally block? The existing reports have depended on `hasLoaded` and `loading` being set together at the end of this `load` method, and setting them individually may cause bugs. 

The `exposed-passwords-report.component` specifically looks like it renders the report contents when the value of `hasLoaded == true`, so this implementation would show the loaded state for that report if any exceptions are thrown.

lastbestdev: Additionally would just like to avoid nested `try` blocks if possible

apps/web/src/app/dirt/reports/pages/exposed-passwords-report.component.ts
```diff
@@ -97,12 +107,20 @@ export class ExposedPasswordsReportComponent extends CipherReportComponent imple
...
+      const exposedCount = await this.auditService.passwordLeaked(login.password);
       if (exposedCount > 0) {
         return { ...cv, exposedXTimes: exposedCount } as ReportResult;
       }
       return null;
-    });
+    } catch (e) {
+      // Do not log the password value or any decrypted vault data.
+      this.logService.error(
+        `[ExposedPasswordsReport] [${this.reportScope}] Failed to check if password was exposed`,
```
lastbestdev: Wondering if logging the cipher's `id` would be a concern here? And possibly the organization `id` as well when applicable?

I'm just thinking for troubleshooting and debugging purposes, logs without any way to identify which cipher is failing the check may not be very useful.

apps/web/src/app/dirt/reports/pages/cipher-report.component.ts
lastbestdev: Looks like claude was on this path already as well https://github.com/bitwarden/clients/pull/21077#discussion_r3391970400

PR #21085, opened 2026-06-05
Perf/incremental shadow overflow

blackwood:
## 🎟️ Tracking

PM-35196 (follow-on)

# PR 3 — perf/incremental-shadow-overflow

PR 1: #21083
PR 2: #21084

(Note: applies last, biggest caveats)

**Net objective:** Remove the worst-case latency spike: on candidate-cap overflow,
scan the kept set incrementally instead of re-walking the whole document.
Touches both files → must rebase onto PRs 1 (#21083) & 2 (#21084).

- **Expected:** Eliminates the full-document-walk spike; freed time may be
  reclaimed by more rebuilds; coverage risk for roots only the full scan found.
- **Result (headline — per-call spike):** S10 `handleNewShadowRoots`: the 
  worst-case freeze in shadow-root discovery went from ~166ms to instant, 
  ~11.7 seconds of total blocking over the minute vanished; not coincidence 
  because overflow kept happening just as often (only its cost changed).

### ⚠️ Tradeoff 1 — work shift, not CPU saving
S10 `queryAutofillForm` 225→415 calls, +11668ms: the ~11.7s freed is almost
exactly reabsorbed by more rebuilds on the clean baseline. On the integrated
stack (PRs 1+2 present) those rebuilds are gated, net win but
measured alone it redistributes (rather than reduces).

### ⚠️ Tradeoff 2 — narrow coverage gap
In-place `attachShadow` on an already-present element (field, no node-addition)
is no longer swept up. Documented by a committed characterization test;
decision was to document and ship.

### ⚠️ Caveat 3 — 1.A subsumes most of its S10 effect
With PR 1 (#21083) landed, 1.A prevents the deep-churn delivery that causes S10 overflow,
so this lever's unique remaining value is light-DOM node floods (e.g. Google Maps tiles), which 1.A can't mask.

## Coverage

S10 fields/forms max held 5/2; late attach via node-addition still discovered.

<details>
<summary><b>Benchmark scenario legend</b></summary>

| Label | Fixture | What it exercises |
|-------|---------|-------------------|
| **S1** | `1-cosmetic-feed-scroll` | Light-DOM cosmetic childList churn (feed scrolling) — no fillable change. |
| **S2** | `2-form-with-mutations` | A real form mutating; fields should plateau at the page's true count. |
| **S3** | `3-late-attached-shadow-form` | A shadow-hosted form attached *after* load — late discovery / coverage. |
| **S4** | `4-detach-reattach-churn` | Elements detaching and reattaching — reaper / leak behavior. |
| **S5** | `5-attribute-toggle-storm` | High-rate attribute toggling across many fields. |
| **S6** | `6-rerender-burst` | Bursty full re-renders (large mutation batches). |
| **S7** | `7-attr-relevance-transition` | A node becoming autofill-relevant via attribute change. _(retired)_ |
| **S8** | `8-shadow-attribute-storm` | Style/transform churn at frame rate *inside a field-bearing shadow root*. |
| **S9** | `9-shadow-tile-churn` | Field-less shadow roots doing deep childList churn (map-tile pattern). |
| **S10** | `10-shadow-scale-stress` | ~90k nodes: 300 field-less tiles + a nested-root chain + sparse forms (scale, overflow, nesting). |

</details>

### DIFF
--- a/apps/browser/src/autofill/services/collect-autofill-content.service.spec.ts
+++ b/apps/browser/src/autofill/services/collect-autofill-content.service.spec.ts
@@ -2875,15 +2875,14 @@ describe("CollectAutofillContentService", () => {
         expect(collectAutofillContentService["pendingMutationAddedElements"].size).toBe(0);
       });
 
-      it("trips the overflow flag at the cap and releases element refs", () => {
+      it("trips the overflow flag at the cap and keeps the capped set for incremental scan", () => {
         const cap = collectAutofillContentService["pendingMutationAddedElementsCap"];
         const widgets = Array.from({ length: cap + 50 }, () => document.createElement("my-widget"));
 
         collectAutofillContentService["collectAddedShadowRootCandidates"]([buildMutation(widgets)]);
 
         expect(collectAutofillContentService["pendingMutationAddedElementsOverflowed"]).toBe(true);
-        // Overflow path clears the Set immediately so refs don't linger until the debounce fires.
-        expect(collectAutofillContentService["pendingMutationAddedElements"].size).toBe(0);
+        expect(collectAutofillContentService["pendingMutationAddedElements"].size).toBe(cap);
       });
 
       it("is a no-op once overflow has been tripped (later batches are ignored)", () => {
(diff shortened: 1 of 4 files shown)


blackwood:
Undergoing experimental changes related to late shadow DOM hydration.

blackwood:
Superseded by https://github.com/bitwarden/clients/pull/21370

PR #20425, opened 2026-04-28
[CL-1042] Design system refresh: Milestone 3

vleague2:
## 🎟️ Tracking

[CL-1042](https://bitwarden.atlassian.net/browse/CL-1042)
^ This is the epic for the tickets that are in this feature branch (although not all of the epic tickets are in here; some targeted `main` directly)

## 📔 Objective

This PR is a feature branch for milestone 3 of the design system refresh, which includes form updates. This feature branch contains all the form-related changes that needed to go out together.

All commits in this PR have been individually code-reviewed and QA'd.

[CL-1042]: https://bitwarden.atlassian.net/browse/CL-1042?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #21092, opened 2026-06-05
[PM-38343] 2 step and require sso and remove unlock and remove export into simple toggle

JaredScar:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-38343

## 📔 Objective

2-Step, Require SSO, Remove Unlock with PIN, Remove Export policies utilize SimpleTogglePolicyComponent going forward for their drawer implementations so long as the feature flag is turned on.

## 📸 Screenshots

image
image
image
image


libs/components/src/switch/switch.component.ts
```diff
@@ -56,6 +57,11 @@ export class SwitchComponent implements ControlValueAccessor, AfterViewInit {
   protected readonly disabled = model(false);
   protected readonly disabledReasonText = input<string | null>(null);
 
+  /**
+   * When true, the toggle appears before the label (left side) in a compact, non-spreading layout.
+   */
+  readonly reversed = input(false, { transform: booleanAttribute });
```
vleague2: We are about to merge a [feature branch](https://github.com/bitwarden/clients/pull/20425) containing updates to the form controls, including switch. This looks like it will give you the behavior you need without you having to make any changes: [feature branch switch story](https://62a88a6de5b807fa98886113-mwphoknvqh.chromatic.com/?path=/story/component-library-form-switch--default )

Can this PR wait until we merge that? (I am aiming for Monday!!) Sorry for the sequencing issue, it has been challenging to keep designs in sync with what we have and haven't merged.

vleague2 (dismissed):

libs/components/src/switch/switch.component.ts
JaredScar: Can totally wait! @vleague2 Thank you for the knowledge!

BTreston (dismissed):
Generally seems fine. I think the design as it stands now suffers from the drawers initiative and this UI re-design running in parallel with each other, so we now have a decent amount of tech debt and feature flagged code paths to clean up. But I guess that's just how it has to be.

As a side note, all the inheritance going on with policies is highly unintuitive IMO, I'll spend some BEEEP time determining if this can be simplified.

BryanCunningham (dismissed):

eliykat (commented):
This and #21097 have overlapping changes which is best to avoid. I didn't realise this PR was also in flight.

I suggest you merge that PR first, then update this one to match. I'm not sure how much divergence there is between the PRs, but for example this still contains changes to existing components (e.g. policy-edit-dialog and multi-step-policy), which needs to be reconciled. (I preferred to avoid changes to non-feature flagged components.) @BTreston you may want to take a peek at that PR for any major differences compared to your feedback here, or just wait until it's merged and this one is updated.

I don't think this one needs my review unless there's anything specific.

mcamirault:
The changes to Tools-owned files got removed at some point, I think. Feel free to add me directly back on if I'm wrong!

BTreston (approved):

PR #21094, opened 2026-06-05
[PM-38617] Fix Targeting Rules not respecting Fill Assist setting changes in some cases

jprusik:
## 🎟️ Tracking

[PM-38617](https://bitwarden.atlassian.net/browse/PM-38617)

## 📔 Objective

Address some instances in code that might not be fully/properly respecting Fill Assist user setting changes.

- clear pageTargetingRules on page location change
- add defensive check of fill assist setting before generating fill scripts
- page details cache should clear when the Fill Assist setting is changed
- subscribe background overlay to fill assist changes and broadcast targeting rules cache invalidation

[PM-38617]: https://bitwarden.atlassian.net/browse/PM-38617?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/browser/src/autofill/services/autofill.service.ts
```diff
@@ -775,8 +788,18 @@ export default class AutofillService implements AutofillServiceInterface {
...
-    if (hasTargetedFields) {
+    const pageHasTargetedFields = pageDetails.fields.some(({ targeted }) => targeted === true);
+
+    if (pageHasTargetedFields) {
+      const fillAssistEnabled = await this.isFillAssistEnabled();
+
+      // We could alternatively retrigger gathering page details with the
+      // heuristic strategy, but this code path is mostly defensive and not
+      // expected to be hit often, since the entrypoints for this workflow
+      // are also expected to be gated.
```
jprusik, line 786: That is, I suspect the possibility of the gating logic at the workflow entry-points to be fragile enough against future changes to justify defense-in-depth here.

jprusik:
BIt failures addressed by https://github.com/bitwarden/browser-interactions-testing/pull/498

apps/browser/src/autofill/background/overlay.background.ts
```diff
@@ -364,6 +364,21 @@ export class OverlayBackground implements OverlayBackgroundInterface {
...
       .subscribe();
+
+    // Dump targeting rules' cached page details when Fill Assist setting is changed,
+    // and signal content scripts to drop their own targeting-rules caches so the
+    // next page-details collection re-evaluates which strategy to use (targeted vs
+    // heuristic).
+    this.domainSettingsService.enableFillAssist$
+      .pipe(
+        // ignore initial empty state
+        skip(1),
```
audreyality: 🤔 The initial state of `enableFillAssist$` should never be empty. `domain-settings.service.ts:214` maps null outputs to false. Skipping the first entry could cause subscribe to hang if there's never an update to the entry. This can happen in Chrome, which only emits state update events when a stored value changes.

apps/browser/src/autofill/background/overlay.background.ts
```diff
@@ -386,6 +401,42 @@ export class OverlayBackground implements OverlayBackgroundInterface {
...
+        // and should not appear in the same pageDetails
+        if (entry.details?.fields?.some((field) => field.targeted === true)) {
+          frameMap.delete(frameId);
+        }
+      }
+      if (frameMap.size === 0) {
+        delete this.pageDetailsForTab[tabId];
+      }
+    }
+  }
```
audreyality, line 429: ♻️ The complexity of this and similar management logic says to me that the cache should probably be factored into an independent type.

apps/browser/src/autofill/content/autofill-init.ts
```diff
@@ -180,6 +181,18 @@ class AutofillInit implements AutofillInitInterface {
...
+  /**
+   * Drops `pageDetails` built via Targeting Rules and init data collection
+   * with updated strategy.
+   *
+   * `sender: "autofillInit"` updates the cached page details without triggering
+   * an autofill (unlike `"autofiller"` or autofill-command senders).
+   */
+  private handleClearTargetingRulesCache(): void {
+    this.collectAutofillContentService.clearCachedTargetingRules();
+    void this.sendExtensionMessage("bgCollectPageDetails", { sender: "autofillInit" });
```
audreyality: ❓ Why not just call `AutofillInit.collectPageDetails(...)`?

apps/browser/src/autofill/services/autofill.service.ts
```diff
@@ -424,6 +425,18 @@ export default class AutofillService implements AutofillServiceInterface {
...
+    const fillAssistFeatureEnabled = await this.configService.getFeatureFlag(
+      FeatureFlag.FillAssistTargetingRules,
+    );
+
+    if (!fillAssistFeatureEnabled) {
+      return false;
+    }
+
+    return await firstValueFrom(this.domainSettingsService.enableFillAssist$);
+  }
```
audreyality: 🎨 `domain-settings-service.ts` already imports `FeatureFlag` and `ConfigService`. Folding this logic into the domain settings keeps the logic encapsulated and ensures all readers of `enableFillAssist$` get a consistent view of its enabled status.


```ts
this.enableFillAssistState = this.stateProvider.getGlobal(ENABLE_FILL_ASSIST);
this.enableFillAssist$ = combineLatest(
  this.configService.getFeatureFlag$(FeatureFlag.FillAssistTargetingRules),
  this.enableFillAssistState.state$
).pipe(
  map(([feature, enabled]) => feature && enabled ?? false),
  distinctUntilChanged()
);
```

audreyality (commented):
📝 Some questions I'd like answered before approving; no blocking feedback.

apps/browser/src/autofill/services/autofill.service.ts
jprusik: Agreed, and is the pattern I've used in the past for Autofill settings. The one bit of nuance is making clear distinctions between the originating state and the resolved state, especially if you need the unresolved state for more than just reflecting in the setting UI (or if you have multiple UIs that control the setting).

apps/browser/src/autofill/content/autofill-init.ts
jprusik: Nice catch - absolutely we should. There are a few other places (unrelated to this) that we should examine to eliminate unnecessary messaging as well, but that's a whole other expedition.

libs/common/src/autofill/services/domain-settings.service.ts
```diff
@@ -213,6 +229,19 @@ export class DefaultDomainSettingsService implements DomainSettingsService {
...
     this.enableFillAssist$ = this.enableFillAssistState.state$.pipe(map((x) => x ?? false));
 
+    // `defer` ensures each subscriber gets a fresh `getFeatureFlag$` call,
+    // so tests that mutate the config mock after service construction see
+    // the updated value.
+    this.resolvedEnableFillAssist$ = defer(() =>
+      combineLatest([
+        this.enableFillAssist$,
+        this.configService.getFeatureFlag$(FeatureFlag.FillAssistTargetingRules),
+      ]),
```
audreyality: 💭 Seems like this `defer` is only needed because you're using `of` in your mocks. If your mocks use a `SubjectLike`, then you can simply emit the signal you want without changing the mock.

apps/browser/src/autofill/background/overlay.background.ts
```diff
@@ -364,6 +366,23 @@ export class OverlayBackground implements OverlayBackgroundInterface {
...
+
+    // Dump targeting rules' cached page details when Fill Assist becomes
+    // disabled, and signal content scripts to drop their own targeting-rules
+    // caches so the next page-details collection re-evaluates which strategy
+    // to use (targeted vs heuristic). Only act on a `true` -> `false`
+    // transition so service-worker cold starts (where the replayed initial
+    // value is `false`) don't broadcast.
+    this.domainSettingsService.resolvedEnableFillAssist$
+      .pipe(
+        distinctUntilChanged(),
```
audreyality: ⛏️ This `distinctUntilChanged()` is handled within `resolvedEnableFillAssist$`

```suggestion
```

audreyality (dismissed):

libs/common/src/autofill/services/domain-settings.service.ts
jprusik: addressed in 853e10654f8d0e5fb5aaae66ed86406c58b9271a

apps/browser/src/autofill/background/overlay.background.ts
jprusik: addressed in 853e10654f8d0e5fb5aaae66ed86406c58b9271a

audreyality (dismissed):

audreyality (approved):

PR #21097, opened 2026-06-05
[PM-38345] block account, remove card, remove free fam, and desktop autotype to simple toggle

JaredScar:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-38345

## 📔 Objective

Fixes up policy drawer UI for block account, remove card, remove free families, and desktop autotype policy to utilize SimpleTogglePolicyComponent going forward

## 📸 Screenshots

image
image
image
image


bitwarden_license/bit-web/src/app/admin-console/policies/policy-edit-definitions/block-claimed-domain-account-creation.component.ts
```diff
@@ -1,20 +1,31 @@
...
   name = "blockClaimedDomainAccountCreation";
+  nameV2 = "blockClaimedDomainAccountCreationTitleV2";
   description = "blockClaimedDomainAccountCreationDesc";
+  descriptionV2 = "blockClaimedDomainAccountCreationDescV2";
   type = PolicyType.BlockClaimedDomainAccountCreation;
   category = PolicyCategory.Authentication;
   priority = 60;
   component = BlockClaimedDomainAccountCreationPolicyComponent;
+  prerequisiteKey = "blockClaimedDomainAccountCreationPrerequisiteV2";
+  prerequisiteLinkHref = "https://bitwarden.com/help/domain-verification/";
```
eliykat: Per the Figma should be: https://bitwarden.com/help/claimed-domains/

bitwarden_license/bit-web/src/app/admin-console/policies/policy-edit-definitions/block-claimed-domain-account-creation.component.ts
```diff
@@ -1,20 +1,31 @@
...
   BasePolicyEditDefinition,
   BasePolicyEditComponent,
   PolicyCategory,
 } from "@bitwarden/web-vault/app/admin-console/organizations/policies";
+import { SimpleTogglePolicyComponent } from "@bitwarden/web-vault/app/admin-console/organizations/policies/policy-edit-definitions/simple-toggle-policy.component";
 import { SharedModule } from "@bitwarden/web-vault/app/shared";
 
 export class BlockClaimedDomainAccountCreationPolicy extends BasePolicyEditDefinition {
   name = "blockClaimedDomainAccountCreation";
+  nameV2 = "blockClaimedDomainAccountCreationTitleV2";
```
eliykat: I think you previously pointed out that this differs from the name in the list (Block account creation for claimed domains vs. Claimed domain restriction). Did you get confirmation from Jillian that this is intended?

bitwarden_license/bit-web/src/app/admin-console/policies/policy-edit-definitions/block-claimed-domain-account-creation.component.ts
eliykat, line 1: The callout on this page is missing the Prerequisite heading.

bitwarden_license/bit-web/src/app/billing/policies/free-families-sponsorship.component.ts
```diff
@@ -1,26 +1,35 @@
...
   BasePolicyEditDefinition,
   BasePolicyEditComponent,
   PolicyCategory,
 } from "@bitwarden/web-vault/app/admin-console/organizations/policies";
-import { SharedModule } from "@bitwarden/web-vault/app/shared";
+import { SimpleTogglePolicyComponent } from "@bitwarden/web-vault/app/admin-console/organizations/policies/policy-edit-definitions/simple-toggle-policy.component";
 
 export class FreeFamiliesSponsorshipPolicy extends BasePolicyEditDefinition {
   name = "freeFamiliesSponsorship";
+  nameV2 = "freeFamiliesSponsorshipPolicyTitleV2";
```
eliykat: This also differs from the main list (option vs. sponsorship) - please double check any divergences and confirm with Jillian.

apps/web/src/app/admin-console/organizations/policies/base-policy-edit.component.ts
```diff
@@ -81,12 +95,37 @@ export abstract class BasePolicyEditDefinition {
...
    * Used by {@link SimpleTogglePolicyComponent} to avoid per-policy component boilerplate.
    */
   warningKey?: string;
 
+  /**
+   * Optional alternative component to use when a feature flag is enabled.
+   * When set, the {@link PolicyEditDialogComponent} checks the flag at dialog open time
+   * and uses this component instead of {@link component} when the flag is on.
+   */
+  flaggedComponent?: { flag: FeatureFlag; component: Constructor<BasePolicyEditComponent> };
```
eliykat: I like the idea of the nested object for the v2 configuration. It keeps the old & new separate rather than scattering v2 properties through the main object.

I suggest renaming this `v2` and then moving all new properties into this object.

apps/web/src/app/admin-console/organizations/policies/base-policy-edit.component.ts
```diff
@@ -81,12 +95,37 @@ export abstract class BasePolicyEditDefinition {
    **/
   showDescription: boolean = true;
 
+  /**
+   * Optional i18n key for a prerequisite info callout rendered above the enable/disable toggle.
+   * Used by {@link SimpleTogglePolicyComponent} when a policy requires another policy to be enabled first.
```
eliykat: This is rendered in the `policy-edit-dialog.component.ts` so I don't think it's accurate that it's only used by the SimpleToggle component.

apps/web/src/app/admin-console/organizations/policies/base-policy-edit.component.ts
```diff
@@ -81,12 +95,37 @@ export abstract class BasePolicyEditDefinition {
...
+   */
+  prerequisiteKey?: string;
```
eliykat: Does this change between policies, or does it always just say "Prerequisite"? If it's always the same it can be hard coded into the component.

apps/web/src/app/admin-console/organizations/policies/policy-edit-dialog.component.html
eliykat, line 1: Please make this a new component, `policy-edit-drawer.component`, which implements the new layout. Then we can switch between which component is opened, rather than shooting this through with conditional logic.

apps/web/src/app/admin-console/organizations/policies/policy-edit-definitions/simple-toggle-policy.component.spec.ts
```diff
@@ -0,0 +1,110 @@
...
+      schemas: [NO_ERRORS_SCHEMA],
+    }).compileComponents();
+
+    fixture = TestBed.createComponent(SimpleTogglePolicyComponent);
+    component = fixture.componentInstance;
+  });
+
+  describe("warning callout", () => {
+    it("should not render a callout when no warningKey is set", () => {
+      const policy = new TwoFactorAuthenticationPolicy(); // warningKey is undefined
```
eliykat: Better to define your own test fixtures. Otherwise if someone actually does add a warningKey to the 2FA policy, this test will fail, even though the behavior hasn't changed.

apps/web/src/app/admin-console/organizations/policies/policy-edit-definitions/simple-toggle-policy.component.spec.ts
```diff
@@ -0,0 +1,110 @@
...
+    it("should render a callout when warningKey is set on the policy", () => {
+      const policy = Object.assign(new TwoFactorAuthenticationPolicy(), {
+        warningKey: "twoStepLoginPolicyWarning",
+      });
+      fixture.componentRef.setInput("policy", policy);
+      fixture.componentRef.setInput("policyResponse", makePolicyResponse(false));
+      component.ngOnInit();
+      fixture.detectChanges();
+
+      expect(fixture.debugElement.query(By.css("bit-callout"))).not.toBeNull();
```
eliykat: I would also check that it has rendered the expected contents. e.g. this passes today, but the heading isn't displayed.

apps/web/src/app/admin-console/organizations/policies/policy-edit-definitions/simple-toggle-policy.component.spec.ts
```diff
@@ -0,0 +1,110 @@
...
+
+      const result = await component.buildRequest(mock<OrgKey>());
+
+      expect(result).toEqual({
+        policy: { enabled: false, data: null },
+        metadata: null,
+      });
+    });
+
+    it("should return enabled: false when policy is disabled and then re-enabled via the form", async () => {
```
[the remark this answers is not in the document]
eliykat: Agree with Claude here, but I would make it more succinct: `should return enabled: true when policy is enabled via the form`.

eliykat (changes requested):
Please also include basic stories for each concrete policy component (just 1 each because they are so simple). Maybe render them inside the outer policy-edit component so that it shows us how they compose. (Less granular but more useful to reflect how they actually appear in the app.)

This also saves you from including screenshots because you can just link to the stories 😁

apps/web/src/app/admin-console/organizations/policies/policy-edit-definitions/remove-unlock-with-pin.component.ts
eliykat, line 1: There are several policies that have been added here but are outside the scope of this ticket. Maybe you forgot to switch branches? Please tidy up to match ticket scope.

apps/web/src/app/admin-console/organizations/policies/policies.component.ts
```diff
@@ -179,13 +180,19 @@ export class PoliciesComponent {
     const dialogComponent: PolicyDialogComponent =
       policy.editDialogComponent ?? PolicyEditDialogComponent;
 
-    if (useDrawer && dialogComponent.openDrawer) {
+    // For custom dialog components (e.g. MultiStep), prefer their own openDrawer.
+    // For standard policies, route all drawer opens through PolicyEditDrawerComponent.
+    const drawerOpener = useDrawer
+      ? (dialogComponent.openDrawer ?? PolicyEditDrawerComponent.openDrawer)
```
eliykat: I think the `?? PolicyEditDrawerComponent.openDrawer` is unnecessary - you already coalesce `policy.editDialogComponent ?? PolicyEditDialogComponent` above, so you either way you should get a component that implements `openDrawer`.

apps/web/src/app/admin-console/organizations/policies/policy-edit-dialog.component.ts
eliykat, line 1: Are these remaining changes necessary now that we have a separate drawer component?

apps/web/src/app/admin-console/organizations/policies/policy-edit-drawer.component.ts
```diff
@@ -0,0 +1,28 @@
...
+import { BasePolicyEditComponent } from "./base-policy-edit.component";
+import { PolicyEditDialogComponent } from "./policy-edit-dialog.component";
+import type { PolicyEditDialogData, PolicyEditDialogResult } from "./policy-edit-dialog.component";
+
+@Component({
+  templateUrl: "policy-edit-drawer.component.html",
+  imports: [SharedModule],
+  changeDetection: ChangeDetectionStrategy.OnPush,
+})
+export class PolicyEditDrawerComponent extends PolicyEditDialogComponent {
```
eliykat: This shouldn't subclass the existing component; it's a new implementation. That should also let you revert the remaining changes to the dialog component as well. Some code duplication here is OK just so that the v2 implementation is self-contained.

apps/web/src/app/admin-console/organizations/policies/policy-edit-drawer.component.html
```diff
@@ -0,0 +1,48 @@
...
+      <div>
+        <i
+          class="bwi bwi-spinner bwi-spin tw-text-muted"
+          title="{{ 'loading' | i18n }}"
+          aria-hidden="true"
+        ></i>
+        <span class="tw-sr-only">{{ "loading" | i18n }}</span>
+      </div>
+    }
+    <div [hidden]="loading()">
```
eliykat: Use `@if` rather than `hidden`. Really it should just be an `else { }` block on the previous `@if`.

apps/web/src/app/admin-console/organizations/policies/policy-edit-drawer.component.html
```diff
@@ -0,0 +1,48 @@
...
+    @if (loading()) {
+      <div>
+        <i
+          class="bwi bwi-spinner bwi-spin tw-text-muted"
+          title="{{ 'loading' | i18n }}"
+          aria-hidden="true"
+        ></i>
+        <span class="tw-sr-only">{{ "loading" | i18n }}</span>
+      </div>
+    }
```
eliykat: Use the `bit-spinner` component: https://components.bitwarden.com/?path=/docs/component-library-spinner--docs

eliykat (changes requested):
All my previous feedback has been addressed, except for:
- policy name changes (waiting on design I think)
- storybooks

Some additional feedback below. Thanks!

bitwarden_license/bit-web/src/app/admin-console/policies/policy-edit-definitions/block-claimed-domain-account-creation.component.ts
JaredScar: Design and product have came to terms with the changes. Updating now!

bitwarden_license/bit-web/src/app/billing/policies/free-families-sponsorship.component.ts
JaredScar: Design and product have came to terms with the changes. Updating now!

apps/web/src/app/admin-console/organizations/policies/base-policy-edit.component.ts
```diff
@@ -87,6 +88,35 @@ export abstract class BasePolicyEditDefinition {
    */
   warningKey?: string;
 
+  /**
+   * When true, the cancel button is hidden in the drawer footer.
+   */
+  hideCancelButton?: boolean;
```
eliykat: I expected that this would be consistent between policies - which policy does it need to be configurable for?

apps/web/src/app/admin-console/organizations/policies/base-policy-edit.component.ts
```diff
@@ -87,6 +88,35 @@ export abstract class BasePolicyEditDefinition {
...
+    /** Drawer-only description. Falls back to {@link description} when not set. */
+    description?: string;
+    /** i18n key for a prerequisite info callout rendered by {@link PolicyEditDialogComponent} above the policy form. */
+    prerequisiteKey?: string;
+    /** URL for an optional "learn more" link inside the prerequisite callout. */
+    prerequisiteLinkHref?: string;
+    /** i18n key for the text of {@link prerequisiteLinkHref}. */
+    prerequisiteLinkTextKey?: string;
+    /** i18n key for the switch label. Defaults to "turnOn" when not set. */
+    labelKey?: string;
```
eliykat: I expected that this would be consistent between policies - which policy does it need to be configurable for?

apps/web/src/app/admin-console/organizations/policies/policy-edit-definitions/require-sso.component.ts
eliykat, line 1: The PR still has edits for unrelated policies - please revert any policies that aren't in scope for the ticket, unless they're needed to prevent a regression or something (but these changes should be additive only).

apps/web/src/locales/en/messages.json
```diff
@@ -13462,13 +13633,13 @@
     "message": "Block account creation for claimed domains"
   },
   "blockClaimedDomainAccountCreationTitleV2": {
-    "message": "Claimed domain restriction"
+    "message": "Block account creation for claimed domains"
```
eliykat: Identical to the existing i18n key - just double check you're not adding any duplicates.

apps/web/src/app/admin-console/organizations/policies/policy-edit-dialogs/multi-step-policy-edit-dialog.component.ts
```diff
@@ -131,8 +122,6 @@ export class MultiStepPolicyEditDialogComponent
     // Read step configuration from child component.
     // Setting policySteps triggers currentStepConfig to recompute, which re-evaluates saveDisabled.
     this.policySteps.set(component.policySteps ?? []);
-
-    await this.setupDiscardGuard();
   }
```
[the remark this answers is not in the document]
eliykat: Please double check this Claude feedback. I haven't looked into it that closely but I am also concerned about removing existing functionality.

apps/web/src/app/admin-console/organizations/policies/policy-edit-dialog.component.ts
```diff
@@ -76,121 +69,23 @@ export class PolicyEditDialogComponent implements AfterViewInit {
...
-   *
-   * Call this once the child policy component has been initialised.
-   */
-  protected async setupDiscardGuard(): Promise<void> {
-    this.discardGuardEnabled.set(
-      await this.configService.getFeatureFlag(FeatureFlag.PolicyDrawers),
-    );
-    if (!this.discardGuardEnabled()) {
-      return;
-    }
```
eliykat, line 133: Again I'm concerned about removing current functionality - but even though the logic applies to both drawers and modals, it has an early return at the top if drawers aren't enabled. How does this work? Did we ever support the discard guard for modals, or is it drawers only?

apps/web/src/app/admin-console/organizations/policies/policy-edit-drawer.component.ts
```diff
@@ -0,0 +1,262 @@
...
+    cancelButtonText: { key: "backToEditing" },
+  };
+
+  private async setupDiscardGuard(): Promise<void> {
+    this.discardGuardEnabled.set(
+      await this.configService.getFeatureFlag(FeatureFlag.PolicyDrawers),
+    );
+    if (!this.discardGuardEnabled()) {
+      return;
+    }
```
eliykat: This component is only used if the flag is enabled, so I think you can remove this.

bitwarden_license/bit-web/src/app/admin-console/policies/policy-edit-definitions/block-claimed-domain-account-creation.component.stories.ts
```diff
@@ -0,0 +1,130 @@
...
+        }
+        <p bitTypography="body1">{{ descriptionKey | i18n }}</p>
+        <app-simple-toggle-policy-edit
+          [policy]="policyDef"
+          [policyResponse]="policyResponse"
+        ></app-simple-toggle-policy-edit>
+      </div>
+    `,
+  };
+}
```
eliykat: This is reimplementing enough template that it's no longer a useful reflection of how the app actually looks. Storybooks should be testing actual components, otherwise we are only testing test code.

My thinking: have the story render the `PolicyEditDrawerComponent`, but using helper code that parameterizes what policy it renders within it. That will give you a real world representation of how the drawer renders. You'll still need 1 story per policy, but they'll all share the same setup so that each story will only be a few lines.

Untested example:

```ts
// Common helper function
export function policyDrawerMeta(title: string, policy: BasePolicyEditDefinition): Meta<PolicyDrawerStoryArgs> {
  return {
    title,
    component: PolicyEditDrawerComponent,
    args: { enabled: false },
    argTypes: { enabled: { control: "boolean" } },
    decorators: [
      moduleMetadata({
        providers: [
          { provide: DIALOG_DATA, useValue: { policy, organization: { id: ORG_ID } } },
          { provide: DialogRef, useValue: { close: () => {}, closePredicate: undefined } },
          // ...AccountService, KeyService, AuthService, DialogService mocks
          // Flag off → setupDiscardGuard short-circuits, no auth/account wiring needed.
          { provide: ConfigService, useValue: { getFeatureFlag: () => Promise.resolve(false) } },
        ],
      }),
      applicationConfig({ providers: [importProvidersFrom(PreloadedEnglishI18nModule)] }),
    ],
    // Per-story override so the `enabled` arg drives the initial load.
    render: (args) => ({
      moduleMetadata: {
        providers: [{
          provide: PolicyApiServiceAbstraction,
          useValue: {
            getPolicy: () => Promise.resolve(new PolicyResponse({ Enabled: args.enabled })),
            putPolicy: () => Promise.resolve(),
          },
        }],
      },
    }),
  };
}
```

```ts
// Per-policy story file

export default policyDrawerMeta(
  "Admin Console/Organizations/Policies/Block Claimed Domain Account Creation",
  new BlockClaimedDomainAccountCreationPolicy(),
);

export const PolicyOff: StoryObj<PolicyDrawerStoryArgs> = {};
export const PolicyOn: StoryObj<PolicyDrawerStoryArgs> = { args: { enabled: true } };
```

eliykat (changes requested):

apps/web/src/app/admin-console/organizations/policies/policy-edit-dialog.component.ts
JaredScar: @eliykat The discard guard was written to handle both modals and drawers, but drawers are the going-forward UX. The guard was also only intentionally scoped to affect drawers only. Therefore ridding of this code actually should be okay!

amorask-bitwarden (dismissed):
Billing changes look good.

libs/components/src/dialog/dialog/dialog.component.html
```diff
@@ -11,12 +11,13 @@
 >
   @let showHeaderBorder = bodyHasScrolledFrom().top;
   <header
-    class="tw-flex tw-justify-between tw-items-center tw-gap-4 tw-border-0 tw-border-b tw-border-solid tw-py-3 tw-ps-6 tw-pe-4"
```
harr1424, line 14: Why is this being changed?

harr1424 (dismissed):
Tools Owned files good with one non-blocking question

libs/components/src/dialog/dialog/dialog.component.html
vleague2: This looks like a small style change that will be covered by our design system refresh updates for this component. We haven't gotten to dialogs/drawers yet. 

We can revert the changes to this file. UIF will do it with the other refresh changes once we get to that component!

apps/web/src/app/admin-console/organizations/policies/base-policy-edit.component.ts
```diff
@@ -87,6 +88,33 @@ export abstract class BasePolicyEditDefinition {
    */
   warningKey?: string;
 
+  /**
+   * @deprecated No longer read by the drawer template. The cancel button is always hidden in drawers.
+   */
+  hideCancelButton?: boolean;
```
eliykat: You are adding this and deprecating it in the same PR. Remove it if you don't need it.

libs/common/src/vault/abstractions/file-upload/cipher-file-upload.service.ts
eliykat, line 1: There are various changes to unrelated code in `libs/common` - please revert. If prettier/eslint is doing this, double check you're using the correct version in `package.json`.

eliykat (commented):
A couple of quick notes, plus [UIF feedback above](https://github.com/bitwarden/clients/pull/21097#discussion_r3422761732) needs to be addressed and possibly clarified with Jillian. 

Please also make sure linting passes and Storybooks are all compiling and look as expected. Then I can re-review to take a closer look at the remaining changes.

libs/components/src/dialog/dialog/dialog.component.html
JaredScar: Currently with this, a 1 pixel border which isn't really visible is added to the dialog, but also affects the drawer. Design didn't like this extra pixel, so they asked for it to be removed. I'm fine with getting rid of the change here, but we definitely want to make sure it's communicated to Design too. I already messaged Jillian about this though and will reiterate her response one I get it!

JaredScar: image
Closing loop here. Jillian confirmed that UIF can handle these updates via their design refresh. Thanks for the comments! @harr1424 @vleague2 -- Removing the changes to this file and marking resolved

vleague2: @JaredScar Thanks for checking!!

apps/web/src/app/admin-console/organizations/policies/policy-edit-dialog.component.ts
eliykat, line 1: The discard guard changes are incomplete:
- the guard has been partially but not fully removed from this component (`setupDiscardGuard` and `discardGuardEnabled` are still here)
- it seems to have been shifted down into the multi-step component, even though I think that is also a dialog (not a drawer) so presumably it doesn't need it? Or is the multi-step treated differently?

I think:
- if neither component needs it - remove it completely
- if the multi-step component does still need it, or if there are any complications - then let's not touch any of these old components, revert any changes and let the discard guard be dead code. Ideally we want all our changes behind feature flags, if we're refactoring old components then that is not covered by feature flags; it also seems like unnecessary refactoring of components that we're going to delete soon.

Let me know if you want to chat about it.

apps/web/src/app/admin-console/organizations/policies/policy-edit-dialogs/multi-step-policy-edit-dialog.component.ts
eliykat, line 1: Previous comment also applies to this file.

apps/web/src/app/admin-console/organizations/policies/policy-edit-drawer.component.ts
```diff
@@ -0,0 +1,250 @@
...
+        }),
+        takeUntilDestroyed(this.destroyRef),
+      )
+      .subscribe(() => {
+        this.discardGuardEnabled.set(false);
+        this.dialogRef.closePredicate = undefined;
+      });
+  }
+
+  protected readonly cancel = async () => {
```
eliykat: Cancel doesn't appear to be called from anywhere.

apps/web/src/app/admin-console/organizations/policies/policy-edit-drawer.component.ts
```diff
@@ -0,0 +1,250 @@
...
+    this.policyComponent.set(component);
+    this.policyEnabled.set(policyResponse.enabled);
+
+    if (component.data) {
+      component.data.statusChanges
+        .pipe(
+          map((status) => status === "INVALID" || !policyResponse.canToggleState),
+          takeUntilDestroyed(this.destroyRef),
+        )
+        .subscribe((disabled) => this._saveDisabled.set(disabled));
```
eliykat: It looks like this logic has been updated in `main` since you opened this PR - please check to see if we need to bring it across here.

eliykat (commented):
Please fix the storybook build and let me know if you need any help with it.

cyprain-okeke (dismissed):
Looks good for billing changes

amorask-bitwarden (dismissed):

harr1424 (dismissed):

eliykat (dismissed):
LGTM, please review and approve the Storybook changes before merging.

eliykat (approved):

harr1424 (approved):

amorask-bitwarden (approved):

PR #21098, opened 2026-06-05
[PM-38635] [PM-37790] fix(fido2): consult vault before transport-only fallback

### LINKED ISSUE #20973: Passkey support completely broken when 1Password extension is co-installed (blocking for users in enforced work environments)
### Steps To Reproduce

1. On macOS (also reproduced on Linux), use Google Chrome 148.0.7778.179.
2. Install the Bitwarden browser extension 2026.4.1.
3. Install the 1Password browser extension alongside it (in many corporate environments this is mandated by policy and cannot be disabled by the end user).
4. Have a passkey stored in Bitwarden for a site that supports passkeys (e.g., github.com).
5. Trigger a passkey flow on that site (e.g., GitHub sudo-mode re-auth, or "Sign in with a passkey").

### Expected Result

Bitwarden either intercepts the WebAuthn request and presents its own passkey picker, or — when another passkey provider is also registered — the browser shows a chooser that allows the user to pick Bitwarden. The Bitwarden-stored passkey should remain usable.

### Actual Result

Chrome's native passkey dialog appears and routes the request to a non-Bitwarden provider (1Password / iCloud Keychain). Bitwarden never gets a chance to surface its passkey, even though the credential exists in the Bitwarden vault and Bitwarden is configured as the default password manager in Chrome. Passkeys stored in Bitwarden become effectively unusable for as long as the 1Password extension is enabled.

Disabling the 1Password extension restores Bitwarden's passkey interception. For users whose organizations mandate 1Password via MDM / work policy, there is **no workaround** — Bitwarden passkeys cannot be used at all.

### Screenshots or Videos

_No response_

### Additional Context

This is a follow-up to #20743, which documents the broader Chrome 146+ passkey-interception regression. The comment thread (see https://github.com/bitwarden/clients/issues/20743#issuecomment-4587926129 and surrounding discussion) makes it clear that the failure mode is now reliably tied to having a second passkey-capable extension (1Password) installed.

Key points that justify a separate report:

- This is not just "two password managers don't coexist nicely" — it is a complete breakage of Bitwarden passkey functionality, not a degraded UX.
- It affects users who **cannot** uninstall or disable 1Password (enforced work profiles, MDM-managed browsers, locked corporate extensions).
- It is reproducible across operating systems (macOS, Linux, Windows reports also in the thread) and is not limited to a single Chrome/Bitwarden version combination.
- A community fix has been proposed in #20849 which migrates Bitwarden to Chrome's official `chrome.webAuthenticationProxy` API (available since Chrome 115). Adopting that API would let Chrome route WebAuthn requests to Bitwarden deterministically instead of relying on content-script interception that loses the race to 1Password / the native picker.
- 1Password support has reportedly told affected users that the responsibility lies with Bitwarden, so a coordinated fix on the Bitwarden side is needed.

Related: #20743, #20849, #14720.

### Operating System

macOS, Linux, Windows

### Operating System Version

macOS 15.7.3 (24G419) — also reproduced on macOS 26.3.1, Linux, and Windows 11 per #20743 thread

### Web Browser

Chrome

### Browser Version

Google Chrome 148.0.7778.179 (Official Build, arm64)

### Environment Versions

Extension 2026.4.1 (bitwarden.com and bitwarden.eu both affected)

### Issue Tracking Info

xrvk:
## 🎟️ Tracking

[PM-37790](https://bitwarden.atlassian.net/browse/PM-37790)

Fixes #20973, related to #20743.

## 📔 Objective

Vault-resident passkeys are unusable in Chrome 146+ when another passkey provider (1Password, iCloud Keychain) is also attached, because Bitwarden hands the request off to the browser before checking whether the credential is in the vault.

[PM-33781](https://bitwarden.atlassian.net/browse/PM-33781) added a fallback at the top of `Fido2ClientService.assertCredential` that throws `FallbackRequestedError` whenever every entry in `allowCredentials` advertises only non-internal transports (`usb` / `nfc` / `ble` / `hybrid`). The intent was correct: a request that only allows roaming-key transports cannot be served by an internal-only authenticator, so the browser should handle it. The check, however, does not consult the vault. Relying parties commonly echo the transports recorded at registration time, so a passkey that was originally registered via cross-device hybrid flow and is now synced into the vault arrives with transports such as `["hybrid"]` and trips the fallback. Chrome 146+ then routes the request to its native picker, which surfaces whichever other provider is attached.

Fix: consult the vault first via `silentCredentialDiscovery(rpId)` and only throw `FallbackRequestedError` when none of the requested credential ids match a vault credential. Behavior for credentials that are genuinely not in the vault is unchanged. Vault credential ids are normalized via `parseCredentialId` so both UUID and `b64.<base64url>` forms are handled. Discovery failures are caught and surfaced as `FallbackRequestedError` to keep the page-script flow deterministic.

Coverage in `fido2-client.service.spec.ts` (`fallback for hardware/roaming keys`):
- vault match in UUID form proceeds with Bitwarden
- vault match in `b64.<base64url>` form proceeds with Bitwarden
- no vault match falls back
- `silentCredentialDiscovery` rejection falls back
- existing tests for internal transports and empty transport lists are unchanged in intent and now also assert that discovery is not invoked

## 📸 Screenshots

N/A.

[PM-33781]: https://bitwarden.atlassian.net/browse/PM-33781?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/common/src/platform/services/fido2/fido2-client.service.spec.ts
+++ b/libs/common/src/platform/services/fido2/fido2-client.service.spec.ts
@@ -654,6 +654,24 @@ describe("FidoAuthenticatorService", () => {
         expect(authenticator.getAssertion).toHaveBeenCalled();
       });
 
+      it("matches vault credentials stored in b64.<base64url> form", async () => {
+        const rawBytes = randomBytes(16);
+        const allowedId = Fido2Utils.arrayToString(rawBytes);
+        const credentialId = "b64." + allowedId;
+        const params = createParams({
+          allowedCredentials: [{ id: allowedId, transports: ["usb"] }],
+          fallbackSupported: true,
+        });
+        authenticator.silentCredentialDiscovery.mockResolvedValue([
+          { credentialId } as Fido2CredentialView,
+        ]);
+        authenticator.getAssertion.mockResolvedValue(createAuthenticatorAssertResult());
+
+        await client.assertCredential(params, windowReference);
+
+        expect(authenticator.getAssertion).toHaveBeenCalled();
+      });
+
       it("falls back when allowCredentials are non-internal and the vault holds different credentials", async () => {
         const params = createParams({
           allowedCredentials: [{ id: "requestedCredId", transports: ["hybrid"] }],
@@ -662,6 +680,19 @@ describe("FidoAuthenticatorService", () => {
         authenticator.silentCredentialDiscovery.mockResolvedValue([
           { credentialId: Utils.newGuid() } as Fido2CredentialView,
         ]);
+
+        await expect(client.assertCredential(params, windowReference)).rejects.toMatchObject({
+          fallbackRequested: true,
+        });
+        expect(authenticator.getAssertion).not.toHaveBeenCalled();
+      });
+
+      it("falls back when silentCredentialDiscovery throws", async () => {
+        const params = createParams({
+          allowedCredentials: [{ id: "requestedCredId", transports: ["hybrid"] }],
+          fallbackSupported: true,
+        });
+        authenticator.silentCredentialDiscovery.mockRejectedValue(new Error("vault unavailable"));
 
         await expect(client.assertCredential(params, windowReference)).rejects.toMatchObject({
           fallbackRequested: true,
--- a/libs/common/src/platform/services/fido2/fido2-client.service.spec.ts
+++ b/libs/common/src/platform/services/fido2/fido2-client.service.spec.ts
@@ -6,6 +6,7 @@ import { AuthenticationStatus } from "../../../auth/enums/authentication-status"
 import { DomainSettingsService } from "../../../autofill/services/domain-settings.service";
 import { Utils } from "../../../platform/misc/utils";
 import { VaultSettingsService } from "../../../vault/abstractions/vault-settings/vault-settings.service";
+import { Fido2CredentialView } from "../../../vault/models/view/fido2-credential.view";
 import { ConfigService } from "../../abstractions/config/config.service";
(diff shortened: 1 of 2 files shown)


micahblut:
Hi @xrvk, thank you for your contribution. Unfortunately, we will not be accepting this as is. It introduces a vulnerability allowing a compromised site to enumerate what credentials Bitwarden does or does not have in the vault. If you look at the comments for #19606, this was initially an issue with that contribution also and something we explicitly avoided.

xrvk:
Thanks for the context on the prior #19606 thread, that makes the constraint clear. Quick note on what's driving this: the goal here is mainly to fix #20973, where Bitwarden passkeys become completely unusable in MDM-locked corporate environments running Chrome 146+ alongside 1Password (users physically cannot disable the other extension).

I've opened #21128 with a revised approach that I believe avoids the enumeration vector. Instead of matching `allowCredentials[].id` values against the vault, it calls `silentCredentialDiscovery(rpId)` and only falls back when the vault has **no** credentials for that rpId at all. The decision is made on the rpId alone, never on individual credential ids.

The browser binds rpId to the calling origin, so a malicious site can only query its own rpId, and the only thing it can learn is "does this user have any Bitwarden passkey for me", which the RP already knows from its own registration records. The sites #19606 was protecting (no Bitwarden passkey at all) still fall back cleanly.

Happy to iterate further if there's still an angle I'm missing.

PR #21124, opened 2026-06-08
[PM-38718] Add dev-only setting to disable process reload

### LINKED ISSUE #11045: Automatic Vault Lock Forces Workspace Change
### Steps To Reproduce

1. Open Bitwarden in any workspace (we'll say workspace 1).
2. Set a vault timeout (Edit > Settings > Security > Vault Timeout).  Mine is set for 2 minutes.
3. Open a browser on workspace 2 and go to a website that requires username and password saved in your vault.
4. Copy and paste login credentials from Bitwarden, and begin doing whatever you need to do on said website.

### Expected Result

Bitwarden locks my vault after 2 minutes.

### Actual Result

Bitwarden does lock my vault after 2 minutes.  _However,_ when it locks, it automatically forces my screen back to workspace 1.  

### Screenshots or Videos

_No response_

### Additional Context

I'm unsure if this is actually a Bitwarden problem.  This behavior is new, but that doesn't necessarily mean that it's from an update to Bitwarden.  My first thought was that it's either a Bitwarden problem or a problem with my window manager (kwin).  From what I've been able to determine so far, it doesn't seem to be a kwin problem, but I'm still investigating.  Perhaps this is neither a Bitwarden nor a kwin problem, but at this time I believe Bitwarden to be the most likely culprit.  I don't know if it would be from upstream Bitwarden or if it has something to do with how the Arch Linux devs packaged the binary in their repos.  I'll be looking into this as well.

### Operating System

Linux

### Operating System Version

Arch Linux using Linux 6.10.9-arch1-2 kernel.

### Installation method

Other

### Build Version

Version 2024.8.1 Shell 31.6.0 Renderer 126.0.6478.234 Node 20.17.0 Architecture x64

### Issue Tracking Info

### LINKED ISSUE #17107: Linux Desktop app spawn to front when timeout
### Steps To Reproduce

The Desktop App Window is spawned to front when it reaches the timeout, which is annoying:
You fill out some password by getting it from the Desktop app. You work on with the app that needed the password. The Bitwarden app window is still open (but not in focus anymore) and might be even hidden by another window. 

### Expected Result

The app should just lock but stay in background/unfocused if it was already in that state.

### Actual Result

When the timeout is reached the Bitwarden app correctly locks. 
However, at the same moment it locks, the app window (with the login) is automatically put frontmost on top of all other open windows, even if you are in the middle of typing in another app. 

### Screenshots or Videos

_No response_

### Additional Context

_No response_

### Operating System

Linux

### Operating System Version

Fedora 42

### Installation method

### Build Version

25.10.0

### Issue Tracking Info

### LINKED ISSUE #18463: Linux Client App Dumps Core: Crashing because hung
### Steps To Reproduce

1. Look at the list of core-dumps with `coredumpctl list`
2. Discover a large amount of crashed instances of `bitwarden-app`
3. Look at the journal and find this:

```
Jan 19 21:54:17 bitwarden[1825]: 21:54:17.709 › [LockService] Locking user some_uuid
Jan 19 21:54:17 bitwarden[1825]: 21:54:17.786 › [NAPI] [INFO] desktop_core::password::password: OS keyring is_locked=false
Jan 19 21:54:17 bitwarden[1825]: 21:54:17.805 › [LockService] Locked user some_uuid
Jan 19 21:54:17 bitwarden[1825]: 21:54:17.851 › Reloading render process
Jan 19 21:54:17 bitwarden[2063]: [2063:0119/215417.857168:FATAL:content/child/child_thread_impl.cc:377] Crashing because hung
Jan 19 21:54:17 bitwarden[1825]: 21:54:17.857 › Reloading render process
Jan 19 21:54:17 kernel: traps: Chrome_ChildIOT[2067] trap int3 ip:55867921262e sp:7f91435f9230 error:0 in bitwarden-app[6c0d62e,558674bda000+9233000]
Jan 19 21:54:17 systemd-coredump[75922]: Process 2063 (bitwarden-app) of user 1000 terminated abnormally with signal 5/TRAP, processing...
```

### Expected Result

Bitwarden Desktop Client should not crash.

### Actual Result

Desktop app crashes upon lock, see above.

### Screenshots or Videos

_No response_

### Additional Context

Desktop Environment: KDE Plasma, Wayland

### Operating System

Linux

### Operating System Version

Arch Linux

### Installation method

Other

### Build Version

2025.12.1

### Issue Tracking Info

quexten:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-38718

## 📔 Objective

Adds an env var to disable process reload. This is useful for debugging certain cases (IPC, Biometrics) which otherwise get cancelled / cleared. We do not currently expose this to users as per products advice.

### DIFF
--- a/apps/desktop/src/main/window.main.ts
+++ b/apps/desktop/src/main/window.main.ts
@@ -58,6 +58,13 @@ export class WindowMain {
     // Perform a hard reload of the render process by crashing it. This is suboptimal but ensures that all memory gets
     // cleared, as the process itself will be completely garbage collected.
     ipcMain.on("reload-process", async () => {
+      if (process.env.DISABLE_PROCESS_RELOAD === "true") {
+        this.logService.info(
+          "Process reload requested, but skipping because DISABLE_PROCESS_RELOAD is set",
+        );
+        return;
+      }
+
       if (isDev()) {
         this.logService.info("Process reload requested, but skipping in development mode");
         return;


quexten:
Moved to draft waiting for product perspective.

apps/desktop/src/main/window.main.ts
```diff
@@ -58,6 +58,13 @@ export class WindowMain {
     // Perform a hard reload of the render process by crashing it. This is suboptimal but ensures that all memory gets
     // cleared, as the process itself will be completely garbage collected.
     ipcMain.on("reload-process", async () => {
+      if (process.env.DISABLE_PROCESS_RELOAD === "true") {
```
djsmith85, line 61: ⚠️  If this is supposed to work as a dev-only flag, then we should combine it with `isDev()`.

Without this protection it can be set on production builds, which would enable potential attacks vectors.

djsmith85 (changes requested):

apps/desktop/src/main/window.main.ts
quexten: This is intentionally scoped to be just an environment variable, and not behind the compile time isDev setting.

There is no new security issue introduced here. Namely, in order for this to have any impact on secrets remaining in memory, the attacker would have to control the env prior or during an unlocked vault. This is a scenario which we do not cover - we only cover the counter case, a locked vault: https://contributing.bitwarden.com/architecture/security/principles/locked-vault-is-secure.

However, there are already paths in which process reload is disabled, for example when using PIN unlock, which would fall under the same issue, and if you use account switching. Both of these prevent process reload.

The comment about this being dev only is stating: "We do not want to give users an explicit easy setting." It is not stating "If this is in production builds, this would be insecure".

quexten:
@djsmith85 Seems this got a little bit lost. Could I get a review?

PR #21127, opened 2026-06-08
PM-36505 impmented policy text changes

bmbitwarden:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-36505

## 📔 Objective

Policy interactions and copy updates


apps/web/src/app/admin-console/organizations/members/components/invite-members-dialog/by-link-tab.component.ts
```diff
@@ -114,6 +115,11 @@ export class ByLinkTabComponent {
         void this.prefillFromVerifiedDomains();
       }
     });
+
+    // Reset the hint once the user edits the autopopulated value.
```
JimmyVo16: Non-blocker: Can you abstract this into its own method?

JimmyVo16 (changes requested):
Can you add test coverage? Just covering your changes should be sufficient.

apps/web/src/app/admin-console/organizations/members/components/invite-members-dialog/by-link-tab.component.html
eliykat, line 1: You've made the change described in the ticket, but to a different  and unrelated component. This component is used to invite new members. The ticket appears to be asking for an update to the send policy.

apps/web/src/locales/en/messages.json
eliykat, line 1: i18n strings should not be updated in place - see the failing `locales lint` CI job.

apps/web/src/app/admin-console/organizations/policies/policy-edit-dialog.component.ts
```diff
@@ -228,9 +228,11 @@ export class PolicyEditDialogComponent implements AfterViewInit {
         this.data.policy.type,
       );
     } catch (e: any) {
-      // No policy exists yet, instantiate an empty one
+      // No policy exists yet, instantiate an empty one.
+      // SendControls defaults to enabled for new organizations.
       if (e.statusCode === 404) {
-        return new PolicyResponse({ Enabled: false });
+        const defaultEnabled = this.data.policy.type === PolicyType.SendControls;
+        return new PolicyResponse({ Enabled: defaultEnabled });
```
eliykat, line 235: This is a new pattern, so a few thoughts/questions here:

1. this will only default to ON for any organizations that have never enabled the policy before. That may include existing organizations, not just new organizations per the ticket. Is that intended? (maybe - we probably don't want to remove Send for orgs that use it today - but it's unclear)
2. this only changes the policy's state when editing it. Anywhere else that reads it through the app would be reading its "real" disabled value. We'd need to consider how to make this change propagate through the system - "no policy record" used to mean disabled, but now it means enabled, but only for this specific policy.
3. we want to avoid policy-specific changes in shared components - again this just goes back to considering how to make this change in an app-wide fashion.

For 2-3 at least, I think you can handle it as follows:
- on edit - see if you can flip the value in the `send-controls.component` instead, so it doesn't pollute this shared component
- on server-side enforcement - you can write the `SendControlsPolicyRequirement` to invert the value
- on client-side enforcement - you can use `policyService.policiesByType$` and treat an empty list as enabled (you may want to abstract this inside your own domain service)

(By enforcement, I mean where the domain logic reads this value to change app behavior - show/hide the Send menu button, for example.)

The ticket describes this as a quick change but I think it requires a bit more than that to function as intended - it may be worth discussing further with the team. (I also recall a bunch of transitional logic to adopt the Send Controls policy, I'm not sure if that may change my answer above.) Happy to discuss further.

eliykat (changes requested):

eliykat:
It looks like a bad merge conflict hit every team with review requests - I have removed those to help everyone's notifications :)

apps/web/src/app/admin-console/organizations/policies/policy-edit-dialog.component.ts
mcamirault: Addressing #1, the intention is to have this policy enabled by default for everyone, new and existing orgs alike. I'm confirming something with @harr1424 (who implemented the Send Control policy initially) about how to proceed with this server-side but I think the client-side changes (both to the policy editor and wherever else in the app the Send Controls policy is relevant) will be far more substantial.

mcamirault:
@eliykat Given the desired behavior I laid out in my comment above (policy is "On" by default for all orgs) I think we have two options for how to proceed:
1. We add code to the server that adds a policy of this type when an org is created. We then run a data migration that adds an enabled Send Controls policy to every org that doesn't have one, though this policy won't be visible to them until the feature flag is enabled. I think just about everything would work as it did before, though we'd probably still want to update our domain service to defensively account for a missing policy as you suggested.
2. We don't perform a data migration; in this case I think we'd have to have policy-specific code - either in the client or the server - to make the policy overview page work properly. I think most of the other spots in the code that interact with the Send Controls policy do so through our domain service, which we can update as in the previous option to account for the policy not being there.

This is based on my current understanding of how the policy overview page displays "On" or "Off", let me know if I'm off base and what your opinion of these two options are. Personally I think 1 is better, though it requires a migration and we'd need to work with @bitwarden/dept-dbops on that.

eliykat:
Handling it in the app logic (option 2) is definitely the better approach, otherwise we are inserting lots of redundant data into our database for something we can logically resolve in code. The real question here is "what does no policy mean?" It previously meant OFF, now it could mean ON for some policy types. The policy system allows for configuration on a per-policy basis, so we should be able to invert this behavior in Send Controls policy-specific code.

I think your approach here is correct:

>   2. We don't perform a data migration; in this case I think we'd have to have policy-specific code - either in the client or the server - to make the policy overview page work properly. I think most of the other spots in the code that interact with the Send Controls policy do so through our domain service, which we can update as in the previous option to account for the policy not being there.

The Send Controls policy is also enforced server-side, so it will need to be both clients and server. See my last message above - I _think_ these are roughly the locations where you need to update the logic:

> I think you can handle it as follows:
> 
>   * on edit - see if you can flip the value in the `send-controls.component` instead, so it doesn't pollute this shared component
> 
>   * on server-side enforcement - you can write the `SendControlsPolicyRequirement` to invert the value
> 
>   * on client-side enforcement - you can use `policyService.policiesByType$` and treat an empty list as enabled (you may want to abstract this inside your own domain service)

I'm happy to help as needed, please send me a calendar invite or Slack message if you'd like to pair on it.

PR #21128, opened 2026-06-08
[PM-38723] [PM-37790] fix(fido2): consult vault by rpId before transport-only fallback

### LINKED ISSUE #20973: Passkey support completely broken when 1Password extension is co-installed (blocking for users in enforced work environments)

xrvk:
## 🎟️ Tracking

[PM-37790](https://bitwarden.atlassian.net/browse/PM-37790)

Fixes #20973, related to #20743.

## 📔 Objective

Vault-resident passkeys are unusable in Chrome 146+ when another passkey provider (1Password, iCloud Keychain) is also attached, because Bitwarden hands the request to the browser before checking whether the credential is in the vault.

[PM-33781](https://bitwarden.atlassian.net/browse/PM-33781) (#19606) added a fallback at the top of `Fido2ClientService.assertCredential` that throws `FallbackRequestedError` whenever every `allowCredentials` entry advertises only non-internal transports. RPs commonly echo the transports recorded at registration, so a passkey originally registered via cross-device hybrid flow arrives with transports such as `["hybrid"]` and trips the fallback even when the vault holds the credential.

Inside the existing non-internal-transports branch, call `silentCredentialDiscovery(rpId)` and only fall back when no requested credential id matches a vault credential. Vault credential ids are normalized via `parseCredentialId` so both UUID and `b64.<base64url>` forms match.

Follow-up to closed PR #21098. The previous attempt was rejected because the id-matching path appeared to enable vault enumeration. The same probe is already available via `transports:["internal"]`, so matching ids inside this branch adds no enumeration surface beyond what already exists. Happy to discuss further in review.

## 📸 Screenshots

N/A.

[PM-33781]: https://bitwarden.atlassian.net/browse/PM-33781?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/common/src/platform/services/fido2/fido2-client.service.spec.ts
+++ b/libs/common/src/platform/services/fido2/fido2-client.service.spec.ts
@@ -6,6 +6,7 @@ import { AuthenticationStatus } from "../../../auth/enums/authentication-status"
 import { DomainSettingsService } from "../../../autofill/services/domain-settings.service";
 import { Utils } from "../../../platform/misc/utils";
 import { VaultSettingsService } from "../../../vault/abstractions/vault-settings/vault-settings.service";
+import { Fido2CredentialView } from "../../../vault/models/view/fido2-credential.view";
 import { ConfigService } from "../../abstractions/config/config.service";
 import {
   ActiveRequest,
@@ -618,7 +619,7 @@ describe("FidoAuthenticatorService", () => {
     });
 
     describe("fallback for hardware/roaming keys", () => {
-      it("throws FallbackRequestedError when all allowCredentials have non-internal transports only", async () => {
+      it("throws FallbackRequestedError when all allowCredentials have non-internal transports only and vault has no credentials for rpId", async () => {
         const params = createParams({
           allowedCredentials: [
             { id: "credId1", transports: ["usb"] },
@@ -626,6 +627,54 @@ describe("FidoAuthenticatorService", () => {
           ],
           fallbackSupported: true,
         });
+        authenticator.silentCredentialDiscovery.mockResolvedValue([]);
+
+        await expect(client.assertCredential(params, windowReference)).rejects.toMatchObject({
+          fallbackRequested: true,
+        });
+        expect(authenticator.silentCredentialDiscovery).toHaveBeenCalledWith(RpId);
+        expect(authenticator.getAssertion).not.toHaveBeenCalled();
+      });
+
+      it("proceeds with Bitwarden when transports are non-internal but the vault has a credential for rpId", async () => {
+        const params = createParams({
+          allowedCredentials: [
+            { id: "credId1", transports: ["hybrid"] },
+            { id: "credId2", transports: ["usb"] },
+          ],
+          fallbackSupported: true,
+        });
+        authenticator.silentCredentialDiscovery.mockResolvedValue([
+          { credentialId: Utils.newGuid() } as Fido2CredentialView,
+        ]);
+        authenticator.getAssertion.mockResolvedValue(createAuthenticatorAssertResult());
+
+        await client.assertCredential(params, windowReference);
+
+        expect(authenticator.silentCredentialDiscovery).toHaveBeenCalledWith(RpId);
+        expect(authenticator.getAssertion).toHaveBeenCalled();
+      });
+
+      it("falls back when the vault is locked, without invoking silentCredentialDiscovery", async () => {
+        authService.activeAccountStatus$ = of(AuthenticationStatus.Locked);
+        const params = createParams({
+          allowedCredentials: [{ id: "credId1", transports: ["hybrid"] }],
+          fallbackSupported: true,
+        });
+
+        await expect(client.assertCredential(params, windowReference)).rejects.toMatchObject({
+          fallbackRequested: true,
+        });
+        expect(authenticator.silentCredentialDiscovery).not.toHaveBeenCalled();
+        expect(authenticator.getAssertion).not.toHaveBeenCalled();
+      });
+
+      it("falls back when silentCredentialDiscovery throws", async () => {
+        const params = createParams({
+          allowedCredentials: [{ id: "credId1", transports: ["hybrid"] }],
+          fallbackSupported: true,
+        });
+        authenticator.silentCredentialDiscovery.mockRejectedValue(new Error("boom"));
 
         await expect(client.assertCredential(params, windowReference)).rejects.toMatchObject({
           fallbackRequested: true,
@@ -642,6 +691,7 @@ describe("FidoAuthenticatorService", () => {
 
         await client.assertCredential(params, windowReference);
 
+        expect(authenticator.silentCredentialDiscovery).not.toHaveBeenCalled();
         expect(authenticator.getAssertion).toHaveBeenCalled();
       });
 
@@ -654,6 +704,7 @@ describe("FidoAuthenticatorService", () => {
 
         await client.assertCredential(params, windowReference);
 
+        expect(authenticator.silentCredentialDiscovery).not.toHaveBeenCalled();
         expect(authenticator.getAssertion).toHaveBeenCalled();
       });
     });
--- a/libs/common/src/platform/services/fido2/fido2-client.service.spec.ts
+++ b/libs/common/src/platform/services/fido2/fido2-client.service.spec.ts
@@ -619,11 +619,11 @@ describe("FidoAuthenticatorService", () => {
     });
 
     describe("fallback for hardware/roaming keys", () => {
-      it("throws FallbackRequestedError when all allowCredentials have non-internal transports only and vault has no credentials for rpId", async () => {
+      it("falls back when transports are non-internal and the vault has no credential for rpId", async () => {
         const params = createParams({
           allowedCredentials: [
-            { id: "credId1", transports: ["usb"] },
-            { id: "credId2", transports: ["nfc"] },
+            { id: Fido2Utils.arrayToString(guidToRawFormat(Utils.newGuid())), transports: ["usb"] },
+            { id: Fido2Utils.arrayToString(guidToRawFormat(Utils.newGuid())), transports: ["nfc"] },
           ],
           fallbackSupported: true,
         });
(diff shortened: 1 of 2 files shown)


xrvk:
cc @micahblut Let me know what you think of this iteration. 

>The previous attempt was rejected because the id-matching path appeared to enable vault enumeration. The same probe is already available via transports:["internal"], so matching ids inside this branch adds no enumeration surface beyond what already exists. Happy to discuss further in review.

coroiu:
Hey @xrvk — thanks for sticking with this and for the detailed write-ups. The underlying problem (Bitwarden passkeys unusable next to a locked-in 1Password) is real and we want it solved, so let me be transparent about where we've landed and where I'd most like your help.

**Why the vault check is a problem — it's the exact case the spec tells us to avoid.** WebAuthn §14.5.2 [Authentication Ceremony Privacy](https://www.w3.org/TR/webauthn-3/#sctn-assertion-privacy) says:

> In order to protect users from being identified without consent, implementations of the `[[DiscoverFromExternalSource]](origin, options, sameOriginWithAncestors)` method need to take care to not leak information that could enable a malicious WebAuthn Relying Party to distinguish between these cases, where "named" means that the credential is listed by the Relying Party in `allowCredentials`:
> - A named credential is not available.
> - A named credential is available, but the user does not consent to use it.
>
> If the above cases are distinguishable, information is leaked by which a malicious Relying Party could identify the user by probing for which credentials are available.

This PR makes exactly those two cases distinguishable: if the requested id is in the vault we proceed (credential available, pending consent); if it isn't, we throw `FallbackRequestedError` and hand off to the browser (credential not available). A site can tell those apart, which is the probing-for-credentials leak the spec calls out. It's also why the platform authenticators (iCloud, Google) don't expose this, and it's the property [#19606](https://github.com/bitwarden/clients/pull/19606) deliberately preserved by deciding solely on RP-supplied transports, never on vault contents.

It matters even same-site: a relying party knows every credential id it has ever issued, across all its users, so "probing for which credentials are available" lets a site test other accounts' ids against the current vault and link accounts a user kept separate — and because the vault is synced, that link crosses browser/profile boundaries cookies never could. (See also §14.6.3 [Privacy leak via credential IDs](https://www.w3.org/TR/webauthn-3/#sctn-credential-id-privacy-leak).)

**On the implementation vs. your description.** Your comment describes an rpId-only check ("only falls back when the vault has no credentials for that rpId at all… never on individual credential ids"), but the code still matches individual ids:

```ts
const requestedIds = params.allowedCredentials.map((c) => Fido2Utils.stringToArray(c.id));
matchedVaultCredential = vaultCredentials.some((vc) => {
  const parsed = parseCredentialId(vc.credentialId);
  return parsed != null && requestedIds.some((id) => compareCredentialIds(parsed, id));
});
```

and your test "falls back when vault has credentials for rpId but none match the requested ids" confirms the branch is on the specific id. So functionally this is the same enumeration surface as [#21098](https://github.com/bitwarden/clients/pull/21098).

**On "the same probe is already available via `transports:["internal"]`".** I don't think the code bears that out. On the `internal` path the request goes through `getAssertion` whether or not the id is in the vault, and without user consent both outcomes end in the same `NotAllowedError` (not-found calls `informCredentialNotFound()` then throws; a dismissed picker throws the same), with a Bitwarden prompt shown either way — which is actually consistent with §14.5.2. This PR is different: `silentCredentialDiscovery` runs with no user interaction, and on a no-match it throws `FallbackRequestedError` immediately — no Bitwarden UI, handed straight to the browser's native WebAuthn — while a match shows Bitwarden's UI. That match-vs-no-match fork is a new, and effectively silent, distinction, so it adds enumeration surface rather than reusing something that already exists.

**On whether this fixes the target issue — where I'd most like your input.** [#20973](https://github.com/bitwarden/clients/issues/20973) reproduces on extension builds from before this transport-fallback code path existed (e.g. 2025.12.1), which points at the older content-script race between co-installed providers. That's what [#20849](https://github.com/bitwarden/clients/pull/20849) (the `webAuthenticationProxy` migration) is meant to fix, and a vault check here wouldn't affect it. Separately, I tried to reproduce the specific case this PR targets — a synced passkey coming back with non-internal transports so we disown it — and couldn't: on Android the credential came back tagged with `internal`, which the current code already handles.

So before changing this logic, could you share a concrete repro? A current build, a credential confirmed to be in the Bitwarden vault, that the RP sends back with non-internal-only transports. If that exists it genuinely changes the picture and we can find the right fix together. Otherwise I think [#20849](https://github.com/bitwarden/clients/pull/20849) is the path for [#20973](https://github.com/bitwarden/clients/issues/20973).

What's the actual setup where you're seeing the breakage? Happy to keep digging with you.

xrvk:
Thanks @coroiu, I really appreciate the team getting back to me. Your post clarified a few things for me, and what I had wrong. I appreciate you taking the time to lay out the reasoning carefully, rather than declining without explanation.

On the two points you raised, we can agree that:

**§14.5.2 is the right read.** The silent match-vs-no-match fork is a new distinguisher. My defense relied on the `transports:["internal"]` path adding the same surface, but as you showed, that path lands on the same `NotAllowedError` with Bitwarden's UI either way. The two paths are not equivalent, and the spec text is clear on why this approach violates it.

**The body did not match the code.** It described an rpId-only check while the implementation matched individual ids, and the test confirms it. The body drifted out of sync with a code change and I missed it on the final read. Functionally this lands at the same surface as #21098.

**On the reproduction steps you asked for.** I tried to capture on macOS Chrome 149 (the platform #20973 reports) and it seems the result is more upstream than the transport-fallback path. With 1Password force-installed alongside Bitwarden, Bitwarden's content-script never installs its hooks. Here it is reproduced on github.com itself:

```
Repro environment
- macOS, fresh Chrome install in a separate user-data-dir
- Chrome 149.0.7827.103 (Official Build) (arm64)
- Bitwarden extension 2026.5.1
- 1Password extension 8.12.22.17 (force-installed via IT policy)

Console output on github.com sign-in:
webauthn-listeners.js:3 Cannot overwrite navigator.credentials.create method while 1Password is enabled
   set @ webauthn-listeners.js:3
   (anonymous) @ fido2-page-script.js:466
webauthn-listeners.js:3 Cannot overwrite navigator.credentials.get method while 1Password is enabled
   set @ webauthn-listeners.js:3
   (anonymous) @ fido2-page-script.js:467
...
webauthn-get-element.ts:214 Uncaught (in promise) NotAllowedError: The operation either timed out or was not allowed. See: https://www.w3.org/TR/webauthn-2/#sctn-privacy-considerations-client.
   prompt @ webauthn-get-element.ts:214
```

image

1Password's `webauthn-listeners.js` installs a property setter on `navigator.credentials.create/get` that throws on overwrite. Bitwarden's `fido2-page-script.js` attempts the monkey-patch at those lines and is rejected. The subsequent passkey prompt on github.com fails with the same `NotAllowedError` end users report in #20973. With the setters in place, transport echo is irrelevant; the call never reaches Bitwarden's authenticator. The fallback path this PR modifies addresses a layer that does not execute in this scenario.

I can agree with closing this PR if you wish.

But overall, it seems that in order to resolve the issue, we should be moving towards #20849 , which I oponed on May 25. This would  update Bitwarden to use `chrome.webAuthenticationProxy`. The fix here works since the proxy API does not depend on content-script injection so 1Password's setters cannot block it, and Chrome routes WebAuthn calls through **a registered chooser** between providers. Unfortunately, CI is waiting on the standard maintainer-approval gate.

Could we get a review pass on #20849? It would be amazing to help us finally get us past this issue.  Please, let me know if there's anyhting I can do to help. 

I promise I'm not trying to be a nuisance here, I just want to be able to use my bitwarden passkeys again 🙏

micahblut:
@xrvk i've pinged the right people to get eyes on #20849 so hopefully we can get your passkeys working again. I'm going to close this contribution, as we're aligned it is not the approach we want to take to resolve the issue.

PR #21133, opened 2026-06-08
Uif/m3/visual regression fixes

BryanCunningham:
## 📔 Objective

- Fix some minor visual regressions introduced with new field designs

@bitwarden/team-admin-console-dev As a heads up: I've updated a couple of the fields in the scim components to be readonly rather than disabled. I've aligned with design on this change. Readonly fields in this scenario are a better experience for both screen reader and keyboard users.

Also, made some minor changes to the scim v2 component to be able to use the switch in this manor. Support for switch for use like this is not built into the component. It's meant to be rendered with a label currently

### DIFF
--- a/bitwarden_license/bit-web/src/app/admin-console/organizations/manage/scim-v2.component.ts
+++ b/bitwarden_license/bit-web/src/app/admin-console/organizations/manage/scim-v2.component.ts
@@ -16,18 +16,14 @@ import { PlatformUtilsService } from "@bitwarden/common/platform/abstractions/pl
 import { OrganizationId } from "@bitwarden/common/types/guid";
 import {
   AriaDisableDirective,
-  BaseCardDirective,
   BitActionDirective,
   BitIconButtonComponent,
-  CardComponent,
   DialogService,
   FormFieldModule,
-  IconComponent,
   LinkComponent,
   SpinnerComponent,
   SwitchComponent,
   ToastService,
-  TypographyModule,
 } from "@bitwarden/components";
 import { I18nPipe } from "@bitwarden/ui-common";
 import { WebHeaderComponent } from "@bitwarden/web-vault/app/layouts/header/web-header.component";
@@ -41,16 +37,12 @@ import { ScimApiKeyDialogComponent } from "./scim-api-key-dialog.component";
   imports: [
     WebHeaderComponent,
     ReactiveFormsModule,
-    BaseCardDirective,
-    CardComponent,
-    TypographyModule,
     LinkComponent,
     AriaDisableDirective,
     SwitchComponent,
     FormFieldModule,
     BitIconButtonComponent,
     BitActionDirective,
-    IconComponent,
     I18nPipe,
     SpinnerComponent,
   ],
@@ -70,8 +62,8 @@ export class ScimV2Component {
 
   protected readonly enabled = new FormControl(false);
   protected readonly formData = new FormGroup({
-    endpointUrl: new FormControl({ value: "", disabled: true }),
-    clientSecret: new FormControl({ value: "", disabled: true }),
+    endpointUrl: new FormControl({ value: "", disabled: false }),
+    clientSecret: new FormControl({ value: "", disabled: false }),
   });
 
   private readonly organizationId: Signal<OrganizationId>;
--- a/bitwarden_license/bit-web/src/app/admin-console/organizations/manage/scim-v2.component.ts
+++ b/bitwarden_license/bit-web/src/app/admin-console/organizations/manage/scim-v2.component.ts
@@ -1,4 +1,12 @@
-import { ChangeDetectionStrategy, Component, effect, inject, Signal, signal } from "@angular/core";
+import {
+  ChangeDetectionStrategy,
+  Component,
+  effect,
+  inject,
+  Signal,
+  signal,
+  viewChild,
+} from "@angular/core";
 import { toSignal } from "@angular/core/rxjs-interop";
 import { FormControl, FormGroup, ReactiveFormsModule } from "@angular/forms";
 import { ActivatedRoute } from "@angular/router";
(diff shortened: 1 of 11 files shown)


bitwarden_license/bit-web/src/app/admin-console/organizations/manage/scim-v2.component.ts
```diff
@@ -70,8 +62,8 @@ export class ScimV2Component {
 
   protected readonly enabled = new FormControl(false);
   protected readonly formData = new FormGroup({
-    endpointUrl: new FormControl({ value: "", disabled: true }),
-    clientSecret: new FormControl({ value: "", disabled: true }),
+    endpointUrl: new FormControl({ value: "", disabled: false }),
+    clientSecret: new FormControl({ value: "", disabled: false }),
```
vleague2: ❓ Not familiar with this code, what is the context for changing the initial value?

libs/components/src/form-control/form-control-card.component.html
```diff
@@ -1,48 +1,51 @@
 @let control = base.formControl();
-<div
-  class="tw-flex tw-items-center tw-relative tw-p-4 tw-border tw-border-border-base tw-rounded-xl tw-justify-between tw-gap-4 has-[label:hover]:tw-border-border-brand-soft has-[input:focus-visible]:tw-ring-2 has-[input:focus-visible]:tw-ring-border-focus has-[input:focus-visible]:tw-border-transparent has-[input:checked]:tw-bg-bg-brand-soft"
->
-  <!-- label z-index to ensure label sits on top of switch thumb control(z-index: 2) -->
```
vleague2, line 5: 🧐 I can't click the switch in the card form again, I am blaming the z index

https://github.com/user-attachments/assets/59a8484f-2311-437d-a8a9-d110f395714d

eliykat:
@BrandonTreston requesting your review because you've worked on both of these recently.

bitwarden_license/bit-web/src/app/admin-console/organizations/manage/scim-v2.component.ts
BryanCunningham: @vleague2 So, these should not be disabled at all. This _seemed_ like the best/least intrusive way to accomplish this. But, I think maybe just doing this also works but, not for sure. Thoughts?
```js
endpointUrl: new FormControl(""),
clientSecret: new FormControl(""),
```

libs/components/src/form-control/form-control-card.component.html
BryanCunningham: Noooo! Ok, will take a look. I messed around with the z-index to fix an a11y error where contrast could not be evaluated due to the label being 'on top' of the text

BryanCunningham: @vleague2 This should be taken care of now. Thanks for the heads up!

bitwarden_license/bit-web/src/app/admin-console/organizations/manage/scim-v2.component.ts
vleague2: Ooh I'm tracking now, this is related to the readonly change haha. Yes, if we're removing the option to disable these in favor of readonly, then we should probably just remove that property here. I am not as familiar with the Angular form API but that seems reasonable?

BryanCunningham: I went ahead and just removed the object and added an empty string as the value

vleague2 (approved):

bitwarden_license/bit-web/src/app/admin-console/organizations/manage/scim-v2.component.ts
BTreston, line 1: There are a lot of mechanics that we're building into the consumer here regarding the switch. Are all of these IDs and aria labels being set here only because the component was not built to have no label? Do we have any plans to support this use case better? I can easily see more design using just a simple toggle switch with no label like this in the future.

BryanCunningham: @BTreston Correct. The Switch component is not currently set up to be used in this way. And, I agree, this is not ideal. 

We have to tie the external content to the switch's `aria-labelledby` and `aria-describedby` attributes in this way to make it accessible. And, attach the actual label to the switch's internal input... it's gross, for sure.

I met with design about this yesterday and in the interest of UIF not blocking any work, we'll go with this solution for now. 

As for plans to better support this. If(when) this particular pattern is used elsewhere, we'll likely need to build a new(bit-disclosure-card???) component to support this use case.

BTreston (approved):

bitwarden_license/bit-web/src/app/admin-console/organizations/manage/scim-v2.component.ts
BryanCunningham: My main concern about allowing switches to be used in this way(without the wrapper `bit-form-control` component controlling the label and hint text rendering) is a11y. We'd have to depend on consuming teams correctly adding aria attributes. And, that is a little bit of a slippery slope. We _could_ allow `bit-label` and `bit-hint` to be passed into the standalone switch, look for those children and associate them automatically under the hood. But, then you'd have duplicate content. One label and/or hint in the switch and one external label/hint. Not a huge deal per se. But, you then run the risk of those things becoming out of sync...

BTreston: Gotcha, I feel like it would make sense to drop this into its own form then so we can take advantage of the built in a11y handling. Though if there's still no label/hint and that still causes issues then maybe not. As an idea (without having dug into it) could we give it a label and add some input flag to hide the label w/ CSS/tailwind so it at least exists inside the component?

Regardless, I understand now why its not ideal. Just wanted to see if there was an easy win to clean it up. If there's not then I'm okay with this since we have all the context here for historical reference. Approved 👍

BryanCunningham: > As an idea (without having dug into it) could we give it a label and add some input flag to hide the label w/ CSS so it at least exists inside the component?

Yeah, that is _sort of_ what I was getting at above with
> We could allow bit-label and bit-hint to be passed into the standalone switch, look for those children and associate them automatically under the hood.

But, we'd then run into that duplicate content/chance of content falling out of sync situation

PR #21136, opened 2026-06-08
[PM-38581] - disable drag for single custom fields

### LINKED ISSUE #21060: Cannot highlight/select text inside a custom field when only one custom field exists
### Steps To Reproduce

Open the Bitwarden Desktop app on Windows.
Create or edit a login item.
Scroll down to Custom fields.
Add one custom field.
Set the field type to Hidden.
Name the field something like Security Question.
Enter text into the field value.
Try to click and drag over the text to highlight/select part of the text.

### Expected Result

I expected to be able to highlight/select text inside the custom field normally when clicking and dragging over the text.

### Actual Result

When there is only one custom field, clicking and dragging over the text does not select/highlight the text. Instead, the whole custom field behaves like it is being dragged/reordered.

If I add a second custom field, the drag/reorder handle icons appear next to the custom fields, and then text selection works normally. I tested this with one hidden field plus one text field, and also with two hidden fields. In both cases, once there are two custom fields, I can highlight/select the text normally.

It seems like when only one custom field exists, the field may still be acting draggable/reorderable even though the drag handle is not shown, which prevents normal text selection.

### Screenshots or Videos

https://github.com/user-attachments/assets/59fd5f82-372f-42c5-8e4c-e9fcd2b1516c

### Additional Context

This happens in the Bitwarden Desktop app on Windows. I noticed it while trying to store a security question/answer in a custom hidden field. The issue seems specific to the case where only one custom field exists. Once a second custom field is added, the UI behaves correctly.

### Operating System

Windows

### Operating System Version

Version 25H2 (OS Build 26200.8524)

### Installation method

### Build Version

Version 2026.5.0 

### Issue Tracking Info

jaasen-livefront:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-38581

Fixes https://github.com/bitwarden/clients/issues/21060

## 📔 Objective

Fixes an issue where an individual custom field could be dragged by clicking and
holding anywhere on its text input when only a single custom field existed.

The drag handle button (`cdkDragHandle`) is only rendered when there are two or
more fields. With a single field, no handle was present, so Angular CDK fell back
to making the entire row draggable — including the text input — which interfered
with normal text entry/selection.

The fix disables `cdkDrag` on the row when there is only one field by adding
`fields.controls.length <= 1` to the `cdkDragDisabled` condition. When multiple
fields are present, behavior is unchanged: the drag-and-drop handle on the far
right works as expected, and the inputs remain fully interactive.

## 📸 Screenshots

https://github.com/user-attachments/assets/9ac47a00-e9ea-4709-9834-16568a900fa8

### DIFF
--- a/libs/vault/src/cipher-form/components/custom-fields/custom-fields.component.html
+++ b/libs/vault/src/cipher-form/components/custom-fields/custom-fields.component.html
@@ -19,7 +19,7 @@
             }"
             [attr.data-testid]="field.value.name + '-entry'"
             cdkDrag
-            [cdkDragDisabled]="!canEdit(field.value.type)"
+            [cdkDragDisabled]="!canEdit(field.value.type) || fields.controls.length <= 1"
             #customFieldRow
           >
             <!-- Text Field -->
--- a/libs/vault/src/cipher-form/components/custom-fields/custom-fields.component.html
+++ b/libs/vault/src/cipher-form/components/custom-fields/custom-fields.component.html
@@ -19,7 +19,7 @@
             }"
             [attr.data-testid]="field.value.name + '-entry'"
             cdkDrag
-            [cdkDragDisabled]="!canEdit(field.value.type) || fields.controls.length <= 1"
+            [cdkDragDisabled]="dragDisabled(field.value.type)"
             #customFieldRow
           >
             <!-- Text Field -->
--- a/libs/vault/src/cipher-form/components/custom-fields/custom-fields.component.ts
+++ b/libs/vault/src/cipher-form/components/custom-fields/custom-fields.component.ts
@@ -163,6 +163,10 @@ export class CustomFieldsComponent implements OnInit, AfterViewInit {
     );
   }
 
+  dragDisabled(type: FieldType): boolean {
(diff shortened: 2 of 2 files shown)


libs/vault/src/cipher-form/components/custom-fields/custom-fields.component.html
```diff
@@ -19,7 +19,7 @@ <h2 bitTypography="h6">{{ "customFields" | i18n }}</h2>
             }"
             [attr.data-testid]="field.value.name + '-entry'"
             cdkDrag
-            [cdkDragDisabled]="!canEdit(field.value.type)"
+            [cdkDragDisabled]="!canEdit(field.value.type) || fields.controls.length <= 1"
```
gbubemismith: 🎨 can you move this logic to the class instead?

jaasen-livefront: Done! https://github.com/bitwarden/clients/pull/21136/commits/645f6d927ebc02908da10baf694593aed91b1280

gbubemismith (approved):

PR #19307, opened 2026-02-28
[PM-32917] fix: detect Firefox by Application Support/Firefox on macOS

sergeypdev:
Firefox reads native messaging manifests from
~/Library/Application Support/Mozilla/, but that directory may not exist if Firefox stores its data in ~/Library/Application Support/Firefox/. Refactor getDarwinNMHS to separate browser detection paths from manifest paths so the manifest is written to the correct Mozilla location when Firefox is detected by either directory. Also use recursive mkdir so the manifest directory tree is created when it doesn't exist.

Fixes #18996
Fixes #19304

## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-32274
https://bitwarden.atlassian.net/browse/PM-32914

## 📔 Objective

When creating manifest file for communicating with the browser extension, Firefox could end up begin not detected. This should fix the issue

PR #21141, opened 2026-06-09
[PM-38760] Recursively create directory when writing manifest

### LINKED ISSUE #18996: [PM-32274] NativeMessagingHosts file seems to be missing in some cases for Firefox (at least on MacOS)
### Description

When you try to set up biometrics for the Firefox browser extension - even though the desktop app was prepared as it should be - several people reported that it would stay on "Awaiting confirmation from desktop". That was at least reported here:

- https://community.bitwarden.com/t/firefox-browser-extension-biometrics-integration-stuck-on-awaiting-confirmation-from-desktop/69712
- https://community.bitwarden.com/t/macos-firefox-bitwarden-extension-rejects-touch-id/93161
- https://github.com/bitwarden/clients/issues/18521

This comment here ([https://github.com/bitwarden/clients/issues/18521#issuecomment-3866182533](https://github.com/bitwarden/clients/issues/18521#issuecomment-3866182533)) suggested that manually creating the missing NativeMessagingHosts file with a certain content (see that comment) would solve the connectivity problem with the desktop app.

At least one other user now reported, that it would solve their biometrics set up issue with Firefox as well ([here](https://github.com/bitwarden/clients/issues/18521#issuecomment-3873191043) and [here](https://community.bitwarden.com/t/macos-firefox-bitwarden-extension-rejects-touch-id/93161/15)). (EDIT: There is another user who wrote that "the entire Mozilla folder" would be missing and the file, and adding the folder and file would fix the issue: https://github.com/bitwarden/clients/issues/18521#issuecomment-3887755970)

This does look like a bug (BTW, the reports I mentioned indicate, that that file is _sometimes there_ and _**sometimes missing**_ on their systems - for whatever reasons - and so it may not be reproducible ever time!!), so please have another look into this, as no regular user can be expected to find out about the NativeMessagingHosts file that had to be added manually in some instances. 

(I personally don't mind -- if you consider or rather: further investigate this report at all -- if this issue here stays open, or if https://github.com/bitwarden/clients/issues/18521 might get reopened. But I do think it should be investigated, why that file is missing in some cases, hindering biometrics for the Firefox extension being set up.)

### Operating System

macOS

### Web Browser

Firefox

### Issue Tracking Info

pschlan:
## 🎟️ Tracking

https://github.com/bitwarden/clients/issues/18996

## 📔 Objective

Might fix #18996 - it seems the `Mozilla` folder is missing entirely in those cases, and it's not enough to just create the leaf directory.


pamperer562580892423:
Whenever this will be worked on -- FYI, there is also another PR that tried to fix this: https://github.com/bitwarden/clients/pull/19307

addisonbeck (approved):

pschlan:
CI fails because of `User pschlan does not have the necessary access for this repository.` - not sure how this is supposed to work for community contributions.

addisonbeck:
Yes, sorry, that's a step for maintainers before merge. It's a bit annoying, and prone to clicking-and-forgetting behavior 😁

Thanks again!

pamperer562580892423:
Thanks for fixing this bug!

@addisonbeck I guess this other PR can be closed then: https://github.com/bitwarden/clients/pull/19307 ?!

PR #21143, opened 2026-06-09
Change configRetrievalIntervalMs

coltonhurst:
Prevents immediate repeated polling when the user is logged out.

This is a quick update following the first PR: https://github.com/bitwarden/clients/pull/21131

### DIFF
--- a/apps/browser/config/development.json
+++ b/apps/browser/config/development.json
@@ -4,6 +4,6 @@
       "base": "https://localhost:8080"
     },
     "skipWelcomeOnInstall": true,
-    "configRetrievalIntervalMs": 1
+    "configRetrievalIntervalMs": 60000
   }
 }
--- a/apps/browser/config/development.json
+++ b/apps/browser/config/development.json
@@ -4,6 +4,6 @@
       "base": "https://localhost:8080"
     },
     "skipWelcomeOnInstall": true,
-    "configRetrievalIntervalMs": 60000
+    "configRetrievalIntervalMs": 5000
   }
 }
--- a/apps/cli/config/development.json
+++ b/apps/cli/config/development.json
@@ -1,6 +1,6 @@
 {
   "flags": {},
   "devFlags": {
-    "configRetrievalIntervalMs": 1
+    "configRetrievalIntervalMs": 60000
   }
 }
--- a/apps/cli/config/development.json
+++ b/apps/cli/config/development.json
@@ -1,6 +1,6 @@
 {
   "flags": {},
   "devFlags": {
-    "configRetrievalIntervalMs": 60000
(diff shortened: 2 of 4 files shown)


apps/browser/config/development.json
```diff
@@ -4,6 +4,6 @@
       "base": "https://localhost:8080"
     },
     "skipWelcomeOnInstall": true,
-    "configRetrievalIntervalMs": 1
+    "configRetrievalIntervalMs": 60000
```
mzieniukbw: Now when i think about it, the 1 millisecond is problematic, because new saved [user config state](https://github.com/bitwarden/clients/blob/662e4c77401173e1d67c5a24e6ea987637ba02c3/libs/common/src/platform/services/config/default-config.service.ts#L232) would cause the re-emit of the state observable [userConfigFor$](https://github.com/bitwarden/clients/blob/662e4c77401173e1d67c5a24e6ea987637ba02c3/libs/common/src/platform/services/config/default-config.service.ts#L111) causing the endless loop of retrievals, since it the delay between the processing of new config for user and another renewal is likely more than few milliseconds.
I see no reason to go with 60 seconds if we want the experience of "no caching", might do 1 seconds, which will give us enough buffer to not cause endless loop of retrievals - the delay between setting new state and re-emit of observable is usually in 1-50ms window (tested in prod when i was debugging a race-condition of similar sort), so i would say 1 second should be more than enough.
```suggestion
    "configRetrievalIntervalMs": 1000
```

mzieniukbw: Another way to make sure we don't end up in a loop of renewals is to additionally pipe [userConfigFor$](https://github.com/bitwarden/clients/blob/662e4c77401173e1d67c5a24e6ea987637ba02c3/libs/common/src/platform/services/config/default-config.service.ts#L111) with `distinctUntilChanged`, although that will likely require custom comparer function for [ServerConfig](https://github.com/bitwarden/clients/blob/b07d6c29a45b8bea60f7e9e11993714a7394df86/libs/common/src/platform/abstractions/config/server-config.ts#L21), as we already do for [Environment](https://github.com/bitwarden/clients/blob/662e4c77401173e1d67c5a24e6ea987637ba02c3/libs/common/src/platform/services/config/default-config.service.ts#L57)

MGibson1 (dismissed):
1 min is 60 times better, I'd go as low as maybe 5 seconds just because we haven't actually stress tested our storage for any speed.

coltonhurst:
@MGibson1 @mzieniukbw Ok, if we're good with it, aligned on `5` seconds here: 4d2a49581afe93e050c724c7db12439791a09bc1

MGibson1 (approved):

mzieniukbw (approved):

PR #21152, opened 2026-06-09
[PM-38182] Fix Trend Chart not appearing in Access Intelligence V2 path

Banrion:
## 🎟️ Tracking

[PM-38182](https://bitwarden.atlassian.net/browse/PM-38182)

## 📔 Objective

This pull request fixes the **Risk over time** chart not appearing in the Access Intelligence **Activity** tab when the flags `pm-31936-access-intelligence-new-architecture` and `pm-26961-access-intelligence-trend-chart` are enabled.

**Changes:**
- Add `TrendWidgetComponent` to Access Intelligence v2 path behind the `pm-26961-access-intelligence-trend-chart` feature flag
    - Matched v1 implementation styling and logic
- Add `DefaultRiskOverTimeService` to routing
- Updated test cases and storybooks

**Additional Changes**
- Consolidated storybook helpers for `TrendWidget` for use in Activity Tab
- Updated and moved `NewAdminWelcomeDialog` storybook

**Note:** I see the _Run report_ button is right-aligned while it is left-aligned in v1. While out of scope of this pull request, I'll confirm this difference with design and add tracking if changes are needed.

## 📸 Screenshots

Access Intelligence with flags enabled:

image

[PM-38182]: https://bitwarden.atlassian.net/browse/PM-38182?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


bitwarden_license/bit-web/src/app/dirt/access-intelligence/v2/activity-tab/activity-tab.component.ts
```diff
@@ -54,6 +79,39 @@ export class ActivityTabComponent {
...
+      const orgId = this.organizationId();
+      if (flag && orgId && !untracked(() => this.trendChartInitialized())) {
+        this.trendChartInitialized.set(true);
+        this.riskOverTimeService.initialize(
+          orgId,
+          DEFAULT_TIME_PERIOD,
+          TrendWidgetViewType.Applications,
+        );
+      }
+    });
```
lastbestdev, line 107: I am wondering if this has to be inside an `effect`, or if we can just run it on component construction?

lastbestdev (dismissed):
Looks good, one comment to clarify need for `effect`

bitwarden_license/bit-web/src/app/dirt/access-intelligence/v2/activity-tab/activity-tab.component.ts
Banrion: Good question 🤔 In general, I lean towards reactivity over imperative checks, but we could move it to an `ngOnInit` function and use `await` on the flag to remove the need for `trendChartInitialized`. On reflection, I'd want this to be responsive if the user changed orgs via navigation (e.g. bookmarked pages) so I'd rather remove the `trendChartInitialized` guard and keep the effect. Re-calling `initialize` is safe since it just pushes onto a `switchMap`'d subject, so the old fetch is cancelled and swapped cleanly.

lastbestdev (approved):

bitwarden_license/bit-web/src/app/dirt/access-intelligence/v2/activity-tab/activity-tab.component.ts
lastbestdev: Good point on handling Org changes, lets keep it!

PR #21153, opened 2026-06-09
[PM-37228] feat: Add price increase callout to organization subscription page

amorask-bitwarden:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-37228

## 📔 Objective

Adds an informational callout to the cloud **Organization → Subscription** page notifying admins
when a price increase is scheduled at their next renewal — the in-product surface for the
existing-customer pricing migration (epic PM-35215).

It binds to a new nullable `scheduledPriceIncrease` object on the existing
`GET organizations/{id}/billing/vnext/warnings` response and renders **only when present**:

- Shows the new per-seat **monthly-equivalent** price (computed server-side — the client does no
  price math) and the **effective date**.
- Copy variant is driven by `cadence`; the annual variant appends "(billed annually)".
- Rendered as a `bit-callout type="info"`, below the subscription price details and above
  "More from Bitwarden".

Implementation notes:

- The effective date is formatted in **UTC** (`longDate`) so it matches the renewal email's
  calendar date and never shows off-by-one west of UTC.
- **No client feature flag** — the server gates the field behind `PM35215_BusinessPlanPriceMigration`
  and returns `null` when it shouldn't display; the client renders iff the field is non-null.
- New standalone, OnPush component using signals (`input.required()`, `inject()`, `toObservable`).
- Includes a small chore commit adding `/.worktrees/` to `.gitignore` (local worktree hygiene).
- Server-side support for the new field ships separately.

## 📸 Screenshots

image


sbrown-livefront (dismissed):
🧹 Clean! Great component tests too ✅

sbrown-livefront (approved):

PR #21154, opened 2026-06-09
[PM-38787] Pin vault bulk-actions bar to the viewport on scroll

nick-livefront:
## 🎟️ Tracking

[PM-38787](https://bitwarden.atlassian.net/browse/PM-38787)

## 📔 Objective

The bulk-actions bar (`bit-vault-batch-action`) didn't stay pinned to the bottom of the viewport — it scrolled away with the vault list.

**Root cause:** the bar is `position: fixed`, but the layout's `<main>` carries `cdk-virtual-scrollable`, whose `contain: strict` makes it a containing block for fixed descendants. So the bar resolved against `<main>` (the scroll container) instead of the viewport and scrolled with the content.

**Fix:** render the bar in a new `bit-layout` footer region — a sibling of `<main>` in the same grid cell but outside the scroll container — mirroring the existing `drawerPortal` pattern:
- `LayoutFooterService` holds a `Portal` signal that `LayoutComponent` renders via `cdkPortalOutlet`, this closely follows how the drawer service here but I'd appreciate any feedback of course!

## 📸 Screenshots

|Before|After|
|-|-|
|<video src="https://github.com/user-attachments/assets/c6b19145-6aba-4103-bc02-a2eb30f757d7" />|<video src="https://github.com/user-attachments/assets/7546d2cc-7064-47be-88be-eb0494d86706" />|

[PM-38787]: https://bitwarden.atlassian.net/browse/PM-38787?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/components/src/layout/index.ts
+++ b/libs/components/src/layout/index.ts
@@ -1,2 +1,3 @@
 export * from "./layout.component";
+export * from "./layout-footer.service";
 export * from "./scroll-layout.directive";
--- a/libs/components/src/layout/layout-footer.service.ts
+++ b/libs/components/src/layout/layout-footer.service.ts
@@ -1,0 +1,27 @@
+import { Portal } from "@angular/cdk/portal";
+import { Injectable, signal } from "@angular/core";
+
+/**
+ * Hosts content pinned to the bottom of the layout's main content area — e.g. a
+ * bulk-actions bar.
+ *
+ * The layout renders this portal in a region that overlaps `<main>` but sits outside its
+ * scroll container, so the content stays pinned to the viewport while `<main>` scrolls and is
+ * horizontally aligned to the main content column. Consumers attach a `TemplatePortal` /
+ * `ComponentPortal` and detach it on destroy.
+ */
+@Injectable({ providedIn: "root" })
+export class LayoutFooterService {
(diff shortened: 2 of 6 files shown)


BryanCunningham:
@nick-livefront This is fine for now probably. I guess my question is why does `<main>` even contain the virtual scroll stuff. Should be a wrapper around the table, I would think. But, that's probably a bigger ask/lift

BryanCunningham (approved):

shane-melton:
@willmartian Any thoughts on this solution in regards to your work in https://github.com/bitwarden/clients/pull/20900 ?

shane-melton (approved):

nick-livefront:
> @willmartian Any thoughts on this solution in regards to your work in #20900 ?

@willmartian We're going to merge this in to keep QA moving on the batch bar work. If you do have feedback or want to discuss on how we can collaborate on the table work, let me know!

willmartian:
I would like this fixed at the UIF level instead of the Vault-level. I'll create a ticket for UIF, and treat the merged resolution as a temporary fix. Thanks for the ping!

PR #21167, opened 2026-06-10
[PM-38821] Fix vault search not finding entries with umlauts/diacritics

### LINKED ISSUE #10738: can't get or search entries with "umlaute" in name
### Steps To Reproduce

1. create an entry with an "Umlaut" in the name (e.g. "test_umlaut_ü") and a random password
2. execute `bw get password "test_umlaut_ü"` or `bw list items --search "test_umlaut_ü"`

Sidenote: `bw list items --search "test_umlaut_"` returns the correct entry and also displays the umlaut in the name correctly.

### Expected Result

`<random password>`

### Actual Result

`Not found`

### Screenshots or Videos

_No response_

### Additional Context

_No response_

### Operating System

Linux

### Operating System Version

Debian Bookworm

### Shell

Bash

### Build Version

2024.8.1

### Issue Tracking Info

electronicpeter:
## 🎟️ Tracking

Fixes #10738 

## 📔 Objective

Vault search failed to find entries when the search query or the entry fields contained umlaut/diacritic characters (ö, ü, ä, é, …).

**Root cause:** `normalizeSearchQuery` was applied to the query, stripping diacritics (e.g. `ö → o`), but the field values compared against were left un-normalized. This caused mismatches like `"muller".indexOf("müller") === -1`.

A second issue existed in the lunr pipeline: `normalizeAccentsPipelineFunction` only ran during indexing, not during search. Adding it to `builder.searchPipeline` was a no-op because the original function's field-metadata check always returned `false` in search context (no field metadata present).

**Changes:**

- `searchCiphersBasic`: normalize `name`, `subtitle`, `notes` and login URIs before `indexOf` comparison, so searches with or without umlauts both match
- `normalizeAccentsPipelineFunction`: preserve field-specific indexing behaviour, but normalize unconditionally when no field metadata is present (search pipeline context)
- `buildCipherIndex`: add `builder.searchPipeline.add(normalizeAccentsPipelineFunction)` so lunr queries with umlauts are normalized before lookup

### DIFF
--- a/libs/common/src/vault/services/lunr-search.service.ts
+++ b/libs/common/src/vault/services/lunr-search.service.ts
@@ -13,8 +13,6 @@ import { CipherType } from "../enums/cipher-type";
 import { FieldType } from "../enums/field-type.enum";
 import { CipherViewLike, CipherViewLikeUtils } from "../utils/cipher-view-like-utils";
 
-import { normalizeSearchQuery } from "./search.service";
-
 export type IndexId = Opaque<string, "IndexId">;
 
 type IndexState = {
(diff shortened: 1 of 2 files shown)


electronicpeter:
I tested the fix by hand with a tiny vault. There everything worked fine.

jaasen-livefront:
Thank you for your contribution! This has since been fixed in https://github.com/bitwarden/clients/issues/10738.

PR #21169, opened 2026-06-10
[PM-38767] Adopt cargo-run-bin for binary tool version pinning in desktop_native

coroiu:
## 🎟️ Tracking

- [PM-38767](https://bitwarden.atlassian.net/browse/PM-38767): Adopt cargo-run-bin for binary tool version pinning in Clients / desktop-native
- Ports the SDK work from [PM-38720](https://bitwarden.atlassian.net/browse/PM-38720) (bitwarden/sdk-internal#1179) and the [PM-24643](https://bitwarden.atlassian.net/browse/PM-24643) lint-runner improvement (bitwarden/sdk-internal#1140) to the `desktop_native` workspace.

## 📔 Objective

The `desktop_native` Rust workspace pinned its tool versions in five inconsistent places:

| Tool | Pinned in |
| --- | --- |
| cargo-deny, cargo-sort, cargo-udeps | `cargo-tool-versions` |
| cargo-llvm-cov | ad-hoc `cargo install` in `test.yml` |
| cargo-xwin | ad-hoc `cargo install` in `build.js` |
| cargo-xwin (again) | error message in `apps/desktop/scripts/appx-cross-build.ps1` |
| Rust nightly | hardcoded date in `lint.yml`, unmanaged; local dev used rolling `nightly` |

Local-dev and CI versions could drift, and there was no single command to reproduce CI's lint checks locally.

This change:

- **Pins every binary tool in one place**, `[workspace.metadata.bin]` in `apps/desktop/desktop_native/Cargo.toml`, and runs them via [`cargo-run-bin`](https://crates.io/crates/cargo-run-bin) (`cargo bin <tool>`), bootstrapped once with `cargo install cargo-run-bin --locked --version 1.7.4`. Renovate-managed.
- **Pins the nightly toolchain the same way the SDK does**, via a non-standard `nightly-channel` key in `rust-toolchain.toml`. The shared Renovate preset this repo already extends (`github>bitwarden/renovate-config`) matches that key in any `rust-toolchain.toml` and updates it from a custom `rust-nightly` datasource, so no `renovate.json5` change is needed. This closes a real gap rather than a cosmetic one: `rustfmt.toml` enables unstable options (`wrap_comments`, `group_imports`, `imports_granularity`) whose output changes between nightlies, so a rolling local nightly could format code that CI's pinned nightly then rejects.
- **Reads both toolchains from that one file** in CI, via `scripts/rust-toolchain.mjs`, so a Renovate bump can no longer leave a stale literal behind in a workflow. This also removes `lint.yml`'s hardcoded `1.96.0`, which duplicated a value Renovate already manages in `rust-toolchain.toml`.
- **Enforces source-only installs** (VULN-613) via a `check-no-binstall` guard: fails CI if cargo-binstall would be used; warns locally.
- **Adds a unified lint runner** (PM-24643 improvement): `npm run lint:rust` (`--fix` / `--only <check>`) over fmt/clippy/sort/udeps/deny, mirroring CI so a local pass means CI passes.
- **Restructures the lint.yml `rust` job** into a per-check × OS matrix with per-check `.bin` caching, plus a stable `Rust lint` aggregator job.
- **Converts the dev/CI scripts to portable Node** (`.mjs`), replacing the duplicated bash + PowerShell `prepare-env` pair with a single `scripts/prepare-env-rust.mjs`, so there is one implementation to keep correct instead of two that can drift apart.

### Notable

- `cargo-xwin` is needed **twice** when cross-building for Windows, and only one of the two can go through `cargo bin`:
  - `cargoBuild()` invokes it as `cargo bin cargo-xwin build …`, omitting the `xwin` cargo-subcommand token, because cargo-run-bin supplies that token itself. The raw binary accepts both `cargo-xwin build` and `cargo-xwin xwin build`, but `cargo bin cargo-xwin xwin build` becomes `xwin xwin` and fails with `error: unrecognized subcommand 'xwin'`.
  - `buildNapiModule()` hands off to the napi CLI, which spawns `cargo xwin` as a PATH cargo-subcommand we can't redirect. So cargo-xwin is *also* installed globally, at the same version, read from `[workspace.metadata.bin]` by a small `pinnedBinVersion()` helper. One pin, two install paths.
- The `cargo-xwin` precheck in `appx-cross-build.ps1` is **removed**. It carried its own copy of `0.23.0` (covered by neither Renovate custom manager) and hard-errored *before* `npm run build-native` could perform the install that satisfies it.
- ⚠️ **Branch protection:** the matrix renames the lint jobs, from `Run Rust lint on <os>` (3 checks) to `Rust <check> (<os>)` (15 checks). Point the protected-branch required status checks at the new `Rust lint` aggregator job and drop the old names, so merges aren't blocked.

### Deliberate divergences from the SDK

The mechanism matches bitwarden/sdk-internal; three details intentionally do not:

- **Node instead of bash.** The SDK's rust lint runs on `ubuntu-24.04` only; ours runs on `ubuntu-24.04`, `macos-14` and `windows-2022`, so a single portable implementation has to be Node.
- **The toolchain is read in Node, not with `grep -oP`.** BSD grep on `macos-14` has no `-P`. (The SDK's own `lint.sh` uses `awk` for the same reason; only its workflow can assume GNU grep.)
- **Check arguments mirror clients' existing CI**, not the SDK's, so this PR doesn't silently change what CI enforces while also restructuring how it runs. `RUSTFLAGS=-D warnings` is scoped to the clippy invocation rather than exported globally, so it can't leak into the from-source tool builds `cargo bin` triggers for the other checks.

### Verified

In CI, on all three runners:

- 15/15 matrix jobs green, plus the `Rust lint` aggregator and `Rust Coverage` (which exercises `cargo bin cargo-llvm-cov` and the standalone binstall guard).
- The resolve step emits real values on each OS: `RUST_TOOLCHAIN=1.96.0`, `RUST_NIGHTLY_TOOLCHAIN=nightly-2026-06-29`.

Locally (macOS):

- All five checks pass via `npm run lint:rust`.
- `prepare-env-rust.mjs` installs the pinned nightly with rustfmt.
- All four branches of the binstall guard behave (local warn, CI error, and both the inline and table forms of a `binstall` alias).
- Pre-commit hooks run the new `cargo bin` commands, with `+nightly` resolving to the pin.

Cross-compiling for Windows, which CI never does (each platform builds natively on its own runner), verified by hand on macOS using the exact form `cargoBuild()` uses:

```
cargo bin cargo-xwin build --bin desktop_proxy --target=x86_64-pc-windows-msvc
→ Finished `dev` profile in 28.91s
→ target/x86_64-pc-windows-msvc/debug/desktop_proxy.exe: PE32+ executable (console) x86-64, for MS Windows
```

**Not verified:** the napi `--cross-compile` hand-off and the Appx packaging around it, which need `brew install msix-packaging osslsigncode`. `prepare-env-rust.mjs` is not run by CI at all (it's a developer bootstrap script, and CI installs toolchains via `dtolnay/rust-toolchain` instead), so its `win32` branches, including the `win.rustup.rs` bootstrap, are only covered by the macOS run above.

[PM-38767]: https://bitwarden.atlassian.net/browse/PM-38767?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[PM-38720]: https://bitwarden.atlassian.net/browse/PM-38720?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[PM-24643]: https://bitwarden.atlassian.net/browse/PM-24643?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


coroiu:
@claude re-review this PR

coroiu:
@claude Thanks, this was a good catch even though the conclusion went the other way. Addressed in 11b9a8e090.

## ❓ The `xwin` token question: checked empirically

I had network and a toolchain available, so I ran it rather than reasoning about `main.rs`. cargo-xwin 0.23.0 accepts the token **either way** when invoked directly, so the code was not backwards:

| Invocation | Result |
| --- | --- |
| `cargo-xwin build --help` (raw, no token) | ✅ works |
| `cargo-xwin xwin build --help` (raw, with token) | ✅ works |
| `cargo xwin build --help` (cargo dispatch) | ✅ works |
| `cargo bin cargo-xwin build …` (what the PR did) | ✅ works |
| `cargo bin cargo-xwin xwin build …` | ❌ `error: unrecognized subcommand 'xwin'` |

But **you were right that the comment's reasoning was wrong**, which is what made it look backwards. It claimed clap "does not strip" the token. What actually happens is that cargo-run-bin *supplies* the token itself, so passing it explicitly yields `xwin xwin`, which is the only form that fails. The comment now says that.

And rather than leave it as "the one line CI can't cover", I closed it end to end on macOS:

```
$ cargo xwin build --bin desktop_proxy --target=x86_64-pc-windows-msvc
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 28.91s

$ file target/x86_64-pc-windows-msvc/debug/desktop_proxy.exe
target/…/desktop_proxy.exe: PE32+ executable (console) x86-64, for MS Windows
```

## ♻️ Duplicate build: adopted, plus a reason you didn't mention

Your secondary point stands and I've taken it. `cargoBuild()` now dispatches `cargo xwin build …` and relies on the pinned global that `installTarget()` installs for napi's benefit, instead of building a second copy of the same pinned version into `.bin`.

There's an additional argument for it: `cargo install --version --locked` is a source install from crates.io, whereas `cargo bin` falls through to cargo-binstall's **prebuilt** binaries when binstall is on PATH. I hit that locally, where `cargo bin cargo-xwin` downloaded a prebuilt artifact from github.com. So for this tool the global install is the better route on VULN-613 grounds too. Re-verified with a real cross-compile through the new form.

The one thing it costs is an ordering dependency: `cargo xwin` needs `installTarget()` to have run first. That holds on every path that can reach a Windows target (both the `--target=` and `cross-platform` flows), and there's now a `NOTE:` on it.

## Non-blocking notes

Fixed:

- **`test.yml:235`** now resolves the toolchain from `rust-toolchain.toml` like `lint.yml`. Agreed this was the last place a stale literal could survive a Renovate bump.
- **`--only` with no value** now exits 2 instead of silently running every check.
- **`lint:rust` location**: added root-level `lint:rust` / `lint:rust:fix` aliases. The runner lives in the root `scripts/` and CI invokes it from the root, so requiring a `cd` was surprising. The `apps/desktop` ones stay for anyone looking there.

Verified, no change needed:

- **`nightly-channel` and the shared preset**: confirmed rather than assumed. `github>bitwarden/renovate-config` → `non-pinned.json` defines a `rust-nightly` custom datasource plus a manager matching `nightly-channel\s*=\s*"nightly-(?<currentValue>.+?)"\n` against `(^|/)rust-toolchain\.toml?$`, which matches our nested path. I tested both preset regexes against the file as written, including the trailing-newline requirement in the nightly pattern. Both match.

Deliberately not changed:

- **`run-cargo-tool.mjs` not calling `checkNoBinstall()`**: agreed it's the one unguarded `cargo bin` entry point, but it only ever runs from pre-commit, where the guard is warn-only by design. Adding it would print a warning on every commit for anyone with binstall installed, without changing any outcome. Happy to add it if you'd rather have the consistency.
- **`node` availability**: noted, no action. Both jobs rely on the runner image's preinstalled Node, which is fine on GitHub-hosted images.

## Still on me

None of the above covers the napi `--cross-compile` hand-off or the Appx packaging around it, since those need `brew install msix-packaging osslsigncode`. The description lists them under **Not verified**. And branch protection still needs the three `Run Rust lint on <os>` checks swapped for the single `Rust lint` aggregator before merge.

apps/desktop/desktop_native/build.js
```diff
@@ -117,9 +140,16 @@ function buildProcessIsolation() {
 
 function installTarget(target) {
     runCommand("rustup", ["target", "add", target]);
-    // Install cargo-xwin for cross-platform builds targeting Windows
+    // Cross-building for Windows needs cargo-xwin, reached two different ways
+    // (see the comments on each install below). Both use the single version
+    // pinned in Cargo.toml under [workspace.metadata.bin].
     if (target.includes('windows') && process.platform !== 'win32') {
-        runCommand("cargo", ["install", "--version", "0.23.0", "--locked", "cargo-xwin"]);
+        runCommand("cargo", ["install", "cargo-run-bin", "--locked", "--version", "1.7.4"]);
```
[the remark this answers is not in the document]
coroiu: Correct, and it's fallout from my own previous fix. Removed in a9f067a747.

Verified the same way you did, plus the runtime call graph: after `11b9a8e0` the only `cargo bin` mentions left in `build.js` were in comments, and nothing it reaches invokes it either (`cargoBuild()` → `cargo xwin build`, `buildNapiModule()` → napi CLI → `cargo xwin` from PATH). So every Windows cross-build was paying for a from-source `cargo install cargo-run-bin` whose output went unused.

Also folded the two-install framing back to one, as you suggested, and added an explicit note that nothing here goes through cargo-run-bin, so the next reader doesn't "restore" the bootstrap.

One knock-on you didn't mention: with that line gone, `build.js` pins no cargo-run-bin version at all, which left two dead entries in the Renovate custom manager. Both are now removed:

- the `apps/desktop/desktop_native/build\.js$` `fileMatch` entry
- the JS-array `matchStrings` pattern `'"cargo-run-bin", "--locked", "--version", "x.y.z"'`, whose only match in the repo was that line (`prepare-env-rust.mjs` uses the `CARGO_RUN_BIN_VERSION` const form, which has its own pattern)

Checked the manager still matches everything that does pin it, and no longer matches `build.js`:

```
MATCH    .github/workflows/lint.yml
MATCH    .github/workflows/test.yml
MATCH    scripts/prepare-env-rust.mjs
MATCH    .devcontainer/common/post-create-command.sh
no-match  apps/desktop/desktop_native/build.js
```

And re-confirmed the cross-build still works with no cargo-run-bin involved at all:

```
$ cargo xwin build --bin desktop_proxy --target=x86_64-pc-windows-msvc
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.56s
→ desktop_proxy.exe: PE32+ executable (console) x86-64, for MS Windows
```

apps/desktop/scripts/appx-cross-build.ps1
```diff
@@ -91,12 +91,10 @@ if (!(Get-Command osslsigncode -ErrorAction SilentlyContinue)) {
...
-if (!(Get-Command cargo-xwin -ErrorAction SilentlyContinue)) {
-    Write-Error "The `cargo-xwin` tool is required to cross-compile Windows native code."
-    Write-Error "You can install with cargo:"
-    Write-Error "  cargo install --version 0.23.0 --locked cargo-xwin"
-    Exit 1
-}
+# No cargo-xwin precheck here: `npm run build-native` (below) installs it at the
+# version pinned in desktop_native/Cargo.toml under [workspace.metadata.bin].
+# Checking for it up front would both fail before that install could run, and
+# duplicate the pinned version number into this file.
```
dani-garcia: Seems a bit unnecessary to add a comment for something we don't do, no?

dani-garcia (dismissed):

apps/desktop/scripts/appx-cross-build.ps1
coroiu: Fair, removed in a4713fe9fe. The neighbouring `makemsix`/`osslsigncode` checks are there because nothing installs those for you, so the absence of a `cargo-xwin` one doesn't really need defending in the file.

dani-garcia (approved):

PR #21177, opened 2026-06-10
[PM-38834]Allow programmatic changes to Firefox settings

dan-livefront:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-38834
## 📔 Objective

Currently, we do not make any programmatic changes to Firefox settings. This PR introduces support for leveraging the exposed `passwordSavingEnabled` endpoint to manage password-saving behavior.

Additionally, Firefox was displaying the permissions dialog behind the popup because, unlike other browsers, the popup does not automatically close when the dialog is shown. To address this, I adopted the existing Chrome approach by using local storage for Firefox as well, resulting in a consistent user experience across browsers. Otherwise, the user would have had to close the pop up to find the hidden dialog, approve it, open the dialog and check the setting again. 

## 📸 Screenshots

https://github.com/user-attachments/assets/89a8087b-2dd1-441e-b2c4-d7da110090ec


apps/browser/src/autofill/utils/pending-default-password-manager.storage.ts
```diff
@@ -0,0 +1,70 @@
...
+ * Used when enabling Bitwarden as default password manager which requires the privacy permission and the popup
+ * may close before the user responds. Uses session storage so the flag survives popup close but is cleared once the session ends.
+ */
+const pendingDefaultPasswordManagerApplyKey = "pendingDefaultPasswordManagerApply";
+
+const isWebExtensionsApi = typeof browser !== "undefined";
+
+async function getSessionStorageValue(storageKey: string): Promise<unknown> {
+  if (isWebExtensionsApi) {
+    const sessionStorage = browser?.storage?.session;
```
dan-livefront, line 13: I contemplated using local storage instead of session storage in the event a user closed the browser after granting permissions, this seems like a real scenario but then Claude called out the switch and made me contemplate going back to session. Open to suggestions on which approach to take here.

jengstrom-bw (approved):
Vault files LGTM

trmartin4:
hey @dan-livefront!  ~I noticed that this is open but there's still a "Parking work" note in the description.  Does Platform need to review this now or wait?  Thanks!~

Looks like we're good to start reviewing based on other feedback 👍 .

dan-livefront:
> hey @dan-livefront! ~I noticed that this is open but there's still a "Parking work" note in the description. Does Platform need to review this now or wait? Thanks!~
> 
> Looks like we're good to start reviewing based on other feedback 👍 .

Hey @trmartin4 ! Sorry about that, forgot to remove the note when setting to ready for review.

bensbits91 (approved):

coroiu (approved):
Feels like this should be autofill owned, but oh well 🤷

PR #21181, opened 2026-06-10
[PM-38863] [PM-17952] - Defer autofocus until the window has focus

### LINKED ISSUE #13241: Bitwarden on macOS 15.3 occasionally doesn't realease Secure Keyboard Input, preventing global hotkeys from triggering for other apps.
### Steps To Reproduce

I've not yet found a way to reliably force this bug to occur (i.e. without having to wait), but the following steps do eventually trigger the bug for me.

1. Set "Start to menu bar" and open Bitwarden without logging in (alternatively set "Close to menu bar" and close the window without logging in). This bug _might_ also occur if you leave the window open without logging in, but I haven't seen that happen myself yet.
2. Wait for some amount of time (it's hard to say exactly how much time, but I think somewhere around about 30 minutes).
3. Try triggering a global hotkey associated with a different app (e.g. the hotkey window of iterm2).

### Expected Result

The hotkey triggers successfully.

### Actual Result

The hotkey does not trigger and a notification is displayed saying that another app has Secure Keyboard Input enabled. Quitting Bitwarden fixes the issue.

### Screenshots or Videos

_No response_

### Additional Context

_No response_

### Operating System

macOS

### Operating System Version

Sequoia 15.3

### Installation method

Mac App Store

### Build Version

Version 2025.1.3 SDK 'main (28c7e29)' Shell 33.2.1 Renderer 130.0.6723.137 Node 20.18.1 Architecture arm64

### Issue Tracking Info

slifty:
## 🎟️ Tracking

  - Fixes https://github.com/bitwarden/clients/issues/13241
  - PM-17952

## 📔 Objective

On macOS, when the vault auto-locks on timeout while Bitwarden is in the background, global hotkeys / keyboard monitoring in other applications (e.g. iTerm2, BetterTouchTool, Hammerspoon, skhd, Keyboard Maestro, etc.) stop working.

**Root cause:** the lock screen's master password field is autofocused when the lock route renders even if the Bitwarden application is not focused at the OS level.

**Regression Source:** The desktop lock component already had a `deferFocus` mechanism meant to prevent this, but it was dead: it focused via `document.getElementById("masterPassword" / "pin")`, and those inputs now render through `bitInput` with auto-generated ids.  This meant the lookup in the original location always returned `null`, and the only thing focusing the field was the `appAutofocus` directive, which had no window-focus guard.

**Fix:** `AutofocusDirective` now only focuses when `document.hasFocus()` is true. When the window isn't focused, it defers focusing until the window regains focus (one-shot `window` `focus` listener, cleaned up on destroy).

This matches expected behavior: the cursor only enters the field once Bitwarden is brought to the foreground.  It also fixes every unlock method, not just master password.

In the common case (window already focused) behavior should be unchanged.

This PR has two commits:

  1. The fix in `AutofocusDirective`, with tests.
  2. Removal of the now-redundant, dead `deferFocus`/`focusInput` code in the lock component.

## 📸 Screenshots

N/A

### DIFF
--- a/libs/components/src/input/autofocus.directive.spec.ts
+++ b/libs/components/src/input/autofocus.directive.spec.ts
@@ -1,0 +1,131 @@
+import {
+  ChangeDetectionStrategy,
+  Component,
+  EventEmitter,
+  Injectable,
+  NgZone,
+} from "@angular/core";
+import { ComponentFixture, fakeAsync, TestBed, tick } from "@angular/core/testing";
+
+import { Utils } from "@bitwarden/common/platform/misc/utils";
+
+import { AutofocusDirective } from "./autofocus.directive";
+
+@Injectable()
+class MockNgZone extends NgZone {
+  override onStable: EventEmitter<any> = new EventEmitter(false);
+  isStable = true;
+
+  constructor() {
+    super({ enableLongStackTrace: false });
+  }
+
+  override run(fn: any): any {
+    return fn();
+  }
+
+  override runOutsideAngular(fn: any): any {
+    return fn();
+  }
+}
+
+@Component({
+  template: `<input appAutofocus />`,
+  imports: [AutofocusDirective],
+  changeDetection: ChangeDetectionStrategy.OnPush,
+})
+class TestHostComponent {}
+
+describe("AutofocusDirective", () => {
+  beforeEach(() => {
+    TestBed.configureTestingModule({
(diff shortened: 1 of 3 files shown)


slifty:
I won't be a pain and regularly bump but just a note that I'll be glad to rebase / do whatever is needed to resolve conflicts and re-test when the time does come for review!

PR #16247, opened 2025-09-01
[PM-27847] Enable biometric IPC on mac dmg

quexten:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-27847

## 📔 Objective

Enables mac dmg biometrics integration. Historically, this was not allowed by gatekeeper, mac's security layer. This can no longer be reproduced, and the integration works.

## ⏰ Reminders before review

- Contributor guidelines followed
- All formatters and local linters executed and passed
- Written new unit and / or integration tests where applicable
- Protected functional changes with optionality (feature flags)
- Used internationalization (i18n) for all UI strings
- CI builds passed
- Communicated to DevOps any deployment requirements
- Updated any necessary documentation (Confluence, contributing docs) or informed the documentation team

## 🦮 Reviewer guidelines

- 👍 (`:+1:`) or similar for great changes
- 📝 (`:memo:`) or ℹ️ (`:information_source:`) for notes or general info
- ❓ (`:question:`) for questions
- 🤔 (`:thinking:`) or 💭 (`:thought_balloon:`) for more open inquiry that's not quite a confirmed issue and could potentially benefit from discussion
- 🎨 (`:art:`) for suggestions / improvements
- ❌ (`:x:`) or ⚠️ (`:warning:`) for more significant problems or concerns needing attention
- 🌱 (`:seedling:`) or ♻️ (`:recycle:`) for future improvements or indications of technical debt
- ⛏ (`:pick:`) for minor or nitpick changes

PR #21195, opened 2026-06-11
[PM-38187] Improve isSnapStore detection

djsmith85:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-38187

### Changes
- **Move all platform/process detection methods into a new platform-utils.main.ts** - 38c6d6848f1b6365778470ac1460f50d93215ec6
  - `utils.ts` was being picked up by the compiler/bundler to be included in the renderer and any methods exposed would fail. The correct way to call them on renderer is using ipc.platform.XXX
- **Tighten isSnapStore detection by checking process.env.SNAP and process.env.SNAP_NAME instead of process.env.SNAP_USER_DATA** - 0bc0f8c750374bc81104ff6c8f6979a4e7bdf334
- **Improve detection of Snap sandbox** - 96993254c2963f7492a57d725a28ddae0ceb5f0b


apps/desktop/src/main/platform-utils.main.ts
```diff
@@ -0,0 +1,54 @@
+/**
+ * We block the browser integration on some unsupported platforms prevents
+ * experimenting with the feature for QA. So this env var allows overriding
+ * the block.
+ */
+export function allowBrowserintegrationOverride() {
```
quexten, line 6: TBH I think at this ponit we can remove this. We rolled out support to flatpak, snap, mac app store, so this goes unused.

apps/desktop/src/main/platform-utils.main.ts
```diff
@@ -0,0 +1,54 @@
...
+    (process.resourcesPath?.indexOf("8bitSolutionsLLC.bitwardendesktop_") > -1 ||
+      process.resourcesPath?.indexOf("8bitSolutionsLLC.BitwardenBeta_") > -1)
+  ) {
+    windowsStore = true;
+  }
+  return windows && windowsStore === true;
+}
+
+export function isFlatpak() {
+  return process.platform === "linux" && process.env.container != null;
```
quexten, line 56: On `desktop_native` we use:
```rust
    std::env::var("container") == Ok("flatpak".to_string())
```

I wonder whether we can consolidate...

quexten (commented):
Since this is mostly a move, these comments are optional, but I'll note them because we don't ofter otherwise touch these files.

apps/desktop/src/main/platform-utils.main.ts
djsmith85: Since this was introduced by KM with https://github.com/bitwarden/clients/pull/16247 and they can best decide on the current state of biometrics could you please create a follow-up ticket.

apps/desktop/src/main/window.main.ts
```diff
@@ -48,6 +42,43 @@ protocol.registerSchemesAsPrivileged([
...
+  if (!isLinux()) {
+    return false;
+  }
+  // execPath is the real binary path, not settable via env
+  if (!SNAP_STORE_NAMES.some((name) => process.execPath.startsWith(`/snap/${name}/`))) {
+    return false;
+  }
+  try {
+    const cgroup = fs.readFileSync("/proc/self/cgroup", "utf8");
+    return /snap\.bitwarden(-beta)?\b/.test(cgroup);
```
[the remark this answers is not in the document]
djsmith85: Addressed with 10e68d14a2d2d1b8758dab9aa0c64062964aaac7

apps/desktop/src/main/window.main.ts
```diff
@@ -48,6 +42,43 @@ protocol.registerSchemesAsPrivileged([
...
+ *
+ * Lives here rather than alongside `isSnapStore()` because reading the cgroup needs `fs`, and
+ * platform-utils.main is bundled into the sandboxed preload where Node builtins fail to load.
+ */
+export function isConfinedSnap() {
+  if (!isLinux()) {
+    return false;
+  }
+  // execPath is the real binary path, not settable via env
+  if (!SNAP_STORE_NAMES.some((name) => process.execPath.startsWith(`/snap/${name}/`))) {
```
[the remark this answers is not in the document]
djsmith85: Using `.includes()` would weaken the detection.

Addressed with 0493b2a88e98b7d9e62ad26b97f1d40336f976ce

coroiu (dismissed):

quexten (dismissed):
Approving for km owned files.

quexten (approved):

coroiu (approved):

PR #21198, opened 2026-06-11
[PM-38958] Add ESLint no package self import rule

quexten:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-38958

## 📔 Objective
We waste precious review time on relative imports. Human reviewers should not do this mechanical work. We have ~200 violations of this rule currently. This task is establishing an eslint rule and fixing existing violations, freeing human review time.

Add a custom ESLint rule `@bitwarden/platform/no-self-package-import` that flags imports referencing a package's own `@bitwarden/*` alias instead of relative paths. The rule self-limits to the file's owning package.

Scoped to `libs/**` and `bitwarden_license/bit-common/src/**` as a `warn` for now. Will be promoted to `error` once existing self-alias imports are converted (`eslint --fix`).

Ref: https://contributing.bitwarden.com/contributing/code-style/web/typescript#imports-within-the-same-package

> [!NOTE]  
> There are 200 existing violations to the rule. We will auto-create follow-up PR's per team.

## 📸 Screenshots

n/a


eslint.config.mjs
```diff
@@ -640,6 +640,16 @@ export default tseslint.config(
     },
   },
 
+  // Within a package, import sibling code via relative paths rather than the package's own
+  // `@bitwarden/*` alias. Scoped to libs here; the rule self-limits to the file's owning package.
+  // TODO: change to "error" once existing self-alias imports are converted (run `eslint --fix`).
+  {
+    files: ["libs/**/*.ts", "bitwarden_license/bit-common/src/**/*.ts"],
+    rules: {
+      "@bitwarden/platform/no-self-package-import": "warn",
```
quexten, line 650: There will be subsequent PR's migrating everyone, then we will make this `error`

coroiu (approved):
Hey, nice, why didn't I think of this!?

mzieniukbw:
Can we revert this change, until all the related PR's are merged ? Merging `main` into a branch will run git hooks, which runs this eslint rule but with `--fix`, which in my PR resulted in ~120 file changes with unrelated import changes.

PR #21211, opened 2026-06-11
feat: add Region.Gov to available environments

addisonbeck:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-38018
https://bitwarden.atlassian.net/browse/PM-38019

## 📔 Objective

Add `bitwarden-gov.com` as a third cloud region behind the `FedRampGovRegion` feature flag.

Commit 1 adds `Region.Gov` to the `Region` const-object and a Gov entry to `PRODUCTION_REGIONS` covering every `bitwarden-gov.com` subdomain. The Gov region becomes globally addressable: `EnvironmentService.availableRegions()` returns it, origin validation accepts it, stored-region rehydration resolves it.

Commit 2 introduces `AvailableRegionsService`, a flag-gated wrapper around `EnvironmentService.availableRegions()` that filters Gov from the UI display source when `FedRampGovRegion` is off. The wrapper exists because `ConfigService` already depends on `EnvironmentService` to resolve API URLs, so wiring the flag check into `EnvironmentService` directly would create a circular dependency.

Commit 3 migrates the three Auth-owned environment-selector components to consume `availableRegions$` for dropdown options while keeping unfiltered `availableRegions()` for region-identity lookups (current region, host match). 

When `FedRampGovRegion` is on, Gov appears in every environment selector alongside US and EU. With the flag off, Gov stays invisible to the UI but remains resolvable for any user who lands on `vault.bitwarden-gov.com` directly.

Shipping the service without the selector migration would briefly expose Gov in every selector with no flag gate, because unmigrated selectors read `EnvironmentService.availableRegions()` directly. For that reason I've introduced the new region, and new Platform service, and gone updated all the callers in Auth code as a single PR.

## 📸 Screenshots

Screenshots taken from the desktop app's environment selector. First is flag off, second is flag on. If you try and log in to gov it fails, since there isn't anything deployed for that URL right now.

Screenshot 2026-06-16 at 4 11 03 PM

Screenshot 2026-06-16 at 4 59 54 PM


djsmith85 (dismissed):
Approving unowned and platform-owned changes

libs/auth/src/angular/registration/registration-env-selector/registration-env-selector.component.ts
JaredSnider-Bitwarden, line 1: ⚠️  Claude (unvetted b/c storybook won't build on this branch): Storybook stories missing `AvailableRegionsService` provider; all `RegistrationStartComponent` stories will throw at render
  - `libs/auth/src/angular/registration/registration-start/registration-start.stories.ts:78-90`

JaredSnider-Bitwarden (changes requested):
The primary changes look good to me for all auth owned env selectors - however, a local Claude review caught one issue we should address - registration start storybook stories need updating.

libs/auth/src/angular/registration/registration-env-selector/registration-env-selector.component.ts
addisonbeck: Corrected on https://github.com/bitwarden/clients/pull/21211/commits/6781afac0400f5de0bfaec948c2d4b52ff9312dc

Thanks for the catch! I don't work in storybook ever so let me know if you think I should do something else, like adding a story that has gov in the mock.

JaredSnider-Bitwarden: Np, thank you for the fix. LGTM!

JaredSnider-Bitwarden (dismissed):

djsmith85 (dismissed):

JaredSnider-Bitwarden (dismissed):

addisonbeck:
@JaredSnider-Bitwarden @djsmith85 sorry for the review pings on this. A [bug](https://bitwarden.atlassian.net/browse/PM-39401) got logged for this that I was able to recreate on the browser build from the original PR's commit log: https://github.com/bitwarden/clients/pull/21211/commits/6781afac0400f5de0bfaec948c2d4b52ff9312dc

I had no clue how the changes in this PR introduced such a bug, and was testing https://github.com/bitwarden/clients/pull/21211/commits/2d51461be9a65892b8832455a5c802285022683c as a potential fix.

This "worked", but I was really confused by it, so I reverted it. The latest build artifacts are from the revert, and they also don't have the bug.

I'm assuming something in the merges from main fixed this, but really I don't know what happened. Regardless: this PR should be in the same state it was in when you originally approved, besides merges from main.

addisonbeck:
Tests seem to be failing on main, and are not related

JaredSnider-Bitwarden (approved):

djsmith85 (approved):
Yeah, tests are failing on main. Changes here are looking good.

addisonbeck:
the chromatic diff demonstrates that there is now a small delay between first render and the environment box default value being populated. This is expected, because the environment list is no longer static (it's filtered through feature flags and emitted), and became async rendered in the component. This is not detectable in normal use.

PR #21212, opened 2026-06-11
[PM-38393] bypass cloud check for stripe checkout for qa

kdenney:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-38393

## 📔 Objective

Adds the QA bypass feature flag `debug-disable-self-host-premium-check` (matching the key used on the mobile side) so QA can exercise the browser/desktop Stripe checkout flow while a client is pointed at an internal (cloud) server, whose region the client still resolves as self-hosted.


connerbw (commented):
@kdenney You'll need similar updates to the conditionals at

https://github.com/bitwarden/clients/blob/billing/pm-38393/disable-self-host-premium-check-flag/libs/common/src/billing/services/subscription-pricing.service.ts#L91

and

https://github.com/bitwarden/clients/blob/billing/pm-38393/disable-self-host-premium-check-flag/libs/common/src/billing/services/subscription-pricing.service.ts#L102

otherwise the price won't display

kdenney:
> @kdenney You'll need similar updates to the conditionals at
> 
> https://github.com/bitwarden/clients/blob/billing/pm-38393/disable-self-host-premium-check-flag/libs/common/src/billing/services/subscription-pricing.service.ts#L91
> 
> and
> 
> https://github.com/bitwarden/clients/blob/billing/pm-38393/disable-self-host-premium-check-flag/libs/common/src/billing/services/subscription-pricing.service.ts#L102
> 
> otherwise the price won't display

@connerbw Great catch! Thanks for pointing these out. I've added checks to this file in the latest push.

connerbw (approved):

PR #21222, opened 2026-06-12
[PM-38867] Fix: add error toast if no owner for a new item 

eliykat:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-38867

## 📔 Objective

With #20377, we are enforcing policies as soon as a user accepts their invite. However, this causes the following problem:

> A user in the accepted state has the Centralize Org Policy enforced against them, so they can’t create personal items, but they also can’t create organization items until they’re confirmed. This is an in-between state we didn’t have previously.

Current behavior:
- `ItemDetailsSectionComponent` does check for this state, but it [throws an error](https://github.com/bitwarden/clients/blob/main/libs/vault/src/cipher-form/components/item-details/item-details-section.component.ts#L223) that is unhandled and leaves the form partially initialized (particularly noticeable when editing an existing item - the load logic never gets executed)
- when the save button is clicked, it shows a `1 field needs attention` error, but with no inline error message.

New behavior:
- remove the unhandled throw - instead default to `null` (personal ownership) in this case. This allows the form to load properly even though it's invalid, and this will be caught on submit below.
- move the check to the submit method and show a toast: "Unable to save item. You must be confirmed to the organization to save items."

This is a quick fix for the `rc` cut on Monday. Product & Design will refine a nicer solution as follow-up work.

## 📸 Screenshots

https://github.com/user-attachments/assets/440563c5-7e70-4add-a75b-acf0de87a336

### DIFF
--- a/apps/browser/src/_locales/en/messages.json
+++ b/apps/browser/src/_locales/en/messages.json
@@ -6681,5 +6681,8 @@
   },
   "errorImportingMyItemsMultiCollection": {
     "message": "You cannot import items into My Items that are part of multiple collections"
+  },
+  "cannotSaveItemNoConfirmedOrgs": {
+    "message": "Unable to save item. You must be confirmed to the organization to save items."
   }
 }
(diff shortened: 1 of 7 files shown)


shane-melton (approved):
Changes look good! Thanks for the tests as well!

PR #21240, opened 2026-06-12
[PM- 38988] Fix Trial Dialog Dismissal Caching

sbrown-livefront:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-38988

## 📔 Objective
During a payment-optional free trial, the trial payment modal was re-appearing every time the user navigated between pages in the admin console (e.g. Members → Collections). Dismissing via the "Later" button, the X button, or clicking outside the modal had no effect on subsequent page navigations.

**Root cause:** `OrganizationWarningsService.showSubscribeBeforeFreeTrialEndsDialog$` had no dismissal tracking. Both `members.component.ts` and `vault.component.ts` subscribe to this method via `switchMap` on `organization$`, so each navigation created a new
      subscription that re-opened the dialog.

**Fix:** Persist per-organization dismissal state using `StateProvider` with a `UserKeyDefinition` stored in `BILLING_DISK_LOCAL` (localStorage). Once a user dismisses the modal for an org — via "Later", the X button, ESC, or clicking outside — it will not appear again. Dismissal is permanent but does not exist if client changes.

### Initial Dismiss of Modal Able to Navigate without Dialog
https://github.com/user-attachments/assets/357ddf2d-09ed-4b9d-810f-edfeb04973ff

### New Client (Incognito) without Session Storage Dialog Opens but Close Persists
https://github.com/user-attachments/assets/489804e1-b6b7-492f-9b48-41270f91fc12

### DIFF
--- a/apps/web/src/app/billing/organizations/warnings/services/organization-warnings.service.ts
+++ b/apps/web/src/app/billing/organizations/warnings/services/organization-warnings.service.ts
@@ -296,13 +296,13 @@ export class OrganizationWarningsService {
             break;
           }
           default: {
-            // Covers CLOSED, X button, ESC, and clicking outside (all emit undefined)
+            // Covers CLOSED, X button, ESC, and clicking outside (all emit undefined).
+            // Uses .update() to read current state at write time, preventing stale overwrites
+            // if multiple tabs dismiss different orgs concurrently.
             try {
-              await this.stateProvider.setUserState(
-                TRIAL_PAYMENT_MODAL_DISMISSED_ORGS_KEY,
-                { ...(dismissedOrgs ?? {}), [organization.id]: true },
-                account.id,
-              );
+              await this.stateProvider
+                .getUser(account.id, TRIAL_PAYMENT_MODAL_DISMISSED_ORGS_KEY)
+                .update((current) => ({ ...(current ?? {}), [organization.id]: true }));
             } catch (error) {
               this.logService.error("Failed to save trial payment modal dismissal state:", error);
             }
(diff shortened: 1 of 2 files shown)


apps/web/src/app/billing/organizations/warnings/services/organization-warnings.service.ts
```diff
@@ -255,9 +287,27 @@ export class OrganizationWarningsService {
...
+            break;
+          }
+          default: {
+            // Covers CLOSED, X button, ESC, and clicking outside (all emit undefined)
+            try {
+              await this.stateProvider.setUserState(
+                TRIAL_PAYMENT_MODAL_DISMISSED_ORGS_KEY,
+                { ...(dismissedOrgs ?? {}), [organization.id]: true },
+                account.id,
+              );
```
connerbw: ⚠️ This has a very small chance of saving stale data. It's very much an edge case, I think, so this is by no means a blocking comment.

`dismissedOrgs` is read up on line 271, then the dialog is open and closed before it's saved here. In some off chance that this user is running two tabs for different orgs, this dialog opens on both tabs, the user clicks close on one of them, then closes the next one, that second close will wipe out the first one. Again, very small chance that this happens. Reading the current stored value at the exact moment before we save should mitigate this. I could also just be wrong, so feel free to dismiss if you think so.


```suggestion
              await this.stateProvider
                .getUser(account.id, TRIAL_PAYMENT_MODAL_DISMISSED_ORGS_KEY)
                .update((current) => ({ ...(current ?? {}), [organization.id]: true }));
```

connerbw (dismissed):

connerbw (approved):

PR #21289, opened 2026-06-16
feat(browser): Add autofill for SSH keys

quexten:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-39609
https://bitwarden.slack.com/archives/C09CE6NSPHC/p1782462926838819

## 📔 Objective

Adds autofill (inline and regular) for SSH-keys. Recorded and tested are GitHub and GitLab, Hetzner, BorgBase, DigitalOcean.

## 📸 Screenshots

image

(Please note, below did not have a fix for the icon missing applied yet, so they are missing the key icon).
https://github.com/user-attachments/assets/251554e6-96a1-4d12-a45a-7764f89a7274

https://github.com/user-attachments/assets/2e002dd1-2bba-4ffb-9b3e-c92dc8539128

borgbase

Note: On hetzner autofill was not available, and the page showed a popup that bitwarden inline autofill disabled itself because the page was interfering.
hetzner

digitialocean

### DIFF
--- a/apps/browser/src/_locales/en/messages.json
+++ b/apps/browser/src/_locales/en/messages.json
@@ -1944,7 +1944,7 @@
   "showInlineMenuCardsLabel": {
     "message": "Display cards as suggestions"
   },
-  "showInlineMenuSshKeysLabel": {
+  "showInlineMenuSSHKeysLabel": {
     "message": "Display SSH keys as suggestions"
   },
   "showInlineMenuOnIconSelectionLabel": {
--- a/apps/browser/src/_locales/en/messages.json
+++ b/apps/browser/src/_locales/en/messages.json
@@ -1944,6 +1944,9 @@
   "showInlineMenuCardsLabel": {
     "message": "Display cards as suggestions"
   },
+  "showInlineMenuSshKeysLabel": {
+    "message": "Display SSH keys as suggestions"
+  },
   "showInlineMenuOnIconSelectionLabel": {
     "message": "Display suggestions when icon is selected"
   },
--- a/apps/browser/src/autofill/background/overlay.background.ts
+++ b/apps/browser/src/autofill/background/overlay.background.ts
@@ -117,7 +117,12 @@ import {
   PasswordGenerateRequestSource,
 } from "./abstractions/overlay.background";
 
-const cardAndIdentityCipherType: CipherType[] = [CipherType.Card, CipherType.Identity];
+// Non-login cipher types that are fetched and cached for the inline menu regardless of URL match.
+const cardAndIdentityCipherType: CipherType[] = [
+  CipherType.Card,
+  CipherType.Identity,
+  CipherType.SshKey,
+];
 
 export class OverlayBackground implements OverlayBackgroundInterface {
   // Assigned as members so jest.spyOn can intercept them in tests
@@ -189,6 +194,7 @@ export class OverlayBackground implements OverlayBackgroundInterface {
       ),
     getInlineMenuCardsVisibility: () => this.getInlineMenuCardsVisibility(),
     getInlineMenuIdentitiesVisibility: () => this.getInlineMenuIdentitiesVisibility(),
+    getInlineMenuSshKeysVisibility: () => this.getInlineMenuSshKeysVisibility(),
     closeAutofillInlineMenu: ({ message, sender }) =>
       void this.withSenderTab(sender, () => this.closeInlineMenu(sender, message)),
     checkAutofillInlineMenuFocused: ({ sender }) =>
@@ -580,6 +586,7 @@ export class OverlayBackground implements OverlayBackgroundInterface {
       await this.cipherService.getAllDecryptedForUrl(currentTab.url || "", userId, [
         CipherType.Card,
         CipherType.Identity,
+        CipherType.SshKey,
       ])
     ).sort((a, b) => this.cipherService.sortCiphersByLastUsedThenName(a, b));
 
@@ -737,6 +744,12 @@ export class OverlayBackground implements OverlayBackgroundInterface {
             continue;
           }
           break;
+
+        case CipherType.SshKey:
+          if (areKeyValuesNull(cipher.sshKey, ["publicKey"])) {
+            continue;
+          }
+          break;
       }
       if (!this.focusedFieldMatchesFillType(cipher.type)) {
         continue;
@@ -885,6 +898,11 @@ export class OverlayBackground implements OverlayBackgroundInterface {
 
     if (cipher.type === CipherType.Card) {
       inlineMenuData.card = cipher.card.subTitle;
+      return inlineMenuData;
+    }
+
+    if (cipher.type === CipherType.SshKey) {
+      inlineMenuData.sshKey = cipher.sshKey?.keyFingerprint;
       return inlineMenuData;
     }
 
@@ -1941,20 +1959,24 @@ export class OverlayBackground implements OverlayBackgroundInterface {
       return {};
     }
 
-    let elementOffset = height * 0.37;
-    if (height >= 35) {
-      elementOffset = height >= 50 ? height * 0.47 : height * 0.42;
+    // Cap the height used for sizing the button so tall fields (e.g. textareas for SSH public
+    // keys) don't scale the icon up. Single-line inputs are at or under this cap and unaffected.
+    const sizingHeight = Math.min(height, 64);
+
+    let elementOffset = sizingHeight * 0.37;
+    if (sizingHeight >= 35) {
+      elementOffset = sizingHeight >= 50 ? sizingHeight * 0.47 : sizingHeight * 0.42;
     }
 
     const fieldPaddingRight = parseInt(paddingRight ?? "", 10);
     const fieldPaddingLeft = parseInt(paddingLeft ?? "", 10);
-    const elementHeight = height - elementOffset;
+    const elementHeight = sizingHeight - elementOffset;
 
     const elementTopPosition = subFrameTopOffset + top + elementOffset / 2;
     const elementLeftPosition =
       fieldPaddingRight > fieldPaddingLeft
-        ? subFrameLeftOffset + left + width - height - (fieldPaddingRight - elementOffset + 2)
-        : subFrameLeftOffset + left + width - height + elementOffset / 2;
+        ? subFrameLeftOffset + left + width - sizingHeight - (fieldPaddingRight - elementOffset + 2)
+        : subFrameLeftOffset + left + width - sizingHeight + elementOffset / 2;
 
     const button = {
       top: Math.round(elementTopPosition),
@@ -2485,6 +2507,13 @@ export class OverlayBackground implements OverlayBackgroundInterface {
    */
   private async getInlineMenuIdentitiesVisibility(): Promise<boolean> {
     return await firstValueFrom(this.autofillSettingsService.showInlineMenuIdentities$);
+  }
+
+  /**
+   * Gets the inline menu's visibility setting for SSH keys from the settings service.
+   */
+  private async getInlineMenuSshKeysVisibility(): Promise<boolean> {
+    return await firstValueFrom(this.autofillSettingsService.showInlineMenuSshKeys$);
   }
 
   /**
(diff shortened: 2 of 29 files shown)


apps/browser/src/_locales/en/messages.json
```diff
@@ -1922,6 +1922,9 @@
   "showInlineMenuCardsLabel": {
     "message": "Display cards as suggestions"
   },
+  "showInlineMenuSshKeysLabel": {
```
jprusik: the nittiest nit (non-blocking):

```suggestion
  "showInlineMenuSSHKeysLabel": {
```

apps/browser/src/autofill/background/overlay.background.ts
```diff
@@ -1941,20 +1959,24 @@ export class OverlayBackground implements OverlayBackgroundInterface {
...
-    const elementHeight = height - elementOffset;
+    const elementHeight = sizingHeight - elementOffset;
 
     const elementTopPosition = subFrameTopOffset + top + elementOffset / 2;
     const elementLeftPosition =
       fieldPaddingRight > fieldPaddingLeft
-        ? subFrameLeftOffset + left + width - height - (fieldPaddingRight - elementOffset + 2)
-        : subFrameLeftOffset + left + width - height + elementOffset / 2;
+        ? subFrameLeftOffset + left + width - sizingHeight - (fieldPaddingRight - elementOffset + 2)
+        : subFrameLeftOffset + left + width - sizingHeight + elementOffset / 2;
```
jprusik: cc @dan-livefront

apps/browser/src/autofill/browser/cipher-context-menu-handler.ts
```diff
@@ -70,6 +70,9 @@ export class CipherContextMenuHandler {
         [CipherType.Login]: [],
         [CipherType.Card]: [],
         [CipherType.Identity]: [],
+        // SSH keys are not surfaced in the right-click context menu yet, but the grouping
+        // record must cover every AutofillCipherTypeId.
+        [CipherType.SshKey]: [],
```
jprusik, line 75: 👍

apps/browser/src/autofill/services/autofill-constants.ts
```diff
@@ -1008,6 +1008,48 @@ export class IdentityAutoFillConstants {
...
+  static readonly SshKeyAttributes: string[] = [
+    "autoCompleteType",
+    "htmlName",
+    "htmlID",
+    "htmlClass",
+    "label-tag",
+    "label-left",
+    "label-top",
+    "placeholder",
+    "title",
```
jprusik: One risk of a broader attribute set here is false positives. If we start seeing that, we might narrow these somewhat.

apps/browser/src/autofill/services/autofill-constants.ts
```diff
@@ -1008,6 +1008,48 @@ export class IdentityAutoFillConstants {
   };
 }
 
+export class SshKeyAutoFillConstants {
+  /** Field attributes scanned to gather matching keywords. */
+  static readonly SshKeyAttributes: string[] = [
+    "autoCompleteType",
```
jprusik: Is there a specific value you had in mind? On quick scan, I didn't [see anything](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Attributes/autocomplete#value) that felt like a strong signpost of what the input content represents. If you agree, we should probably drop this one.

apps/browser/src/autofill/services/autofill-constants.ts
```diff
@@ -1008,6 +1008,48 @@ export class IdentityAutoFillConstants {
...
+    "ssh-rsa",
+    "ssh-ed25519",
+    "ssh-dss",
+    "ecdsa-sha2-",
+    "sk-ssh-ed25519",
+    "sk-ecdsa-sha2-",
+  ];
+
+  static readonly PublicKeyFieldNames: string[] = [
+    "key",
```
jprusik: It might be prudent to drop this one; I suspect it's likely to trigger false-positives on keyword value collision of unrelated concerns (we try to avoid <4 character values with our current string matching strategy for this reason).

apps/browser/src/autofill/services/autofill-overlay-content.service.ts
```diff
@@ -1171,6 +1173,16 @@ export class AutofillOverlayContentService implements AutofillOverlayContentServ
       return false;
     }
 
+    // SSH key fields must be checked before the ignored-types early return, since the SSH
+    // public key field is a textarea (which is otherwise an ignored inline menu field type).
```
jprusik, line 1195: 👍

apps/browser/src/autofill/services/autofill.service.ts
```diff
@@ -1884,6 +1888,163 @@ export default class AutofillService implements AutofillServiceInterface {
...
+    const keywords: Set<string> = new Set();
+    for (let index = 0; index < SshKeyAutoFillConstants.SshKeyAttributes.length; index++) {
+      const attribute = SshKeyAutoFillConstants.SshKeyAttributes[index];
+      const value = field[attribute];
+      if (value != null && typeof value === "string") {
+        keywords.add(
+          value
+            .trim()
+            .toLowerCase()
+            .replace(/[^a-zA-Z0-9]+/g, ""),
```
jprusik: nit, somewhat silly regex optimization (could probably get even more clever, to the point of causing more issues 🤷 )

```suggestion
            .replace(/[^a-z0-9]+/gi, ""),
```

jprusik: thinking out loud: also worth noting, since we're dealing with attribute values which can have arbitrary values, we should probably support extended/unicode cases, though we don't have any match values of that sort at the moment, so a premature optimization for sure

apps/browser/src/autofill/services/autofill.service.ts
```diff
@@ -1884,6 +1888,163 @@ export default class AutofillService implements AutofillServiceInterface {
...
+
+  /**
+   * Identifies if a field is the SSH key title field. Limited to non-textarea inputs whose
+   * keywords match a title field name. Callers should additionally confirm a public key field
+   * is present before filling.
+   *
+   * @param field - The field to check
+   */
+  private isSshTitleField(field: AutofillField): boolean {
+    if (field.tagName === "textarea" || this.isSshPublicKeyField(field)) {
```
jprusik, line 2099: We could probably get more clever about the `tagName` check to early return, but this is fine for now.

apps/browser/src/autofill/background/overlay.background.ts
```diff
@@ -117,7 +117,12 @@ import {
   PasswordGenerateRequestSource,
 } from "./abstractions/overlay.background";
 
-const cardAndIdentityCipherType: CipherType[] = [CipherType.Card, CipherType.Identity];
+// Non-login cipher types that are fetched and cached for the inline menu regardless of URL match.
+const cardAndIdentityCipherType: CipherType[] = [
```
jprusik: hrm, perhaps a rename is in order? 🤔 

```suggestion
const nonLoginInlineCipherType: CipherType[] = [
```

apps/browser/src/autofill/services/autofill-constants.ts
```diff
@@ -1008,6 +1008,48 @@ export class IdentityAutoFillConstants {
...
+    "public-key",
+    "publickey",
+    "ssh-key",
```
jprusik: While our current string-matching strategy will strip the `-`, yielding some duplicate values, our practice has been to capture real unique value patterns in anticipation of other matching strategies in the future (what you've done here is desireable).

jprusik:
Custom fields are enabled for SSH ciphers; is this intentional? Is there any expectation of those values autofilling in real scenarios?

Image

apps/browser/src/autofill/background/overlay.background.ts
jprusik, line 1: When there is no SSH cipher and an SSH field is selected, "New item" appears. Clicking that does nothing:

https://github.com/user-attachments/assets/5617a14d-8442-4db4-8514-6626c7b44269

I'm fine with deferring this, but we should capture any followup in a ticket.

apps/browser/src/autofill/background/overlay.background.ts
```diff
@@ -888,6 +901,11 @@ export class OverlayBackground implements OverlayBackgroundInterface {
       return inlineMenuData;
     }
 
+    if (cipher.type === CipherType.SshKey) {
+      inlineMenuData.sshKey = cipher.sshKey?.keyFingerprint;
```
jprusik: Image

I'm somewhat apprehensive about exposing any real part of the ssh key when a cipher title already allows disambiguation, as a predictable/targetable concern and as a privacy/fingerprinting concern (the less we have to inject into the page inline menu experience, the better)

We do similar things elsewhere with username and identity but those are typically required contextually since the default titles are the domain. SSH ciphers are domainless and so are likely to have meaningful names (and a user isn't likely to have memorized their key fingerprint anyway).

Is including the fingerprint an important part of the UX here, or can we drop it?

apps/browser/src/autofill/background/overlay.background.ts
jprusik, line 1: It looks like there's some odd background coloring on the inline menu icon we should take a look at:

Image

jprusik (commented):
I did some smoke-testing with this, and it's really nice! Overall, things are put together very thoughtfully. I have a few questions, but this is nearly there, imo.

jprusik:
Not part of the work here, but I was curious; when creating an SSH cipher, it auto-generates a key for me; I hadn't expected that and it threw me at first. I wonder if we should consider some kind of positioning that the keys are being _generated_, not provided by the user.

I also expected to be able to generate my own values and store them as an SSH cipher, which doesn't seem to be possible; they are read only, even in edit mode. I _have_ to use the extension keys for this cipher type.

apps/browser/src/_locales/en/messages.json
quexten: Just noting, in prior SDK work I got the opposite feedback: "SSH -> Ssh". (I prefer SSH too though...)

apps/browser/src/autofill/background/overlay.background.ts
quexten: @jprusik  Is there an actionable change you want from my side here?

quexten:
> Not part of the work here, but I was curious; when creating an SSH cipher, it auto-generates a key for me; I hadn't expected that and it threw me at first. I wonder if we should consider some kind of positioning that the keys are being generated, not provided by the user.

> I also expected to be able to generate my own values and store them as an SSH cipher, which doesn't seem to be possible; they are read only, even in edit mode. I have to use the extension keys for this cipher type. 

Actually, both are supported, even if that is not super clear. You can import via a button. For this you would create -> It generates -> you import -> it overrides.

The UI is sub-par currently :( @maxkpower Do you know when / who would work on an updated UI for creating? Tools?

quexten:
@jprusik Adding new items works now; the weird icon rendering is fixed. I'm not too familiar with the UI code here, and just tampered around with it until it worked, so please beware / double check. 

https://github.com/user-attachments/assets/6d75951f-170e-4e18-a024-fc9acb395556

apps/browser/src/autofill/background/overlay.background.ts
quexten: Fixed, see recording.

apps/browser/src/autofill/background/overlay.background.ts
quexten: Dropped.

apps/browser/src/autofill/services/autofill-constants.ts
quexten: Do you mean `autoCompleteType` specifically? I think I just copied it from the constants for cards. Removed.

apps/browser/src/autofill/background/overlay.background.ts
jprusik: nope, just giving Dan a heads since this may impact some component work in this area

apps/browser/src/autofill/services/autofill-constants.ts
jprusik: right - there's no "ssh" kind of value for the `autocomplete` attribute

jprusik (approved):

gbubemismith (approved):
Vault changes look good

quexten:
Merging since `rc` has been cut.

PR #21290, opened 2026-06-16
[PM-27041] - CLI - respect passphrase param in /generate

jaasen-livefront:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-27041

## 📔 Objective

Fix a bug in the CLI serve `/generate` route where query-string boolean values were handled by truthiness instead of boolean parsing.

Previously, when `passphrase` was present as a query parameter, `"false"` was treated as truthy, so the command generated a passphrase even for:

`/generate?password=true&...&passphrase=false`

This PR updates option normalization to parse `passphrase` via `CliUtils.convertBooleanOption(...)` before selecting generation type, so:
- `passphrase=true` => passphrase
- `passphrase=false` => password

Also adds regression tests in `GenerateCommand` to lock this behavior.

## 📸 Screenshots

N/A (no UI changes)

### DIFF
--- a/apps/cli/src/tools/generate.command.spec.ts
+++ b/apps/cli/src/tools/generate.command.spec.ts
@@ -1,0 +1,54 @@
+// FIXME: Update this file to be type safe and remove this and next line
+// @ts-strict-ignore
+import { mock } from "jest-mock-extended";
+import { of } from "rxjs";
+
+import { AccountService } from "@bitwarden/common/auth/abstractions/account.service";
+import { TokenService } from "@bitwarden/common/auth/abstractions/token.service";
+import { PasswordGenerationServiceAbstraction } from "@bitwarden/generator-legacy";
+
+import { GenerateCommand } from "./generate.command";
+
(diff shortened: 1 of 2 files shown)


apps/cli/src/tools/generate.command.spec.ts
```diff
@@ -0,0 +1,54 @@
+// FIXME: Update this file to be type safe and remove this and next line
+// @ts-strict-ignore
```
harr1424: For new files we try to avoid doing this, is it even necessary here? 

For a spec file I won't consider it a blocker, but if it can be removed, it would be preferred.

jaasen-livefront: @harr1424 Totally agree. Removed https://github.com/bitwarden/clients/pull/21290/commits/6d1ac8ba4fa088d308b5e969d3893106ae3f7769

harr1424 (approved):

PR #21291, opened 2026-06-16
[PM-34023] Remove 2FA account recovery flag

eliykat:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-34023
## 📔 Objective

Remove the 2FA account recovery flag and associated cleanup.

- Delete feature flag and all conditional logic
- Move types that we still use out of the v1 dialog and into the v2 dialog
- Delete v1 dialog (not renamed here to preserve git history)
- Delete unused and obsolete api endpoint

- **Platform:** you used one of our feature flags in your ConfigService tests. I have updated it to use one of yours.
- **Auth:** update to code comment only to refer to the new method.

Server PR: https://github.com/bitwarden/server/pull/7815


djsmith85 (approved):
Approving for platform owned code.

:thought_balloon: It does seem odd that these [tests](https://github.com/bitwarden/clients/pull/21291/changes#diff-4b0e04f2ad4fcc76baf354df2b7daab827093e0dfaaba81a4decd7c9cb61e67aR398) rely on existing feature flags. Maybe we could assign a test list of feature flags instead. Then this file wouldn't need to change at all.

JaredSnider-Bitwarden (approved):
Auth file change is just a comment tweak - LGTM!

libs/common/src/platform/services/config/config.service.spec.ts
```diff
@@ -395,9 +395,9 @@ describe("ConfigService", () => {
     });
 
     it("returns the override value when one is present", async () => {
-      overrideState.stateSubject.next({ [FeatureFlag.AdminResetTwoFactor]: true });
+      overrideState.stateSubject.next({ [FeatureFlag.FedRampGovRegion]: true });
```
JimmyVo16, line 398: Since this is test code, I’m guessing this is just a placeholder?

JimmyVo16 (approved):

libs/common/src/platform/services/config/config.service.spec.ts
eliykat: Yes - see @djsmith85 's comment above. These should really use mock values, but I didn't want to figure that out here. I just changed it to a Platform-owned flag so that they can manage it.

PR #21294, opened 2026-06-16
[PM-39131] [Shared Unlock] Safari - Desktop IPC Implementation

quexten:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-39131

## 📔 Objective

Implements IPC transport for Safari. This enables biometric unlock via the desktop app, and shared unlock for Safari.

Safari extensions cannot hold a persistent connection to the desktop, so the desktop hosts a buffered IPC server over the shared App Group socket (native Rust implementation in `desktop_native`, exposed via NAPI). The extension background sends messages through the Safari native handler and polls the buffered socket to receive desktop-initiated messages.

Please note, eventually we should switch this to XPC, which should be a reasonably sized lift. However, I've moved that out of scope for this PR because:
- We should migrate it together with `desktop_proxy`
- This approach was easier to implement and easier to debug
- The complexity introduced here is limited to two files and can be easily replaced
  - It is reasonably easy to replace with XPC later on. MAS is distributed together with Safari, so we don't even need to do any compatibility checking. We can just have one PR replacing (parts of) the swift module, the one rust file. The rest of the code stays the same.
- (My signing setup didn't work and it would be a scope increase to set up proper signing)
- (There is a prototype for XPC here https://github.com/bitwarden/clients/pull/21297/changes, but I don't see a need to shepherd it through here. Please feel free to take it over)

### DIFF
--- a/apps/browser/src/platform/ipc/transports/index.ts
+++ b/apps/browser/src/platform/ipc/transports/index.ts
@@ -1,0 +1,3 @@
+export { WebIpcTransport } from "./web";
+export { SafariTransport } from "./safari";
+export { NativeMessagingTransport } from "./native-messaging";
--- a/apps/browser/src/platform/ipc/transports/safari.ts
+++ b/apps/browser/src/platform/ipc/transports/safari.ts
@@ -1,0 +1,115 @@
+import { LogService } from "@bitwarden/common/platform/abstractions/log.service";
+import { IpcCommunicationBackend, IncomingMessage, OutgoingMessage } from "@bitwarden/sdk-internal";
+
+import { SafariApp } from "../../../browser/safariApp";
+
+/** Poll interval while the channel is active (a recent send, or messages just arrived). */
+const SAFARI_FAST_POLL_MS = 200;
+/** Idle heartbeat so desktop-initiated pushes (e.g. shared unlock) arrive promptly. */
+const SAFARI_IDLE_POLL_MS = 1000;
+/** How long after activity to keep polling fast. */
+const SAFARI_ACTIVE_WINDOW_MS = 5000;
+
+/** Envelope serialized over the buffered Safari socket. Payloads are opaque (SDK-encrypted). */
+interface SafariIpcEnvelope {
+  destination: unknown;
+  source?: unknown;
+  payload: number[];
+  topic?: string;
+}
+
+/**
+ * Transport for communicating with the desktop app on Safari.
+ *
+ * Safari cannot maintain a persistent native messaging connection to the desktop, so messages are
+ * sent through the Safari Swift module and desktop -> extension messages are received by polling a
+ * buffered socket.
+ */
+export class SafariTransport {
+  private pollTimeout?: ReturnType<typeof setTimeout>;
+  private lastActivity = 0;
+
+  constructor(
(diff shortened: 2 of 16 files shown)


apps/browser/src/platform/ipc/transports/index.ts
```diff
@@ -0,0 +1,3 @@
+export { WebIpcTransport } from "./web";
```
quexten, line 1: Note to platform: This was getting hard to read / reason about so I split these apart into their own files.

apps/browser/src/platform/ipc/transports/safari.ts
```diff
@@ -0,0 +1,115 @@
+import { LogService } from "@bitwarden/common/platform/abstractions/log.service";
+import { IpcCommunicationBackend, IncomingMessage, OutgoingMessage } from "@bitwarden/sdk-internal";
+
+import { SafariApp } from "../../../browser/safariApp";
+
+/** Poll interval while the channel is active (a recent send, or messages just arrived). */
+const SAFARI_FAST_POLL_MS = 200;
```
quexten, line 7: As far as I can see it we have three states:
- Disconnected entirely (desktop app closed)
  - With default auto-start should not happen much hopefully
- Idle (most of the time)
- Active

We have RPC calls that may timeout, so we cannot wait longer than the timeout *if* desktop initiates the RPC. Active is there to reduce latency in case of things that require a few messages (cryptographic handshake), as to not take multiple seconds.

apps/desktop/desktop_native/core/src/ipc/safari_ipc_server.rs
```diff
@@ -0,0 +1,381 @@
...
+use std::{
+    collections::VecDeque,
+    error::Error,
+    path::PathBuf,
+    sync::{Arc, Mutex},
+    time::{Duration, Instant},
+};
+
+use anyhow::Result;
+use futures::{SinkExt, StreamExt, TryFutureExt};
```
quexten, line 22: Platform question: Should we restrict listening to:
- Electron dev build
- MAS store build

I don't want to create a scenario in which you switch to XPC and suddenly things are unsupported. I'm not sure if XPC is supported if we sign the DMG properly.

apps/browser/src/safari/safari/SafariWebExtensionHandler.swift
```diff
@@ -407,6 +446,114 @@ func jsonDeserialize<T: Decodable>(json: String?) -> T? {
...
+        }
+        defer { close(fd) }
+
+        var addr = sockaddr_un()
+        addr.sun_family = sa_family_t(AF_UNIX)
+        let pathBytes = path.utf8CString
+        // sun_path is a fixed-size buffer (104 bytes on Darwin); the path must fit including its
+        // null terminator.
+        if pathBytes.count > MemoryLayout.size(ofValue: addr.sun_path) {
+            os_log(.error, "[BufferedSocketClient] Socket path too long (%d bytes, max %d)", pathBytes.count, MemoryLayout.size(ofValue: addr.sun_path))
```
quexten, line 493: We had issues with long user names. They will be the same here, just noting.

PR #21301, opened 2026-06-16
[CL-1223] simplify banner component

BryanCunningham:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/CL-1223

## 📔 Objective

Update banner alignment so that icon and close button are not center aligned


libs/components/src/banner/banner.component.html
```diff
@@ -1,64 +1,107 @@
...
+            [variant]="variant()"
+            size="sm"
+            [class]="showActions() ? '@3xl:tw-hidden' : ''"
+            aria-hidden="true"
+          />
+          <bit-icon-tile
+            [icon]="icon"
+            [variant]="variant()"
+            size="lg"
+            class="tw-hidden @3xl:tw-block"
```
vleague2, line 31: Does this one also need a `showActions` check?

BryanCunningham: No. That top one doesn't actually need it either honestly. `showActions` is just a computed signal that returns true/false based on if the title is present

BryanCunningham: I'll take another look here and see if I can clean this up some more

BryanCunningham: @vleague2 removed `showActions` entirely

vleague2 (approved):

PR #21311, opened 2026-06-16
[PM-39070] Create new permissions screen

dan-livefront:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-39070
## 📔 Objective

Create the new permissions screen for the conflicting settings work. I utilized existing patterns in the login carousel in order to achieve a similar feel since this screen will come before it. 
This screen is not wired and is just UI for now 

> [!NOTE]  
> This is a dead code path for now that will be wired in PM-39071

> [!IMPORTANT]  
> `* eslint-disable @bitwarden/components/require-theme-colors-in-svg */` was added to svg file because the gradient does not map exactly to theme colors. This approach is required or we will need to change assets. 
## 📸 Screenshots

Screenshot 2026-06-16 at 2 09 53 PM
Screenshot 2026-06-16 at 2 10 20 PM


apps/browser/src/autofill/popup/default-password-manager/default-password-manager-illustrations.ts
jprusik, line 1: non-blocking nit: run these SVGs through SVGO?

jprusik (dismissed):

apps/browser/src/autofill/popup/default-password-manager/default-password-manager-illustrations.ts
dan-livefront: Yep, these are the optimized ones 😅 when I tweaked the settings on float precision, the image quality went down a bit


Edit: was able to play with the settings a bit and reduce them further

jprusik (approved):

PR #21326, opened 2026-06-17
[PM-39182] Migrate to relative imports - @bitwarden/team-tools-dev

quexten:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-39182

## 📔 Objective

Imports within the same package must be relative imports. This is documented here: https://contributing.bitwarden.com/contributing/code-style/web/typescript#imports-within-the-same-package . To enforce this requirement, an eslint rule has been added in https://github.com/bitwarden/clients/pull/21198 . This PR set is, per team, introducing relative self imports. These are for the most part auto-generated by the eslint rule, with some manual tweaks for e.g. import order issues after the autofix was applied.

### DIFF
--- a/libs/angular/src/tools/password-strength/password-strength-v2.component.ts
+++ b/libs/angular/src/tools/password-strength/password-strength-v2.component.ts
@@ -11,10 +11,11 @@ import {
   SimpleChanges,
 } from "@angular/core";
 
-import { JslibModule } from "@bitwarden/angular/jslib.module";
 import { I18nService } from "@bitwarden/common/platform/abstractions/i18n.service";
 import { PasswordStrengthServiceAbstraction } from "@bitwarden/common/tools/password-strength";
 import { ProgressBarComponent, FormFieldModule } from "@bitwarden/components";
+
+import { JslibModule } from "../../jslib.module";
 
 export interface PasswordColorText {
   color: BackgroundTypes;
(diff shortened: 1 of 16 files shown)


libs/angular/src/tools/password-strength/password-strength-v2.component.ts
```diff
@@ -11,11 +11,12 @@ import {
   SimpleChanges,
 } from "@angular/core";
 
-import { JslibModule } from "@bitwarden/angular/jslib.module";
 import { I18nService } from "@bitwarden/common/platform/abstractions/i18n.service";
 import { PasswordStrengthServiceAbstraction } from "@bitwarden/common/tools/password-strength";
 import { ProgressBarComponent, FormFieldModule } from "@bitwarden/components";
 
+import { JslibModule } from "../../jslib.module";
```
harr1424, line 18: We try and replace this deprecated library with a non-deprecated alternative when we edit these files, but I can appreciate this may be out of scope for this work item.

harr1424 (approved):

libs/angular/src/tools/password-strength/password-strength-v2.component.ts
quexten: Sorry, yeah fully automated changes with one manual review pass by me, I did not look at each change individually and look for alternatives or refactors. It was over 220 files >.<

PR #21331, opened 2026-06-17
[PM-39182] Migrate to relative imports - @bitwarden/team-vault-dev

quexten:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-39182

## 📔 Objective

Imports within the same package must be relative imports. This is documented here: https://contributing.bitwarden.com/contributing/code-style/web/typescript#imports-within-the-same-package . To enforce this requirement, an eslint rule has been added in https://github.com/bitwarden/clients/pull/21198 . This PR set is, per team, introducing relative self imports. These are for the most part auto-generated by the eslint rule, with some manual tweaks for e.g. import order issues after the autofix was applied.


gbubemismith (approved):
Thanks for working on this @quexten

PR #21345, opened 2026-06-17
[PM-37807] Implement `VaultBatchBarComponent` on organization vault

gbubemismith:
## 🎟️ Tracking
https://bitwarden.atlassian.net/browse/PM-37807

## 📔 Objective
Enable admin console users to perform bulk collection operations through the floating batch bar.

## 📸 Screenshots

https://github.com/user-attachments/assets/378993a1-830c-426b-9963-dbd8c9fdcb82

### Updated dialogs

https://github.com/user-attachments/assets/0e52d689-81db-48c9-a71b-dacf0e4deeb2


apps/web/src/app/admin-console/organizations/collections/vault-v2.component.ts
```diff
@@ -372,6 +407,19 @@ export class VaultV2Component implements OnInit, OnDestroy {
...
+    combineLatest([this.organization$, this.allCollections$, this.ciphers$])
+      .pipe(takeUntil(this.destroy$))
+      .subscribe(([organization, allCollections, ciphers]) => {
+        this.vaultBatchBarService.setConfig({
+          isOrgVault: true,
+          organization,
+          allCollections,
+          hasCiphers: ciphers.length > 0,
+        });
+      });
```
BTreston, line 422: Why do we need `isOrgVault`? would organization being `Organization | undefined` not satisfy that? Also it feels a bit backward to be managing the service state from the component. I think it would be clearer to derive the config internally in `VaultBatchBarService` from our existing services (`OrganizationService` and `VaultCollectionService`). I understand that `ciphers$` here is still managed internally by our component and not a dedicated service which, complicates this. But a lift and shift of the ciphers logic might just resolve that. Any thoughts on that? We should really strive for truly presentational components, and all the business logic should live in the services.

BTreston (commented):
2 questions about the interface for the config of this batch bar service

apps/web/src/app/admin-console/organizations/collections/vault-v2.component.ts
gbubemismith: `isOrgVault` feeds `cipherAuthorizationService.canDeleteCipher$/canRestoreCipher$` which checks if an action is performed from admin console, it gates the elevated admin permission paths. That's why I'd keep it explicit rather than infer it from organization != null; on the individual vault, we always go through the member path, even for ciphers that belong to an org. So presence of an organization isn't the same question as `is this an admin-console action?`. I'll prefer we keep this separate

BTreston (dismissed):
LGTM  🌱  The component managing the service's state is not ideal, but non blocking. We can revisit this tech-debt on the AC side in the future as we continue to clean up our org vault component.

BTreston (dismissed):

nick-livefront (dismissed):

nick-livefront (approved):

BTreston (approved):

BryanCunningham (approved):

PR #21349, opened 2026-06-17
[PM-27844] Set email on sync

ike-kottlowski:
## 🎟️ Tracking

[PM-27844](https://bitwarden.atlassian.net/browse/PM-27844)

### Primary: Update the cached account email on sync (feature flagged)

When a user changes their email, the local account state could become stale until the next full re-login. This PR makes the client update the cached account email as part of the normal sync flow, so the new email is reflected immediately across the UI.

- Adds the `PM30806_SelfServiceChangeEmailCommand` (`pm-30806-self-service-change-email-command`) feature flag, defaulting to `FALSE`.
- `DefaultSyncService` now takes a `ConfigService` and, when the flag is enabled, calls `accountService.setAccountEmail(...)` with the email returned in the sync response. Wired up in `main.background.ts`, the CLI `service-container`, and `jslib-services.module.ts`.
- `ChangeEmailComponent` no longer forces a logout when the flag is enabled. Instead, on a successful email change it sets the new account email directly and shows a simpler success toast. The legacy "log back in" + `logout` behavior is preserved when the flag is off.

All new behavior is gated behind the flag, so it is functional but off by default.

### Secondary: Modernize the touched components

While working in this area, the components that were touched were modernized to current Bitwarden Angular conventions (standalone, `OnPush`, signal inputs, `@if` control flow), clearing their `CL-764`/`CL-903` FIXMEs and `@ts-strict-ignore` headers. Two of these modernizations also fixed real display bugs:

- **`ChangeEmailComponent`** — migrated to `OnPush` change detection and signals, and renamed the multi-step form controls/methods to describe their intent (`userVerificationAndNewEmail`, `emailOwnershipVerification`, etc.) instead of generic `step1`/`token` naming.
- **`dynamic-avatar.component`** — migrated `@Input()`s to signal `input()`s and `OnPush`. This fixed a stale-display bug: the email shown in the avatar `title` would not update reactively, so it showed an outdated value until the page was refreshed. With signal inputs the title now reflects the current email.
- **`account-fingerprint.component`** — migrated to signal inputs, a `signal` for the generated fingerprint, `OnPush`, and `@if` control flow. This fixed a display bug introduced when the parent component was modernized: the fingerprint would not render unless the page was refreshed. Making the fingerprint reactive resolved the stale display.
- **`profile.component`** — linked to the email signal so it updates on server success, addressed an outstanding FIXME, and fixed a `bwi` icon warning by moving the button's icon from a nested `<bit-icon name="bwi-pencil-square">` to the `startIcon` attribute.

## 📸 Screenshots

https://github.com/user-attachments/assets/3284eac7-142a-417f-8c3f-a2792dd4a6d1

A video shows a user logging in and navigating to account settings. User then proceeds to initiate email change by inputting a password in the new email. User gets email OTP and inserts it into the corresponding input box. User saves the new email successfully. Visual components that showed the old email are updated to reflect the new email.

User changes the dynamic avatar color and saves with the success.

The user also types in the name input box displaying the reactive nature of the dynamic avatar. The user saves a name change and shows that the dynamic avatar dates based on the name change.

[PM-27844]: https://bitwarden.atlassian.net/browse/PM-27844?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mzieniukbw (dismissed):

libs/common/src/platform/sync/default-sync.service.ts
```diff
@@ -277,6 +280,12 @@ export class DefaultSyncService extends CoreSyncService {
     await this.accountService.setAccountCreationDate(response.id, new Date(response.creationDate));
     await this.accountService.setAccountVerifyNewDeviceLogin(response.id, response.verifyDevices);
 
+    if (
+      await this.configService.getFeatureFlag(FeatureFlag.PM30806_SelfServiceChangeEmailCommand)
+    ) {
+      await this.accountService.setAccountEmail(response.id, response.email);
+    }
```
JaredSnider-Bitwarden, line 287: ❓ : doesn't this missing set mean that if a user updates their email on web, other clients don't update on sync - only on logout and login? If so... it might literally be worth either (1) merging in that as a bugfix PR to main separately so we can get the benefits of this fix in between now and when we turn this flag on or (2) just remove the feature flag check + test on feature?

apps/web/src/app/auth/settings/account/change-email.component.ts
```diff
@@ -1,123 +1,135 @@
...
+  protected readonly selfServiceChangeEmailEnabled: Signal<boolean>;
+
+  readonly formGroup = this.formBuilder.group({
+    userVerificationAndNewEmail: this.formBuilder.group({
       masterPassword: ["", [Validators.required]],
       newEmail: ["", [Validators.required, Validators.email]],
     }),
-    token: [{ value: "", disabled: true }, [Validators.required]],
+    emailOwnershipVerification: [{ value: "", disabled: true }, [Validators.required]],
   });
```
JaredSnider-Bitwarden, line 40: 🥇 these renames are really nice. Thank you!

apps/web/src/app/auth/settings/account/change-email.component.ts
```diff
@@ -1,123 +1,135 @@
...
-  activateStep2() {
-    this.formGroup.controls.step1.disable();
-    this.formGroup.controls.token.enable();
+  disableUserVerificationAndNewEmailForm() {
+    this.formGroup.controls.userVerificationAndNewEmail.disable();
+    this.formGroup.controls.emailOwnershipVerification.enable();
 
-    this.tokenSent = true;
+    this.userVerificationSuccessful.set(true);
   }
```
JaredSnider-Bitwarden: 🎨 : this method does more than just what the method name suggests. 
What about `advanceToEmailOwnershipVerification` so it's suggestive of the second stage of what the user has to do?

JaredSnider-Bitwarden (changes requested):
Overall, this is looking great! 2 points to discuss below:

libs/common/src/platform/sync/default-sync.service.ts
ike-kottlowski: The way email change works today is it changes the cryptographic mechanisms. So, an email change forces a log out everywhere since it modifies the master password hash which can/did change the keys. There is a warning when changing your email today that states you will be logged out but other sessions may work for up to an hour, implying they will be logged out.

I did a quick tests, and this is the case on us-dev. I changed the email on web and was logged out shortly after on browser.

JaredSnider-Bitwarden: Oh, right. Thank you for checking. I completely forgot about the pushed logout.

apps/web/src/app/auth/settings/account/change-email.component.ts
ike-kottlowski: method name changed. https://github.com/bitwarden/clients/pull/21349/commits/ae1d4e8afce3d93f6876514c289ef42e2c393790

JaredSnider-Bitwarden (approved):
Excellent work, and thank you for tackling the modernization efforts as well!

quexten (approved):
Re-approving given @mzieniukbw already approved it I'm not reviewing the PR, given the only difference since the last approval is a merge of main.

addisonbeck (approved):

PR #21351, opened 2026-06-17
[PM-39200] - fix notification bar behavior with save and fill

jaasen-livefront:
## 🎟️ Tracking

[PM-39200](https://bitwarden.atlassian.net/browse/PM-39200)

## 📔 Objective

Fixes the post-save behavior in the browser add/edit vault item flow when the Save and Fill feature (`PM29968_FillAfterSave`) is disabled.

Previously, `AddEditComponent.onCipherSaved()` always sent the `showLoginSavedNotification` message and closed the single-action popout immediately, regardless of the feature flag. This surfaced the save-and-fill notification flow in contexts where the feature is off and removed the brief delay that gives users visual confirmation of the save.

Changes:
- Gate the saved-notification flow behind the feature flag: only call `showLoginSavedNotification` and close the popout immediately when `saveAndFillEnabled` is true; otherwise close the popout with a 1000ms delay and skip the notification.
- Consolidated `showSaveAndFill` and the locally-scoped `saveAndFillEnabled` into a single `saveAndFillEnabled` field so the template (Save and Fill button visibility) and the save flow are driven by the same source of truth.
- Updated the `onCipherSaved` unit tests to cover both the enabled and disabled paths.

[PM-39200]: https://bitwarden.atlassian.net/browse/PM-39200?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/browser/src/vault/popup/components/vault/add-edit/add-edit.component.html
+++ b/apps/browser/src/vault/popup/components/vault/add-edit/add-edit.component.html
@@ -31,7 +31,7 @@
   }
 
   <popup-footer slot="footer">
-    @if (showSaveAndFill) {
+    @if (saveAndFillEnabled) {
       <button [bitAction]="submitAndFill" type="button" bitButton buttonType="primary">
         {{ "saveAndFill" | i18n }}
       </button>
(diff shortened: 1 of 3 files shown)


jengstrom-bw (approved):
LGTM

PR #21370, opened 2026-06-18
Fix/late hydration shadow detection

### LINKED ISSUE #21453: High CPU use when a page has a lot of input elements
### Steps To Reproduce

Open a PR with a lot of files changed (~75 but sometimes even as little as 28).

### Expected Result

Normal interaction with the PR

### Actual Result

The browser extension causes the web tab to frequently spike to high CPU usage (70 up to 100%).

### Screenshots or Videos

Image

### Additional Context

I first thought it was a github issue, but when profiling the page it became clear that it was a specific querySelectorAll call from the bitwarden extension that is frequently fired, seemingly without any real correlation to user input (i.e. it triggers without clicking on anything, just scrolling and often even not doing anything will spike the cpu while on the tab).

Disabling the extension removed all cpu spikes.

I was able to replicate this on both Vivaldi and Chrome.

### Operating System

Windows

### Operating System Version

_No response_

### Web Browser

Vivaldi

### Browser Version

_No response_

### Environment Versions

Version: 2026.5.1

SDK: 'main (e13c4fc)'

Server version: 2026.6.1

### Issue Tracking Info

blackwood:
## 🎟️ Tracking

[PM-38775](https://bitwarden.atlassian.net/browse/PM-38775)

# PR 3B — fix/late-hydration-shadow-detection (incremental shadow overflow + late-hydration detection)

PR 1: #21083
PR 2: #21084
~~PR 3: #21085~~ (obviates)

**Net objective**

Restore detection of fields in late-hydrating shadow DOM, and keep shadow-candidate discovery off the `O(document)` path — on both the passive scan and the fill path — without the detached-root memory growth the old full-rescan caused. All in `dom-query.service.ts` / `collect-autofill-content.service.ts`; the pieces are paired — cheap incremental discovery is what lets late-hydration tracking re-scan held candidates without a whole-document walk.

**What changed (relative to main)**

- **Late-hydration detection.** `buildTreeWalkerNodesQueryResults` sinks hyphenated shadow-less elements as unresolved candidates; `trackUnresolvedShadowHosts` holds them so a later scan catches the root `attachShadow()` adds silently — the fields the old code missed.
- **Enroll-at-scan.** `checkForNewShadowRoots` takes the `MutationObserver` and enrolls discovered roots (`enrollShadowRoot`) during the cheap batch scan, not only on the whole-document `getPageDetails` walk.
- **Overflow stays cheap.** Overflow re-scans the held candidate batch instead of re-walking the document, and no longer retains detached roots — closing the memory growth.
- **Fill path stays on the cache.** `prepareForExplicitCollection` refreshes the latch only on `noFieldsFound || unresolvedShadowHosts.size > 0`, not `hostsAwaitingDefinition`; where framework pages park every unregistered selector (`<app-root>`), which otherwise forced an `O(document)` scan on the awaited `collectPageDetails` in `fillInlineMenuCipher`. Late hydration still gates: a parked host that flips `:defined` promotes into `unresolvedShadowHosts`.
- **Finding 1 — own UI excluded.** `isOwnedShadowHost` guards the sink and observation, so the extension no longer scans its own inline-menu roots (by identity, not tag — covers Firefox's `<div>` host).
- **Finding 2 — parked hosts bounded.** `parkHostAwaitingDefinition` stamps a 60s deadline; `enrollUpgradedParkedHosts` tombstones past it, so a never-registered element expires instead of arming the ≤8s retry forever.

## incremental shadow-candidate overflow

- **Goal:** When the added-node batch exceeds the cap, scan the kept candidates incrementally
  instead of re-walking the whole document for shadow roots.
- **Expected:** `handleNewShadowRoots` cost collapses on scale/overflow scenarios; no detached-root retention.
- **Result:** S10 `handleNewShadowRoots` total **11,767 ms → 0.5 ms** (full-document re-walk eliminated).
- **Tradeoff:** Discovery is now batch-scoped — a root reachable only outside the batch is caught on
  the next mutation cycle rather than immediately.

## late-hydration shadow detection

- **Goal:** Detect form fields in custom-element hosts whose shadow root attaches *after* the host is
  scanned (lazy hydration emits no observable mutation).
- **Expected:** Late-attached shadow forms become fillable; tracking is bounded (deadline + cap +
  `whenDefined` parking, and parked hosts now carry their own deadline) so it can't leak — neither memory nor the retry timer.
- **Result:** S3 (`late-attached-shadow-form`) coverage restored. 
- **Tradeoff:** A held host re-scans until it hydrates or its deadline expires; user-initiated
  collection bypasses the no-fields cache while hosts are unresolved.

---
Review follow-ups: shadow roots are now enrolled (observed) during the cheap batch scan, not only the later collection walk, so observation rides the incremental path. Two leaks were closed: the unresolved-host sinks now exclude the extension's own owned inline-menu elements, and parked (awaiting-definition) hosts expire after 60 s so a tag that never registers can't keep the retry timer alive. Deeper cleanup — folding the tracking into a ShadowHostHydrationTracker and revisiting the whenDefined hook — is tracked as follow-up tech debt ([PM-40947](https://bitwarden.atlassian.net/browse/PM-40947)).

## Coverage

255 autofill specs pass; held shadow coverage (S3/S9/S10) and memory discipline.

<details>
<summary><b>Benchmark scenario legend</b></summary>

| Label | Fixture | What it exercises |
|-------|---------|-------------------|
| **S3** | `3-late-attached-shadow-form` | A shadow-hosted form attached *after* load — late discovery / coverage. |
| **S9** | `9-shadow-tile-churn` | Field-less shadow roots doing deep childList churn (map-tile pattern). |
| **S10** | `10-shadow-scale-stress` | ~90k nodes: field-less tiles + nested-root chain + sparse forms (scale, overflow, nesting). |

</details>

[PM-38775]: https://bitwarden.atlassian.net/browse/PM-38775?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[PM-40947]: https://bitwarden.atlassian.net/browse/PM-40947?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/browser/src/autofill/services/collect-autofill-content.service.ts
audreyality, line 1: 🤔 It seems like there's a lot of low-level logic threaded through the collect autofill content service. The large number of configuration settings and interleaving of scheduling logic (e.g. `setTimeout`) with decision logic (tracking and enrollment) leads me to believe there's a missing abstraction.

apps/browser/src/autofill/services/collect-autofill-content.service.ts
```diff
@@ -81,6 +105,29 @@ export class CollectAutofillContentService implements CollectAutofillContentServ
...
+  private readonly monitorUpdateCachedVisibility = createMeter("updateCachedVisibility");
+  private readonly monitorSetupOverlayOnField = createMeter("setupOverlayOnField");
+  // Gate-branch-architecture meters (main's queue-swap path): adaptive backoff + 256-cap overflow.
+  private readonly monitorBackoff = createMeter("mutationBackoff", "burst", "adaptiveMs");
+  private readonly monitorCandidateOverflow = createMeter("shadowCandidateOverflow");
+  private readonly monitorUnresolvedShadowHosts = createMeter(
+    "unresolvedShadowHosts",
+    "tracked",
+    "retryRound",
+  );
```
audreyality: 💭 This is a large number of meters to add in a changeset focused on optimization. Which are key health indicators and which are leftovers from your experiments?

apps/browser/src/autofill/services/collect-autofill-content.service.ts
```diff
@@ -1512,6 +1603,13 @@ export class CollectAutofillContentService implements CollectAutofillContentServ
         if (this.domRecentlyMutated) {
           this.updateAutofillElementsAfterMutation();
         }
+        this.monitorMapSizes(
+          this.autofillFieldElements.size,
+          this.autofillFieldsByOpid.size,
+          this._autofillFormElements.size,
+          (performance as unknown as { memory?: { usedJSHeapSize: number } }).memory
+            ?.usedJSHeapSize ?? 0,
```
audreyality: ⚠️ Focus your monitoring on indicators intrinsic to your logic. We should measure holistic indicators, such as heap size, using higher-level data collection.

apps/browser/src/autofill/services/collect-autofill-content.service.ts
```diff
@@ -1579,22 +1677,228 @@ export class CollectAutofillContentService implements CollectAutofillContentServ
...
+        batch.push(element);
+      }
+    }
+
+    const unresolvedHostSink = new Set<Element>();
+    const hasNewShadowRoots = measure("shadowRootScan", () =>
+      this.domQueryService.checkForNewShadowRoots(batch, unresolvedHostSink),
+    );
     if (hasNewShadowRoots) {
+      this.monitorRequireUpdateNewShadowRoot();
```
audreyality: ⚠️ In order to properly contextualize `monitorRequireUpdateNewShadowRoot()`, you have to know the broader environment in which it's embedded. It's not a context-free measurement, and thus is not a meter. It's an event. Event monitors are out of scope of the metric collection interfaces.

🦖 I have an event monitoring interface planned.

apps/browser/src/autofill/services/collect-autofill-content.service.ts
```diff
@@ -1579,22 +1677,228 @@ export class CollectAutofillContentService implements CollectAutofillContentServ
...
+        }
+        if (!element.matches(":defined")) {
+          continue;
+        }
+        hosts.delete(element);
+        enrolled = true;
+        if (this.unresolvedShadowHosts.size < this.unresolvedShadowHostTrackingCap) {
+          this.unresolvedShadowHosts.set(
+            element,
+            performance.now() + this.unresolvedShadowHostLifetimeMs,
```
audreyality: 🤔 This line is suspect to me. It's not clear why you would need a high-resolution timestamp to track the age of a shadow host.

apps/browser/src/autofill/services/collect-autofill-content.service.ts
```diff
@@ -1579,22 +1677,228 @@ export class CollectAutofillContentService implements CollectAutofillContentServ
...
+  // where no such event exists.
+  private hookCustomElementDefinition(tagName: string) {
+    if (this.hookedCustomElementTags.has(tagName) || !globalThis.customElements) {
+      return;
+    }
+    this.hookedCustomElementTags.add(tagName);
+    globalThis.customElements.whenDefined(tagName).then(
+      () => this.enrollDefinedHosts(tagName),
+      () => {},
+    );
```
audreyality: ⚠️ This drops an error.

apps/browser/src/autofill/services/dom-query.service.ts
```diff
@@ -388,35 +409,56 @@ export class DomQueryService implements DomQueryServiceInterface {
...
-      if (root) {
-        if (!this.knownShadowRoots.has(root)) {
-          return true;
-        }
-        if (this.scanForNewShadowRootInSubtree(root, depth + 1)) {
+      if (this.visitShadowHostCandidate(subtree, depth, unresolvedHostSink)) {
+        if (!unresolvedHostSink) {
           return true;
         }
+        found = true;
```
audreyality: ⚠️ Mutating `unresolvedHostSink` within the body of this method is an antipattern. It turns a pure function into an effectful method. Instead, gather unresolved host sinks using a separate call chain. This callchain can omit `unresolvedHostSink`, and instead return the full set of unresolved host sinks. This callchain will be easier to read and, in all likelhihood, will have more predictable runtime behavior.

apps/browser/src/autofill/services/dom-query.service.ts
```diff
@@ -596,7 +641,17 @@ export class DomQueryService implements DomQueryServiceInterface {
...
           );
+        } else if (
+          unresolvedHostSink &&
+          unresolvedHostSink.size < MAX_UNRESOLVED_SHADOW_HOSTS &&
+          el.tagName.includes("-") &&
+          !el.shadowRoot
+        ) {
+          // Hosts that predate observer attachment never appear as mutation
+          // candidates; the collection walk is their only enrollment source.
+          unresolvedHostSink.add(el);
```
audreyality: 📝 See above.

audreyality (changes requested):

apps/browser/src/autofill/services/collect-autofill-content.service.ts
blackwood: Removed all experimental meters in `Remove meter/bench instrumentation.` 01ab84f

apps/browser/src/autofill/services/collect-autofill-content.service.ts
blackwood: Handled in `Remove meter/bench instrumentation.` 01ab84f

apps/browser/src/autofill/services/collect-autofill-content.service.ts
blackwood: All high-res timestamps converted to `Date.now` 7ff587a

apps/browser/src/autofill/services/dom-query.service.ts
blackwood: Handled in `Refactor unresoldvedHostSink to return-value object.` ade685a

blackwood: > unresoldved

😑

apps/browser/src/autofill/services/dom-query.service.ts
blackwood: Also ade685a

apps/browser/src/autofill/services/dom-query.service.ts
audreyality: That revision renames `unresolvedHostSink` to `unresolvedHosts` and makes it a required argument. `scanForNewShadowRootInSubtree` is still mutating the set and returning a response to a query.

I think the thing that's bothering me is that it seems like this should be simpler. The scanning code should only perform a scan. The query code should only perform a query. There shouldn't need to be a boolean answering a state about the DOM here, because any action it takes should be observable through one of the sets it alters.

Take the following method, for example:

```ts
  private visitShadowHostCandidate = (
    element: Element,
    depth: number,
    unresolvedHosts: Set<Element>,
  ): boolean => {
    const root = this.getShadowRoot(element);
    if (!root) {
      if (unresolvedHosts.size < MAX_UNRESOLVED_SHADOW_HOSTS && element.tagName.includes("-")) {
        unresolvedHosts.add(element);
      }
      return false;
    }
    const foundNewRoot = !this.knownShadowRoots.has(root);
    // Descend even into a new root: its own un-hydrated hosts belong in the set.
    return this.scanForNewShadowRootInSubtree(root, depth + 1, unresolvedHosts) || foundNewRoot;
  };
```

This is mixing two concerns. One of those is being tracked by `unresolvedHosts` and the other is being tracked by `knownShadowRoots`. The concern tracked in `unresolvedHosts` is "where might I need to re-scan in the future?" and the concern tracked by `knownShadowRoots` is "where have I positively identified a new shadow root?". Both of these are related. It makes sense to track them together. 

There's a stem off of `knownShadowRoots`, however: "Did I find a new shadow root while visiting this list of candidates?" This concern is part of the visitor because it's inherited from `scanForNewShadowRootInSubtree`. Not only is there a hardcoded `return false`, there's also a `foundNewRoot` variable introduced in order to account for the side-effects of `scanForNewShadowRootInSubtree`. ~~That function mutuates both sets, which would invalidate a direct `knownShadowRoots` query after-the-fact.~~ The mutation-and-query problem still exists. It's just been made indirect.

I think there's a reasonable way to bridge this, though, that makes both functions less expensive. Hoist the query out of the scan completely. `scanForNewShadowRootInSubtree` can perform a full exhaustive scan and return `void`, which properly communicates it's executed for its side effects. We'll need to rely on those side effects. One in particular: `scanForNewShadowRootInSubtree` only ever increases the size of `knownShadowRoots`. (This is presently true.)

`findNewShadowRootInBatch` is similarly mixing concerns. A similar transformation can be applied. Then we reach `checkForNewShadowRoots`, which can capture the length of `knownShadowRoots` before it executes the scan and compare it to the length afterwards. If the length increases, it found a new shadow root and mark the shadow DOM present.

EDIT: Talked out of band. Turns out that `scanForNewShadowRootInSubtree` is **not** mutating `knownShadowRoots` and I misread the diff. Since it's performing an exhaustive search, however, I think mutating `knownShadowRoots` might be appropriate. That way, you're not traversing the DOM twice in order to discover new roots.

blackwood:
If not autofill, feel free to ignore if you got pinged for this, `prettier` reformatting drift on `main` merge tripped things up.

apps/browser/src/autofill/services/collect-autofill-content.service.ts
blackwood: Deferring this refactor ^, along with modifications to hot path (`getPageDetails`) that would allow this PR's introduced eager work of observing shadow roots in incremental mutation batches from being run redundantly. 

See tech debt ticket [PM-40947](https://bitwarden.atlassian.net/browse/PM-40947) newly created

apps/browser/src/autofill/services/dom-query.service.ts
```diff
@@ -395,35 +413,52 @@ export class DomQueryService implements DomQueryServiceInterface {
...
+
+  /** Sorts one candidate into the scan's bins, descending through any root to reach nested hosts. */
+  private visitShadowHostCandidate = (element: Element, depth: number, scan: ShadowScan): void => {
+    const root = this.getShadowRoot(element);
+    if (!root) {
+      if (
+        scan.unresolvedHosts.size < MAX_UNRESOLVED_SHADOW_HOSTS &&
+        element.tagName.includes("-")
+      ) {
+        scan.unresolvedHosts.add(element);
```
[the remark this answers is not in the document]
blackwood: [6fb1d0b](https://github.com/bitwarden/clients/pull/21370/commits/6fb1d0b393b40c220cc8ce3c118e5cdbf4172da1)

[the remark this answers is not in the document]
blackwood: [6fb1d0b](https://github.com/bitwarden/clients/pull/21370/commits/6fb1d0b393b40c220cc8ce3c118e5cdbf4172da1)

apps/browser/src/autofill/services/collect-autofill-content.service.ts
```diff
@@ -1668,22 +1715,204 @@ export class CollectAutofillContentService implements CollectAutofillContentServ
...
+            // Clamp the exponent: `<<` is a 32-bit shift, so an unclamped round would
+            // eventually wrap to a tiny delay. 5 already exceeds the cap (500 << 5 = 16s).
+            this.shadowDomCheckTimeoutMs << Math.min(this.unresolvedShadowHostRetryRound, 5),
+            this.unresolvedShadowHostRetryCapMs,
+          );
+    this.unresolvedShadowHostRetryRound++;
+    this.unresolvedShadowHostRetryTimeout = setTimeout(() => {
+      this.unresolvedShadowHostRetryTimeout = null;
+      this.handleNewShadowRoots();
+    }, delay);
```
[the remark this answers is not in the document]
blackwood: [6fb1d0b](https://github.com/bitwarden/clients/pull/21370/commits/6fb1d0b393b40c220cc8ce3c118e5cdbf4172da1)

audreyality (commented):
📝  Looks like there's one new Claude feedback item to address. I'm leaving this unapproved until there's a response to it. 

👍🏻 My feedback is resolved! Thanks Miles!

apps/browser/src/autofill/services/collect-autofill-content.service.ts
```diff
@@ -194,6 +224,20 @@ export class CollectAutofillContentService implements CollectAutofillContentServ
     return this._autofillFormElements;
   }
 
+  // Explicit request: refresh the latch and force a fresh query when the cache is
+  // "no fields" or hosts are unresolved; cached populated results stay served.
+  prepareForExplicitCollection = () => {
```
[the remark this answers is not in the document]
audreyality: Addressed by [86ef1c1](https://github.com/bitwarden/clients/pull/21370/commits/86ef1c1f8f1b78adc2b373b12afc69079fbb8f6e) w/ tech debt added to [PM-40947](https://bitwarden.atlassian.net/browse/PM-40947)

apps/browser/src/autofill/services/collect-autofill-content.service.ts
```diff
@@ -194,6 +224,22 @@ export class CollectAutofillContentService implements CollectAutofillContentServ
...
+    if (
+      this.noFieldsFound ||
+      this.unresolvedShadowHosts.size > 0 ||
+      this.hostsAwaitingDefinition.size > 0
+    ) {
+      this.domQueryService.refreshShadowDomStateForUserRequest();
+      this.noFieldsFound = false;
+      this.domRecentlyMutated = true;
+    }
+  };
```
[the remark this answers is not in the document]
audreyality: Another possibility is to introduce a heuristic that removes common exceptions from the "hostsAwaitingDefinition" list.

apps/browser/src/autofill/services/shadow-host-hydration-tracker.ts
```diff
@@ -0,0 +1,301 @@
...
+   *   found, rather than waiting for the next whole-document walk
+   */
+  constructor(
+    private readonly domQueryService: DomQueryService,
+    private readonly mutationObserver: MutationObserver,
+    private readonly requestPageDetailsUpdate: () => void,
+    private readonly now: () => EpochMs = () => Date.now(),
+  ) {}
+
+  noteAddedNodes(mutations: MutationRecord[]): void {
```
audreyality, line 57: ⛏️ Missing documentation.

apps/browser/src/autofill/services/shadow-host-hydration-tracker.ts
```diff
@@ -0,0 +1,301 @@
...
+    return node.firstElementChild !== null;
+  }
+
+  private clearTimer(field: "retryTimeout" | "scanTimeout"): void {
+    const handle = this[field];
+    if (handle) {
+      clearTimeout(handle as ReturnType<typeof setTimeout>);
+      this[field] = null;
+    }
+  }
```
audreyality, line 300: 🤔 This method is coupled to the structure of the class in a way that disables typechecking, which makes the type assertion necessary. It would be better to just inline this and keep the typecheck.

apps/browser/src/autofill/services/shadow-host-hydration-tracker.ts
```diff
@@ -0,0 +1,301 @@
...
+          continue;
+        }
+        this.pendingMutationAddedElements.add(node);
+        if (this.pendingMutationAddedElements.size >= this.pendingMutationAddedElementsCap) {
+          this.pendingMutationAddedElementsOverflowed = true;
+          // Keep the capped set so the debounced scan works through it incrementally.
+          return;
+        }
+      }
+    }
```
audreyality, line 277: 🤔 The list of mutations isn't changing here. Aren't you reprocessing the mutations each time you run this function? Or, if you aren't reprocessing it, are you losing nodes?

💭 I wonder whether a better strategy might be to use `splice` to hack off a fixed set of mutations to iterate through. You can then process that set, and move parked/pending/awaiting/etc mutations into their own queues. You could even append ones that are ambiguous back onto `mutations`.

apps/browser/src/autofill/services/shadow-host-hydration-tracker.ts
blackwood: 204ab54cd1

Docstring and param added

apps/browser/src/autofill/services/shadow-host-hydration-tracker.ts
blackwood: 204ab54cd1

`clearTimer` inlined at all 4 sites and deleted.

apps/browser/src/autofill/services/shadow-host-hydration-tracker.ts
blackwood: No change here: neither reprocessed nor lost on this path: fresh `MutationRecord[]` per callback with dedupe;  did catch lost nodes in the overflow tail, though (ddf766b22c)

apps/browser/src/autofill/services/collect-autofill-content.service.ts
```diff
@@ -1690,8 +1941,7 @@ export class CollectAutofillContentService implements CollectAutofillContentServ
         this.pendingMutationAddedElements.add(node);
         if (this.pendingMutationAddedElements.size >= this.pendingMutationAddedElementsCap) {
           this.pendingMutationAddedElementsOverflowed = true;
-          // Release element refs immediately; we won't process them this window.
-          this.pendingMutationAddedElements.clear();
+          // Keep the capped set so the debounced check scans it incrementally, not a re-walk.
           return;
         }
```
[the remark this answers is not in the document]
blackwood: ddf766b22c

collection walk requested when window overflowed

blackwood:
two pieces added outside of feedback loop:

be2c07713e : backoff. `sawNewHost` was decided before `admitHost` decided whether the host actually took a tracking slot, so hosts that overflowed past the 64-entry cap counted as new work on every scan and reset the backoff. That pinned the sweep at its 500 ms floor for the page's life on component-heavy pages.

91a2daaa50 : set union. Two loops pushing into an array needed a manual has guard to avoid scanning a host twice. A set union dedupes by construction, so the guard goes away.

apps/browser/src/autofill/services/shadow-host-hydration-tracker.ts
```diff
@@ -0,0 +1,336 @@
...
+    let enrolled = false;
+    for (const [element, parkDeadline] of this.hostsAwaitingDefinition) {
+      if (!element.isConnected) {
+        this.hostsAwaitingDefinition.delete(element);
+        continue;
+      }
+      if (element.matches(":defined")) {
+        this.hostsAwaitingDefinition.delete(element);
+        // Overwrites an earlier `Abandoned`: proof of definition repairs a wrong verdict.
+        this.tagVerdicts.set(element.tagName, TagVerdict.Defined);
```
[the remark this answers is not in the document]
blackwood: https://github.com/bitwarden/clients/pull/21370/commits/9495703f07121298289e21ed003f19c2693154b0

apps/browser/src/autofill/services/shadow-host-hydration-tracker.ts
audreyality: ⚠️ That docstring describes something related to the design of the method, but doesn't explain what the method is doing.

apps/browser/src/autofill/services/shadow-host-hydration-tracker.ts
audreyality: ❓ I'm missing something. Why is it a fresh batch of records?

🤔 It looks like you're "re-enrolling" by performing a full page scan with `requestPageDetailsUpdate`? You already have the records that you didn't process in this method. Why do a full scan when you could instead queue the records in a separate batch?

apps/browser/src/autofill/services/shadow-host-hydration-tracker.ts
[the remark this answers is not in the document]
audreyality: ⛏️ `Cap` looks like it's short for capacity? I'd consider using the full word.

apps/browser/src/autofill/services/shadow-host-hydration-tracker.ts
```diff
@@ -0,0 +1,343 @@
...
+
+    // Hosts added by mutation may have been removed during the scan debounce.
+    const batch = [
+      ...new Set([...this.pendingMutationAddedElements, ...this.hostsAwaitingShadowRoot.keys()]),
+    ].filter((element) => element.isConnected);
+
+    const { foundNewRoot, unresolvedHosts } = this.domQueryService.checkForNewShadowRoots(
+      batch,
+      this.mutationObserver,
+    );
```
audreyality: 🎨 Since the purpose of this code is speed, I'd avoid allocating and copying arrays to perform the set-union.

Here's an alternative that filters using a generator with only a single copy (from the set):

```suggestion
    function* connected(elements: Iterable<Element>) {
      for(const e of elements) {
        if(e.isConnected) {
          yield e;
        }
      }
    }
    
    // Hosts added by mutation may have been removed during the scan debounce.
    const batch = new Set(connected(this.pendingMutationAddedElements));
    for(const k of connected(this.hostsAwaitingShadowRoot.keys())) {
      batch.add(k);
    }

    const { foundNewRoot, unresolvedHosts } = this.domQueryService.checkForNewShadowRoots(
      Array.from(batch),
      this.mutationObserver,
    );
```

🦖 I'm not sure if we can rely on the ES2025 baseline. If we can, you can do it without the generator:
```suggestion
    // Hosts added by mutation may have been removed during the scan debounce.
    const batch = new Set(
      Iterator.concat<Element>(
        this.pendingMutationAddedElements,
        this.hostsAwaitingShadowRoot.keys()
      ).filter((e) => e.isConnected)
    );

    const { foundNewRoot, unresolvedHosts } = this.domQueryService.checkForNewShadowRoots(
      Array.from(batch),
      this.mutationObserver,
    );
```

audreyality (commented):
⛏️ The SDLC gate is failing because you can only have a single `t:` label on a PR.

📝 The 🎨 in this section is just a suggestion. There's a ⛏️ and a ⚠️ in response to prior reviews (one Claude, one mine) that should be fixed.

apps/browser/src/autofill/services/shadow-host-hydration-tracker.ts
audreyality: 📝 This is answered by a commit in response to Claude's feedback, above: https://github.com/bitwarden/clients/commit/ddf766b22c0507baf7af811d8e85bb1c4fb6d5fb.

apps/browser/src/autofill/services/shadow-host-hydration-tracker.ts
blackwood: some lines got swallowed up by a comment trimming routine! restored by a02c719a.

apps/browser/src/autofill/services/shadow-host-hydration-tracker.ts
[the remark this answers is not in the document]
blackwood: using cap as in "limit" here--if that's a better term I can swap it in; this did prompt me to rename `retryCapMs` to `maxRetryDelayMs` though. 14021fe4

audreyality (dismissed):

apps/browser/src/autofill/services/shadow-host-hydration-tracker.ts
blackwood: I'm not certain about our polyfill/transpilation status for those two features, but this version saves on allocations and utilizes the fact that `pendingMutationAddedElements` is already a Set:

b3ecb045

```ts
    for (const element of this.pendingMutationAddedElements) {
      if (element.isConnected) {
        batch.push(element);
      }
    }
    for (const host of this.hostsAwaitingShadowRoot.keys()) {
      if (!this.pendingMutationAddedElements.has(host) && host.isConnected) {
        batch.push(host);
      }
    }
```

Also, ordering `!has(...)` before `isConnected` means an overlapping host skips the DOM read entirely, so `isConnected` is called once per unique element instead of per membership.

Also added a test case to assert a tracked host re-reported through `noteAddedNodes` creates no duplicates.

audreyality: Since `this.pendingMutationAddedElements` is a set and `this.hostAwaitingShadowRoot` is a map (and thus set-like), you can use [union](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set/union) in ES2024.

```ts
const batch = new Array(this.hostsAwaitingShadowRoot.size + this.hostsAwaitingShadowRoot.size);
let i = 0;
const candidates = this.pendingMutationAddedElements
  .union(this.hostsAwaitingShadowRoot)
  .forEach((e) => {
    if(e.isConnected) {
      batch[i++] = e;
    }
  });
```

But there's no need to actually do this. I'm just geeking out over APIs at this point.

audreyality (approved):

blackwood:
Rust lint failures inherited from main.

PR #21390, opened 2026-06-19
[PM-39204] Add extra spacing to vault items when batch bar is visible

nick-livefront:
## 🎟️ Tracking

[PM-39204](https://bitwarden.atlassian.net/browse/PM-39204)

## 📔 Objective

When the vault batch action bar is visible, vault items near the bottom of the list can be obscured by the floating bar. This PR adds dynamic bottom padding to the virtual scroll viewport so items remain accessible while the bar is shown.

**Changes:**

- Moves the `PM37785_VaultBatchBar` feature flag signal from `VaultItemsComponent` into `VaultBatchBarService`
  - Added `barVisible()` signal which is true when the flag is on and at least one item is selected (i.e., the bar is actually rendered)
- Applies `tw-pb-[7.5rem]` (120px) to the scroll viewport when `barVisible()` is true, transitioning back to `tw-pb-8` when the bar is dismissed.

> Note: The designs specify 96px of bottom padding when the batch bar is visible. 120px (`7.5rem`) is used here because it produces the same visual clearance as the designs intend.

## 📸 Screenshots

<video src="https://github.com/user-attachments/assets/759bd092-23c1-453c-b064-64d3bf29b109" />

[PM-39204]: https://bitwarden.atlassian.net/browse/PM-39204?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/vault/src/services/vault-batch-bar.service.ts
+++ b/libs/vault/src/services/vault-batch-bar.service.ts
@@ -18,6 +18,8 @@ import { CollectionView, Unassigned } from "@bitwarden/common/admin-console/mode
 import { Organization } from "@bitwarden/common/admin-console/models/domain/organization";
 import { AccountService } from "@bitwarden/common/auth/abstractions/account.service";
 import { getUserId } from "@bitwarden/common/auth/services/account.service";
+import { FeatureFlag } from "@bitwarden/common/enums/feature-flag.enum";
+import { ConfigService } from "@bitwarden/common/platform/abstractions/config/config.service";
 import { I18nService } from "@bitwarden/common/platform/abstractions/i18n.service";
 import { LogService } from "@bitwarden/common/platform/abstractions/log.service";
 import { uuidAsString } from "@bitwarden/common/platform/abstractions/sdk/sdk.service";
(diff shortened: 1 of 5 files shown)


libs/vault/src/services/vault-batch-bar.service.ts
```diff
@@ -147,6 +150,17 @@ export class VaultBatchBarService<C extends CipherViewLike> {
...
+  private readonly batchBarFlag = toSignal(
+    this.configService.getFeatureFlag$(FeatureFlag.PM37785_VaultBatchBar),
+    { initialValue: false },
+  );
+
+  /** True when the batch bar feature flag is enabled. */
+  readonly enabled = computed(() => this.batchBarFlag());
+
+  /** True when the batch bar is actively visible: feature flag on and at least one item selected. */
+  readonly barVisible = computed(() => this.batchBarFlag() && this.selectedCount() > 0);
```
jaasen-livefront, line 162: ⛏️ Not sure if these comment are necessary. Seems like the code is pretty clear here. Totally up to you if you prefer to leave them in. ;)

jaasen-livefront (approved):
Looks good! Just one small stylistic nit.

libs/vault/src/services/vault-batch-bar.service.ts
nick-livefront: Agreed, I'll leave them for now. They don't hurt!

PR #21394, opened 2026-06-19
[PM-38481] Add Staged Filter for Organization Members

sven-bitwarden:
## 🎟️ Tracking
[PM-38481](https://bitwarden.atlassian.net/browse/PM-38481)

## 📔 Objective
We are introducing Staged as a new OrganizationUserStatusType to support automatically-provisioned users. Following our current UI practice, we're adding a new filter to the organization users members page to represent this new status.

> [!NOTE]
> Bitwarden SDK does not yet support this status. This PR introduces a temporary map between the SDKs OrganizationUserStatusType, and the frontend's. The area affected does not apply to Staged users, so it is a safe bet, but we could alternatively wait until the SDK's updated before merging this PR. 

#### Staged tooltip (hover works I swear)
image

#### Staged disappears when I remove 'em all
image

[PM-38481]: https://bitwarden.atlassian.net/browse/PM-38481?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/common/src/admin-console/services/policy/default-new-policy.service.ts
```diff
@@ -87,11 +90,33 @@ export class DefaultNewPolicyService implements InternalNewPolicyService {
...
+      case OrganizationUserStatusType.Accepted:
+        return SdkOrganizationUserStatusType.Accepted;
+      case OrganizationUserStatusType.Confirmed:
+        return SdkOrganizationUserStatusType.Confirmed;
+      case OrganizationUserStatusType.Revoked:
+        return SdkOrganizationUserStatusType.Revoked;
+      default:
+        return SdkOrganizationUserStatusType.Invited;
+    }
+  }
```
eliykat: We shouldn't adopt workarounds like this; the SDK is part of our stack now. Please update the SDK first and then bump the version inside this PR if that's what's needed.

apps/web/src/app/admin-console/organizations/members/members.component.html
```diff
@@ -44,6 +44,11 @@
           <bit-berry [value]="dataSource.activeUserCount"></bit-berry>
         </bit-toggle>
 
+        <bit-toggle [value]="userStatusType.Staged">
+          {{ "staged" | i18n }}
+          <bit-berry [value]="dataSource.stagedUserCount"></bit-berry>
+        </bit-toggle>
```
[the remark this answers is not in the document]
eliykat: Claude feedback looks relevant.

eliykat (changes requested):

libs/common/src/admin-console/services/policy/default-new-policy.service.ts
sven-bitwarden: Yeah I called this out in my PR body - we just need to wait for the Server -> SDK -> Client chain to go through

eliykat (commented):
Changes LGTM, just waiting on SDK changes as discussed. You should be able to:
- update the [enum definition](https://github.com/bitwarden/sdk-internal/blob/main/crates/bitwarden-organizations/src/lib.rs#L22) in the SDK
- after your SDK changes have merged, bump the SDK version used by your PR here to bring in the latest changes (see [SDK README](https://github.com/bitwarden/sdk-internal#integrating-into-clients-from-published-artifacts) for how SDK changes are published)
- bonus points! delete the TS definition of this enum and re-export the SDK definition from the same location, so that we only rely on the SDK definition instead of redefining it. I did this for the `PolicyType` enum.

libs/common/src/admin-console/services/policy/default-new-policy.service.ts
sven-bitwarden: Done!

sven-bitwarden:
@eliykat Since your review I've added the labels that are missing along with the hover-tooltip, and from our x-team discussion I've hidden the filter when there are no staged users.

sven-bitwarden:
@claude have another review you beautiful little menace.

libs/common/src/admin-console/enums/organization-user-status-type.enum.ts
```diff
@@ -1,8 +1,4 @@
...
-export enum OrganizationUserStatusType {
-  Invited = 0,
-  Accepted = 1,
-  Confirmed = 2,
-  Revoked = -1,
-}
+import { OrganizationUserStatusType as SdkOrganizationUserStatusType } from "@bitwarden/sdk-internal";
+
+// Re-export the SdkOrganizationUserStatusType so we have a single source of truth
+export { SdkOrganizationUserStatusType as OrganizationUserStatusType };
```
eliykat: Please mark as deprecated - this is for backwards compatibility only. Example because it took me a moment to realise that the jsdoc needs to be inside the export parens:

```suggestion
// Re-export the SdkOrganizationUserStatusType so we have a single source of truth
export { 
  /** @deprecated Import directly from @bitwarden/sdk-internal instead. **/
  SdkOrganizationUserStatusType as OrganizationUserStatusType
};
```

apps/web/src/locales/en/messages.json
```diff
@@ -6718,6 +6718,12 @@
   "accepted": {
     "message": "Accepted"
   },
+  "staged": {
+    "message": "Staged"
+  },
+  "stagedMemberTooltip": {
+    "message": "Admin needs to send an invite"
+  },
```
eliykat, line 6850: Is this an accurate / helpful tooltip given that admins can't actually send invites to Staged users yet? Maybe this should be "Admin needs to send an invite _link_"?

eliykat (changes requested):
Thanks for doing the SDK dance. 🦀

apps/web/src/locales/en/messages.json
sven-bitwarden: Checked in with Jillian and changed it to `Pending account invitation` - good call

eliykat (dismissed):
LGTM once CI issues are resolved.

BTreston (approved):
lgtm

PR #21427, opened 2026-06-22
[PM-33740] Remove unlock service for password flagged logic

mzieniukbw:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-33740

## 📔 Objective

Unlock service for password feature flag has been enabled in prod and the feature flagged code can now be removed from clients.


JaredSnider-Bitwarden (approved):
So nice to see us moving towards this separation b/w auth and KM! Thank you!

PR #21430, opened 2026-06-22
[PM-38904] - Managed to Claimed Rename

jrmccannon:
## 🎟️ Tracking
[PM-38904](https://bitwarden.atlassian.net/browse/PM-38904)

## 📔 Objective
The properties `managedbyOrganization` and `userIsManagedByOrganization` have been deprecated and switched to `claimedByOrganization` and userIsManagedByOrganization`.

[PM-38904]: https://bitwarden.atlassian.net/browse/PM-38904?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/common/src/admin-console/models/data/organization.data.ts
+++ b/libs/common/src/admin-console/models/data/organization.data.ts
@@ -62,7 +62,7 @@ export class OrganizationData {
   limitCollectionDeletion: boolean;
   limitItemDeletion: boolean;
   allowAdminAccessToAllCollectionItems: boolean;
-  userIsManagedByOrganization: boolean;
+  userIsClaimedByOrganization: boolean;
   useAccessIntelligence: boolean;
   useAdminSponsoredFamilies: boolean;
   useDisableSMAdsForUsers: boolean;
@@ -135,7 +135,7 @@ export class OrganizationData {
     this.limitCollectionDeletion = response.limitCollectionDeletion;
     this.limitItemDeletion = response.limitItemDeletion;
     this.allowAdminAccessToAllCollectionItems = response.allowAdminAccessToAllCollectionItems;
-    this.userIsManagedByOrganization = response.userIsManagedByOrganization;
+    this.userIsClaimedByOrganization = response.userIsClaimedByOrganization;
     this.useAccessIntelligence = response.useAccessIntelligence;
     this.useAdminSponsoredFamilies = response.useAdminSponsoredFamilies;
     this.useDisableSMAdsForUsers = response.useDisableSMAdsForUsers ?? false;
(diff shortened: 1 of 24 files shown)


jengstrom-bw (dismissed):
Vault files look good

eliykat (dismissed):
Please ensure you delete the now-unused response properties on server as well. (Preferably after 1 release.) Thanks!

libs/common/src/admin-console/models/data/organization.data.ts
```diff
@@ -135,7 +135,7 @@ export class OrganizationData {
     this.limitCollectionDeletion = response.limitCollectionDeletion;
     this.limitItemDeletion = response.limitItemDeletion;
     this.allowAdminAccessToAllCollectionItems = response.allowAdminAccessToAllCollectionItems;
-    this.userIsManagedByOrganization = response.userIsManagedByOrganization;
```
JaredSnider-Bitwarden, line 138: A local claude review caught this potential issue: 

`OrganizationData` is persisted via the `ORGANIZATIONS` `UserKeyDefinition` (see `libs/common/src/admin-console/services/organization/organization.state.ts`) to `ORGANIZATIONS_DISK`. Existing clients have records on disk written with the old key name `userIsManagedByOrganization`.

On the first app load after this change ships, `OrganizationData.fromJSON` runs:

```ts
static fromJSON(obj: Jsonify<OrganizationData>) {
  return Object.assign(new OrganizationData(), obj, { /* …dates… */ });
}
```

`Object.assign` copies the old `userIsManagedByOrganization` property onto the new instance as an extra (unused) field, and `userIsClaimedByOrganization` ends up `undefined`. Until a full sync overwrites the cache from a (renamed) server response, downstream checks like `organizations.some((o) => o.userIsClaimedByOrganization === true)` evaluate to false.

That boolean gates `showDeleteAccount$` and `showPurgeVault$` in `account.component.ts` and the blocking dialog in `account-deletion.service.ts`, so during the cache→sync window a claimed user would see those destructive actions enabled — fail-open semantics on the client.

A few options:
- Add a state-migration step that renames the cached property
- Read both names in `OrganizationData` constructor (`response.userIsClaimedByOrganization ?? response.userIsManagedByOrganization`) for one release as a transparent fallback
- Confirm the org sync is awaited before any of these UI checks run, so the disk-cache window cannot be observed in practice

jrmccannon: Good catch. Addressed

JaredSnider-Bitwarden (approved):
LGTM!

jengstrom-bw (approved):

eliykat (approved):

PR #21432, opened 2026-06-23
added vault health reports

maxkpower:
## 🎟️ Tracking
https://bitwarden.atlassian.net/browse/PM-39404

## 📔 Objective

Adds a new bw report password-health CLI command.

For every Login item with a password, the report returns:

- Password strength: score from 0 (weakest) to 4 (strongest)
- Reuse count: how many vault items share the password (null when unique or not evaluated)
- Exposure: whether the password appears in a known breach via Have I Been Pwned, the breach count, and a per-item error message if that lookup failed

Usage:
```
bw report password-health # full report (includes HIBP check)
bw report password-health --no-check-exposed  # skip HIBP network calls
bw report password-health --pretty # pretty-printed JSON
```

Details:

- Output uses the standard Response/ListResponse objects, so it respects BW_RESPONSE/--pretty and stays valid JSON on stdout.
- Because the HIBP check makes a network request per login, a progress indicator is written to stderr (stdout is reserved for the result). It animates on a TTY, degrades to a single static line when piped, and is silenced under BW_QUIET.
- The risk computation reuses the existing shared CipherRiskService (buildPasswordReuseMap + computeRiskForCiphers) rather than adding CLI-specific risk logic. The riskable-login predicate is extracted to an exported isRiskableLoginCipher helper so the command and the service evaluate the same set.


apps/cli/src/dirt/commands/report.command.ts
```diff
@@ -0,0 +1,98 @@
...
+import { uuidAsString } from "@bitwarden/common/platform/abstractions/sdk/sdk.service";
+import { CipherRiskService } from "@bitwarden/common/vault/abstractions/cipher-risk.service";
+import { CipherService } from "@bitwarden/common/vault/abstractions/cipher.service";
+import { isRiskableLoginCipher } from "@bitwarden/common/vault/services/default-cipher-risk.service";
+
+import { Response } from "../models/response";
+import { ListResponse } from "../models/response/list.response";
+import { PasswordHealthResponse } from "../models/response/password-health.response";
+
+const SPINNER_FRAMES = ["|", "/", "-", "\\"];
```
gbubemismith, line 15: 👏🏼 this is a nice addition. I don't think we have this anywhere else 👍🏼

gbubemismith (dismissed):
Nice work @maxkpower, this looks good

djsmith85 (commented):
@maxkpower: IMHO this should be owned and reviewed by @bitwarden/team-data-insights-and-reporting-dev

gbubemismith:
@maxkpower, can you also update the `CODEOWNERS` for these newly added files to be owned by `DIRT`?

AlexRubik (dismissed):

apps/cli/src/dirt/models/response/password-health.response.ts
```diff
@@ -0,0 +1,35 @@
...
+    this.name = name;
+    this.passwordStrength = result.password_strength;
+    // SDK reports occurrence count (1 = unique), so only >1 counts as reused.
+    this.reuseCount =
+      result.reuse_count != null && result.reuse_count > 1 ? result.reuse_count : null;
+
+    const exposed = result.exposed_result;
+    this.exposed = exposed.type === "Found" && exposed.value > 0;
+    this.exposedCount = exposed.type === "Found" && exposed.value > 0 ? exposed.value : null;
+    this.exposedError = exposed.type === "Error" ? exposed.value : null;
```
Banrion, line 33: ❓ Exposed where checked and none are found looks the same in a response where exposed was not checked. Looks intentional, but wanted to call it out in case we wanted to differentiate between exposed count is 0 (safe) vs exposed passwords not being counted at all (null).

Banrion (approved):

gbubemismith (approved):

theMickster (approved):
Approving codeowners changes on behalf of tech leads.

PR #21433, opened 2026-06-23
feat(biometrics): add windows biometrics format based on sdk cryptography

quexten:
## 🎟️ Tracking

https://github.com/bitwarden/sdk-internal/pull/1204
https://bitwarden.atlassian.net/browse/PM-35105

## 📔 Objective

Splits up the windows-hello biometric unlock.

Most of the diff here is cargo lock. Moves the encryption into a separate file and adds a new format based on SDK cryptography. 

Overview of crypto: We store a per-enrollement challenge. This is used via keycredentialmanager to sign the challenge deterministically. The signature is hashed to produce a high-entropy-secret "PRF". This is fed into the secretkeyenvelope seal / unseal. The secretkeyenvelope directly contains the user's user key.

## 📸 Screenshots

<!-- Required for any UI changes; delete if not applicable. Use fixed width images for better display. 

## QA
Tested in a windows VM, but needs QA testing before merge.

### DIFF
--- a/apps/desktop/desktop_native/Cargo.toml
+++ b/apps/desktop/desktop_native/Cargo.toml
@@ -30,6 +30,9 @@ arboard = { version = "=3.6.1", default-features = false }
 ashpd = "=0.13.11"
 async-trait = "=0.1.89"
 base64 = "=0.22.1"
+bitwarden-crypto = { git = "https://github.com/bitwarden/sdk-internal", rev = "f76b51450582cff9be767c9d9f454c42a84a56c1" }
+bitwarden-random = { git = "https://github.com/bitwarden/sdk-internal", rev = "f76b51450582cff9be767c9d9f454c42a84a56c1" }
+bitwarden-sensitive-value = { git = "https://github.com/bitwarden/sdk-internal", rev = "f76b51450582cff9be767c9d9f454c42a84a56c1" }
 byteorder = "=1.5.0"
 bytes = "=1.11.1"
 cbc = "=0.1.2"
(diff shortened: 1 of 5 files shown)


apps/desktop/desktop_native/biometric/src/windows/keycredentialmanager/mod.rs
```diff
@@ -0,0 +1,82 @@
+//! Windows-hello-based PRF derivation via KeyCredentialManager
+//!
+//! This module owns the interaction with Windows Hello: creating/opening the signing credential,
+//! selecting a signing backend, and turning the resulting signature into a [`WindowsHelloPrf`]. The
+//! cryptography that consumes the PRF lives in [`super::encryption`].
+
+mod backend_legacy;
```
quexten: Note: We would add support for the new function with window handle without manual focus here. Sadly, the rust crate does not have a stable release yet, so we cannot add it in this PR.

dani-garcia (dismissed):
Platform changes lgtm

apps/desktop/desktop_native/Cargo.toml
```diff
@@ -53,6 +56,9 @@ napi-derive = "=3.5.4"
 oo7 = "=0.5.0"
 pin-project = "=1.1.10"
 rand = "=0.9.3"
+# rand_core 0.10 (a distinct major from `rand` 0.9) provides the `Rng` trait needed to draw bytes
+# from `bitwarden_random::rng()`, whose RNG type is built on the 0.10 rand-core traits.
+rand_core = "=0.10.1"
```
eligrubb, line 61: ❓ Could `bitwarden-random` re-export the `Rng` trait, and the biometrics crate can import from there? Then we could eliminate the need for a direct dependency on `rand_core` for biometrics. Might even be able to remove the `rand` crate direct dependency from `desktop_native` too.

Probably best as future tech debt.

eligrubb (dismissed):
Good stuff!! One non-blocking thought is all.

apps/desktop/desktop_native/Cargo.toml
quexten: Ohh, interesting. Yes I will get this tracked.

quexten: https://bitwarden.atlassian.net/browse/PM-40804

quexten:
Given the next RC cut is for web/server and not desktop, I'll merge and we'll QA on `main`.

dani-garcia (approved):

eligrubb (approved):

PR #21439, opened 2026-06-23
[PM-38896] Enhance DefaultCollectionService tests and improve decryption

JaredScar:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-38896

## 📔 Objective

Fixes infinite loop issue for subscription which happens because logged out state still tries to decrypt even though user is logged out


eliykat (changes requested):
I assume this works but it's not my preferred solution - it's using the presence of keys as a proxy for the account state in a way that breaks the abstraction. That is - we should be able to ask, "is this account unlocked/active?", not "do we have encryption keys available?" Checking for keys is a bit of a hack.

What's odd is that the general pattern is to subscribe to `accountService.activeAccount$` to get the `userId` to then call these methods. So I would expect that any time it's called, it's with an active UserId. That said, account state can be unstable during transitions.

Have you / can you identify the subscription that's throwing this error? Is it subscribing to the `activeAccount$`?

Assuming that checks out - can you follow up with Platform or maybe in `#sdk` about what the recommended way of safely accessing `userClient$` is?

jrmccannon:
I believe the package-lock.json file is causing issues.

eliykat (commented):
Holding any further reviews on this until we get further guidance from Platform next week.

jrmccannon (dismissed):

eliykat (changes requested):
As discussed in Slack, please handle batch level decryption failures in a way that avoids an infinite loop -> logs filling up -> crashing.

libs/admin-console/src/common/collections/services/default-collection.service.ts
```diff
@@ -127,6 +134,7 @@ export class DefaultCollectionService implements CollectionService {
         ]).pipe(
           switchMap(([collections, orgKeys]) =>
             this.decryptMany$(collections, orgKeys).pipe(
+              catchError((error: unknown) => this.handleBatchDecryptionFailure(error, userId)),
```
eliykat: Even though the current (non-feature flagged) implementation theoretically suffers from this issue as well, we've never seen it in practice. Let's revert this line so that we're not making unnecessary changes to the current implementation. The new error handling should be on the new code only.

libs/admin-console/src/common/collections/services/default-collection.service.ts
```diff
@@ -135,6 +143,31 @@ export class DefaultCollectionService implements CollectionService {
...
+   * Without this, a rejected/errored decryption batch would propagate an error through
+   * `decryptedCollections$`, which is cached per-user via `shareReplay`. Since that cache resets
+   * on error, the next subscriber (e.g. a component re-rendering) would immediately re-trigger
+   * decryption of the same batch, fail again, and log again - repeating indefinitely, filling up
+   * logs, and potentially crashing the app.
+   *
+   * Instead, we log the failure once here and fall back to an empty collection list, which is
+   * persisted via `setDecryptedCollections`. This prevents re-initialization (and therefore
+   * repeated decryption attempts) until the cache is explicitly invalidated, e.g. by `replace()`
+   * on the next full sync.
```
eliykat: This is too far in the other direction. These failure are transient (e.g. page has loaded but SDK hasn't initialized yet), but this fix will persist the empty array in memory and won't try again, even after the transient state has passed. We want to allow the observable time to settle without crashing the app, but also we _do_ want it to retry once the transient state has passed.

eliykat (changes requested):
I don't think this error handling will work - see comments below. These observables are genuinely very difficult to reason about so I wanted to make a concrete suggestion - have a look at https://github.com/bitwarden/clients/pull/21624.

It's very close to what you have, but the main difference is that `catchError` is moved after the `delayWhen` which writes the value to cache - so an error will propagate down the chain, the cache won't be updated, and then the error will be caught and emitted as an empty array. That catch + emission prevents an infinite retry loop on the same subscriber, but it's also not written to cache so it won't persist. If the encrypted collections or the key state changes it'll try again.

Please tell me if this makes sense, if you agree with the fix then can you please manually test it to make sure it's stable and solves the defect we were having. If not we can sync up. I want to emphasise that I haven't tested this against the actual defect.

eliykat (dismissed):
LGTM, let me know how you go with testing it.

eliykat:
In our conversation this morning, I meant only to remove the duplicate `logService` call in `DefaultCollectionService`, because the error is already logged where it occurs in `DefaultCollectionEncryptionService`. We can't remove the empty array emission, because that is critical to recovering from the error and not having `shareReplay` continually re-subscribe to try again, which causes the churn.

I've pushed some commits to show you. However I can't approve my own commits so you'll just have to grab someone else from the team to do the final review. You can see the net changes since my last review here: https://github.com/bitwarden/clients/pull/21439/changes/469e35b821c104d79ace31ce45545ebde4027f73..fcf42f6d7be57948df8771b3d6b44680ba0f3e39#top

BTreston (approved):

PR #21450, opened 2026-06-24
[PM-39451] Add IPC reachability across browser, desktop, and web

quexten:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-39451
SDK change: https://github.com/bitwarden/sdk-internal/pull/1213

## 📔 Objective

Adds a lightweight plaintext ping/pong keepalive protocol across the IPC layer so each participant can check whether its leader is reachable before attempting a crypto-channel (Noise) send.


apps/web/src/app/platform/ipc/web-ipc.service.ts
```diff
@@ -5,23 +5,35 @@ import { PlatformUtilsService } from "@bitwarden/common/platform/abstractions/pl
...
+import {
+  Endpoint,
   IncomingMessage,
   IpcClient,
   IpcCommunicationBackend,
   ipcRegisterDiscoverHandler,
   OutgoingMessage,
 } from "@bitwarden/sdk-internal";
 
+// Web's only leader is the browser extension background.
```
quexten: I don't think leader is the right term here because this is not the shared unlock context. It's more "direct peers" (since we don't do forwarding here), *but*, for desktop, it would not reach out to the direct peer, it would get reached out by the direct peer.

apps/desktop/src/services/biometric-message-handler.service.ts
```diff
@@ -105,6 +105,11 @@ export class BiometricMessageHandlerService {
   async handleMessage(msg: LegacyMessageWrapper) {
     const { appId, message: rawMessage } = msg as LegacyMessageWrapper;
 
+    if (rawMessage == null) {
```
quexten, line 108: Not strictly required since we catch this elsewhere now, but still a good defensive measure.

PR #21455, opened 2026-06-24
[PM-36407] Ts strict bulk member components

BTreston:
## 🎟️ Tracking
https://bitwarden.atlassian.net/browse/PM-36407

## 📔 Objective
Refactors the member bulk action dialogs to be compliant with ts-strict + use modern angular in adherence to our ADRs and best practices.


apps/web/src/app/admin-console/organizations/members/components/bulk/bulk-confirm-dialog.component.ts
```diff
@@ -32,53 +39,66 @@ type BulkConfirmDialogParams = {
...
+    this.organizationKey$ = this.accountService.activeAccount$.pipe(
+      getUserId,
       switchMap((userId) => this.keyService.orgKeys$(userId)),
-      map((organizationKeysById) => organizationKeysById[this.organization.id as OrganizationId]),
-      takeUntilDestroyed(),
+      map((orgKeys) => {
+        const orgId = this.dialogParams.organization.id;
+        const orgKey = orgKeys?.[orgId] ?? undefined;
+        if (orgKey == null) {
+          throw new Error(`Organization key not found for org ${orgId}`);
```
JimmyVo16, line 79: I saw a few new error checks like this. I’m curious: do we surface this error to the user as is? Are there any concerns about localization or users not understanding the technical nature of the error?

JimmyVo16 (commented):
Pretty much approved, just one question.

apps/web/src/app/admin-console/organizations/members/components/bulk/bulk-confirm-dialog.component.ts
BTreston: I don't think its a concern, there's not much precedence for i18n'ing these kinds of errors in the code base that I can see. Maybe worth bringing up on the team sync though to see if there's a preferred direction from product. This won't merge until next sprint because we're at the tail end and this is non-feature flagged so I don't mind holding until the next team sync.

Banrion (approved):
✅ DIRT files

AlexRubik (approved):

JimmyVo16 (approved):

PR #21465, opened 2026-06-24
PM-38122 resolved fastmail checkbox

bmbitwarden:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-38122

## 📔 Objective

Add checkbox to Generator page to specify Fastmail website prefix

## 📸 Screenshots

image


libs/tools/generator/core/src/integration/fastmail.ts
```diff
@@ -71,7 +71,7 @@ const createForwardingEmail = Object.freeze({
                 state: "enabled",
                 description: "",
                 forDomain: context.website(request),
-                emailPrefix: "",
+                emailPrefix: context.prefixEnabled() ? context.website(request) : "",
```
mcamirault: ⚠️ You'll want to get the domain instead of the full website so that the behavior matches the description (use `Utils.getDomain`). According to [Fastmail's API docs](https://www.fastmail.com/dev/) the `emailPrefix` field has some limitations:

> The string MUST be <= 64 characters in length and MUST only contain characters a-z, 0-9 and _ (underscore)

So I'd recommend taking the domain string, lowercase it, replace any non-compliant character with `_`, replace any runs of more than one `_` in a row with a single `_`, and cut it at length 64. That should give you pretty readable prefixes that always comply with the API spec.

mcamirault (changes requested):
One change needed so the behavior matches the description

mcamirault (approved):
LGTM

PR #21477, opened 2026-06-24
[PM-39269] Fix At-Risk Members CSV export including extra columns

AlexRubik:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-39269

## 📔 Objective

The V2 Access Intelligence "At-Risk Members" CSV export (behind the `pm-31936-access-intelligence-new-architecture` flag) incorrectly included `userName` and `userGuid` columns. The export should contain only the member email and the at-risk application count.

Root cause: `DrawerMemberData` carried `userName`/`userGuid` fields that were written by the page-component content builders but never read by the drawer template or any logic. Because the shared `exportToCSV` helper emits every key present on each row object, these unused fields leaked into the CSV as extra columns.

This change removes the two dead fields from `DrawerMemberData` and stops populating them in `mapMembersToDrawerData` and `getCriticalAtRiskMembersContent`, so the export now contains only `email` and `atRiskApplicationCount`. The shared `exportToCSV` helper is intentionally left untouched (it backs several other exports), keeping the change scoped to the at-risk-members path. The member registry and report serialization are unaffected. The drawer download spec assertion is tightened to the two-column header and a regression test guards against the fields reappearing.

### DIFF
--- a/bitwarden_license/bit-web/src/app/dirt/access-intelligence/v2/shared/access-intelligence-drawer-v2/access-intelligence-drawer-v2.component.spec.ts
+++ b/bitwarden_license/bit-web/src/app/dirt/access-intelligence/v2/shared/access-intelligence-drawer-v2/access-intelligence-drawer-v2.component.spec.ts
@@ -29,13 +29,8 @@ describe("AccessIntelligenceDrawerV2Component", () => {
 
   /** Sample members used in drawer data */
   const sampleMembers: DrawerMemberData[] = [
-    {
-      email: "alice@example.com",
-      userName: "Alice Smith",
-      userGuid: "u1",
-      atRiskApplicationCount: 5,
-    },
-    { email: "bob@example.com", userName: "Bob Jones", userGuid: "u2", atRiskApplicationCount: 3 },
+    { email: "alice@example.com", atRiskApplicationCount: 5 },
+    { email: "bob@example.com", atRiskApplicationCount: 3 },
   ];
 
   /** Sample applications used in drawer data */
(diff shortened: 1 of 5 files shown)


bitwarden_license/bit-web/src/app/dirt/access-intelligence/v2/shared/access-intelligence-drawer-v2/access-intelligence-drawer-v2.component.spec.ts
```diff
@@ -122,7 +117,22 @@ describe("AccessIntelligenceDrawerV2Component", () => {
...
+        type: DrawerType.OrgAtRiskMembers,
+        members: sampleMembers,
+      };
+
+      component.downloadAtRiskMembers();
+
+      const callArg = mockFileDownloadService.download.mock.calls[0][0];
+      const header = (callArg.blobData as string).split("\n")[0];
+      expect(header).not.toContain("userName");
+      expect(header).not.toContain("userGuid");
```
lastbestdev: This test is redundant and can be removed. You have updated the test case above this one to cover that the headers are exactly what we expect and nothing more. 👍🏻

lastbestdev (dismissed):
Doesn't appear that the `userName, userGuid` properties were being used by the drawers at all, so this LGTM. One unit test should be removed.

bitwarden_license/bit-web/src/app/dirt/access-intelligence/v2/shared/access-intelligence-drawer-v2/access-intelligence-drawer-v2.component.spec.ts
AlexRubik: Good catch ty , https://github.com/bitwarden/clients/pull/21477/commits/1bac11b1aeb0109e7e8e247f5dce285352512c41

Banrion (approved):

lastbestdev (approved):

PR #21498, opened 2026-06-25
[PM-39593] reduce calls to `nodeIsElement` in `query` (and by extension `queryAutofillFormAndFieldElements`)

### LINKED ISSUE #21453: High CPU use when a page has a lot of input elements

datdenkikniet:
This function (which seems to be in the hot path of some of the autofill checking), called for a lot of nodes, checks that a value of type `Element` is an element, which is redundant.

Reduce the amount of redundant checks by propagating the actual type of the objects that are being filtered (`Element`) and refactoring the filter callbacks accordingly.

The one check that is unchanged is the on in `elementIsInstanceOf`, as there are some indirect callers of that function that may not yet _actually_ only call that function with an `Element`.

## 🎟️ Tracking

N/A

## 📔 Objective

In a (novice) attempt to figure out possible causes for the high CPU usage some users are experiencing (including myself, and other examples such as #21453), I stumbled across this seemingly simple optimization. In some of the performance traces I've been looking at, `elementIsInstanceOf` made up 484 out of 7380 non-idle samples, and half of those were `nodeIsElement` samples.

I'm mostly creating this PR to see if there are any tests that I have not managed to find/run clearly throw an error for this change.


datdenkikniet:
Hmm, okay, when checking the history for this file, e973d72b0170a9fe40a7f4a0ac383c34bc84e2f0 seems to have added exactly the check that I've removed here to `elementIsInstanceOf`. Perhaps someone internal to bitwarden has better insight into this (seeminly rolled up?) commit, adding this potentially expensive check (due to being on the hot path for `buildTreeWalkerNodesQueryResults`) to a function that really doesn't appear to need it.

I guess it may be due to the TS becoming public as a JS API through the functions turned `export` in that commit. Is there a way to avoid the additional, expensive check in TS in that case (beyond having some sort of `elementIsInstanceOfInternal` that skips the `isElement` check)?

audreyality:
Hi @datdenkikniet! Thanks for your interest in contributing to Bitwarden!

This check is performed because a variable's type is erased when typescript compiles, and downstream code requires that the element is a real DOM element. While it is called often, checking the type does not have an appreciable effect on performance in our tests. 

I'm inclined to believe that the "this is a real DOM element" guarantee enforced by this function is worth the cost. If you have a test that can illustrate its cost, I'd like to see it! 

Of particular interest are real-world examples. A [synthetic benchmark in BIT](https://github.com/bitwarden/browser-interactions-testing/tree/main/benchmarks) would also be useful.

In the meantime, I'm going to switch this contribution to a draft. That communicates to the rest of my team that the PR isn't yet ready for a formal review. 😄

datdenkikniet:
I guess that makes sense.

The real-world usecase where (I think) this actually is observable is the review page of [gerrit](https://www.gerritcodereview.com/) change requests with semi-large changes (1000+ lines changed).

When adding comments to a file (by expanding it, selecting some text, and pressing `c`), like 1/10th of the time bitwarden spends is in `elementIsInstanceOf`, and over half of that time is spent on `nodeIsElement`. It is perhaps not a particularly reproducible or super-often-occuring scenario, and the actual time spent in this code may be acceptable, but it is not great.

For example, is on Firefox 151, using v2026.5.1 of the extension. I'm adding a comment to the +968 file on [this page](https://gerrit-review.googlesource.com/c/git-repo/+/596721). Even without the extension, it's not fast, but activating the extensions definitely doesn't improve things. 

I've attached the profile for good measure (I don't expect you to look at it: I'm mostly uploading it to show that there are some cases in which this may have some effect, and so that I have an easier time finding it back in the future). It shows the behaviour I describe above.

I agree that keeping this as a draft is a good idea. I will try to see if I can cook up a benchmark that shows this, somehow.

[Firefox 2026-06-25 23.14 profile.json.gz](https://github.com/user-attachments/files/29358253/Firefox.2026-06-25.23.14.profile.json.gz)

audreyality:
> The real-world usecase where (I think) this actually is observable is the review page of [gerrit](https://www.gerritcodereview.com/) change requests with semi-large changes (1000+ lines changed).

~~Do you have a publicly reachable page I could observe this with?~~

Nevermind. I scanned too quickly and didn't notice you already did this.

apps/browser/src/autofill/utils/index.ts
```diff
@@ -213,7 +213,7 @@ export function elementIsInstanceOf<T extends Element>(
   element: Element,
   tagName: string,
 ): element is T {
-  return nodeIsElement(element) && element.tagName.toLowerCase() === tagName;
+  return element.tagName.toLowerCase() === tagName;
```
audreyality: 🤔 So, this change is a little too localized to provide the kind of benefit that we'd need to accept the change. It shows up on your profile because we call it frequently. That's likely due to inefficiencies elsewhere.

I'm wondering whether there's opportunities to address how many times we call this method further up the call graph. If we can identify those, then we could remove the `nodeIsElement(...)` checks selectively where we have better guarantees of type safety.

If you'd like to build towards that, we can explore it here. It won't be the quick-win you were hoping for, but it could help you feel less like a novice!

datdenkikniet: > where we have better guarantees of type safety

I'm not sure I understand _why_ we need "better guarantees of type safety". While I can imagine that being extra lenient can be handy, this specific case has me puzzled: the API is fully typed TS expecting an `Element`, called only by other TS. Why is it undesirable to push the responsibility of providing a value of the correct type to the caller? Are there JS users that might call this function and not know the type requirement? Additionally: where do we draw the line? `nodeIsElement` expects a `Node`, and if we cannot be certain that the `element` passed to `elementIsInstanceOf` is an `Element`, how can we be sure that it is a `Node`?

I guess the main thing that makes me think that there should be some improvement here is that the current code practically forces the following pattern:

1. Some code has a `node: Node`
2. To be allowed to call `elementIsInstanceOf<T>` from TS, it must first "cast" it to an `Element` by calling `nodeIsElement(node)`
3. Then, it calls `elementIsInstanceOf<T>(node)`, which immediately calls `nodeIsElement(node)` again.

Examples are all four of the `nodeIsXElement` functions [here](https://github.com/bitwarden/clients/blob/main/apps/browser/src/autofill/utils/index.ts#L319). For example, `nodeIsFormElement` shows up quite a bit in the profile, and goes through the exact dance I describe above.

In and of itself this may not be super expensive, but this code clearly gets called a lot in some scenarios, so doing double work seems pretty undesirable.

Another alternative could be to add a `nodeIsInstanceOf<T extends Node>` that is called in the `nodeIsXElement` functions instead. Duplicating the `node.TagName` check across multiple functions feels unpleasant, but perhaps a necessary evil.

datdenkikniet: Also, I appreciate that you are willing to engage in a bit of a dialog on this, thank you for that!

audreyality: > I'm not sure I understand why we need "better guarantees of type safety". 

The reason why we need to worry about type safety guarantees is because this code has a history that pre-dates our use of typescript. There are places in your profile's flame graph where we're processing large numbers of elements, but we use `CallableFunction` as the type, which erases type signatures. The problem here is that `CallableFunction` is a concrete type, but [it's argument accepts any](https://github.com/microsoft/TypeScript/blob/637d5746b70257028fb95aad32ddec6b26ab0a14/src/lib/es5.d.ts#L263) (note: [CallableFunction extends Function](https://github.com/microsoft/TypeScript/blob/637d5746b70257028fb95aad32ddec6b26ab0a14/src/lib/es5.d.ts#L313))!

For one such example, check out [buildTreeWalkerNodesQueryResults(...)](https://github.com/bitwarden/clients/blob/main/apps/browser/src/autofill/services/dom-query.service.ts#L554-L560). The call graph can end up here indirectly\* through [queryAutofillFormAndFieldElements(...)](https://github.com/bitwarden/clients/blob/783d4159b2dcf9b303cf194fee9f0d5498ea18c3/apps/browser/src/autofill/services/collect-autofill-content.service.ts#L1326-L1343). See that `Node` type on line 1329? That's an assumption that typescript allows because `(t: Node) => boolean` is assignable to `(t: any) => any`. The type system looks like it's working, but it's actually short-circuiting!

In order to fix these kinds of problems, we need to structure the algorithms so that we can trust the type signatures. It's not as simple, however, as just updating the typechecks. There's too many places we'd need to update. The PR would end up too large to review competently. So, instead, we find smaller chunks of work to do. (In the world of weird metaphors: "How do you eat an elephant? One bite at a time.)

> I guess the main thing that makes me think that there should be some improvement here is that the current code practically forces the following pattern...

Yes. That's why I'm asking how interested you are in jumping down the rabbit hole. It looks like you've found an interesting optimization problem. Solving it involves a bunch of [yak shaving](https://en.wiktionary.org/wiki/yak_shaving).

For what it's worth, there's almost always yak shaving involved in this kind of work. Performance problems for code as mature as Bitwarden's are often structural. If it were as easy as removing a type guard, we'd have already done it. 🫠 

---

\* I found that by tracing the path through the code manually. It's not included in your profile, because your code is JIT optimized, and SpiderMonkey is inlining the intermediate function calls.

datdenkikniet: > The type system looks like it's working, but it's not!

Ah, that's unfortunate!

So, to check that I've understood this correctly: there's no concern about _external_ code running into issues with this function, there's "just" internal type-erasure shenanigans that make the additional check necessary. _That_ sounds like something I can fix.

Would properly typing up all call paths leading to `elementIsInstanceOf` be sufficient to tighten the type-safety guarantees? I'm definitely up for shaving some yaks if it means that BitWarden becomes a bit better.

audreyality: It's unlikely to be enough. Check out, for example, from about 6.3s to 6.7s in the flame chart, only looking at scripts. You'll see that `nodeIsElement` is taking up roughly 30-40ms several times throughout that call. Now, not all of those are processing the same list of changes (at least one of them is processing nodes in a closed shadow root). However! There's a good chance that a few of those calls are processing *substantially the same list*. 

image

That's, in fact, something we know for sure, because the code says as much. All of that processing is happening in procedural code. For each item that `buildTreewalkerNodesQueryResults` processes that's an input in that hot path, we're calling it at least 3 times:

* [One](https://github.com/bitwarden/clients/blob/783d4159b2dcf9b303cf194fee9f0d5498ea18c3/apps/browser/src/autofill/services/collect-autofill-content.service.ts#L1330C39-L1333) (indirect)
* [Two](https://github.com/bitwarden/clients/blob/783d4159b2dcf9b303cf194fee9f0d5498ea18c3/apps/browser/src/autofill/services/collect-autofill-content.service.ts#L1371-L1373)
* [Three](https://github.com/bitwarden/clients/blob/main/apps/browser/src/autofill/services/dom-query.service.ts#L577)

Swapping out the function call won't be enough to solve this problem, because the indirection presents a structural barrier. Ideally, we'd want to take the list of all elements returned by a selector, then repeatedly reduce it to smaller and smaller lists. As the list gets more specific, we run different, more expensive tests. 

```
const nodes: Node[] = ...;
const elements = nodes.filter(isElement);

// `isInteresting` here is a fast filter that may not be right, but is never wrong
const interesting = elements.filter(isInterestingElement);

// fields will outnumber forms by a wide margin, so look for them first
const fields = new Set(interesting.filter((e) => expensiveHeuristic(e)));

// then turn around and `use Set.has()` as an `interestingElement` check!
const forms = interesting.filter((e) => fields.has(e) && isFormElement(e));
```

What we're doing now is running substantially-similar tests on the same, large list. Why? As written, this is all that our algorithm can do. The type erasure isn't the only problem. The indirection is also causing trouble, because it prevents us from using the kinds of tricks I alluded to in the above code block.

The yak shaving is first in figuring out what functions produce good `isInterestingElement` filters. Then it's in figuring out how to go from what we have now to what we need. Oh, yeah, and in between you've got to convince sticklers like me that you've shaved the right yak.

audreyality: That all being said, it's now pretty late in my day. I'd like to confirm with my fellow team members that they'd like to pursue this optimization. I'll get back to you early next week, after we've talked it over.

datdenkikniet: One additional friction point I've come across now, which could qualify as a decent `isInterestingElement` thing, is the fact that the collect-autofill-content service goes to some lengths to construct a solid `queryString` to pass to `query`, which is then promptly ignored entirely (since the code path that actually uses said `queryString` requires passing `forceDeepQueryAttempt = true` to `query`, which from a brief inspection, a total of 0 `query` callers actually do. It seems that this option was used when `query` was introduced to the codebase, but it no longer is).

So, the filtering logic is duplicated in some sense (once for constructing the query string, and once for the filtering function), but the first (from what I can tell more efficient?) instance of the logic is effectively unused. Assuming my analysis is correct, it seems that there's some dead code that can be shaved. That's probably for a different PR, though.

Not sure how/if this fits into the discussion we're having here (I assume we could see some perf improvements if we get better at filtering sooner), but I saw it and figured I'd comment on it.

> I'll get back to you early next week, after we've talked it over.

SGTM!

datdenkikniet: Ah, but it gets better: all of the (indirect) callers of `buildTreeWalkerNodesQueryResults` assume that it might call the filter with non-`Element` nodes. However, `NodeFilter.SHOW_ELEMENT` is passed to `document.createTreeWalker` in that function, meaning that _all_ nodes it yields are already `Element`s... So in effect, every single call to `nodeIsElement` is redundant, since all of them happen in the call stack of a `query` filter, which is always a `buildTreeWalkerNodesQueryResults`, which will only ever call the filter function with `Element`s. In other words, `query<T>` is really `query<T extends Element>`, and the filter callback is really `(element: Element) => bool`, and all `isNodeElement` and `isElementXElement` functions (appear to be) called only in `query` calls.

This is a complicated yak, best not to hold it upside down I guess :P

datdenkikniet: Ah, but it is, as you said, not as easy as I make it sound: `NodeFilter.SHOW_ELEMENT` _also_ accepts shadow roots, which aren't `Element`s... 

From what I can tell: `ShadowRoot.nodeType === Node.DOCUMENT_FRAGMENT_NODE`, but it is still yielded from a tree walker with `NodeFilter.SHOW_ELEMENT` (on firefox at least),

But, a single "is this an Element" check should still suffice, as opposed to multiple ones.

ETA: ahh, okay, of course: `currentNode` is the root node if no iterator-ish function has been called, which is always: the root node of the document OR a shadow root. So the walker's filter is never applied to the root node, so it is observed because the search uses `let currentNode = walker.currentNode` (and the spec doesn't say anything about applying the filter before setting `currentNode`) as starting condition. All non-root nodes _are_ `Element`s.

datdenkikniet:
> A [synthetic benchmark in BIT](https://github.com/bitwarden/browser-interactions-testing/tree/main/benchmarks) would also be useful.

I've tried to get BIT to do what I want for a few hours, but am stuck on getting it running at all (`npm run setup:all` fails without a clear error. Some commands fail in `npm run build:extension`, but I cannot figure out how to get it to tell me why/how). 

I'm fairly confident that a test that does nothing but load [this file](https://github.com/user-attachments/files/29437182/basic_diff.html), however, will demonstrate the problem: it's simply a large HTML file, making up the contents of the shadow DOM element that is loaded in on the gerrit page I linked earlier.

Just opening the page/file in firefox while running a capture shows that bitwarden (on my machine) spends 194 ms in `query`, and of that time, a little over half the samples (52%) are spent in `nodeIsElement`.

datdenkikniet:
Pushed to fix lints (code style) & conflicts

datdenkikniet:
Has there been any movement on this? I still think it's somewhat important to get this change merged at some point.

I've tried to get BIT working for a few more hours, but it just won't seem to even run the included benchmark (can't log in to the extension, and if you comment out some lines to make it ignore actually loading it still fails to find some start button, it seems).

As an alternative, I've just applied a patch that adds a 100x loop to `query`, and some `performance.mark` measurements:

<details>
<summary>patch</summary>

```diff
+++ b/apps/browser/src/autofill/services/dom-query.service.ts
@@ -70,6 +70,21 @@ export class DomQueryService implements DomQueryServiceInterface {
     const ignoredTreeWalkerNodes = ignoredTreeWalkerNodesOverride || this.ignoredTreeWalkerNodes;
 
     if (!forceDeepQueryAttempt) {
+      performance.mark("start queryAllTreeWalkerNodes loop")
+      for (var i = 0; i < 100; i++) {
+        performance.mark("start queryAllTreeWalkerNodes")
+        this.queryAllTreeWalkerNodes<T>(
+          root,
+          treeWalkerFilter,
+          ignoredTreeWalkerNodes,
+          mutationObserver,
+        );
+        performance.mark("end queryAllTreeWalkerNodes")
+        performance.measure("queryAllTreeWalkerNodes", "start queryAllTreeWalkerNodes", "end queryAllTreeWalkerNodes")
+      }
+      performance.mark("end queryAllTreeWalkerNodes loop")
+      performance.measure("full loop", "start queryAllTreeWalkerNodes loop", "end queryAllTreeWalkerNodes loop")
+
       return this.queryAllTreeWalkerNodes<T>(
         root,
         treeWalkerFilter,
```

</details>

to measure the impact of this (now rather substantial) change.

I get the following traces when loading the HTML page I linked [in my comment above](https://github.com/bitwarden/clients/pull/21498#issuecomment-4826184959): [this-branch.json.gz](https://github.com/user-attachments/files/29760945/this-branch.json.gz) and [main.json.gz](https://github.com/user-attachments/files/29760950/main.json.gz) (and for fun, without `nodeIsElement` in the path at all: [no-more-nodeiselement.json.gz](https://github.com/user-attachments/files/29760964/no-more-nodeiselement.json.gz)).

Selecting the duration of the loop clearly shows that the client goes from spending 53% of it's `query`-ing time in `nodeIsElement` to "only" 29%. That's still a lot of time spent double-checking something that is already (almost) certain, but at least it's better. Naturally, _actually_ fully removing `nodeIsElement` is still high on my wishlist ;)

I've tried my best to divvy up the changes into acceptable, single-purpose commits, but not sure if the PR would ever be acceptable in its current form.

apps/browser/src/autofill/services/collect-autofill-content.service.ts
```diff
@@ -1363,26 +1361,22 @@ export class CollectAutofillContentService implements CollectAutofillContentServ
...
    * @returns {boolean}
    * @private
    */
-  private isNodeFormFieldElement(node: Node): boolean {
-    if (!nodeIsElement(node)) {
-      return false;
-    }
-
-    const nodeTagName = node.tagName.toLowerCase();
+  private isElementFormFieldElement(element: Element): boolean {
```
audreyality, line 1395: ⛏️ The JSDoc is out of date

apps/browser/src/autofill/services/dom-query.service.ts
```diff
@@ -563,20 +563,29 @@ export class DomQueryService implements DomQueryServiceInterface {
...
+      // `currentNode` can be one of two things: the root node (which is a `Node`),
+      // or an `Element` (due to the `NodeFilter.SHOW_ELEMENT`). Therefore,
+      // `currentNode` is an `Element` if it is not the root node, or if it
+      // is an element.
+      let currentElement: Element | null = null;
+      if (currentNode != treeWalker.root || nodeIsElement(currentNode)) {
+        currentElement = currentNode as Element;
+      }
+
+      if (currentElement != null && filterCallback(currentElement)) {
```
audreyality: 👍🏻 This is a pretty clever bit of machinery that avoids the `Element` check, except in the case of the root node, which could be any kind of a node.

⚠️ The repeated `currentNode != null` checks are a code smell. You can eliminate them if you continue when the node is not an element.

```ts
if (currentNode != treeWalker.root || nodeIsElement(currentNode)) {
  currentElement = currentNode as Element;
} else {
  currentNode = treeWalker?.nextNode();
  continue;
}
```

apps/browser/src/autofill/services/autofill-overlay-content.service.ts
```diff
@@ -599,15 +599,17 @@ export class AutofillOverlayContentService implements AutofillOverlayContentServ
...
       return submitButtonElement;
     }
 
     // If the submit button is not a traditional button element, check for an anchor element that contains submission keywords.
-    const submitAnchorElement = await this.querySubmitButtonElement(element, "a", (node: Node) =>
-      nodeIsAnchorElement(node),
+    const submitAnchorElement = await this.querySubmitButtonElement(
+      element,
+      "a",
+      (element: Element) => elementIsAnchorElement(element),
```
audreyality: ⛏️  I think you can eliminate the type information here; it's inferred by `querySubmitButtonElement`.

```suggestion
      (element) => elementIsAnchorElement(element),
```

📝 The same is true for lines 602 and 593.

audreyality (commented):
Sorry for taking so long to get back to you. I had to go heads down on some work, and lost track of how much time had passed. 😓

This is looking pretty good! There's just a little cleanup needed.

apps/browser/src/autofill/services/autofill-overlay-content.service.ts
datdenkikniet: Fixed.

I've also repeated `(element: Element)` in all callers of `query`, but I guess those can also be inferred? Let me know if I should fix those as well.

apps/browser/src/autofill/services/collect-autofill-content.service.ts
datdenkikniet: Updated

apps/browser/src/autofill/services/dom-query.service.ts
datdenkikniet: I've rewritten this as a do-while loop (with the commit message explaining the reasoning). A bit larger than just your suggestion, but hopefully still OK.

datdenkikniet (commented):
Thank you for taking the time to take another look. No worries, there's always _something_ going on so I wasn't just twiddling my thumbs in the mean time ;)

I've added additional commits to address the PR feedback. I can amend earlier commits if that is preferred, but now at least it's obvious what I've changed.

audreyality (commented):
📝 There's a code formatting error reported. It could be the result of the merge I did to run the build. You can fix it by running `npm run prettier` in the root folder.

⛏️ You can fix the other explicit `Element` types if you'd like. It won't block the merge.

datdenkikniet:
Fixed the formatting.

> ⛏️ You can fix the other explicit Element types if you'd like. It won't block the merge.

I will skip over this for now, in that case :)

audreyality (approved):

audreyality:
Thanks for your patch @datdenkikniet! It just passed QA!

PR #21500, opened 2026-06-25
[PM-33362] Autotype Edit Cipher Initial Changes

coltonhurst:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-33362

## 📔 Objective

This PR implements the following initial changes for editing a cipher to pair an application for use with Autotype:
- `+ website` becomes `+ website or app`
- When clicking the above changed button, a dropdown appears, allowing the user to select `website` or `app`
  - If `website` is selected, everything works as is
  - If `app` is selected, the text `desktopapp://` is appended which identifies this uri as a desktop app. This will be expanded on in future pr's (and changed to become a select instead of a text input field), and the schema is subject to change while the v1 is still being ironed out.
- Change the title text for the field based on if it's a website or an app
- Feature flag the change, and make sure it is only for windows desktop (initially)
- UI tests added with the `// Autotype App Tests` comment so they can easily be identified and updated later

There will be more PR's for this ticket, the PR's are being broken up into logical chunks to avoid having one massive PR that accomplishes many things.

## 📸 Screenshots

The dropdown:

Dropdown

The "App" label working, retaining the incrementing numbers:

New App

## 🧪 Testing

Locally I have verified:
- In the dev environment these changes are active and working, only on Windows (not active on macOS)
- In the prod environment these changes are not active, as they are behind the autotype GA FF

Discussed with QA- no QA required for each PR until all PR's for this ticket are in, as the changes are behind a FF.

### DIFF
--- a/libs/vault/src/cipher-form/components/autofill-options/uri-option.component.ts
+++ b/libs/vault/src/cipher-form/components/autofill-options/uri-option.component.ts
@@ -135,6 +135,15 @@ export class UriOptionComponent implements ControlValueAccessor {
   // eslint-disable-next-line @angular-eslint/prefer-signals
   @Input({ required: true }) index: number;
 
+  /**
+   * When true, URIs prefixed with the desktop app scheme will display the "App (URI)" label
+   * instead of "Website (URI)". Should only be true when the WindowsDesktopAutotypeGA feature
+   * flag is enabled.
+   */
+  // FIXME(https://bitwarden.atlassian.net/browse/CL-903): Migrate to Signals
+  // eslint-disable-next-line @angular-eslint/prefer-signals
+  @Input() showAppLabel: boolean = false;
+
   // FIXME(https://bitwarden.atlassian.net/browse/CL-903): Migrate to Signals
   // eslint-disable-next-line @angular-eslint/prefer-output-emitter-ref
   @Output()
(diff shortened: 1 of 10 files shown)


libs/vault/src/cipher-form/components/autofill-options/uri-option.component.ts
```diff
@@ -133,6 +135,15 @@ export class UriOptionComponent implements ControlValueAccessor {
...
   @Input({ required: true }) index: number;
 
+  /**
+   * When true, URIs prefixed with the desktop app scheme will display the "App (URI)" label
+   * instead of "Website (URI)". Should only be true when the WindowsDesktopAutotypeGA feature
+   * flag is enabled.
+   */
+  // FIXME(https://bitwarden.atlassian.net/browse/CL-903): Migrate to Signals
+  // eslint-disable-next-line @angular-eslint/prefer-signals
+  @Input() showAppLabel: boolean = false;
```
nick-livefront, line 145: ⛏️ (non-blocking): This follows the current state of the component, but adding signals will cut down on the amount of things that we need to change later on.

nick-livefront (dismissed):
👏

libs/vault/src/cipher-form/components/autofill-options/uri-option.component.ts
coltonhurst: @nick-livefront yeah, I opted to keep it the same to match the file. I'm happy to create a follow up PR to update this 👍

Thank you!

Edit: PR is here: https://github.com/bitwarden/clients/pull/21638

coltonhurst:
@nick-livefront I needed to update the storybook file for the cipher form component, did so in: https://github.com/bitwarden/clients/pull/21500/changes/861c10beca3fcd9fc190d2987b1ac33a83beec5e

The failing CI is due to a RUSTSEC advisory, unrelated to this PR.

nick-livefront (approved):

PR #21505, opened 2026-06-26
[PM-39349]] - make ssh key cipher fields readonly

jaasen-livefront:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-39439

## 📔 Objective

SSH key cipher fields were using Angular's `FormGroup.disable()` to prevent user input. Disabled form controls render with low-contrast text in dark mode, making the field values nearly unreadable.

This PR changes the SSH key fields (`privateKey`, `publicKey`, `keyFingerprint`) from disabled to `readonly`. The `readonly` HTML attribute prevents user input without altering the field's visual styling, so the values remain legible in both light and dark mode.

**Changes:**
- Removed `sshKeyForm.disable()` and the `formStatusChange$` re-disable subscription from `ngOnInit` — the form stays enabled so values flow correctly through `valueChanges` and programmatic updates (e.g. import from clipboard) continue to work
- Added `readonly` attribute to all three SSH key inputs in the template
- Removed the now-unused `disabled` input, `DestroyRef`, and `inject` from the component
- Removed the `[disabled]` binding from the parent `cipher-form.component.html`
- Updated tests to remove assertions on the (now-removed) disabled behavior

## 📸 Screenshots

Screenshot 2026-06-25 at 5 14 03 PM

### DIFF
--- a/libs/vault/src/cipher-form/components/cipher-form.component.html
+++ b/libs/vault/src/cipher-form/components/cipher-form.component.html
@@ -28,10 +28,7 @@
     }
 
     @if (config.cipherType === CipherType.SshKey) {
-      <vault-sshkey-section
-        [disabled]="config.mode === 'partial-edit'"
-        [originalCipherView]="originalCipherView"
-      ></vault-sshkey-section>
+      <vault-sshkey-section [originalCipherView]="originalCipherView"></vault-sshkey-section>
     }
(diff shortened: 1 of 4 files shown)


jengstrom-bw (approved):
LGTM!

PR #21511, opened 2026-06-26
Fix/autofill toplayer listener dedup

blackwood:
## 🎟️ Tracking

[PM-40125](https://bitwarden.atlassian.net/browse/PM-40125)

## 📔 Objective

Fixes toggle-listener accumulation on top-layer autofill candidates (dialogs/popovers).

- Dedup listener attachment: `setupTopLayerCandidateListener` re-attached a toggle listener every time an element was re-observed, leaking one handler per mutation pass. Now gated by a `topLayerListenedElements` `WeakSet` so each element is listened to at most once; `WeakSet` lets detached nodes GC.
- `continue` instead of `return`: when clearing cached field handlers, a missing memo for one event bailed out of the whole loop, leaving the remaining events' listeners un-removed. Switched to continue so each event is handled independently.

[PM-40125]: https://bitwarden.atlassian.net/browse/PM-40125?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


jprusik (approved):
nice, good catches on both counts

blackwood:
Rust lint issues are pollution from main (again).

PR #21514, opened 2026-06-26
Add claude snap permissions rule

quexten:
## 🎟️ Tracking

n/a

## 📔 Objective

Adds a rule reminding the author that we have to obtain permissions for new plugs. This should happen *before* the PR merges.


.claude/rules/snap-permissions.md
```diff
@@ -0,0 +1,29 @@
...
+
+# Snap Permissions (Plugs)
+
+`snapcraft.yaml` declares the [plugs](https://snapcraft.io/docs/supported-interfaces) (interfaces)
+that the snap requests at runtime. Snap uses strict confinement, so every plug grants additional
+access to the host system.
+
+## Adding new plugs
+
+When a PR adds a new entry to the `plugs:` list, the code reviewer **must** confirm one of the
```
theMickster, line 22: ♻️ / 🎨 The rules metadata outlines that we are adding constraints for both YAML and JSON files. However, we only describe the YAML list and not the JSON array.  

I recommend that we generalize the skill body to cover both declaration sites:

  ```markdown
  Snap plugs are declared in two places:
  - `apps/cli/stores/snap/snapcraft.yaml` — under `apps.<app>.plugs:` (YAML list)
  - `apps/desktop/electron-builder.json` — under `snap.plugs` (JSON array)
  ```

.claude/rules/snap-permissions.md
```diff
@@ -0,0 +1,29 @@
+---
+paths:
+  - "apps/cli/stores/snap/snapcraft.yaml"
+  - "apps/desktop/electron-builder.json"
+  - "apps/desktop/electron-builder.beta.json"
```
theMickster: ♻️  A local multi agent code review caught the following finding. I believe that is it correct and we should consider removing the constraint until such time as we add a snap section to said file. 

<details>
<summary> `electron-builder.beta.json` has no snap configuration — path reference is a dead entry </summary>

<br/>`.claude/rules/snap-permissions.md:5`
<br/>**Caught by:** Architecture agent


`electron-builder.beta.json` contains no `snap` block and no `plugs` declaration. The beta builder targets only Windows formats (`portable`, `nsis-web`, `appx`) — the entire `snap` object present in `electron-builder.json` is absent from the beta variant. Listing it in `paths:` means the snap-permissions rule fires whenever the beta builder config is edited, even though snap plug guidance has no bearing on that file.

  **Suggested fix:** Remove `"apps/desktop/electron-builder.beta.json"` from the `paths:` frontmatter:

  ```yaml
  paths:
    - "apps/cli/stores/snap/snapcraft.yaml"
    - "apps/desktop/electron-builder.json"
  ```

  Re-add it only if a `snap` section is ever introduced in `electron-builder.beta.json`.

  </details>

.claude/rules/snap-permissions.md
```diff
@@ -0,0 +1,29 @@
...
+---
+
+# Snap Permissions (Plugs)
+
+`snapcraft.yaml` declares the [plugs](https://snapcraft.io/docs/supported-interfaces) (interfaces)
```
theMickster: ⚠️/♻️ We never mention the JSON file that's constrained in the rule frontmatter.  Please consider updating the opening sentence to include both files. 

  > `` `snapcraft.yaml` and the `snap.plugs` section of `electron-builder.json` declare the [plugs](https://snapcraft.io/docs/supported-interfaces) that the snap requests at runtime. ``

.claude/rules/snap-permissions.md
theMickster, line 22: ❓ I think snapcraft allows for `plugs:` at both the top-level and individual app level (like we currently have at the apps.bw in `apps/cli/stores/snap/snapcraft.yaml`).  Should we consider also flagging if a plugs list (or in the case of JSON an array) is added to the top-level configuration?

theMickster (changes requested):
I think we need to improve on lines 10 & 16.  The comment on the frontmatter is open for discussion/closure if there are forthcoming plans to add snap rules to the `apps/desktop/electron-builder.beta.json`

.claude/rules/snap-permissions.md
quexten: What does top level mean here? We have two snap manifests, one for the CLI app one for the desktop app (via the electron builder json). These are separate apps.

.claude/rules/snap-permissions.md
quexten: Makes sense. My intent was that possibly the person adding the snap build to the beta would not know about this claude rule. But maybe that's being too defensive.

.claude/rules/snap-permissions.md
quexten: Adressed

theMickster (approved):

.claude/rules/snap-permissions.md
theMickster: Sorry, wasn't clear. I was trying to think about adding a case for Claude to raise a finding on a code review if it detected that the `plugs:` list was moved in the file from this

```yaml
apps:
  bw:
    command: bw
    environment:
      XDG_CONFIG_HOME: $SNAP_USER_DATA
    plugs: [network, network-bind, desktop, home, removable-media]

```

to this

```yaml

apps:
  bw:
    command: bw
    environment:
      XDG_CONFIG_HOME: $SNAP_USER_DATA
plugs: [network, network-bind, desktop, home, removable-media]
```
Because I _think_ that would still technically be valid, but not a desired change.
I don't think that would be likely to happen and shouldn't be a concern at this point.

PR #21518, opened 2026-06-26
[PM-39387] fix: Auto-sync premium status after returning from Stripe checkout

cyprain-okeke:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-39387

## 📔 Objective

Premium status was not reflected in the client after a user completed payment in an external Stripe checkout flow. The server sends a one-time `PremiumStatusChanged` push, but the client misses it while it is in the background during checkout — so the user's premium entitlements don't appear until a later full sync.

This PR recovers the missed push by syncing when the client returns from checkout:

- Adds `PremiumCheckoutPendingService` (abstraction + `DefaultPremiumCheckoutPendingService`) backed by a memory-scoped, account-scoped state flag (`clearOn: ["logout"]`).
- `PremiumUpgradeDialogComponent` marks checkout as pending when it launches the external Stripe checkout URI.
- Desktop consumes the flag on `windowIsFocused` and browser popup consumes it on init, triggering `fullSync(true)` to pull the updated premium status.
- Registered the service in `jslib-services.module.ts` (Angular clients) and `service-container.ts` (CLI).
- Adds unit tests for the new service and the browser/desktop/dialog wiring.


sbrown-livefront (dismissed):
✅ Looks good. Great thorough job ✨ 🧹 . Have these been manually tested for both?

dani-garcia (approved):

sbrown-livefront (approved):

PR #21523, opened 2026-06-26
PM-16780 SecureSafe export format

bmbitwarden:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-16780

## 📔 Objective

SecureSafe has changed their export format - no longer compatible with Bitwarden


mcamirault (approved):
LGTM

PR #21524, opened 2026-06-26
[PM-32384] Update copy new to create/add browser

jengstrom-bw:
## 🎟️ Tracking

[Jira](https://bitwarden.atlassian.net/browse/PM-32384)

## 📔 Objective

Update copy from `New ...` to `Add ...` or `Create ...`

## 📸 Screenshots

https://github.com/user-attachments/assets/0499c99c-be37-464e-b0bc-f18a3c94de07


harr1424 (dismissed):
Tools owned changes look good.

apps/browser/src/vault/popup/components/vault/add-edit/add-edit.component.ts
```diff
@@ -520,24 +525,60 @@ export class AddEditComponent implements OnInit, OnDestroy {
...
-        : "newItemHeaderDriversLicense",
-      [CipherType.Passport]: isEditMode ? "editItemHeaderPassport" : "newItemHeaderPassport",
+        : this.btnTextAddCreateFeatureFlag()
+          ? "addItemHeaderDriversLicense"
+          : "newItemHeaderDriversLicense",
+      [CipherType.Passport]: isEditMode
+        ? "editItemHeaderPassport"
+        : this.btnTextAddCreateFeatureFlag()
+          ? "addItemHeaderPassport"
+          : "newItemHeaderPassport",
```
nikwithak, line 581: 🎨 (optional): Nested ternary operators can get a  little bit confusing.. since you're doing the same pattern for all of these with the same feature flags, you might consider a helper function that selects the appropriate key

nikwithak (approved):
One suggestion to reduce the use of nested ternaries - your call

apps/browser/src/vault/popup/components/vault/add-edit/add-edit.component.ts
jengstrom-bw: I think it will be much easier to unwind the feature flag with these nested layers, should be able to be clean line deletions.

harr1424 (approved):

PR #21529, opened 2026-06-26
[PM-38909] [BEEEP] Add QA cloud builds to desktop and browser

kdenney:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-38909

## 📔 Objective

Adds QA cloud-region builds for both the desktop and browser extensions so QA can test against the internal QA servers (USQA, USQA2, EUQA) without needing to manually enter QA urls and with having the app actually consider the server as a cloud-hosted server (important to test certain billing setups).

**Desktop**

- Added `apps/desktop/config/qa.json` with the three QA regions
- Fixed `webpack.base.js` to select the environment config by `ENV` (falling back to `NODE_ENV`), decoupling which config file loads from the webpack production/development mode
- Added `build:qa`, `dist:mac:qa`, and `dist:win:qa` npm scripts, plus matching `qa` Nx configurations in `project.json`

**Browser**

- Added `apps/browser/config/qa.json` with the same three QA regions
- Fixed `webpack.base.js` to separate `ENV` (config selector) from `NODE_ENV` (webpack mode); `ADDITIONAL_REGIONS` and `NODE_ENV` are now injected via `EnvironmentPlugin`
- Fixed `popup/main.ts` (OSS and commercial) to gate Angular prod mode on `NODE_ENV` instead of `ENV`, so `ENV=qa` builds still run in Angular production mode
- Added `build:qa` / `dist:qa` scripts (and `bit` twins) for chrome, edge, firefox, and opera, plus `package:qa:dev:chrome` / `package:bit:qa:dev:chrome` for side-loadable QA dev zips, plus matching `-qa` Nx configurations in `project.json`

**CI**

- Folded the QA builds into the existing matrices instead of standalone jobs: browser `build` gains a `build_type` (prod/qa) dimension, and desktop `windows` / `macos-package-github` gain the same. QA artifacts are `-QA` named; the desktop prod leg still builds without signing secrets while the QA leg only runs when secrets are present

## 📸 Screenshots

https://github.com/user-attachments/assets/97757e08-a6c7-434c-a8f9-9bc4184e3149

https://github.com/user-attachments/assets/f07f7206-46d2-4d78-81b2-a2eba4eb396d


.github/workflows/build-browser.yml
```diff
@@ -374,6 +374,64 @@ jobs:
           if-no-files-found: error
 
 
+  # Standalone job (not part of the build matrix): builds the dev-signed, QA-region Chrome
+  # extension for QA side-loading. Consumes the commercial build-source relay (the same sanitized
+  # source snapshot the production commercial build uses) for consistency with the other build jobs.
+  build-qa:
```
dereknance: Including QA builds as part of the `build` matrix above, perhaps under a new `release_type` key, seems to make a lot of sense to me. What was your intent with singling-out commercial Chrome for QA in this job, as opposed to any other configuration, as well as defining a standalone job?

The GitHub Actions docs mention a way to [exclude certain matrix configurations](https://docs.github.com/en/actions/how-tos/write-workflows/choose-what-workflows-do/run-job-variations#excluding-matrix-configurations), so you could use that to exclude the OSS Chrome and (OSS/Commercial) Edge, Firefox[-MV3], and Opera-MV3 in case any of those are undesirable as QA builds.

.github/workflows/build-desktop.yml
```diff
@@ -736,6 +736,121 @@ jobs:
           path: apps/desktop/dist/nsis-web/${{ needs.setup.outputs.release_channel }}.yml
           if-no-files-found: error
 
+  windows-qa:
```
dereknance: Similar to [this comment](https://github.com/bitwarden/clients/pull/21529/changes#r3507679798), adding a distinct build job for QA, instead of adding a build matrix to the `windows` job and gating relevant steps behind an `if:`, results in quite a bit of duplication.

The same goes for the `macos-qa` job below too.

apps/browser/package.json
dereknance, line 1: It is Platform's intent to migrate builds to use Nx, so it would be good to add these new builds to [`apps/browser/project.json`](https://github.com/bitwarden/clients/blob/5ac81980c6fab44cd9c83fffe1a4c00e4b254623/apps/browser/project.json).

apps/desktop/package.json
dereknance, line 1: [This comment](https://github.com/bitwarden/clients/pull/21529/changes#r3507772145) also applies here.

dereknance (commented):
I think this is a really cool idea! I do have a couple questions/concerns, mostly around job duplication and Nx-equivalents of the new build scripts.

apps/browser/package.json
kdenney: Good call; added in latest push for all the new qa builds I added in the npm scripts.

.github/workflows/build-browser.yml
kdenney: Thank you for reviewing and leaving feedback!

The singling out of Chrome for QA was unintentionally left behind from when I was testing and wanted to limit scope for the time being; I meant to come back and expand it but forgot. Great catch!

I have pushed new changes that removes all the standalone builds and incorporates the qa env builds into the existing matrix under a new `build_type` key. Let me know if you see anything I missed or if you want anything else changed. Appreciate it! 🙌

.github/workflows/build-desktop.yml
kdenney: Addressed in my reply above; fixed in latest push.

apps/desktop/package.json
kdenney: Addressed in my reply above; fixed in latest push.

.github/workflows/build-browser.yml
```diff
@@ -225,52 +225,62 @@ jobs:
...
             type: "commercial"
+        build_type:
+          - name: "prod"
+            readable: "Production"
+            command_qa_segment: ""
+            archive_qa_segment: ""
+          - name: "qa"
+            readable: "QA"
+            command_qa_segment: "qa:"
+            archive_qa_segment: "-QA"
```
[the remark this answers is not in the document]
kdenney: Yes QA should build on every PR

.github/workflows/build-desktop.yml
```diff
@@ -675,92 +689,92 @@ jobs:
         if: ${{ needs.setup.outputs.has_secrets == 'true' }}
         uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
         with:
-          name: Bitwarden-Installer-${{ env._PACKAGE_VERSION }}.exe
+          name: Bitwarden-Installer${{ matrix.build_type.installer_qa_segment }}-${{ env._PACKAGE_VERSION }}.exe
```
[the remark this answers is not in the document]
kdenney: Good catch, and confirmed. The `-QA` installer was an `nsis-web` downloader stub, and its `.nsis.7z` payload was never uploaded on the QA leg, so it had nothing to install. Even if we did upload the payload, the stub resolves it against the production publish URL baked into `electron-builder.json`, so a QA installer would still either fail to download or pull the production build.

Rather than wire up the web installer for QA, I went with your second suggestion and shipped a self-contained artifact instead. The Windows QA leg now uploads `Bitwarden-Portable-QA-<version>.exe` (with a `-QA` suffix to avoid an artifact-name collision with the prod portable), and the installer stub upload is now restricted to prod only, so the QA leg no longer produces the broken installer. This mirrors what the macOS QA leg already does with its self-contained `.dmg`.

Fixed in 278805cd5c.

dereknance (dismissed):
LGTM!

apps/desktop/package.json
```diff
@@ -65,9 +66,11 @@
     "dist:lin": "npm run build && npm run pack:lin",
     "dist:lin:arm64": "npm run build && npm run pack:lin:arm64",
     "dist:mac": "npm run build && npm run pack:mac",
+    "dist:mac:qa": "npm run build:qa && npm run pack:mac",
```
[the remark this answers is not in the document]
kdenney: Confirmed, that's right. The `-QA` was only on the uploaded artifact name; the app carried prod identity, so it replaced `/Applications/Bitwarden.app` and shared prod's userData.

Fixed by adding `apps/desktop/electron-builder.qa.json`, which extends the base config and overrides identity only (`com.bitwarden.desktop.qa`, `Bitwarden QA`, `bitwarden-qa`), mirroring beta's side-by-side install with its own state. The QA pack scripts and CI legs now build with it.

Notes: Windows narrows to portable via `-c.win.target=portable` in the pack script (extends concatenates target arrays, so it can't go in the JSON); dmg/portable `artifactName` are pinned so CI upload paths are unchanged. macOS signing is kept and works under the `.qa` appId like beta's `.beta`.

apps/desktop/electron-builder.qa.json
```diff
@@ -0,0 +1,15 @@
+{
+  "$schema": "https://raw.githubusercontent.com/electron-userland/electron-builder/master/packages/app-builder-lib/scheme.json",
+  "extends": "electron-builder.json",
+  "appId": "com.bitwarden.desktop.qa",
+  "productName": "Bitwarden QA",
```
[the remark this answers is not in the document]
kdenney: Fixed, qa should not auto update

.github/workflows/build-desktop.yml
```diff
@@ -602,19 +616,20 @@ jobs:
...
-        if: steps.cache.outputs.cache-hit != 'true'
+        if: ${{ (matrix.build_type.name == 'prod' || needs.setup.outputs.has_secrets == 'true') && steps.cache.outputs.cache-hit != 'true' }}
         working-directory: apps/desktop/desktop_native
         env:
           MODE: ${{ github.event_name == 'workflow_call' && '--release' || '' }}
         run: node build.js cross-platform "$env:MODE"
 
       - name: Build
-        run: npm run build
+        if: ${{ matrix.build_type.name == 'prod' || needs.setup.outputs.has_secrets == 'true' }}
```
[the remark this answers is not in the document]
kdenney: This suggestion was wrong. Github doesn't allow "if" at this level.

kdenney:
@dereknance Fixed a few issues found by the bots, need another stamp. 🙌

dereknance (dismissed):

apps/desktop/config/qa.json
```diff
@@ -0,0 +1,29 @@
+{
+  "additionalRegions": [
+    {
+      "key": "USQA",
+      "domain": "qa.bitwarden.pw",
+      "urls": {
+        "webVault": "https://vault.qa.bitwarden.pw",
+        "base": "https://vault.qa.bitwarden.pw"
```
mgaineybw: Flagging this before I start the QA pass. I had Claude Code diff the new region configs against what the QA environments actually serve, and it found one gap — the regions omit `notifications`, which the URL-derivation fallback cannot cover.

Each region in `apps/{desktop,browser}/config/qa.json` sets only `base` and `webVault`. Every other service URL is then derived by `UrlEnvironment.getUrl()` as `base + "/<suffix>"`, so `getNotificationsUrl()` (`libs/common/src/platform/services/default-environment.service.ts:429`) resolves to `https://vault.qa.bitwarden.pw/notifications`, and SignalR appends `/hub`.

Probing all three environments (2026-09-17):

| URL | Status |
| --- | --- |
| `vault.<env>.bitwarden.pw/api/config` | 200 |
| `vault.<env>.bitwarden.pw/identity/.well-known/openid-configuration` | 200 |
| `vault.<env>.bitwarden.pw/icons/...` | 200 |
| `vault.<env>.bitwarden.pw/events/collect` | 401 (routed, needs auth) |
| **`vault.<env>.bitwarden.pw/notifications/hub`** | **404** |
| `notifications.<env>.bitwarden.pw/hub` | 401 (alive, needs token) |

...identical on `qa`, `qa2`, and `euqa`. Path-based routing off the vault host covers api/identity/icons/events, but notifications is only served on its own subdomain — which matches what `/api/config` advertises (`environment.notifications: https://notifications.qa.bitwarden.pw`).

Expected impact:

- **Desktop** — the derived URL is not `DISABLED_NOTIFICATIONS_URL`, so `default-server-notifications.service.ts:100` proceeds and SignalR retries a 404 indefinitely. Push-driven auto-sync (cipher changes, `PremiumStatusChanged`, remote logout) silently stops working. The console signature is a misleading "WebSocket failed to connect… proxy blocking WebSockets".
- **Both platforms** — login with device / auth-request approval uses the same URL via `anonymous-hub.service.ts:36`, so that flow should fail outright.
- **Browser MV3 may mask it** — QA reports `pushTechnology: 1` with a VAPID key, so Chrome likely takes the WebPush path and syncs fine. A passing browser result would not clear desktop.

Scope note: this only affects the `ENV=qa` builds — prod builds compile an empty `additionalRegions`, confirmed against the prod artifact from the same run — so it's a QA-tooling defect, not a production one.

The fix looks like one line per region, in both config files:

```json
"urls": {
  "webVault": "https://vault.qa.bitwarden.pw",
  "base": "https://vault.qa.bitwarden.pw",
  "notifications": "https://notifications.qa.bitwarden.pw"
}
```

For reference, `getScimUrl()` falls back to `vault.<env>/scim/v2`, which also 404s — leaving it as-is seems fine since neither client calls SCIM, but noting it in case you'd rather set all of them explicitly the way `PRODUCTION_REGIONS` does.

(Library line references verified against `main` @ `79ef6575`; endpoint probes are live as of today.)

mgaineybw (commented):
Small comment on notifications settings for QA envs.

apps/desktop/config/qa.json
kdenney: Nice catch! I've added the notification service urls to the qa config files in my latest push. @dereknance I need another stamp. 😄

mgaineybw:
I think I'm good to approve this, then I can get started on the QA parts. Claude pointed out a couple extra questions though, when you have a minute. 

❓: QA build's macOS entitlements and provisioning profile still assert the production bundle ID; asking whether biometric unlock has been verified on an installed QA build
  - `apps/desktop/electron-builder.qa.json:4`
❓: `bitwarden://` protocol registration is shared with the production install; assuming intended, flagging as a documented QA gotcha
  - `apps/desktop/electron-builder.qa.json:3`

mgaineybw (approved):
Per Claude Review: Reviewed the QA cloud-region build support for desktop and browser at `21442bc9c`, covering the region configs, the `ENV`/`NODE_ENV` webpack separation, the npm and Nx build targets, both CI workflow matrices, and the QA electron-builder config. The previously reported missing `notifications` URL is fixed correctly across all six region blocks, and the earlier round of build-matrix findings all check out as genuinely resolved rather than only marked resolved. Two questions left inline, neither blocking — QA will validate the macOS side-by-side behavior during the test pass.

PR #21530, opened 2026-06-26
introduce `AutofillLifecycleService`

audreyality:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-39587

## 📔 Objective

* Replace autofiller.ts “collectPageDetails” messages with “pageTransitionDetected” messages. "collectPageDetails" messages are now only emitted by the extension.
* Capture the “pageTransitionDetected” message in the extension as a “pending message”. There should be only 1 pending message per (tab, frame). Only keep the most recent detection.
* When enabling monitoring on a tab (e.g. transitioning from a logged out state to a logged in state), process any pending “pageTransitionDetection” messages related to that tab.

> [!NOTE]
> While this is mostly cleaning up tech debt, it has a nice side-effect of fixing a race condition that could cause autofill-on-load to take its 1.5s fallback branch as a result of script injection latency.

## 🦖 System Evolution

This PR introduces the `AutofillLifecycleService` to manage autofill lifecycle events, and insofar as is possible keep `AutofillService` focused on autofill heuristics.

Check out [lifecycle.design.md](https://github.com/bitwarden/clients/blob/autofill/pm-39587/autofill-lifecycle-refinement/apps/browser/src/autofill/lifecycle.design.md) to learn more than you ever wanted to know about how the whole thing works.


apps/browser/src/autofill/services/autofill-lifecycle.service.ts
```diff
@@ -0,0 +1,379 @@
...
+          const key = this.monitorFrameKey(event.tabId, event.frameId);
+          if (event.active) {
+            state.set(key, true);
+          } else {
+            state.delete(key);
+          }
+          return state;
+        }, new Map<string, boolean>()),
+      )
+      .subscribe(this.monitoringState);
```
audreyality, line 116: @blackwood - These are the `scan()`s I thought you might enjoy!

bensbits91 (dismissed):
Looks good Audrey, no notes :)

bensbits91 (approved):

audreyality:
📝 The lint failure is due to `cargo-deny` in the rust code. It's unrelated to this PR.

PR #21561, opened 2026-06-29
[PM-37988] Add SessionTimeoutPolicyV2 component usage for new policy drawers

JaredScar:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-37988

## 📔 Objective
Implement new UI design for Session Timeout Policy using V2 component

## 📸 Screenshots

image

### DIFF
--- a/bitwarden_license/bit-web/src/app/key-management/policies/session-timeout-v2.component.html
+++ b/bitwarden_license/bit-web/src/app/key-management/policies/session-timeout-v2.component.html
@@ -1,5 +1,5 @@
 <bit-callout title="{{ 'prerequisite' | i18n }}">
-  {{ "requireSsoPolicyReq" | i18n }}
+  {{ "requireSsoPolicyReqV2" | i18n }}
 </bit-callout>
 
 <p>{{ "sessionTimeoutPolicyDescription" | i18n }}</p>
--- a/bitwarden_license/bit-web/src/app/key-management/policies/session-timeout-v2.component.html
+++ b/bitwarden_license/bit-web/src/app/key-management/policies/session-timeout-v2.component.html
@@ -1,0 +1,41 @@
+<bit-callout title="{{ 'prerequisite' | i18n }}">
+  {{ "requireSsoPolicyReq" | i18n }}
+</bit-callout>
+
+<p>{{ "sessionTimeoutPolicyDescription" | i18n }}</p>
+
+<bit-form-control>
+  <bit-switch [formControl]="enabled" id="enabled"></bit-switch>
+  <bit-label>{{ "enablePolicy" | i18n }}</bit-label>
+</bit-form-control>
+
+<div [formGroup]="data">
+  <div class="tw-grid tw-grid-cols-12 tw-gap-4">
+    <bit-form-field class="tw-col-span-12 !tw-mb-0">
+      <bit-label>{{ "maximumAllowedTimeout" | i18n }}</bit-label>
+      <bit-select formControlName="type">
+        @for (option of typeOptions; track option.value) {
+          <bit-option [value]="option.value" [label]="option.name"></bit-option>
+        }
+      </bit-select>
+    </bit-form-field>
+    @if (data.value.type === "custom") {
+      <bit-form-field class="tw-col-span-6 tw-self-start !tw-mb-0">
+        <bit-label>{{ "hours" | i18n }}</bit-label>
+        <input bitInput type="number" min="0" formControlName="hours" />
+      </bit-form-field>
+      <bit-form-field class="tw-col-span-6 tw-self-start !tw-mb-0">
+        <bit-label>{{ "minutes" | i18n }}</bit-label>
+        <input bitInput type="number" min="0" max="59" formControlName="minutes" />
+      </bit-form-field>
+    }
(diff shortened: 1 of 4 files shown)


bitwarden_license/bit-web/src/app/key-management/policies/session-timeout-v2.component.ts
```diff
@@ -0,0 +1,220 @@
...
+  imports: [
+    CalloutComponent,
+    FormControlModule,
+    FormFieldModule,
+    ReactiveFormsModule,
+    SelectModule,
+    SwitchComponent,
+    I18nPipe,
+  ],
+  changeDetection: ChangeDetectionStrategy.OnPush,
```
mzieniukbw, line 77: ⚠️  This will likely break few things. This component is not ready for on push change detection. I suggest to revert it, KM will handle modernisation in the future.

bitwarden_license/bit-web/src/app/key-management/policies/session-timeout-v2.component.html
```diff
@@ -0,0 +1,41 @@
+<bit-callout title="{{ 'prerequisite' | i18n }}">
+  {{ "requireSsoPolicyReq" | i18n }}
+</bit-callout>
+
+<p>{{ "sessionTimeoutPolicyDescription" | i18n }}</p>
+
+<bit-form-control>
+  <bit-switch [formControl]="enabled" id="enabled"></bit-switch>
+  <bit-label>{{ "enablePolicy" | i18n }}</bit-label>
+</bit-form-control>
```
mzieniukbw, line 10: Instead of creating v2 component from scratch, which is no different than "v1", just gate these behind the policy drawers feature flag. 
```suggestion
@if (drawerEnabled()) {
<bit-form-control>
  <bit-switch [formControl]="enabled" id="enabled"></bit-switch>
  <bit-label>{{ "enablePolicy" | i18n }}</bit-label>
</bit-form-control>
} @else {
<p>{{ "sessionTimeoutPolicyDescription" | i18n }}</p>

<bit-form-control>
  <input type="checkbox" id="enabled" bitCheckbox [formControl]="enabled" />
  <bit-label>{{ "turnOn" | i18n }}</bit-label>
</bit-form-control>
}
```
That way there is no need to duplicate the .html and .ts, with most of the .ts file unchanged.

bitwarden_license/bit-web/src/app/key-management/policies/session-timeout-v2.component.ts
```diff
@@ -0,0 +1,220 @@
...
+import { SessionTimeoutConfirmationNeverComponent } from "./session-timeout-confirmation-never.component";
+
+export class SessionTimeoutPolicyV2 extends BasePolicyEditDefinition {
+  name = "sessionTimeoutPolicyTitle";
+  description = "sessionTimeoutPolicyDescription";
+  type = PolicyType.MaximumVaultTimeout;
+  category = PolicyCategory.Authentication;
+  priority = 70;
+  component = SessionTimeoutPolicyV2Component;
+  showDescription = false;
```
mzieniukbw, line 54: See my other suggestion in .html file, which will make this redundant (default `showDescription`).

bitwarden_license/bit-web/src/app/key-management/policies/session-timeout-v2.component.ts
```diff
@@ -0,0 +1,220 @@
...
+  showEnabledBadge = true;
```
mzieniukbw, line 55: As i understand this is no-op for this policy, since it is not multi step policy. See my other suggestion in .html file, which will make this entire change much smaller.

mzieniukbw (changes requested):
This PR could be much smaller.

bitwarden_license/bit-web/src/app/key-management/policies/session-timeout-v2.component.html
JaredScar: I believe we're doing this pattern for all of the tickets to keep consistency across all the policy changes for the Policy Drawers initiative. Whilst I agree it could be much simpler, sometimes components have more tech debt, so we don't want to change the original code of the component to the point where it can't be easily reverted via a feature flag. By using V2 components, we feature flag the route, so the original V1 components stay unchanged.

Parent ticket for reference: https://bitwarden.atlassian.net/browse/PM-34804

bitwarden_license/bit-web/src/app/key-management/policies/session-timeout-v2.component.ts
JaredScar: From my inspection, this should be fine with a `markForCheck` added after the `typeControl.setValue` on line 139. Let me know if you think otherwise?

r-tome (approved):
LGTM, I just wonder if we should create a specific test file for the V2 component

mzieniukbw (approved):

JaredScar:
> LGTM, I just wonder if we should create a specific test file for the V2 component

Only the text is changing. Functionality still stays the same besides the drawer aspect. The drawers do have spec test for drawer functionality too. Also made another ticket to make sure the spec tests are kept when we move V2 back to be the original component

PR #21563, opened 2026-06-29
[PM-33228] Add screen reader only to enhance a11y behavior

BryanCunningham:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-33228

## 📔 Objective

- Adds a SR only section for separated values wrapped in `<code>` tags so that the SR does not attempt to group them into words and announces punctuation chars correctly. Punctuation is generally skipped by SRs. Putting the chars in code blocks forces the SR to announce them more consistently. 
- SR will not announce `space` chars so we substitute them for the actual word 'space'


vleague2 (approved):

jengstrom-bw (approved):
Vault changes look good!

PR #21573, opened 2026-06-30
Auth/PM-39774 - SSO - Improve error for existing users without org membership

JaredSnider-Bitwarden:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-39774
Server PR: https://github.com/bitwarden/server/pull/7890

## 📔 Objective

Adds client-side handling for the new `OrgMembershipRequired` SSO redirect error code from the server, which fires when an existing Bitwarden user attempts SSO into an org where they have no `OrganizationUser` row at all (vs. `InviteAcceptanceRequired`, where they have an `Invited` row pending acceptance). Both rejection lanes share the same client recovery contract — match a stashed invite for the rejected org by `organizationId` + email, auto-progress to MP entry on match, or warn and stay at email entry on no-match — so they're routed through one handler.

A TODO is in place to revisit the stash-matching logic once the open-invite-link work lands; today the stash only ever holds direct (email) invites.

## 📸 Screenshots

Old behavior:

https://github.com/user-attachments/assets/aa18a252-e99c-4118-9813-15e58de9d82b

New behavior: 

https://github.com/user-attachments/assets/9520554c-1f29-4007-b7a4-877e44e8889f


Patrick-Pimentel-Bitwarden (approved):
Great work! Looks good.

PR #21578, opened 2026-06-30
[PM-39787] Promote exact hostname-match cipher to top of autofill list

AbdelftahZowail:
## 📔 Objective

Before this change, the browser extension's inline autofill menu and
context menu ordered matching ciphers purely by last-used recency. This
means a cipher saved for a specific subdomain — e.g.
`https://admin.example.com` — would almost always appear at the bottom of
the list when visiting that subdomain, because it is used less frequently
than the main domain's cipher. The user has to scroll past unrelated
entries to find the most relevant match.

This adds a hostname-awareness step: after the initial last-used sort,
`CipherViewLikeUtils.sortCiphersForUrl()` finds the first cipher whose
login URI matches the current page's exact hostname and moves it to
position 0. So the subdomain-specific cipher now appears first, and all
other matches follow in their last-used order.

### What changed

- **`libs/common/src/vault/utils/cipher-view-like-utils.ts`**: New static
  `sortCiphersForUrl<T>(ciphers, url)` — pure utility that promotes the
  exact-hostname match cipher to the front of the array.

- **3 call sites updated**:
  - `overlay.background.ts` — `getCipherViews` (login-only inline menu update)
    and `getAllCipherTypeViews` (full card/identity refresh)
  - `cipher-context-menu-handler.ts` — right-click context menu cipher list

- **Tests**: 5 unit tests in `cipher-view-like-utils.spec.ts` covering
  no-hostname, match-at-front, match-promoted, no-match, and no-login.

### Design notes

- Placed on `CipherViewLikeUtils` (not `CipherService`) because the method
  is a pure function with zero service-state dependencies — it only uses
  `Utils.getHostname` and `CipherViewLikeUtils.getLogin`, both already
  available in that module. This follows the existing pattern of static
  cipher utilities (`matchesUri`, `canLaunch`, `getUriHostname`).
- Mutates the array in place and returns it. Safe for current callers
  (all pass a fresh `.filter()`-ed array from `getAllDecryptedForUrl`).

### Behavior change

Before: autofill list ordered purely by last-used-then-name — a
subdomain-specific cipher typically appears at the bottom.

After: exact-hostname match promoted to position 0, then last-used-then-name
for the rest.

## 📸 Screenshots

No UI changes — ordering only.

### DIFF
--- a/apps/browser/src/autofill/background/overlay.background.spec.ts
+++ b/apps/browser/src/autofill/background/overlay.background.spec.ts
@@ -41,6 +41,7 @@ import { VaultSettingsService } from "@bitwarden/common/vault/abstractions/vault
 import { CipherRepromptType, CipherType } from "@bitwarden/common/vault/enums";
 import { CipherView } from "@bitwarden/common/vault/models/view/cipher.view";
 import { Fido2CredentialView } from "@bitwarden/common/vault/models/view/fido2-credential.view";
+import { CipherViewLikeUtils } from "@bitwarden/common/vault/utils/cipher-view-like-utils";
 import { CredentialGeneratorService } from "@bitwarden/generator-core";
 import { GeneratedCredential, GeneratorHistoryService } from "@bitwarden/generator-history";
 
@@ -187,6 +188,8 @@ describe("OverlayBackground", () => {
     cipherService = mock<CipherService>({
       getAllDecryptedForUrl: jest.fn().mockResolvedValue([]),
     });
+
+    jest.spyOn(CipherViewLikeUtils, "sortCiphersForUrl").mockImplementation((ciphers) => ciphers);
     enableNotificationAnimationMock$ = new BehaviorSubject(true);
     enableInlineMenuAnimationMock$ = new BehaviorSubject(true);
     autofillService = mock<AutofillService>();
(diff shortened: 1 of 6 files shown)


audreyality:
Hi @AbdelftahZowail, thanks for your interest in contributing to Bitwarden!

After reviewing this adjustment with the team, we won't be accepting this PR. If you'd like to limit your logins by hostname, you can instead change your [default match detection](https://bitwarden.com/help/uri-match-detection/#default-match-detection).

AbdelftahZowail:
@audreyality Appreciate the review. I think there's a meaningful difference here though.

Setting match detection to "Host" removes entries — I'd lose my main-domain logins on subdomains entirely. Sometimes I share credentials across subdomains, or I just want to see all my options.

This PR doesn't remove anything. It just nudges the cipher whose hostname actually matches to the top. The very next entry in the list is the most-recently-used one, so nothing is lost — the old behavior is literally position #2 now. The downside is zero; the upside is that on pages with a subdomain-specific login, you stop having to dig for it.

I've been running this change locally and it's a genuine quality-of-life improvement. The "Host" workaround would force me to choose between seeing the right entry or seeing all my entries — this lets me have both

PR #21579, opened 2026-06-30
[PM-31068] Wire send SDK repository for edit/removePassword

adudek-bw:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-39432

## 📔 Objective

Register a client-managed send repository with the SDK so SDK-backed send edit and remove-password work.

With the pm-30110-sdk-sends-api flag on, SendClient.edit/get/list/remove_password load the existing send from the SDK's send state via state().get::<Send>() (to recover the send key and re-encrypt — SendEditRequest carries no key). But nothing populated that state, because there was no send repository registered — so editing a send hung on an infinite spinner and remove-password failed, while create (which needs no prior state) worked.

sdk-internal changes: https://github.com/bitwarden/sdk-internal/pull/1217


libs/common/src/tools/send/models/domain/send-text.ts
```diff
@@ -42,4 +44,32 @@ export class SendText extends Domain {
...
+    return {
+      text: this.text?.toSdk(),
+      hidden: this.hidden,
+    };
+  }
+
+  /** Maps an SDK `SendText` back to a domain `SendText`. */
+  static fromSdk(obj?: SdkSendText): SendText {
+    if (obj == null) {
+      return null;
```
harr1424, line 59: ❓ I see `undefined` used a lot elsewhere in this PR, and was curious will the difference be a behavioral one here?

harr1424 (commented):
Looks good overall! I see a lot of nullish coalescing operators that fallback to `null` in some cases, and `undefined` in others, and was curious why. 

In file `libs/common/src/tools/send/models/domain/send.ts` `toSdkSend()` prefers `undefined`, `fromSdkSend()` mostly prefers `null`, and `toSendData()` strictly prefers `null`. 

Not a blocker, only a curiosity on my part. I do think the value assignment/mapping logic will be complicated to troubleshoot/debug, but I also don't know if there is any better alternative.

mcamirault (dismissed):

libs/common/src/tools/send/models/domain/send-text.ts
adudek-bw: Good question — the mix is intentional and keyed to the boundary direction:

- **`toSdk*()` → `undefined`**: the SDK's generated TS interfaces type every optional field as `T | undefined` (never `| null`), and that's what the serde/WASM boundary expects for an absent optional.
- **`toSendData()` → `null`**: `SendData` is the JSON-serialized storage shape, and our storage convention is `null` (JSON preserves `null` but drops `undefined`).
- **`fromSdk*()` → `null`**: maps back into the domain layer, which uses `null` throughout (all the `fromJSON` methods, `buildDomainModel` defaults, etc.).

On the specific line here: no behavioral difference. `SendText.fromSdk` returns `null` on null input to match the other domain `fromJSON`/`fromSdk` methods, and downstream everything uses `?.`/`??`/falsy checks that treat `null` and `undefined` the same — plus `fromSdkSend` guards with `obj.text != null ?` before ever calling it. You did catch a real inconsistency in `fromSdkSend` though: `text`/`file` were using `undefined` while their siblings used `null`, so I've normalized those to `null`.

harr1424 (dismissed):

djsmith85 (dismissed):
Approving platform-owned changes

mcamirault (dismissed):

mcamirault (dismissed):

mcamirault (dismissed):

djsmith85 (dismissed):

harr1424 (dismissed):

harr1424 (approved):

djsmith85 (approved):

PR #21588, opened 2026-06-30
build: Add arm64-only MacOS pack step

iinuwa:
## 📔 Objective

Adds a build step to compile just the arm64 version of the MAS-development profile of the desktop app. This makes it faster for local compilations on macOS with Apple Silicon.


iinuwa:
Note: I think the proliferation of all these flags means that perhaps package.json should be refactored to make these options composable. But that can be done separately.

abergs (approved):

PR #21589, opened 2026-06-30
[CL-1210] use overflow directive to pack primary actions when needed

BryanCunningham:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/CL-1210

## 📔 Objective

Use existing overflow directive to pack actions into additional actions menu as viewport can no longer support bar width


libs/components/src/bulk-actions-bar/bulk-actions-bar.component.ts
```diff
@@ -302,3 +368,22 @@ export class BulkActionsBarComponent {
...
+function reveal(el: HTMLElement): () => void {
+  const prevHidden = el.hidden;
+  const prevDisplay = el.style.display;
+  el.hidden = false;
+  el.style.display = "";
+  return () => {
+    el.hidden = prevHidden;
+    el.style.display = prevDisplay;
+  };
+}
```
willmartian: I believe these are duplicated from `overflow-list.directive`? https://github.com/bitwarden/clients/pull/21589/changes#diff-8a39322023540ef97493802215471a6024aca34ec53c142fbbd1b4bf7e6de121R215-R234

willmartian (changes requested):
Looks good overall!

libs/components/src/bulk-actions-bar/bulk-actions-bar.component.ts
BryanCunningham: @willmartian Deduplicated that logic [here](https://github.com/bitwarden/clients/pull/21589/changes/96ece69ee7fcc17f8017e494fc1ab63705fbab18). Let me know what you thinkg

willmartian (dismissed):
Looks good! Thank you :shipit:

libs/components/src/bulk-actions-bar/bulk-actions-bar.component.ts
```diff
@@ -92,7 +111,16 @@ export class BulkActionsBarComponent {
...
+  protected readonly visible = computed(() => this.effectiveCount() > 0);
+
+  /**
+   * Mirrors the directive's trigger-hiding condition. Derived rather than read off
+   * `nativeElement.hidden`, which the directive may not have written yet.
+   */
+  private readonly triggerHidden = computed(() => {
+    const list = this.overflowList();
+    return list.ready() && list.overflow().length === 0 && !this.hasAdditionalActions();
+  });
```
willmartian: Perhaps we should instead expose this on `OverflowListDirective` and read it from there?

libs/components/src/bulk-actions-bar/bulk-actions-bar.component.ts
```diff
@@ -182,6 +209,19 @@ export class BulkActionsBarComponent {
...
+    // with it, but the directive only remeasures on item-set changes. Without
+    // this, widths cached while wide would drive packing once we narrow.
+    // Same `ready()` gate as above.
+    effect(() => {
+      this.compact();
+      const list = this.overflowList();
+      if (!list.ready()) {
+        return;
+      }
+      list.remeasure();
```
willmartian: It is weird that we need to guard calling remeasure like this from the bulk actions bar code. If `remeasure` needs to be guarded based on the `ready` state, should that not happen within `remeasure`?

libs/components/src/bulk-actions-bar/bulk-actions-bar.component.ts
```diff
@@ -210,9 +251,48 @@ export class BulkActionsBarComponent {
 
       onCleanup(() => manager.destroy());
     });
+
+    // Packing a button into the menu — or hiding the trigger once nothing
+    // overflows — hides it without rebuilding the manager, so a roving tabindex
+    // left on it drops the toolbar out of the tab order.
+    effect(() => {
+      const list = this.overflowList();
```
willmartian: Similar question for this `effect`, this doesn't seem specific to the bulk-actions-bar. Does it make sense to move much of this effect to the overflow list instead?

libs/components/src/bulk-actions-bar/bulk-actions-bar.mdx
willmartian, line 1: I feel like the documentation here kind of overexplains the internals / repeats what is seen in the story. What do you think? As a consumer, what all do I need to know when using the bulk actions bar? MDX documentation should be any necessary info that compliments _usage_.

libs/components/src/bulk-actions-bar/bulk-actions-bar.component.ts
BryanCunningham: @willmartian I don't disagree. I think a lot of this overflow list code is going to be re-written with the changes that needed to happen for the breadcrumbs. Since this branch has already been QA'd, my plan currently is to get this merged and then do some refactoring in the breadcrumbs branch. I can revisit this there, if that works for you? Otherwise, happy to make the change here

libs/components/src/bulk-actions-bar/bulk-actions-bar.component.ts
BryanCunningham: @willmartian Sort of the same comment as above re: fast follow refactor of the overflow directive and it's utils

libs/components/src/bulk-actions-bar/bulk-actions-bar.mdx
BryanCunningham: @willmartian Ok, I'll take another pass and remove anything that isn't really necessary

libs/components/src/bulk-actions-bar/bulk-actions-bar.component.ts
BryanCunningham: Never mind. I'll just resolve this here

vleague2 (approved):

PR #21593, opened 2026-06-30
[PM-39814] Firefox passkey login using WebAuthn fallback relay

RWebster-Noble:
## 📔 Objective

We want to enable passkey (WebAuthn/FIDO2) login and PRF-based vault unlock for the Firefox browser extension.

Firefox extensions cannot call `navigator.credentials.get()` with a custom RP ID, so we reused the same relay pattern that our existing **WebAuthn 2FA fallback connector** already uses. A web-vault connector page hosted on the vault domain performs the WebAuthn ceremony that we cannot run inside the extension, then returns the result to us through the existing content-script bridge. We extended that flow so it also fetches the assertion options, evaluates PRF, and encrypts the raw PRF output with an ephemeral ECDH key exchange before it crosses the extension boundary.

### Key changes
- New `passkey-connector.html` / `passkey-connector.ts` page for login and unlock.
- Extension background ECDH session management and `passkeyLoginResult`/`passkeyUnlockResult` handling.
- New `PasskeyRelayService` bridging background and popup contexts.
- New `LoginWithPasskeyResultComponent` and `UnlockWithPasskeyResultComponent` result popouts.
- `LoginViaWebAuthnComponentService` and `UnlockViaWebAuthnComponentService` abstractions.
- Firefox passkey login enabled in `extension-login-component.service.ts`.

### Security
- ECDH P-256 + HKDF/AES-GCM encryption for PRF bytes in transit.
- Referrer validated against known vault hostnames from `event.origin`.
- 5-minute expiry and immediate discard of the ephemeral ECDH private key.
- Raw PRF bytes are zeroed after key derivation.

### Security

- **ECDH encryption:** The raw PRF output is encrypted under a one-time P-256 ECDH shared secret (AES-GCM via HKDF) before crossing `window.postMessage` or the content script.
- **Referrer validation:** The background validates the result came from a known vault hostname via `isValidVaultReferrer(referrer)`, where `referrer` is taken from `event.origin`, not the message payload.
- **Single-use keys:** The ECDH private key is discarded immediately after decryption; pending sessions expire after 5 minutes.
- **Buffer zeroing:** Raw PRF bytes are zeroed after `createSymmetricKeyFromPrf()`/`decrypt` operations.
- **Token/session note:** This flow relies on `/webauthn/assertion-options` returning a stateless bearer token usable by the extension context. 

## 📸 Screenshots

Firefox extension login screen with "Log in with passkey" enabled
image

Firefox extension unlock screen with "Log in with passkey" enabled
image

Web vault connector page prompting for passkey authentication
image

Extension result popout showing "Logging in..." / "Unlocking..." state
image

### DIFF
--- a/apps/browser/src/auth/popup/guards/platform-popout.guard.ts
+++ b/apps/browser/src/auth/popup/guards/platform-popout.guard.ts
@@ -16,6 +16,15 @@ export function platformPopoutGuard(
   forcePopout: boolean = false,
 ): CanActivateFn {
   return async (route: ActivatedRouteSnapshot, state: RouterStateSnapshot) => {
+    // Skip popout for Firefox — WebAuthn is handled via a relay tab (not in-popup),
+    // so the popout window is unnecessary and would be left orphaned.
+    const isFirefox =
+      navigator.userAgent.indexOf(" Firefox/") !== -1 ||
+      navigator.userAgent.indexOf(" Gecko/") !== -1;
+    if (isFirefox && !forcePopout) {
+      return true;
+    }
+
     // Check if current platform matches
     const platformInfo = await BrowserApi.getPlatformInfo();
     const isPlatformMatch = platforms.includes(platformInfo.os);
--- a/apps/browser/src/auth/popup/login-with-passkey-result/login-with-passkey-result.component.html
+++ b/apps/browser/src/auth/popup/login-with-passkey-result/login-with-passkey-result.component.html
@@ -1,0 +1,27 @@
+<div class="tw-flex tw-flex-col tw-items-center tw-justify-center tw-min-h-[300px]">
+  @if (currentState() === "loggingIn") {
+    <div class="tw-text-center">
+      <div class="tw-mb-4">
+        <bit-icon name="bwi-spinner" class="tw-text-3xl tw-animate-spin"></bit-icon>
+      </div>
+      <p class="tw-text-lg">{{ "loggingIn" | i18n }}</p>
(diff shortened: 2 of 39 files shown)


micahblut:
@RWebster-Noble thank you for your contribution and your interest in expanding the passkey unlock feature to Firefox.

Unfortunately, we cannot accept the contribution. 

The PRF output is vault decryption key material. Before the connector can encrypt it, it holds it as a plaintext. The connector runs in a page served by the web vault server. If the server is ever compromised, a malicious passkey-connector.ts could be served that exfiltrates prfOutput before it is encrypted.

This introduces a vulnerability that would violate the principle that Bitwarden infrastructure cannot access unencrypted user data.

Thank you again for the contribution.

RWebster-Noble:
Thank you @micahblut for the feedback. I get your concern about exposing the PRF key in JavaScript served by the web vault. However, the web vault already handles PRF output in its own server-delivered code during passkey login. A compromised server could serve malicious JavaScript that exfiltrates the key during that flow, so the vulnerability is not introduced by this PR. It already exists in the architecture you ship today.

Also, the existing webauthn connector is also served from the web vault domain. If the server were compromised, it could modify the existing webauthn connector page to request the PRF extension and exfiltrate the output for extension users (that use the connector)

Since the web vault and extensions already operate under this same trust model, I believe this PR should be reconsidered. It simply extends an existing, accepted architecture and security requirements for Firefox users rather than introducing any new vulnerabilities.

PR #21594, opened 2026-06-30
[PM-39060] Fix new org button not showing up

BTreston:
## 🎟️ Tracking
https://bitwarden.atlassian.net/browse/PM-39060

## 📔 Objective
The change to enforce policies on accepted users caused an edge case here. Rendering these filter nodes is gated by there being at least one child filter node, which normally was always the "My Vault" node. However, with organizationDataOwnership policy enabled, individual vault node is removed. Since the user is accepted they don’t have an organization node. The end result is the whole section is hidden because there are no nodes, even though the new org button node should be allowed to appear.

also refactors some class members and getters into readonly signals.

### DIFF
--- a/apps/web/src/app/vault/individual-vault/vault-filter/shared/components/vault-filter-section.component.ts
+++ b/apps/web/src/app/vault/individual-vault/vault-filter/shared/components/vault-filter-section.component.ts
@@ -51,6 +51,24 @@ export class VaultFilterSectionComponent {
     initialValue: new Set<string>(),
   });
 
+  protected readonly headerNode = computed(() => this.data()!);
+  protected readonly headerInfo = computed(() => this.section().header);
+  protected readonly filters = computed(() => this.data()?.children);
+  protected readonly isOrganizationFilter = computed(
+    () => this.data()?.node instanceof Organization,
+  );
+  protected readonly isAllVaultsSelected = computed(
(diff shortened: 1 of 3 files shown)


apps/web/src/app/vault/individual-vault/vault-filter/shared/components/vault-filter-section.component.ts
```diff
@@ -51,6 +51,24 @@ export class VaultFilterSectionComponent {
...
+    () => this.isOrganizationFilter() && !this.activeFilter.selectedOrganizationNode,
+  );
+  protected readonly editInfo = computed(() => this.section().edit);
+  protected readonly addInfo = computed(() => this.section().add);
+  protected readonly showAddLink = computed(() => !!this.section().add?.route);
+  protected readonly optionsInfo = computed(() => this.section().options);
+  protected readonly premiumFeature = computed(
+    () => this.section().premiumOptions?.showBadgeForNonPremium,
+  );
+  protected readonly divider = computed(() => this.section().divider);
```
BTreston, line 70: Claude wanted to refactor these to comply with ADRs and I went along with it. If we want to remove these I am fine with that since it's  not AC owned code.

gbubemismith: Thanks. This is fine

gbubemismith (approved):

PR #21597, opened 2026-06-30
[PM-37963] - Add Staged Member Actions

sven-bitwarden:
## 🎟️ Tracking
[PM-37963](https://bitwarden.atlassian.net/browse/PM-37963)

## 📔 Objective
Steamrolling on with Milestone2, we need to be able to gate Staged member actions to just revoke and remove.

## 📸 Screenshots

[PM-37963]: https://bitwarden.atlassian.net/browse/PM-37963?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/web/src/app/admin-console/organizations/members/services/member-actions/member-actions.service.ts
```diff
@@ -198,6 +196,30 @@ export class MemberActionsService {
...
+  }
+
+  canRemove(user: OrganizationUserView): boolean {
+    return !user.claimedByOrganization;
+  }
+
+  canManageMember(user: OrganizationUserView): boolean {
+    return user.status !== OrganizationUserStatusType.Staged;
+  }
+
```
BTreston: I feel like these may be better suited living in the `OrganizationUserView` class, similar to the getters in the `Organization` class. These are generally useful and more portable when baked into the class IMO.

BTreston: ~Building on that (I looked in the template) these can probably be readonly signals that we init in the ctor of that class rather than functions.~ 

Fun tip: function calls like `foo()` in the template in angular is an anti-pattern and causes many re-calculations to be run during change detection (if not using onPush change detection strategy, which is not guaranteed at the consumer). Generally always prefer Observables + async pipe or Signals

apps/web/src/app/admin-console/organizations/members/members.component.ts
```diff
@@ -304,6 +304,30 @@ export class MembersComponent {
...
+  }
+
+  canRemove(user: OrganizationUserView): boolean {
+    return this.memberActionsService.canRemove(user);
+  }
+
+  canManageMember(user: OrganizationUserView): boolean {
+    return this.memberActionsService.canManageMember(user);
+  }
+
```
BTreston: The previous comments will let you get rid of all this

apps/web/src/app/admin-console/organizations/members/members.component.ts
```diff
@@ -519,11 +543,11 @@ export class MembersComponent {
...
-      showBulkConfirmUsers: members.every((m) => m.status == OrganizationUserStatusType.Accepted),
-      showBulkReinviteUsers: members.every((m) => m.status == OrganizationUserStatusType.Invited),
-      showBulkRestoreUsers: members.every((m) => m.status == OrganizationUserStatusType.Revoked),
-      showBulkRevokeUsers: members.every((m) => m.status != OrganizationUserStatusType.Revoked),
-      showBulkRemoveUsers: members.every((m) => !m.claimedByOrganization),
+      showBulkConfirmUsers: members.every((m) => this.memberActionsService.canConfirm(m)),
+      showBulkReinviteUsers: members.every((m) => this.memberActionsService.canReinvite(m)),
+      showBulkRestoreUsers: members.every((m) => this.memberActionsService.canRestore(m)),
+      showBulkRevokeUsers: members.every((m) => this.memberActionsService.canRevoke(m)),
+      showBulkRemoveUsers: members.every((m) => this.memberActionsService.canRemove(m)),
```
BTreston: these would end up being something like:

      `showBulkRemoveUsers: members.every((m) => m.canRemove())`

BTreston (changes requested):
Looks good, a few tips to tidy this up

apps/web/src/app/admin-console/organizations/members/services/member-actions/member-actions.service.ts
BTreston: NVM, The signals don't work like that. They still need to be getters ):

apps/web/src/app/admin-console/organizations/core/views/organization-user.view.ts
```diff
@@ -54,6 +59,14 @@ export class OrganizationUserView {
...
     this.avatarColor = c.avatarColor;
     this.name = c.name;
+    this.claimedByOrganization = c.claimedByOrganization ?? false;
+
+    this.canConfirm = this.status === OrganizationUserStatusType.Accepted;
+    this.canReinvite = this.status === OrganizationUserStatusType.Invited;
+    this.canRestore = this.status === OrganizationUserStatusType.Revoked;
+    this.canRevoke = this.status !== OrganizationUserStatusType.Revoked;
+    this.canRemove = !this.claimedByOrganization;
+    this.canManageMember = this.status !== OrganizationUserStatusType.Staged;
```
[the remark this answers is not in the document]
BTreston: @sven-bitwarden I think we do actually have to address this. Can we vet if these are actually treated as immutable properties? If we are mutating them as suggested here then this will cause issues.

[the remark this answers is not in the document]
sven-bitwarden: Now reloading post-confirm

BTreston (dismissed):

eliykat (commented):
Mostly relying on @BTreston 's review but it looks tidy to me 👍

eliykat (approved):

PR #21600, opened 2026-06-30
PM-36838 update template to show deviceName instead of deviceType

voommen-livefront:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-36838

## 📔 Objective

Update the HEC template to replace `deviceType` to `deviceName`
No changes were made to the server code.

## 📸 Screenshots

Datadog will show "deviceName"
Datadog also shows "deviceType" as it is computed from "device" (see screen shot below)
image

image


lastbestdev (approved):
Verified event data in Splunk has also been updated to use `deviceName` 

Image

Banrion (approved):

PR #21610, opened 2026-07-01
[PM-39841] Fix unlocking crashing the desktop app

quexten:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-39841
https://bitwarden.slack.com/archives/C07KT8VK3UG/p1782812931638739

## 📔 Objective

We introduced debounceTime as a means to prevent frequent state writes (e.g. several writes during a sync) from causing multiple SDK re-inits. However, this leads to a case where setting the userkey to state leads to the clients counting as unlocked, and the cipher observables already starting the decryption process, while the SDK is not actually replaced by an SDK instance that is initialized. This leads to all decrypts failing.

Further, this failure in some cases leads to re-attempts, which blows up memory usage. On electron, the memory usage was up to 10 gigabytes in the renderer before it died.

The temporary fix is to use throttleTime instead of debounceTime which allows through the first emission and then blocks. The long-term fix is to get rid of the observable pattern and to have a single unlocked/locked sdk instance.

### DIFF
--- a/libs/common/src/platform/services/sdk/default-sdk.service.ts
+++ b/libs/common/src/platform/services/sdk/default-sdk.service.ts
@@ -16,8 +16,8 @@ import {
   takeWhile,
   throwIfEmpty,
   firstValueFrom,
-  debounceTime,
   filter,
+  throttleTime,
 } from "rxjs";
 
 // This import has been flagged as unallowed for this class. It may be involved in a circular dependency loop.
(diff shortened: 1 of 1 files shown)


dereknance (dismissed):

quexten:
@dereknance Looks like there was a bug remaining here, which was caught by the tests suite. Added a fix.

dereknance (approved):

PR #21612, opened 2026-07-01
[CL-1158] update browser footer

BryanCunningham:
## 🎟️ Tracking
https://bitwarden.atlassian.net/browse/CL-1158

## 📔 Objective
- updated icons and colors for browser tab bar

## 📸 Screenshots
image

image


willmartian (changes requested):
Could we go ahead and move this to the CL as `bit-bottom-nav`?

willmartian (changes requested):
I left a change request in Chromatic

libs/components/src/bottom-navigation/bottom-navigation.stories.ts
```diff
@@ -0,0 +1,151 @@
...
+  VaultInactive,
+} from "@bitwarden/assets/svg";
+import { I18nService } from "@bitwarden/common/platform/abstractions/i18n.service";
+
+import { I18nMockService } from "../utils";
+
+import { BottomNavigationButton, BottomNavigationComponent } from "./bottom-navigation.component";
+
+// FIXME(https://bitwarden.atlassian.net/browse/CL-764): Migrate to OnPush
+// eslint-disable-next-line @angular-eslint/prefer-on-push-component-change-detection
```
vleague2: Obligatory "can we fix these while we're in here" comment 😂

libs/components/src/bottom-navigation/bottom-navigation.stories.ts
```diff
@@ -0,0 +1,153 @@
...
+type Story = StoryObj<BottomNavigationComponent>;
+
+export const Default: Story = {
+  render: (args) => ({
+    props: args,
+    template: /*html*/ `
+      <story-container>
+        <div class="tw-flex-1 tw-overflow-y-auto">
+          <router-outlet></router-outlet>
+        </div>
```
vleague2, line 130: Not a blocker, but do we even need this page container part of the story? Could we show just the bottom nav buttons? (Not sure if having a "page" context is better visualization perhaps)

BryanCunningham: To be able to show the 'active' state, we would need it.... I think

willmartian (approved):
LGTM

PR #21613, opened 2026-07-01
[CL-51] Simplify file upload — two composed wrappers over bit-form-field

willmartian:
## 🎟️ Tracking

Stacked on #20899 (`uif/CL-51/file-upload`). Draft — an alternative take on that PR's component-library integration for discussion.

## 📔 Objective

Same two components (`bit-file-upload`, `bit-file-dropzone`), far less machinery. Each is a **composed wrapper over `bit-form-field`** that is itself a `ControlValueAccessor`, so to end users they're **normal reactive-form controls** — `formControlName` / `[formControl]` / `[(ngModel)]` work exactly as before.

### What this removes (vs. the base PR)

- ❌ `BitFormControlProxyDirective` — the fake `NgControl` that lied to `bit-form-field` to avoid reactive forms' `setUpControl`.
- ❌ the ghost `<input type="hidden" bitInput>` whose only job was to satisfy `contentChild.required(BitFormFieldControlDirective)`.
- ❌ the self-stabilizing `effect` that re-pointed `labelForId` every cycle.
- ❌ `BitCustomInputDirective` and the `dropzone` / `file-list` sub-components (inlined).
- ❌ two competing form-field integration strategies — now there's one.

Net **~1,200 fewer lines** than the base approach.

### How it works instead

- `bit-file-upload` and `bit-file-dropzone` each implement `ControlValueAccessor` for `File[]`, so `required` / error / `touched` come from the consumer's own control — no proxy.
- `bit-form-field` gains a **control-less mode**: when no `BitFormFieldControlDirective` is projected, it renders only the label / hint / error chrome around a `[bitCustomInput]` slot, and takes a single **`[control]` + `[labelForId]`** — it derives `required` / `hasError` / `error` from that control (mirroring `BitFormFieldControlDirective`) and exposes `describedById` for the wrapper to wire `aria-describedby`. Standard usages (`bitInput`, `bit-select`, …) are unchanged.

## ⏰ Reminders before contributing

- [x] Consumers, stories, MDX, and specs updated.
- [x] `lib` type-check, ESLint (ts + html), and the file-upload/form-field Jest suites pass.

## 🦮 Accessibility

Preserved from the base PR: label association, `aria-live` add/remove announcements, upload-count status via `aria-describedby`, and roving focus after deletion.


libs/components/src/file-upload/file-upload.component.html
```diff
@@ -1,61 +1,32 @@
...
+  <!-- Filename readout in the field's default content. Clickable so the whole field opens the
+       picker; not a focus stop. -->
   <div
-    class="tw-flex tw-items-center tw-size-full"
+    class="tw-flex tw-size-full tw-items-center"
     [class.tw-cursor-pointer]="!disabled()"
     [class.tw-cursor-not-allowed]="disabled()"
-    (click)="openFilePicker()"
+    (click)="openPicker()"
   >
```
BryanCunningham, line 27: @willmartian You want to handle this key down in this PR or have me add it once this is merged?

BryanCunningham (approved):

PR #21614, opened 2026-07-01
[PM-36888] Do auto confirm when setting is enabled

BTreston:
## 🎟️ Tracking
https://bitwarden.atlassian.net/browse/PM-36888

## 📔 Objective
Auto-confirm functionality only works for users who move into an accepted state after the auto-confirm feature has been turned on in the browser. Users who were in that state prior to auto-confirm being turned on are not auto-confirmed, and must be manually confirmed. This fixes that.

### DIFF
--- a/libs/auto-confirm/src/services/default-auto-confirm.service.ts
+++ b/libs/auto-confirm/src/services/default-auto-confirm.service.ts
@@ -65,12 +65,10 @@ export class DefaultAutomaticUserConfirmationService implements AutomaticUserCon
             catchError(() => EMPTY),
           ),
         ),
+        switchMap((userId) => this.bulkAutoConfirmPendingUsers(userId)),
+        catchError(() => EMPTY),
       )
-      .subscribe((userId) => {
-        this.bulkAutoConfirmPendingUsers(userId).catch(() => {
-          // intentionally swallowed — errors are transient (network, etc.)
-        });
-      });
+      .subscribe();
   }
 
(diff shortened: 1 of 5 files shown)


jengstrom-bw (dismissed):

libs/auto-confirm/src/services/default-auto-confirm.service.ts
```diff
@@ -65,12 +65,10 @@ export class DefaultAutomaticUserConfirmationService implements AutomaticUserCon
             catchError(() => EMPTY),
           ),
         ),
+        switchMap((userId) => this.bulkAutoConfirmPendingUsers(userId)),
```
r-tome: Shouldn't we use `mergeMap` here? We could end up with a few requests being cancelled if the user unlocks with another account mid-flight

libs/auto-confirm/src/services/default-auto-confirm.service.ts
```diff
@@ -65,12 +65,10 @@ export class DefaultAutomaticUserConfirmationService implements AutomaticUserCon
...
+        catchError(() => EMPTY),
```
r-tome: The previous implementation swallowed errors returned by `bulkAutoConfirmPendingUsers` whereas here we're killing the whole subscription, right?

BTreston: Yes, good catch on these. I'll fix the merge map and this should be on the inner observable

r-tome (approved):

jengstrom-bw (approved):

PR #21620, opened 2026-07-01
[PM-17584] Delinea XML parser

mcamirault:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-17584

## 📔 Objective

This PR adds a new importer that handles the Delinea XML export format. All Secrets are imported as Secure Notes with their information stored as Text type custom fields (except passwords, which are imported as Hidden type)

## 📸 Screenshots

https://github.com/user-attachments/assets/753c4e43-c966-4414-be2b-68b373c3701c

### DIFF
--- a/libs/importer/src/components/import.component.html
+++ b/libs/importer/src/components/import.component.html
@@ -448,6 +448,11 @@
           entries should be included in the export &rarr; Click on next to export into the location
           previously chosen.
         </ng-container>
+        <ng-container *ngIf="format === 'delineaxml'">
+          Log into your organization's Secret Server instance. Search for Export/Import using the
+          search box at the top of the screen and click the Export button to go to the Export page.
+          Follow the instructions on this page to export an XML file.
+        </ng-container>
       </bit-callout>
       @if (showLastPassOptions) {
         <import-lastpass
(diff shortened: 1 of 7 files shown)


harr1424 (dismissed):
Code changes look good! Two questions: 

1. Did this undergo design review? Should it? 
2. Was there a reason we didn't [implement this in the SDK?](https://github.com/bitwarden/clients/pull/21052#pullrequestreview-4429198732)

mcamirault:
> Code changes look good! Two questions:
> 
>     1. Did this undergo design review? Should it?
> 
>     2. Was there a reason we didn't [implement this in the SDK?](https://github.com/bitwarden/clients/pull/21052#pullrequestreview-4429198732)

@harr1424 

1. Since I'm just adding a new entry to our importers list (and not the featured section) I didn't think it was really necessary. The only exception might be the instructions, which I wrote based on the [documentation here](https://docs.delinea.com/online-help/secret-server/secret-operations/secret-import-and-export/exporting-secrets/index.htm) since I don't have access to an instance of Secret Server. I can check with Product Design, see if what I wrote was acceptable
2. Habit, mostly 😛 But also since your KDBX importer is the first one in the SDK there's a bunch of support code you had to write to handle the registration, the difference between the SDK and non-SDK importers, etc. that I'd rather not duplicate here. Given how simple this importer is I don't think a Rust rewrite after your KDBX change goes in would be too much work, and since I'm the one saying so I think it's fair to say that I'm volunteering for it

harr1424 (approved):

PR #21623, opened 2026-07-01
[PM-39270] - [Safari][Browser] Settings icons and title pages are misaligned 

jaasen-livefront:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-39270

## 📔 Objective

Fix item content rows rendering with extra vertical height when macOS scrollbars are visible.

`bit-item-content` was applying `tw-overflow-auto` to non-truncated text content. When persistent scrollbars are enabled, that can reserve scrollbar space inside the text wrapper, causing affected item rows to become taller even when the text itself does not need scrolling.

This removes `tw-overflow-auto` from wrapped item content while preserving text wrapping and word breaking behavior.

Verified with:
- macOS scrollbar setting set to `Always`
- Browser extension settings popup in non-compact mode
- Tailwind CSS build check

## 📸 Screenshots

Before: affected settings rows showed extra vertical height when scrollbars were always visible.

After: settings rows render at the expected height without the extra space.

Screenshot 2026-07-01 at 3 40 43 PM

### DIFF
--- a/libs/components/src/item/item-content.component.html
+++ b/libs/components/src/item/item-content.component.html
@@ -9,7 +9,7 @@
       <div
         [ngClass]="{
           'tw-truncate': truncate(),
-          'tw-text-wrap tw-overflow-auto tw-break-words': !truncate(),
+          'tw-text-wrap tw-break-words': !truncate(),
         }"
       >
         <ng-content></ng-content>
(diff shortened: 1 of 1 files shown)


willmartian (changes requested):
Please see the failing UI test from Chromatic

willmartian (commented):
Is this reproducible/visible in any of the existing stories at https://components.bitwarden.com/?path=/docs/component-library-item--docs?

jaasen-livefront:
> Is this reproducible/visible in any of the existing stories at https://components.bitwarden.com/?path=/docs/component-library-item--docs?

In my case it required setting sshow scroll bars to "Always" in macos settings. Looks like the cure is worse than the disease here. I'll see what else I can come up with.

jaasen-livefront:
Closing. Fixed in https://bitwarden.atlassian.net/browse/CL-982

PR #21632, opened 2026-07-02
[PM-39554] Create default collection when admin is demoted to user

BTreston:
## 🎟️ Tracking
https://bitwarden.atlassian.net/browse/PM-39554

## 📔 Objective
This PR fixes an issue when an admin's role is demoted to a user, no "My Items" collection is created for them.

### DIFF
--- a/apps/web/src/app/admin-console/organizations/core/services/user-admin.service.ts
+++ b/apps/web/src/app/admin-console/organizations/core/services/user-admin.service.ts
@@ -1,10 +1,16 @@
 import { Injectable } from "@angular/core";
+import { firstValueFrom, switchMap } from "rxjs";
 
 import {
   OrganizationUserApiService,
   OrganizationUserInviteRequest,
+  OrganizationUserService,
   OrganizationUserUpdateRequest,
 } from "@bitwarden/admin-console/common";
+import { OrganizationService } from "@bitwarden/common/admin-console/abstractions/organization/organization.service.abstraction";
+import { AccountService } from "@bitwarden/common/auth/abstractions/account.service";
+import { getUserId } from "@bitwarden/common/auth/services/account.service";
+import { getById } from "@bitwarden/common/platform/misc";
 import { Guid, OrganizationId } from "@bitwarden/common/types/guid";
 
 import { CoreOrganizationModule } from "../core-organization.module";
@@ -12,7 +18,12 @@ import { OrganizationUserAdminView } from "../views/organization-user-admin-view
 
 @Injectable({ providedIn: CoreOrganizationModule })
 export class UserAdminService {
-  constructor(private organizationUserApiService: OrganizationUserApiService) {}
+  constructor(
+    private organizationUserApiService: OrganizationUserApiService,
+    private organizationUserService: OrganizationUserService,
+    private organizationService: OrganizationService,
+    private accountService: AccountService,
+  ) {}
 
   async get(
     organizationId: OrganizationId,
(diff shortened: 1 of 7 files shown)


libs/admin-console/src/common/organization-user/services/default-organization-user.service.ts
```diff
@@ -119,6 +121,30 @@ export class DefaultOrganizationUserService implements OrganizationUserService {
     );
   }
 
+  updateUser(
+    organization: Organization,
+    userId: string,
+    request: OrganizationUserUpdateRequest,
+  ): Observable<void> {
+    const exempt =
+      request.type === OrganizationUserType.Owner || request.type === OrganizationUserType.Admin;
```
eliykat, line 130: We should also check `organization.UseMyItems && organization.UsePolicies` and that the centralize organization ownership policy itself is enabled.

What's neat is that the SDK policy interface is stateless, so you could actually construct and provide this info to `sdkService.client$.policies.filter_by_type` and it would handle all this logic for you - even though it's for a future state for a different user. I think that is the best solution, otherwise you are effectively reimplementing policy logic here. We haven't used it like that before though, so if that doesn't work it's not a blocker.

apps/web/src/app/admin-console/organizations/core/services/user-admin.service.ts
```diff
@@ -52,7 +63,20 @@ export class UserAdminService {
     userId: Guid,
     organizationId: OrganizationId,
   ): Promise<void> {
-    await this.organizationUserApiService.putOrganizationUser(organizationId, userId, request);
+    await firstValueFrom(
+      this.accountService.activeAccount$.pipe(
```
eliykat: We shouldn't inject `activeAccount` in the service layer, this should be passed in.

As an aside I don't understand why this service method exists to just reroute to a different service, but it's an existing pattern so it's not a blocker.

apps/web/src/app/admin-console/organizations/core/services/user-admin.service.ts
```diff
@@ -52,7 +63,20 @@ export class UserAdminService {
     userId: Guid,
```
eliykat: `userId` appears to refer to the target organizationUserId - this should be clarified.

eliykat (changes requested):

apps/web/src/app/admin-console/organizations/core/services/user-admin.service.ts
BTreston: I did this, needed to preserve this service until the member dialog is removed since the new dialog is gated behind the invite link feature flag. It would be more work to try and refactor the existing dialog that's going to be removed anyway than to just wire up these wrappers. I planned to just remove this service altogether once that was done.

libs/admin-console/src/common/organization-user/services/default-organization-user.service.ts
BTreston: I don't think the SDK method (or the existing policy service for that matter) will work for this case. This is another example of trying to read the policy state for the non-active user. We don't fetch the policies for the target user as far as I am aware. This has come up as a limitation quite a few times in the past. These checks are also done on the server side, but I'll also add them here too.

eliykat: This is what I meant by it being stateless. It will evaluate whatever arguments you pass in, not just for the active user. In this case:
- we have the policy because we're part of the organization as well
- we have the details for the target user we're updating

So you can give that to the SDK client directly (bypassing policyService, which is tied to the active user), and it'll give you an accurate result.

I acknowledge we may want a better interface for this though so that we're not injecting the SDK client everywhere.

I did also wonder if we should just do it unconditionally on the client-side and leave it up to the server, but I wasn't sure.

No action required.

eliykat (approved):

PR #21634, opened 2026-07-02
[PM-29790] Windows passkey plugin: unify signature verification 

iinuwa:
## 🎟️ Tracking

[PM-29790](https://bitwarden.atlassian.net/browse/PM-29790)

## 📔 Objective

Following up from a [previous review comment](https://github.com/bitwarden/clients/pull/21622#discussion_r3513676812), the context strings between get/create requests and cancel requests are different. That requires the desktop app to parse the transaction ID instead of just storing the context string opaquely for cancellation. (The user verification request still does need to parse the request context.)

This unifies this by adding the request hash to the context string.

To get there though, we have to add the request hash to the context, causing extra plumbing during the initial verification. To simplify that, we also use a single function to verify an operation request, whether it is a get/create/cancel.

This makes cancel operation behave much more like get/create.

[PM-29790]: https://bitwarden.atlassian.net/browse/PM-29790?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/desktop/desktop_native/win_webauthn/src/api/plugin/com.rs
+++ b/apps/desktop/desktop_native/win_webauthn/src/api/plugin/com.rs
@@ -5,6 +5,7 @@
 use std::{
     alloc,
     mem::{size_of, ManuallyDrop, MaybeUninit},
+    ops::DerefMut,
     ptr::{self, NonNull},
     sync::{Arc, Mutex, OnceLock},
 };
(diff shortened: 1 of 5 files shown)


iinuwa:
The CI failure is unrelated and is being addressed in [PM-39889](https://bitwarden.atlassian.net/browse/PM-39889).

[PM-39889]: https://bitwarden.atlassian.net/browse/PM-39889?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

abergs (approved):

PR #21640, opened 2026-07-02
[PM-39556] Fix access selector not saving changes when the dialog field is modified

BTreston:
## 🎟️ Tracking
https://bitwarden.atlassian.net/browse/PM-39556

## 📔 Objective
When changes are made to collection permission when editing a member with the access selector table, if the collection permission field above is modified the access selector changes are lost. This PR fixes this and decouples pending changes to the field from the access selector


r-tome (dismissed):
Great testing on this!

r-tome (approved):

PR #21646, opened 2026-07-02
[PM-37196] Tooltip expand max-width

JaredScar:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-37196

## 📔 Objective

Expands the max-width from `12rem` for tooltips to `12.25rem` so the tooltip for revocation reasons is 2 liner instead of 3 liner...

## 📸 Screenshots

Old:
image

New:
image


willmartian (approved):
LGTM!

PR #21650, opened 2026-07-03
Add Early Access option on desktop, extension, and web clients

trmartin4:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-39893

## 📔 Objective

Adds an option in the desktop, extension, and web clients to allow "early access" to features not yet ready for production release, by setting the `Is-Prerelease` header on requests if the user chooses to do so.

This header is passed up on:
1. Authenticated `/api/config` requests, which is important for client-flagged changes
2. All authenticated API requests, which is important if any Early Access features are flagged server-side.

### DIFF
--- a/apps/browser/src/_locales/en/messages.json
+++ b/apps/browser/src/_locales/en/messages.json
@@ -6889,19 +6889,19 @@
   "cannotSaveItemNoConfirmedOrgs": {
     "message": "Unable to save item. You must be confirmed to the organization to save items."
   },
-  "betaMode": {
-    "message": "Beta mode"
+  "earlyAccess": {
+    "message": "Early access"
   },
-  "betaModeDesc": {
+  "earlyAccessDesc": {
     "message": "Enable early access to features currently in prerelease testing"
   },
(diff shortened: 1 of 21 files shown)


willmartian (commented):
Nothing egregious jumped out to me in the frontend implementation. 👍 

Thoughts:
- My _guess_ is that Design might want to use [a switch](https://components.bitwarden.com/?path=/story/component-library-form-switch--form-control-card) for something like this, but that can be sussed out during design review. 
- I haven't personally used/seen this HTTP header approach used for another feature in the app before. On search, I do see there is one use. So that jumps out to me as something to be verified by @bitwarden/team-platform. 
  - I assume the reason we went with this is to have client-owned state, shared with the server? Instead of server-owned state, set by the client. And that is because we don't want to sync this setting between multiple clients? Would an alternate approach be to allow the storage of arbitrary client state on the server, in the users table? `clientId123: { some json blob }`.  How might that compare in complexity & other pros/cons? (Just rubber ducking--not making a claim that this approach is better.)

PR #21654, opened 2026-07-03
SSH Agent v2: restructure the SignRequest

neuronull:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-37774

## 📔 Objective

The primary changes of interest are in auth_policy.rs , where we now have an improved data structure representation of the sign request information.

The original representation had flattened all data and thus the Options on them resulted in convoluted scenarios in the upstream that are never actually possible due to how the data is collected.

The new representation groups the data fields by origin, and the napi layer translates it for Electron.

## 📸 Testing done

- validated that sign requests still work


neuronull:
note- the Lint Rust failure for cargo deny is unrelated

quexten (approved):

PR #21701, opened 2026-07-07
[PM-39783] Add override for policy enabled state

eliykat:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-39783

## 📔 Objective

Example code for how we can support the proposed behavior of the new Send Controls policy.

Expected behavior:
- the policy toggle turns the Send _feature_ on or off
- toggle off -> Send is disabled
- toggle on -> Send is enabled, and the additional controls apply (e.g. allowed send types)
- default state: toggle is on

This doesn't match usual behavior, because usually the toggle being off means that the policy is disabled and has no effect. But here, it has effects in both states.

The approach in this PR: we treat the policy as always-enabled behind the scenes (so that it always has some effect), and the `policy.data.disableSend` value maps to the toggle state. This avoids additional data migrations and means we don't have to make adjustments to policy enforcement logic to support disabled policies doing something. It preserves the mental model of "enabled = potentially enforced".


eliykat:
Changes were incorporated into the main feature PR, no longer required.

PR #21702, opened 2026-07-07
Prevent connect to desktop in ipc background if native messaging permission is missing

quexten:
## 🎟️ Tracking

[no ticket]

## 📔 Objective

connectToDesktop should not attempt to send if the native messaging permission is missing. If it is not grated, it should just do nothing, as to prevent error spam.

### DIFF
--- a/apps/browser/src/platform/ipc/ipc-background.service.ts
+++ b/apps/browser/src/platform/ipc/ipc-background.service.ts
@@ -150,9 +150,13 @@ export class IpcBackgroundService extends IpcService {
 
   /**
    * Starts a connection to the desktop app. This function attempts to establish a connection with the desktop application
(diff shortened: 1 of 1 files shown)


djsmith85 (approved):
Changes are looking good. 

Please do create a ticket so we can track these changes and let QA decide if this requires verification

quexten:
> Please do create a ticket so we can track these changes and let QA decide if this requires verification

I would, however, IPC background service is only used by shared unlock and we have a collective task to QA it. This means that this code is not to production code / can be considered feature flagged.

It can only be QA'd collectively as part of the full shared unlock qa.

@djsmith85 Please let me know if you have concerns with this approach, I do not see another way to effectively QA this.

PR #21703, opened 2026-07-07
[PM-40060] Add native messaging permission dialog

quexten:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-40060

## 📔 Objective

Adds a new dialog. This will be responsible for asking for the native messaging permission. The design is confirmed by product.

The flow will be:
1. User clicks on feature that needs permission
2. If popup -> convert to popout and show dialog
3. If popout -> show dialog
4. -> The dialog primes the user.
5. After the user clicks continue, ask for permission
6. After permission is granted, return true
7. -> The caller is expected to set / unset the setting in the UI thread it was called from. If the result was true, they must also do an extension reload for the permission to become active.

## 📸 Screenshots

image

### DIFF
--- a/apps/browser/src/key-management/shared-unlock/popup/native-messaging-permission-dialog.component.ts
+++ b/apps/browser/src/key-management/shared-unlock/popup/native-messaging-permission-dialog.component.ts
@@ -63,7 +63,9 @@ export class NativeMessagingPermissionDialogComponent {
     optional: true,
   });
   protected readonly descriptionKey =
-    DESCRIPTION_KEY_BY_TYPE[this.params?.type ?? NativeMessagingPermissionDialogType.Biometrics];
+    DESCRIPTION_KEY_BY_TYPE[
+      (this.params?.type as NativeMessagingPermissionDialogType | undefined) ?? NativeMessagingPermissionDialogType.Biometrics
+    ];
 
   async continue() {
     let granted = false;
@@ -81,7 +83,6 @@ export class NativeMessagingPermissionDialogComponent {
           type: "info",
         });
       }
-      granted = false;
     }
     await this.dialogRef.close(granted);
   }
--- a/apps/browser/src/key-management/shared-unlock/popup/native-messaging-permission-dialog.component.ts
+++ b/apps/browser/src/key-management/shared-unlock/popup/native-messaging-permission-dialog.component.ts
@@ -83,7 +83,7 @@ export class NativeMessagingPermissionDialogComponent {
       }
       granted = false;
     }
-    void this.dialogRef.close(granted);
+    await this.dialogRef.close(granted);
   }
 
   /**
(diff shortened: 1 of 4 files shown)


apps/browser/src/key-management/shared-unlock/popup/native-messaging-permission-dialog.component.ts
```diff
@@ -0,0 +1,110 @@
...
+   * extension on grant, which would close a popup or sidebar; the flow must be popped out to a
+   * stable window first.
+   */
+  static open(
+    dialogService: DialogService,
+    params: NativeMessagingPermissionDialogParams,
+  ): DialogRef<boolean> {
+    if (!BrowserPopupUtils.inPopout(window)) {
+      throw new Error(
+        "NativeMessagingPermissionDialogComponent must be opened from a popout, so the extension " +
```
mzieniukbw, line 102: ❓ Does it need internationalisation ? I have a feeling this will be shown as a toast to the user, worth checking.

apps/browser/src/key-management/shared-unlock/popup/native-messaging-permission-dialog.component.ts
```diff
@@ -0,0 +1,110 @@
...
+      if (this.platformUtilsService.isFirefox() && BrowserPopupUtils.inSidebar(window)) {
+        await this.dialogService.openSimpleDialog({
+          title: { key: "nativeMessaginPermissionSidebarTitle" },
+          content: { key: "nativeMessaginPermissionSidebarDesc" },
+          acceptButtonText: { key: "ok" },
+          cancelButtonText: null,
+          type: "info",
+        });
+      }
+      granted = false;
```
mzieniukbw: default is false, this does not do anything.

apps/browser/src/key-management/shared-unlock/popup/native-messaging-permission-dialog.component.ts
```diff
@@ -0,0 +1,110 @@
...
+    }
+    void this.dialogRef.close(granted);
```
mzieniukbw: It is better to use await here.
```suggestion
    await this.dialogRef.close(granted);
```

mzieniukbw (changes requested):

apps/browser/src/key-management/shared-unlock/popup/native-messaging-permission-dialog.component.ts
quexten: No, this should not be shown to users. The caller must make sure that the API is only called from a popout.

apps/browser/src/key-management/shared-unlock/popup/native-messaging-permission-dialog.component.ts
quexten: I guess :shrug:

quexten: [24f3c18](https://github.com/bitwarden/clients/pull/21703/commits/24f3c185dc233f426e2cf86badf2d20894105b3e)

mzieniukbw (dismissed):

mzieniukbw (approved):

PR #21705, opened 2026-07-07
[PM-40060] Use native messaging dialog in browser extension settings

quexten:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-40060
Blocked by: https://github.com/bitwarden/clients/pull/21782

## 📔 Objective

Uses the new native messaging dialog for setup of biometric unlock and shared unlock. Also, removes the confirmation dialog for biometrics. We will remove the entire biometric unlock setting soon, and shared unlock does not have a similar mechanism.


apps/browser/src/auth/popup/settings/account-security.component.ts
```diff
@@ -330,50 +373,47 @@ export class AccountSecurityComponent implements OnInit, OnDestroy {
...
-          this.form.controls.biometric.setValue(false);
+      const hadPermission = await BrowserApi.permissionsGranted(["nativeMessaging"]);
+      if (!hadPermission) {
+        if (!BrowserPopupUtils.inPopout(window)) {
+          // Requesting the permission outside a popout would close the window (granting reloads
+          // the extension), so pop out to a stable window. The popped-out window presents the
+          // dialog and requests the permission. With hash routing, Angular reads query params
+          // from after the hash route.
+          const url = new URL(window.location.href);
+          url.hash += (url.hash.includes("?") ? "&" : "?") + "autoRequestBiometrics=true";
```
mzieniukbw, line 390: Query params are part of [URL's search](https://developer.mozilla.org/en-US/docs/Web/API/URL/search)

apps/browser/src/auth/popup/settings/account-security.component.ts
```diff
@@ -330,50 +373,47 @@ export class AccountSecurityComponent implements OnInit, OnDestroy {
...
+          await BrowserPopupUtils.openCurrentPagePopout(window, url.href);
```
mzieniukbw, line 391: ❓  do we need to set the form's biometrics back to false ?

apps/browser/src/auth/popup/settings/account-security.component.ts
```diff
@@ -330,50 +373,47 @@ export class AccountSecurityComponent implements OnInit, OnDestroy {
...
-        this.validationService.showError(error);
+      await this.biometricStateService.setBiometricUnlockEnabled(true, userId);
+
+      if (!hadPermission) {
+        // The nativeMessaging permission was just granted. The extension must reload to register
+        // the native messaging host. State is saved above so it survives the reload. The
+        // background performs the reload once all popups/popouts have closed; this popout won't
+        // receive its own broadcast, so close it explicitly to unblock that reload.
+        this.messagingService.send("reloadExtension");
+        await BrowserPopupUtils.closeCurrentPopupOrPopout(window);
```
mzieniukbw, line 416: ❓ Why not `await this.processReloadService.startProcessReload();`. This feels more correct and is doing the same thing.

quexten: Processreload is prevented if there are unlocked accounts currently... I guess we could instead modify the api to accept a (ignore_unlocked_accounts) param.  Any preference?

apps/browser/src/auth/popup/settings/account-security.component.ts
quexten: https://github.com/user-attachments/assets/cf878f6a-7eb3-4997-8eff-15c616974f11

Here is a recording. Popping out resets the form to false, but because the params are set, it auto-requests the permission dialog, which - after accepting - will set it to true again.

apps/browser/src/auth/popup/settings/account-security.component.ts
quexten: I looked into this, and search is nicer to work with, however unfortunately the application overwrites anything I put into search with a ui-popout location property. I don't know if it's worth it to dig much more into this though..

mzieniukbw (approved):

ike-kottlowski (approved):
Auth changes are good.

PR #21708, opened 2026-07-07
[PM-39259] Migrate vault export and generator to SDK random number client

quexten:
https://bitwarden.atlassian.net/browse/PM-39259

Migrates tools off of cryptofunctionservice randomness, to the sdk random implementation. Additionally, this replaces the PureCrypto randomness with the new SdkRandomNumberClient.


libs/tools/generator/extensions/legacy/src/create-legacy-username-generation-service.ts
```diff
@@ -14,7 +14,7 @@ import { KeyService } from "@bitwarden/key-management";
 import { LegacyUsernameGenerationService } from "./legacy-username-generation.service";
 import { UsernameGenerationServiceAbstraction } from "./username-generation.service.abstraction";
```
quexten, line 15: @adudek-bw Do you happen to know if tools has work to remove the remaining use-cases / port them to the SDK?

harr1424: @quexten Tools Team is working on a [tech-breakdown](https://github.com/bitwarden/tech-breakdowns/pull/19) describing all such work, but I am still discovering gaps in this document. I don't want to speak for @adudek-bw , but yes, we are planning to migrate generator logic **completely** into the SDK and eventually first deprecate, then remove the unused TS code.

harr1424 (dismissed):

libs/tools/generator/extensions/legacy/src/create-legacy-username-generation-service.ts
quexten: Thanks @harr1424!

harr1424 (approved):

PR #21715, opened 2026-07-07
[PM-39876] Implement targeting rules form category

jprusik:
## 🎟️ Tracking

[PM-39876](https://bitwarden.atlassian.net/browse/PM-39876)

## 📔 Objective

The extension treats username as a shared data concept (email/username), but map-the-web treats concepts semantically, expecting consumers to implement concerns like autofill on the contextual basis of the form category. In our case, we can have login forms that have email and/or phone number fields and no username field.

Using the Forms map category context, we are now able to map discrete fields to a shared username concept for stored login ciphers.

[PM-39876]: https://bitwarden.atlassian.net/browse/PM-39876?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/browser/src/autofill/models/autofill-field.ts
+++ b/apps/browser/src/autofill/models/autofill-field.ts
@@ -1,3 +1,5 @@
+import { AutofillTargetingRuleType, FormPurposeCategory } from "@bitwarden/common/autofill/types";
+
 import { FieldRect } from "../background/abstractions/overlay.background";
 import { AutofillFieldQualifierType } from "../enums/autofill-field.enums";
 import {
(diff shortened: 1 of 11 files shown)


apps/browser/src/autofill/models/autofill-field.ts
```diff
@@ -124,13 +126,20 @@ export default class AutofillField {
...
 
   /**
    * Indicates this field was qualified by targeting rules rather than heuristics
    */
   targeted?: boolean;
 
+  /**
+   * The `category` of the targeting-rule form this field was resolved from
+   * (e.g. `account-login`). Only set for targeted fields; undefined for
+   * heuristic fields and for targeted fields routed across an iframe boundary.
```
bensbits91: Since the second commit ([pass formCategory through iframe traversals](https://github.com/bitwarden/clients/pull/21715/commits/6cc5678c78f06646c994543201ebb15816f3270d)) now threads this across sub-frames, should this comment be updated to match? Maybe something like:

```
/**
 * The `category` of the targeting-rule form this field was resolved from
 * (e.g. `account-login`). Set for targeted fields and propagated across
 * iframe routing, so sub-frame-resolved fields retain the source form's
 * category. Undefined for heuristic fields.
 */
```

bensbits91 (dismissed):
Looks good. One question/suggestion about a JSDoc comment.

apps/browser/src/autofill/models/autofill-field.ts
jprusik: ah, good call - I had originally planned to follow up the traversal in another task, and changed my mind when it ended up being small.

fixed!

bensbits91 (approved):

PR #21717, opened 2026-07-07
[PM-40074] Refactor policy definitions to remove deprecated versions and streamline components

JaredScar:
## 🎟️ Tracking
https://bitwarden.atlassian.net/browse/PM-40074

## 📔 Objective

Updates to utilize v2 nesting pattern so policy updates stay consistent. Also fix up the spec tests to point to v2 components when possible


BTreston (changes requested):
I don't think this is correct. The way these are implemented will leak out new design changes intended for the policy drawers into the existing components UI. For example: the OrganizationDataOwnership policy will show the new enabled badge with the feature flag turned off. Unless there's a misunderstanding on my part and we DO want to show the new design in the dialogs w/out the feature flag enabled. If that is the case then please feel free to correct me.

For reference, this is with the policy drawers flag set to `false`:
Screenshot 2026-07-07 at 1 13 09 PM

JaredScar:
> I don't think this is correct. The way these are implemented will leak out new design changes intended for the policy drawers into the existing components UI. For example: the OrganizationDataOwnership policy will show the new enabled badge with the feature flag turned off. Unless there's a misunderstanding on my part and we DO want to show the new design in the dialogs w/out the feature flag enabled. If that is the case then please feel free to correct me.
> 
> For reference, this is with the policy drawers flag set to `false`: Screenshot 2026-07-07 at 1 13 09 PM

With it off now after previous changes:
image

@BTreston

BTreston (changes requested):
There are still new design changes leaking out into the existing dialogs. See the master password requirements policy.

Please, fix all of the components and validate that the designs have not leaked out from behind the flag. At this stage, it may also be helpful to write stories for the old policies and the new so we can detect these issues with CI rather than relying on manual testing.

BTreston (changes requested):
Centralize organization data ownership story is broken.

BTreston (approved):

cyprain-okeke (approved):
Looks Good for billing changes

JimmyVo16:
Removing myself as a reviewer since Brandon is already reviewing this and has more context.

PR #17052, opened 2025-10-27
[PM-27495] Do not show true length of password in password mask

Mauritz8:
## 🎟️ Tracking

https://github.com/bitwarden/clients/issues/13867

## 📔 Objective

This PR fixes https://github.com/bitwarden/clients/issues/13867. When a password is hidden, the password mask is now a fixed length of 8 characters.

## ⏰ Reminders before review

- Contributor guidelines followed
- All formatters and local linters executed and passed
- Written new unit and / or integration tests where applicable
- Protected functional changes with optionality (feature flags)
- Used internationalization (i18n) for all UI strings
- CI builds passed
- Communicated to DevOps any deployment requirements
- Updated any necessary documentation (Confluence, contributing docs) or informed the documentation team

## 🦮 Reviewer guidelines

- 👍 (`:+1:`) or similar for great changes
- 📝 (`:memo:`) or ℹ️ (`:information_source:`) for notes or general info
- ❓ (`:question:`) for questions
- 🤔 (`:thinking:`) or 💭 (`:thought_balloon:`) for more open inquiry that's not quite a confirmed issue and could potentially benefit from discussion
- 🎨 (`:art:`) for suggestions / improvements
- ❌ (`:x:`) or ⚠️ (`:warning:`) for more significant problems or concerns needing attention
- 🌱 (`:seedling:`) or ♻️ (`:recycle:`) for future improvements or indications of technical debt
- ⛏ (`:pick:`) for minor or nitpick changes

PR #21722, opened 2026-07-07
[PM-19263] fix: do not show true length of password in password mask

nick-livefront:
## 🎟️ Tracking

[PM-19263](https://bitwarden.atlassian.net/browse/PM-19263)

Picks up #17052 (originally by @Mauritz8), rebased onto current `main`.

## 📔 Objective

The password mask's dot-count matched the real password length, leaking length at a glance. Fixed by using `LoginView.maskedPassword` (a fixed-length `"••••••••"` mask) instead of the raw password when not revealed. Merged the original fix forward and updated the component's tests accordingly.

## 📸 Screenshots

https://github.com/user-attachments/assets/ee34a8f9-0a91-4aea-bedb-612dd1fbd598

[PM-19263]: https://bitwarden.atlassian.net/browse/PM-19263?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


djsmith85:
@nick-livefront If we decide to carry a community PR, we should ideally just rebase/merge `main` and apply any required changes on the contributors branch. This ensures that they receive the full attribution on GH and the automation for the release notes can pick up the PR for the praise section in the release notes. Sometimes their forks/branches are write protected, but they usually aren't.

The attribution will now only be partially picked up due to their commit, which will get squashed on merge.

Please try and update @Mauritz8's [PR](https://github.com/bitwarden/clients/pull/17052) and only if that does not work, continue with your current approach.

Kind regards,
Daniel

nick-livefront:
> Please try and update @Mauritz8's [PR](https://github.com/bitwarden/clients/pull/17052) and only if that does not work, continue with your current approach.
> 
> Kind regards, Daniel

@djsmith85 I was blocked from pushing to the contributors branch yesterday, but it worked this morning after I cloned their entire repo rather than just checking out the PR. I'll close this out! Thanks for the nudge!

PR #21723, opened 2026-07-07
[PM-39218] Use global environment for self-hosted env dialog and SSO URL

trmartin4:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-39218

## 📔 Objective

See https://github.com/bitwarden/clients/pull/21721 for larger context for this change.

Directly addresses the issue in https://bitwarden.atlassian.net/browse/PM-39218, which occurs in the self-hosted URL selection dialog as well as the SSO URL identification during the SSO login process.

We do this here by moving from the (to be deprecated) `environment$` to the `globalEnvironment$` in these cases, as these are by definition pre-authentication uses.

Note that this PR also moves the `sso-localhost-callback.service.ts` to Auth ownership.

Is this a partial fix?  Yes.  There are multiple other PRs to address other cases where the environment usage can be clearer:
1. In `ApiService`: https://github.com/bitwarden/clients/pull/21724
2. In remaining uses pre-authentication: https://github.com/bitwarden/clients/pull/21350

### Self-hosted URL dialog persists correctly
https://github.com/user-attachments/assets/bf902f68-f3ca-47ae-91d0-257d86400320

### SSO URL for global environment is used on SSO redirect
https://github.com/user-attachments/assets/87bde66c-861e-4751-98bd-278c1c6a874b


dani-garcia (approved):
LGTM, no platform changes other than some import reorder

ike-kottlowski (approved):
Auth changes are good. Agree with Claude's assessment.

PR #21725, opened 2026-07-07
[PM-38392] Route password-protected Send saves through the SDK

adudek-bw:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-38392

## 📔 Objective

Route password-protected Send saves through the SDK (behind the pm-30110-sdk-sends-api flag).


apps/cli/src/tools/send/commands/create.command.ts
```diff
@@ -162,7 +162,9 @@ export class SendCreateCommand {
 
       const sendView = SendResponse.toView(req);
       const [encSend, fileData] = await this.sendService.encrypt(sendView, fileBuffer, password);
-      await this.sendApiService.save([encSend, fileData]);
+      // Forward the plaintext so the SDK path can derive the send password over the key it
```
harr1424:
```suggestion
      // Forward the plaintext password so the SDK path can derive the send password over the key it
```
Would this be more accurate?

apps/cli/src/tools/send/commands/edit.command.ts
```diff
@@ -107,7 +107,9 @@ export class SendEditCommand {
 
     try {
       const [encSend, encFileData] = await this.sendService.encrypt(sendView, null, req.password);
-      await this.sendApiService.save([encSend, encFileData]);
+      // Forward the plaintext (null when preserving an existing password) so the SDK path can
```
harr1424:
```suggestion
      // Forward the plaintext  password (null when preserving an existing password) so the SDK path can
```

harr1424 (dismissed):

harr1424 (approved):

PR #21729, opened 2026-07-08
[PM-39259] Remove typescript cipherkey generation

quexten:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-39259

## 📔 Objective

We want to only use SDK generated keys for fedramp, to audit where the randomness originates. Rather than porting cipherkey generation, I just removed it since the code is unused.


eligrubb (dismissed):
🔥🔥

jengstrom-bw (approved):
vault item look good!

eligrubb (approved):

PR #21730, opened 2026-07-08
fix(biometrics): Biometric unlock does not work over SDK

quexten:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-40092

## 📔 Objective

SDK biometrics does not use the native messaging background, but it's own connection separately in the ipc background service. Thus, the check is just incorrect, leading to an always disconnected status. This PR fixes it by using the timeout mechanics of the requests that the IPC layer offers. For non-interactive requests such as status, a short timeout is chosen. For interactive requests - biometric unlock - a long timeout is chosen.


apps/browser/src/key-management/biometrics/background-browser-biometrics.service.ts
```diff
@@ -83,10 +85,13 @@ export class BackgroundBrowserBiometricsService extends BiometricsService {
 
   async authenticateWithBiometrics(): Promise<boolean> {
     if (await this.configService().getFeatureFlag(FeatureFlag.BiometricsSDKIPC)) {
-      if (!this.nativeMessagingBackground().connected) {
+      try {
+        return await ipcRequestAuthenticateBiometrics(
+          this.ipcService().client,
+          AbortSignal.timeout(this.BIOMETRICS_INTERACTION_TIMEOUT),
```
mzieniukbw, line 51: ❓ We are only giving users 10 seconds to confirm biometrics prompt ? I don't think that's enough.

apps/browser/src/key-management/biometrics/background-browser-biometrics.service.ts
```diff
@@ -199,14 +196,15 @@ export class BackgroundBrowserBiometricsService extends BiometricsService {
...
   async getBiometricsStatusForUser(id: UserId): Promise<BiometricsStatus> {
     if (await this.configService().getFeatureFlag(FeatureFlag.BiometricsSDKIPC)) {
-      if (!this.nativeMessagingBackground().connected) {
-        return BiometricsStatus.DesktopDisconnected;
-      } else {
+      try {
         const status = await ipcRequestGetBiometricsStatus(
           this.ipcService().client,
           fromTsUserId(id),
+          AbortSignal.timeout(this.BIOMETRICS_NO_INTERACTION_TIMEOUT),
```
mzieniukbw, line 167: ❓ Are there any cases where 0.5 second is not enough ? I.e. browser extension or desktop app just started / unlocked - like ipc connection not established yet

apps/browser/src/key-management/biometrics/background-browser-biometrics.service.ts
quexten: Bumped to 60 seconds!

apps/browser/src/key-management/biometrics/background-browser-biometrics.service.ts
quexten: IPC should establish in the background, completely unrelated to this service. It should never be the case that the ipcRequestGet triggers the connection establishment. And even if so, then "DesktopDisconnected" seems reasonable to give as a status, because it is still connecting (and available at a later date).

mzieniukbw (approved):

PR #21731, opened 2026-07-08
fix(unlock): Biometric unlock shows error dialog when cancelling

### LINKED ISSUE #20721: Bitwarden Desktop (macOS) reports extraneous error when TouchID auth is denied / fails
### Steps To Reproduce

1. Open Bitwarden Desktop - locked vault, with TouchId Unlock configured
2. Unlock with Touch ID
3. Reject the request -- hit Esc or Cancel

### Expected Result

Desktop client should do nothing.

### Actual Result

App pops up an actively distracting, interfering, generic, non-responsive popup. It requires an additional action to dismiss.

### Screenshots or Videos

Image

Image

### Additional Context

Bitwarden makes an attempt to auth unlock when the app opens up—-I don't want my vault open always. I like the app open so I can open the vault quickly--_when I want it open_.

This behavior seems new, even though I haven't updated the app, viz `v2026.3.1` with latest release ([v2026.4.1](https://github.com/bitwarden/clients/releases/tag/browser-v2026.4.1)). I'm wondering if the next phase of the project has arrived, and hope there's still some value for you to capture.

### Operating System

macOS

### Operating System Version

_No response_

### Installation method

Homebrew

### Build Version

Version 2026.3.1 (58867)

### Issue Tracking Info

quexten:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-37737

## 📔 Objective

Currently, when cancelling biometric unlock by denying the authentication, an error shows. Denying authentication is not an error condition and we can drop the dialog. 

There should not be valid conditions that result in an error dialog. When the desktop app is not connected, the button is disabled.


libs/key-management-ui/src/lock/components/lock.component.ts
```diff
@@ -444,41 +444,8 @@ export class LockComponent implements OnInit, OnDestroy {
...
-
-      this.logService.error("[LockComponent] Failed to unlock via biometrics.", e);
-
-      let biometricTranslatedErrorDesc;
-
-      if (this.clientType === "browser") {
-        const biometricErrorDescTranslationKey = this.lockComponentService.getBiometricsError(e);
-
-        if (biometricErrorDescTranslationKey) {
-          biometricTranslatedErrorDesc = this.i18nService.t(biometricErrorDescTranslationKey);
```
mzieniukbw, line 460: ❓  We use this component in web and desktop too, but looks like we might not need some of the internationalisation messages anymore on those ? (still used by extension)

mzieniukbw (approved):

libs/key-management-ui/src/lock/components/lock.component.ts
quexten: Will make a follow-up PR after this PR and the extension split share unlock & use native messaging dialog PR merge. After this we can remove the i18n messages because we should no longer use these errors anywhere.

PR #21736, opened 2026-07-08
Tidy Desktop Autofill Service

iinuwa:
## 📔 Objective

- Remove ts-strict-ignore flag
- Filter null cipher maps, which occur when the vault is actively decrypting.


abergs (approved):
Looks good!

coltonhurst (approved):
Approving for changes in the `preload.ts` file

PR #21742, opened 2026-07-08
[CL-1194] responsive breadcrumbs

BryanCunningham:
Builds on/pulls in some changes from https://github.com/bitwarden/clients/pull/21589 to reduce merge conflicts

## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/CL-1194

## 📔 Objective

- Improve breadcrumb responsive behavior
- Add slot to render bit-icon-tile inside breadcrumb

#### With `bit-icon-tile` in start slot
image

#### Overflow
image


BryanCunningham:
Holding for now as I just realized the focus behavior on breadcrumbs is incorrect

libs/components/src/breadcrumbs/breadcrumb.component.ts
```diff
@@ -93,6 +105,15 @@ export class BreadcrumbComponent {
...
       .subscribe((_) => this.checkActiveRoute());
+
+    // Drive the projected icon tile's size from the crumb size (pushed by the parent
+    // `bit-breadcrumbs`) so it stays in sync. Runs once the content query resolves.
+    effect(() => {
+      const tile = this.startIconTile();
+      if (tile) {
+        tile.size.set(this.size() === "small" ? "xs" : "sm");
+      }
+    });
```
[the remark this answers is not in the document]
BryanCunningham: breadcumb always wins. Fixed docs

libs/components/src/breadcrumbs/breadcrumb.component.html
```diff
@@ -1,6 +1,9 @@
 <ng-template>
-  @if (icon(); as icon) {
-    <bit-icon [name]="icon" class="tw-me-2" />
-  }
-  <ng-content></ng-content>
+  <div class="tw-flex tw-items-center tw-gap-2">
+    <ng-content select="[slot=start]"></ng-content>
+    @if (icon(); as icon) {
+      <bit-icon [name]="icon" />
+    }
```
vleague2, line 6: Wonder if we should commit to one approach -- either open slot to pass in whatever, or only accept icon or icon tile

libs/components/src/breadcrumbs/breadcrumbs.component.spec.ts
```diff
@@ -0,0 +1,80 @@
...
+  function tileContainerClass(): string {
+    const tile = fixture.debugElement.query(By.directive(IconTileComponent));
+    return (tile.nativeElement as HTMLElement).firstElementChild!.className;
+  }
+
+  it("sizes a projected start-slot icon tile to `sm` when the breadcrumbs are `base`", () => {
+    fixture.componentInstance.size.set("base");
+    fixture.detectChanges();
+
+    expect(tileContainerClass()).toContain("tw-w-6"); // sm
```
vleague2: I don't love testing for css classes

libs/components/src/breadcrumbs/breadcrumbs.component.ts
```diff
@@ -15,12 +20,19 @@ import { I18nPipe } from "@bitwarden/ui-common";
...
+  OverflowItemDirective,
+  OverflowListDirective,
+  OverflowTriggerDirective,
+} from "../overflow-list";
 import { TypographyModule } from "../typography";
 
 import { BreadcrumbComponent } from "./breadcrumb.component";
 
+/** Approximate width reserved for the trailing separator arrow (icon + margins), in pixels. */
+const TRAILING_ARROW_RESERVE_PX = 48;
```
vleague2: Does this change in small mode?

libs/components/src/overflow-list/overflow-list.directive.ts
```diff
@@ -92,85 +104,160 @@ export class OverflowListDirective {
...
+function revealForMeasurement(el: HTMLElement): () => void {
+  const prevHidden = el.hidden;
+  const prevDisplay = el.style.display;
+  el.hidden = false;
+  el.style.display = "";
+  return () => {
+    el.hidden = prevHidden;
+    el.style.display = prevDisplay;
+  };
+}
```
vleague2, line 277: I imagine we'll want to do some regression testing around the other components using this directive since it has a lot of changes

libs/components/src/breadcrumbs/breadcrumbs.stories.ts
```diff
@@ -192,3 +230,28 @@ export const WithTrailingArrow: Story = {
     showTrailingArrow: true,
   },
 };
+
+/**
+ * Breadcrumb overflow is width-driven: as the container narrows, the middle crumbs
```
vleague2: These story comments feel like documentation that should be on the mdx page. I don't think these show up anywhere?

libs/components/src/breadcrumbs/breadcrumbs.component.spec.ts
BryanCunningham: Me neither. Stupid Claude. I'll fix

libs/components/src/breadcrumbs/breadcrumb.component.html
BryanCunningham: @vleague2 What would 'only accept icon or icon tile' look like in your mind? I thought about adding another input for icon tile but, we'd then need one more input for icon tile variant, if that is what you are meaning

vleague2: Yeah that's what I was thinking too, which I don't love. I just think it is confusing to have 2 ways to do it -- I could see devs using the start slot for icons too since it's possible, so that's what made me think to standardize it one way or the other

BryanCunningham: @vleague2 I don't have strong feelings about it. I can switch it to only have a slot, if we want. Would just be more consumer code changes

vleague2: Okay I have decided this is not worth pursuing, we are already going the "icon or open slot" on the form control cards anyway

libs/components/src/breadcrumbs/breadcrumbs.component.ts
BryanCunningham: Yes, I've pushed a commit to change it in small mode. It doesn't necessarily break with a constant width here but, it would cause items to pack slightly early

vleague2:
I think we need truncation of the active breadcrumb for the super small screen size use case
Screenshot 2026-07-17 at 9 45 21 AM

vleague2 (dismissed):

vleague2 (dismissed):

vleague2 (dismissed):

libs/components/src/breadcrumbs/breadcrumbs.component.ts
```diff
@@ -70,41 +97,23 @@ export class BreadcrumbsComponent {
...
-
-  /** The last breadcrumb, shown after the overflow menu */
-  protected readonly afterOverflow = computed(() => this.breadcrumbs().at(-1));
+  constructor() {
+    // Push our size down to each child crumb so they can size projected icon tiles in step.
+    effect(() => {
+      const size = this.size();
+      this.breadcrumbs().forEach((breadcrumb) => breadcrumb.size.set(size));
+    });
+  }
```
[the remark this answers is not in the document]
vleague2: @BryanCunningham seems worth fixing?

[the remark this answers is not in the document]
BryanCunningham: Yeah, I'll take a look

vleague2 (dismissed):

vleague2 (approved):

PR #21753, opened 2026-07-08
[PM-40138] desktop-autofill: refactor Autofill IPC Registration

iinuwa:
## 🎟️ Tracking

[PM-40138](https://bitwarden.atlassian.net/browse/PM-40138)

## 📔 Objective

The Autofill IPC mechanism is getting more use now, so we should refactor to make it easier to add more to it.

This PR refactors the Autofill IPC registration:

- Refactors identical registration methods into a single function for each of main process, preload shim, and the renderer process
- Moves correlation between incoming and outgoing channel names, and request and response types into a single file and adds type-checking.
- Uses named callbacks instead of order-dependent function arguments. 

[PM-40138]: https://bitwarden.atlassian.net/browse/PM-40138?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/desktop/desktop_native/napi/index.d.ts
+++ b/apps/desktop/desktop_native/napi/index.d.ts
@@ -184,13 +184,13 @@ export declare namespace autofill {
   }
   export interface AutofillIpcCallbacks {
     /** Function to execute when a passkey registration request is received. */
-    registrationCallback: (error: null | Error, clientId: number, sequenceNumber: number, message: PasskeyRegistrationRequest) => void
-    /** Function to execute when a passkey assertion request is received. */
-    assertionCallback: (error: null | Error, clientId: number, sequenceNumber: number, message: PasskeyAssertionRequest) => void
-    /** Function to execute when a passkey assertion request is received and the UI must not be shown. */
-    assertionWithoutUserInterfaceCallback: (error: null | Error, clientId: number, sequenceNumber: number, message: PasskeyAssertionWithoutUserInterfaceRequest) => void
-    /** Function to execute when a notification of the autofill provider's status is received. */
-    nativeStatusCallback: (error: null | Error, clientId: number, sequenceNumber: number, message: NativeStatus) => void
+  registrationCallback: { (error: null, clientId: number, sequenceNumber: number, message: PasskeyRegistrationRequest): void; (error: Error, clientId: number, sequenceNumber: number, message: null): void; }
+  /** Function to execute when a passkey assertion request is received. */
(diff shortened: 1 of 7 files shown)


iinuwa:
Depends on #21736

apps/desktop/desktop_native/napi/index.d.ts
```diff
@@ -170,9 +170,10 @@ export declare namespace autofill {
      *
      * @param name The endpoint name to listen on. This name uniquely identifies the IPC
      * connection and must be the same for both the server and client. @param callback
-     * This function will be called whenever a message is received from a client.
+     * The functions that will be called whenever a message of the
+     * corresponding type is received from a client.
      */
-    static listen(name: string, registrationCallback: (error: null | Error, clientId: number, sequenceNumber: number, message: PasskeyRegistrationRequest) => void, assertionCallback: (error: null | Error, clientId: number, sequenceNumber: number, message: PasskeyAssertionRequest) => void, assertionWithoutUserInterfaceCallback: (error: null | Error, clientId: number, sequenceNumber: number, message: PasskeyAssertionWithoutUserInterfaceRequest) => void, nativeStatusCallback: (error: null | Error, clientId: number, sequenceNumber: number, message: NativeStatus) => void): Promise<AutofillIpcServer>
```
iinuwa, line 175: @dani-garcia, since you're tagged for review, these NAPI index.d.ts changes are just moving the `listen()` method to use named callbacks instead of a list of arguments to make it a little clearer at the call site which method is being registered as the number of callbacks grows. Cf. https://github.com/bitwarden/clients/pull/21753/changes#diff-e7c048771b6299bb0d0264342ff22c5e3bb19761e34dbfd641a2888b4101bed2R93-R116

abergs (approved):
Finally a sane refactor of this spaghetti 🍝 (of which I am guilty of adding to).

Thank you

dani-garcia (approved):
LGTM, thanks!

apps/desktop/src/autofill/models/ipc-handler.type.ts
```diff
@@ -0,0 +1,21 @@
+export type CompletionCallback<Response> = {
```
neuronull, line 1: Should this file be added to CODEOWNERS as shared ownership ?

neuronull (approved):
`ipc-handler.types.ts` LGTM

apps/desktop/src/autofill/models/ipc-handler.type.ts
iinuwa: Yeah, it could be, but I missed it. To avoid another round of review bingo I'll go ahead and merge and add CODEOWNERS the next time I need to update it. (I don't think I'll need to anytime soon.)

PR #21756, opened 2026-07-08
[PM-37964] Add Ability to Disable Automatic Invitations in SCIM

sven-bitwarden:
## 🎟️ Tracking
[PM-37964](https://bitwarden.atlassian.net/browse/PM-37964)

## 📔 Objective
The penultimate PR for the staged saga, this PR surfaces the UI to provision staged users. Users are able to _disable_ this toggle to prevent provisioned users from being invited, wherein they will land in a staged state.

## 📸 Screenshots
image

[PM-37964]: https://bitwarden.atlassian.net/browse/PM-37964?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/bitwarden_license/bit-web/src/app/admin-console/organizations/manage/scim-v2.component.ts
+++ b/bitwarden_license/bit-web/src/app/admin-console/organizations/manage/scim-v2.component.ts
@@ -19,11 +19,14 @@ import { ScimConfigApi } from "@bitwarden/common/admin-console/models/api/scim-c
 import { OrganizationConnectionRequest } from "@bitwarden/common/admin-console/models/request/organization-connection.request";
 import { ScimConfigRequest } from "@bitwarden/common/admin-console/models/request/scim-config.request";
 import { OrganizationConnectionResponse } from "@bitwarden/common/admin-console/models/response/organization-connection.response";
+import { FeatureFlag } from "@bitwarden/common/enums/feature-flag.enum";
+import { ConfigService } from "@bitwarden/common/platform/abstractions/config/config.service";
 import { EnvironmentService } from "@bitwarden/common/platform/abstractions/environment.service";
 import { I18nService } from "@bitwarden/common/platform/abstractions/i18n.service";
 import { PlatformUtilsService } from "@bitwarden/common/platform/abstractions/platform-utils.service";
 import { OrganizationId } from "@bitwarden/common/types/guid";
 import {
+  A11yTitleDirective,
   AriaDisableDirective,
   BaseCardDirective,
   BitActionDirective,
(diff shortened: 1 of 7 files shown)


bitwarden_license/bit-web/src/app/admin-console/organizations/manage/scim-v2.component.ts
```diff
@@ -201,7 +215,11 @@ export class ScimV2Component {
       this.organizationId(),
       OrganizationConnectionType.Scim,
       true,
-      new ScimConfigRequest(this.enabled.value ?? false),
+      new ScimConfigRequest(
+        this.enabled.value ?? false,
+        null,
```
BTreston: nit: use `undefined` instead of `null`

libs/common/src/admin-console/models/api/scim-config.api.ts
```diff
@@ -6,6 +6,7 @@ import { ScimProviderType } from "../../enums";
 export class ScimConfigApi extends BaseResponse {
   enabled: boolean;
   scimProvider: ScimProviderType;
+  inviteUsersAfterProvisioning: boolean | null;
```
BTreston: ⛏️ `null` => `undefined`

libs/common/src/admin-console/models/request/scim-config.request.ts
```diff
@@ -5,6 +5,7 @@ import { ScimProviderType } from "../../enums";
 export class ScimConfigRequest {
   constructor(
     private enabled: boolean,
-    private scimProvider: ScimProviderType = null,
+    private scimProvider: ScimProviderType | null = null,
```
BTreston: ⛏️ `null` => `undefined`

libs/common/src/admin-console/models/request/scim-config.request.ts
BTreston: non blocking: we could also address this boy scout style since it looks super easy.

package-lock.json
BTreston, line 1: Did you mean to commit this file? We probably don't need it looking at the changes in the PR

BTreston (changes requested):
Looks 99% good. Just some very minor nit ⛏️ s

package-lock.json
sven-bitwarden: I was getting dependency install errors (see: https://github.com/bitwarden/clients/actions/runs/28977514964/job/85988200695?pr=21756). I'm assuming it had to do with the node upgrade yesterday morning, so I upgraded my local node, regenerated, pushed up the new package lock, it seemed much happier after that

sven-bitwarden: I've removed it from the changelist - the one on main works juuust fine now after rebasing.

BTreston (approved):
🚀

PR #21761, opened 2026-07-09
[PM-34857] Remove PoliciesInAcceptedState feature flag

eliykat:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-34857

## 📔 Objective

Remove the `pm-34145-policies-in-accepted-state` feature flag.

This PR is intentionally narrow - it removes the feature flag, and moves the new SDK logic into the main policy service. However, it keeps the new policy service to manage the new policy state, and continues tracking the old policy state separately. 

This is because the old policy state is still being used by the `policies$` observable, which is a public interface - we need to review those consumers and make sure that changing this over to the new state (accepted AND confirmed policies) won't affect them. Then we need to migrate the state keys to drop the new suffix. This follow-up work is tracked in follow-up tasks in https://bitwarden.atlassian.net/browse/PM-39299. For now, just get rid of the feature flag.

Server PR: https://github.com/bitwarden/server/pull/7947


dereknance (dismissed):
✅ Platform

dereknance (approved):

sven-bitwarden (approved):

PR #21764, opened 2026-07-09
PM-14801 replace UserKeyEncryptor with SubjectKeyEncryptor in UserSta…

bmbitwarden:
…teSubject

## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-14801

## 📔 Objective

Replace `UserKeyEncryptor` with `SubjectKeyEncryptor` in `UserStateSubject`

### DIFF
--- a/libs/common/src/tools/cryptography/legacy-encryptor-provider.ts
+++ b/libs/common/src/tools/cryptography/legacy-encryptor-provider.ts
@@ -10,6 +10,7 @@ import {
 } from "../dependencies";
 
 import { OrganizationEncryptor } from "./organization-encryptor.abstraction";
+import { SubjectKeyEncryptor } from "./subject-key-encryptor.abstraction";
 import { UserEncryptor } from "./user-encryptor.abstraction";
 
 /** Creates encryptors
@@ -28,6 +29,21 @@ export abstract class LegacyEncryptorProvider {
     frameSize: number,
     dependencies: SingleUserDependency,
   ) => Observable<UserBound<"encryptor", UserEncryptor>>;
+
+  /** Retrieves an encryptor bound to a specific subject within the user's data.
+   *  The encryptor uses a padded data packer to encode data and is identified
+   *  by a subject id that uniquely identifies the subject being encrypted.
+   *  @param frameSize length of the padded data packer's frames.
+   *  @param subjectId uniquely identifies the subject being encrypted within the user's data.
+   *  @param dependencies.singleUserId$ identifies the user to which the encryptor is bound
+   *  @returns an observable that emits when the key becomes available and completes
+   *    when the key becomes unavailable.
+   */
+  subjectEncryptor$: (
+    frameSize: number,
+    subjectId: string,
+    dependencies: SingleUserDependency,
+  ) => Observable<UserBound<"encryptor", SubjectKeyEncryptor>>;
 
   /** Retrieves an encryptor populated with the organization's most recent key instance that
    *  uses a padded data packer to encode data.
--- a/libs/common/src/tools/cryptography/subject-key-encryptor.abstraction.ts
+++ b/libs/common/src/tools/cryptography/subject-key-encryptor.abstraction.ts
@@ -1,0 +1,41 @@
+// FIXME: Update this file to be type safe and remove this and next line
+// @ts-strict-ignore
+import { Jsonify } from "type-fest";
+
+import { EncString } from "../../key-management/crypto/models/enc-string";
+import { UserId } from "../../types/guid";
+
+/** An encryption strategy that protects a type's secrets using a key
+ *  specific to the subject being encrypted. This strategy is bound to a
+ *  specific user and a specific subject within that user's data.
+ *
+ *  The subject id uniquely identifies the subject being encrypted and is
+ *  used to identify the key(s) that perform decryption on behalf of the
+ *  subject. This allows the UserStateSubject to grow to support collections
+ *  like Record<Guid,T> and Array<T>.
+ */
+export abstract class SubjectKeyEncryptor {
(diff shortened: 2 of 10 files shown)


libs/common/src/tools/cryptography/user-subject-key-encryptor.ts
```diff
@@ -0,0 +1,52 @@
+// FIXME: Update this file to be type safe and remove this and next line
+// @ts-strict-ignore
```
quexten, line 2: Is there a reason we are adding new code that has ts-strict ignore on?

libs/common/src/tools/cryptography/user-subject-key-encryptor.ts
```diff
@@ -0,0 +1,52 @@
...
+import { Jsonify } from "type-fest";
+
+import { EncString } from "../../key-management/crypto/models/enc-string";
+import { UserId } from "../../types/guid";
+
+import { SubjectKeyEncryptor } from "./subject-key-encryptor.abstraction";
+import { UserEncryptor } from "./user-encryptor.abstraction";
+
+/** A classification strategy that protects a type's secrets by encrypting them
+ *  with the user key. The user key is provided by a UserEncryptor.
```
quexten, line 12: This comment is not correct. UserEncrypt encrypts with the localUserDataKey, not the UserKey. This was changed a while ago, and during the migration period they may be the same key, but this will break / not be true at some point.

libs/common/src/tools/cryptography/legacy-encryptor-provider.ts
```diff
@@ -29,6 +30,21 @@ export abstract class LegacyEncryptorProvider {
     dependencies: SingleUserDependency,
   ) => Observable<UserBound<"encryptor", UserEncryptor>>;
 
+  /** Retrieves an encryptor bound to a specific subject within the user's data.
```
quexten, line 33: Actually, "bound to a specific subject" is not correct, if we use the user encryptor, since it is bound to both the subject (the user) *and* the device.

quexten (commented):
Only some comment concerns, I think since the jira ticket got created things changed which is why it doesn't seem up-to-date?

We should be clear about the terms here because otherwise we risk confusion / incorrect usage in later PRs. The userkey is user-scoped and may be changed by another device during key rotation. The user encryptor encrypts locally scoped data. Locally scoped data cannot be encrypted by a user-scoped key, since the user-scoped key may changed later, and since we don't keep a history of old keys.

Because of this, these keys were split here: https://github.com/bitwarden/clients/pull/19433 and these are two separate keys - "UserKey", "LocalUserDataKey".

(These used to be the same key, however the linked PR changed that).

libs/common/src/tools/cryptography/legacy-encryptor-provider.ts
audreyality: As I'd originally envisioned this, the subject is the data being encrypted (e.g. email forwarder settings). More specifically, "subject" refers to [UserStateSubject](https://github.com/bitwarden/clients/blob/main/libs/common/src/tools/state/user-state-subject.ts), an [rxjs SubjectLike](https://rxjs.dev/api/index/interface/SubjectLike) wrapping our [SingleUserState](https://github.com/bitwarden/clients/blob/main/libs/state/src/core/user-state.ts#L46). The subject encryptor uses an intermediate key, which it protects using the user's encryption key.

The intermediate key could be device specific, ofc. I'd originally envisioned it would be that way once we supported it. I just want to make sure we keep our terms straight. :)

When I wrote the ticket, the encryption logic was still all managed through JS, which is why I'd envisioned this would accept a user encryptor. If there's now a way that the encryption APIs can decrypt the intermediate key implicitly, then there's no need to handle the user-level encryption key, and the user encryptor can be replaced with a direct reference to an account id. That, IMO, is a more secure design, and makes the implementation much simpler.

libs/common/src/tools/cryptography/user-subject-key-encryptor.ts
```diff
@@ -0,0 +1,52 @@
...
+    this.assertHasValue("userEncryptor", userEncryptor);
+  }
+
+  async encrypt<Secret>(secret: Jsonify<Secret>): Promise<EncString> {
+    return this.userEncryptor.encrypt(secret);
+  }
+
+  async decrypt<Secret>(secret: EncString): Promise<Jsonify<Secret>> {
+    return this.userEncryptor.decrypt(secret);
+  }
```
audreyality, line 45: ⚠️ The purpose of this ticket is to introduce an intermediate key, which is itself encrypted by the user key. As-written, this just introduces indirection. It doesn't implement the acceptance criteria.

libs/common/src/tools/cryptography/subject-key-encryptor.abstraction.ts
```diff
@@ -0,0 +1,41 @@
...
+/** An encryption strategy that protects a type's secrets using a key
+ *  specific to the subject being encrypted. This strategy is bound to a
+ *  specific user and a specific subject within that user's data.
+ *
+ *  The subject id uniquely identifies the subject being encrypted and is
+ *  used to identify the key(s) that perform decryption on behalf of the
+ *  subject. This allows the UserStateSubject to grow to support collections
+ *  like Record<Guid,T> and Array<T>.
+ */
+export abstract class SubjectKeyEncryptor {
```
audreyality, line 17: ❌ This interface needs to support key rotation. There are no endpoints available to perform this operation.

audreyality (changes requested):

bmbitwarden:
@itsadrago  we decided to put this ticket in draft to further discuss recent related progress since this ticket was first created

PR #21765, opened 2026-07-09
desktop-autofill: Move autofill services and models from platform team

iinuwa:
## 📔 Objective

Moves Desktop Autofill main process code into Desktop Native + Skunkworks ownership from platform.


dani-garcia (dismissed):

iinuwa:
Added Skunkworks as preload owners for desktop-autofill features and fixed merge conflict.

coltonhurst (commented):
@iinuwa overall looks good. Two questions, just to double check:

1. For the files we are taking ownership of from platform, is Skunkworks taking priority responsibility on those while native passkeys are being worked on (like discussed in the agreement), and then we can include them in the transition when we transition things over to desktop native?
2. Ideally there are no issues, but were you able to do some "sanity check" testing?

iinuwa:
@coltonhurst, yes to both. The main risk is that the app doesn't start at all or the SSH agent doesn't start because of a typo. I confirmed that both of these are still working, as well as the desktop autofill code when the feature flag is enabled.

coltonhurst (approved):

abergs (approved):

dani-garcia (approved):

PR #21768, opened 2026-07-09
[PM-40259] introduce `AutofillOrchestrator`

audreyality:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-40259

## 🏁 Objective

All "do an autofill" signals will now run through `AutofillOrchestrator`. This class reconciles user requests for autofills, including "autofill on page load" requests. The benefit of this is that `AutofillOrchestrator` can decide "is it the right moment to autofill?" It provides a single-source-of-truth for the decision, ensuring that monitoring and autofill signals are properly interleaved and providing a convenient site for monitoring.

This logic was originally handled by `RuntimeBackground`. It's been moved into its own class because running autofill only on active tabs will make the decision considerably more complex. 

> [!NOTE]
> This PR refactors the existing autofill support. Active tab management and its attendant retry logic are not included in this PR.

## 📚 Documentation

* [autofill README](https://github.com/bitwarden/clients/blob/autofill/pm-39579/only-monitor-active-tabs/apps/browser/src/autofill/README.md) (new)
* [lifecycle documentation](https://github.com/bitwarden/clients/blob/autofill/pm-39579/only-monitor-active-tabs/apps/browser/src/autofill/lifecycle.design.md)
* [autofill documentation](https://github.com/bitwarden/clients/blob/autofill/pm-39579/only-monitor-active-tabs/apps/browser/src/autofill/autofill.design.md) (new)

> [!IMPORTANT]
> These documents reflect the design after the only-monitor-active-tab optimization is complete. This PR follows the architecture described in the above documents, but **does not** implement the active tab optimization described in the lifecycle documentation.

## 🚧 Dependencies

#21909


apps/browser/src/autofill/services/autofill.service.ts
```diff
@@ -117,13 +117,15 @@ export default class AutofillService implements AutofillServiceInterface {
...
+   * Collects page details from a tab. Returns an observable that builds the results from the
+   * collectPageDetailsResponse messages the tab's frames send back. When `frameId` is given, the
+   * collection is scoped to that one frame — only it is asked, and only its response is admitted, so
+   * a concurrent collection on another frame of the same tab cannot bleed in.
    *
    * @param tab The tab to collect page details from
+   * @param frameId When set, collect only this frame; otherwise collect every frame
    */
-  collectPageDetailsFromTab$(tab: chrome.tabs.Tab): Observable<PageDetail[]> {
+  collectPageDetailsFromTab$(tab: chrome.tabs.Tab, frameId?: number): Observable<PageDetail[]> {
```
audreyality, line 129: 📝 Added `frameId` to `collectPageDetailsFromTab$` so that page load triggers, which are frame specific, don't cause a full-page collection for every frame on the page.

apps/browser/src/autofill/rx/sync-scope-flag.ts
audreyality, line 1: 📝 `sync-scope-flag` is an rx utility that detects if someone adds asynchronous code into a pipeline.

apps/browser/src/autofill/background/autofill-orchestrator.ts
```diff
@@ -0,0 +1,271 @@
...
+          if (liveTab == null) {
+            return;
+          }
+
+          // FIXME (PM-39579): the tab gate replaces this. Until then, keep filling only the active
+          // tab, since resolving the target by id would otherwise fill a background tab.
+          const activeTab = await BrowserApi.getTabFromCurrentWindow();
+          if (activeTab?.id !== request.tabId) {
+            return;
+          }
```
audreyality, line 189: 📝 The "tab gate" this refers to is only running monitoring on active tabs. When dispatch only runs as a consequence of the user switching tabs, this check ends up being a prerequisite for pageTransitionResolved$ to emit, so it can be removed from the dispatch body.

apps/browser/src/autofill/services/autofill-lifecycle.service.ts
```diff
@@ -55,18 +65,102 @@ export class DefaultAutofillLifecycleService implements AutofillLifecycleService
...
+  private readonly assertSynchronousConnectedPorts = assertSynchronousScope(
+    "autofill-lifecycle:connectedPorts-fold",
+    (message) => this.logService.error(message),
+  );
+
+  /** Report error if a future change introduces async behaviors to `this.monitorLifecycle`. */
+  private readonly assertSynchronousMonitoringState = assertSynchronousScope(
+    "autofill-lifecycle:monitoringState-fold",
+    (message) => this.logService.error(message),
+  );
```
audreyality, line 154: 📝 These assertions ensure that events linked to ports remain synchronous, which is important for cases where you want to reply to a message.

🎨 A throw could be used in the callabck instead of logging. Doing so would tear down the port subscriptions, which would disable fill page on load completely if the guarded regions were to become asynchronous.

apps/browser/src/autofill/services/autofill-lifecycle.service.ts
```diff
@@ -55,18 +65,102 @@ export class DefaultAutofillLifecycleService implements AutofillLifecycleService
...
+              (lifecycle) =>
+                !lifecycle.active &&
+                this.monitorFrameKey(lifecycle.tabId, lifecycle.frameId) === key,
+            ),
+          ),
+        ),
+      );
+    }),
+    share({ resetOnRefCountZero: true }),
+  );
```
audreyality, line 129: 📝 Moved from line 119 because this is now stateless

apps/browser/src/autofill/background/autofill-orchestrator.ts
```diff
@@ -0,0 +1,271 @@
...
+
+  /**
+   * Re-resolves the reported frame's tab live by id and confirms the frame still shows the URL it
+   * reported. Returns the live tab when the fill may proceed, or undefined to abandon.
+   */
+  private async resolveFreshTarget(
+    request: Extract<FillRequest, { kind: "pageLoad" }>,
+  ): Promise<chrome.tabs.Tab | undefined> {
+    // `getTab` may return null synchronously (no tab id) or a promise that rejects (tab gone);
+    // both collapse to a clean abandon rather than a logged error.
```
bensbits91, line 238: Question / nit: 

This comment suggests that `getTab` might return `null` synchronously. However, `getTab` is `async`so it always returns a `Promise` — the `?.catch` optional-chain short-circuit won't trigger.

I think the behavior is correct but the comment is a tiny bit misleading. Please disregard if I'm wrong or just too nit-picky.

bensbits91 (dismissed):
Looks good! One minor question/nit about a comment.

apps/browser/src/autofill/background/autofill-orchestrator.ts
audreyality: This is a consequence of its type signature. The code cannot return `null` synchronously, but the return type says that it can. "May" here means "is allowed to".

https://github.com/bitwarden/clients/blob/c786476c8dc3afd7b443c2530d20ed3ea2ca366d/apps/browser/src/platform/browser/browser-api.ts#L250

bensbits91: Makes sense, thanks for explaining. My approval stands :)

apps/browser/src/autofill/README.md
jprusik, line 1: Do we have a strategy for ensuring these docs stay up-to-date?

apps/browser/src/autofill/autofill.design.md
```diff
@@ -0,0 +1,89 @@
...
+
+- **The autofill-on-page-load setting must be enabled.** It is off by default, and a user who has not
+  opted in gets no page-load fill even on a committed, monitored frame. The monitoring lifecycle
+  gates the _autofiller's injection_ on this same setting, but an already-injected autofiller is not
+  re-evaluated when the setting changes — it keeps reporting transitions until logout or context
+  loss — so this fill-time check, not the injection-time gate, is what enforces the setting when a
+  user toggles it off mid-session.
+- **A cipher must match the frame's page.** With no match there is nothing to fill, and the
+  opportunity is discarded without collecting page details.
+- **The frame's trust must permit filling.** A page-load fill into an untrusted iframe is refused as
```
jprusik, line 34: Presumably this triggers the existing user-prompt about cross-origin frame fills (that is, the user still has the option to continue to the fill)?

apps/browser/src/autofill/README.md
audreyality: Not yet. The idea is for these docs to serve a dual-role. One side is as orientation to explore and understand autofill systems. The other is as reference/context material for LLMs.

The trouble is that there's a chicken-and-egg situation. You need to guide the LLM how to write the docs, but you need the docs to know how to guide the LLM.

The approach I'm taking is to do the docs first. Once we have them seeded, then we can add the skills and rules needed to keep them up-to-date. I have a few things that can serve as guidance on how to write the docs starting to coalesce. Rules and skills are on the horizon. :)

apps/browser/src/autofill/lifecycle.design.md
```diff
@@ -72,35 +73,72 @@ Services that own both reversible and terminal work expose both methods. The ide
 
 This composition keeps each scope focused. Anything reversible belongs to monitoring; anything that requires graph-wide teardown belongs to disposal. The two never entangle.
 
+## The tab lifecycle
+
+The browser reports one fact about attention: which tab is active. A tab is active when it is the active tab **in its own window**, and autofill reads that fact per window — window focus is not consulted, so alt-tabbing between two windows changes neither window's active tab and leaves autofill undisturbed.
```
jprusik, line 78: > window focus is not consulted, so alt-tabbing between two windows changes neither window's active tab and leaves autofill undisturbed.

I think this is potentially undesirable (we do care about window focus as well); could this lead to a situation where the inline menu is hanging open for an extended time while the user works in another (non-browser) window? Or is focus considered elsewhere?

Another thought; can there ever be more than one active tab? Can there be _no_ active tab?

apps/browser/src/autofill/lifecycle.design.md
```diff
@@ -72,35 +73,72 @@ Services that own both reversible and terminal work expose both methods. The ide
...
+
+Autofill does not act on that raw fact directly. It maintains three states over it:
+
+- **Away** — the resting state of any tab the user is not working in. Autofill does nothing: no monitoring runs, and the tab holds none of the monitoring-scoped resources `stopMonitoring()` clears, so it costs nothing on its host page. A freshly-activated tab stays away until it has been active long enough to commit.
+- **Committed** — the tab autofill treats as the one the user is working in. A tab becomes committed once it has stayed active long enough to be more than a tab passed through. Monitoring runs, and filling happens only here, so a fill lands where the user is actually looking.
+- **Cooling down** — a committed tab the user has just left. Autofill keeps monitoring it briefly, so a quick return finds monitoring already in flight, but the tab is no longer committed, so nothing fills there. When the cool-down elapses the tab falls back to away and monitoring stands down.
```
jprusik, line 84: > Autofill keeps monitoring it briefly... but the tab is no longer committed, so nothing fills there

If think this is answered in the affirmative later in the doc, but just to be sure: does the tab have to "cool" fully before accruing active time to become "committed"?

apps/browser/src/autofill/lifecycle.design.md
```diff
@@ -72,35 +73,72 @@ Services that own both reversible and terminal work expose both methods. The ide
...
+
 ## The page lifecycle
 
-A page-lifecycle monitor watches for the moments a page becomes ready to act on — its load, and the navigations that follow — and reports each as a transition. It does not examine field data and is **not** an `AutofillMonitor`. The autofiller (`apps/browser/src/autofill/content/autofiller.ts`) is the current monitor of this lifecycle: it polls for URL changes and, on each transition, reports a `pageTransitionDetected` fact to the background.
+A page-lifecycle monitor watches for the moments a page becomes ready to act on — its load, and the navigations that follow — and reports each as a transition. It does not examine field data and is **not** an `AutofillMonitor`. The autofiller (`apps/browser/src/autofill/content/autofiller.ts`) is the current monitor of this lifecycle. The browser surfaces no reliable signal for single-page-app navigation, so the autofiller synthesizes these transitions itself: it polls for URL changes and reports each as a `pageTransitionDetected` fact to the background. Like the tab lifecycle's committed and cooling-down states, a page transition is a virtual state autofill maintains, not a fact the browser hands it.
```
jprusik, line 92: > The browser surfaces no reliable signal for single-page-app navigation

Mostly agree, but it depends on the SPA. Some emit events like [PopState](https://developer.mozilla.org/en-US/docs/Web/API/PopStateEvent).

apps/browser/src/autofill/lifecycle.design.md
```diff
@@ -72,35 +73,72 @@ Services that own both reversible and terminal work expose both methods. The ide
...
 
 There is no `enableAutofiller` message. Re-enabling happens by re-injection on the next page-load when the user is unlocked. The autofiller's content-script lifecycle, in full: _inject (when unlocked) → report transitions → (disable on logout | dispose on context loss)_.
 
 ### Buffering transitions
 
-Autofill can only fill a frame that is monitoring — monitoring is what makes the page details available to act on. Autofill-on-load therefore depends on monitoring, and the two are not ordered against each other: injection adds the autofiller, which begins reporting at page load, while the `start monitors` command follows separately. A transition can be reported before monitoring has started on a freshly-injected frame.
+Autofill can only fill a frame that is monitoring — monitoring is what makes the page details available to act on. A page-load fill therefore depends on monitoring, and the two are not ordered against each other: injection adds the autofiller, which begins reporting at page load, while the `start monitors` command follows separately. A transition can be reported before monitoring has started on a freshly-injected frame.
```
jprusik, line 107: Does any of this meaningfully change if autofill-on-page-load doesn't exist?

apps/browser/src/autofill/lifecycle.design.md
jprusik, line 1: Is this design compatible with our MV2 builds?

apps/browser/src/autofill/autofill.design.md
audreyality: ❓ ~~I didn't change the way that we determine whether a frame is trusted, so this should continue working. That said, there doesn't seem to be a test-case for an untrusted iframe on `webtests.dev`, so I didn't test for it explicitly. Is there a good website I can use to try this out?~~

🦖 The one case where the new code would reject this is [if the untrusted frame's URL changes](https://github.com/bitwarden/clients/blob/286eed657e06e9f5f3f69e0d087481ae6969b025/apps/browser/src/autofill/background/autofill-orchestrator.ts#L202-L204) between the time the autofill request was issued and when it was processed. [doAutofill checks to make sure the tab hasn't changed](https://github.com/bitwarden/clients/blob/286eed657e06e9f5f3f69e0d087481ae6969b025/apps/browser/src/autofill/services/autofill.service.ts#L473-L475). This is necessary to prevent an attacker page from racing an `iframe src` change against autofill data collection.

apps/browser/src/autofill/lifecycle.design.md
audreyality: > could this lead to a situation where the inline menu is hanging open for an extended time while the user works in another (non-browser) window?

Changes to the tab lifecycle only change monitor state. Monitors close the inline menu when they stop. They do not control the menu when they are running. The inline menu remains responsible for managing its own open/closed state. This doesn't change that behavior.

> Another thought; can there ever be more than one active tab?

The active tab is the one that's selected in each window. You have at least as many active tabs as you do windows. I suspect that with FF split windows, you can have more than one active tab per window, but I haven't confirmed this.

The tab lifecycle is used to determine when we are running monitors and when we aren't. The implication to "window focus is not consulted" is that we run monitors on all active tabs, whether they're the user's focus or not. That is, the consequence of one or more unfocused active tabs is that we're consuming compute and memory resources on all of them.

> Can there be no active tab?

When there are no windows, there are no active tabs. There are also no active content scripts, so there's no monitors to track or toggle.

apps/browser/src/autofill/lifecycle.design.md
audreyality: It becomes committed as soon as the user returns to it. The purpose of the cooling down state is two-fold:

* Avoid cold-starts when the user is flipping between tabs.
* Prohibit background fills of the tab.

apps/browser/src/autofill/lifecycle.design.md
audreyality: That signal is unreliable because it depends upon the SPA to use the history API. The reason why the autofiller needs to synthesize the transition is because it is not reliable across SPAs.

apps/browser/src/autofill/lifecycle.design.md
audreyality: If the transitions are never reported (e.g. autofiller.ts is not loaded), then this is inert.

apps/browser/src/autofill/lifecycle.design.md
jprusik: So it sounds like "no", actually? That is, the cool state can be active alongside a committed state (this feels better to me)?

apps/browser/src/autofill/lifecycle.design.md
audreyality: As far as I'm aware, yes. MV2 is easier to work with than MV3, since all of the extension APIs this relies upon are cross-compatible and MV2 has an always-on background page. This doc only calls out MV3 because it requires we think about losing state when the extension's service worker shuts down. 

If there are specific concerns you have regarding MV2, I can review them.

apps/browser/src/autofill/lifecycle.design.md
audreyality: Correct. On any given tab, only one state exists. One tab can be cooling down while another is committed.

jprusik: > only one state exists. One tab can be cooling down while another is committed

Sorry, I meant on a single tab context. If a user leave a tab and it begins cooling down, and they return to the tab before the cool down has completed, what happens? My read thus far is that it will exit the "cooling" state immediately once the "committed" state has begun (or does the committed state "wait" for the "cooling" state to end... or something else?).

jprusik (dismissed):

apps/browser/src/autofill/autofill.design.md
jprusik: > Is there a good website I can use to try this out?

I've DMed you an example

apps/browser/src/autofill/lifecycle.design.md
audreyality: It'll immediately switch back to committed. If the tab you just left was committed, it will enter cooling down. If the tab you just left was away, nothing happens.

apps/browser/src/autofill/autofill.design.md
audreyality: 📝 I worked through the example. The behavior hasn't changed.

* When a credential is tracked to `_top` but not the iframe URL, a confirmation popup is displayed.
* When a credential is tracked to an iframe and `_top` is not associated with the credential, there is no valid credential to fill.
* When a credential is tracked to `_top` and *also* the iframe URL, we fill.

blackwood (approved):

PR #21769, opened 2026-07-09
[PM-38189] Wire InitiationPath to Organizations to Correct Trial Metadata

sbrown-livefront:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-38189

## 📔 Objective

Existing users who receive a trial via `trial/send-verification-email` land on `/#/create-organization` (the `trial-initiation` routes are behind `unauthGuardFn()`). Completing org creation there tagged their Stripe subscription `trialInitiationPath = product-initiated` instead of `marketing-initiated`.

The server derives `trialInitiationPath` from the org create request's `initiationPath`. `OrganizationPlansComponent.createCloudHosted` hardcoded it to `"New organization creation in-product"`, with no way to signal a marketing origin.

**Fix:** derive `initiationPath` from the `product` query param — the same signal the new-user trial flow already uses — and thread it through:

- `OrganizationPlansComponent` gets an `initiationPath` input (defaulting to `NewOrganizationCreationInProduct`, so other consumers are unchanged) and uses it in `createCloudHosted`.
- `CreateOrganizationComponent` sets the marketing value when a `product` param is present (in-product navigations here are param-less, so its presence marks a marketing trial).

### Notes:
- Reuses the existing `initiationPath` field and the `product` param the deep link already carries. 
- Consolidates the duplicated `InitiationPath` type into a const enum-like at `libs/common/src/billing/enums/initiation-path.enum.ts`.


libs/common/src/models/request/reference-event.request.ts
sbrown-livefront, line 1: 📝 I'm thinking this can be deleted. I don't see it being used in the code, but wanted to make sure I didn't overlook any hidden usage.

jrmccannon (dismissed):
Looks fine from AC's perspective

amorask-bitwarden (dismissed):

amorask-bitwarden (approved):

jrmccannon (approved):

PR #21774, opened 2026-07-09
desktop: Add function to retrieve native window handle

iinuwa:
## 🎟️ Tracking

[PM-29792](https://bitwarden.atlassian.net/browse/PM-29792)

## 📔 Objective

This adds an IPC method to get details about the native OS window. This is needed specifically in the Windows passkey plugin authenticator for passing focus between the passkey client (browser), Electron app, and the Windows Hello user verification dialog.

Cf. https://github.com/bitwarden/clients/blob/skunkworks/PM-29792/autofill-query-window-and-lock-status/apps/desktop/src/autofill/services/desktop-autofill.service.ts#L283 for where we intend to call it, and the eventual consumption in https://github.com/bitwarden/clients/blob/main/apps/desktop/desktop_native/windows_plugin_authenticator/src/authenticator.rs#L244-L258.

[PM-29792]: https://bitwarden.atlassian.net/browse/PM-29792?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/desktop/src/main/window.main.ts
+++ b/apps/desktop/src/main/window.main.ts
@@ -135,6 +135,14 @@ export class WindowMain {
       }
     });
 
+    ipcMain.handle("get-native-window-details", (_event) => {
+      return {
+        isVisible: this.win.isVisible(),
+        isFocused: this.win.isFocused(),
+        handle: this.win.getNativeWindowHandle().toString("base64"),
+      };
+    });
+
     this.desktopSettingsService.modalMode$
       .pipe(
         pairwise(),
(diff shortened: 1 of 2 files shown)


iinuwa:
Now that I've refactored this, I realize that my use case doesn't need to expose this to other platformss, I already have a reference to WindowMain in my service, so I can call these directly.

Closing.

PR #21776, opened 2026-07-09
[CL-972] card updates

BryanCunningham:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/CL-972

## 📔 Objective
- Minor style updates to cards
- Adds a segmented card

## 📸 Screenshots
image

image


libs/components/src/card/segmented-card.component.ts
```diff
@@ -0,0 +1,24 @@
...
+ * its direct children. Makes no assumptions about segment content — project any
+ * elements you like; each segment is responsible for its own padding and layout
+ * (`<bit-card-content>` is a convenient choice for standard card padding).
+ *
+ * Note: dividers are drawn between *direct* children (Tailwind `divide-y`), so each
+ * segment must be a direct child of `bit-segmented-card`. `@if` / `@for` blocks are fine;
+ * wrapping segments in an intermediate element removes the dividers.
+ */
+@Component({
+  selector: "bit-segmented-card",
```
vleague2: Does this need to be a fully separate component? Is this something we could build into the card, where using multiple `bit-card-content` children automatically applies the segmented appearance, while just one of those children is the normal card?

BryanCunningham: This just felt more discoverable/explicit to me. Using multiple card content children _could_ work. But, feels sort of magical and might come as a surprise if, for some reason folks didn't want/need the segments. I'm open to whatever we think is best.

libs/components/src/card/card.component.ts
vleague2, line 21: Can we also add a jsdoc comment to this component directing people to the segmented one for that use case? For more discoverability on the dev side

BryanCunningham: Yep, I can get something wrote up

vleague2 (approved):

PR #21777, opened 2026-07-09
[PM-27060] feat: Call SDK for password prelogin

ike-kottlowski:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-27060
related server PR: https://github.com/bitwarden/server/pull/7954

## 📔 Objective

Adds a feature-flagged path for `DefaultPasswordPreloginService` to fetch prelogin data (KDF config) via the SDK's `auth().login().get_password_prelogin()` instead of the legacy API call. Behavior is gated behind `FeatureFlag.PM27060_PasswordPreloginFromSdk` (default off) so the SDK path can be rolled out safely; the legacy API path remains the fallback. `DefaultPasswordPreloginService` now takes `SdkService`, `EnvironmentService`, and `ConfigService`.   DI wiring updated in the CLI service container and the Angular `jslib-services.module.ts`.

## 📸 Screenshots

https://github.com/user-attachments/assets/65b23a9a-85ad-47ca-af01-60281f94cdb0

Chromium browser with dev tools opens shows a user trying to log in. The user clicks continue and in dev tools we can see the break point for the feature flag being on is hit indicating that the `PasswordPreLoginRequest` goes through the SDK. In the network tab with the dev-tools we can see the `PasswordPreLoginRequest` comes back as we expect. The user continues to login successfully showing functional parody with legacy.


enmande (approved):
Looks like a clean stand-in; thank you for the tests that express the flagging behavior!

PR #21790, opened 2026-07-10
[CL-1274] Add bit-page layout region for bit-layout main content

willmartian:
## Summary

Introduces `bit-page`: a full-height flex column for a `bit-layout`'s main content area, whose body fills the available height and owns scrolling.

- **Fill content** dropped in (e.g. a `height="fill"` table) grows to the available height and scrolls internally rather than growing the page; regular flowing content scrolls in the body.
- `bit-layout` drops its own main-content padding when a `bit-page` is present (`has-[bit-page]:tw-p-0`), so `bit-page` owns the body padding and the scrollbar sits at the true edge.
- **Not promoted to `<main>` yet** — `bit-layout` still renders the page's `<main>` landmark, and two would be invalid. `bit-page` becomes `<main>` once every page adopts it and `bit-layout` drops its own.

## Notes

- Split out of #20900 (bit-table-v2) as an independent, standalone change. It's the intended fill container for `bit-table-v2 height="fill"`, but stands alone as a general layout primitive.
- The one-line `has-[bit-page]:tw-p-0` hook on `bit-layout` is progressive enhancement — inert on pages that don't use `bit-page`.

## Testing

- Storybook: **Component Library/Layout/Page → Default**.

### DIFF
--- a/libs/components/src/layout/index.ts
+++ b/libs/components/src/layout/index.ts
@@ -1,3 +1,4 @@
 export * from "./layout.component";
+export * from "./page.component";
 export * from "./layout-footer.service";
 export * from "./scroll-layout.directive";
--- a/libs/components/src/layout/layout.component.html
+++ b/libs/components/src/layout/layout.component.html
@@ -48,7 +48,7 @@
     [id]="mainContentId"
     tabindex="-1"
     bitScrollLayoutHost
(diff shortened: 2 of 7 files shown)


BryanCunningham:
@willmartian This seems fine. Did we double check that this change doesn't have any negative effect on drawers?

BryanCunningham (dismissed):

willmartian:
> @willmartian This seems fine. Did we double check that this change doesn't have any negative effect on drawers?

@BryanCunningham Yes, verified.

BryanCunningham (approved):

PR #21797, opened 2026-07-10
[PM-38796] Add link confirm API

JimmyVo16:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-38796

## 📔 Objective

1. Add a method to call the link confirmation endpoint.
2. Add logic to map server errors to OrganizationInviteLinkConfirmRequest so Auth can consume them and translate the errors into localized, user-friendly messages.
3. Add test coverage.


libs/organization-invite-link/src/models/requests/organization-invite-link-confirm.request.ts
```diff
@@ -0,0 +1,18 @@
+export class OrganizationInviteLinkConfirmRequest {
+  code: string;
+  orgUserKey: string;
+  resetPasswordKey?: string;
```
quexten, line 4: Ideally, we would give these more clear names and types, currently this is a bit confusing to read / easy to mistake. I don't know what an "orgUserKey" is.
```suggestion
  // The organization symmetric key, encapsulated for the user's public key
  encapsulatedOrganizationKey: UnsignedSharedKey;
  // The user symmetric key, encapsulated for the organization's public key
  // for account recovery
  encapsulatedUserKey?: UnsignedSharedKey;
```

Eventually, we want AC to act as a transport layer for higher level cryptographic objects, that AC does not need to / want to know about. We want to provide Toolbox 1 (https://bitwarden.atlassian.net/wiki/spaces/EN/pages/3092250645/KM+Feature+Team+API+Vision#Toolbox-1%3A-Domain-Based-Cryptographic-Objects) high level objects `OrganizationMembership`, `AccountRecoveryEnrollemnt` here.

```suggestion
  organizationMembership: {
    encapsulatedOrganizationKey: UnsignedSharedKey
  }
  accountRecoveryEnrollment: {
    encapsulatedUserKey?: UnsignedSharedKey;
  }
```

^ This would be the eventual state. @eliykat do we already want to move towards this direction, or do we want to keep the current standard and later migrate all touching points in one go?

JimmyVo16: 1. We can update the types and add comments. That should be easy and is only scoped to this PR.

2. We can change the field names in this PR, but these are the common field names used across our server, including the database. We can make the change in this PR, which will require updating the endpoint, but that should be a small change. However, updating this across the entire server would be a much larger effort.

3. It seems like the long-term solution is to use an object instead of a string. Based on the doc, it looks like each object will have its own database table. This would require a major lift.

quexten: 2/3, okay, lets table these for later. I do think this is where we want to head, but I understand if this is not the time/place to do it.

eliykat: @quexten I agree that we need to move in this direction, but we should figure out new cross-team interfaces + terminology for membership before we start making changes. For the scope of this PR I agree with @JimmyVo16 that it's better to be consistent with existing practices. (e.g. here `orgUserKey` is the `Key` on the `OrganizationUser` table which is aligned with existing terminology)

libs/organization-invite-link/src/models/requests/organization-invite-link-confirm.request.ts
```diff
@@ -0,0 +1,18 @@
+export class OrganizationInviteLinkConfirmRequest {
```
r-tome, line 1: This should include `OrganizationId`

JimmyVo16:
Holding off on this ticket/PR until the changes from Milestone 1 are completed. I'll make sure to incorporate the code review feedback once I start this PR back up again.

JimmyVo16:
This is no longer needed since we're making the server call in the SDK.

PR #21798, opened 2026-07-10
Ac/pm 38906 decrypt list with failures for collections

JaredScar:
## 🎟️ Tracking

## 📔 Objective


libs/admin-console/src/common/collections/services/default-collection-admin.service.ts
```diff
@@ -146,8 +192,119 @@ export class DefaultCollectionAdminService implements CollectionAdminService {
...
+   * {@link FeatureFlag.CollectionAdminBulkDecrypt} until this path has proven out in production.
+   *
+   * Unlike the personal-vault decryption path, a collection that fails to decrypt is never
+   * silently dropped here: it's shown with a placeholder name so admins can still see, manage,
+   * and delete it from the Admin Console.
+   */
+  private async decryptManyBulk(
+    collections: CollectionAccessDetailsResponse[],
+    userId: UserId,
+  ): Promise<CollectionAdminView[]> {
```
BTreston: We should return an Observable instead

libs/admin-console/src/common/collections/services/default-collection-encryption.service.ts
```diff
@@ -25,11 +29,38 @@ export class DefaultCollectionEncryptionService implements CollectionEncryptionS
   }
 
   async decryptMany(collections: Collection[], userId: UserId): Promise<CollectionView[]> {
+    return (await this.decryptManyWithFailures(collections, userId))[0];
+  }
+
+  async decryptManyWithFailures(
+    collections: Collection[],
+    userId: UserId,
+  ): Promise<[CollectionView[], Collection[]]> {
```
BTreston: This return type is confusing to the consumers. It would be clearer to define a result object type instead.

libs/admin-console/src/common/collections/services/default-collection-encryption.service.ts
```diff
@@ -25,11 +29,38 @@ export class DefaultCollectionEncryptionService implements CollectionEncryptionS
...
+      ? this.decryptManyWithFailuresBulk(collections, userId)
+      : this.decryptManyWithFailuresOriginal(collections, userId);
+  }
+
+  /**
+   * Original implementation: decrypts each collection individually via the SDK, one at a time.
+   * A collection that fails to decrypt is logged and returned in the failures array instead of
+   * aborting the rest of the batch.
+   */
+  private async decryptManyWithFailuresOriginal(
```
BTreston: We should probably use our standard v2 naming convention

libs/admin-console/src/common/collections/services/default-collection-encryption.service.ts
BTreston, line 1: A lot of `Promise` return types here. We should try to stick with `Observables` where possible for these new operations I think.

libs/admin-console/src/common/collections/services/default-collection-encryption.service.ts
```diff
@@ -39,21 +70,122 @@ export class DefaultCollectionEncryptionService implements CollectionEncryptionS
...
+  /**
+   * Batched implementation using the SDK's `decrypt_list_with_failures`, which parallelizes
+   * decryption of the whole list for better performance on large lists. Gated behind
+   * {@link FeatureFlag.CollectionBulkDecryptWithFailures} until the SDK bindings that expose
+   * this method have rolled out everywhere this service is used.
+   */
+  private async decryptManyWithFailuresBulk(
+    collections: Collection[],
+    userId: UserId,
+  ): Promise<[CollectionView[], Collection[]]> {
```
BTreston: same feedback as the other function. Something like 
```typescript
type DecryptionResult  = {
    success: CollectionView[],
    failure: Collection[]
}
....
  private async decryptManyWithFailuresBulk(
    collections: Collection[],
    userId: UserId,
  ): Observable<DecryptionResult>
```
or similar

BTreston: This is also a pretty massive function that can probably be decomposed a bit.

libs/common/src/enums/feature-flag.enum.ts
```diff
@@ -13,6 +13,8 @@ export enum FeatureFlag {
   /* Admin Console Team */
   GenerateInviteLink = "pm-32497-generate-invite-link",
   PM35153CollectionSdkDecryption = "pm-35153-collection-sdk-decryption",
+  CollectionBulkDecryptWithFailures = "collection-bulk-decrypt-with-failures",
+  CollectionAdminBulkDecrypt = "collection-admin-bulk-decrypt",
```
BTreston, line 16: 2 feature flags in one PR?

libs/admin-console/src/common/collections/services/default-collection-admin.service.ts
BTreston, line 1: This applies to all the files: prefer `undefined` over `null`

BTreston (commented):
Just some very high level things that caught my eye, not an in depth review

libs/admin-console/src/common/collections/services/default-collection-encryption.service.ts
```diff
@@ -25,11 +29,38 @@ export class DefaultCollectionEncryptionService implements CollectionEncryptionS
...
     if (!collections || collections.length === 0) {
-      return [];
+      return [[], []];
     }
 
-    return firstValueFrom(
+    const bulkDecryptEnabled = await this.configService.getFeatureFlag(
+      FeatureFlag.CollectionBulkDecryptWithFailures,
+    );
```
eliykat: This is currently a nested feature flag: your first SDK feature flag switches at the `collectionService` level, then this switches within `collectionEncryptionService` beneath that. We want to avoid nested flags because it becomes difficult to reason about. It's OK for a draft, but we should wait until your first feature flag is removed before merging this.

libs/admin-console/src/common/collections/services/default-collection-encryption.service.ts
```diff
@@ -25,11 +29,38 @@ export class DefaultCollectionEncryptionService implements CollectionEncryptionS
...
+   * A collection that fails to decrypt is logged and returned in the failures array instead of
+   * aborting the rest of the batch.
+   */
+  private async decryptManyWithFailuresOriginal(
+    collections: Collection[],
+    userId: UserId,
+  ): Promise<[CollectionView[], Collection[]]> {
+    const startTime = performance.now();
+
+    const result = await firstValueFrom(
```
eliykat: To Brandon's comment - avoid using `firstValueFrom`, let the observables flow through the method calls.

libs/admin-console/src/common/collections/services/default-collection-encryption.service.ts
```diff
@@ -39,21 +70,122 @@ export class DefaultCollectionEncryptionService implements CollectionEncryptionS
...
+          for (const collection of collections) {
+            if (collectionMap.has(collection.id)) {
+              // Decrypted views are re-associated with their source `Collection` by id below
+              // (needed to preserve `defaultUserCollectionEmail`, which gates the security
+              // restriction in `CollectionView.canEditName()`). A duplicate id would make that
+              // re-association ambiguous and could silently attach the wrong collection's
+              // metadata to a decrypted view, so fail closed instead of guessing.
+              throw new Error(
+                `Duplicate collection id passed to decryptManyWithFailures: ${collection.id}`,
+              );
```
eliykat: No need to throw here, I think we can just silently drop duplicate `id`s (as that should indicate a duplicate collection). A `Map` doesn't support duplicate keys, so you get that for free if you just do a 1-liner: `new Map(collections.map(c => [c.id, c]));` (last duplicate wins).

libs/admin-console/src/common/collections/services/default-collection-encryption.service.ts
```diff
@@ -39,21 +70,122 @@ export class DefaultCollectionEncryptionService implements CollectionEncryptionS
...
+          }
+
+          const result: DecryptCollectionListResult = ref.value
+            .vault()
+            .collections()
+            .decrypt_list_with_failures(collections.map((c) => c.toSdkCollection()));
+
+          const views = result.successes
+            .map((sdkView) => {
+              const id = sdkView.id ? uuidAsString(sdkView.id) : "";
```
eliykat: Incorrect null handling - an empty string is not a valid ID.

libs/admin-console/src/common/collections/services/default-collection-encryption.service.ts
```diff
@@ -39,21 +70,122 @@ export class DefaultCollectionEncryptionService implements CollectionEncryptionS
...
+              const original = collectionMap.get(id);
+              if (!original) {
+                return null;
+              }
```
eliykat: This should never happen as far as I can tell.

libs/admin-console/src/common/collections/services/default-collection-encryption.service.ts
```diff
@@ -39,21 +70,122 @@ export class DefaultCollectionEncryptionService implements CollectionEncryptionS
...
+
+          const failedCollections = result.failures.map((sdkCollection) =>
+            Collection.fromSdkCollection(sdkCollection),
+          );
+
+          for (const failure of failedCollections) {
+            this.logService.error(`Failed to decrypt collection ${failure.id}`);
+          }
+
+          return [views, failedCollections] as [CollectionView[], Collection[]];
```
eliykat: Unnecessary cast (or if it is necessary then something else is wrong)

libs/admin-console/src/common/collections/services/default-collection-encryption.service.ts
```diff
@@ -39,21 +70,122 @@ export class DefaultCollectionEncryptionService implements CollectionEncryptionS
...
+            })
+            .filter((v): v is CollectionView => v !== null);
+
+          const failedCollections = result.failures.map((sdkCollection) =>
+            Collection.fromSdkCollection(sdkCollection),
+          );
+
+          for (const failure of failedCollections) {
+            this.logService.error(`Failed to decrypt collection ${failure.id}`);
+          }
```
eliykat: If lots of collections fail, this is going to spam the log.

I think a better approach is to let the caller choose how to handle failures. (We will probably just log it, but the caller can decide that - and probably concatenate all the ids into a single log message.)

libs/admin-console/src/common/collections/services/default-collection-encryption.service.ts
```diff
@@ -39,21 +70,122 @@ export class DefaultCollectionEncryptionService implements CollectionEncryptionS
...
+
+          using ref = sdk.take();
+
+          const collectionMap = new Map<string, Collection>();
+          for (const collection of collections) {
+            if (collectionMap.has(collection.id)) {
+              // Decrypted views are re-associated with their source `Collection` by id below
+              // (needed to preserve `defaultUserCollectionEmail`, which gates the security
+              // restriction in `CollectionView.canEditName()`). A duplicate id would make that
```
eliykat: No changes required for now - but we have a few patterns here I'm dissatisfied with:
- we have to keep this `defaultUserCollectionEmail` value for permission checks to work
- we have to map the SDK value back to `CollectionAdminView` to use the permission checks that live on that object
- those permission checks require you to pass in the correct `organization`, which is a bad interface
- putting the permission logic on the object also makes it difficult to feature flag (this has come up in the past)

I think (but I'm not sure) that I would prefer an SDK interface that accepts all your collections + all your organizations and then computes all these permissions for you on the return object. This could return an enriched collectionView, or a parallel CollectionPermissions object (with id + permissions).

Thoughts @BTreston ?

libs/admin-console/src/common/collections/services/default-collection-admin.service.ts
```diff
@@ -146,8 +192,119 @@ export class DefaultCollectionAdminService implements CollectionAdminService {
...
+        id: c.id as CollectionId,
+        name: new EncString(c.name),
+        organizationId: c.organizationId,
+      });
+      collection.externalId = c.externalId;
+      collection.readOnly = c.readOnly;
+      collection.hidePasswords = c.hidePasswords;
+      collection.manage = c.manage;
+      collection.type = c.type;
+      collection.defaultUserCollectionEmail = c.defaultUserCollectionEmail;
```
eliykat: This mapping should live inside one of the models.

libs/admin-console/src/common/collections/services/default-collection-admin.service.ts
eliykat: Please align the decryption paths - `CollectionEncryptionService` should be responsible for all collection decryption, and both `CollectionService` and `CollectionAdminService` should use it rather than implementing their own SDK calls and mapping logic.

Ideally there is 1 core method that decrypts `Collection` -> `CollectionView`, then there can be a wrapper around that that transforms it back to the `CollectionAdminView` (but that's another layer of mapping, not calling the SDK).

eliykat (changes requested):

dani-garcia (dismissed):
Platform changes LGTM

JimmyVo16:
Removing myself as a reviewer since we already have two AC members looking at this.

eliykat (changes requested):
I agree with your suggestion in Slack to break this into 2 PRs - one for each feature flag. That will make the changes easier to review and approve. I suggest focusing on the admin path first, because that will be quicker and easier to review, test and deploy.

JaredScar:
Closing in favor of the new PRs I'm adding

PR #21810, opened 2026-07-13
[PM-40335] PAM - Add access rules domain (1/2)

Hinton:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-40335

## 📔 Objective

Introduces the **foundation and logic layer** for the Privileged Access Management (PAM)
**access-rules** feature, extracted from the PAM POC (`pam/poc` + `pam/sdk-access-rules`)
into a standalone, production-shaped feature. All access-rule CRUD is served by the Rust
SDK (`client.commercial().pam().access_rules()`) rather than hand-rolled HTTP/DTOs.

> **This is part 1 of 2.** To keep review tractable, the ~4.9k-line feature is split into
> a foundation/logic PR (this one) and a UI/wiring PR. **The Angular components, routing,
> org-nav entry, and provider wiring land in a stacked follow-up PR** (based on this
> branch). Nothing in this PR is reachable at runtime yet — there are no components and no
> routes; it's dead code until the follow-up mounts it, and even then it's gated behind the
> `pm-37044-pam-v-0` flag (default **FALSE**) and the new `Organization.canManageAccessRules`
> permission.

### What's included (this PR)

- **Feature flag & permission** — `FeatureFlag.Pam` (`pm-37044-pam-v-0`, default **FALSE**)
  and `Organization.canManageAccessRules` (`isAdmin && usePam`, so the org must have PAM
  enabled server-side).
- **SDK dependency** — bumps `@bitwarden/sdk-internal` and
  `@bitwarden/commercial-sdk-internal` to `0.2.0-main.897` (which carry the `bitwarden-pam`
  crate types) and adds `@bufbuild/protobuf`.
- **Framework-agnostic contract layer** (`abstractions/`, `helpers/`, `date/relative-time`,
  re-exported via `index.ts`) — domain-type re-exports, the abstract `AccessRuleSdkService`
  (5 CRUD methods), the flat `AccessRuleError` helpers, and pure
  table/request/duration/collection-name helpers. No Angular APIs, so it stays
  unit-testable without a TestBed. This was originally its own `@bitwarden/bit-pam`
  package; it's now folded into `bit-web`, since `bit-web` was its only consumer.
- **SDK-backed service layer** (`services/`):
  - `access-rules-sdk.service.ts` — the SDK-backed `AccessRuleSdkService` implementation,
    following the canonical per-call SDK-consumption pattern (`SdkService.userClient$` →
    `using ref = sdk.take()`).
  - `access-rules.service.ts` — a page data service that owns the list + collections state
    and the CRUD round-trips (view concerns stay in the components, which arrive in part 2).

All layers ship with unit tests (67 tests across 8 suites).

### Deferred to the follow-up PR (part 2)

- Access-rules **list** (status filtering, per-row/bulk enable-disable-delete, badges, empty state)
- Routed **create/edit page** (`access-rules`, `access-rules/new`, `access-rules/:accessRuleId`)
  with templates and duration presets
- **IP-allowlist CIDR editor** (delegates to the SDK's commercial-only `is_valid_cidr`)
- **Org-nav entry** (`app-pam-org-nav-slot`) and localized strings
- Route module, `provide-pam.ts`, and `AppModule` wiring

### Notes

- SDK domain shapes (`AccessCondition`, `AccessRuleAddEditRequest`, `AccessRuleView`, …) are
  re-exported with `export type` (not `export`) — type-only and erased at compile time, so
  jest never resolves the wasm SDK when running this directory's unit tests. This layer has
  **no runtime SDK imports**; the commercial-only runtime surface (`is_valid_cidr`) is
  introduced in part 2.


Hinton:
Breaking this up into two PRs.

amorask-bitwarden (dismissed):

bitwarden_license/bit-web/src/app/pam/date/relative-time.ts
```diff
@@ -0,0 +1,32 @@
+/**
+ * Format an instant as a localized relative phrase, e.g. "5 min. ago" or
+ * "in 2 hr.", by walking from seconds up to years and emitting the first unit
+ * the delta fits within.
+ *
+ * The {@link Intl.RelativeTimeFormat} is passed in (not constructed here) so the
+ * caller controls locale and can cache the formatter across many rows. Both
+ * times are epoch milliseconds; a non-finite result returns the empty string.
+ */
+export function formatRelativeTime(
```
kspearrin, line 10: Is this not something we would want to use elsewhere, other than PAM?

Hinton: Probably, but at that time we can lift it out. (I raised the topic a few days ago but have not gotten any commitment to how extensive this pattern will be)

kspearrin (approved):

BTreston (approved):

quexten (approved):
Approving CODEOWNERS changes for @bitwarden/tech-leads

PR #21824, opened 2026-07-13
[PM-37991] Implement Automatic App Login Policy v2 and associated components

JaredScar:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-37991

## 📔 Objective

Updates to use policy drawers for policy drawers initiative and follows new figma mockups.

## 📸 Screenshots
image


bitwarden_license/bit-web/src/app/admin-console/policies/policy-edit-definitions/automatic-app-login.component.spec.ts
```diff
@@ -0,0 +1,102 @@
...
+  });
+
+  it("renders the v2 component inside the drawer, with its own drawer description", () => {
+    const policy = new AutomaticAppLoginPolicy();
+
+    expect(policy.v2?.component).toBe(AutomaticAppLoginPolicyV2Component);
+    expect(policy.v2?.component).not.toBe(policy.component);
+    expect(policy.v2?.description).toBe("automaticAppLoginWithSSODescV2");
+  });
+});
```
eliykat: These are low quality tests imo because they test basic rendering which is better covered by Storybook. I suggest they can be deleted.

bitwarden_license/bit-web/src/app/admin-console/policies/policy-edit-definitions/automatic-app-login.component.spec.ts
```diff
@@ -0,0 +1,102 @@
...
+    }).compileComponents();
+
+    fixture = TestBed.createComponent(componentClass);
+    component = fixture.componentInstance;
+  });
+
+  it("starts with the idpHost control empty", () => {
+    expect(component.data.controls.idpHost.value).toBeNull();
+  });
+
```
eliykat: Also covered by storybook.

bitwarden_license/bit-web/src/app/admin-console/policies/policy-edit-definitions/automatic-app-login-v2.component.html
```diff
@@ -0,0 +1,14 @@
+<bit-form-control>
+  <bit-switch [formControl]="enabled" id="enabled"></bit-switch>
+  <bit-label>{{ "enablePolicy" | i18n }}</bit-label>
+</bit-form-control>
+
+<div [formGroup]="data">
+  <div class="tw-grid tw-grid-cols-12 tw-gap-4">
+    <bit-form-field class="tw-col-span-12 !tw-mb-0">
```
eliykat: There's no need to use `tw-grid` just to have a single full width column and nothing else. In my testing it can be removed with no impact on display. `tw-gap` can also be removed because it doesn't do anything without a grid.

Why is `!tw-mb-0` needed? Our component library forms have standard whitespace padding and design requirements should align with this. We generally should not override CL defaults when using its components. Please remove if not needed or we can discuss with design if there is a conflict between the CL and designs. (If this pattern has been followed for all drawers then this can be non-blocking, as it's probably better to be consistent and we can follow up separately.)

eliykat (changes requested):

eliykat (approved):

PR #21829, opened 2026-07-13
[PM-40236] fix: return locked SDK client when key material is unavailable during logout race

ike-kottlowski:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-40236

## 📔 Objective

Fixes a race condition that could leave a user stuck in a locked state during logout.

When the SDK client's crypto is being initialized, key material can become unavailable mid-flight if a logout is processing concurrently. Previously, `initializeClientCrypto` would throw an `EncryptionSettingsError` with the `UserKeyStateRetrievalFailed` variant, and that error propagated out of the client factory as a hard failure.

This change wraps the `initializeClientCrypto` call in a try/catch and treats `UserKeyStateRetrievalFailed` exactly like the existing null-`userKey` case: return the locked client instead of throwing. Any other initialization failure is still a genuine error and continues to propagate.

Added unit coverage in `default-sdk.service.spec.ts` verifying that a `UserKeyStateRetrievalFailed` error yields a locked client while other errors propagate.

## 📸 Screenshots

https://github.com/user-attachments/assets/cc8c98cc-7aae-43ba-82f7-ba7e8ec2349c

A user logs into the browser extension in Chrome and then logs out.


mzieniukbw:
I advise against this. This will only hide legitimate issues. Just like we had decryption errors for many years but they have not surfaced anywhere (suppressed) hence we did not know about them.

ike-kottlowski:
addressed in #21869, closing.

PR #21836, opened 2026-07-14
[PM-40303] Fix logout when underprivileged user accesses Manage Send policy 

harr1424:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-40303

## 📔 Objective

The PR resolves a regression whereby an org user with custom permissions including "Manage Policies" but not including "Manage SSO" would be logged out when opening the "Manage Send" policy drawer. `fetchClaimedDomains()` calls an API endpoint requiring `canManageDomainVerification` privileges. `ApiService` will log a user out after more than a single `401` response, and so the early return added in this PR avoids the logout.


itsadrago (approved):

eliykat (approved):
Approving to unblock the release, but as discussed we may consider expanding who can read claimed domains so that we don't have differing behavior.

PR #21841, opened 2026-07-14
[PM-40335] PAM - Add access rules UI (2/2)

Hinton:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-40335

## 📔 Objective

**Part 2 of 2** of the PAM access-rules feature. This PR is **stacked on #21810** and
targets its branch (`pam/access-rules-sdk`), so the diff here shows only the Angular UI and
wiring built on top of the contract/service layer introduced there. Please review/merge
#21810 first; once it merges, this PR's base will be retargeted to `main`.

The whole feature stays gated behind the `pm-37044-pam-v-0` flag (`FeatureFlag.Pam`,
default **FALSE**) and the `Organization.canManageAccessRules` permission (`isAdmin && usePam`),
so nothing is reachable in production until both are on.

### What's included

- **Access-rules list** — table with status filtering, per-row and bulk
  enable/disable/delete, collection/condition badges, and an empty state.
- **Routed create/edit page** — mounted at `access-rules`, `access-rules/new`, and
  `access-rules/:accessRuleId`, with access-rule templates and duration presets.
- **IP-allowlist CIDR editor** — the `ip_allowlist` condition's editor, a thin view over a
  `FormArray` owned by the edit page's form group. CIDR validation delegates to the SDK's
  commercial-only `is_valid_cidr` (full parse, host-bits rejection, IPv6) instead of regexes.
- **Org-nav entry** — `apps/web/src/app/pam/org-nav-slot/`. A self-contained
  `app-pam-org-nav-slot` that renders the PAM nav group in the org admin-console side nav
  only when the flag is on and the org can manage access rules. The OSS layout plugs it in
  with a single tag and no PAM-specific symbols.
- **Wiring** — the lazy-loaded `PamRoutingModule` (guarded by `FeatureFlag.Pam`),
  `provide-pam.ts` binding the SDK-backed services + CIDR validator, and the `AppModule`
  `providePam()` registration. Plus Storybook stories and localized strings.


kspearrin (dismissed):

eliykat (dismissed):
AC changes look fine. Do we need to do something to hide it from the navigation if the user is running an OSS build?

Hinton:
> AC changes look fine. Do we need to do something to hide it from the navigation if the user is running an OSS build?

I don't think so. You shouldn't get the org capability in OSS builds.

Edit: You could technically have a license granting you and your users PAM _and_ be running an OSS build, but the same applies to SSO config.

patriksvensson (approved):
LGTM 👍

jrmccannon (approved):

PR #21848, opened 2026-07-14
[PM-40377] Fix bug with form field to have textarea behave the same as an input when disabled

jengstrom-bw:
## 🎟️ Tracking

[Jira](https://bitwarden.atlassian.net/browse/PM-40377)

## 📔 Objective

`textarea` fields were not displaying correctly when disabled, this updates it to behave the same as an `input` field

## 📸 Screenshots

Before:
Screenshot 2026-07-14 at 9 24 59 AM

After:
Screenshot 2026-07-14 at 9 24 16 AM

### DIFF
--- a/libs/components/src/form-field/field-container.directive.ts
+++ b/libs/components/src/form-field/field-container.directive.ts
@@ -26,7 +26,7 @@ export class BitFieldContainerDirective {
       "tw-border-border-strong",
       "tw-bg-bg-secondary",
       "tw-placeholder-fg-body-subtle",
-      "has-[input:disabled]:tw-border-border-base",
+      "has-[:is(input,textarea):disabled]:tw-border-border-base",
       "tw-transition-colors",
       "has-[:focus-visible]:tw-border-border-brand",
       "has-[.tw-test-focus-visible]:tw-border-border-brand",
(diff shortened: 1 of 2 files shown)


libs/components/src/form-field/field-container.directive.ts
```diff
@@ -26,7 +26,7 @@ export class BitFieldContainerDirective {
       "tw-border-border-strong",
       "tw-bg-bg-secondary",
       "tw-placeholder-fg-body-subtle",
-      "has-[input:disabled]:tw-border-border-base",
+      "has-[:is(input,textarea):disabled]:tw-border-border-base",
```
BryanCunningham, line 29: I don't think we need the extra `:is` selector here. The field container will never actually *be* an input or textarea IIRC

willmartian: @BryanCunningham  The `:is` here is basically saying input OR textarea, right? So the logic is:

**has**  (input OR textarea)

BryanCunningham: Yeah, I guess it's a shorthand way of writing `has-[textarea:disabled,input:disabled]:tw-border-border-base`. We don't have to repeat the `:disabled` part. We probably could just say `has-[:disabled]...` here. And just cover all the use cases

BryanCunningham: Now that I've looked at it more closely, either way is fine with me

BryanCunningham (approved):
Thanks for catching this @jengstrom-bw

libs/components/src/form-field/field-container.directive.ts
willmartian: Ty for taking a look 🙇

PR #21849, opened 2026-07-14
[PM-40241] Bugfix - Rehydrate localData lastUsedDate on local cipher views

### LINKED ISSUE #21794: Bitwarden inline autofill suggestions differ from extension popup on Brave (Chrome extension)
### Steps To Reproduce

When using Bitwarden as my primary password manager with the Chrome extension in the Brave browser on macOS, the inline autofill suggestions shown when focusing on username/password fields do not match the items shown in the Bitwarden extension popup for the same page.

Specifically, the “highest ranking” login entries (correct URL match, frequently used, clearly relevant) appear in the extension popup, but they either do not appear or are ranked much lower in the inline dropdown that appears on the field itself. This makes inline autofill unreliable and forces me to open the extension popup every time to select the correct credential, which is significantly less convenient than the expected behavior.

I have already:

- Updated Brave to the latest stable version.
- Updated the Bitwarden extension to the latest version.
- Verified that Brave’s built‑in password manager/autofill is disabled.
- Tried toggling Bitwarden’s “Make Bitwarden your default password manager” and reinstalling the extension.
- Restarted Brave and the machine after each change.
- The issue persists across multiple sites and is reproducible consistently.

**Environment setup**
- Platform: macOS.
- Browser: Brave (1.92.139).
- Bitwarden: (Version: 2026.6.1 / SDK: 'main (c5d5bba)' / Server version: 2026.6.2) Chrome extension installed from the Chrome Web Store, logged in and vault unlocked.
- Browser’s built‑in password manager disabled (no saved passwords in Brave).

**Steps to reproduce**
- Create or verify test login
- In Bitwarden, create or confirm a login entry for a test site (e.g., https://example.com or any common login site you use).
- Ensure the URI/URL for the login exactly matches the site you will test on (e.g., https://example.com/login).
- Confirm this login is the one you expect to be the “top match” (frequently used, correct URL, no competing entries).
- Open the test site in Brave
- Navigate to the login page (e.g., https://example.com/login) in a regular Brave window.
- Make sure the Bitwarden extension is unlocked.
- Observe inline autofill behavior
- Click/focus the username field.
- Observe the inline Bitwarden autofill dropdown (the overlay on or near the field).
- Note which logins are suggested and in what order.
- Compare with extension popup
- Without changing the page, click the Bitwarden extension icon in the toolbar to open the popup.
- Check the list of suggested logins for the current page in the popup.
- Confirm that the expected “top” login appears here and that the ordering is different from the inline suggestions.
- Confirm mismatch
- If the expected login appears at or near the top in the extension popup, but does not appear or appears much lower in the inline dropdown, note this mismatch.
- Repeat steps 3–5 on a second and third site with multiple logins stored (e.g., multiple accounts for the same domain) to confirm this is consistent and not site‑specific.

### Expected Result

- The inline autofill dropdown on focused fields should show the same matching logins, with similar ranking, as the Bitwarden extension popup for the current page.
- The most relevant login (correct URL, frequently used) should appear at the top of both the inline dropdown and the extension popup, allowing quick autofill without opening the popup.

### Actual Result

- Inline autofill suggestions are incomplete or differently ranked compared to the Bitwarden extension popup.
- The most relevant login appears correctly in the extension popup but is missing or de‑prioritized in the inline dropdown, forcing manual selection from the popup each time.

### Screenshots or Videos

_No response_

### Additional Context

_No response_

### Operating System

macOS

### Operating System Version

_No response_

### Web Browser

Brave

### Browser Version

_No response_

### Environment Versions

Version: 2026.6.1

SDK: 'main (c5d5bba)'

Server version: 2026.6.2

### Issue Tracking Info

jprusik:
## 🎟️ Tracking

[PM-40241](https://bitwarden.atlassian.net/browse/PM-40241), addresses https://github.com/bitwarden/clients/issues/21794, relates to https://github.com/bitwarden/clients/pull/16810

## 📔 Objective

CipherViews from the SDK are missing `localData` and do not hydrate the local models consumed in downstream experiences. This PR adds a re-hydration util and invokes it at the `getAllDecryptedUsingSdk` callsite, addressing ordering in the inline menu and context menu.

The autofill suggestion in the extension view has it's own logic resolving last used order where the typing differs from the `localData` shape.

Not addressed:
- [A similar sorting case in Desktop client](https://github.com/bitwarden/clients/blob/main/apps/desktop/src/services/encrypted-message-handler.service.ts#L121-L122)

## 📸 Screenshots

**with localData**

Screenshot 2026-07-14 at 12 39 47 PM

**without localData**

Screenshot 2026-07-14 at 12 39 35 PM

---

**broken cipher order**

https://github.com/user-attachments/assets/5072f464-54e1-4e34-8d91-6456aec6c0d0

**fixed cipher order**

https://github.com/user-attachments/assets/b88d9fab-d42b-4903-a358-008729102d2a

[PM-40241]: https://bitwarden.atlassian.net/browse/PM-40241?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/common/src/vault/utils/hydrate-ciphers-with-local-data.ts
```diff
@@ -0,0 +1,32 @@
...
+/**
+ * Re-attaches client-only {@link LocalData} to decrypted {@link CipherView}s and returns
+ * them (`localData` is a client-side-only field. The SDK decryption path does not
+ * populate it on the views it returns).
+ *
+ * @param ciphers - The decrypted cipher views to hydrate. Mutated in place.
+ * @param localData - The per-cipher local data map, e.g. from `CipherService.localData$`.
+ * @returns The hydrated `ciphers` array
+ */
+export function hydrateCiphersWithLocalData(
```
jprusik, line 14: I intentionally didn't handle this in `getAllDecryptedForUrl` since nothing about that method signals intention or expectation around sorting strategy.

dan-livefront (dismissed):
Nit: Might be cleaner to hydrate once in `getAllDecryptedUsingSdk` before caching so every consumer gets `localData` for free and we don’t need the multiple call site patches.

That isn't blocking though, just a callout for potential future enhancement.

libs/common/src/vault/utils/hydrate-ciphers-with-local-data.spec.ts
nick-livefront, line 1: Thank you for the unit tests!

libs/common/src/vault/utils/hydrate-ciphers-with-local-data.ts
nick-livefront: This makes sense to me, keep everything single purpose!

nick-livefront (dismissed):
I agree with @dan-livefront's suggestion and that it is non-blocking. We haven't had any other issues regarding this that I know of.

jprusik:
> I agree with @dan-livefront's suggestion and that it is non-blocking. We haven't had any other issues regarding this that I know of.

I agree as well; making adjustments now and will push up shortly

libs/common/src/vault/utils/hydrate-ciphers-with-local-data.ts
jprusik: I wrote this when this method was also handling sorting as well. But, since callsites requiring the hydrated data don't _necessarily_ have the same sorting concerns, I've decoupled sorting from hydration (mutation in place allows sort chaining)

libs/common/src/vault/services/cipher.service.ts
```diff
@@ -2067,6 +2067,14 @@ export class CipherService implements CipherServiceAbstraction {
         return null;
       }
 
+      // `getAllDecryptedForUrl` already hydrates localData at the source, but this path caches the
+      // result in `sortedCiphersCache` and is read immediately after a lastLaunched/lastUsed update
+      // (the launch-then-autofill flow, PM-26302) before `cipherViews$` necessarily re-decrypts.
+      // Re-hydrate from the current localData$ so the correct cipher is selected regardless of that
+      // emission timing.
+      const localData = await firstValueFrom(this.localData$(userId));
+      ciphers = hydrateCiphersWithLocalData(ciphers, localData);
```
jprusik, line 2082: I did test the case this originally covered (from https://github.com/bitwarden/clients/pull/16810) and the general hydration worked fine, but given the high-risk of a bad cipher/site handoff, I thought the extra defensiveness here was warranted.

libs/common/src/vault/services/cipher.service.ts
```diff
@@ -352,10 +353,24 @@ export class CipherService implements CipherServiceAbstraction {
...
+    // decrypted views. Re-attach it here so every consumer of `cipherViews$`/`getAllDecrypted`
+    // sees it, matching the legacy and list-view decryption paths.
+    let localData: Record<CipherId, LocalData> | undefined;
+
+    // Wrap `localData` in try-catch so a failure to hydrate data doesn't cascade to a larger failed experience
+    try {
+      localData = await firstValueFrom(this.localData$(userId));
+    } catch {
+      localData = undefined;
+    }
```
nick-livefront, line 372: 👏

nick-livefront (dismissed):

dan-livefront (dismissed):

nick-livefront (approved):

dan-livefront (approved):

PR #21857, opened 2026-07-14
[PM-39772] Detect shadow hosts created from iframes

bensbits91:
## 🎟️ Tracking

[PM-39772](https://bitwarden.atlassian.net/browse/PM-39772)

### Problem

When a shadow host is created inside an iframe and then moved to the top document, the autofill scanner misses it on the first pass. Users see this as one of three things when they click a form field inside the moved host: no inline menu at all, a delayed menu, or a menu that arrives so quickly the bug looks like a non-issue.

Root cause: two spots on the shadow-host detection pipeline used `instanceof Element` to filter candidates.

- `isShadowRootCandidate` (mutation filter) — `collect-autofill-content.service.ts`
- `scanForNewShadowRootInSubtree` (batch host check) — `dom-query.service.ts`

`instanceof` compares against the **current realm's** constructor. An element created inside an iframe keeps the iframe realm's `Element` constructor even after being moved to the top document, so top-frame `instanceof Element` returns `false` and the host is filtered out. Detection then falls to the next full-document rescan, which is realm-agnostic, so it eventually works, but only on the rescan's schedule (often after the user has already clicked into the field).

### Fix

Replace `instanceof Element` in both spots with the existing realm-agnostic helper `nodeIsElement(node)` (which checks `node.nodeType === Node.ELEMENT_NODE`). `Node.ELEMENT_NODE` is a spec constant (`1`) identical in every realm.

The same swap was already applied to attribute-mutation handling — see the sister comment at `collect-autofill-content.service.ts:1453`:

> `nodeType === 1` instead of `instanceof Element` — works across realms (adopted-from-iframe).

This PR extends that pattern to the two remaining sites in the shadow-host detection path.

### Why this is hard to reproduce manually — and how the tests help

The ticket's own repro steps state: *"...no inline menu, a delayed resolution, **or a seemingly quick resolution**."* All three outcomes are the same bug, because the full-document rescan is the safety net. On fast hardware the rescan window is short enough that the menu appears essentially in time with the click. The bug looks absent when it's actually just been raced.

Two Jest tests were added that make the failure deterministic by constructing a genuinely foreign-realm host in jsdom:

```ts
const iframe = document.createElement("iframe");
document.body.appendChild(iframe);
const host = iframe.contentDocument!.createElement("foreign-host");
host.attachShadow({ mode: "open" });

expect(host instanceof Element).toBe(false); // precondition guard
```

The precondition (`instanceof Element` is `false`) proves the setup actually reproduces the cross-realm condition — if jsdom ever changed and stopped giving us a separate realm, the test would fail loudly rather than silently pass.

Both tests failed on `main` before the changes introduce by this PR, and pass after. These tests provide a durable regression guard for a class of bug the manual repro can't be trusted to catch.

[PM-39772]: https://bitwarden.atlassian.net/browse/PM-39772?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/browser/src/autofill/services/collect-autofill-content.service.ts
```diff
@@ -1730,9 +1730,9 @@ export class CollectAutofillContentService implements CollectAutofillContentServ
   }
 
   private isShadowRootCandidate(node: Node): node is Element {
-    // FIXME (PM-39772): same-realm only — iframe-adopted (foreign-realm) hosts fall through here
-    // and are detected only later, not on insert.
-    if (!(node instanceof Element)) {
+    // `nodeType === 1` instead of `instanceof Element` — works across realms
+    // (hosts adopted from an iframe keep the iframe's `Element` constructor).
```
audreyality: ⛏️ 
```suggestion
```

apps/browser/src/autofill/services/dom-query.service.ts
```diff
@@ -400,7 +400,9 @@ export class DomQueryService implements DomQueryServiceInterface {
       return false;
     }
     // Host check — `querySelectorAll("*")` excludes the scope element.
-    if (subtree instanceof Element) {
+    // `nodeType === 1` instead of `instanceof Element` — works across realms
+    // (hosts adopted from an iframe keep the iframe's `Element` constructor).
```
audreyality: ⛏️ 

```suggestion
```

audreyality (dismissed):
👍🏻 Just some nits!

apps/browser/src/autofill/services/collect-autofill-content.service.ts
bensbits91: I agree that this comment is unnecessary and potentially fragile. Removed in https://github.com/bitwarden/clients/pull/21857/commits/a87ef98d426dfa81b18f97b7495f97de381174c7

apps/browser/src/autofill/services/dom-query.service.ts
bensbits91: I agree that this comment is unnecessary and potentially fragile. Removed in https://github.com/bitwarden/clients/pull/21857/commits/a87ef98d426dfa81b18f97b7495f97de381174c7

audreyality (approved):

PR #21858, opened 2026-07-14
[PM-40236] Logging out throws error mid-logout due to a race condition. User left in locked state

mzieniukbw:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-40236

## 📔 Objective

Race condition between clearing pin and clearing user key causes pin logout function to throw error. The logout procedure never finishes successfully, which leaves user in locked state.

Root cause: pin logout function uses default sdk client (`userClient$`), which initialises crypto using client managed state, and can fail with error, when user key is not present. Initialisation of user client is conditioned by user key presence, but since clearing of user key and pin key happens in parallel, asynchronously, sometimes the user init crypto is called with user key being present, only to be no longer present moments later within SDK (wasm).

Alternatives:
- Change pin service logout function to use register sdk client - which does not init user crypto. This is fine, since sdk's [unset_pin](https://github.com/bitwarden/sdk-internal/blob/fe351b43b7f6ace9cd9b6be4964bc75265b7782a/crates/bitwarden-core/src/key_management/pin_lock_system.rs#L312) does not use or expect any user crypto state.
- Leave the pin state clearing to `clearOn` event: https://github.com/bitwarden/clients/blob/9e73ef9b4816e50f5aa0519848e035ad1eb39e9c/libs/common/src/key-management/pin/pin.state.ts
- Suppress errors thrown due to user crypto init: #21829. This would cause pin state to be left in memory and on disk, which is not the right approach.

## 📸 Screenshots

https://github.com/user-attachments/assets/496c2a3c-fe3d-4c79-a49e-2739e942f07b


quexten (approved):
I think the long term solution to this is to get rid of the SDK emissions. Clients needs to behave the same as with mobile. Considering the bugs are piling up, we may need to consider undertaking this effort earlier to eliminate this bug class.

Nice find.

PR #21859, opened 2026-07-14
[PM-16487] Remove Add/edit v2 component

jengstrom-bw:
## 🎟️ Tracking

[Jira](https://bitwarden.atlassian.net/browse/PM-16487)

## 📔 Objective

`add-edit-v2` needs to be removed. It isn't used anywhere.


shane-melton (approved):
🧹 🚀

PR #21860, opened 2026-07-14
[PM-40286] - Add Vfo1TerminologyService and VFO1I18nPipe infrastructure

nick-livefront:
## 🎟️ Tracking

[PM-40286](https://bitwarden.atlassian.net/browse/PM-40286)

## 📔 Objective

Introduces the foundational infrastructure for the VFO1 Phase 1 terminology shift

- **`FeatureFlag.VFO1Foundation`** — new feature flag (`"vfo1-foundation"`)
- `Vfo1TerminologyService` and `Vfo1I18nPipe`: resolves between a legacy i18n key and a new key based on the flag; fails safe to legacy copy before the flag resolves
- Barrel exports added to `libs/vault/src/index.ts`

[PM-40286]: https://bitwarden.atlassian.net/browse/PM-40286?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/vault/src/services/vault-terminology.service.ts
```diff
@@ -0,0 +1,14 @@
+import { inject, Injectable, Signal } from "@angular/core";
+import { toSignal } from "@angular/core/rxjs-interop";
+
+import { FeatureFlag } from "@bitwarden/common/enums/feature-flag.enum";
+import { ConfigService } from "@bitwarden/common/platform/abstractions/config/config.service";
+
+@Injectable({ providedIn: "root" })
+export class VaultTerminologyService {
```
shane-melton: ❓ Should we also rename this to `Vfo1TerminologyService` since we renamed the pipe?

shane-melton (dismissed):
One naming question, otherwise I think we're good!

Also, I'm good with Vault owning this since we're one of the primary teams impacted/going to use it. But we could also move to a more shared space if others want ownership.

libs/vault/src/services/vault-terminology.service.ts
nick-livefront: Great call. I'll leave it owned by us for now but we can adjust: [dac9942](https://github.com/bitwarden/clients/pull/21860/commits/dac9942f49af0e649bc6282261101124199b3ec3)

shane-melton (approved):
Looks good! Though I'm not sure why these changes would be causing Chromatic to be failing 👀 😕

nick-livefront:
> Looks good! Though I'm not sure why these changes would be causing Chromatic to be failing 👀 😕

Must have been temporary, merge from main fixed it

PR #21876, opened 2026-07-15
[PM-40193] feat: add vfo1Icon pipe for flag-gated icon swaps

jaasen-livefront:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-40193

## 📔 Objective

Part of the VFO1 terminology work (epic PM-39699) that renames "collections" to "shared folders" and "folders" to "my folders" behind the `vfo1-foundation` flag.

Text terms are already handled by `Vfo1I18nPipe`, which selects a legacy vs. new i18n key off `Vfo1TerminologyService.enabled()`. The one thing that pipe can't express is the collection icon swap — a CSS class, not a translatable string. Rather than add a per-icon `collectionIconClass` signal, this PR mirrors the text pattern with a generic icon pipe:

- **`Vfo1IconPipe` (`vfo1Icon`)** — an impure pipe (flag state can change at runtime) that takes a legacy icon class and returns its VFO1 replacement when the flag is on, caching its last result to avoid recomputing on unchanged inputs.
- **`Vfo1TerminologyService.iconClass(iconClass)`** — resolves the mapping, backed by a frozen `VFO1_ICON_MAP` (currently `bwi-collection-shared` → `bwi-shared-folder`). Icon classes with no mapping pass through unchanged.

Behavior for the collection icon:

- flag off → `bwi-collection-shared`
- flag on  → `bwi-shared-folder`

The map lives on the terminology service because it's the non-text half of the same flag-gated terminology swap, and it reads the same synchronous `enabled` signal the `vfo1I18n` pipe uses. Using a generic pipe + map (instead of one named signal per icon) lets future icon swaps be added by extending the map alone — no new service members or exports. `Vfo1IconPipe` is exported from `libs/vault/src/index.ts` alongside `Vfo1TerminologyService` and `Vfo1I18nPipe`. Unit tests cover both flag states and pass-through for unmapped classes.

Note: the `bwi-shared-folder` glyph is added under separate icon-update work in PM-39699; this PR only emits the class name. No UI change yet.

### DIFF
--- a/libs/vault/src/index.ts
+++ b/libs/vault/src/index.ts
@@ -104,3 +104,4 @@ export { VaultBatchActionComponent } from "./components/vault-batch-bar/vault-ba
 
 export { Vfo1TerminologyService } from "./services/vfo1-terminology.service";
 export { Vfo1I18nPipe } from "./pipes/vfo1-i18n.pipe";
+export { Vfo1IconPipe } from "./pipes/vfo1-icon.pipe";
--- a/libs/vault/src/pipes/vfo1-icon.pipe.spec.ts
+++ b/libs/vault/src/pipes/vfo1-icon.pipe.spec.ts
@@ -1,0 +1,60 @@
+import { TestBed } from "@angular/core/testing";
+
+import { Vfo1TerminologyService } from "../services/vfo1-terminology.service";
+
+import { Vfo1IconPipe } from "./vfo1-icon.pipe";
+
+describe("Vfo1IconPipe", () => {
+  let iconClassFn: jest.Mock<string, [string]>;
+  let enabledFn: jest.Mock<boolean, []>;
+  let pipe: Vfo1IconPipe;
+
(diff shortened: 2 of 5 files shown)


shane-melton (commented):
Should we also create a reusable `pipe` for this like the `Vfo1I18nPipe`? Otherwise, I see use needing to import/inject the `Vfo1TerminologyService` everywhere these are used in templates.

I would imagine the usage is something like:

```html
 <bit-icon
  [name]="'bwi-collection-shared' | vfo1Icon"
  ></bit-icon>
```

Where `vfo1Icon` has a map that checks the input icon name and updates it to the corresponding VFO1 icon when the flag is enabled, otherwise leaves it alone.

jaasen-livefront:
@shane-melton Ready for another look when you have a moment. ;)

shane-melton (approved):

PR #21881, opened 2026-07-15
Fix SSH Agent concurrent requests

### LINKED ISSUE #21839: ssh-agent throttling
### Steps To Reproduce

I'm using Bitwarden Desktop on Linux, installed via flatpak.
`Bitwarden	com.bitwarden.desktop	2026.6.1	stable	system`

It feels like there's throttling in the ssh-agent when it's enabled. To witness the problem
```sh
# make sure SSH_AUTH_SOCK is set and some keys are there beforehand
# export one key
ssh-add -L | head -n 1 > test-key.pub
# hammer the socket
time (for i in {1..5}; do ssh-add -T test-key.pub >/dev/null 2>&1 & done; wait)
```

This hangs for 60 seconds. While using the default agent would be very fast. The use case that made me spot this: I scripted tmux to open several SSH connections to different servers, and only one worked consistently while others hanged and then failed.

This could be a red-herring but maybe worth checking out:

> Root cause: The renderer approval pipeline in apps/desktop/src/autofill/services/ssh-agent.service.ts (shared V1/V2) processes all sign requests through one subscription using switchMap (lines 117 and 158). switchMap cancels the previous in-flight request when a new one arrives, so concurrent requests drop all-but-one. Dropped requests never call signRequestResponse, so the native confirm (core/src/ssh_agent/mod.rs:139) blocks until SIGN_TIMEOUT = 60_000 (autofill/main/main-ssh-agent.service.ts:21) returns false. The code comment at lines 111-116 shows the cancelling behavior was intended as serialization.

### Expected Result

I expected the socket to handle ssh-agent -T instantly instead of hanging for a minute.

### Actual Result

```sh
time (for i in {1..5}; do ssh-add -T test-key.pub >/dev/null 2>&1 & done; wait)
# It took a minute, a likely timeout
real	1m0,029s
user	0m0,124s
sys	0m0,011s
```

### Screenshots or Videos

_No response_

### Additional Context

_No response_

### Operating System

Linux

### Operating System Version

Tumbleweed

### Installation method

Flatpak

### Build Version

Version 2026.6.1 SDK 'main (c5d5bba)' Shell 39.8.5 Renderer 142.0.7444.265 Node 22.22.1 Architecture x64

### Issue Tracking Info

neuronull:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-40340

## 📔 Objective

Fixes: #21839 

There is an edge case bug with concurrent requests that results in any request that comes in before the current request is resolved, being dropped.

The root cause is the use of the `switchMap` where `concatMap` should have been used.

## 📸 Testing done

// the new unit test fails on the non-fix

```
  SshAgentService – concurrent sign requests
    ✕ when two sign requests arrive before getAllDecrypted resolves, both receive signRequestResponse (8 ms)
  SshAgentService – concurrent list keys requests
    ✕ when two list requests arrive before getAllDecrypted resolves, both receive listRequestResponse (5 ms)

  ● SshAgentService – concurrent sign requests › when two sign requests arrive before getAllDecrypted resolves, both receive signRequestResponse
  ```

// the new unit test passes with the fix

```
  SshAgentService – concurrent sign requests
    ✓ when two sign requests arrive before getAllDecrypted resolves, both receive signRequestResponse (12 ms)
  SshAgentService – concurrent list keys requests
    ✓ when two list requests arrive before getAllDecrypted resolves, both receive listRequestResponse (5 ms)

Test Suites: 1 passed, 1 total
```

### CODE REVIEW

**Review by coltonhurst — APPROVED** [review]

File: apps/desktop/src/autofill/services/ssh-agent.service.ts [hunk 1]
```diff
@@ -154,8 +154,7 @@ export class SshAgentService implements OnDestroy {
 
           return of([message, account.id]);
         }),
-        // This switchMap handles fetching the ciphers from the vault.
-        switchMap(([message, userId]: [Record<string, unknown>, UserId]) =>
+        concatMap(([message, userId]: [Record<string, unknown>, UserId]) =>
```
quexten, line 157: IMO, the observable chain is still pretty hard to reason about. Would be nice to simplify at some point.

quexten (approved):

apps/desktop/src/autofill/services/ssh-agent.service.ts
neuronull: 100%. This particular block is nasty because it has v1 and v2 logic combined. Per the comment on 116, it should clean up quite a bit just by removing the v1. Will for sure look for areas to improve in that task as well.

PR #21887, opened 2026-07-15
[PM-18638] Prevent collection/cipher creation for suspended organizations

JaredScar:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-18638

## 📔 Objective

Resource not found error that now shows up in a toast is ugly and not user friendly... This changes the behavior to no longer be an error message toast at all. We disable the new cipher/collection buttons if a organization is disabled, since those should not be actioned on anyway.

## 📸 Screenshots

image
image


jengstrom-bw (dismissed):
Looks good, thanks Jared!

JimmyVo16 (approved):

jengstrom-bw (approved):

PR #21889, opened 2026-07-16
[PM-38557] Netwrix Password Secure importer should map URL-like fields to Bitwarden website URI

harr1424:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-38557

## 📔 Objective

When importing from Netwrix Password Secure, some fields expected to map to URIs were silently dropped. This PR adds support for multi-lingual representation of URI containing fields.


libs/importer/src/importers/netwrix/netwrix-passwordsecure-csv-importer.spec.ts
```diff
@@ -66,6 +93,55 @@ describe("Netwrix Password Secure CSV Importer", () => {
...
+
+  it("should not crash on an enclosed-in-quotes export whose only row is empty", async () => {
+    // Regression: the re-parse used to collapse to just the header and return null, throwing a
+    // TypeError in parse(). It must now resolve gracefully instead.
+    const result = await importer.parse(credentialsDataEnclosedInQuotesEmptyRow);
+    expect(result).not.toBeNull();
+    expect(result.success).toBe(true);
+  });
+
+  it("should map the English 'Organisational unit' header to a folder", async () => {
```
mcamirault, line 151: ❓ By setting `organizationId` on the next line this test is relying on the `moveFoldersToCollections` call on line 85 of the importer and checking for folders in a more roundabout fashion than I'd expect. Is there a reason we can't omit `organizationId` and check the folders and folderRelationships directly?

harr1424: I agree, I will push changes to correct this.

mcamirault (approved):

PR #21895, opened 2026-07-16
feat(shared-unlock): support trusted devices

quexten:
https://bitwarden.atlassian.net/browse/PM-37678

Add support for TDE devices during shared unlock. The intent is that:
1. Only trusted devices share unlock. Untrusted devices for TDE accounts *DO NOT* share unlock
2. If you trust your device, all devices unlocked by it will be trusted.
   - I.e "Device" here really means device, not "client" (what we call device internally).
3. For this, we must handle the unlock event emitted by shared unlock, trust the device and navigate away on the login decryption options component
4. If we extend to shared login, the idea is the same, you trust your device, and all clients on the device will be logged in and trusted. Shared login is not implemented here, but we are planning for how we will support it in the future.

### DIFF
--- a/libs/auth/src/angular/login-decryption-options/login-decryption-options.component.ts
+++ b/libs/auth/src/angular/login-decryption-options/login-decryption-options.component.ts
@@ -9,17 +9,18 @@ import {
   catchError,
   concatMap,
   defer,
+  filter,
   firstValueFrom,
   from,
   map,
   of,
   switchMap,
+  take,
   throwError,
 } from "rxjs";
 
 import { JslibModule } from "@bitwarden/angular/jslib.module";
 import {
-  LoginEmailServiceAbstraction,
   LogoutService,
   UserDecryptionOptions,
   UserDecryptionOptionsServiceAbstraction,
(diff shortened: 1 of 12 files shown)


libs/auth/src/angular/login-decryption-options/login-decryption-options.component.ts
```diff
@@ -123,7 +126,6 @@ export class LoginDecryptionOptionsComponent implements OnInit {
     private i18nService: I18nService,
     private keyService: KeyService,
     private loginDecryptionOptionsService: LoginDecryptionOptionsService,
-    private loginEmailService: LoginEmailServiceAbstraction,
```
quexten, line 126: Unrelated to the PR but this was unused

libs/auth/src/angular/login-decryption-options/login-decryption-options.component.ts
```diff
@@ -135,10 +137,11 @@ export class LoginDecryptionOptionsComponent implements OnInit {
     private validationService: ValidationService,
     private logoutService: LogoutService,
     private registerSdkService: RegisterSdkService,
-    private securityStateService: SecurityStateService,
```
quexten, line 138: Unrelated to the PR but this was unused

eligrubb (approved):

enmande (approved):

PR #21910, opened 2026-07-16
Autotype AppData path normalization

neuronull:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-40101

## 📔 Objective

Laying some foundation for the AppData mechanism that will be used in the pairing and verification of apps for autotype, this PR introduces the `path` component of the AppData, where paths can be normalized and de-normalized, using the cross-platform (existing) dep crate `dirs`.

The normalization is presently focused on user-specific paths to apps. That is, paths in the user's space, not the OS system paths.

This solution takes an approach that allows the path normalizer to be compiled and tested for non-native platforms to the host.

Normalization and denormalization are best-effort. A failure for the `dirs` crate to resolve the well-known paths would be very rare and unexpected.

## 📸 Testing

unit tests added


apps/desktop/desktop_native/autotype/src/app_data/path/platform_policy.rs
```diff
@@ -0,0 +1,57 @@
...
+
+/// Separator and case conventions for a target platform's paths.
+#[derive(Debug, Clone, Copy)]
+pub struct PlatformPolicy {
+    /// Canonical separator; all separators are rewritten to this.
+    pub canonical_sep: char,
+    /// Characters treated as path separators.
+    pub separators: &'static [char],
+    /// Whether prefix matching ignores ASCII case or not.
+    pub case_insensitive: bool,
```
coltonhurst: Nit: I feel like `case_sensitive` (going with affirmative boolean naming) would be easier to immediately reason about, compared to `case_insensitive`.

coltonhurst (commented):
I like this implementation, nice! Only one small nit. If you disagree with the nit that's fine, let me know and I'll approve.

coltonhurst (approved):

PR #21920, opened 2026-07-17
feat(automation-driver): add automation biometrics service

quexten:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-39560

## 📔 Objective

Adds an implemention of the automation biometrics service. The intent is to be used by the automation driver to expose the biometric subsystem at runtime to e.g. an agent or test automation. The biometrics requests can be read and answered by injecting commands onto the devtools, which end-to-end tests, as well as agents can use at runtime to simulate biometrics interaction.

## 📸 Screenshots

---

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>

### DIFF
--- a/apps/desktop/src/key-management/biometrics/automation-biometrics.service.ts
+++ b/apps/desktop/src/key-management/biometrics/automation-biometrics.service.ts
@@ -1,0 +1,135 @@
+import { LogService } from "@bitwarden/common/platform/abstractions/log.service";
+import { SymmetricCryptoKey } from "@bitwarden/common/platform/models/domain/symmetric-crypto-key";
+import { UserId } from "@bitwarden/common/types/guid";
+import { BiometricsStatus } from "@bitwarden/key-management";
+
+import { OsBiometricService } from "./os-biometrics.service";
+
+export type AutomationBiometricRequestType = "authenticate" | "unlock";
+
+export interface AutomationBiometricRequest {
+  id: string;
+  type: AutomationBiometricRequestType;
+  userId?: UserId;
+}
+
+interface PendingRequest extends AutomationBiometricRequest {
+  resolve: (approved: boolean) => void;
+}
+
+/**
+ * A fake {@link OsBiometricService} used for automation (E2E tests, manual automation). It replaces
+ * the real OS biometric service on the desktop main process when running in dev mode with the
+ * `USE_AUTOMATION_BIOMETRICS` environment variable set, so the native OS prompt never fires.
+ *
+ * Biometric requests are queued and block until automation approves or denies them, allowing tests
+ * to deterministically simulate the user accepting or rejecting the native prompt. The reported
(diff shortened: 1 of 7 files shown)


apps/desktop/src/key-management/biometrics/automation-biometrics.service.ts
```diff
@@ -0,0 +1,132 @@
...
+    return this.mockStatus !== BiometricsStatus.HardwareUnavailable;
+  }
+
+  async needsSetup(): Promise<boolean> {
+    return false;
+  }
+
+  async canAutoSetup(): Promise<boolean> {
+    return false;
+  }
```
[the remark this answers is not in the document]
quexten: Follow-up improvement. Setup is linux only.

apps/desktop/src/key-management/biometrics/automation-biometrics.service.ts
```diff
@@ -0,0 +1,132 @@
...
+ * biometric status is settable. Biometric keys are held in memory only (no OS keychain), so they do
+ * not persist across restarts.
+ */
+export class AutomationBiometricsService implements OsBiometricService {
+  private mockStatus = BiometricsStatus.Available;
+  private keys = new Map<UserId, SymmetricCryptoKey>();
+  private pendingRequests: PendingRequest[] = [];
+  private nextRequestId = 1;
+
+  constructor(private logService: LogService) {}
```
mzieniukbw: nit: 
```suggestion
  constructor(private readonly logService: LogService) {}
```

apps/desktop/src/key-management/biometrics/automation-biometrics.service.ts
```diff
@@ -0,0 +1,132 @@
...
+  private resolveRequests(id: string | undefined, approved: boolean): void {
+    const matches =
+      id == null
+        ? this.pendingRequests.splice(0, this.pendingRequests.length)
+        : this.pendingRequests
+            .filter((r) => r.id === id)
+            .map((r) => {
+              this.pendingRequests.splice(this.pendingRequests.indexOf(r), 1);
+              return r;
+            });
```
mzieniukbw: I think this is a bit easier to understand, given ids are unique:
```suggestion
  let matches: PendingRequest[];                                                                                                
  if (id == null) {                                                                                                             
    matches = this.pendingRequests.splice(0, this.pendingRequests.length);                                                      
  } else {                                                                                                                      
    const index = this.pendingRequests.findIndex((r) => r.id === id);                                                           
    matches = index === -1 ? [] : this.pendingRequests.splice(index, 1);                                                        
  } 
```

mzieniukbw (dismissed):

mzieniukbw (dismissed):

mzieniukbw (approved):

PR #21926, opened 2026-07-17
refactor: gate desktop_native core allocator and windows deps

quexten:
The follow-up PR to this adds `bitwarden-crypto` to the desktop_native workspace. `bitwarden-crypto` has its own zeroizing alloc, which means we need to remove it for windows.

### DIFF
--- a/apps/desktop/desktop_native/core/Cargo.toml
+++ b/apps/desktop/desktop_native/core/Cargo.toml
@@ -42,6 +42,8 @@ tokio = { workspace = true, features = ["io-util", "sync", "macros", "net"] }
 tokio-util = { workspace = true, features = ["codec"] }
 tracing = { workspace = true }
 typenum = { workspace = true }
+
+[target.'cfg(not(windows))'.dependencies]
 zeroizing-alloc = { workspace = true }
 
 [target.'cfg(target_os = "linux")'.dependencies]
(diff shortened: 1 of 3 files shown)


apps/desktop/desktop_native/core/Cargo.toml
```diff
@@ -65,19 +67,10 @@ scopeguard = { workspace = true }
 secmem-proc = { workspace = true }
 widestring = { workspace = true, optional = true }
 windows = { workspace = true, features = [
-    "Foundation",
```
quexten, line 68: These features are unused

apps/desktop/desktop_native/core/src/lib.rs
```diff
@@ -22,7 +22,10 @@ pub mod process_isolation;
 #[allow(missing_docs)] // staged to be removed
 pub mod ssh_agent;
 
+#[cfg(not(target_os = "windows"))]
 use zeroizing_alloc::ZeroAlloc;
```
quexten, line 26: Please note, this will not be merged by itself. We want to keep the zeroizing alloc, obviously, and the follow-up PR on the stack adds it again via bitwarden-crypto.

dani-garcia (dismissed):
LGTM, why not import bitwarden_crypto on all platforms though? I assume we'll need it eventually anyway

quexten:
> LGTM, why not import bitwarden_crypto on all platforms though? I assume we'll need it eventually anyway

@dani-garcia won't cargo udeps complain? I guess we could supress the warning, but on the other hand we could also just wait until we use it..

dani-garcia:
> > LGTM, why not import bitwarden_crypto on all platforms though? I assume we'll need it eventually anyway
> 
> @dani-garcia won't cargo udeps complain? I guess we could supress the warning, but on the other hand we could also just wait until we use it..

Hmm good point, the `global_allocator` means the crate will be used, but there's a chance that udeps treats it as a false positive. We can definitely just wait until it's needed.

dani-garcia (dismissed):

dani-garcia (approved):

PR #21939, opened 2026-07-17
[PM-38333] feat: Annual billing savings offer on cancel

kdenney:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-38333

## Associated PRs
- https://github.com/bitwarden/server/pull/8010

## 📔 Objective

Adds the client side of the annual billing savings offer. When an owner of a Teams or Enterprise monthly organization opens the cancellation offboarding survey and selects the cost reason, an inline callout shows the projected annual savings and an option to switch to annual billing at the next renewal. Accepting calls the redeem endpoint and closes the flow; declining continues the normal cancellation. The organization subscription page also reflects a pending switch during the window between redemption and renewal.

### Changes

- `OrganizationBillingClient`: methods to fetch the annual upgrade offer and redeem it.
- Offboarding survey: loads the offer on init (failures are swallowed so the survey still works), shows the savings callout on the cost cancellation reason, and redeems on accept. The callout is gated on the cancellation reason whose label refers to cost, accounting for the legacy reason value codes not matching their displayed labels. Layout and copy match the design (two-column label and right-aligned amount).
- Organization subscription page: parses a new `pendingAnnualUpgrade` field on the subscription response and, when present, shows the scheduled annual plan name, per-year cadence, and annual amounts with an effective-date callout, instead of the stale monthly values.
- `subscription-status` component: derives the displayed plan name from `pendingAnnualUpgrade` when a switch is scheduled.
- New English locale strings for the callout and the subscription-page pending state.
- Unit tests for the billing client methods, the offboarding survey callout behavior, the subscription page, the subscription-status component, and the response model.

## 📸 Screenshots

image

image

image


apps/web/src/app/billing/organizations/subscription-status.component.ts
```diff
@@ -52,7 +52,10 @@ export class SubscriptionStatusComponent {
   }
 
   get planName() {
-    return this.organizationSubscriptionResponse.plan.name;
+    return (
```
sbrown-livefront, line 55: ⛏️ Do you think this would be a good chance to modernize this component or would it be too far out of scope?

sbrown-livefront (dismissed):
✅ small question of modernizing a component.

apps/web/src/app/billing/organizations/subscription-status.component.ts
kdenney: Too far out of scope imo. This ticket has already ballooned quite a lot so I'll leave this for a future ticket lol

apps/web/src/app/billing/shared/offboarding-survey.component.ts
```diff
@@ -159,6 +189,68 @@ export class OffboardingSurveyComponent {
...
+    }
+  }
+
+  private async loadAnnualUpgradeOffer(organizationId: string): Promise<void> {
+    // Checked here as well as on the server. The endpoint answers 404 when the flag is off, and
+    // the catch below swallows and logs that, so an unguarded call would put an error in the log
+    // for every cancellation dialog in a flag-off environment.
+    if (!(await this.configService.getFeatureFlag(FeatureFlag.PM38333_AnnualBillingSavings))) {
+      return;
+    }
```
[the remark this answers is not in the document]
kdenney: I hadn't pushed the server repo yet. It's there now

apps/web/src/app/billing/shared/offboarding-survey.component.ts
```diff
@@ -159,6 +190,68 @@ export class OffboardingSurveyComponent {
...
+      );
+    } catch (e) {
+      this.logService.error(e);
+    }
+  }
+
+  protected formatCurrency(amount: number): string {
+    // Mirror the price-increase-warning precedent: whole-dollar amounts show no cents,
+    // fractional amounts show two decimals.
+    const digitsInfo = Number.isInteger(amount) ? "1.0-0" : "1.2-2";
```
sbrown-livefront, line 223: ⛏️ We use this same formatting call 2 - 3 different places. Maybe making this an accessible function could be helpful in the future. Not a blocking change.

apps/web/src/app/billing/shared/offboarding-survey.component.html
```diff
@@ -43,7 +82,13 @@
     </bit-form-field>
   </div>
   <ng-container bitDialogFooter>
-    <button bitButton bitFormButton buttonType="primary" type="submit" [disabled]="!reason()">
+    <button
+      bitButton
+      bitFormButton
+      buttonType="primary"
+      type="submit"
+      [disabled]="!reason() || annualUpgradeRedeemLoading()"
```
sbrown-livefront, line 90: ❓ Do we need cancellationLoading on this also?

sbrown-livefront (approved):
✅ A few non-blocking comments

apps/web/src/app/billing/shared/offboarding-survey.component.html
kdenney: No, the bitSubmit/bitFormButton directives already control the loading and disable it for the whole cancellation. cancellationLoading is only needed to cross-disable the redeem button.

apps/web/src/app/billing/shared/offboarding-survey.component.ts
kdenney: Good idea and I think it could be useful to define as a new angular pipe too but I am going to defer this one to later in the interest of getting this ticket unblocked. Thanks for the suggestion though!

PR #21957, opened 2026-07-20
fix(key-connector): fix sdk key-connector login not working

quexten:

The sdk login path incorrectly detected users that have not yet migrated with key-connector, but should migrate, as 'canUnlockWithkeyConnector()', which lead to a failed unlock attempt, when unlocking with the SDK key-connector feature-flag enabled.


ike-kottlowski (approved):

quexten:
Changes are behind the unreleased flag so merging.

PR #21958, opened 2026-07-20
refactor(login-strategy): clean up crypto-related functions and deprecation notices

quexten:
follow-up to https://bitwarden.atlassian.net/browse/PM-30584; noticed these clean-up opportunities (each in a separate commit, feel free to review commit-by-commit).


libs/auth/src/common/login-strategies/sso-login.strategy.ts
```diff
@@ -346,14 +346,7 @@ export class SsoLoginStrategy extends LoginStrategy {
   protected override async setAccountCryptographicState(
```
ike-kottlowski: ❓: In all other strategies this method was removed is there a reason it's left in the SSO strategy?

quexten: Removed!

ike-kottlowski (dismissed):

ike-kottlowski (approved):

PR #21968, opened 2026-07-20
[PM-40510] feat: rename vault-filter query param to sharedFolderId

jaasen-livefront:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-40510

## 📔 Objective

Behind the `vfo1-foundation` feature flag (off by default), rename the vault-filter query param from `collectionId` to `sharedFolderId` in the shared `RoutedVaultFilterService`:

- **Reader** (`filter$`) accepts **both** `sharedFolderId` and legacy `collectionId` (preferring the new one), so existing/bookmarked `?collectionId=` links keep working during the transition. The `RoutedVaultFilterModel.collectionId` field name is unchanged — only the URL param is renamed.
- **`createRoute()`** writes `sharedFolderId` and nulls `collectionId` when the flag is on; writes `collectionId` and nulls `sharedFolderId` when off (`queryParamsHandling: "merge"` clears the stale param).

Split from PM-40252 (terminology-only) — it is the only routing/behavioral change and is coupled to the per-client reader/writer migrations. App-level direct query-param readers/writers (apps/web `vault.component`, `vault-collection-actions.service`; apps/browser `new-item-page`) are migrated in the per-client tickets; until then the flag stays off, so no broken window.


jengstrom-bw (approved):
LGTM!

PR #21969, opened 2026-07-20
[PM-40654] feat: gate org-creation default collection name behind VFO1 terminology

jaasen-livefront:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-40654
https://bitwarden.atlassian.net/browse/PM-40882

Part of epic PM-39699 (Change "collections" to "shared folders"). Independent of PM-40252 — branches off `main`.

## 📔 Objective

The collection auto-created at organization setup is named via the `defaultCollection` i18n key ("Default collection"). Behind the `vfo1-foundation` flag, it should instead read **"Default shared folder"**.

This gates the name at all five org-creation call sites:

- `organization-billing.service.ts` (libs/common) — purchase / free / restart flows
- `default-organization-invite.service.ts` (libs/common) — accept-and-init flow
- `premium-org-upgrade.service.ts` (web) — premium → org upgrade
- `organization-self-hosting-license-uploader.component.ts` (web) — self-host license upload
- `web-provider.service.ts` (bit-web) — MSP/provider client-org creation

Each site resolves `ConfigService.getFeatureFlag(VFO1Foundation)` and selects the `defaultSharedFolder` / `defaultCollection` key. The two `libs/common` services gain a `ConfigService` dependency (both `deps` arrays updated in `jslib-services.module.ts`). `WebProviderService` also gains a `ConfigService` dependency; it uses Angular constructor injection with no `deps` array, so it resolves automatically.

Unlike the display-only terminology work elsewhere in the epic, this renames the **persisted** collection name, so it reads the resolved flag value rather than the `Vfo1TerminologyService` signal (which defaults to `true` before the flag resolves). New `defaultSharedFolder` key added to web's `en` locale; `defaultCollection` retained for the flag-off path.

Behavior only changes for orgs created after deploy.

### DIFF
--- a/apps/web/src/app/billing/individual/upgrade/premium-org-upgrade-payment/services/premium-org-upgrade.service.spec.ts
+++ b/apps/web/src/app/billing/individual/upgrade/premium-org-upgrade-payment/services/premium-org-upgrade.service.spec.ts
@@ -7,8 +7,10 @@ import {
   BusinessSubscriptionPricingTierIds,
   PersonalSubscriptionPricingTierIds,
 } from "@bitwarden/common/billing/types/subscription-pricing-tier";
+import { FeatureFlag } from "@bitwarden/common/enums/feature-flag.enum";
 import { EncryptService } from "@bitwarden/common/key-management/crypto/abstractions/encrypt.service";
 import { EncString } from "@bitwarden/common/key-management/crypto/models/enc-string";
+import { ConfigService } from "@bitwarden/common/platform/abstractions/config/config.service";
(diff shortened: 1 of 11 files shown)


enmande (dismissed):
Auth changes lgtm!
Nice, minimal change hooked into the VFO1Foundation flag itself, feeds straight through to the Organization Init flow for the collection name.
Thank you for the tests expressing behaviors with and without the flag on! 🧪

JimmyVo16 (approved):

enmande (approved):

cyprain-okeke (approved):
Looks good for billing

PR #21974, opened 2026-07-21
feat(data-recovery): add fido2 step

quexten:
Add a step to detect and recover from fido2 errors. The assumption currently is that specifically only the fido2 part may be getting corrupted. This is not yet confirmed, and this step will help to confirm it. It is also entirely possible that these are field-encrypted-cipher-corruption-issues.

### DIFF
--- a/apps/web/src/app/key-management/data-recovery/steps/corruption-log.ts
+++ b/apps/web/src/app/key-management/data-recovery/steps/corruption-log.ts
@@ -1,0 +1,14 @@
+import { Cipher } from "@bitwarden/common/vault/models/domain/cipher";
+
+import { LogRecorder } from "../log-recorder";
+
+/**
+ * Records diagnostic details about a corrupt cipher: its id and whether it carries a per-item
+ * vault item key (cipher key). No decrypted data is logged.
+ */
+export function logCipherCorruption(cipher: Cipher, logger: LogRecorder, note?: string): void {
+  const cipherKeyPresent = cipher.key != null;
+  logger.record(
+    `Corrupt cipher ${cipher.id}${note ? ` (${note})` : ""}: cipherKeyPresent=${cipherKeyPresent}`,
+  );
+}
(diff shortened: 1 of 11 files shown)


apps/web/src/app/key-management/data-recovery/steps/corruption-log.ts
```diff
@@ -0,0 +1,14 @@
+import { Cipher } from "@bitwarden/common/vault/models/domain/cipher";
+
+import { LogRecorder } from "../log-recorder";
+
+/**
+ * Records diagnostic details about a corrupt cipher: its id and whether it carries a per-item
+ * vault item key (cipher key). No decrypted data is logged.
+ */
+export function logCipherCorruption(cipher: Cipher, logger: LogRecorder, note?: string): void {
+  const cipherKeyPresent = cipher.key != null;
```
quexten, line 10: At a later stage we will log encrypted-by data here, to trace the key hierarchy.

apps/web/src/app/key-management/data-recovery/steps/fido2-step.ts
```diff
@@ -0,0 +1,163 @@
...
+import { DialogService } from "@bitwarden/components";
+import { UserId } from "@bitwarden/user-core";
+
+import { LogRecorder } from "../log-recorder";
+
+import { logCipherCorruption } from "./corruption-log";
+import { RecoveryStep, RecoveryWorkingData } from "./recovery-step";
+
+/**
+ * Detects Login ciphers whose only decryption failure is corrupt FIDO2 credentials, and repairs
```
quexten, line 15: Please note: It is unclear whether this is actually a case that is present. We currently think it *may* be. If the theory turns out to be wrong, this step will be removed again.

apps/web/src/app/key-management/data-recovery/steps/fido2-step.ts
```diff
@@ -0,0 +1,163 @@
...
+    logger.record(
+      `Successfully removed FIDO2 credentials from ${workingData.fido2CorruptCipherIds.length} ciphers`,
+    );
+  }
+
+  /**
+   * Returns true when the cipher decrypts without a decryption failure, false on any error or a
+   * flagged decryption failure.
+   */
+  private async decryptSucceeds(cipher: Cipher, userId: UserId): Promise<boolean> {
```
quexten, line 142: Note: We should consider porting this to SDK. The problem will be transferring the step progress, but I'm sure we can figure something out there...

PR #21980, opened 2026-07-21
[PM-40685] Fix forwarder settings cache collision

cyfrit:
## 🎟️ Tracking

Community-reported browser extension bug. No matching GitHub issue was found after searching existing reports.

https://bitwarden.atlassian.net/browse/PM-40685

## 📔 Objective

Email forwarder algorithms use object identifiers containing their vendor IDs. Interpolating these objects directly into the generator settings cache key causes every forwarder to share the same cache entry.

After generation fails for an unconfigured forwarder, switching to a configured forwarder can reuse the previous forwarder's empty settings and incorrectly report an invalid API token.

This change derives forwarder cache keys from their vendor IDs and adds a regression test covering recovery from a failed generation followed by switching forwarders.

>I selected an email forwarding provider that I had not configured and accidentally clicked Generate. As expected, this produced an invalid API token error. I then switched to a different provider that was correctly configured and clicked Generate again, but it also reported an invalid API token despite having valid settings.

>After investigating, I found that the empty settings loaded for the first provider were being reused for the second provider because their cache keys collided.


harr1424:
Hi @cyfrit thanks for taking the time to submit this bug fix. Your code changes look good to me. Could you please link to where this issue was first reported? Or at least where you became aware of it?

cyfrit:
Hi @harr1424, I first noticed this myself while using the Bitwarden browser extension, so there is no external report to link.

Specifically, I selected an email forwarding provider that I had not configured and accidentally clicked Generate. As expected, this produced an invalid API token error. I then switched to a different provider that was correctly configured and clicked Generate again, but it also reported an invalid API token despite having valid settings.

After investigating, I found that the empty settings loaded for the first provider were being reused for the second provider because their cache keys collided.

harr1424:
@cyfrit thank you for providing the exact replication steps, I am working on replicating this now and will then begin testing your contribution.

harr1424 (approved):

PR #21982, opened 2026-07-21
[PM-40682] desktop-autofill: Graceful Windows plugin registration

iinuwa:
## 🎟️ Tracking

[PM-40682](https://bitwarden.atlassian.net/browse/PM-40682)

## 📔 Objective

The Windows plugin doesn't seem to provide a lookup or upsert for plugin details, so we check whether the add method fails due to already existing, and then call update. This avoids unnecessary errors in the electron app, and keeps it up to date every time the app starts.

[PM-40682]: https://bitwarden.atlassian.net/browse/PM-40682?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/desktop/desktop_native/win_webauthn/src/api/plugin/mod.rs
+++ b/apps/desktop/desktop_native/win_webauthn/src/api/plugin/mod.rs
@@ -347,10 +347,13 @@ impl WebAuthnPlugin {
 
     /// Adds this implementation as a Windows WebAuthn plugin.
     ///
+    /// Returns None if the authenticator already is registered; call
+    /// [WebAuthnPlugin::update_authenticator_details] to update the details.
+    ///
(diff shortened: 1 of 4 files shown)


abergs (approved):

abergs:
I thought there was a querying API, we might want to check if that was removed

iinuwa:
🤦 I checked this like 8 times to make sure I wasn't crazy, but I missed it every time: [WebAuthNPluginGetAuthenticatorState()](https://learn.microsoft.com/en-us/windows/win32/api/webauthnplugin/nf-webauthnplugin-webauthnplugingetauthenticatorstate).

Regardless, it's good for us to update the credential details.

PR #21984, opened 2026-07-21
[PM-40449] create lit cipher items and list

dan-livefront:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-40449
## 📔 Objective

Add Lit components for inline menu cipher items and lists with Storybook coverage. Doing exact parity of what we have today but there will be design changes after these components get adopted. 
## 📸 Screenshots

Screenshot 2026-07-21 at 1 37 08 PM
Screenshot 2026-07-21 at 1 36 32 PM
Screenshot 2026-07-21 at 1 37 16 PM
Screenshot 2026-07-21 at 1 37 24 PM


apps/browser/src/autofill/content/components/inline-menu/cipher-item.ts
audreyality, line 1: ⛏️ Seems to me like each of the components should be stored in its own file.

audreyality (dismissed):
Just one nit. 

❓ Were tests omitted because of the pending UI changes?

apps/browser/src/autofill/content/components/inline-menu/cipher-item.ts
dan-livefront: Great call out, I considered splitting these into separate row components for login, TOTP, and passkeys, but kept a single cipher item since the row structure, accessibility, and keyboard navigation are shared so I tried not to duplicate that logic.

What do you think about extracting the type specific details instead like the TOTP countdown for example while keeping one shared wrapper?

dan-livefront: >❓ Were tests omitted because of the pending UI changes?

Yep! I am planning on testing everything as a final pass

audreyality: > What do you think about extracting the type specific details instead like the TOTP countdown for example while keeping one shared wrapper?

I think you're saying the same thing I am? The decomposition is fine, it's just that `InlineMenuCipherItem`, `CipherDetails`, `TotpCountdown`, and `TotpCipherInfo` should each be in their own file.

dan-livefront: Cool beans I will do that

audreyality (dismissed):

audreyality (approved):

PR #21988, opened 2026-07-21
[PM-40726] Add filter-as-you-type to autofill inline menu

chiefbrain:
This pull request implements the same filter-as-you-type behavior for the autofill inline menu as #20543. The two take different approaches; the main differences are below.

**Where filtering is applied**
- This pull request builds the full inline-menu cipher list once, caches it, and applies the filter to the cached built data on each keystroke.
- #20543 filters the source cipher list at the start of `getInlineMenuCipherData` and rebuilds the matching entries on each keystroke (including regenerating TOTP codes via `buildCipherData`).

**Cipher count tracking**
- This pull request keeps two counts: `currentInlineMenuCiphersCount` (total built) and `filteredInlineMenuCiphersCount` (after filtering). `checkIsInlineMenuCiphersPopulated` reads the total; the show/hide decision reads the filtered count.
- #20543 sets `currentInlineMenuCiphersCount` to the filtered length and reads that value both for the show/hide decision and in `checkIsInlineMenuCiphersPopulated`.

**Matched fields**
- This pull request matches: cipher name, login username, login URI, passkey userName, passkey rpName, card subtitle (brand + last four), identity full name, identity username.
- #20543 matches: cipher name, login username, card number, identity full name.

**Card matching**
- This pull request matches the card subtitle as shown in the list.
- #20543 matches the full card number.

**Text normalization**
- This pull request normalizes the query and candidate values with Unicode NFD + combining-mark stripping in addition to lowercasing (mirroring `normalizeSearchQuery` from the vault search service).
- #20543 lowercases the query and candidate values.

**Filter state and lifecycle**
- This pull request stores the active filter value as instance state and clears it on focus change and on a full cipher refresh (`handleOverlayCiphersUpdate`).
- #20543 passes the filter value as a message/method parameter and does not store it.

**Empty-result handling**
- This pull request hides the list via `toggleInlineMenuHidden` when the filtered count is zero.
- #20543 closes the list via `closeInlineMenu` with `forceCloseInlineMenu` when the count is zero and a filter is set.

**Passkey section labels**
- This pull request recomputes `showPasskeysLabelsWithinInlineMenu` from the filtered result set.
- #20543 sets it while building from the already-filtered source list.

**Content-script input handling**
- This pull request sends `filterAutofillInlineMenu` with the field value and, when the field is empty, also sends `openAutofillInlineMenu`.
- #20543 sends `filterInlineMenuCiphers` with the current field value (empty string when the field is empty).

**Tests**
- This pull request adds background tests for filter matching across login/card/identity ciphers, URI and diacritic matching, the total-vs-filtered count, reuse of the built data while filtering, account-creation filtering, and the filter reset on focus/refresh, alongside content-service tests.
- #20543 updates the content-service tests for the new message.

### DIFF
--- a/apps/browser/src/autofill/background/abstractions/overlay.background.ts
+++ b/apps/browser/src/autofill/background/abstractions/overlay.background.ts
@@ -132,6 +132,10 @@ export type CloseInlineMenuMessage = {
   overlayElement?: string;
 };
 
+export type FilterInlineMenuMessage = {
+  filterValue?: string;
+};
+
 export type ToggleInlineMenuHiddenMessage = {
   isInlineMenuHidden?: boolean;
   setTransparentInlineMenu?: boolean;
@@ -161,6 +165,7 @@ export type OverlayBackgroundExtensionMessage = {
   iframeTargetedFields?: { selector: string; fieldType: string; formCategory?: string }[];
 } & OverlayAddNewItemMessage &
   CloseInlineMenuMessage &
+  FilterInlineMenuMessage &
   ToggleInlineMenuHiddenMessage &
   UpdateInlineMenuVisibilityMessage;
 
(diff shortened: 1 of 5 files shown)


pamperer562580892423:
@chiefbrain I'm just another user, but since you compare your PR so meticulously to that other PR - did you see that there is a third PR trying to implement the same thing: https://github.com/bitwarden/clients/pull/20836 ?!

chiefbrain:
> @chiefbrain I'm just another user, but since you compare your PR so meticulously to that other PR - did you see that there is a third PR trying to implement the same thing: #20836 ?!

If have seen it but just had a brief look before. On evaluating it now, it seems as the most advanced PR, my PR could only contribute by using a cache (not rebuilding the whole list on every keystroke) and by adding card/identity fields.

As mentioned in #20543, the use of this PR is mostly a reference. It could be useful if a decision in the internal ticket is made that a cache is needed.

jprusik:
I reassigned this one to @audreyality since we have a few PRs with different approaches to this and she's presently coordinating the community PRs and our internal Product team developments around this ask.

audreyality:
Hi @chiefbrain,

PR #21992, opened 2026-07-22
[PM-39452] fix: Show premium membership status from desktop Account menu

### LINKED ISSUE #21452: Premium membership not reflected in macOS desktop application
### Steps To Reproduce

1. Upgrade account to premium via browser plugin and thence the bitwarden website.
2. Log out of Bitwarden application on macOS.
3. Quit Bitwarden application.
4. Restart Bitwarden application.
5. Log in.
6. Account -> Premium membership shows "Upgrade to premium".
7. Click the "Upgrade now" button.
8. The website shows that the account has already been upgraded.

### Expected Result

I would expect Account -> Premium membership to show some evidence that premium membership is in force.

### Actual Result

Account -> Premium membership shows "Upgrade to premium".

### Screenshots or Videos

Image

### Additional Context

_No response_

### Operating System

macOS

### Operating System Version

15.7.7 (24G720)

### Installation method

### Build Version

Version 2026.5.0 (62263)

### Issue Tracking Info

cyprain-okeke:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-39452

Resolves #21452

## 📔 Objective

Fixes the desktop regression where **Account → Premium membership** showed the "Upgrade to premium" dialog to users who already have premium (reported on desktop v2026.5.0, direct download).

**Root cause:** #20287 removed the `PM23713_PremiumBadgeOpensNewPremiumUpgradeDialog` feature flag (first shipped in desktop v2026.5.0), which made `DesktopPremiumUpgradePromptService.promptForPremium()` unconditionally open the status-blind `PremiumUpgradeDialogComponent`. Only direct-download builds are affected — store builds hide the menu item entirely (`visible: !isWindowsStore() && !isMacAppStore()`).

**Fix:** make `promptForPremium()` premium-status-aware, mirroring the guard in `WebVaultPremiumUpgradePromptService`:

- Resolves the active account; logs a warning and no-ops if none (unexpected — the menu item is disabled while locked).
- Checks `BillingAccountProfileStateService.hasPremiumFromAnySource$(account.id)`.
- **Premium** → opens the status-aware `PremiumComponent` membership view (restores the pre-#20287 behavior).
- **Non-premium** → unchanged upgrade dialog.
- Body wrapped in try/catch with `LogService.error` — the caller's message loop (`app.component.ts` `openPremium` case) floats the promise, so a rejection would otherwise vanish unlogged.

**Testing:**

- Unit spec extended 1 → 5 tests: both dialog branches with mutual-exclusion assertions, premium check pinned to the active account id, no-account warning path, and error-path resolution (no unhandled rejection).
- E2E-verified on macOS against a local server: a premium user sees the "You are a premium member!" membership view on fresh login **and** after quit → restart → unlock (the reporter's exact repro); a non-premium user still gets the upgrade dialog.

**QA note:** please visually verify the premium-user path (Account → Premium membership while premium) renders the membership dialog correctly — unit specs mock `DialogService`, so component rendering inside the CDK dialog is only covered by manual/e2e verification.

## 📸 Screenshots

**Before** (from #21452): premium user shown "Upgrade to premium".

**After (local e2e):** premium user now sees the membership view; non-premium user still sees the upgrade dialog.


apps/desktop/src/billing/services/desktop-premium-upgrade-prompt.service.ts
```diff
@@ -1,19 +1,50 @@
...
+        );
+        return;
+      }
+
+      const hasPremium = await firstValueFrom(
+        this.billingAccountProfileStateService.hasPremiumFromAnySource$(account.id),
+      );
+
+      if (hasPremium) {
+        // Users with premium should see the status-aware membership view rather than the upgrade pitch.
```
amorask-bitwarden: ❓ What is the status-aware membership view? On line 36, you're checking `hasPremiumFromAnySource$`, but if they have premium from an organization, they're not going to have a status or membership.

cyprain-okeke: Good question — that comment was confusing, so I reworded it.

The "status-aware membership view" is PremiumComponent — the dialog the Account menu opened before PM-29073. It branches on its own isPremium$ (also hasPremiumFromAnySource$), so org-premium users see the "premium member" view regardless of what this service checks — same as pre-#20287 and web's guard. Checking hasPremiumPersonally$ here would instead pitch a purchase to users whose premium features already work.

amorask-bitwarden (dismissed):

amorask-bitwarden (approved):

PR #21994, opened 2026-07-22
[PM-40766] Handle punycode-unicode domain matching for "Host" URI matching strategy

### LINKED ISSUE #4372: Internationalized Domain Names (IDN) unsupported
### Step 1
1. Create a new entry:
2. URI 1: medæøå.dk
3. Username: Bitwarden
4. Password: Hr*UX6$gmBx@qJJZKjYj!mvBcHHYZD6mwmya4sNhb%zVVpNE6D^nFyeVqyv8@ui%6#sWmc$mPB8z

Please note: Site is set up only for this example, and will at some point die. 

### Step 2
Go to [medæøå.dk/wp-login.php](https://medæøå.dk/wp-login.php) and click the Bitwarden extension to open possible entries.

### Expected Result

The login info for user Bitwarden is available in the browser extension ready for autofill immediately.

### Actual Result

No login info is available in the browser extension. Searching for "medæøå" shows the login info but you have to open the item and choose "Auto-fill". 

### Screenshots or Videos

_No response_

### Additional Context

It appears that when Bitwarden matches uris it does so without respecting IDN, leaving out native support for users with non-ASCII keyboards (compassing large parts of EMEA and Asia). Browsers started a long time ago to convert from Punycode to Unicode in the address bar to support native users.

As I am Danish, my main focus has been on the Danish special characters æ, ø, and å. As far as I know the support of these are no different from many European languages, Cyrillic, Arabic, Chinese Simplified, and many others (see [Wikipedia on IDN](https://en.wikipedia.org/wiki/Internationalized_domain_name)).

If a user adds the login through the browser extension the uri is saved correctly and autofill works as intended. 

### Operating System

Windows

### Operating System Version

10

### Web Browser

Chrome, Firefox

### Browser Version

Firefox 108.0.1, Chrome 108.0.5359.125.

### Build Version

Version: 2022.12.1  Server Version: 2022.12.0 

### Issue Tracking Info

jprusik:
## 🎟️ Tracking

[PM-40766](https://bitwarden.atlassian.net/browse/PM-40766)

## 📔 Objective

Addressed https://github.com/bitwarden/clients/issues/4372

> we support IDN/punycode comparison when using the "Host" URI match detection strategy, but not when the "Base domain" (the out-of-the-box default) strategy is used

[PM-40766]: https://bitwarden.atlassian.net/browse/PM-40766?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/common/src/platform/misc/utils.spec.ts
+++ b/libs/common/src/platform/misc/utils.spec.ts
@@ -221,6 +221,42 @@ describe("Utils Service", () => {
       expect(Utils.getHostname("https://subdomain.bütwarden.com")).toBe("subdomain.bütwarden.com");
     });
 
+    // Test examples here gathered from https://www.iana.org/domains/reserved
+    it("should support urls containing unicode characters", () => {
+      // Arabic Arabic
+      expect(Utils.getHostname("https://www.إختبار.com")).toBe("www.إختبار.com");
+
+      // Persian Arabic
+      expect(Utils.getHostname("https://www.آزمایشی.com")).toBe("www.آزمایشی.com");
+
+      // Chinese Han (Simplified variant)
+      expect(Utils.getHostname("https://www.测试.com")).toBe("www.测试.com");
+
+      // Chinese Han (Traditional variant)
+      expect(Utils.getHostname("https://www.測試.com")).toBe("www.測試.com");
+
+      // Russian Cyrillic
+      expect(Utils.getHostname("https://www.испытание.com")).toBe("www.испытание.com");
+
+      // Hindi Devanagari (Nagari)
+      expect(Utils.getHostname("https://www.परीक्षा.com")).toBe("www.परीक्षा.com");
+
+      // Greek, Modern (1453-) Greek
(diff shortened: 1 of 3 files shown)


libs/common/src/platform/misc/utils.spec.ts
jprusik, line 1: These additions are strictly unrelated, but during my investigation I felt like these were worthwhile defensive measures against future IDN edge regressions.

jprusik:
We might also consider how we want to handle punycode/unicode matching in the other URI matching strategies, but they strike me as more concerned with character equivalence over compound value equivalence.

Elsewhere, we probably want to take a look at these comparison cases for phishing domains, equivalent domains, and blocked/excluded domains.

(I explicitly treated these concerns as out-of-scope for this task)

dani-garcia (approved):

gbubemismith (approved):

Issue #19858, opened 2026-03-30
[PM-34358] Unable to create/store a PRF-capable login-passkey with encryption on Windows Hello


pamperer562580892423:
### Steps To Reproduce

**(important: see "Additional Context" also for the necessary conditions to reproduce this!)**

1. Log in to your web vault.
2. Go to Settings --> Security --> Master password --> Log in with passkey
3. Try to add a new login-passkey and store it in Windows Hello
4. Follow that procedure until the passkey is successfully created and stored.

### Expected Result

On the latest Windows 11 and Chrome version (see the "Additional Context" section for important details!), it should be possible to create and store a PRF-capable login-passkey with Windows Hello.

So, I should get asked to create that login-passkey "with encryption" - and in the end, it should be shown in the web vault as "Used for encryption".

### Actual Result

In the creation-process, I don't get asked for "with encryption?". In the end, the web vault says "Encryption not supported" for the just created login-passkey.

### Screenshots or Videos

https://github.com/user-attachments/assets/2a524be9-cb55-4ecf-b9c9-896369a30338

&nbsp;

Overview to the video:
- 0:00 - 1:07 Log in & passkey creation
- 1:07 - 1:15 Chrome & Win 11 versions
- 1:15 - 1:44 Corbado PRF-Test with Windows Hello
- 1:44 - 1:54 Both created passkeys (for BW and the one for Corbado) listed in Windows Hello
- 1:54 - 2:05 (nothing relevant, really...)
- 2:05 - 2:18 Using the just created Windows Hello login-passkey without PRF fails with an error (but that's a separate issue - see https://github.com/bitwarden/clients/issues/16075)

&nbsp;

Some remarks to the video:
- I logged in to the web vault via an existing login-passkey on a hardware security key (I was on the US server first, and hat to change to the EU server region, that's why the video starts a bit "confusing")
- it's a free BW test account I use, so the things that are not redacted are not relevant
- the other existing Windows Hello passkey in the web vault is old and from another Win 11 computer - and not relevant for that video or issue
- recognize, that in the whole passkey creation process, I don't get asked "with encryption?", though the Corbado test etc. shows PRF is possible with Windows Hello
- just like I described it here, the login-passkey is then just shown as "Encryption not supported" (though, again, Windows Hello is PRF-capable and _encryption_ should be supported)
- versions when I recorded the video (I only added the video, video overview and these remarks here on Apr 14, 2026, while I submitted the original report on Mar 30, 2026):
[...]

PR #21998, opened 2026-07-22
webauthn: Pass PRF input on unlock passkey creation

iinuwa:
## 🎟️ Tracking

[PM-40775](https://bitwarden.atlassian.net/browse/PM-40775)

## 📔 Objective

Two goals here:
- WebAuthn has been updated to allow passing PRF inputs on create, so you don't have to do two authentication ceremonies.
- Windows Hello supports PRF, but there's a bug where they don't signal PRF support if you don't pass PRF input on create, which is not compliant with the spec. This means that our users couldn't use Windows Hello passkeys to unlock their vaults.

Fixing the former also fixes the latter. 😎 

## 📸 Screenshots

https://github.com/user-attachments/assets/ddfd3eee-4973-40b3-a9f6-26569953cbd0

[PM-40775]: https://bitwarden.atlassian.net/browse/PM-40775?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/web/src/app/auth/core/services/webauthn-login/webauthn-login-admin.service.ts
+++ b/apps/web/src/app/auth/core/services/webauthn-login/webauthn-login-admin.service.ts
@@ -116,19 +116,26 @@ export class WebauthnLoginAdminService implements UserKeyRotationDataProvider<We
     const nativeOptions: CredentialCreationOptions = {
       publicKey: credentialOptions.options,
     };
-    // TODO: Remove `any` when typescript typings add support for PRF
     nativeOptions.publicKey.extensions = {
       prf: { eval: { first: await this.webAuthnLoginPrfKeyService.getLoginWithPrfSalt() } },
-    } as any;
+    };
 
     try {
       const response = await this.navigatorCredentials.create(nativeOptions);
       if (!(response instanceof PublicKeyCredential)) {
         return undefined;
       }
-      // TODO: Remove `any` when typescript typings add support for PRF
-      const supportsPrf = Boolean((response.getClientExtensionResults() as any).prf?.enabled);
-      return new PendingWebauthnLoginCredentialView(credentialOptions, response, supportsPrf);
+
+      const prfExtensionOutput = response.getClientExtensionResults().prf;
+      const supportsPrf = Boolean(prfExtensionOutput?.enabled);
+      const prfResult = prfExtensionOutput?.results?.first;
+
+      return new PendingWebauthnLoginCredentialView(
+        credentialOptions,
+        response,
+        supportsPrf,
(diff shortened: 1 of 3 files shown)


apps/web/src/app/auth/core/services/webauthn-login/webauthn-login-admin.service.ts
```diff
@@ -116,19 +116,26 @@ export class WebauthnLoginAdminService implements UserKeyRotationDataProvider<We
     const nativeOptions: CredentialCreationOptions = {
       publicKey: credentialOptions.options,
     };
-    // TODO: Remove `any` when typescript typings add support for PRF
     nativeOptions.publicKey.extensions = {
-      prf: {},
+      prf: { eval: { first: await this.webAuthnLoginPrfKeyService.getLoginWithPrfSalt() } },
     } as any;
```
rr-bw: I notice you removed the TODO comment. Can we also remove corresponding `as any` type cast?

apps/web/src/app/auth/core/services/webauthn-login/webauthn-login-admin.service.ts
rr-bw: Can we also remove this TODO comment and the corresponding `as any` cast?

rr-bw, line 157: Could you add some test coverage per the Claude review note:

> The new direct-PRF branch in createKeySet (using pendingCredential.prfResult and skipping .get()) has no unit-test coverage in webauthn-login-admin.service.spec.ts — the existing tests only cover createCredential, enableCredentialEncryption, and rotation. Since this is an encryption-key-derivation path, a test exercising both the direct and fallback branches would harden the change.

rr-bw (changes requested):

apps/web/src/app/auth/core/services/webauthn-login/webauthn-login-admin.service.ts
iinuwa: Added some tests that cover when the authenticator supports PRF results on get (existing functionality) and on create (added in this PR).

pamperer562580892423:
Sorry for the intrusion, but could you link my bug report (https://github.com/bitwarden/clients/issues/19858) - of course only in case this PR indeed will fix this issue? Thanks for working on this anyway! 🙏

rr-bw (approved):

abergs (approved):
Nice

PR #22006, opened 2026-07-23
[PM-40846] Support Fido2 v4 server update

ike-kottlowski:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-40846

🟥 Blocks: https://bitwarden.atlassian.net/browse/PM-25806, https://github.com/bitwarden/server/pull/6328

## 📔 Objective

In order to merge the Fido2 changes in this renovate PR we have to make modifications to the web client's implementations. The new version of Fido2 is stricter than the earlier version requiring an addition of the `transports` property. 

The class also used a generic utils class, which has now been changed to use the Fido2 specific utils class.

- 🤖  `WebauthnLoginAttestationResponseRequest` / `WebauthnLoginAuthenticatorResponseRequest`: replace `Utils.fromBufferToUrlB64` / `Utils.fromBufferToB64` with `Fido2Utils.bufferSourceToUint8Array` + `Fido2Utils.arrayToString`; populate `transports` from `credential.response.getTransports()`.
- 🤖 `DefaultTwoFactorApiService`: same conversion swap for `rawId`, `AttestationObject`, and `clientDataJson`; fixes `rawId` incorrectly being derived from `btoa(request.deviceResponse.id)` instead of the actual raw credential ID buffer; adds `transports` to the device response sent to the server.
- 🤖 Updates the corresponding spec to mock `getTransports()` and assert on the new `rawId`/`transports` fields.

## 📸 Screenshots
Backwards compatibility verification

### Webauthn/Passkey login and registration

https://github.com/user-attachments/assets/40a2eb66-1284-44b1-8249-629930fa511f

The user logs in using a passkey. The user follows the system prompts select the correct passkey. User successfully logs in and navigates to the master password component in the account settings. User disables Passkey login. User re-enables passkey login and logs out. User then logs in successfully using the newly registered passkey.

### Webauthn MFA

https://github.com/user-attachments/assets/bd0e93c2-dd5a-4cfd-8336-2d3bde9d2700

User logs in with master password and is prompted for webauthn MFA. User follows prompts for MFA and logs in successfully. User navigates to MFA settings and disables webauthn MFA. User re-enables passkey MFA with a new passkey. User logs out. User logs in with master password and is prompted for passkey MFA. User selects newly registered passkey and logs in successfully.


libs/common/src/auth/two-factor/services/default-two-factor-api.service.ts
```diff
@@ -239,12 +239,19 @@ export class DefaultTwoFactorApiService implements TwoFactorApiService {
...
-      rawId: btoa(request.deviceResponse.id),
+      rawId: Fido2Utils.arrayToString(
+        Fido2Utils.bufferSourceToUint8Array(request.deviceResponse.rawId),
+      ),
       type: request.deviceResponse.type,
       extensions: request.deviceResponse.getClientExtensionResults(),
       response: {
-        AttestationObject: Utils.fromBufferToB64(deviceResponse.attestationObject),
-        clientDataJson: Utils.fromBufferToB64(deviceResponse.clientDataJSON),
+        AttestationObject: Fido2Utils.arrayToString(
```
rr-bw, line 248:
```suggestion
        attestationObject: Fido2Utils.arrayToString(
```

libs/common/src/auth/two-factor/services/default-two-factor-api.service.spec.ts
```diff
@@ -579,19 +581,21 @@ describe("TwoFactorApiService", () => {
 
         const result = await twoFactorApiService.putTwoFactorWebAuthn(request);
 
+        const base64UrlPattern = expect.not.stringMatching(/[+/=]/); // no standard base64 chars/padding
```
rr-bw, line 586: According to a callout from Claude, this test may still pass even if given Base64 (not Base64URL):

> This assertion uses a negative regex match: if any of the three characters that distinguish standard Base64 from Base64URL (+, /, =) appear in the output, the test fails. This is a reasonable pattern, but worth noting: the test data bytes ([1,2,3], [4,5,6], [7,8,9]) are small enough that their Base64 output (AQID, BAUG, BwgJ) contains none of those characters regardless of whether standard Base64 or Base64URL is used. So this assertion would pass even if the code were still using Utils.fromBufferToB64. It guards against regressions with other inputs, but doesn't actively prove the current test data is encoded differently.

> Test data containing bytes like [255, 254, 253] (which produce //79 in standard Base64 vs __-9 in Base64URL) would exercise the difference more meaningfully.

libs/common/src/auth/two-factor/services/default-two-factor-api.service.spec.ts
```diff
@@ -579,19 +581,21 @@ describe("TwoFactorApiService", () => {
...
             deviceResponse: expect.objectContaining({
               id: "credential-id",
-              rawId: expect.any(String), // base64 encoded
+              rawId: base64UrlPattern,
               type: "public-key",
               extensions: {},
               response: expect.objectContaining({
-                AttestationObject: expect.any(String), // base64 encoded
-                clientDataJson: expect.any(String), // base64 encoded
+                AttestationObject: base64UrlPattern,
```
rr-bw, line 598:
```suggestion
                attestationObject: base64UrlPattern,
```

rr-bw (dismissed):
Approved with a few minor comments.

rr-bw (approved):

PR #22009, opened 2026-07-23
[PM-39227] Add the browser vault-health reports service (categorize, score, emit)

AlexRubik:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-39227

## 📔 Objective

Adds the browser extension's vault-health orchestration service for the new Health tab — the aggregation layer on top of the existing per-login risk metrics.

`VaultHealthReportService` loads the user's personal-vault logins reactively from `cipherViews$`, composes the Vault-owned SDK-backed `CipherRiskService` (it does not re-implement report logic), and emits a typed `VaultHealthReportView` that:

- categorizes each at-risk login and deduplicates it into a single highest-risk category (Exposed > Weak > Reused), counted once;
- also retains every category each login is at risk in, for views needing cross-category context;
- computes the vault-health score as unique at-risk logins ÷ total logins;
- re-emits when the vault changes, and skips recompute (and its HIBP call) when the scoped login set is unchanged.

Scope is personal-vault Login items with a password (organization, deleted, and passwordless items are excluded). This is the service, its models, and unit tests only — no UI, no persistence, no API, and no SDK changes. The consuming views (PM-35945, PM-35946) and the scan trigger / failure states (PM-39223) build on this service. Risk-computation errors propagate to the caller so the scan story can route to its failure state.

The feature lives in the licensed package (`bitwarden_license/bit-common/src/dirt/vault-health/`) and is registered in the licensed browser build; the `pm-35928-premium-user-health-reports` flag is added (default off) for the consuming stories to gate against.

## 📸 Screenshots

Not applicable — no UI changes (service, models, and unit tests only).

### DIFF
--- a/bitwarden_license/bit-browser/src/popup/app.module.ts
+++ b/bitwarden_license/bit-browser/src/popup/app.module.ts
@@ -13,7 +13,6 @@ import { AppRoutingModule as OssRoutingModule } from "@bitwarden/browser/popup/a
 import { AppModule as OssModule } from "@bitwarden/browser/popup/app.module";
 // import { WildcardRoutingModule } from "@bitwarden/browser/popup/wildcard-routing.module";
 import { CipherRiskService } from "@bitwarden/common/vault/abstractions/cipher-risk.service";
-import { CipherService } from "@bitwarden/common/vault/abstractions/cipher.service";
 import { safeProvider } from "@bitwarden/ui-common";
 
 import { AppRoutingModule } from "./app-routing.module";
@@ -46,7 +45,7 @@ import { AppComponent } from "./app.component";
     safeProvider({
       provide: VaultHealthReportService,
       useClass: DefaultVaultHealthReportService,
-      deps: [CipherService, CipherRiskService],
+      deps: [CipherRiskService],
     }),
   ],
 })
--- a/bitwarden_license/bit-browser/src/popup/app.module.ts
+++ b/bitwarden_license/bit-browser/src/popup/app.module.ts
@@ -8,6 +8,13 @@ import { JslibModule } from "@bitwarden/angular/jslib.module";
 import { AppRoutingModule as OssRoutingModule } from "@bitwarden/browser/popup/app-routing.module";
 import { AppModule as OssModule } from "@bitwarden/browser/popup/app.module";
 // import { WildcardRoutingModule } from "@bitwarden/browser/popup/wildcard-routing.module";
+import {
+  DefaultVaultHealthReportService,
+  VaultHealthReportService,
+} from "@bitwarden/bit-common/dirt/vault-health/services";
+import { CipherRiskService } from "@bitwarden/common/vault/abstractions/cipher-risk.service";
+import { CipherService } from "@bitwarden/common/vault/abstractions/cipher.service";
+import { safeProvider } from "@bitwarden/ui-common";
 
 import { AppRoutingModule } from "./app-routing.module";
 import { AppComponent } from "./app.component";
@@ -35,5 +42,12 @@ import { AppComponent } from "./app.component";
   ],
   declarations: [AppComponent],
   bootstrap: [AppComponent],
+  providers: [
+    safeProvider({
+      provide: VaultHealthReportService,
+      useClass: DefaultVaultHealthReportService,
+      deps: [CipherService, CipherRiskService],
+    }),
+  ],
 })
 export class AppModule {}
--- a/bitwarden_license/bit-browser/src/popup/app.module.ts
+++ b/bitwarden_license/bit-browser/src/popup/app.module.ts
@@ -5,13 +5,13 @@ import { RouterModule } from "@angular/router";
 
 import { JslibModule } from "@bitwarden/angular/jslib.module";
 // import { AppRoutingAnimationsModule } from "@bitwarden/browser/popup/app-routing-animations";
-import { AppRoutingModule as OssRoutingModule } from "@bitwarden/browser/popup/app-routing.module";
-import { AppModule as OssModule } from "@bitwarden/browser/popup/app.module";
-// import { WildcardRoutingModule } from "@bitwarden/browser/popup/wildcard-routing.module";
 import {
   DefaultVaultHealthReportService,
   VaultHealthReportService,
 } from "@bitwarden/bit-common/dirt/vault-health/services";
+import { AppRoutingModule as OssRoutingModule } from "@bitwarden/browser/popup/app-routing.module";
+import { AppModule as OssModule } from "@bitwarden/browser/popup/app.module";
+// import { WildcardRoutingModule } from "@bitwarden/browser/popup/wildcard-routing.module";
 import { CipherRiskService } from "@bitwarden/common/vault/abstractions/cipher-risk.service";
 import { CipherService } from "@bitwarden/common/vault/abstractions/cipher.service";
 import { safeProvider } from "@bitwarden/ui-common";
--- a/bitwarden_license/bit-common/src/dirt/vault-health/models/risk-category.ts
+++ b/bitwarden_license/bit-common/src/dirt/vault-health/models/risk-category.ts
@@ -1,0 +1,22 @@
+/**
+ * The risk categories a login can fall into on the vault-health report.
+ *
+ * Uses a const object rather than a TypeScript enum per ADR-0025.
+ */
+export const RiskCategory = Object.freeze({
+  Exposed: "exposed",
+  Weak: "weak",
+  Reused: "reused",
+} as const);
+
+export type RiskCategory = (typeof RiskCategory)[keyof typeof RiskCategory];
+
+/**
+ * Highest-risk-wins priority order for deduplicating a login into a single
+ * category: Exposed beats Weak beats Reused (PM-39227 / PM-35945).
+ */
+export const RISK_CATEGORY_PRIORITY: readonly RiskCategory[] = [
+  RiskCategory.Exposed,
+  RiskCategory.Weak,
+  RiskCategory.Reused,
+] as const;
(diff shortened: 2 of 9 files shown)


bitwarden_license/bit-browser/src/popup/app.module.ts
lastbestdev, line 1: File looks good itself, but a general question on file locations for files added in this PR. Do we want them under `bitwarden_license/`? 

What is the criteria for placing files here as opposed to keeping them in the `apps/browser/` directory instead?

bitwarden_license/bit-common/src/dirt/vault-health/models/view/vault-health-report.view.ts
```diff
@@ -0,0 +1,34 @@
...
+  /** Unique logins at risk in any category (the score numerator). */
+  atRiskCount = 0;
+  /** atRiskCount / totalCount; 0 when totalCount is 0. */
+  score = 0;
+  /** Deduplicated, highest-risk-wins count per category. */
+  categoryCounts: CategoryRecord<number> = { exposed: 0, weak: 0, reused: 0 };
+  /** The at-risk logins placed in each category (highest-risk-wins). */
+  categoryItems: CategoryRecord<CipherHealthView[]> = { exposed: [], weak: [], reused: [] };
+  /** Full per-login breakdown: every category each at-risk login falls under. */
+  cipherHealth: CipherHealthView[] = [];
```
lastbestdev: Nit: `cipherHealth` list may be unnecessary, since the CipherHealthView objects in the `categoryItems` category lists contain properties for seeing each at risk category a cipher was flagged for. We can consider removing it.

libs/common/src/enums/feature-flag.enum.ts
lastbestdev, line 1: Because this service does not use/hook up the new service anywhere, we can wait to add this flag. It will be added in the extension Health tab implementation where it is also used #22045

bitwarden_license/bit-common/src/dirt/vault-health/models/risk-category.ts
```diff
@@ -0,0 +1,22 @@
...
+
+/**
+ * Highest-risk-wins priority order for deduplicating a login into a single
+ * category: Exposed beats Weak beats Reused (PM-39227 / PM-35945).
+ */
+export const RISK_CATEGORY_PRIORITY: readonly RiskCategory[] = [
+  RiskCategory.Exposed,
+  RiskCategory.Weak,
+  RiskCategory.Reused,
+] as const;
```
lastbestdev: This appears to be unused and can be removed.

bitwarden_license/bit-common/src/dirt/vault-health/services/implementations/default-vault-health-report.service.ts
```diff
@@ -0,0 +1,125 @@
...
+  ) {}
+
+  vaultHealthReport$(userId: UserId): Observable<VaultHealthReportView> {
+    return this.cipherService.cipherViews$(userId).pipe(
+      map((ciphers) => this.filterScopedLogins(ciphers)),
+      // cipherViews$ re-emits on any vault change, including edits to items this
+      // report ignores (cards, org items, deleted items). Only recompute when the
+      // scoped login set actually changes, so we don't fire a fresh HIBP request
+      // for an identical set. Still re-emits whenever the vault changes (the AC).
+      distinctUntilChanged((prev, curr) => this.scopeSignature(prev) === this.scopeSignature(curr)),
```
lastbestdev: I think we could pare down the scope of responsibility for this service a bit by removing the logic for fetching Ciphers. Instead, we would pass the list of Ciphers and UserId in as parameters to this method, and let the Health tab root component pull the ciphers and provide them to this service. Something like this:

```
buildVaultHealthReport$(ciphers: CipherView[], userId: UserId) 
```

Thoughts on this approach?

bitwarden_license/bit-common/src/dirt/vault-health/services/implementations/default-vault-health-report.service.ts
```diff
@@ -0,0 +1,125 @@
...
+  /**
+   * Personal-vault logins with a password: Login type, no organization,
+   * not deleted, non-empty password. A superset of the SDK's own predicate,
+   * so every login passed to the risk service qualifies.
+   */
+  private filterScopedLogins(ciphers: CipherView[] | null): CipherView[] {
+    return (ciphers ?? []).filter(
+      (c) =>
+        c.type === CipherType.Login &&
+        c.organizationId == null &&
```
lastbestdev, line 41: I think we will want to clarify if Organization owned Ciphers should be included in these reports. I am under the impression they should be.

bitwarden_license/bit-common/src/dirt/vault-health/services/implementations/default-vault-health-report.service.ts
lastbestdev: For comparison of strings made from cipher data here to work, we are relying on `cipherViews$` emitting the list in the same order each time. Are we confident this would work as expected?

bitwarden_license/bit-common/src/dirt/vault-health/models/view/vault-health-report.view.ts
```diff
@@ -0,0 +1,34 @@
...
+ */
+export class VaultHealthReportView implements View {
+  /** Personal-vault logins with a password (the score denominator). */
+  totalCount = 0;
+  /** Unique logins at risk in any category (the score numerator). */
+  atRiskCount = 0;
+  /** atRiskCount / totalCount; 0 when totalCount is 0. */
+  score = 0;
+  /** Deduplicated, highest-risk-wins count per category. */
+  categoryCounts: CategoryRecord<number> = { exposed: 0, weak: 0, reused: 0 };
```
lastbestdev: Nit: similarly, the counts maintained here could be removed since we can read `.length` from each of the category lists in `categoryItems`.

lastbestdev (commented):
Looking good. A few comments I'm curious to hear your thoughts on/things we can clarify with team

libs/common/src/enums/feature-flag.enum.ts
AlexRubik: https://github.com/bitwarden/clients/pull/22009/commits/965c63a38420d577521f2d6f80cf22ad86e376cb

bitwarden_license/bit-common/src/dirt/vault-health/models/risk-category.ts
AlexRubik: https://github.com/bitwarden/clients/pull/22009/commits/a9ac372a8bf0a06f8c39172d28192817d6b58b51

bitwarden_license/bit-common/src/dirt/vault-health/services/implementations/default-vault-health-report.service.ts
AlexRubik: Good catch. Sorted the signature tuples by id so emission order can't cause spurious recomputes: [91acec7f](https://github.com/bitwarden/clients/pull/22009/commits/91acec7fc636f197859751f7c5982e228d738b6f). If we later move to passing the cipher list into the service as a parameter, this signature/dedup code goes away entirely.

bitwarden_license/bit-common/src/dirt/vault-health/models/view/vault-health-report.view.ts
AlexRubik: Agreed, categoryCounts duplicates categoryItems[cat].length, so I dropped it: [0ed23004](https://github.com/bitwarden/clients/pull/22009/commits/0ed230046c55fdc1089819f3f7c548089a5f1d9a). Kept atRiskCount since the AC calls it out.

bitwarden_license/bit-common/src/dirt/vault-health/services/implementations/default-vault-health-report.service.ts
AlexRubik: The jira ticket currently says this: "Scope: personal-vault logins with a password; exclude organization items..."

Has this changed since ticket creation?

bitwarden_license/bit-browser/src/popup/app.module.ts
AlexRubik: This location was decided by Tom + Leslie. I'm not sure what the criteria is.
From claude: "Criteria: enterprise/premium and Access-Intelligence-adjacent code goes under bitwarden_license/; basic/OSS reports stay in apps/browser/libs/common"

bitwarden_license/bit-common/src/dirt/vault-health/services/implementations/default-vault-health-report.service.ts
AlexRubik: Sounds good to me. We may need a small clarification in [PM-39223](https://bitwarden.atlassian.net/browse/PM-39223) if we go this route.

bitwarden_license/bit-common/src/dirt/vault-health/services/implementations/default-vault-health-report.service.ts
lastbestdev: You are correct, the ticket spec says it should be personal items only. Disregard 👍🏻

bitwarden_license/bit-browser/src/popup/app.module.ts
lastbestdev: Got it. I may be missing something but since this feature is not an enterprise feature, it would make sense to me for it to be kept in `apps/browser/`. 

@Banrion can we get your input here?

AlexRubik: Team discussion clarified that we are viewing health tab as paid service so it goes in license dir.

lastbestdev (approved):
LGTM, thanks for making those changes! 🚀

PR #22011, opened 2026-07-23
[PM-40679] Autotype MVP Naming Refactor

coltonhurst:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-40679

## 📔 Objective

Refactoring the Autotype MVP code so that it's very clear which code belongs to the MVP, and which code belongs to GA. _**There should be no functionality changes.**_

An upcoming PR (PM-40679 part 2) will add the GA code, which is based on the MVP code. These are separated to reduce complexity for reviews.

**Original Plan**

- Leave the front end components in the render process as is (not renaming them). Feature flag logic will have to be used to change how certain parts are displayed. There may be some small changes (but no large refactors) required for these components based on the changes listed in the bullet points following this one.
- Refactor the IPC channels, appending the following text: `“Mvp”` (or something similar to this)
- Refactor the `MainDesktopAutotypeService` (in the main process). Ideally we would create a different service, so MVP and GA are separated. However, when this service is initialized (in main.ts), we don’t currently have easy access to feature flag values. This PR has two separate functions within this service that handle registering the appropriate IPC listeners (MVP or GA) after getting the feature flag values from the render process (this is a new IPC channel).
- Refactor the `desktop_native` autotype crate rust code (which will end up including napi exports, etc) so that MVP code is in its own MVP module. The main process then calls these appropriately based on the details mentioned in the above bullet point.

## ✨ Key Changes

- **Rust (`desktop_native/autotype`)**: moved the existing windowing/typing implementation into a new
  `mvp` module (`autotype::mvp::{get_foreground_window_title, type_input}`), leaving `lib.rs` as a thin
  shell that just re-exports `app_data` helpers and the `mvp` module.
- **NAPI bindings**: renamed the exposed `autotype` namespace to `autotype_mvp` (`autotype.rs` →
  `autotype_mvp.rs`), updating `index.d.ts` accordingly.
- **IPC channels** (`apps/desktop/src/autofill/models/ipc-channels.ts`): split `AUTOTYPE_IPC_CHANNELS`
  into a small shared `AUTOTYPE_IPC_CHANNELS` (currently just `STATE`, used to report which
  implementation is active to the main process) and an MVP-only `AUTOTYPE_MVP_IPC_CHANNELS`
  (`TOGGLE`/`CONFIGURE`/`LISTEN`/`EXECUTION_ERROR`/`EXECUTE`).
- **New `AutotypeState` enum** (`disabled` | `mvp` | `ga`) is now computed centrally in
  `DesktopAutotypeService.autotypeState$` from the user setting, the `WindowsDesktopAutotype` (MVP) and
  `WindowsDesktopAutotypeGA` (GA) feature flags, unlock status, and Premium status. MVP takes precedence
  over GA when both flags are on. The resolved state is reported to the main process over the new
  `STATE` IPC channel so `MainDesktopAutotypeService` knows which implementation's shortcut behavior to
  run.
- **Settings UI** (`settings.component.ts`, `settings-dialog.component.ts`): the "show autotype toggle"
  check now uses a new `autotypeMvpOrGaEnabled$` helper instead of checking the MVP flag directly, so
  the setting becomes visible once either the MVP or GA flag is on.
- **`DesktopAutotypeDefaultSettingPolicy`**: also switched to `autotypeMvpOrGaEnabled$` so the org
  default-enable policy applies regardless of which implementation is flagged on.
- **`preload.ts`**: added `reportAutotypeState`, and renamed the MVP-specific bridge methods
  (`configureAutotypeMvp`, `toggleAutotypeMvp`, `listenAutotypeRequestMvp`,
  `stopListeningAutotypeRequestMvp`). `listenAutotypeRequestMvp` registration is now idempotent — it
  unbinds any previously-registered listener before adding a new one — so repeated enable/disable
  transitions can't stack duplicate IPC listeners.
- **`MainDesktopAutotypeService`**: now listens for `STATE` reports and only runs MVP's foreground-window
  lookup when the reported state is `Mvp`; logs a placeholder message for `Ga` until GA is implemented.

## 🧪 Testing

- Full QA testing on Autotype MVP is required to prove there are no changes in the MVP functionality
- I've tested the following so far, with more tests incoming:
    - Autotype is correctly off and disabled in prod (when both MVP and GA FF's are false)
    - Autotype MVP correctly works in QA (when the MVP flag is on and GA is off)
    - Setting the Autotype shortcut still works as it has
    - Going from MVP where Autotype is enabled, logging out, switching to Prod, and logging back in - Autotype is correctly disabled


dani-garcia (dismissed):
Platform changes lgtm

apps/desktop/src/autofill/services/desktop-autotype-policy.service.ts
```diff
@@ -8,9 +8,10 @@ import { AccountService } from "@bitwarden/common/auth/abstractions/account.serv
 import { AuthService } from "@bitwarden/common/auth/abstractions/auth.service";
 import { AuthenticationStatus } from "@bitwarden/common/auth/enums/authentication-status";
 import { getUserId } from "@bitwarden/common/auth/services/account.service";
-import { FeatureFlag } from "@bitwarden/common/enums/feature-flag.enum";
 import { ConfigService } from "@bitwarden/common/platform/abstractions/config/config.service";
 
+import { autotypeMvpOrGaEnabled$ } from "./desktop-autotype.service";
```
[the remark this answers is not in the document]
coltonhurst: Resolved in bc8d3ac73e9736b425371112f67d444ad86d99e3

apps/desktop/src/autofill/services/autotype-feature-flags.ts
```diff
@@ -0,0 +1,35 @@
...
+//
+// `autotypeState$` and `autotypeMvpOrGaEnabled$` both consume this.
+export function autotypeFeatureFlags$(
+  configService: ConfigService,
+): Observable<[boolean, boolean]> {
+  return combineLatest([
+    configService.getFeatureFlag$(FeatureFlag.WindowsDesktopAutotype), // mvp
+    configService.getFeatureFlag$(FeatureFlag.WindowsDesktopAutotypeGA), // ga
+  ]);
+}
```
neuronull, line 17: It looks like this PR is conflating the MVP refactor with also introducing new logic to do stuff around the GA feature flag.

Would it be possible to separate out the GA related stuff to a separate PR?

It seems like adding logic around the GA shouldn't be necessary for this task.

apps/desktop/src/app/accounts/settings.component.spec.ts
```diff
@@ -906,5 +907,20 @@ describe("SettingsComponent", () => {
       // `showEnableAutotype` should be false
       expect(component.showEnableAutotype).toBe(false);
     });
+
+    it("shows the autotype toggle on windows when only the GA flag is enabled (MVP off)", async () => {
```
neuronull, line 911: Apart from the other comment about separating out GA related additions to a new PR, 

Is `app/accounts` the correct place for Autotype specific unit test ? 🤔 Seems like the autotype unit tests would be better suited to be contained in DN team owned files.

neuronull (changes requested):

apps/desktop/src/app/accounts/settings.component.spec.ts
coltonhurst: This is actually adding a test specific to the settings component, which lives in `app/accounts`. It tests that if the GA flag for Autotype is on, the setting to enable Autotype still appears.

neuronull: my concern was about codeownership aspect of it, but I just checked the file and see that a ton of feature specific logic is tested (owned by disparate teams) so my concern is nullified.

apps/desktop/src/autofill/services/autotype-feature-flags.ts
neuronull: We discussed and decided to proceed with the one pr approach

coltonhurst:
Closing this PR, more incoming to replace this one. Going to change the plan so that MVP and GA also have different render and main process services. Because I was attempting to use the same services, there were multiple race conditions I had to keep adjusting for, which continued to increase complexity.

PR #22012, opened 2026-07-23
only serve desktop bundle through bitwarden-desktop-file://

addisonbeck:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-40845

## 📔 Objective

In #19208 we started using a custom url protocol to load the desktop app bundle. We left some backwards compatibility logic in when switching because of independent server requirements. See bitwarden/server#7080 for more info on these.

All non-compliant versions have since rolled out of support, so we can cut this condition now. This commit does that.


addisonbeck:
@dani-garcia I think this captures everything we needed to do here, but let me know if I missed anything. Also, when we talked you mentioned a specific environment setup for QA for this. Would you mind adding that infonto the QA notes on the ticket?

dani-garcia (approved):

PR #22015, opened 2026-07-23
[PM-40693] Hide copy button on file Sends

mcamirault:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-40693

## 📔 Objective

This PR fixes a bug where the "Make a Copy" button was incorrectly displayed for File Sends, which can't be copied. It also fixes two unrelated bugs, one that would incorrectly show the "Send is Disabled" banner after the "Make a Copy" button was pressed on the browser extension, and one that would prevent the "Make a Copy" button from properly re-initializing the Send edit drawer on web.

### DIFF
--- a/apps/browser/src/tools/popup/send-v2/add-edit/send-add-edit.component.ts
+++ b/apps/browser/src/tools/popup/send-v2/add-edit/send-add-edit.component.ts
@@ -235,9 +235,6 @@ export class SendAddEditComponent {
       const sendDisabledReason = await this.sendPolicyService.sendDisabledReason(
         this.config.originalSend,
       );
-      const showMakeCopyButton =
-        sendDisabledReason !== SendDisabledReason.None &&
-        this.config.originalSend.type === SendType.Text;
       if (sendDisabledReason === SendDisabledReason.RestrictedType) {
         this.disabledSendConfig.set({
           title:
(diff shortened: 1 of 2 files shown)


harr1424 (dismissed):
It took me a while to grasp these code changes, but mostly because the PR title led me to believe we were hiding the copy button on disabled file Sends only. I hope it's OK I've updated the title to better reflect we are disabling the copy button on file Sends all-together, since they can't be copied (this was conveyed in the PR description, my fault for not reading it). 

❓ Would it make sense to add test coverage? Or would this only be testing the mock capabilities of Jest?

harr1424 (dismissed):

apps/browser/src/tools/popup/send-v2/add-edit/send-add-edit.component.ts
```diff
@@ -235,24 +235,29 @@ export class SendAddEditComponent {
       const sendDisabledReason = await this.sendPolicyService.sendDisabledReason(
         this.config.originalSend,
       );
+      // We can make a copy of a disabled Send only if two conditions are met
+      // 1. The Send doesn't violate the SendType restriction of the policy (if Text
+      // Sends aren't allowed we'd have to turn the copy into a File Send)
```
harr1424: >(if Text Sends aren't allowed we'd have to turn the copy into a File Send)

This comment almost implies we might do this, I would re-word it to state more plainly that if text Sends are disallowed by policy we just won't allow a copy operation. 

If you agree, I would also update the comment in the other file.

harr1424 (dismissed):

harr1424 (approved):

PR #22032, opened 2026-07-24
chore(kdf-change): port change kdf to sdk

quexten:
Drops a lot of TS code. We can drop even more once key connector is fully in sdk, and we can drop the master-key compatibility logic.


libs/common/src/key-management/utils.ts
```diff
@@ -40,6 +43,29 @@ export async function withPasswordManagerSdk<TResult>(
...
+/**
+ * Keeps the legacy locally-cached master key and master-key-wrapped user key in sync with the
+ * persisted master-password unlock data, so that master-key based unlock verification etc. keeps
+ * working after the SDK re-derives them (e.g. on a KDF change). The SDK has already written the new
+ * unlock data to state, so we read it back to derive the master key.
+ *
+ * TODO: Drop this helper and all of its callers once key connector runs via the SDK, at which point
+ * ownership of this state moves into the SDK and it no longer needs to be mirrored client-side.
+ */
+export async function syncLegacyMasterKeyState(
```
mzieniukbw: This should be part of `InternalMasterPasswordServiceAbstraction`, not in utils.

mzieniukbw (changes requested):

libs/common/src/key-management/utils.ts
quexten: I guess. We will drop it soon either way soon after rolling out sdk-based key-connector migration.

quexten: Moved.

mzieniukbw (approved):

PR #22065, opened 2026-07-27
feat(data-recovery): add detection of field decryption ciphers

quexten:
Add field-level-cipher-corruption detection. Namely, we probe a cipher by trying to decrypt the "name" field. Please note, that the name field *may* be absent for v2 ciphers that use blob encryption. This recovery step is skipped, exactly if the user-key is a v2 key.

https://bitwarden.atlassian.net/browse/PM-40990

Mainly, we want to detect the "passkey corrupted" vault items, all of which have no cipher-key. We want the information to be present whether there is a passkey, and whether there is a cipher-key in the logs.

### DIFF
--- a/apps/web/src/app/key-management/data-recovery/steps/cipher-step.ts
+++ b/apps/web/src/app/key-management/data-recovery/steps/cipher-step.ts
@@ -1,10 +1,22 @@
 import { ApiService } from "@bitwarden/common/abstractions/api.service";
+import { EncryptService } from "@bitwarden/common/key-management/crypto/abstractions/encrypt.service";
+import { EncryptionType } from "@bitwarden/common/platform/enums";
+import { SymmetricCryptoKey } from "@bitwarden/common/platform/models/domain/symmetric-crypto-key";
+import { UserKey } from "@bitwarden/common/types/key";
 import { CipherEncryptionService } from "@bitwarden/common/vault/abstractions/cipher-encryption.service";
+import { Cipher } from "@bitwarden/common/vault/models/domain/cipher";
 import { DialogService } from "@bitwarden/components";
 
 import { LogRecorder } from "../log-recorder";
 
 import { RecoveryStep, RecoveryWorkingData } from "./recovery-step";
+
+/** Prefix of an AesCbc256_HmacSha256_B64 encrypted string. */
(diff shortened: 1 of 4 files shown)


apps/web/src/app/key-management/data-recovery/steps/cipher-step.ts
```diff
@@ -31,20 +44,75 @@ export class CipherStep implements RecoveryStep {
...
   }
 
+  /**
+   * Probes the cipher's name field for field-level corruption. Returns a failure reason, or null
+   * when the name decrypts or is not a form we probe.
+   *
+   * Null means that the check did not fail. A string value means the check failed.
+   */
+  private async checkNameDecrypts(cipher: Cipher, userKey: UserKey | null): Promise<string | null> {
+    const encryptedName = cipher.name?.encryptedString;
```
quexten, line 88: As a follow-up we will make a better API on the SDK side, but for detecting this issue, we will just go via TS.

apps/web/src/app/key-management/data-recovery/steps/cipher-step.ts
```diff
@@ -31,20 +44,75 @@ export class CipherStep implements RecoveryStep {
...
         this.decryptableCipherIds.push(cipher.id);
-      } catch {
-        logger.record(`Cipher ID ${cipher.id} was undecryptable`);
-        this.undecryptableCipherIds.push(cipher.id);
+        continue;
       }
+
+      const cipherKeyState = cipher.key != null ? "present" : "absent";
+      const itemType = cipher.type != null ? cipher.type : "unknown";
+      const fido2CredentialsPresent = cipher.login?.fido2Credentials != null ? "present" : "absent";
```
mzieniukbw, line 69: ❓ should we also log attachments presence ?

mzieniukbw (approved):

apps/web/src/app/key-management/data-recovery/steps/cipher-step.ts
quexten: I made a follow up ticket that tracks attachment decrypt errors in a separate step here: https://bitwarden.atlassian.net/browse/PM-41011

PR #22068, opened 2026-07-27
PM-39981 browser extension share code

bmbitwarden:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-39981

## 📔 Objective

Create the Share panel screen displayed in the browser extension when a user clicks "Share via link" on a vault item

## 📸 Screenshots

Screenshot 2026-07-27 at 8 09 23 AM


apps/browser/src/tools/popup/share/share-item.component.ts
```diff
@@ -0,0 +1,328 @@
...
+    if (nonEmpty.length === 0) {
+      return { required: true };
+    }
+    const emailRegex = /^[^\s@]+@[^\s@]+\.[^\s@]+$/;
+    const invalid = nonEmpty.filter((e) => !emailRegex.test(e));
+    if (invalid.length > 0) {
+      return { multipleEmails: true };
+    }
+    return null;
+  };
```
harr1424, line 180: ❓ This seems to mostly duplicate similar logic located in the [send-details-componet](https://github.com/bitwarden/clients/blob/2ce471bfde972155e23004819c7d10100d129b12/libs/tools/send/send-ui/src/send-form/components/send-details/send-details.component.ts#L391), but doesn't enforce any check against allowed domains configured as part of a relevant Send policy. 

Should the domain check be occurring here? If so, would it make sense to create a shared utility to use in both locations?

apps/browser/src/tools/popup/share/share-item.component.ts
```diff
@@ -0,0 +1,328 @@
...
+  ): ValidationErrors | null => {
+    if (typeof control.value !== "string" || control.value.length < 2500) {
+      return null;
+    }
+    return {
+      emailsMaxLength: {
+        message: this.i18nService.t("sendEmailsCharacterLimitReached"),
+      },
+    };
+  };
```
harr1424, line 193: ❓ This looks like an exact copy of [this](https://github.com/bitwarden/clients/blob/2ce471bfde972155e23004819c7d10100d129b12/libs/tools/send/send-ui/src/send-form/components/send-details/send-details.component.ts#L427), should a shared utility be extracted?

apps/browser/src/tools/popup/share/share-link.service.ts
```diff
@@ -0,0 +1,95 @@
...
+  emails: string[];
+  expiresAt: Date;
+  oneTimeShare: boolean;
+  url: string;
+}
+
+/**
+ * Stub service for managing share links.
+ * Uses in-memory state until the backend API is available.
+ */
```
harr1424, line 21: ❓ How are we going to send this for QA if the backend API isn't available?

apps/browser/src/tools/popup/share/share-link.service.ts
```diff
@@ -0,0 +1,95 @@
...
+@Injectable({ providedIn: "root" })
+export class ShareLinkService {
+  private nextId = 3;
```
harr1424, line 24: ❓ Where does the number 3 come from here?  What will it be used for?

apps/browser/src/tools/popup/share/share-link.service.ts
```diff
@@ -0,0 +1,95 @@
...
+    },
+    {
+      id: "share-link-1",
+      cipherId: "mock-cipher-id" as CipherId,
+      emails: ["sauron@mordor.com"],
+      expiresAt: new Date("2026-05-26"),
+      oneTimeShare: true,
+      url: "https://vault.bitwarden.com/share/share-link-1",
+    },
+  ]);
```
harr1424, line 42: ⚠️ I am going to stop reviewing at this point, we should probably have an offline conversation about how a service class is hardcoding values like this. I wouldn't expect to see the word mock outside of a spec file.

harr1424 (changes requested):
I don't feel comfortable approving this PR on the basis that it uses mocks in a service class as a way to avoid connecting to our actual back-end implementation. Would it make more sense to target the actual back-end once it hits `main` and perform a single review/QA cycle instead?

PR #22071, opened 2026-07-27
Fix AuthEdit SDK break

adudek-bw:
## 📔 Objective

Fixes a build break introduced by bumping `@bitwarden/sdk-internal` (and `@bitwarden/commercial-sdk-internal`, aliased in for `bit-*` clients) to `0.2.0-main.919`. That SDK version changes `SendEditRequest.auth` from a bare `SendAuthType` to a new `AuthEdit` union (`{ type: "preserve" }` or `{ type: "set"; auth: SendAuthType }`), so `SendSdkApiService.buildSendEditRequest` no longer type-checks.


libs/common/src/tools/send/services/send-sdk-api.service.ts
```diff
@@ -74,16 +76,7 @@ export class SendSdkApiService implements SendApiServiceAbstraction {
...
-    // On a password-preserving edit the caller passes no plaintext, and the freshly-encrypted
-    // `send` carries no `keyB64` (SendService.encrypt only derives one when given a plaintext).
-    // The existing proof-of-knowledge lives on the stored send, so recover it from state to
-    // forward verbatim. Only needed for a password-protected edit: absent on create (no existing
-    // send) and unused for non-password sends or when a plaintext is given.
-    const existingPassword =
-      send.id != null && send.authType === AuthType.Password && plaintextPassword == null
-        ? (await this.sendService.getFromState(send.id))?.password
-        : undefined;
-    const sdkView = await this.mutateSend(sendView, userId, plaintextPassword, existingPassword);
```
harr1424, line 86: ❓ Was this removed because the bumped version instead already wraps the existing password? (Using `preserve` ?)

harr1424 (approved):

libs/common/src/tools/send/services/send-sdk-api.service.ts
adudek-bw: Yup!

PR #22080, opened 2026-07-27
[PM-35791] Remove old Send access endpoints

mcamirault:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-35791

## 📔 Objective

This PR removes the two endpoints that used to be used for Send access, removes the V2 suffix we were using on a couple functions to distinguish the new flow from the old, and removes some leftover fields in the Send auth component from the old access flow. There are no changes to functionality, this is pure cleanup.

## 📸 Screenshots

No UI changes

### DIFF
--- a/apps/web/src/app/tools/send/send-access/access.component.ts
+++ b/apps/web/src/app/tools/send/send-access/access.component.ts
@@ -7,8 +7,6 @@ import { takeUntilDestroyed } from "@angular/core/rxjs-interop";
 import { ActivatedRoute } from "@angular/router";
 
 import { SendAccessToken } from "@bitwarden/common/auth/send-access";
-import { SendAccessRequest } from "@bitwarden/common/tools/send/models/request/send-access.request";
-import { SendAccessResponse } from "@bitwarden/common/tools/send/models/response/send-access.response";
 
 import { SharedModule } from "../../../shared";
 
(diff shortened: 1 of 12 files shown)


apps/web/src/app/tools/send/send-access/access.component.ts
```diff
@@ -49,9 +45,7 @@ export class AccessComponent implements OnInit {
       // when pasting sequential Send URLs into the browser,
       // Angular reuses the SendAuthComponent instance
       // Reset state so child components are recreated with fresh data
-      this.sendAccessResponse = null;
       this.sendAccessToken = null;
-      this.sendAccessRequest = new SendAccessRequest();
```
harr1424, line 54: ❓  Do the comments relate to the deleted code? 

Said another way, can we delete those lines and not worry about the former defect where pasting a new Send URL and hitting enter would continue to show the old Send?

mcamirault: `sendAccessToken` is the only field of the three that is passed to a child, the others are simply leftover from the old auth flow. Since we're leaving the line that resets `sendAccessToken` that former defect shouldn't reappear.

harr1424 (approved):

PR #22102, opened 2026-07-28
[CL-1245] updated icon tile colors

BryanCunningham:
## 🎟️ Tracking
https://bitwarden.atlassian.net/browse/CL-1245

## 📔 Objective
- Adds new icon tile non-semantic variants
- Adds new icon tile 'bold' variants and colors
- Adds ability to pass a custom color to icon tile

## 📸 Screenshots
image

image


willmartian (dismissed):
1. I don't like how subtle is both a variant value and emphasis value. Worth changing? :P 


2. `icon-tile.component.ts` and the Tailwind config felt kind of repetitive to me, so I asked Claude to investigate. It recommended deriving the styles with CSS variables, which gets around Tailwind's limitation of not supporting dynamic class names. 

(I'm okay with the current approach too, this is just more for your consideration.)

Like so: 

# 🤖 🤖 🤖 

## icon-tile.component.ts

```ts
type SemanticVariant = "primary" | "success" | "danger" | "warning" | "subtle" | "dark";
type DecorativeVariant = "brand" | "teal" | "green" | "orange" | "red" | "purple" | "gray";

export type IconTileVariant = SemanticVariant | DecorativeVariant;
export type IconTileEmphasis = "subtle" | "bold";

/** The background, border, and foreground colors for one tile appearance. */
type TileColors = { background: string; border: string; text: string };

/** Decorative tokens are formulaic, so the triple is derived rather than enumerated. */
const decorativeColors = (family: DecorativeVariant, emphasis: IconTileEmphasis): TileColors => {
  const suffix = emphasis === "bold" ? "-bold" : "";
  return {
    background: `var(--color-bg-decorative-${family}${suffix})`,
    border: `var(--color-border-decorative-${family}${suffix})`,
    text: `var(--color-fg-decorative-${family}${suffix})`,
  };
};

// Semantic variants that render identically to a decorative family.
const decorativeAliases: Partial<Record<IconTileVariant, DecorativeVariant>> = {
  primary: "brand",
  success: "green",
  danger: "red",
  warning: "orange",
};

// Semantic variants with no decorative equivalent; unaffected by emphasis.
const uniqueVariantColors: Partial<Record<IconTileVariant, TileColors>> = {
  subtle: {
    background: "var(--color-bg-quaternary)",
    border: "var(--color-border-base)",
    text: "var(--color-fg-body)",
  },
  dark: {
    background: "var(--color-bg-contrast)",
    border: "var(--color-border-strong)",
    text: "var(--color-fg-contrast)",
  },
};
```

```ts
protected readonly colorStyles = computed<TileColors>(() => {
  const custom = this.color()?.trim();

  if (custom) {
    // "black" or "white" — svgTextFill omits `!important` so the value is valid in a style binding.
    const text = Utils.pickTextColorBasedOnBgColor(custom, 135, true);
    const lightness = text === "black" ? "calc(l - 15)" : "calc(l + 15)";
    return { background: custom, border: `hsl(from ${custom} h s ${lightness})`, text };
  }

  const variant = this.variant();
  return (
    uniqueVariantColors[variant] ??
    decorativeColors(decorativeAliases[variant] ?? (variant as DecorativeVariant), this.emphasis())
  );
});

protected readonly containerClasses = computed(() => [
  "tw-inline-flex",
  "tw-items-center",
  "tw-justify-center",
  "tw-flex-shrink-0",
  "tw-border",
  ...sizeStyles[this.size()].container,
  ...borderRadius[this.size()],
]);
```

## icon-tile.component.html

```html
<div
  [class]="containerClasses()"
  [style.background-color]="colorStyles().background"
  [style.border-color]="colorStyles().border"
  [style.color]="colorStyles().text"
  [attr.aria-label]="ariaLabel()"
  [attr.role]="ariaLabel() ? 'img' : null"
></div>
```

libs/components/src/icon-tile/icon-tile.component.ts
```diff
@@ -85,17 +214,47 @@ export class IconTileComponent {
...
+    background: string;
+    border: string;
+    text: string;
+  } | null>(() => {
+    const color = this.color();
+    if (color == null || color.trim() === "") {
+      return null;
+    }
+
+    // "black" or "white" — svgTextFill omits `!important` so the value is valid in a style binding.
```
vleague2: Is the `svgTextFill` reference a holdover from the avatar component...?

vleague2 (commented):
I'm going to pull in latest so that Chromatic will run (I think Github actions had issues yesterday)

libs/components/src/icon-tile/icon-tile.stories.ts
```diff
@@ -95,10 +94,6 @@ export const AllVariants: Story = {
               <bit-icon-tile icon="bwi-clock" variant="warning"></bit-icon-tile>
               <span class="tw-text-sm tw-text-muted">Warning</span>
             </div>
-            <div class="tw-flex tw-flex-col tw-items-center tw-gap-2">
-              <bit-icon-tile icon="bwi-clock" variant="subtle"></bit-icon-tile>
-              <span class="tw-text-sm tw-text-muted">Subtle</span>
-            </div>
```
vleague2: Should this be replaced by the gray color now?

vleague2: Nvm I see that there is no semantic equivalent now

vleague2 (approved):

PR #22109, opened 2026-07-28
desktop-autofill: Implement Windows plugin credential sync [PM-41025]

iinuwa:
## 🎟️ Tracking

[PM-41025](https://bitwarden.atlassian.net/browse/PM-41025)

## 📔 Objective

Synchronizes credentials from Bitwarden to Windows autofill store.
- Adds support to query plugin enabled/disabled state via webauthn.dll
- Moves functions to read Appx resources from windows_plugin_authenticator to `desktop_core::autofill`, since they are shared by both `desktop_core` and 
- Implements the native side of the status request from DesktopAutofillService, informing whether the credentials should sync or not
- Implements the conversion from Bitwarden to Windows native credentials

This enables passkey assertion with conditional mediation and from the Windows Hello autofill selection dialog, as shown in the demo below.

Note this work is still behind a feature flag; the UI glitches in the demo below will be addressed in future PRs.

## 📸 Screenshots

https://github.com/user-attachments/assets/2c8352cc-0850-4dbd-8a80-570068bd8339

[PM-41025]: https://bitwarden.atlassian.net/browse/PM-41025?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/desktop/desktop_native/core/src/autofill/windows/sync.rs
+++ b/apps/desktop/desktop_native/core/src/autofill/windows/sync.rs
@@ -1,0 +1,405 @@
+use anyhow::{anyhow, Context, Result};
+use serde::{Deserialize, Serialize};
+use serde_with::{
+    base64::{Base64, UrlSafe},
+    formats::Unpadded,
+    serde_as,
+};
+use win_webauthn::{
+    plugin::{Clsid, PluginCredentialDetails, WebAuthnPlugin},
+    CredentialId, UserId,
+};
+
+use super::{get_clsid, CommandResponse};
+
+pub(super) fn handle_sync_request(params: SyncParameters) -> Result<SyncResponse> {
+    let Some(clsid) = get_clsid()? else {
+        return Err(anyhow!(
+            "[core::autofill::sync] Sync requested, but CLSID not found"
+        ));
+    };
+
+    let win_credentials = to_plugin_credentials(params.credentials)?;
+    let added = sync_credentials_to_windows(win_credentials, clsid)?;
+    Ok(SyncResponse { added })
+}
+
+/// Drops credential types Windows cannot store, converts the rest, and skips any
+/// individual credential Windows would reject rather than failing the whole batch.
+fn to_plugin_credentials(credentials: Vec<SyncCredential>) -> Result<Vec<PluginCredentialDetails>> {
+    let win_credentials =
+        credentials
+            .into_iter()
+            .enumerate()
+            .filter_map(|(i, c)| match c {
+                SyncCredential::Fido2(cred) => Some((i, cred)),
+                SyncCredential::Password => None,
+            })
+            .filter_map(|(i, cred)| {
+                PluginCredentialDetails::try_from(cred)
+                    .map_err(|err| {
+                        tracing::warn!(
+                            "[core::autofill] Skipping sync of credential {i} due to error: {err:#}"
+                        );
+                    })
+                    .ok()
+            })
+            // Windows array lengths are represented with u32, so we can't submit more than that.
+            .take(u32::MAX.try_into().context(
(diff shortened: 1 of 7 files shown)


apps/desktop/desktop_native/core/src/autofill/windows/sync.rs
```diff
@@ -0,0 +1,401 @@
...
+}
+
+#[serde_as]
+#[derive(Debug, Deserialize)]
+#[serde(rename_all = "camelCase")]
+struct SyncFido2Credential {
+    rp_id: String,
+
+    #[serde_as(as = "Base64<UrlSafe, Unpadded>")]
+    credential_id: Vec<u8>,
```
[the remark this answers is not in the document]
iinuwa: Not addressing; if this happens, something else has gone horribly wrong, and it needs more attention elsewhere.

apps/desktop/desktop_native/core/src/autofill/windows/sync.rs
```diff
@@ -0,0 +1,405 @@
...
+                    .map_err(|err| {
+                        tracing::warn!(
+                            "[core::autofill] Skipping sync of credential {i} due to error: {err:#}"
+                        );
+                    })
+                    .ok()
+            })
+            // Windows array lengths are represented with u32, so we can't submit more than that.
+            .take(u32::MAX.try_into().context(
+                "Windows binary unexpectedly running on an architecture where usize < u32",
```
abergs, line 49: How did you find this edge case?

apps/desktop/desktop_native/win_webauthn/src/api/plugin/mod.rs
```diff
@@ -406,6 +406,15 @@ impl WebAuthnPlugin {
...
         };
 
+        // With nothing to add, removing is the whole operation - signing out syncs an empty
+        // list to clear the store - so a failure here cannot be reported as a successful sync.
+        if credentials.is_empty() {
+            tracing::debug!("No credentials to add to Windows, removing all existing ones...");
+            remove_all_credentials(self.clsid)?;
+            tracing::debug!("Successfully removed existing credentials - sync completed");
+            return Ok(());
+        }
```
abergs, line 416: This is ok for now, but I think this wont't work great for multi account setups.

E.g.
account1 syncs 5 credentials
User switches account to account2, which has no passkeys. Sends empty sync.

Windows now has no idea there are passkeys for account1 until the user goes back and switches account in bw.

abergs (approved):

apps/desktop/desktop_native/core/src/autofill/windows/sync.rs
iinuwa: It's part of the function signature: Windows can only take a u32. We're probably constrained by memory on the machine before we hit this limit though. :)

dani-garcia (approved):

apps/desktop/desktop_native/win_webauthn/src/api/plugin/mod.rs
iinuwa: Yeah, we'll need to come to a decision about that. IIRC, on our other clients, passkey autofill entries are only populated for the active account, so we're following the same pattern here.

On Windows, the OS checks the lock status of the vault, and will show a generic Bitwarden entry so that the user could unlock the vault. macOS will also show the "More from Bitwarden..." button to open the vault.

Created [PM-41290](https://bitwarden.atlassian.net/browse/PM-41290) to follow up later.

abergs: Great, thank you.

PR #22119, opened 2026-07-28
PM-40304: Add the conditional Vaults nav-section logic and vault-type color mapping

nikwithak:
## 🎟️ Tracking
https://bitwarden.atlassian.net/browse/PM-40304

## 📔 Objective
Introduce the Vault-owned service that derives the conditional Vaults-section navigation state from existing organization, policy, and account-plan data, plus the vault-type to icon-tile color mapping. Shared foundation both the web and desktop side navs render from.


vleague2 (commented):
Thoughts on putting that `avatar-color` utility into the `libs/components/src/avatar` folder? Is there any particular reason to move it out to common/platform?

gbubemismith:
> Thoughts on putting that `avatar-color` utility into the `libs/components/src/avatar` folder? Is there any particular reason to move it out to common/platform?

I completely agree with this. Nice catch! Thanks @vleague2

libs/vault/src/services/default-vault-nav.service.ts
```diff
@@ -0,0 +1,127 @@
...
+import { VaultNavService } from "./vault-nav.service";
+
+const EMPTY_VIEW_MODEL: VaultsNavViewModel = {
+  showVaultsHeader: false,
+  vaults: [],
+  myVaultItem: null,
+  allItemsItem: null,
+  orgDefaultExpanded: false,
+  showMyItemsGroup: false,
+};
```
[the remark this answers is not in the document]
gbubemismith: Fixed with readonly VaultNavItemViewModel[]

libs/vault/src/services/default-vault-nav.service.ts
```diff
@@ -0,0 +1,127 @@
...
+        label: org.name,
+        color: this.orgColor(org),
+        type:
+          org.productTierType === ProductTierType.Families
+            ? VaultNavItemType.Family
+            : VaultNavItemType.Organization,
+      }));
+
+    return {
+      showVaultsHeader: !dataOwnership,
```
shane-melton: ❓ Shouldn't the "Vaults" section header be shown whenever there are multiple vaults available to the user?  If so, we can probably just drop the `showVaultsHeader` property entirely and derive it in the template. e.g. `@if(vaults.length > 1)`

libs/vault/src/models/vault-nav-view-model.ts
```diff
@@ -0,0 +1,59 @@
...
+   * Ordered vault items. Populated when the user has org memberships; personal vault is always
+   * first, org vaults follow alphabetically by name.
+   */
+  vaults: readonly VaultNavItemViewModel[];
+
+  /**
+   * Non-null when the user has no org memberships and is on a free plan.
+   * The nav renders "My vault" as a plain top-level item (no section header).
+   */
+  myVaultItem: VaultNavItemViewModel | null;
```
shane-melton: ❓ This seems like it'll be redundant with the personal vault that is always first in `vaults` above. I think we can probably drop this, especially if the `showVaultsHeader` is calculated using `vaults.length`.

libs/vault/src/models/vault-nav-view-model.ts
```diff
@@ -0,0 +1,59 @@
...
+
+  /**
+   * Non-null when the user has no org memberships and has premium from any source.
+   * The nav renders "All items" as the top-level item (no section header).
+   */
+  allItemsItem: VaultNavItemViewModel | null;
```
shane-melton: ❓ Same thing here -- If we're only and always going to show the "All Items" entry when the user belongs to multiple vaults, we can just check the `vaults.length` again.

shane-melton (dismissed):
A few clarifying questions

libs/vault/src/services/default-vault-nav.service.ts
gbubemismith: Simplified model 3127d4191bf7cef1bd7e0e1dd40bb0aed2b5c70a

libs/vault/src/models/vault-nav-view-model.ts
gbubemismith: Simplified model https://github.com/bitwarden/clients/commit/3127d4191bf7cef1bd7e0e1dd40bb0aed2b5c70a

libs/vault/src/models/vault-nav-view-model.ts
gbubemismith: Simplified model https://github.com/bitwarden/clients/commit/3127d4191bf7cef1bd7e0e1dd40bb0aed2b5c70a

libs/vault/src/services/default-vault-nav.service.spec.ts
```diff
@@ -0,0 +1,195 @@
...
+        { provide: PolicyService, useValue: policyService },
+        { provide: AvatarService, useValue: avatarService },
+        { provide: I18nService, useValue: i18nService },
+      ],
+    });
+
+    service = TestBed.inject(DefaultVaultNavService);
+  });
+
+  describe("Scenario: No organization memberships", () => {
```
nick-livefront: ⛏️ (non-blocking) more writing this to shake my fists at Claude, `Scenario: ` here isn't a pattern that we have, `describe` intuitively describes the scenario 🤷

libs/vault/src/services/default-vault-nav.service.ts
```diff
@@ -0,0 +1,94 @@
...
+    const sortedOrgItems: VaultNavItemViewModel[] = [...orgs]
+      .sort((a, b) => a.name.localeCompare(b.name))
+      .map((org) => ({
+        id: org.id,
+        label: org.name,
+        color: this.orgColor(org),
+        type:
+          org.productTierType === ProductTierType.Families
+            ? VaultNavItemType.Family
+            : VaultNavItemType.Organization,
```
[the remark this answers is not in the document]
nick-livefront: I believe that Claude is correct here that `Free` should use the family icon. Agree?

nick-livefront (dismissed):

BryanCunningham (dismissed):

libs/vault/src/services/default-vault-nav.service.ts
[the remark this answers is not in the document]
gbubemismith: Yeah also returning the icon string directly as well

libs/vault/src/services/default-vault-nav.service.spec.ts
gbubemismith: Lol I think this should be blocking

nick-livefront (dismissed):

nick-livefront (approved):

BryanCunningham (approved):

PR #22128, opened 2026-07-29
feat(crypto): add user-key id tracking

quexten:
https://bitwarden.atlassian.net/browse/PM-39454

https://github.com/bitwarden/server/pull/8076
https://github.com/bitwarden/sdk-internal/pull/1307
https://github.com/bitwarden/clients/pull/22128

We want to track the current key ID of the user-key. This is required for encryption stability. With follow-up work this will allow us to:
- Reject KDF Changes, Password changes, Account Recovery Enrollments, Emergency Access Enrollements, Vault Item Modifications, Vault Item Creations, that have a wrong user-key contained / used for encryption. The cryptographic metadata is compared against the stored key id.
- Find bugs easily leading to the above encryption stability

The user key id is set as part of the registration for new users. For existing users an upgrade endpoint that can be used *once* is provided. Key rotations will also set the key id.

Password changes, and kdf changes may not change the key ID and will reject in case the user had a wrong user-key in their client. 

## Client Specific notes

Syncing logic of KM is *mostly* where possible moved to the SDK. We may move other syncing logic (organization keys, webauthn unlock) to the SDK into the same KM handler in a future PR. However, we do copy over all syncing logic, I have added GitHub comments referencing the copied over logic.

## Screenshot
image

### DIFF
--- a/libs/common/src/platform/sync/default-sync.service.ts
+++ b/libs/common/src/platform/sync/default-sync.service.ts
@@ -7,8 +7,8 @@ import { firstValueFrom, map } from "rxjs";
 import { CollectionService } from "@bitwarden/admin-console/common";
 // This import has been flagged as unallowed for this class. It may be involved in a circular dependency loop.
 // eslint-disable-next-line no-restricted-imports
-import { KdfConfigService, KeyService } from "@bitwarden/key-management";
-import { EncString as SdkEncString } from "@bitwarden/sdk-internal";
+import { KeyService } from "@bitwarden/key-management";
+import { KeyId as SdkKeyId } from "@bitwarden/sdk-internal";
 
 // This import has been flagged as unallowed for this class. It may be involved in a circular dependency loop.
 // eslint-disable-next-line no-restricted-imports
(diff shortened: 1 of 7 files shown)


libs/common/src/platform/sync/default-sync.service.ts
```diff
@@ -246,27 +243,11 @@ export class DefaultSyncService extends CoreSyncService {
...
     }
 
-    // Users with no master password will not have a key.
+    // This is for key-connector users
     if (response?.key) {
       await this.masterPasswordService.setMasterKeyEncryptedUserKey(response.key, response.id);
     }
 
-    // Cleanup: Only the first branch should be kept after the server always returns accountKeys https://bitwarden.atlassian.net/browse/PM-21768
-    if (response.accountKeys != null) {
```
quexten, line 255: Replaced by https://github.com/bitwarden/sdk-internal/blob/f5d5ef02bb4253f047302c980bfddcb0a71d6dcb/crates/bitwarden-km-sync-handler/src/km_sync_handler.rs#L120

libs/common/src/platform/sync/default-sync.service.ts
```diff
@@ -457,22 +438,46 @@ export class DefaultSyncService extends CoreSyncService {
...
   private async syncUserDecryption(
     userId: UserId,
     userDecryption: UserDecryptionResponse | undefined,
   ) {
     if (userDecryption == null) {
       return;
     }
-    if (userDecryption.masterPasswordUnlock != null) {
-      const masterPasswordUnlockData =
-        userDecryption.masterPasswordUnlock.toMasterPasswordUnlockData();
```
quexten, line 469: Replaced by https://github.com/bitwarden/sdk-internal/blob/f5d5ef02bb4253f047302c980bfddcb0a71d6dcb/crates/bitwarden-km-sync-handler/src/km_sync_handler.rs#L106

libs/common/src/platform/sync/default-sync.service.ts
```diff
@@ -457,22 +438,46 @@ export class DefaultSyncService extends CoreSyncService {
...
-      await this.masterPasswordService.setMasterPasswordUnlockData(
-        masterPasswordUnlockData,
-        userId,
-      );
-      await this.kdfConfigService.setKdfConfig(userId, masterPasswordUnlockData.kdf);
```
quexten, line 474: Todo: link replacement; do not merge before replacement is linked.

libs/common/src/platform/sync/default-sync.service.ts
```diff
@@ -511,14 +516,5 @@ export class DefaultSyncService extends CoreSyncService {
...
       }
     }
-
-    if (userDecryption.v2UpgradeToken != null) {
-      await this.v2UpgradeTokenStateService.setV2UpgradeToken(
-        userDecryption.v2UpgradeToken.toV2UpgradeToken(),
-        userId,
-      );
-    } else {
-      await this.v2UpgradeTokenStateService.clearV2UpgradeToken(userId);
```
quexten, line 521: Replaced here: https://github.com/bitwarden/sdk-internal/blob/f5d5ef02bb4253f047302c980bfddcb0a71d6dcb/crates/bitwarden-km-sync-handler/src/km_sync_handler.rs#L112

libs/common/src/platform/sync/default-sync.service.ts
quexten: https://github.com/bitwarden/sdk-internal/blob/51c596e37832f45cb6e37d0f7b6e69ecdf5f4982/crates/bitwarden-km-sync-handler/src/km_sync_handler.rs#L100

PR #22132, opened 2026-07-29
feat(vault): enable SSH key import from clipboard on web vault

maxkpower:
## 🎟️ Tracking
https://bitwarden.atlassian.net/browse/PM-41099

## 📔 Objective

Enable SSH key import from clipboard in the web vault. It was previously blocked by a ClientType.Web check even though desktop and browser already support it.

- Implement readFromClipboard() in WebPlatformUtilsService (Clipboard API with a legacy execCommand fallback). Was a stub that threw.
- Remove the ClientType.Web exclusion in SshKeySectionComponent.
- Add unit test coverage for both.

Manually verified: valid key import populates private key, public key, and fingerprint correctly; invalid clipboard content shows an error toast without corrupting the form; import button shows correctly in create and edit modes.

### DIFF
--- a/apps/web/src/app/core/web-platform-utils.service.ts
+++ b/apps/web/src/app/core/web-platform-utils.service.ts
@@ -212,8 +212,44 @@ export class WebPlatformUtilsService implements PlatformUtilsService {
     }
   }
 
-  readFromClipboard(options?: any): Promise<string> {
-    throw new Error("Cannot read from clipboard on web.");
+  async readFromClipboard(): Promise<string> {
+    if (this.isClipboardApiSupported(window, "readText")) {
+      try {
+        return await window.navigator.clipboard.readText();
+      } catch (e) {
+        this.logService.debug(
+          `Error reading from clipboard using the clipboard API, attempting legacy method: ${e}`,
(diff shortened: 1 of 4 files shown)


apps/web/src/app/core/web-platform-utils.service.ts
```diff
@@ -212,8 +212,44 @@ export class WebPlatformUtilsService implements PlatformUtilsService {
     }
   }
 
-  readFromClipboard(options?: any): Promise<string> {
-    throw new Error("Cannot read from clipboard on web.");
+  async readFromClipboard(): Promise<string> {
```
djsmith85, line 215: ⚠️ Missing documentation on public method

djsmith85 (changes requested):
Changes are looking good, but I was curious if we already did something similar elsewhere.

The `browser-platform-utils.service` also implements the `readFromClipboard()` method and calls a `BrowserClipboardService` to verify and use the available APIs. So instead of re-implementing this, we should move the `browser-clipboard.service.ts` into a shareable location and also use it from `web-platform-utils.service`.

PR #22137, opened 2026-07-29
[PM-41027] fix: label add-item tile as shared folder under VFO1 terminology

gbubemismith:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-41027

Addresses the first bullet only ("+ Add new item tile"). The Bulk Batch Bar and item
overflow menu bullets on the same ticket are not covered here.

## 📔 Objective

With `vfo1-foundation` enabled, the Collection tile in the add-item grid still read "Collection" instead of "Shared folder".

`AddItemGridComponent` was already resolving the tile's *icon* through `Vfo1TerminologyService.iconClass()`, but its `labelKey` was hard-coded to `"collection"`. The dropdown-menu path in `new-cipher-menu.component.html` already handled this via the `vfo1I18n` pipe — the dialog path was the gap. The label now swaps to the existing `sharedFolder` key when the flag is on.

## 📸 Screenshots
image


jengstrom-bw (approved):
LGTM

PR #22142, opened 2026-07-29
Add t:deps label for SDK update PRs automatically

trmartin4:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-41116

## 📔 Objective

Adds `t:deps` label automatically when creating a PR for SDK updates in `clients`, so that the label requirement check doesn't fail and require manually adding the label on every PR.


djsmith85 (approved):
t:deps would also be added, once https://github.com/bitwarden/clients/pull/22127 merges, but adding it on creation is better.

PR #22157, opened 2026-07-30
[BRE-2116] Bump actions/checkout to v7.0.1 in non-privileged workflows

brandonbiete:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/BRE-2116

## 📔 Objective

Mechanical version bump (`v6.0.2` → `v7.0.1`, SHA-pinned) across 13 workflows that use `pull_request` / `push` / `schedule` / `workflow_run` triggers — none of which are the privileged `pull_request_target` fork-checkout pattern that v7 guards. The `pull_request` workflows here still check out and run fork PRs normally; they're just not privileged, so v7's fork-block doesn't apply and no `allow-unsafe-pr-checkout` is needed. The privileged `pull_request_target` build workflows are handled separately.


djsmith85 (commented):
@brandonbiete: I had started looking into this non-fork change due to https://github.com/bitwarden/clients/pull/21682.
Changes are looking good so far, just one question regarding `lint.yml` and `test.yml`. I would expect both of those to also be run on community contributions. In case we do go forward with this, do we have a way to trigger these independently on a fork?

AmyLGalles (approved):
LGTM

brandonbiete:
> @brandonbiete: I had started looking into this non-fork change due to #21682. Changes are looking good so far, just one question regarding `lint.yml` and `test.yml`. I would expect both of those to also be run on community contributions. In case we do go forward with this, do we have a way to trigger these independently on a fork?

@djsmith85 From my understanding after looking at `test.yml` and `lint.yml`, they should trigger for forked PRs. That's how they function today. This upgrade should not affect that. I may be misinterpreting your question but hopefully that answers your concern.

djsmith85 (approved):
Discussed my concerns offline and happy to approve

jprusik (approved):
Will want to watch the BIT workflow after merge, but otherwise LGTM

coltonhurst (approved):

PR #22159, opened 2026-07-30
Pm 39982 web share

bmbitwarden:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-39982

## 📔 Objective

Create the Share drawer (side panel) displayed in the web app when a user clicks "Share via link" on a vault item. Layout mirrors the browser extension Share panel.

## 📸 Screenshots

Screenshot 2026-08-31 at 09 45 19

**NOTE**
The build errors are expected until we get [this SDK change](https://github.com/bitwarden/sdk-internal/pull/1445) in


libs/angular/src/scss/bwicons/styles/style.css
```diff
@@ -109,6 +109,10 @@
   content: "\f175";
 }
 
+.bwi-add-circle:before {
+  content: "\f173";
+}
```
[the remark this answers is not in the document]
vleague2: Just fyi, this is an auto-generated file and there is a script for it https://github.com/bitwarden/clients/blob/main/scripts/material-icons/README.md

libs/common/src/tools/send/services/send.service.ts
```diff
@@ -147,6 +147,17 @@ export class SendService implements InternalSendServiceAbstraction {
...
+        const encryptionContext = await this.cipherEncryptionService.encryptCipherForRotation(
+          model.data.data,
+          userId,
+          model.cryptoKey as any,
+        );
+        send.data = {
+          data: encryptionContext.cipher,
+        };
+      }
     }
```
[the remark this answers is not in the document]
mcamirault: This is intentional

libs/common/src/tools/send/models/domain/send.ts
```diff
@@ -23,13 +25,14 @@ import { SendView } from "../view/send.view";
 import { SendFile } from "./send-file";
 import { SendText } from "./send-text";
 
-const SEND_TYPE_TO_SDK: Record<SendType, SdkSendType> = {
+// SendType.Item (2) has no SDK equivalent yet — cast to allow the partial mapping.
+const SEND_TYPE_TO_SDK = {
```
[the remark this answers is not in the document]
mcamirault: Fixed

libs/angular/src/scss/bwicons/styles/style.css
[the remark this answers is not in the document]
mcamirault: I believe this should be fixed now

libs/common/src/tools/send/services/send.service.ts
```diff
@@ -52,7 +52,7 @@ export class SendService implements InternalSendServiceAbstraction {
     private keyGenerationService: KeyGenerationService,
     private stateProvider: SendStateProvider,
     private encryptService: EncryptService,
-    private configService: ConfigService,
+    private cipherEncryptionService: CipherEncryptionService,
```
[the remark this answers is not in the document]
mcamirault: Fixed

libs/vault/src/share-link.service.ts
```diff
@@ -0,0 +1,147 @@
...
+    if (oneTimeShare) {
+      sendView.maxAccessCount = 1;
+    }
+    const sharedCipherView = new CipherView();
+    Object.assign(sharedCipherView, cipherView);
+    sharedCipherView.attachments = [];
+    if (sharedCipherView.login) {
+      sharedCipherView.login.fido2Credentials = [];
+    }
+    delete sharedCipherView.key;
```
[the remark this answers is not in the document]
mcamirault: Done

apps/web/src/app/admin-console/organizations/policies/policy-edit-definitions/send-controls.component.ts
```diff
@@ -130,6 +130,13 @@ export class SendControlsPolicyComponent extends BasePolicyEditComponent impleme
       labelName: this.i18nService.t("sendTypeFile"),
       value: SendType.File,
     },
+    {
+      id: SendType.Item.toString(),
+      icon: "bwi-lock",
+      listName: this.i18nService.t("item"),
+      labelName: this.i18nService.t("item"),
+      value: SendType.Item,
+    },
```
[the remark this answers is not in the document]
mcamirault: Fixed

apps/web/src/app/tools/send/send.component.ts
```diff
@@ -149,15 +149,44 @@ export class SendComponent implements OnDestroy {
...
+    combineLatest([this.sendPolicyService.allowedSendTypes$, this.hasPremium$]).pipe(
+      takeUntilDestroyed(),
+      map(([allowedSendTypes, hasPremium]) => {
+        if (!hasPremium) {
+          return allowedSendTypes.filter((st) => st !== SendType.File && st !== SendType.Item);
+        }
+        return allowedSendTypes;
+      }),
+    ),
+  );
```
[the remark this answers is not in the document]
mcamirault: Fixed

apps/web/src/app/vault/individual-vault/vault.component.ts
```diff
@@ -1086,6 +1090,17 @@ export class VaultComponent<C extends CipherViewLike> implements OnInit, OnDestr
...
+  async shareViaLink(cipher: CipherViewLike) {
+    const activeUserId = await firstValueFrom(getUserId(this.accountService.activeAccount$));
+    const cipherView = await firstValueFrom(
+      this.cipherService.cipherView$(activeUserId, cipher.id as CipherId),
+    );
+    if (!cipherView) {
+      return;
+    }
+    await this.dialogService.openDrawer(ShareItemDrawerComponent, { data: { cipher: cipherView } });
+  }
```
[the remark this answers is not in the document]
mcamirault: Fixed

apps/cli/src/service-container/service-container.ts
```diff
@@ -673,7 +673,7 @@ export class ServiceContainer {
       this.keyGenerationService,
       this.sendStateProvider,
       this.encryptService,
-      this.configService,
+      this.cipherEncryptionService,
```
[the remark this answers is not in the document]
mcamirault: Fixed

apps/desktop/src/vault/app/vault-v3/vault-list.component.ts
```diff
@@ -116,6 +116,7 @@ export class VaultListComponent<C extends CipherViewLike> {
   protected readonly allCollections = input<CollectionView[]>([]);
   protected readonly userCanArchive = input<boolean>();
   protected readonly enforceOrgDataOwnershipPolicy = input<boolean>();
+  protected readonly showShareViaLink = input<boolean>(false);
```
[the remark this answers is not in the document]
mcamirault: Fixed

apps/desktop/src/locales/en/messages.json
```diff
@@ -6317,6 +6317,10 @@
   "duplicateAttachmentNameError": {
     "message": "Items cannot have multiple attachments with the same file name"
   },
+  "shareViaLink": {
+    "message": "Share via link",
+    "description": "Menu option to share a vault item via a shareable link"
+  },
```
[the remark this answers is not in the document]
mcamirault: Fixed

libs/vault/src/share-item-form/share-item-form.component.ts
```diff
@@ -0,0 +1,305 @@
...
+      });
+    } else {
+      this.toastService.showToast({
+        variant: "error",
+        message: this.i18nService.t("linkSavedCopyFailed"),
+      });
+    }
+
+    this.form.controls.emails.reset();
+  }
```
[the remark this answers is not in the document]
mcamirault: Fixed

libs/vault/src/vault-item-dialog/vault-item-dialog.component.ts
```diff
@@ -319,6 +321,10 @@ export class VaultItemDialogComponent implements OnInit, OnDestroy {
     { initialValue: false },
   );
 
+  protected readonly showShareButton = toSignal(
+    this.shareLinkService.cipherCanBeShared(this.cipher),
+  );
```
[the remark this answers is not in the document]
mcamirault: Fixed

apps/web/src/app/vault/components/vault-items/vault-cipher-row.component.ts
```diff
@@ -165,17 +169,20 @@ export class VaultCipherRowComponent<C extends CipherViewLike> implements OnInit
...
     private accountService: AccountService,
     private cipherService: CipherService,
     private platformUtilsService: PlatformUtilsService,
     private configService: ConfigService,
+    private shareLinkService: ShareLinkService,
   ) {
     this.showCopyAndLaunchActions$ = this.configService.getFeatureFlag$(
       FeatureFlag.PM28091_AddCopyAndQuickLaunchActions,
     );
+    this.showShareViaLink$ = this.shareLinkService.cipherCanBeShared(this.cipher);
```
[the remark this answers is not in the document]
mcamirault: Fixed

libs/vault/src/share-item-form/share-item-form.component.ts
```diff
@@ -0,0 +1,315 @@
...
+      .subscribe((sendControlsPolicies) => {
+        const policyWithDeletionDate = sendControlsPolicies.find(
+          (scp) => scp.data.deletionHours != undefined,
+        );
+        if (policyWithDeletionDate) {
+          const expiryHoursFormControl = this.form.get("expiryHours");
+          expiryHoursFormControl?.setValue(policyWithDeletionDate.data.deletionHours);
+          expiryHoursFormControl?.disable();
+        }
+      });
```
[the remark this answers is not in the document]
mcamirault: Should be fixed now

apps/browser/src/tools/popup/share/share-item.component.ts
```diff
@@ -0,0 +1,69 @@
...
+  ],
+})
+export class ShareItemComponent {
+  private readonly route = inject(ActivatedRoute);
+  private readonly location = inject(Location);
+  private readonly cipherService = inject(CipherService);
+  private readonly accountService = inject(AccountService);
+
+  protected readonly cipher = signal<CipherView | null>(null);
+  protected readonly shareItemForm = viewChild.required(ShareItemFormComponent);
```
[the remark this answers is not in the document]
mcamirault: Fixed

apps/desktop/src/vault/app/vault-v3/vault-list.component.ts
```diff
@@ -175,6 +177,10 @@ export class VaultListComponent<C extends CipherViewLike> {
     { initialValue: false },
   );
 
+  private readonly temporaryItemSharingEnabled = toSignal(
+    this.configService.getFeatureFlag$(FeatureFlag.PM34203TemporaryItemSharing),
+  );
```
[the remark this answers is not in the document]
mcamirault: Removed

apps/desktop/src/vault/app/vault-v3/vault-list.component.html
```diff
@@ -94,6 +94,7 @@
               (onEvent)="event($event)"
               [userCanArchive]="userCanArchive()"
               [enforceOrgDataOwnershipPolicy]="enforceOrgDataOwnershipPolicy()"
+              [showShareViaLinkOption]="shareLinkService.cipherCanBeShared(item.cipher) | async"
```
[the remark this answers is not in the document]
mcamirault: Fixed

libs/vault/src/share-link.service.ts
```diff
@@ -0,0 +1,191 @@
...
+          return false;
+        }
+        if (CipherViewLikeUtils.getType(c) === CipherType.SshKey) {
+          return false;
+        }
+        if (c.collectionIds.length === 0) {
+          return true;
+        }
+        return c.collectionIds.some((cId) => collections.some((c) => c.id === cId && !c.readOnly));
+      }),
```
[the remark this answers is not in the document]
mcamirault: Fixed logic

apps/web/src/app/admin-console/organizations/policies/policy-edit-definitions/send-controls.component.ts
```diff
@@ -117,22 +117,39 @@ export class SendControlsPolicyComponent extends BasePolicyEditComponent impleme
...
+            icon: "bwi-lock",
+            listName: this.i18nService.t("vaultItems"),
+            labelName: this.i18nService.t("vaultItems"),
+            value: SendType.Item,
+          });
+        }
+        return options;
+      }),
+    );
+  protected readonly allSendTypeOptions = toSignal(this.allSendTypeOptions$, { initialValue: [] });
```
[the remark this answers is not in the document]
mcamirault: Saving closes the drawer, so this is currently moot

apps/web/src/app/vault/components/vault-items/vault-cipher-row.component.html
```diff
@@ -323,6 +323,12 @@
             {{ "assignToCollections" | vfo1I18n: "addToSharedFolder" }}
           </button>
         }
+        @if (showShareViaLink$ | async) {
```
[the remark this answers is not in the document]
mcamirault: Handled in Admin Console vault component

libs/tools/share/src/services/share-link.service.ts
```diff
@@ -0,0 +1,208 @@
...
+        }
+        if (CipherViewLikeUtils.getType(c) === CipherType.SshKey) {
+          return false;
+        }
+        if (c.collectionIds.length === 0) {
+          return true;
+        }
+        return c.collectionIds.some((cId) =>
+          collections.some((col) => col.id === cId && !col.readOnly),
+        );
```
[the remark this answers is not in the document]
mcamirault: Fixed logic

apps/web/src/app/tools/send/send-access/send-access-item-field.component.html
```diff
@@ -0,0 +1,33 @@
...
+      ></button>
+    }
+    @if (copyable()) {
+      <button
+        bitIconButton="bwi-clone"
+        bitSuffix
+        type="button"
+        [appCopyClick]="value()"
+        showToast
+        [valueLabel]="label() | i18n"
```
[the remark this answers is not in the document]
mcamirault: Fixed

libs/common/src/tools/send/models/domain/send-item.ts
```diff
@@ -0,0 +1,70 @@
...
+
+  /** Maps an SDK `SendItem` back to a domain `SendItem`. */
+  static fromSdk(obj?: SdkSendItem): SendItem {
+    if (obj == null) {
+      return null;
+    }
+
+    return Object.assign(new SendItem(), {
+      encryptionVersion: obj.encryptionVersion,
+      data: obj.data,
```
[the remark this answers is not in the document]
mcamirault: Fixed

libs/tools/share/src/components/share-item-form/share-item-form.component.ts
```diff
@@ -0,0 +1,315 @@
...
+  }
+
+  protected async deleteLink(link: ShareLink): Promise<void> {
+    await this.shareLinkService.deleteLink(link.sendId);
+    this.toastService.showToast({
+      variant: "success",
+      title: undefined,
+      message: this.i18nService.t("shareLinkDeleted"),
+    });
+  }
```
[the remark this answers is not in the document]
mcamirault: Fixed

apps/web/src/app/admin-console/organizations/collections/vault.component.ts
```diff
@@ -737,6 +741,10 @@ export class VaultComponent implements OnInit, OnDestroy {
       ...navigateOptions,
     });
   }
+
+  private async shareViaLink(cipher: CipherView) {
+    await this.dialogService.openDrawer(ShareItemDrawerComponent, { data: { cipher } });
+  }
```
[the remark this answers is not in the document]
mcamirault: Fixed

vleague2 (dismissed):
UIF icon changes look good!

libs/tools/share/src/components/share-item-drawer/share-item-drawer.component.html
gbubemismith, line 1: I don't think vault should own these components. They should be owned by tools and then reused when needed. What do you think?

libs/tools/share/src/components/share-item-form/share-item-form.component.html
gbubemismith, line 1: Same here. I don't think vault should own this

libs/tools/share/src/services/share-link.service.ts
gbubemismith, line 1: Same as this service. This shouldn't be owned by Vault

libs/tools/share/src/types/share-link.types.ts
gbubemismith, line 1: Same issue here

gbubemismith (changes requested):

apps/desktop/src/app/tools/share-item/share-item-desktop.component.html
gbubemismith, line 1: I think tools can own this too

gbubemismith:
@bmbitwarden I see a lot of new components created under vault. I think they should be owned by tools and only reused on vault owned components.

libs/tools/share/src/components/share-item-form/share-item-form.component.ts
```diff
@@ -0,0 +1,340 @@
...
+  });
+
+  private readonly activeUserId$ = this.accountService.activeAccount$.pipe(getUserId);
+  private readonly policyAllowedDomains = signal<string[] | null>(null);
+
+  constructor() {
+    this.sendPolicyService.deletionDatePolicyInfo$.pipe(takeUntilDestroyed()).subscribe((dh) => {
+      if (dh) {
+        const expiryHoursFormControl = this.form.get("expiryHours");
+        expiryHoursFormControl?.setValue((dh as any) ?? ExpiryOption.SevenDays);
```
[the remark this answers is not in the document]
mcamirault: Fixed

libs/tools/share/project.json
```diff
@@ -0,0 +1,41 @@
...
+      "executor": "nx:run-script",
+      "options": {
+        "script": "clean"
+      }
+    },
+    "lint": {
+      "executor": "@nx/eslint:lint",
+      "outputs": ["{options.outputFile}"],
+      "options": {
+        "lintFilePatterns": ["libs/tools/share/**/*.ts"]
```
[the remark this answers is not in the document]
mcamirault: Jest setup added

apps/browser/src/vault/popup/components/vault/item-more-options/item-more-options.component.ts
```diff
@@ -32,7 +32,7 @@ import {
   ToastService,
   IconModule,
 } from "@bitwarden/components";
-import { PasswordRepromptService, Vfo1I18nPipe } from "@bitwarden/vault";
+import { PasswordRepromptService, ShareLinkService, Vfo1I18nPipe } from "@bitwarden/vault";
```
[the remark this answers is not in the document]
mcamirault: Fixed remaining import sites

apps/browser/src/_locales/en/messages.json
```diff
@@ -6032,6 +6032,50 @@
...
   },
+  "share": {
+    "message": "Share",
+    "description": "Page title for the share vault item panel"
+  },
+  "shareViaLink": {
+    "message": "Share via link",
+    "description": "Menu option to share a vault item via a shareable link"
+  },
+  "expiryOneHour": {
```
itsadrago: I think you can reduce a lot of time related messages that just reflect a range of time to something like the following. 

```suggestion
"timeRange": {
  "message": "$NUMBER$ $UNIT_OF_TIME$",
  "placeholders": {
    "number": {
      "content": "$1",
      "example": "1"
    },
    "unit_of_time": {
      "content": "$2",
      "example": "hour"
    }
  }
}
```
This would allow you to keep the number and unit_of_time values in the code and not bake them into the content. It also results in fewer messages for translators to manage. You may need to add translations for "hour", "hours", "day", "days", etc... though if they are missing

mcamirault: We actually don't need separate message at all, these expiration options are the exact same as the presets we use for Send deletion dates so we can reuse the messages from there.

libs/tools/share/src/components/share-item-form/share-item-form.component.spec.ts
```diff
@@ -0,0 +1,184 @@
...
+        },
+        { provide: PlatformUtilsService, useValue: platformUtilsService },
+        { provide: ToastService, useValue: toastService },
+        { provide: I18nService, useValue: i18nService },
+        { provide: ShareLinkService, useValue: shareLinkService },
+        { provide: CollectionService, useValue: collectionService },
+        { provide: FolderService, useValue: folderService },
+        { provide: PolicyService, useValue: policyService },
+        { provide: LogService, useValue: mock<LogService>() },
+      ],
```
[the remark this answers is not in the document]
mcamirault: Fixed

libs/tools/share/jest.config.js
```diff
@@ -0,0 +1,19 @@
...
+  ...sharedConfig,
+  displayName: "tools/share tests",
+  setupFilesAfterEnv: ["<rootDir>/test.setup.ts"],
+  moduleNameMapper: pathsToModuleNameMapper(
+    // lets us use @bitwarden/common/spec in tests
+    { "@bitwarden/common/spec": ["libs/common/spec"], ...(compilerOptions?.paths ?? {}) },
+    {
+      prefix: "<rootDir>/../../",
+    },
+  ),
```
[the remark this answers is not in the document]
mcamirault: Fixed the wiring

apps/web/src/app/vault/components/vault-items/vault-items.stories.ts
```diff
@@ -48,7 +48,11 @@ import { LayoutComponent, StorybookGlobalStateProvider, ToastService } from "@bi
 // eslint-disable-next-line no-restricted-imports
 import { SymmetricCryptoKey } from "@bitwarden/legacy-crypto";
 import { GlobalStateProvider } from "@bitwarden/state";
-import { RoutedVaultFilterService, PasswordRepromptService } from "@bitwarden/vault";
+import {
+  RoutedVaultFilterService,
+  PasswordRepromptService,
+  ShareLinkService,
+} from "@bitwarden/vault";
```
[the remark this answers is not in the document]
mcamirault: Fixed

libs/tools/share/project.json
```diff
@@ -0,0 +1,41 @@
+{
+  "name": "@bitwarden/tools-share",
+  "$schema": "../../../../node_modules/nx/schemas/project-schema.json",
```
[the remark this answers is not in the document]
mcamirault: Fixed

libs/tools/share/jest.config.js
```diff
@@ -0,0 +1,30 @@
+const { pathsToModuleNameMapper } = require("ts-jest");
+
+const { compilerOptions } = require("../../../tsconfig.base");
+
+const { createCjsPreset } = require("jest-preset-angular/presets");
+
+// FIXME: Should use the shared config!
```
coroiu, line 7: suggestion(non-blocking): might wanna reorganize your libs to fix this

coroiu (dismissed):
approved with a non-blocking comment

apps/browser/src/tools/popup/share/share-item.component.ts
```diff
@@ -0,0 +1,72 @@
...
+    ShareItemFormComponent,
+  ],
+})
+export class ShareItemComponent {
+  private readonly route = inject(ActivatedRoute);
+  private readonly location = inject(Location);
+  private readonly cipherService = inject(CipherService);
+  private readonly accountService = inject(AccountService);
+
+  protected readonly cipher = signal<CipherView | null>(null);
```
BTreston: per [ADR](https://contributing.bitwarden.com/architecture/adr/typescript-strict#guidelines), avoid `null`, prefer `undefined`.

BTreston: 🎨 We can probably initialize this with `toSignal` here rather than use `.subscribe()` and `set()`. Also we should avoid nesting pipes. We already have a `activeUserId$` that we can probably reuse here :

```typescript
protected readonly cipher = toSignal(
  combineLatest([this.activeUserId$, this.cipherId$]).pipe(
    switchMap(([userId, cipherId]) => this.cipherService.cipherView$(userId, cipherId)),
  ),
  { initialValue: undefined },
);
```

apps/web/src/app/admin-console/organizations/policies/policy-edit-definitions/send-controls.component.html
```diff
@@ -8,7 +8,7 @@
       <bit-label>{{ "sendTypes" | i18n }}</bit-label>
       <bit-multi-select
         [formControl]="allowedSendTypesMultiSelectControl"
-        [baseItems]="allSendTypeOptions()"
+        [baseItems]="allSendTypeOptions$ | async"
```
BTreston: Should this use the `allSendTypeOptions` signal? We have 2 properties for this data, it may be more clear to inline the observable login in the `toSignal` and eliminate the observable class member

libs/tools/share/project.json
```diff
@@ -0,0 +1,41 @@
+{
+  "name": "@bitwarden/tools-share",
```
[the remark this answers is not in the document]
mcamirault: Since this doesn't affect the components as currently written I'm not going to fix this yet to avoid dismissing peoples' reviews. Making a note to fix this as an immediate follow-up

apps/web/src/app/admin-console/organizations/policies/policy-edit-definitions/send-controls.component.ts
```diff
@@ -243,12 +287,8 @@ export class SendControlsPolicyComponent extends BasePolicyEditComponent impleme
     this.data.patchValue(policyResponseData);
 
     // The two separate form controls (enabled toggle and Send Types multi-select) must be initialized separately
+    // The Send Types multi-select is initialized by the `effect` in the constructor
```
jrmccannon: First line says that both form controls are present then the second line says that the multi-select has moved. 

Comment can probably just be removed.

jrmccannon (dismissed):

apps/web/src/app/admin-console/organizations/policies/policy-edit-definitions/send-controls.component.ts
mcamirault: I think it's important to note that two of the controls are initialized separately from the rest of the policy form, and the only reason that they are separated right now is because one depends on a feature flag and must be done async. Once the feature flag is removed we'll be able to move that initialization back down here.

apps/web/src/app/admin-console/organizations/policies/policy-edit-definitions/send-controls.component.ts
```diff
@@ -160,10 +180,15 @@ export class SendControlsPolicyComponent extends BasePolicyEditComponent impleme
 
   protected readonly showDeletionHours = new FormControl<boolean>(false);
 
+  protected readonly temporaryItemSharingEnabled = toSignal(
+    this.configService.getFeatureFlag$(FeatureFlag.PM34203TemporaryItemSharing),
+  );
```
BTreston: we're reading this flag from the config service 3x through out this component. We should just use this signal here rather than re-read every time.

itsadrago (dismissed):

libs/tools/share/jest.config.js
mcamirault: Yeah I copied this file from another of our libraries, I'll make a note of this as tech debt

libs/vault/src/index.ts
```diff
@@ -124,5 +124,8 @@ export { VaultBatchActionComponent } from "./components/vault-batch-bar/vault-ba
 export { Vfo1TerminologyService } from "./services/vfo1-terminology.service";
 export { Vfo1I18nPipe } from "./pipes/vfo1-i18n.pipe";
 export { Vfo1IconPipe } from "./pipes/vfo1-icon.pipe";
+
+export * from "./components/totp-countdown/totp-countdown.component";
+export * from "./pipes/credit-card-number.pipe";
```
gbubemismith, line 142: ❓Why are we exporting these? are they being used?

mcamirault: Yes, we're using them in the recipient view for these Item Sends (`apps/web/src/app/tools/send/send-access/send-access-item.component.html`), which displays cipher information similarly to how the vault does but the design is  _just_ different enough that we can't use the existing CipherView component. Here's what a login will look like when shared, which uses the TOTP countdown

Screenshot 2026-08-31 at 11 07 45

gbubemismith: Oh thanks.

apps/web/src/app/admin-console/organizations/policies/policy-edit-definitions/send-controls.component.ts
```diff
@@ -175,6 +200,22 @@ export class SendControlsPolicyComponent extends BasePolicyEditComponent impleme
...
+      if (policyResponse) {
+        this.allowedSendTypesMultiSelectControl.patchValue(
+          allSendTypeOptions.filter((sto) =>
+            (policyResponse.data.allowedSendTypes ?? []).includes(sto.value),
+          ),
+        );
+      } else {
+        this.allowedSendTypesMultiSelectControl.patchValue(allSendTypeOptions);
+      }
+    });
```
BTreston, line 204: This is not an intuitive pattern, we're effectively using change detection->signal emission as imperative clockwork to update this form state. I think this should be handled in `loadData()` once and then the effect can be used to initialize the data once `policyData()` resolves. I think it should look something like:

```typescript
constructor(...) {
  super();
  effect(() => {
    if (this.allSendTypeOptions().length > 0 && this.policyResponse() !== undefined) {
      this.loadData(); // this function handles all the setup we'd need.
    }
  });
}
```

Also I would note that it looks like we're setting these options twice, once in `loadData()` (kinda like what I am proposing) and also inside of this effect which seems redundant.

apps/web/src/app/admin-console/organizations/policies/policy-edit-definitions/send-controls.component.ts
```diff
@@ -117,22 +120,39 @@ export class SendControlsPolicyComponent extends BasePolicyEditComponent impleme
...
+        ];
+        if (isEnabled) {
+          options.push({
+            id: SendType.Item.toString(),
+            icon: "bwi-lock",
+            listName: this.i18nService.t("vaultItems"),
+            labelName: this.i18nService.t("vaultItems"),
+            value: SendType.Item,
+          });
+        }
```
BTreston: I think these options should be defined as the initial value of the signal (with its own type that extends `SelectItemView`) and the feature flag logic here would only push in the new value if enabled.
```typescript
  protected readonly allSendTypeOptions: WritableSignal<NewSelectItemViewType> = toSignal(this.allSendTypeOptions$, { initialValue: [
{
            id: SendType.Text.toString(),
            icon: "bwi-file-text",
            listName: this.i18nService.t("sendTypeText"),
            labelName: this.i18nService.t("sendTypeText"),
            value: SendType.Text,
          },
          {
            id: SendType.File.toString(),
            icon: "bwi-file",
            listName: this.i18nService.t("sendTypeFile"),
            labelName: this.i18nService.t("sendTypeFile"),
            value: SendType.File,
          },
] });
``

File: libs/common/src/tools/send/models/domain/send-item.ts [hunk 44]
```diff
@@ -0,0 +1,70 @@
...
+    if (obj == null) {
+      return;
+    }
+
+    this.encryptionVersion = obj.encryptionVersion;
+    this.data = Cipher.fromJSON(JSON.parse(obj.data));
+  }
+
+  async decrypt(key: SymmetricCryptoKey): Promise<SendItemView> {
+    const cipherView = await this.data.decrypt(key);
```
quexten: Cipher's `decrypt` JSDoc says:
https://github.com/bitwarden/clients/blob/78d082ed393d608410ae6102bd81c7680d9cc8a8/libs/common/src/vault/models/domain/cipher.ts#L144
```
  /**
   * @deprecated WARNING: This API may fail to decrypt ciphers if they are using blob encryption.
   * If you are using this, please migrate off of it immediately! This function will be removed
   * in a near release.
   */
```

Considering it specifically says to migrate off of it immediately, we should not add new callers to it.

libs/common/src/tools/send/models/domain/send-item.ts
```diff
@@ -0,0 +1,70 @@
+// FIXME: Update this file to be type safe and remove this and next line
+// @ts-strict-ignore
```
quexten: Why are we ignoring ts-strict in a new file?

vleague2 (dismissed):

quexten (changes requested):
I'm concerned that we are adding new usages to deprecated functions which directly state that they should not be used and adding new files that have ts-strict disabled. The 'ts-strict-ignore` was only ever meant to help out with not yet migrated files, but should never be used on new files.

libs/common/src/tools/send/models/api/send-item.api.ts
```diff
@@ -0,0 +1,19 @@
+// FIXME: Update this file to be type safe and remove this and next line
+// @ts-strict-ignore
```
quexten: Ts strict ignore should never be used in new files.

```suggestion

```

libs/common/src/tools/send/models/data/send-item.data.ts
```diff
@@ -0,0 +1,19 @@
+// FIXME: Update this file to be type safe and remove this and next line
+// @ts-strict-ignore
```
quexten:
```suggestion
```

Ts strict ignore should never be used in new files.

libs/common/src/tools/send/services/send.service.ts
```diff
@@ -151,6 +152,15 @@ export class SendService implements InternalSendServiceAbstraction {
           model.cryptoKey,
         );
       }
+    } else if (send.type === SendType.Item) {
+      // Since we're encrypting the cipher under the Send's key, not the cipher's, we need to use this method
+      const encryptionContext = await this.cipherEncryptionService.encryptCipherForRotation(
```
quexten: This is not a supported use-case of this api / is mis-using the api. The api clearly is meant for key rotation *not* for cipher sends. Please expose an SDK function through a client appropriate for the purpose.

libs/common/src/tools/send/models/view/send-item.view.ts
```diff
@@ -0,0 +1,17 @@
+// FIXME: Update this file to be type safe and remove this and next line
+// @ts-strict-ignore
```
quexten:
```suggestion
```

Ts strict ignore should not be used in new files.

libs/common/src/tools/send/services/send.service.ts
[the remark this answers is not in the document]
quexten: Sorry but this is directly and intentionally violating the api contract of this function. We require the user-key here for a reason. I'll unresolve this review because i consider it accurate.

libs/vault/src/vault-item-dialog/vault-item-dialog.component.html
```diff
@@ -13,7 +13,22 @@
         [cipher]="cipher"
         [collections]="collections"
         [isAdminConsole]="formConfig.isAdminConsole"
-      ></app-cipher-view>
+      >
+        @if (showShareButton$ | async) {
```
gbubemismith: ❓ is sharing org owned items in scope?

libs/tools/share/src/services/share-link.service.ts
```diff
@@ -0,0 +1,207 @@
...
+    const sendView = new SendView();
+    sendView.name = this.i18nService.t("itemSendTitle", cipherView.name);
+    sendView.type = SendType.Item;
+    sendView.authType = AuthType.Email;
+    sendView.emails = emails;
+    sendView.deletionDate = new Date(Date.now() + expiryHours * 60 * 60 * 1000);
+    if (oneTimeShare) {
+      sendView.maxAccessCount = 1;
+    }
+    const sharedCipherView = CipherView.fromJSON(JSON.parse(JSON.stringify(cipherView)));
```
gbubemismith: ❓ Looks like the whole `cipherView` goes into the send, except the four fields you remove. Should the payload be trimmed to what is needed by the recipient? I think we can do this instead 

``` ts
var shared = new Cipherview()?
shared.name = cipherView.name;
shared.type = cipherView.type;
shared.notes = cipherView.notes;
shared.login = cipherView.login;

```

apps/desktop/src/vault/app/vault-v3/vault-items/vault-cipher-row.component.ts
```diff
@@ -98,6 +110,14 @@ export class VaultCipherRowComponent<C extends CipherViewLike> {
   private i18nService = inject(I18nService);
   private vaultCopyButtonsService = inject(VaultCopyButtonsService);
   private configService = inject(ConfigService);
+  private shareLinkService = inject(ShareLinkService);
+  protected showShareViaLink$: Observable<boolean>;
+
+  constructor() {
+    effect(() => {
```
gbubemismith: ⚠️ Assigning from effect means it's undefined on first render and only gets filled on next change detection. I think we can use `toObservable` instead of effect 

``` ts
protected showShareViaLink$ = toObservable(this.cipher).pipe(
   switchMap((c) => this.shareLinkService.cipherCanBeShared$(c)),
);
```

apps/web/src/app/vault/individual-vault/vault.component.ts
```diff
@@ -1086,6 +1090,20 @@ export class VaultComponent<C extends CipherViewLike> implements OnInit, OnDestr
...
 
+  async shareViaLink(cipher: CipherViewLike) {
+    if (!(await this.repromptCipher([cipher]))) {
+      return;
+    }
+    const activeUserId = await firstValueFrom(getUserId(this.accountService.activeAccount$));
+    const cipherView = await firstValueFrom(
+      this.cipherService.cipherView$(activeUserId, cipher.id as CipherId),
+    );
+    if (!cipherView) {
```
gbubemismith: ❓ do we need an error toast here?

libs/vault/src/vault-item-dialog/vault-item-dialog.component.ts
```diff
@@ -665,6 +671,13 @@ export class VaultItemDialogComponent implements OnInit, OnDestroy {
     }
   };
 
+  protected async openSharePanel() {
+    await this.dialogService.openDrawer(ShareItemDrawerComponent, {
```
gbubemismith: ❓ Desktop opens this dialog too, but this always opens the drawer while desktop's row menu opens a modal. Is the split intentional?

We have run into something similar in vault and we approached this with a token that each client fills in with it's own dialog. Vault declared the token [assign-collections-dialog.token](https://github.com/bitwarden/clients/blob/main/libs/vault/src/tokens/assign-collections-dialog.token.ts ), [token](https://github.com/bitwarden/clients/blob/main/libs/vault/src/services/vault-batch-bar.service.ts#L104) is injected instead of the component, each client then provides it's own adapter. I think we can follow a similar pattern

libs/common/src/tools/send/models/api/send-item.api.ts
BTreston: We shouldn't add new ts-strict-ignores

libs/common/src/tools/send/models/data/send-item.data.ts
BTreston: same as previous comment

libs/common/src/tools/send/models/domain/send-item.ts
BTreston: same as previous comment

libs/common/src/tools/send/models/view/send-item.view.ts
BTreston: same as previous comment

libs/tools/share/src/components/share-item-form/share-item-form.component.ts
```diff
@@ -0,0 +1,341 @@
...
+  private readonly shareLinkService = inject(ShareLinkService);
+  private readonly sendPolicyService = inject(SendPolicyService);
+  private readonly logService = inject(LogService);
+
+  /** The cipher to share. Provided by the shell component. */
+  readonly cipher = input.required<CipherView>();
+
+  protected readonly activeLinks = signal<ShareLink[]>([]);
+  protected readonly collections = signal<CollectionView[]>([]);
+  protected readonly folder = signal<FolderView | null>(null);
```
BTreston: `null` -> `undefined`

libs/tools/share/src/components/share-item-form/share-item-form.component.ts
```diff
@@ -0,0 +1,341 @@
...
+  ): ValidationErrors | null => {
+    if (!control.value) {
+      return null;
+    }
+    const emails = control.value.split(",").map((e: string) => e.trim());
+    const nonEmptyEmails = emails.filter((e: string) => e.length > 0);
+    if (nonEmptyEmails.length === 0) {
+      return { required: true };
+    }
+    const emailRegex = /^[^\s@]+@[^\s@]+\.[^\s@]+$/;
```
BTreston, line 146: non-blocking❓ can `apps/web/src/app/admin-console/organizations/members/components/member-dialog/validators/comma-separated-emails.validator.ts` be reused for this? Or can we at least centralize this one for re-use?

libs/tools/share/src/components/share-item-form/share-item-form.component.ts
```diff
@@ -0,0 +1,341 @@
...
+      validators: [Validators.required, this.emailListValidator, this.emailsMaxLengthValidator],
+    }),
+    expiryHours: new FormControl<SendDeletionDatePreset>(SendDeletionDatePreset.SevenDays, {
+      nonNullable: true,
+    }),
+    oneTimeShare: new FormControl(false, { nonNullable: true }),
+  });
+
+  private readonly activeUserId$ = this.accountService.activeAccount$.pipe(getUserId);
+  private readonly policyAllowedDomains = signal<string[] | null>(null);
```
BTreston: `null` -> `undefined`

BTreston (changes requested):
Looks mostly good, a few points of feedback and questions. Side bar: this PR is really large and would be better as multiple smaller PRs. Granted a lot of it is NX library boilerplate and non-product code. But a good chunk of it is new components, services, and models. It would help reviewers give the best feedback if the scope was narrower.

libs/vault/src/vault-item-dialog/vault-item-dialog.component.html
mcamirault: Yes, with the condition that a user can only share an org-owned cipher if they have edit permission on at least one collection that the cipher belongs to.

libs/tools/share/src/services/share-link.service.ts
mcamirault: All item types besides SSH Keys will be shareable, so we'd have to include all those fields too at a minimum. It seemed simpler to include everything and trim the parts that we were specifically directed to exclude. Are there other cipher fields that you think definitely shouldn't be a part of this?

libs/vault/src/vault-item-dialog/vault-item-dialog.component.ts
mcamirault: The design we were given didn't have a desktop-specific image for this section, but web used a drawer and browser opened on a new page instead of a modal so I assumed that a drawer-like experience for all clients was intended. I'll confirm with the designer.

libs/tools/share/src/components/share-item-form/share-item-form.component.ts
mcamirault: The new validator also includes logic for restricting which domains the emails can belong to (which can be applied via Send Control policy). I agree it'd be good to centralize this, but I'd rather not add to the PR by adding the domain validation to the shared validator right now. OK to add this as a tech debt ticket somewhere?

libs/tools/share/src/components/share-item-form/share-item-form.component.ts
mcamirault: Fixed

libs/tools/share/src/components/share-item-form/share-item-form.component.ts
mcamirault: Fixed

libs/common/src/tools/send/models/view/send-item.view.ts
mcamirault: Fixed

libs/common/src/tools/send/models/domain/send-item.ts
mcamirault: Fixed

libs/common/src/tools/send/models/data/send-item.data.ts
mcamirault: Fixed

libs/common/src/tools/send/models/api/send-item.api.ts
mcamirault: Fixed

apps/web/src/app/admin-console/organizations/policies/policy-edit-definitions/send-controls.component.ts
mcamirault: I turned it into a computed signal so it plays nicer with the feature flag signal

apps/web/src/app/admin-console/organizations/policies/policy-edit-definitions/send-controls.component.ts
mcamirault: Done. The form field for the allowedSendTypes is technically being set twice, but it might not be for the reason you had in mind. Since the multi-select component outputs an array of full SelectItemView objects and our policy data is an array of SendType enums we have two different elements to the form here; one is separated from the rest of the form and is used to populate the multi-select component we display and the user interacts with (this is what the `effect` was setting) and the other is part of the actual form, has no display component, and is set to a value computed from the multi-select form control whenever its value changes (this is what `loadData` sets). Since we're setting both elements the latter does get updated twice, but since the effect took into account the policy data it should be set to the same thing both times.

apps/web/src/app/admin-console/organizations/policies/policy-edit-definitions/send-controls.component.ts
mcamirault: Done

apps/web/src/app/admin-console/organizations/policies/policy-edit-definitions/send-controls.component.html
mcamirault: Yep, changed it

apps/browser/src/tools/popup/share/share-item.component.ts
mcamirault: It needed to take the query parameters into account as well but I've made this change

mcamirault: Done

apps/web/src/app/vault/individual-vault/vault.component.ts
mcamirault: I added in the same `handleUnknownCipher()` call that other actions have

apps/desktop/src/vault/app/vault-v3/vault-items/vault-cipher-row.component.ts
mcamirault: Changed

libs/common/src/tools/send/models/view/send-item.view.ts
mcamirault: Fixed

libs/common/src/tools/send/services/send.service.ts
mcamirault: Done, the relevant SDK PR is up [here](https://github.com/bitwarden/sdk-internal/pull/1445)

libs/common/src/tools/send/models/data/send-item.data.ts
mcamirault: Fixed

libs/common/src/tools/send/models/api/send-item.api.ts
mcamirault: Fixed

libs/common/src/tools/send/models/domain/send-item.ts
mcamirault: Fixed

libs/common/src/tools/send/models/domain/send-item.ts
mcamirault: Decryption can now go through the SDK with the exposure of [two new functions](https://github.com/bitwarden/sdk-internal/pull/1445)

libs/angular/src/scss/bwicons/styles/style.css
```diff
@@ -94,393 +94,397 @@
 }
 
 .bwi-accessibility:before {
-  content: "\f178";
+  content: "\f179";
```
quexten, line 97: Is this large style diff intended? Could you comment on what it does?

libs/common/src/tools/send/services/send-sdk-decryption.service.ts
```diff
@@ -0,0 +1,41 @@
...
+          }
+          using ref = sdk.take();
+          const sendsClient = ref.value.sends();
+          return sendsClient.decrypt_send(send.toSdkSend());
+        }),
+      ),
+    );
+  }
+
+  async decryptSendAccess(sendAccess: SendAccessResponse, key: string): Promise<SdkSendAccessView> {
```
quexten: Nit (non-blocking): Probably want to use symmetric crypto key instead of string here. (Or the SDK's symmetric key).

libs/tools/share/src/services/share-link.service.ts
```diff
@@ -0,0 +1,208 @@
...
+    if (!sharedCipherView) {
+      throw new Error(this.i18nService.t("linkSaveFailed"));
+    }
+
+    // Strip attachments, passkeys, password history, and encryption key
+    sharedCipherView.attachments = [];
```
quexten: Nit (non-blocking): We should probably make a helper function such as "toShareView()" which is clearer than just doing these transforms inline.

libs/tools/share/src/services/share-link.service.ts
```diff
@@ -0,0 +1,208 @@
...
+    sendView.name = this.i18nService.t("itemSendTitle", cipherView.name);
+    sendView.type = SendType.Item;
+    sendView.authType = AuthType.Email;
+    sendView.emails = emails;
+    sendView.deletionDate = new Date(Date.now() + expiryHours * 60 * 60 * 1000);
+    sendView.expirationDate = sendView.deletionDate;
+    if (oneTimeShare) {
+      sendView.maxAccessCount = 1;
+    }
+    const sharedCipherView = CipherView.fromJSON(JSON.parse(JSON.stringify(cipherView)));
```
quexten: Nit: Moving this to a helper function CipherView.deepCopy() would be clearer to read.

quexten (dismissed):
Thanks for addressing the concerns. Approving for crypto.

djsmith85 (commented):
Hey @bitwarden/team-tools-dev I hope all of you are doing well.

**TLDR: Non-blocking but hopefully useful feedback to break up the work and speed up the delivery of item sharing.**

Due to the size and tagged teams, this is going to be hard to get approved by all. For QA a full regression on text/file sends due to the SDK changes will be required including all testing for the new item-sharing feature. This PR will stay blocked for a long time until QA completes and will likely have merge-conflicts due to the sheer amount of changed files.

May I suggest breaking this work up into smaller PRs?

Some parts I have identified and can likely be split into multiple PRs. Most of the PRs can be opened, reviewed, and merged in parallel and will not stack/rely on each other.

### Tools
- ~~Introducing the `PM34203TemporaryItemSharing`-flag. Tiny PR and can be used by any following feature PR where needed.~~ - Completed with https://github.com/bitwarden/clients/pull/22921

- ~~Split the addition of `FeatureFlag.Pm30110SdkSendsApi` to route all Send-logic through the SDK. This also affects text- and file Sends, which can be separately tested~~ - Completed with https://github.com/bitwarden/clients/pull/22923

- Extending view-models and responses (`libs/common/src/tools/send/models/*`) can be merged into main without any code using it. `Item.data` when used for item sharing will be separately tested by QA - WIP with https://github.com/bitwarden/clients/pull/23060

- Changes to the SendAccess-page, none of the new components (`SendAccessItemFieldComponent` and `SendAccessItemComponent`) will be visible to the user, as they won't be able to create ItemShares yet. Can be reviewed and merged and will be tested by QA once the create share actions are available

### Tools/Platform
- ~~`libs/tools/share` can be its own PR, as it introduces the library, including all the monorepo wiring~~
  - ~~As this new functionality is not used yet, it can be merged without testing and only requires approval by @bitwarden/team-tools-dev~~ 
  - Completed with https://github.com/bitwarden/clients/pull/22926 

### Vault/AC
- Creating separate components owned by tools, rather than extending @bitwarden/team-vault-dev owned code, besides the markup.

  The driving idea here is encapsulation, and in case the logic to hide/show the menu item or the action changes, none of the Vault/AC components need to change, and the code only changes in one place.
  Think about what happens when your side of the code needs to change and how that affects the current classes/components. Adding logging and future policies that restrict UX, or once PAM is in place, the ShareLinkService might need to be replaced by a generic PAM service. 

  - `apps/web/src/app/vault/components/vault-items/vault-cipher-row.component.ts`
  Create a `ShareViaLinkMenuItem`-component which receives the Cipher(Id) and uses their own ShareLinkService to figure out if the the MenuItem is shown/disabled. It also takes care of whatever needs to happen on click (emitting an event or opening a dialog). This makes it a smaller change in Vault and AC code by only importing the new component and adding the menu item to the markup without the additional logic and ShareLinkService.
  - `apps/desktop/src/vault/app/vault-v3/vault-items/vault-cipher-row.component.ts` The same menu item component introduced above can be re-used here
  - The same pattern can be applied to `libs/vault/src/vault-item-dialog/vault-item-dialog.component.ts` which will create the `ShareViaLinkButton`-component and besides handling enabling/disabling the button, will take care of opening the dialog/drawer when clicked.

  - Also `apps/browser/src/vault/popup/components/vault/item-more-options/item-more-options.component.ts`

Use the `PM34203TemporaryItemSharing`flag within the share actions(button/menu item) to not show themselves. The removal of the feature flag will not require another teams review.

### UI Foundation
- ~~Adding the new SVG/icon, extending icon.ts, fonts, and non-used styles can be a small targeted PR, reviewed by @bitwarden/team-ui-foundation and merged into main without consequences.~~  - Completed with https://github.com/bitwarden/clients/pull/22934

In case it seems useful to further break up the work, either for review or QA purposes, introduce the new feature per client. All the actions to create a item share will be guarded by the feature flag.

**These are just some ideas and guidance, and none of this feedback is required to unblock this PR if you decide to keep it.**

libs/tools/share/src/components/share-item-form/share-item-form.component.ts
BTreston: Perfectly fine, thanks!

libs/common/src/tools/send/services/send.service.ts
```diff
@@ -393,7 +410,16 @@ export class SendService implements InternalSendServiceAbstraction {
...
+    const useSdkForSends = await this.configService.getFeatureFlag(FeatureFlag.Pm30110SdkSendsApi);
+    const decryptSendPromises = sends.map((s) => {
+      if (useSdkForSends) {
+        return this.sendSdkDecryptionService
+          .decryptSend(s, userId)
+          .then((s) => SendView.fromSdkSend(s));
+      } else {
+        return s.decrypt(userId);
+      }
+    });
```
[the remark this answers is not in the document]
mcamirault: We won't be turning this feature on until that feature flag is enabled

libs/angular/src/scss/bwicons/styles/style.css
mcamirault: We're adding a new icon and since it got added into the middle of the list it shifted everything after it up by one

libs/tools/share/src/services/share-link.service.ts
mcamirault: Moved to the `toShareView` helper function with a comment

libs/tools/share/src/services/share-link.service.ts
mcamirault: Added helper function

libs/common/src/tools/send/services/send-sdk-decryption.service.ts
mcamirault: Changed

shane-melton:
I just want to echo @djsmith85's [comment above](https://github.com/bitwarden/clients/pull/22159#pullrequestreview-5088277170). 

Vault would strongly prefer this be broken up with the share logic encapsulated in Tools owned component(s) as described above. This will allow you all to make future adjustments as necessary without requiring us to review/approve everything and have to try and understand the full scope of the feature.

Also, I wanted to callout out that it doesn't appear any of the changes have been applied to the new VFO1 work vault is working for the re-design. Many of these table/row changes will be lost on VFO1's release.

libs/vault/src/vault-item-dialog/vault-item-dialog.component.ts
mcamirault: Confirmed, they want a drawer on desktop as well

mcamirault:
The breakup of work from this PR is now complete, with all outstanding work contained in [these](https://github.com/bitwarden/clients/pull/23060) [three](https://github.com/bitwarden/clients/pull/23079) [PRs](https://github.com/bitwarden/clients/pull/23152); major thanks to @djsmith85 for his help breaking this down and to everyone else who reviewed this PR originally. Closing now.

PR #22161, opened 2026-07-30
introduce `AutofillResult`

audreyality:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-41230

This work was split from https://bitwarden.atlassian.net/browse/PM-39579 to ease review.

## 📔 Objective

`doAutoFill` returned `string | null` and overloaded exceptions to mean "failure to fill" and "an unexpected error occurred". Catching errors and continuing works well with imperative code, but poorly with streaming code. An uncaught error tears down the stream early. 

The `AutoFillResult` union separates the two concerns: `didAutofill` reports whether a fill happened, and totp reports whether there was a code to copy. When an error is thrown, something unexpected occurred. This allows streams to divert control cleanly, and is often faster than stack unwinding for imperative code.

## 🧰 New Approach

`AutofillService.doAutoFill` previously signaled "nothing was filled" by throwing. It now returns a discriminated union instead:

```ts
export type AutoFillResult =
  | { didAutofill: false }
  | { didAutofill: true; totp?: string };
```

This type splits the autofill flag from totp while ensuring that callers that need to access the totp (which is most of them) confirm that autofill succeeded before accessing it.

`doAutoFill`, `doAutoFillOnTab`, and `doAutoFillActiveTab` all return `Promise<AutoFillResult>`. Genuine mid-fill errors still reject the promise. Only "no fill happened" became a value.


apps/browser/src/vault/popup/services/vault-popup-autofill.service.spec.ts
audreyality, line 1: 📝 Testing `doAutofill` requires mock resets, but this file contained tests that relied upon the state of prior tests. Those tests were fixed.

audreyality (commented):
📝 So much of this is just changing return values. 😅

apps/browser/src/autofill/background/overlay.background.ts
```diff
@@ -1477,8 +1477,8 @@ export class OverlayBackground implements OverlayBackgroundInterface {
       inlineMenuFillType: this.focusedFieldData?.inlineMenuFillType,
     });
 
-    if (totpCode) {
-      this.platformUtilsService.copyToClipboard(totpCode);
+    if (result.didAutofill && result.totp) {
+      this.platformUtilsService.copyToClipboard(result.totp);
     }
 
     this.updateLastUsedInlineMenuCipher(inlineMenuCipherId, cipher);
```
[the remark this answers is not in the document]
audreyality: It's not. I spotted that, too, and thought it was fixed. Perhaps it was a different callsite. 🤦🏻‍♀️ 

Addressed by d1afc1c7b77edfe58.

apps/browser/src/autofill/services/autofill.service.spec.ts
```diff
@@ -1083,7 +1035,7 @@ describe("AutofillService", () => {
 
       const autofillResult = await autofillService.doAutoFill(autofillOptions);
 
-      expect(autofillResult).toBeNull();
+      expect(autofillResult).toEqual({ didAutofill: true });
```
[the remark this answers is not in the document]
audreyality: Addressed by 54175414f07af993aa5be0ece44.

apps/browser/src/vault/popup/services/vault-popup-autofill.service.ts
```diff
@@ -307,22 +307,36 @@ export class VaultPopupAutofillService {
         allowTotpAutofill: true,
       });
 
-      if (totpCode != null) {
-        this.platformUtilService.copyToClipboard(totpCode, { window: window });
+      if (!result.didAutofill) {
+        this._reportAutofillFailure();
+        return false;
```
jprusik, line 312: This will result in new behaviour where if autofill fails, the TOTP will not be copied to the clipboard (where it did before)? I don't have issue with this, just want to call it out as a potential difference in the UX.

audreyality: The original threw an exception which was handled in the `catch` and skipped the TOTP copy. That branch is preserving the original behavior. :)

jprusik (approved):

jengstrom-bw (approved):
vault items look good

bensbits91 (approved):

audreyality:
📝 The lint failure is a `cargo deny` check that's unrelated to this PR.

PR #22168, opened 2026-07-31
feat(platform): let an SdkRecordMapper exclude records from SDK-visible state

Hinton:
## 🎟️ Tracking

Prerequisite for the PAM cipher-gating stack (#22169 → #22170 → #22171). Small and
generic — no PAM knowledge needed to review.

## 📔 Objective

Adds an optional `shouldInclude` predicate to `SdkRecordMapper`. When it returns false for
a record, `RepositoryRecord.get` returns `null` and `list()` omits it, so the SDK never has
to parse a record it cannot represent.

Writes are deliberately **unfiltered**: the record still round-trips through client state
and stays readable via the client's own services. Only the SDK-facing view is narrowed.

The consumer is #22169, where PAM-gated ciphers are hidden from the SDK — it has no
partial-data decrypt path and rejects sparsely-populated ciphers with a serde
`invalid type: unit value` error.

## Verification

`npm test -- libs/common/src/platform/services/sdk/client-managed-state.spec.ts` — 40
passed (35 existing + 5 new, covering both `get`/`list` filtering, the no-predicate
default, and that writes still land).

## 🚨 Breaking Changes

None — the predicate is optional and existing mappers include everything, exactly as before.

### DIFF
--- a/libs/common/src/platform/services/sdk/client-managed-state.spec.ts
+++ b/libs/common/src/platform/services/sdk/client-managed-state.spec.ts
@@ -91,6 +91,63 @@ describe("RepositoryRecord", () => {
     });
   });
 
+  describe("shouldInclude filtering", () => {
+    // `excluded:`-prefixed values stand in for records the SDK cannot represent.
+    function createFilteringMapper(): SdkRecordMapper<ClientType, SdkType> {
+      return {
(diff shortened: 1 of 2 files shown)


Hinton:

PR #22169, opened 2026-07-31
feat(vault): plumb the PAM partialData gating marker through the cipher model

Hinton:
## 🎟️ Tracking

PAM cipher gating, 2/4. Stacked on #22168 (needs its `shouldInclude` hook).
Server side: bitwarden/server#8114 (field) and #8115 (behavior).

## 📔 Objective

The server withholds a PAM-gated cipher's sensitive fields and returns a reduced blob
instead — the encrypted name plus, for logins, the encrypted URIs. This carries that blob
from the wire down to `CipherView`, where UI surfaces can use its presence as the "this row
is gated" signal.

- **`CipherResponse`** reads `PartialData` (accepting a string or an already-parsed object)
  and lifts the encrypted `Name` onto `name` so the normal decrypt path handles it. For
  logins it also lifts `Uris`, building the login from **`Uris` alone** — a deliberate
  allowlist, so an over-sharing blob can never leak a password or TOTP onto a gated view.
  A malformed blob keeps the marker and drops the name: failing to parse must not un-gate
  the row.
- The marker is carried through `CipherData` → `Cipher` → `CipherView`.
- **`CipherRecordMapper.shouldInclude`** hides gated rows from the SDK (no partial-data
  decrypt path; it rejects sparse ciphers). They stay in client state.
- **`DefaultCipherEncryptionService`** re-attaches the marker after the SDK round trip
  drops it, on both the single and bulk decrypt paths.

### Note for reviewers

This is **additive to the existing `data` field**, which carries the full blob-encrypted
payload (PM-32696). The two are unrelated and both are kept.

Also adds `leaseGated`, the transient companion marker for a cipher served under active
access. It is deliberately never persisted to `CipherData` nor serialized in `fromJSON`,
and has **no producer yet** — declared here because the cipher-view banner seam (#22171)
forwards it.

## Verification

`npm test -- libs/common/src/vault libs/common/src/platform/services/sdk` — 784 passed
across 47 suites. `npm run test:types` clean.

New `cipher.response.spec.ts` covers the name/URI lift, that no secret login field is ever
lifted, the malformed-blob and already-parsed-object paths, and that a real top-level `Name`
wins over the blob's.

## 🚨 Breaking Changes

None — purely additive. With no server sending `PartialData`, `partialData` is always
absent and every path behaves exactly as before.

### DIFF
--- a/libs/common/src/vault/models/data/cipher.data.ts
+++ b/libs/common/src/vault/models/data/cipher.data.ts
@@ -48,6 +48,12 @@ export class CipherData {
   reprompt: CipherRepromptType = CipherRepromptType.None;
   key?: string;
   data?: string;
+  /**
+   * Raw JSON-string partial-data payload for PAM-gated rows. See {@link CipherResponse.partialData}.
+   * Widened to allow `null` so a `CipherResponse` (whose `partialData` is `string | null`)
+   * stays structurally assignable to `CipherData`.
+   */
+  partialData?: string | null;
 
   constructor(response?: CipherResponse, collectionIds?: string[]) {
     if (response == null) {
@@ -74,6 +80,7 @@ export class CipherData {
     this.reprompt = response.reprompt;
     this.key = response.key;
     this.data = response.data;
+    this.partialData = response.partialData ?? undefined;
 
     switch (this.type) {
       case CipherType.Login:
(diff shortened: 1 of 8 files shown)


Hinton:

PR #22170, opened 2026-07-31
feat(vault): decrypt PAM gated rows outside the SDK

Hinton:
## 🎟️ Tracking

PAM cipher gating, 3/4. Stacked on #22169.

## 📔 Objective

`decryptCiphersWithSdk` now partitions gated rows out before calling the SDK and merges the
results back in, so gated ciphers still appear in the vault list. Without this they would
**vanish**: the SDK has no partial-data decrypt path and rejects sparsely-populated ciphers.

The encrypted name has already been lifted out of the partial blob onto `cipher.name`
(#22169), so the standard decrypt path handles it; every other field is null, so the
type-specific branches short-circuit and the view ends up with a decrypted name (plus login
URIs) and nothing else.

When no key is available for the cipher's scope the row renders with an **empty name rather
than being dropped** — silently hiding a vault item is worse than showing it nameless.

## ⚠️ Reviewer note: this uses a deprecated API

This becomes the only caller of `Cipher.decrypt`, which `main` marks
`@deprecated … will be removed in a near release`. I'd like a second opinion, but the case
for it:

- The deprecation is about **blob encryption**, and a gated cipher never carries a blob —
  the server only strips ciphers whose data is *not* blob-encrypted.
- Going through the SDK is precisely what fails here.
- Hand-decrypting the fields would mean duplicating the **login-URI checksum validation**,
  which guards against a compromised server tampering with URIs. Reimplementing that
  security-sensitive logic seemed clearly worse than using the deprecated path.

If `Cipher.decrypt` is removed before the real gate ships, this needs a replacement — most
likely an SDK-side partial-decrypt path, which would also let #22169 drop its
`shouldInclude` filter.

## Verification

`npm test -- libs/common/src/vault` — 731 passed. `npm run test:types` clean.

This path had **no test coverage in the reference implementation**; added tests for
partitioning on both the full and lightweight paths, the merge, the missing-key fallback,
and that no key fetch happens when nothing is gated.

## 🚨 Breaking Changes

None — with no gated ciphers present, the partition is empty and behavior is unchanged.

### DIFF
--- a/libs/common/src/vault/services/cipher.service.ts
+++ b/libs/common/src/vault/services/cipher.service.ts
@@ -2164,16 +2164,25 @@ export class CipherService implements CipherServiceAbstraction {
       return [[], []];
     }
 
+    // PAM gated rows can't go through the SDK (no partial-data decrypt path), so decrypt
+    // them separately and merge the results back in, keeping them in the vault list.
+    const partialCiphers = ciphers.filter((c) => c.partialData != null);
+    const decryptableCiphers = ciphers.filter((c) => c.partialData == null);
+    const partialViews = await this.decryptPartialCiphers(partialCiphers, userId);
+
     if (fullDecryption) {
       const [decryptedViews, failedViews] = await this.cipherEncryptionService.decryptManyLegacy(
-        ciphers,
+        decryptableCiphers,
         userId,
       );
-      return [decryptedViews.sort(this.getLocaleSortingFunction()), failedViews];
+      return [
+        [...partialViews, ...decryptedViews].sort(this.getLocaleSortingFunction()),
+        failedViews,
+      ];
     }
 
     const [decrypted, failures] = await this.cipherEncryptionService.decryptManyWithFailures(
-      ciphers,
+      decryptableCiphers,
       userId,
     );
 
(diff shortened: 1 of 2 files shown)


libs/common/src/vault/services/cipher.service.ts
```diff
@@ -2184,7 +2193,48 @@ export class CipherService implements CipherServiceAbstraction {
...
+
+        if (key == null) {
+          // No key for this cipher's scope. Render the row with an empty name rather than
+          // dropping it — the gating marker is preserved either way.
+          const view = new CipherView(cipher);
+          view.name = "";
+          return view;
+        }
+
+        return cipher.decrypt(key);
```
quexten, line 2235: https://github.com/bitwarden/clients/blob/ee4f0abbc74367d34af5159340c0135570d66f5d/libs/common/src/vault/models/domain/cipher.ts#L145

Please do not add callers to this function. It is deprecated and should not be used. We want to remove calls to it, and remove the implementation imminently.

quexten: A temporary solution would be to expose a decrypt_with_key function in the sdk which would already be better than adding a caller to this deprecated function.

quexten (changes requested):
It's still in draft, but putting this early because I may forget to loop back around to the PR.

libs/common/src/vault/services/cipher.service.ts
Hinton: Agreed, we would probably lift this into the SDK somehow.

Hinton:

Issue #1359, opened 2020-08-14
Trying to get the status code.


lmk07:
Hi guys,
            i'm trying to fetch the details from network activity in browser using javascript, i need to get the status code of each and every resource while page loading and interaction happens in page. Any one can help me to get this... Thanks in advance.

![Capture](https://user-images.githubusercontent.com/66107540/90236045-d836d700-de3f-11ea-9fbb-0a522a0cf3b1.JPG)

Greenderella:
Hi, could you help us understand why do you need those status codes?

Greenderella:
If you are trying to get the network details in your browser regarding the extension itself, you will need to go to chrome://extensions and turn on Developer Mode:

![image](https://user-images.githubusercontent.com/5418577/90269063-4edfce80-de58-11ea-891d-d28adcc89779.png)

Then you will see a link to inspect the background.html page, which can lead them to the network traffic/communication and JS world of our browser extension.

lmk07:
@Greenderella  Hi, i want it as javascript code, which should work in all the browser. Thank you.

cscharf:
Hi @lmk07 , hopefully that gets you the information you need. For now I'm going to close this issue as it looks like Alex has answered your question. Please let us know if you run into any issues or feel free to re-open this if necessary.

PR #22171, opened 2026-07-31
feat(vault): route PAM gated ciphers through the SDK (partial-cipher pivot)

Hinton:
## 🎟️ Tracking

Strategy pivot for PAM cipher gating. **Supersedes #22168, #22169, #22170** (now closed). Depends on **bitwarden/sdk-internal#1359**.

> ⚠️ **Draft / blocked on the SDK:** consumes types (`Cipher.partialData`, the `partial` view flag, and the `RestrictedCipherData` field allowlist) from sdk-internal#1359. The `@bitwarden/sdk-internal` version bump lands once that publishes; until then CI typecheck fails on three lines (`cipher.ts` `toSdkCipher`/`fromSdkCipher`, `cipher.view.ts` `fromSdkCipherView`). Reviewable now against a local WASM link. The branch is already squashed to a single commit and rebased onto latest `main`; the SDK version bump is the only remaining pre-merge step.

## 📔 Objective

Route **PAM-gated ("partial") ciphers** through the SDK and surface them **only in the web vault**, where they render read-only with a "Controlled access" badge. A partial cipher is a row whose sensitive fields the server suppressed (the caller lacks a privileged-access lease); it ships a reduced `partialData` envelope (encrypted name + login URIs only) in place of the full payload, and the SDK decrypts it into a view marked `partial`.

Everywhere outside the web vault list, partials are excluded — they must never reach autofill, export, reports, Fido2, key rotation, or the CLI, and must never be modifiable (editing/re-encrypting a gated row would clobber the server-suppressed fields with blanks).

The security-critical field allowlist (a gated view exposes only name + login URIs, never password/TOTP) lives in the SDK (`RestrictedCipherData` in #1359), not in this repo.

### What changed

**Model plumbing.** `CipherResponse.partialData` (optional) flows verbatim → `CipherData` → domain `Cipher.partialData`, round-tripping losslessly through `toSdkCipher`/`fromSdkCipher`. The SDK decrypts gated rows like any other cipher (no client-side partition, no `SdkRecordMapper.shouldInclude` filter, no deprecated `Cipher.decrypt`) and stamps `partial: boolean` on the decrypted `CipherView`/`CipherListView`.

**Web-vault-only, excluded-by-default streams.** `CipherService` splits its decrypted streams so partials are excluded by default and only the web list opts in:

- `getAllDecrypted()` excludes partials at the root (a private `getAllDecryptedIncludingPartials()` retains them solely as the shared decrypt source), covering every imperative consumer — export, reports, Fido2, ssh-agent, autofill card/identity suggestions, key rotation, CLI.
- `cipherViews$` / `cipherListViews$` exclude partials, derived from private partials-inclusive sources so decryption stays single-pass.
- The web vault list consumes the new opt-in `cipherListViewsWithPartials$`.
- `CipherViewLikeUtils.isPartial` centralizes the flag read.

**Read-only rendering + badge seams.** A gated cipher opens read-only in the vault-item dialog (Edit hidden, form-mode redirected to view) and cipher-view, where a host can render an optional banner through the `CIPHER_VIEW_BANNER` seam (handed the `CipherView` directly). A host-provided **"Controlled access" badge column** (`VAULT_ROW_LEASE_BADGE` seam) appears only when the `Pam` feature flag is on, a PAM-enabled org (`Organization.usePam`) is in view, **and** a host provides the badge — otherwise the table is unchanged.

**Non-modifiable in web.** Gated rows are non-selectable (checkbox disabled + excluded from `editableItems`/select-all) and expose no modify menu actions; both bulk-action paths (classic emitters + `VaultBatchBarService`) filter them defensively; encrypted export excludes them via the `partialData` marker.

**Filter tree.** The web filter (type filter + folder tree, via a web-only `VaultFilterService` subclass) includes partials so a type/folder present only as gated rows still surfaces; other clients keep the excluded stream. Collections are unaffected.

### Verification

`test:types`, `lint`, and prettier are clean on the affected projects (save the three SDK-blocked lines above). Touched + model specs pass, including new coverage for the stream split/exclusion, `isPartial`, the read-only rendering, the badge column gate, non-selectability, bulk-action + export exclusion, and key-rotation exclusion. E2E stays a no-op until a real PAM provider binds the badge seam and the server emits `partialData`.

## 🚨 Breaking Changes

None — inert until a PAM provider binds the badge seam and the server emits `partialData`. With PAM off, no cipher is `partial`, so every added filter is a no-op and behavior is unchanged.

### DIFF
--- a/apps/web/src/app/vault/components/vault-items/vault-items.component.ts
+++ b/apps/web/src/app/vault/components/vault-items/vault-items.component.ts
@@ -1,8 +1,8 @@
 // FIXME: Update this file to be type safe and remove this and next line
 // @ts-strict-ignore
 import { SelectionModel } from "@angular/cdk/collections";
-import { Component, EventEmitter, Input, Output, inject } from "@angular/core";
-import { takeUntilDestroyed } from "@angular/core/rxjs-interop";
+import { Component, EventEmitter, Input, Output, Signal, inject } from "@angular/core";
+import { takeUntilDestroyed, toSignal } from "@angular/core/rxjs-interop";
 import {
   Observable,
   combineLatest,
@@ -47,6 +47,7 @@ import {
   convertToPermission,
 } from "./../../../admin-console/organizations/shared/components/access-selector/access-selector.models";
 import { VaultItemEvent } from "./vault-item-event";
+import { VAULT_ROW_LEASE_BADGE } from "./vault-row-lease-badge.token";
 
 // Fixed manual row height required due to how cdk-virtual-scroll works
 export const RowHeight = 76.5;
@@ -161,6 +162,13 @@ export class VaultItemsComponent<C extends CipherViewLike> {
     optional: true,
   }) as VaultBatchBarService<C> | null;
 
+  /**
+   * Host-provided "Controlled access" badge seam. Its presence (a privileged-access feature is
+   * installed) is what surfaces the Controlled access column; unprovided, the column is absent
+   * and the table is unchanged.
+   */
+  protected readonly leaseBadge = inject(VAULT_ROW_LEASE_BADGE, { optional: true });
+
   protected editableItems: VaultItem<C>[] = [];
   protected dataSource = new TableDataSource<VaultItem<C>>();
   private readonly _localSelection = new SelectionModel<VaultItem<C>>(
@@ -178,6 +186,7 @@ export class VaultItemsComponent<C extends CipherViewLike> {
   protected showCopyAndLaunchActions$: Observable<boolean>;
   protected showQuickCopyActions$: Observable<boolean>;
   private restrictedTypes: RestrictedCipherType[] = [];
+  private readonly pamEnabled: Signal<boolean>;
 
   private readonly vaultCopyButtonsService = inject(VaultCopyButtonsService);
 
@@ -195,6 +204,10 @@ export class VaultItemsComponent<C extends CipherViewLike> {
       this.configService.getFeatureFlag$(FeatureFlag.PM40435_QuickCopyIconSetting),
       this.vaultCopyButtonsService.showQuickCopyActions$,
     ]).pipe(map(([flagEnabled, settingEnabled]) => flagEnabled && settingEnabled));
+
+    this.pamEnabled = toSignal(this.configService.getFeatureFlag$(FeatureFlag.Pam), {
+      initialValue: false,
+    });
     this.canDeleteSelected$ = this.selection.changed.pipe(
       startWith(null),
       switchMap(() => {
@@ -286,7 +299,19 @@ export class VaultItemsComponent<C extends CipherViewLike> {
   }
 
   get showExtraColumn() {
-    return this.showCollections || this.showGroups || this.showOwner;
+    return this.showCollections || this.showGroups || this.showOwner || this.showControlledAccess;
+  }
+
+  /**
+   * Whether to render the "Controlled access" column. Shown only when the PAM feature flag is
+   * enabled, the viewer actually has PAM enabled — i.e. at least one organization in view has the
+   * Privileged Access capability (`usePam`) — and a host provides the badge seam. Otherwise the
+   * column is absent and the table is unchanged.
+   */
+  get showControlledAccess() {
+    return (
+      this.pamEnabled() && this.leaseBadge != null && this.allOrganizations.some((o) => o.usePam)
+    );
   }
 
   /**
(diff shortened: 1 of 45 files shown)


apps/web/src/app/vault/components/vault-items/vault-items.component.ts
```diff
@@ -277,7 +285,17 @@ export class VaultItemsComponent<C extends CipherViewLike> {
...
+  }
+
+  /**
+   * Whether to render the "Controlled access" column. Shown only when the viewer actually has
+   * PAM enabled — i.e. at least one organization in view has the Privileged Access capability
+   * (`usePam`) — and a host provides the badge seam. Without a PAM-enabled org the column is
+   * absent and the table is unchanged.
+   */
+  get showControlledAccess() {
+    return this.leaseBadge != null && this.allOrganizations.some((o) => o.usePam);
```
Hinton: I think we need to check the feature flag too.

apps/web/src/app/vault/individual-vault/cipher-open-gate.ts
Hinton, line 1: I think it would be nice to strip out the open gate for this PR. We can do that in a follow up.

libs/common/src/platform/services/sdk/client-managed-state.ts
```diff
@@ -67,7 +67,10 @@ export class RepositoryRecord<ClientType, SdkType> implements Repository<SdkType
   async get(id: string): Promise<SdkType | null> {
     const record = await this.getRecord();
     const element = record[id];
-    return element ? this.mapper.toSdk(element) : null;
+    if (!element) {
+      return null;
+    }
+    return this.mapper.toSdk(element);
```
Hinton: Why this change?

libs/common/src/vault/models/data/cipher.data.ts
```diff
@@ -48,6 +48,12 @@ export class CipherData {
   reprompt: CipherRepromptType = CipherRepromptType.None;
   key?: string;
   data?: string;
+  /**
+   * Raw JSON-string partial-data payload for PAM-gated rows. See {@link CipherResponse.partialData}.
+   * Widened to allow `null` so a `CipherResponse` (whose `partialData` is `string | null`)
+   * stays structurally assignable to `CipherData`.
+   */
+  partialData?: string | null;
```
Hinton: Can we not do ` ?? undefined`?

libs/common/src/vault/models/response/cipher.response.ts
```diff
@@ -53,6 +53,14 @@ export class CipherResponse extends BaseResponse {
...
   key: string;
   data?: string;
+  /**
+   * Raw JSON-string payload the server returns on PAM-gated rows in place of the
+   * sensitive fields: the encrypted name and, for logins, the encrypted URIs.
+   * Its presence is the "this row is gated" marker used by the vault-row badge
+   * and the cipher-open gate. Distinct from {@link data}, which carries the full
+   * (blob-encrypted) payload for ciphers the caller may open.
+   */
+  partialData: string | null = null;
```
Hinton: Can we not do `?` rather than use null?

libs/vault/src/tokens/gated-cipher-reloader.token.ts
Hinton, line 1: I think we can drop this too.

apps/web/src/app/vault/components/vault-items/vault-items.component.ts
Hinton: Done in 2986059 — `showControlledAccess` now also gates on `FeatureFlag.Pam` (alongside the badge seam and a `usePam` org).

apps/web/src/app/vault/individual-vault/cipher-open-gate.ts
Hinton: Done in 2986059 — stripped the open gate (file + wiring in vault.component.ts); left as a follow-up.

libs/common/src/platform/services/sdk/client-managed-state.ts
Hinton: Reverted in 2986059 — `RepositoryRecord.get` is back to the original ternary; that change wasn't needed.

libs/common/src/vault/models/data/cipher.data.ts
Hinton: Done in 2986059 — direct assignment now that `partialData` is `string | undefined` on both models (also removed the two matching `?? undefined` in cipher.ts).

libs/common/src/vault/models/response/cipher.response.ts
Hinton: Done in 2986059 — `partialData?: string` instead of `string | null = null`.

libs/vault/src/tokens/gated-cipher-reloader.token.ts
Hinton: Done in 2986059 — dropped the reloader token, its index export, and the vault-item-dialog wiring; the read-only `partial` rendering seam stays.

libs/vault/src/cipher-view/cipher-view.component.html
```diff
@@ -1,4 +1,17 @@
 @if (!!cipher()) {
+  @if (bannerComponent) {
+    <ng-container
+      *ngComponentOutlet="
+        bannerComponent;
+        inputs: {
+          cipherId: cipher().id,
+          partial: cipher().partial,
+          leaseGated: cipher().leaseGated,
```
Hinton: We should just provide the cipher directly.

libs/common/src/vault/services/cipher.service.ts
```diff
@@ -139,19 +139,63 @@ export class CipherService implements CipherServiceAbstraction {
...
 
   /**
-   * Observable that emits an array of decrypted ciphers for given userId.
-   * This observable will not emit until the encrypted ciphers have either been loaded from state or after sync.
+   * Emits the fully decrypted views for the given user, INCLUDING PAM-gated ("partial") rows.
+   * Does not emit until the encrypted ciphers have loaded from state or after sync.
    *
-   * This uses the SDK for decryption, when the `PM22134SdkCipherListView` feature flag is disabled the full `cipherViews$` observable will be emitted.
-   * Usage of the {@link CipherViewLike} type is recommended to ensure both `CipherView` and `CipherListView` are supported.
+   * Private source for {@link cipherViews$} and the non-SDK fallback of
```
Hinton: This reads as if we used it without filtering.

libs/common/src/vault/services/cipher.service.ts
```diff
@@ -139,19 +139,63 @@ export class CipherService implements CipherServiceAbstraction {
...
+    },
+    this.clearCipherViewsForUser$,
+  );
+
+  /**
+   * Emits the fully decrypted views for the given user, with PAM-gated ("partial") rows EXCLUDED.
+   * The default full-view stream for every consumer. Derived from {@link cipherViewsWithPartials$},
+   * so a gated cipher never reaches autofill, export, key rotation, and similar flows.
+   *
+   * A `null` value indicates decryption is in progress; the decrypted views follow once complete.
```
Hinton: I think we should include the information this previously contained.

libs/common/src/vault/services/cipher.service.ts
```diff
@@ -189,26 +233,30 @@ export class CipherService implements CipherServiceAbstraction {
...
 
   /**
-   * Observable that emits an array of decrypted ciphers for the active user.
-   * This observable will not emit until the encrypted ciphers have either been loaded from state or after sync.
-   *
-   * A `null` value indicates that the latest encrypted ciphers have not been decrypted yet and that
-   * decryption is in progress. The latest decrypted ciphers will be emitted once decryption is complete.
+   * Emits the decrypted list views for the given user, with PAM-gated ("partial") rows EXCLUDED.
+   * The default list-view stream for every consumer except the vault list. Derived from
+   * {@link cipherListViewsWithPartials$}, so decryption runs once no matter which is subscribed.
```
Hinton, line 248: This should include the information it had before.

libs/common/src/vault/models/domain/cipher.ts
Hinton, line 1: Some of the added comments in this file seems somewhat obvious.

libs/common/src/vault/services/cipher.service.ts
```diff
@@ -325,11 +386,26 @@ export class CipherService implements CipherServiceAbstraction {
...
+   * returned.
+   *
+   * Partials are filtered here so a gated cipher never reaches the imperative consumers of this
+   * accessor (export, reports, autofill, Fido2, key rotation, CLI). The vault list gets partials
+   * via {@link cipherListViewsWithPartials$}.
    * @deprecated Use `cipherViews$` observable instead
    */
   async getAllDecrypted(userId: UserId): Promise<CipherView[]> {
+    const ciphers = await this.getAllDecryptedIncludingPartials(userId);
+    return ciphers.filter((cipher) => !CipherViewLikeUtils.isPartial(cipher));
```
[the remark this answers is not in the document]
Hinton: This is already i place in the SDK.

PR #22177, opened 2026-07-31
desktop-autofill: Verify user without UI on assertion when possible [PM-29791]

iinuwa:
## 🎟️ Tracking

## 📔 Objective

This implements the business for native OS user verification for passkeys.

- Defines business logic for user verification, centralizing logic for verifying via master password reprompt, vault unlock method or OS native user verification
- Allows keeping the Bitwarden window hidden on assertion if 1 passkey is selected and the vault is unlocked
- If the vault is unlocked for the passkey ceremony, the user is not re-prompted for user verification
- Adds a user verification service abstraction that each OS will implement/override.

For now, there is a no-op UV implementation that always returns false. macOS and Windows implementations will come in subsequent PRs.

### DIFF
--- a/apps/desktop/src/autofill/services/desktop-fido2-user-interface.service.ts
+++ b/apps/desktop/src/autofill/services/desktop-fido2-user-interface.service.ts
@@ -199,42 +199,41 @@ export class DesktopFido2UserInterfaceSession implements Fido2UserInterfaceSessi
   private uiShown = false;
 
   // Method implementation
-  async pickCredential({
-    cipherIds,
-    userVerification,
-    assumeUserPresence,
-    masterPasswordRepromptRequired,
-  }: PickCredentialParams): Promise<{ cipherId: string | undefined; userVerified: boolean }> {
-    this.logService.debug("pickCredential desktop function", {
-      cipherIds,
-      userVerification,
-      assumeUserPresence,
-      masterPasswordRepromptRequired,
-    });
+  async pickCredential(
+    params: PickCredentialParams,
+  ): Promise<{ cipherId: string | undefined; userVerified: boolean }> {
+    this.logService.debug("pickCredential desktop function", params);
 
     try {
-      // Check if we can return the credential without user interaction
-      if (assumeUserPresence && cipherIds.length === 1 && !masterPasswordRepromptRequired) {
-        this.logService.debug(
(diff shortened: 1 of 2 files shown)


libs/common/src/platform/services/fido2/fido2-authenticator.service.ts
iinuwa, line 1: I plan to address this in a follow-up for the desktop implementation, but I haven't checked whether changing this to InvalidState would affect the browser, so I'm leaving that for a separate PR.

libs/common/src/platform/abstractions/fido2/fido2-user-interface.service.abstraction.ts
```diff
@@ -84,7 +84,7 @@ export abstract class Fido2UserInterfaceSession {
    */
   abstract pickCredential(
     params: PickCredentialParams,
-  ): Promise<{ cipherId: string; userVerified: boolean }>;
+  ): Promise<{ cipherId: string | undefined; userVerified: boolean }>;
```
iinuwa, line 87: This change allows strict type checking to pass on the desktop implementation by conforming to the `@return` documentation.

apps/desktop/src/autofill/services/desktop-fido2-user-interface.service.ts
```diff
@@ -268,6 +260,84 @@ export class DesktopFido2UserInterfaceSession implements Fido2UserInterfaceSessi
...
+    } else if (params.assumeUserPresence) {
+      this.logService.debug(
+        "[DesktopFido2UserInterfaceSession]",
+        "shortcut - Assuming user presence and returning cipherId",
+        selectedCipherId,
+      );
+      return { cipherId: selectedCipherId, userVerified: false };
+    }
+  }
+
```
abergs, line 371: Do we need more inline documentation in this method? There's a lot of things happening.

abergs (dismissed):

apps/desktop/src/autofill/services/desktop-fido2-user-interface.service.ts
iinuwa: Added some more docs to this method.

abergs (approved):

PR #22178, opened 2026-07-31
desktop-autofill: macOS native user verification [PM-29791]

iinuwa:
## 🎟️ Tracking

## 📔 Objective

Implements native macOS user verification

## 📸 Screenshots

Register and assert with a username.

https://github.com/user-attachments/assets/dcf05344-ad3b-4c15-8077-c3f2f90b7d5d

Assert without a username/discoverable credential

https://github.com/user-attachments/assets/b316817c-0bf2-4197-bfe1-b589e4887f2b

### DIFF
--- a/apps/desktop/desktop_native/objc/src/native/autofill/commands/user_verification.h
+++ b/apps/desktop/desktop_native/objc/src/native/autofill/commands/user_verification.h
@@ -1,0 +1,8 @@
+#ifndef USER_VERIFICATION_H
+#define USER_VERIFICATION_H
+
+#import <Foundation/Foundation.h>
+
+void userVerification(void *context, NSDictionary *params);
+
+#endif
--- a/apps/desktop/desktop_native/objc/src/native/autofill/commands/user_verification.m
+++ b/apps/desktop/desktop_native/objc/src/native/autofill/commands/user_verification.m
@@ -1,0 +1,41 @@
+#import <Foundation/Foundation.h>
+#import <LocalAuthentication/LocalAuthentication.h>
+#import "../../interop.h"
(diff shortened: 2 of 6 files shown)


apps/desktop/desktop_native/objc/src/native/autofill/commands/user_verification.h
abergs, line 1: My gut feeling is that we have a preference to use swift over obj-c, but I may be wrong.

Would that complicate things?

abergs:
We should log a follow up task to build a custom prompt/window  that embeds uv/touchid instead of firing the system biometric prompt That would allow for even nicer messaging/layout. Can you log that @iinuwa?

abergs (dismissed):

apps/desktop/desktop_native/objc/src/native/autofill/commands/user_verification.h
iinuwa: Platform had preferred Objective-C over Swift due to its simplicity in FFI, but I think we could probably use Swift in a few more places. Since we have the infrastructure here to do Objective-C, and this is so small, I think it's fine to do Objective-C here.

(For another task, adding passkey client support to the desktop apps, I am considering building out a Swift framework as the Objective-C code is more complex than setting up Swift in that case.)

iinuwa:
> We should log a follow up task to build a custom prompt/window that embeds uv/touchid instead of firing the system biometric prompt That would allow for even nicer messaging/layout. Can you log that @iinuwa?

Created https://bitwarden.atlassian.net/browse/PM-41788

abergs (approved):

PR #22191, opened 2026-08-01
[PM-40157] Add extension beta icons and badge

trmartin4:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-40157

## 📔 Objective

Adds the following, to support a better beta experience for our users:
1. Beta-specific icon for the extension
2. Beta icon in the unauthenticated header for our browser extension.  **Changes to the authenticated header will be handled separately, as the header will be refactored for the redesign.**. ❓ The beta icon includes a gradient, which I don't believe has been used before.  I followed the SVG [instructions](https://components.bitwarden.com/?path=/docs/component-library-svg--docs) to the best of my knowledge.
4. Extension name is changed to "Bitwarden Password Manager Beta"
5. Helper text on hover of extension icon says "Bitwarden Beta"
6. Welcome page does not display for prerelease builds, as users are already assumed to be familiar with Bitwarden.

The icons are named the same as the non-beta icons, with the exception of a `_beta` suffix, so the manifest creation substitutes the files at build time, allowing all icon-specific logic to remain unchanged. 

See https://github.com/bitwarden/clients/pull/22445/ and https://github.com/bitwarden/clients/pull/22595 for corresponding changes for desktop client.

### Hover text shows "Bitwarden Beta"
hover

### Unauthenticated header shows beta icon
beta_anonlayout

### Locked header shows beta icon
image

### Extension details shows beta icon and title as "beta" extension
(Red icon is because the extension is installed from an unpackaged source - that would not be there if installed from the store.)
beta_detials


libs/common/src/platform/misc/flags.ts
```diff
@@ -26,14 +26,14 @@ function getFlags<T>(envFlags: string | T): T {
...
- * All flags default to "on" (true).
+ * Flags default to "off" (false) when not populated.
  * Only use for shared code in `libs`, otherwise use the client-specific function.
  * @param flag The name of the feature flag to check
  * @returns The value of the flag
  */
 export function flagEnabled<Flags extends SharedFlags>(flag: keyof Flags): boolean {
   const flags = getFlags<Flags>(process.env.FLAGS) ?? ({} as Flags);
-  return flags[flag] == null || !!flags[flag];
+  return !!flags[flag];
```
djsmith85, line 37: 🎨  Could you please split these changes into a separate PR. It's not bound to the beta-work and having a separate PR documents this change better. It also makes it easier to revert in case we'd need to.

djsmith85 (changes requested):
Platform-owned changes overall are looking good. Just as request to split a part into a separate PR.

libs/common/src/platform/misc/flags.ts
trmartin4: Absolutely.  Great callout.  I've split it out into a separate PR and stacked this PR on top of it so that it can use that logic against the new flag logic.

trmartin4: @djsmith85 This is now ready for review, built on top of the flag changes previously merged.  Thank you for suggesting a separate PR!

trmartin4:
@vleague2 tagging you for UIF review.  Looking at the pending UI Review and UI Tests items, it looks like it's because the SVG directory changed.  Can you sanity check me and make sure you agree those are good for me to accept?

apps/browser/src/popup/components/extension-anon-layout-wrapper/extension-anon-layout-wrapper.component.ts
```diff
@@ -74,7 +75,7 @@ export class ExtensionAnonLayoutWrapperComponent implements OnInit, OnDestroy {
   protected secondaryContentLocation?: SecondaryContentLocationType;
 
   protected theme: string;
-  protected logo = BitwardenLogo;
+  protected logo = flagEnabled("prereleaseBuild") ? BitwardenLogoBeta : BitwardenLogo;
```
vleague2, line 78: I don't think I've seen this kind of flag check function before 🤔 Is this a different kind of flag than a feature flag?

libs/assets/src/svg/svgs/bitwarden-logo-beta.icon.ts
```diff
@@ -0,0 +1,53 @@
...
+    <g filter="url(#bwlb-a)">
+      <g filter="url(#bwlb-b)">
+        <path fill="url(#bwlb-c)" d="M1.021 12.042c-.01.508.381.889 1.016.889 0 0 11.937.155 13.334 0s3.556-1.778 3.937-4.19 0-5.461 0-5.461c-.03.777-.163 1.09-.89 1.143H5.213c-1.524 0-3.598 1.65-3.937 3.682s-.244 3.43-.254 3.937"/>
+        <path fill="url(#bwlb-d)" d="M18.419 4.423c.726-.052.86-.366.889-1.143 0 0-.762-1.778-.89-1.143-.126.635-.6 1.56-1.142 2.286z"/>
+        <path style="fill: var(--color-pink-800)" d="M19.302 3.27C19.175 2.727 18.678 2 18.54 2c-.127 0-.15.097-.254.508-.127.508-.762 1.524-1.016 1.905-.07.105 1.397 0 1.651-.127.419-.21.488-.56.381-1.016"/>
```
vleague2, line 16: Normally we don't want to reference color vars directly in svgs, but since this is a special case, we can leave it. Do you mind leaving a jsdoc comment at the top of this `const` saying that we should not use direct color variable references and that this is an exception? (That'll help Claude when it's pulling references for future svg additions)

vleague2 (commented):
Chromatic diffs look good, I approved those

djsmith85 (dismissed):

libs/assets/src/svg/svgs/bitwarden-logo-beta.icon.ts
trmartin4: Good call. Added a JSDoc block in https://github.com/bitwarden/clients/pull/22191/commits/3fc7dca8b68a43ae13c572294840a6ec308e6ad7.

apps/browser/src/popup/components/extension-anon-layout-wrapper/extension-anon-layout-wrapper.component.ts
trmartin4: Yes, this is a compile-time flag, set in `flags.json` on the client.  This is because the fact that the client is a beta is baked into the compilation of the artifact, versus a feature flag that we consume from the server to enable functionality.

djsmith85 (approved):

vleague2 (approved):

PR #22197, opened 2026-08-03
fix: update sdk and add support for kdf in state bridge

quexten:
https://github.com/bitwarden/sdk-internal/pull/1297#discussion_r3703282785 added the KDF config as a separate entry to the state bridge. While in the long term we will remove the KDF state, for now we need to support it.

This PR updates the SDK and fixes the breakage caused by the new state bridge entry.


package-lock.json
mzieniukbw, line 1: The changes in the lock file seems out of order compared to the `package.json` changes.

mzieniukbw (changes requested):

package-lock.json
quexten: Looks like my local npm version was too old.

mzieniukbw (approved):

PR #22200, opened 2026-08-03
SSH Agent: fix all-or-nothing key parsing

### LINKED ISSUE #22158: Linux SSH Agent repeatedly logs Ed25519 parse errors and fills syslog
## Environment

- Bitwarden Desktop: `2026.6.0`
- OS: Ubuntu Linux (GNOME desktop)
- Installation: direct Linux installation under `/opt/Bitwarden`
- Bitwarden process: `/opt/Bitwarden/bitwarden-app`
- SSH Agent enabled
- SSH key type involved: OpenSSH Ed25519
- System logging: `rsyslogd` forwarding system messages to `/var/log/syslog`

## Problem

With the Bitwarden SSH Agent enabled, the desktop application repeatedly emits:

```text
Failed to parse ed25519 key
```

The message was emitted continuously and flooded the user/system logs. The observed rates were:

- `10,826` matching messages in approximately one minute
- `1,955` matching messages in approximately ten seconds

The Bitwarden vault contained multiple SSH Key items. The agent appears to process all SSH keys in the vault, but the error does not identify which vault item caused the parsing failure.

## Impact

`rsyslogd` kept an unlinked (deleted) `/var/log/syslog` file open:

```text
PID: 1234
FD: 7
Path: /var/log/syslog (deleted)
Size: 43,618,512,896 bytes (40.62 GiB)
```

The root filesystem reached 100% usage and applications began failing with `No space left on device`.

After the disk filled, GNOME Tracker produced a secondary error loop because its SQLite database could not be opened. Examples included:

```text
Could not open sqlite3 database:
'~/.cache/tracker3/files/meta.db':
unable to open database file

There is not enough space on the file system for update operations
```

The Tracker messages appear to be a consequence of the full filesystem, while the Bitwarden SSH-agent parse errors were observed before the disk-full condition.

## Key validation

The local encrypted Ed25519 key itself is valid:

```text
ssh-keygen -y -f ~/.ssh/id_ed25519
```

successfully produced the corresponding `ssh-ed25519` public key after the passphrase was entered.

The private key begins with the standard OpenSSH header:

```text
-----BEGIN OPENSSH PRIVATE KEY-----
```

No private key material, passphrase, vault export, or complete public-key data is included in this report.

## Vault details

The vault contained 13 SSH Key items, including Ed25519 and RSA items. Each displayed a private-key header, public-key field, and fingerprint. No obvious missing fingerprint was found from the decrypted item metadata.

A test SSH Key item was later created through the Bitwarden CLI and then removed. The CLI item was created by copying a passphrase-protected OpenSSH private-key string directly into the item. This may be relevant because the desktop import flow prompts for the passphrase and may normalize the key differently. However, the repeated parse errors had already been observed with the existing vault configuration.

## Expected behavior

If an SSH Key item is malformed, encrypted in an unsupported way, or otherwise unparsable, the agent should:

1. Identify the item name or ID in the diagnostic message.
2. Skip that item and continue loading valid keys.
3. Log the failure once or with rate limiting.
4. Avoid continuously writing the same error to system logs.

## Requested investigation

Please investigate:

- Whether the SSH Agent attempts to parse every SSH Key item repeatedly.
- Whether passphrase-protected OpenSSH private keys are supported in vault SSH items and by the agent.
- Why the parser failure is emitted continuously instead of being rate-limited.
- Whether the failing item name/fingerprint can be included in the diagnostic output.
- Whether the Linux desktop app should log this only to its application log instead of flooding journald/rsyslog.

## Temporary workaround

Disabling/stopping the Bitwarden SSH Agent prevents the log flood. Using the native Linux `ssh-agent` avoids the issue while retaining the SSH keys in the vault.

The deleted log file was released by restarting `rsyslogd`.

neuronull:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-41249

fixes: #22158

## 📔 Objective

The release of the v2 ssh agent revealed a bug in our import code path, which allows the import of SSH key types that the agent doesn't parse and load into it's keystore. The import bug was not introduced by the v2 but has been present during the lifetime of v1, and will be addressed separately.

The import bug was revealed by user who'd been exposed to it, with the symptom of the v2 agent failing to parse _all_ keys from the vault when the list of keys contained _any_ unsupported one. Additionally, the key parsing occurred in a loop despite the key data in the vault not changing.

This fix:
- trues up the behavior of the v2 agent to match the v1 regarding the parsing of keys- that is, if a key in the list can't be parsed, it is skipped and the remaining keys are still loaded
- adds further protections in the render service to not get in an error loop if unexpected errors occur which are deterministic and wouldn't self-resolve by retries.

Follow-up changes in separate PR(s):
- not importing unsupported key types
- robustness fixes to the electron services call chains that aren't directly part of this bug's path

## 🧪  Testing Done

manual testing:
- added an `sk-ssh-ed25519` key to the vault. confirmed that the key is not parsed but remaining keys are, and that the failure occurs only once
- manually injected IPC errors in the main service callbacks. Confirmed that a single error occurs from the render 

unit tests:
- added new unit tests on the DN and Electron side
- for the Electron side, temporarily ran the new unit tests against the un-fixed render service changes to confirm they catch the buggy behaviour

// demonstration of the app logs that occur with the fix, when an un-supported key is in the vault.

```
[Elec] 10:14:03.119 › [NAPI] [WARN] ssh_agent::storage::keydata: Skipping un-parseable key error=Unsupported key type: sk-ssh-ed25519@openssh.com cipher_id=<redacted>
[Elec] 10:14:03.119 › [NAPI] [WARN] ssh_agent::storage::keydata: Not all SSH keys parsed skipped=1 total=4
[Elec] 10:14:03.119 › [NAPI] [DEBUG] ssh_agent::agent: Replacing key data.
[Elec] 10:14:03.119 › [NAPI] [INFO] ssh_agent::agent: Key data replaced.
```


apps/desktop/src/autofill/services/ssh-agent.service.ts
```diff
@@ -412,15 +420,24 @@ export class SshAgentService implements OnDestroy {
...
           }),
-          catchError((error: unknown, source) => {
+          catchError((error: unknown) => {
             this.logService.error("Unexpected error in SSH agent replace keys", error);
-            return source;
+            // If unexpected errors reach here, they are not recoverable by retrying.
+            // Return EMPTY so that we don't resubscribe, which would use the same cached upstream
+            // values in a loop.
+            return EMPTY;
           }),
```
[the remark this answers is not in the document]
neuronull: addressed in https://github.com/bitwarden/clients/pull/22200/commits/6ccda48ad32cc6cfba228a6862adf2010447cfd4

note that there are other code paths not touched in this PR, which will also need a similar adjustment. I've intentionally left them out for a follow-up PR to contain the scope of this PR to fixing the specific bug path.

apps/desktop/desktop_native/ssh_agent/src/storage/keydata.rs
```diff
@@ -90,6 +102,29 @@ impl SSHKeyData {
...
 
+    /// Parses a batch of vault SSH keys, dropping the ones that can't be parsed..
+    #[must_use]
+    pub fn from_private_key_pems(keys: Vec<EncodedSSHKeyData>) -> Vec<Self> {
+        let total = keys.len();
+        let parsed: Vec<Self> = keys
+            .into_iter()
+            .filter_map(|k| {
+                Self::from_private_key_pem(&k.private_key_pem, k.name, k.cipher_id)
+                    .inspect_err(|error| warn!(%error, "Skipping un-parseable key"))
```
miensol: Is it reasonable to log the human-readable `k.name` in the error message to make the diagnosis easier?

neuronull: I avoided this to adhere to our policy to not log vault data (of any kind). But I believe that a debug level log with that information would not invalidate the policy. Good suggestion 👍

apps/desktop/desktop_native/ssh_agent/src/storage/keydata.rs
```diff
@@ -27,6 +28,17 @@ pub trait QueryableKeyData: Send + Sync {
     fn cipher_id(&self) -> &String;
 }
 
+/// An intermediary struct representing an SSH key from the vault,
+/// before its private key has been parsed.
+pub struct EncodedSSHKeyData {
```
coltonhurst: Nit: Regarding naming, using `raw` or `unparsed` might be better than `encoded`, as it's not really encoded, it's just not parsed, right?

Not a deal breaker though.

coltonhurst (dismissed):
Looks good, one small nit comment, feel free to take it or leave it!

apps/desktop/desktop_native/ssh_agent/src/storage/keydata.rs
neuronull: Also, I didn't mention this in the PR description but there is likely going to be a follow-up work to surface/signal un-parsed keys to the user, than having to check the logs. The debug log is still valuable as a short term improvement though.

neuronull: Actually, it'll have to be the `cipher_id`. This is something that can be cross referenced using the bitwarden CLI.  Also, the `error` reason printed in that warn log will include the algorithm kind.

Note again that reliance on this is temporary, as a follow up task to surface it to the UI is planned.

coltonhurst (approved):

PR #22214, opened 2026-08-04
refactor(unlock): route all unlocks through unlocks service

quexten:
There is a loop involved in unlock / lock of shared unlock. If a follower gets a lockstate update, and subsequently locks / unlocks, it sends this back up as a manual lock / unlock to the leader role. This can in some cases lead to an instant re-lock and should be prevented. To do so, we will track the lock / unlock source.

To prepare this, all locks and unlocks *must* go through the unlock service and are not allowed to go through the key service. The setUserKey function is removed. Further, setting and using of the auto-key is extracted into a service. 


apps/cli/src/vault/create.command.ts
```diff
@@ -182,14 +182,6 @@ export class CreateCommand {
       return Response.error("Premium status is required to use this feature.");
     }
 
-    const userKey = await this.keyService.getUserKey();
```
quexten, line 185: Don't need this, users without user-keys have been unsupported for a year now.

apps/browser/src/key-management/biometrics/background-browser-biometrics.service.ts
```diff
@@ -106,13 +106,13 @@ export class BackgroundBrowserBiometricsService extends BiometricsService {
...
           const userKey = SymmetricCryptoKey.fromSdk(response.user_key) as UserKey;
-          if (!(await this.keyService.validateUserKey(userKey, userId))) {
+          if (!(await this.keyService().validateUserKey(userKey, userId))) {
             this.logService.info("Biometric unlock for user failed: invalid user key");
             return null;
           }
 
           await this.biometricStateService.setBiometricUnlockEnabled(true, userId);
-          await this.keyService.setUserKey(userKey, userId);
+          await this.unlockService!.unlockWithDecryptedUserKey(userId, userKey);
```
quexten, line 115: At some point we can delete this once shared unlock has rolled out.

libs/auth/src/common/login-strategies/auth-request-login.strategy.ts
```diff
@@ -78,7 +80,11 @@ export class AuthRequestLoginStrategy extends LoginStrategy {
   protected override async unlock(response: IdentityTokenResponse, userId: UserId): Promise<void> {
     const authRequestCredentials = this.cache.value.authRequestCredentials;
     await this.masterPasswordService.setMasterKeyEncryptedUserKey(response.key, userId);
-    await this.keyService.setUserKey(authRequestCredentials.decryptedUserKey, userId);
+    // Login with device: the approving device supplies an already-decrypted user key.
+    await this.unlockService.unlockWithDecryptedUserKey(
```
quexten, line 84: Should port this to a real unlock method at some point, but not in this PR.

libs/key-management/src/abstractions/key.service.ts
```diff
@@ -44,38 +44,19 @@ export type CipherDecryptionKeys = {
 };
 
 export abstract class KeyService {
+  /**
+   * Provides the {@link UnlockService} used by {@link initAccount} to unlock a newly created
+   * account. Called by `DefaultUnlockService` on construction; it cannot be DI injected because of
+   * the circular dependency between the two services.
```
quexten, line 50: Removing this after v2 register rollout, then the dep chain will become simple again.

apps/desktop/src/app/services/init.service.ts
```diff
@@ -85,16 +83,14 @@ export class InitService {
...
-        // if the auto user key is set to avoid race conditions of any code trying to access
-        // the user key from mem.
-        setUserKeyInMemoryPromises.push(
-          this.userAutoUnlockKeyService.setUserKeyInMemoryIfAutoUserKeySet(userId),
-        );
+        try {
+          await this.unlockService.unlockWithAutoUnlockKey(userId);
+        } catch (e) {
+          // eslint-disable-next-line no-console
+          console.error(`Error occurred while unlocking user key for user ${userId}:`, e);
```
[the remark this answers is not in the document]
quexten: Accepted, because this is so early in the bootstrap.

PR #22217, opened 2026-08-04
chore(crypto): port crypto sync handler to sdk

quexten:
Ports the crypto sync handler to go via the sdk. Part of https://bitwarden.atlassian.net/browse/PM-39454 / https://bitwarden.atlassian.net/browse/PM-41528.


libs/common/src/key-management/state-bridge.ts
```diff
@@ -1,14 +1,24 @@
 import { filter, firstValueFrom, map, race, timer } from "rxjs";
 
+// There is no way to prevent this restricted import currently. These should be extracted out into a separate package.
+// eslint-disable-next-line no-restricted-imports
+import {
+  USER_DECRYPTION_OPTIONS,
```
quexten: For the KM and auth reviewers; IMO we should extract these, so that the PRF webauthn state is entirely separate from the decryption options. LMK if you agree, and I'll add this as a follow-up task. This would prevent this dependency loop.

This is out of scope for this ticket.

libs/common/src/platform/sync/default-sync.service.ts
```diff
@@ -457,68 +438,38 @@ export class DefaultSyncService extends CoreSyncService {
...
+              : {
+                  masterPasswordUnlock: userDecryption.masterPasswordUnlock
+                    ?.toMasterPasswordUnlockData()
+                    .toSdk(),
+                  v2UpgradeToken: userDecryption.v2UpgradeToken?.toV2UpgradeToken(),
+                  webAuthnPrfOptions: userDecryption.webAuthnPrfOptions
+                    ?.map((option) => option.toWebAuthnPrfUnlockOption())
+                    .filter((option) => option != null),
+                },
+          accountCryptographicState: profile.accountKeys?.toWrappedAccountCryptographicState(),
```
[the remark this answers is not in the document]
quexten: Yes. We have been sending account keys for long enough that no servers should return account keys null anymore.

libs/common/src/platform/sync/default-sync.service.ts
```diff
@@ -457,68 +438,38 @@ export class DefaultSyncService extends CoreSyncService {
...
+        }),
       );
-    } else {
-      await this.v2UpgradeTokenStateService.clearV2UpgradeToken(userId);
+    } catch (error) {
+      this.logService.error("[Sync] Key management sync handler failed:", error);
     }
```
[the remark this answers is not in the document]
quexten: Ok, fixing

libs/common/src/key-management/state-bridge.ts
mzieniukbw: (observation) I agree, but there are two things to notice, once user decryption options storage right now is a "catch all" for some storage options, but there is no reason for this, they all should be separate.
Secondly, it's owned by Auth, but should be owned by KM

libs/common/src/key-management/state-bridge.ts
```diff
@@ -124,6 +156,44 @@ export class JsWasmStateBridge implements WasmStateBridge {
...
+  ): Promise<void> {
+    const current = await readAtomic(this.stateProvider, this.userId, USER_DECRYPTION_OPTIONS);
+    if (current == null) {
+      return;
+    }
+
+    const updated = Object.assign(new UserDecryptionOptions(), current);
+    updated.webAuthnPrfOptions = prfOptions;
+
+    await writeAtomic(this.stateProvider, this.userId, USER_DECRYPTION_OPTIONS, updated);
```
mzieniukbw: nit: We expand the scope of who owns this state, which is not great.
Maybe it would be better idea to use the service instead `InternalUserDecryptionOptionsServiceAbstraction` ?
Or maybe this should be an opportunity to take out the WebAuthn state out of decryption options state ?

libs/common/src/platform/sync/default-sync.service.ts
```diff
@@ -189,7 +180,13 @@ export class DefaultSyncService extends CoreSyncService {
 
       await this.cipherService.clear(response.profile.id);
 
-      await this.syncUserDecryption(response.profile.id, response.userDecryption);
+      // The crypto sync handler *MUST* be the first sync handler to run. It reserves
+      // the option to reject a sync, should the data be inconsitent. In this case ,it will throw.
+      await this.runCryptoSyncHandler(
```
mzieniukbw, line 183: We should run this before `await this.cipherService.clear`, otherwise, crypto error during sync would cause user's vault to vanish.

libs/common/src/platform/sync/default-sync.service.ts
```diff
@@ -189,7 +180,13 @@ export class DefaultSyncService extends CoreSyncService {
 
       await this.cipherService.clear(response.profile.id);
 
-      await this.syncUserDecryption(response.profile.id, response.userDecryption);
+      // The crypto sync handler *MUST* be the first sync handler to run. It reserves
+      // the option to reject a sync, should the data be inconsitent. In this case ,it will throw.
```
mzieniukbw: typo
```suggestion
      // the option to reject a sync, should the data be inconsitent. In this case, it will throw.
```

libs/common/src/platform/sync/default-sync.service.ts
```diff
@@ -457,68 +438,37 @@ export class DefaultSyncService extends CoreSyncService {
...
-        userDecryption.v2UpgradeToken.toV2UpgradeToken(),
-        userId,
-      );
-    } else {
-      await this.v2UpgradeTokenStateService.clearV2UpgradeToken(userId);
-    }
+    await withPasswordManagerSdk(userId, this.sdkService, (sdk) =>
+      sdk.crypto_sync_handler().on_sync({
+        userDecryption:
+          userDecryption == null
```
dani-garcia: I wonder if we can also move this conversion to a `toSdk()` function in `UserDecryptionResponse` like we do with the others?

dani-garcia (dismissed):

JaredSnider-Bitwarden (dismissed):
Auth file LGTM!

libs/common/src/platform/sync/default-sync.service.ts
```diff
@@ -457,68 +438,37 @@ export class DefaultSyncService extends CoreSyncService {
...
+            : {
+                masterPasswordUnlock: userDecryption.masterPasswordUnlock
+                  ?.toMasterPasswordUnlockData()
+                  .toSdk(),
+                v2UpgradeToken: userDecryption.v2UpgradeToken?.toV2UpgradeToken(),
+                webAuthnPrfOptions:
+                  userDecryption.webAuthnPrfOptions == null
+                    ? []
+                    : userDecryption.webAuthnPrfOptions
+                        ?.map((option) => option.toWebAuthnPrfUnlockOption())
```
mzieniukbw: The `userDecryption.webAuthnPrfOptions` cannot be null
```suggestion
                        .map((option) => option.toWebAuthnPrfUnlockOption())
```

mzieniukbw (changes requested):

libs/common/src/platform/sync/default-sync.service.ts
quexten: Good catch. i moved it into the cipher sync handler since the only bit where it is relevant is just before ciphers. It shouldn't be in the top level.

libs/common/src/platform/sync/default-sync.service.ts
quexten: > (observation) I agree, but there are two things to notice, once user decryption options storage right now is a "catch all" for some storage options, but there is no reason for this, they all should be separate.
Secondly, it's owned by Auth, but should be owned by KM

Agree on both. I'll make a follow-up ticket. I think the decryption options are meant as a temporary item during the login flow, not permanent for unlock, but the PRF work entangled them.. :(

libs/common/src/key-management/state-bridge.ts
quexten: Okay, expanding scope of this ticket. I'll move things to a central `state-definitions.ts` file. As we start deleting features we no longer have in `ts` (pin state right now, maybe account cryptographic state soon, maybe upgrade token soon), we only need this one central file and no longer need the feature folders.

libs/key-management-ui/src/lock/services/default-webauthn-prf-unlock.service.ts
```diff
@@ -67,14 +65,13 @@ export class DefaultWebAuthnPrfUnlockService implements WebAuthnPrfUnlockService
...
-      if (!userDecryptionOptions?.webAuthnPrfOptions) {
-        return [];
-      }
-      return userDecryptionOptions.webAuthnPrfOptions.map((option) => ({
-        credentialId: option.credentialId,
-        transports: option.transports,
-      }));
+      const prfOptions = await this.getPrfOptions(userId);
+      return prfOptions
+        .filter((option) => option.credentialId != null)
```
mzieniukbw, line 70: ❓ Doesn't this potentially change the `isPrfUnlockAvailable` behaviour ?

mzieniukbw (approved):

libs/key-management-ui/src/lock/services/default-webauthn-prf-unlock.service.ts
quexten: Yes, but unlock happens only after a sync, so a sync must have happened, which means that the value must be populated. For old values the migration should apply.

dani-garcia (approved):

JaredSnider-Bitwarden (approved):

PR #22223, opened 2026-08-04
Move browser Health tab files to commercial license

lastbestdev:
## 🎟️ Tracking
N/A

## 📔 Objective
The Health report feature is commercial, but its browser extension files were placed in the OSS app. Move them to bitwarden_license/bit-browser and register the tab from the commercial app module so it only ships with the commercial extension.

OSS cannot import from bitwarden_license/, so the two OSS touchpoints get seams instead of direct references:

- The `tabs/health` route moves to the commercial routing module, re-using TabsV2Component for the same bottom navigation chrome.
- TabsV2Component reads the Health tab from the HEALTH_TAB_NAV_BUTTON injection token, which the commercial app module provides. It is absent in the OSS extension, which hides the tab.

## 📸 Screenshots
N/A


AlexRubik (approved):
lgtm

PR #22225, opened 2026-08-04
[PM-41482] Revert PM-32696: Add data field to Cipher object

YongDo-Hyun:
Reverts 409aefcf7268405343e039f86b3e3c781829addc.

This change introduced a regression in the browser client causing a WASM SDK type conversion failure:

invalid type: JsValue(Object(...)), expected a string

This restores the previous cipher mapping behavior until the blob encryption flow is fixed.


pamperer562580892423:
@YongDo-Hyun I think you should rather update your Vaultwarden server - see here: https://github.com/dani-garcia/vaultwarden/discussions/7473

PS: Here an explanation why it is a Vaultwarden issue and which PR fixed it: https://github.com/dani-garcia/vaultwarden/discussions/7473#discussioncomment-17771043

quexten (changes requested):
Sorry but what exactly is the bug this is fixing? The feature works as expected with current official servers.

PR #22242, opened 2026-08-05
[CL-982] update bit-item styles

BryanCunningham:
## 🎟️ Tracking
https://bitwarden.atlassian.net/browse/CL-982

## 📔 Objective
Updated item styles

## 📸 Screenshots
image


libs/components/src/item/item-action.component.ts
```diff
@@ -7,12 +7,7 @@ import { Component } from "@angular/core";
   imports: [],
   template: `<ng-content></ng-content>`,
   host: {
-    class:
-      /**
-       * `top` and `bottom` units should be kept in sync with `item-content.component.ts`'s y-axis padding.
-       * we want this `:after` element to be the same height as the `item-content`
-       */
-      "[&>button]:tw-relative [&>button:not([bit-item-content])]:after:tw-content-[''] [&>button]:after:tw-absolute [&>button]:after:tw-block bit-compact:[&>button]:after:tw-top-[-0.7rem] bit-compact:[&>button]:after:tw-bottom-[-0.7rem] [&>button]:after:tw-top-[-0.8rem] [&>button]:after:tw-bottom-[-0.80rem] [&>button]:after:tw-right-0 [&>button]:after:tw-left-0",
+    class: "[&>button]:tw-relative",
```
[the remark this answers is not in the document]
vleague2: I know we talked about not needing the extra click area for the main item button/content, but we might want to keep them for the end slot icon buttons. It is kind of a small click area otherwise

libs/components/src/item/item.component.ts
```diff
@@ -15,30 +9,34 @@ import { ItemActionComponent } from "./item-action.component";
...
 })
 export class ItemComponent {
   /**
    * We have `:focus-within` and `:focus-visible` but no `:focus-visible-within`
    */
   protected readonly focusVisibleWithin = signal(false);
-  @HostListener("focusin", ["$event.target"])
-  onFocusIn(target: EventTarget) {
+
+  protected readonly hostClasses = computed(() => {
```
vleague2, line 45: Any reason to have this bound to `[class]` and also properties in `class` already?

libs/components/src/item/item-content.component.ts
```diff
@@ -21,7 +21,7 @@ import { TypographyModule } from "../typography";
        * y-axis padding should be kept in sync with `item-action.component.ts`'s `top` and `bottom` units.
        * we want this to be the same height as the `item-action`'s `:after` element
        */
-      "tw-outline-none tw-text-main hover:tw-text-main tw-no-underline hover:tw-no-underline tw-text-base tw-py-2 tw-px-4 bit-compact:tw-py-1.5 bit-compact:tw-px-2 tw-bg-transparent tw-w-full tw-border-none tw-flex tw-gap-4 tw-items-center tw-justify-between disabled:tw-cursor-not-allowed [&[disabled]_[bittypography]]:!tw-text-secondary-300 [&[disabled]_i]:!tw-text-secondary-300",
+      "tw-outline-none tw-text-main hover:tw-text-main tw-no-underline hover:tw-no-underline tw-text-base tw-py-3 tw-ps-4 bit-compact:tw-py-2 bit-compact:tw-ps-3 tw-bg-transparent tw-w-full tw-border-none tw-flex tw-gap-4 tw-items-center tw-justify-between disabled:tw-cursor-not-allowed [&[disabled]_[bittypography]]:!tw-text-fg-inactive [&[disabled]_i]:!tw-text-fg-inactive",
```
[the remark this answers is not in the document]
vleague2: Looks legit, I can see it in chromatic

libs/components/src/item/item.component.ts
```diff
@@ -15,30 +9,34 @@ import { ItemActionComponent } from "./item-action.component";
...
+    const structural =
+      "tw-border [&:not(bit-layout_*)]:tw-rounded-lg bit-compact:[&:not(bit-layout_*)]:tw-rounded-none bit-compact:[&:not(bit-layout_*)]:last-of-type:tw-rounded-b-lg bit-compact:[&:not(bit-layout_*)]:first-of-type:tw-rounded-t-lg tw-mb-1.5 bit-compact:tw-mb-0 bit-compact:[&+&]:tw-border-t-0";
```
vleague2: I think we can probably remove the check for being a child of `bit-layout`. The only reason we had the non-rounded border was because the curved bottom border didn't look good on a background of the same color. Having the full border should remove the problem entirely, I would think

libs/components/src/item/item-content.component.ts
vleague2: The items are a lot taller now.... I have a feeling this will not be received well as users have been unhappy with added "whitespace" so far. Can we ask Emily about this?

libs/components/src/item/item.component.ts
willmartian: +1, I don't believe we are maintaining any intentional differences between clients going forward

libs/components/src/item/item-content.component.ts
willmartian: Furthermore, if this PR changes the height of items at all, we need to make sure that the designated height of any virtual scroll containers using items is also updated in kind in this PR.

libs/components/src/item/item-action.component.ts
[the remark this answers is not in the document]
willmartian: +1. We should not change the effective size of the click target without explicit go-ahead from Design. (It _is_ okay if it is refactored, if you deem that necessary.)

libs/components/src/item/item-content.component.ts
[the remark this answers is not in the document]
BryanCunningham: Fixed

libs/components/src/item/item-action.component.ts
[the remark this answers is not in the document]
BryanCunningham: These click areas are (now)the same as all of our other icon buttons. I personally don't see a need for them to be bigger than normal

libs/components/src/item/item.component.ts
BryanCunningham: I think this was a hold over from the previous branch's logic where we were passing in the `joined` value and doing some logic around that. I can see if we can simplify and/or just move all the class logic here

libs/components/src/item/item-content.component.ts
BryanCunningham: This is what design specified. I can take this back to them and double check this is actually what they want. I'll mention the height change causing significant refactoring/added QA.

libs/components/src/item/item-action.component.ts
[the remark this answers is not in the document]
willmartian: I generally agree, but I would still check in with design because IIRC this was added to the icon buttons after complaints about the click target in the extension. Not sure whether that is still relevant after updates to icon button and the item, but still worth confirming.

libs/components/src/item/item-content.component.ts
BryanCunningham: For posterity: We're adding a `size` input. The `default` size will match the previous item's padding. We'll have a `lg` size that adds this extra top/bottom padding

libs/components/src/item/item-action.component.ts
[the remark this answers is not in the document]
BryanCunningham: For posterity: After chatting with design, we are good to remove this added click area height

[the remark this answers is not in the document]
willmartian: Nice, thank you.

libs/components/src/item/item.component.ts
BryanCunningham: @vleague2 I moved all the classes in here

libs/components/src/item/item.component.ts
BryanCunningham: I removed references to bit-layout

jengstrom-bw (dismissed):
vault items look good

libs/components/src/item/item.component.ts
```diff
@@ -1,44 +1,70 @@
...
    * We have `:focus-within` and `:focus-visible` but no `:focus-visible-within`
    */
   protected readonly focusVisibleWithin = signal(false);
-  @HostListener("focusin", ["$event.target"])
-  onFocusIn(target: EventTarget) {
-    this.focusVisibleWithin.set((target as HTMLElement).matches("[data-fvw-target]:focus-visible"));
+
+  protected readonly hostClasses = computed(() => {
+    const base =
+      "tw-flex tw-rounded-lg tw-gap-3 tw-justify-between tw-overflow-hidden tw-bg-bg-primary [&:has([data-item-main-content]_button:hover,[data-item-main-content]_a:hover)]:tw-cursor-pointer [&:has([data-item-main-content]_button:enabled:hover,[data-item-main-content]_a:hover,[data-item-main-content]_button:enabled:focus-visible,[data-item-main-content]_a:focus-visible)]:tw-bg-bg-brand-softer tw-text-fg-heading tw-border-solid tw-border-border-base tw-min-h-9";
```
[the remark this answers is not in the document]
willmartian: I agree we can remove the story / docs that were previously capturing this. @BryanCunningham

willmartian (dismissed):
Looks good! Just one old story that needs to be removed:

https://github.com/bitwarden/clients/pull/22242#discussion_r3750526340

jengstrom-bw (dismissed):

vleague2 (dismissed):

jengstrom-bw (dismissed):

vleague2 (dismissed):

vleague2 (approved):

jengstrom-bw (approved):

PR #22251, opened 2026-08-06
feat(dev-tools): add feature flag override menu

quexten:
Adds a dev-tools library, and as part of this library the "feature flag overrides UI". This is a dialog that is enabled when:
- Running in dev mode
- Running with a special env var in production
- Pasting a command into the dev tools in production

Doing so, the feature flag menu becomes available, which allows introspecting the server-sent state, and overriding it. This is useful when testing featureflags against a local server quicker, and for testing feature flags against the production server.

Local feature flags are already available by editing data.json / state, and this just adds a UI layer on top. The gating is not meant to fully prevent access by end users, just make it non-discoverable by default.

https://github.com/user-attachments/assets/bef48772-b7a4-4467-bd10-ca98fb135e15

https://bitwarden.atlassian.net/browse/PM-41526

Please note: No I18N is added since this is a dev-only menu.


libs/dev-tools/package.json
```diff
@@ -0,0 +1,11 @@
+{
+  "name": "@bitwarden/dev-tools",
```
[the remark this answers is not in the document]
quexten: Yeah, that seems ok. Dev tools can be made by all teams.

libs/dev-tools/src/feature-flag-overrides/services/feature-flag-override-menu.service.ts
```diff
@@ -0,0 +1,58 @@
...
+
+  /**
+   * Exposes `enableFeatureFlagOverrideMenu()` / `disableFeatureFlagOverrideMenu()` on the global
+   * scope so the menu can be turned on from the developer console in a build where it is off by
+   * default.
+   */
+  installGlobalHook(): void {
+    globalThis.enableFeatureFlagOverrideMenu = () => this.setEnabled(true);
+    globalThis.disableFeatureFlagOverrideMenu = () => this.setEnabled(false);
+  }
```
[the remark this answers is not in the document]
quexten: Appsec does not review specific security implementations. No, it should not, you can already today set the feature flags by editing data.json / opening browser dev tools, so there is no security change here.

djsmith85:
@quexten Might wanna sync up with @jprusik as he has worked on something similar in the past. He might be able to point out some gotchas he ran into or share some ideas that expand on this work.

harr1424 (dismissed):
Changes to Tools-owned files look good

jprusik:
Here's the PoC I put together years back:
https://github.com/bitwarden/clients/compare/main...refs/heads/pm-12767

quexten:
@djsmith85 @jprusik thanks, but I think this is a different use-case. The branch from @jprusik looks to add a user-facing labs settings, which are not a 1:1 mapping of the feature flags. This PR adds a dev-facing toggle for *all flags*. This is not intended to be used by regular users.

apps/desktop/src/utils.ts
```diff
@@ -83,6 +83,14 @@ export function allowBrowserintegrationOverride() {
...
 }
 
+/**
+ * Offers the developer-only feature flag override menu in a build where it would otherwise be
+ * hidden, so a flag can be flipped locally against a production build.
+ */
+export function featureFlagOverrideMenuEnvEnabled() {
+  return process.env.ENABLE_FEATURE_FLAG_OVERRIDE_MENU === "true";
+}
+
```
djsmith85: ⚠️ Please move this to `apps/desktop/src/main/platform-utils.main.ts` as this should only be used from the main process. The renderer can access this via ipc.

djsmith85 (changes requested):
Only a partial review unfortunately, I haven't gotten around to it yet, but I also wanted to have a look at the unowned code. I'll discuss with the team if platform should own it.

Some conflicts need resolving and please move `featureFlagOverrideMenuEnvEnabled()`

apps/desktop/src/utils.ts
quexten: Damn, sorry this slipped by. At this point it's been a while since I worked on it so I don't exactly recall, but it looks like I only tested in dev mode, and just assumed the feature flag check would work?

I'll fix it at any rate

djsmith85: In addition you do expose the value via ipc (preload) so it should work, it's just shouldn't be part of the compiled renderer code. The issue why this file is being pulled in is `isDev()`.

harr1424 (dismissed):

harr1424 (approved):

PR #22257, opened 2026-08-06
desktop-autofill: Refactor UV reprompt and window management [PM-29791]

iinuwa:
## 🎟️ Tracking

## 📔 Objective

Refactoring for future user verification PRs.

Moves master password reprompts out of the UI components into the DesktopFido2UserInterfaceSession, so that this can be grouped with vault unlock and native OS user verification.

This also transfers responsibility to close the modal to the session rather than the UI component.

### DIFF
--- a/apps/desktop/src/app/services/services.module.ts
+++ b/apps/desktop/src/app/services/services.module.ts
@@ -149,6 +149,7 @@ import {
   RoutedVaultFilterBridgeService,
   VAULT_FILTER_BASE_ROUTE,
   Vfo1TerminologyService,
+  PasswordRepromptService,
 } from "@bitwarden/vault";
 
 import { DesktopLoginComponentService } from "../../auth/login/desktop-login-component.service";
@@ -448,6 +449,7 @@ const safeProviders: SafeProvider[] = [
       MessagingServiceAbstraction,
       Router,
       DesktopSettingsService,
+      PasswordRepromptService,
     ],
   }),
   safeProvider({
--- a/apps/desktop/src/autofill/modal/credentials/fido2-create.component.spec.ts
+++ b/apps/desktop/src/autofill/modal/credentials/fido2-create.component.spec.ts
@@ -13,7 +13,6 @@ import { CipherService } from "@bitwarden/common/vault/abstractions/cipher.servi
 import { CipherRepromptType, CipherType } from "@bitwarden/common/vault/enums";
 import { CipherView } from "@bitwarden/common/vault/models/view/cipher.view";
 import { DialogService } from "@bitwarden/components";
-import { PasswordRepromptService } from "@bitwarden/vault";
 
 import { DesktopAutofillService } from "../../../autofill/services/desktop-autofill.service";
 import { DesktopSettingsService } from "../../../platform/services/desktop-settings.service";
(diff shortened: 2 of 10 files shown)


apps/desktop/src/autofill/services/desktop-fido2-user-interface.service.ts
```diff
@@ -479,4 +494,47 @@ export class DesktopFido2UserInterfaceSession implements Fido2UserInterfaceSessi
...
+   * actually took place.
+   */
+  private async repromptMasterPassword(): Promise<boolean> {
+    if (!(await this.passwordRepromptService.enabled())) {
+      this.logService.info(
+        "[DesktopFido2UserInterfaceSession]",
+        "Cannot reprompt for the master password because the account does not have one",
+      );
+      return false;
+    }
```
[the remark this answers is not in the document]
iinuwa: will be handled in #22259. This is feature-flagged so no user-visible behavior yet

libs/common/src/platform/abstractions/fido2/fido2-user-interface.service.abstraction.ts
```diff
@@ -84,7 +84,7 @@ export abstract class Fido2UserInterfaceSession {
    */
   abstract pickCredential(
     params: PickCredentialParams,
-  ): Promise<{ cipherId: string; userVerified: boolean }>;
+  ): Promise<{ cipherId: string | undefined; userVerified: boolean }>;
```
iinuwa, line 87: This is to satisfy strict type-checking and conforms to the method's doc comment

apps/desktop/src/app/services/services.module.ts
iinuwa, line 1: To platform reviewers: this is moving PasswordReprompt responsibility from a UI component to the desktop UI interface service, so we're adding the dependency here.

apps/desktop/src/autofill/modal/credentials/fido2-create.component.spec.ts
```diff
@@ -159,16 +155,6 @@ describe("Fido2CreateComponent", () => {
...
-    it("should not add passkey when password reprompt is cancelled", async () => {
-      const cipher = createMockCiphers()[0];
-      cipher.reprompt = CipherRepromptType.Password;
-      mockPasswordRepromptService.showPasswordPrompt.mockResolvedValue(false);
-
-      await component.addCredentialToCipher(cipher);
-
-      expect(mockSession.notifyConfirmCreateCredential).toHaveBeenCalledWith(false, cipher);
-    });
-
```
abergs, line 171: note: This is replaced by a test in the service instead.

abergs (approved):
nice

djsmith85 (approved):
Discussed the differences offline between PasswordReprompt and UserVerification. UV might be better suited, but for now we are going with PasswordRepromptService.

With context on the following PR https://github.com/bitwarden/clients/pull/22258 and  https://github.com/bitwarden/clients/pull/22259 I raised concerns about disparity for user verification on browser and desktop. This is noted and is done because the existing abstractions also need to be re-worked and logic needs to move to the SDK. Goal is to get this feature set released and then continue refining and ensuring parity of the implementation across clients.

Thanks @iinuwa for the great discussion yesterday.

Approving for platform

PR #22258, opened 2026-08-06
desktop-autofill: Define the desktop FIDO2 user verification service [PM-29791]

iinuwa:
## 🎟️ Tracking

## 📔 Objective

Creates the abstraction for implementing user verification for FIDO2 ceremonies on the desktop client. Implementation will be added in a subsequent PR.

### DIFF
--- a/apps/desktop/src/autofill/services/desktop-fido2-unsupported-user-verification.service.ts
+++ b/apps/desktop/src/autofill/services/desktop-fido2-unsupported-user-verification.service.ts
@@ -1,0 +1,21 @@
+import { LogService } from "@bitwarden/logging";
+
+import { DesktopFido2UserVerificationService } from "./desktop-fido2-user-verification.service.abstraction";
+
+/**
+ * Used on platforms with no OS user verification integration.
+ *
+ * Verification always fails rather than throwing, so ceremonies that require it
+ * are rejected by the authenticator while ceremonies that don't are unaffected.
+ */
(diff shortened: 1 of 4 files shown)


iinuwa:
I'm not sure what GitHub is talking about with a merge conflict... maybe it thinks the two service.module.ts modifications in this PR and the previous one in the stack conflict? They should apply cleanly, so I don't know what's going on there. I'm also not sure why the other checks aren't running. I'll look into this some more.

apps/desktop/src/autofill/services/desktop-fido2-unsupported-user-verification.service.ts
```diff
@@ -0,0 +1,21 @@
+import { LogService } from "@bitwarden/logging";
+
+import { DesktopFido2UserVerificationService } from "./desktop-fido2-user-verification.service.abstraction";
+
+/**
+ * Used on platforms with no OS user verification integration.
```
abergs, line 6: Linux only?

abergs (dismissed):

apps/desktop/src/autofill/services/desktop-fido2-unsupported-user-verification.service.ts
iinuwa: yes, but more specifically sandboxed Linux packages that don't have access to issue Polkit prompts (there's not currently an API to do user verification for sandboxed apps)

abergs (approved):

PR #22267, opened 2026-08-06
[PM-41472] feat(web): add the My folders page

gbubemismith:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-41472

Server side: bitwarden/server#8157 adds the `DELETE /folders` endpoint this consumes.

## 📔 Objective

Adds the My folders page, reached from the Manage section of the Password Manager side nav.
Lists the user's folders with item counts, search and sort, per-row edit and delete, and
multiselect delete. Both the route and the nav entry are gated on `vfo1-foundation`.

## 📸 Screenshots

https://github.com/user-attachments/assets/2d237568-2083-4aad-8f79-0976c84d5bed

#### With updated side nav
image


apps/web/src/app/vault/my-folders/my-folders.component.html
```diff
@@ -0,0 +1,68 @@
+<app-header></app-header>
+
+<bit-table-v2 [tableDef]="table" [filter]="filter" [selection]="selection" [loading]="loading()">
```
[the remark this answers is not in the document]
gbubemismith: New tables should use `bit-table-v2`

BryanCunningham (dismissed):

apps/web/src/app/layouts/user-layout.component.html
```diff
@@ -58,7 +58,7 @@
       </bit-nav-section>
 
       <bit-nav-section [label]="'manage' | i18n">
-        <bit-nav-item icon="bwi-folder" [text]="'myFolders' | i18n"></bit-nav-item>
+        <bit-nav-item icon="bwi-folder" [text]="'myFolders' | i18n" route="folders"></bit-nav-item>
```
nick-livefront, line 21: ❓ Does this need to be hidden behind the feature flag? My instincts say yes but I could be missing a piece of info here!

apps/web/src/app/vault/my-folders/my-folders.component.html
```diff
@@ -0,0 +1,74 @@
...
+        (click)="editFolder(row)"
+      ></button>
+      <button
+        slot="end"
+        type="button"
+        bitIconButton="bwi-trash"
+        buttonType="dangerGhost"
+        size="small"
+        [id]="'my-folders_button_delete-' + row.id"
+        [label]="'delete' | i18n"
```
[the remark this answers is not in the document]
nick-livefront: I'd agree with claude that this would be a nice improvement while we're here

libs/vault/src/components/delete-folder-dialog/delete-folder-dialog.component.ts
```diff
@@ -0,0 +1,70 @@
...
+    "folderName" in this.params
+      ? this.i18nService.t("deleteFolder")
+      : this.i18nService.t("deleteFoldersCount", this.params.count);
+
+  // The single-folder description is one localized sentence with the name in the middle, so
+  // translators control word order. It's split around the marker to render the name in an italic
+  // span between the surrounding text.
+  private readonly descriptionParts = this.folderName
+    ? this.i18nService.t("deleteFolderDescription", NAME_MARKER).split(NAME_MARKER)
+    : [this.i18nService.t("deleteFoldersDescription")];
```
nick-livefront, line 58: This is pretty unique, so I appreciate the comment explanation!

libs/vault/src/components/delete-folder-dialog/delete-folder-dialog.component.ts
nick-livefront, line 1: ❓ Will this be used on the desktop app? If so could this be moved to `libs/vault` for reuse?

libs/vault/src/components/my-folders/my-folders.component.spec.ts
```diff
@@ -0,0 +1,304 @@
...
+import { HeaderModule } from "../../layouts/header/header.module";
+
+import * as deleteFolderDialog from "./delete-folder-dialog/delete-folder-dialog.component";
+import { buildFolderRows, MyFoldersComponent } from "./my-folders.component";
+
+global.ResizeObserver = jest.fn().mockImplementation(() => ({
+  observe: jest.fn(),
+  unobserve: jest.fn(),
+  disconnect: jest.fn(),
+}));
```
nick-livefront, line 33: ❓ What was this needed for?

apps/web/src/app/vault/my-folders/my-folders.component.ts
```diff
@@ -0,0 +1,244 @@
...
+
+import { HeaderModule } from "../../layouts/header/header.module";
+
+import { openDeleteFolderDialog } from "./delete-folder-dialog/delete-folder-dialog.component";
+
+export type FolderTableRow = {
+  id: string;
+  name: string;
+  /** An em-dash stands in for a folder with no name. */
+  displayName: string;
```
nick-livefront: ⛏️ (non-blocking): Should this be marked as optional given there is a default value?


```suggestion
  /** 
   * Name displayed for at the folder
   * @default em-dash
   */
  displayName?: string;
```

libs/vault/src/components/add-edit-folder-dialog/add-edit-folder-dialog.component.ts
```diff
@@ -78,14 +83,52 @@ export class AddEditFolderDialogComponent implements AfterViewInit, OnInit {
...
+    this.configService.getFeatureFlag$(FeatureFlag.VFO1Foundation),
+    { initialValue: false },
+  );
+
   folder: FolderView = new FolderView();
 
   variant: "add" | "edit" = "add";
 
+  /** Applies the VFO1 name rules when the flag is on, a bare required check otherwise. */
+  private readonly nameValidator: ValidatorFn = (control) => {
```
nick-livefront, line 93: ❓ Once VFO1 goes live, this will mean anyone who edits a folder name longer than 50 characters will need to shrink it down. Is that a known scenario? 

No change needed, just calling it out!

nick-livefront (changes requested):
Only requesting changes for the feature flag on the nav item, if it is needed!

apps/web/src/app/layouts/user-layout.component.html
gbubemismith: It's handled [here](https://github.com/bitwarden/clients/pull/22283/changes#diff-78b15d292e37b3d21a41e036b63f9f88064b51b5ba6e8de6b3672e4991a0aee6). The whole side nav is gated and `folders` is just an item on the side nav

libs/vault/src/components/delete-folder-dialog/delete-folder-dialog.component.ts
gbubemismith: I was caught up between handling it in this ticket or doing it in https://bitwarden.atlassian.net/browse/PM-41475. I think I'll just handle it on this ticket

libs/vault/src/components/my-folders/my-folders.component.spec.ts
gbubemismith: `bit-table-v2` calls new `ResizeObserver(...)` internally and jsdom does not provide it, and the test crashes without it. I have seen other tests put it in `beforeEach`,  I think this approach is a bit tidier

apps/web/src/app/vault/my-folders/my-folders.component.ts
gbubemismith: This is non-optional on purpose, displayName is always populated in `buildFolderRows`

libs/vault/src/components/add-edit-folder-dialog/add-edit-folder-dialog.component.ts
gbubemismith: Got product to drop this. Nice catch!

libs/vault/src/components/add-edit-folder-dialog/add-edit-folder-dialog.component.ts
```diff
@@ -127,10 +149,12 @@ export class AddEditFolderDialogComponent implements AfterViewInit, OnInit {
   /** Submit the new folder */
   submit = async () => {
     if (this.folderForm.invalid) {
+      this.folderForm.markAllAsTouched();
       return;
     }
 
-    this.folder.name = this.folderForm.controls.name.value ?? "";
+    const name = this.folderForm.controls.name.value ?? "";
+    this.folder.name = this.vfo1Enabled() ? name.trim() : name;
```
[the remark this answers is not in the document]
gbubemismith: Addressed 033636030f1e524a8ac87cd21b015d2d8e737c2d

nick-livefront (dismissed):

apps/web/src/app/vault/my-folders/my-folders.component.html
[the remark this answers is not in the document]
BryanCunningham: I think we should add this as screen reader only text and tie it to the button via `arai-describedby`. Otherwise the item name in the tooltip could be really long and make it hard to read for sighted users. I'd like to make a similar change to the other instances where we do this as well

BryanCunningham (dismissed):

BryanCunningham (dismissed):

nick-livefront (dismissed):

nick-livefront (approved):

BryanCunningham (approved):

PR #22268, opened 2026-08-07
[PM-37839] Browser: Delete item from at risk category view

lastbestdev:
## 🎟️ Tracking
https://bitwarden.atlassian.net/browse/PM-37839

## 📔 Objective
From within the Health report categories, users are able to select items to (soft) delete from their vault. These changes implement a new dialog that will handle deleting the items from their vault, and therefore from appearing within their Health report. The dialog includes a section to display all at risk categories the selected item falls under before they confirm deletion.

Deleting an item from within the health report detail view will soft delete the item from the user's vault, as well as immediately remove it from the last health report that was run, without triggering a new report to be built.

## 📸 Screenshots

https://github.com/user-attachments/assets/f23018ef-091b-46f3-a1ff-0991ae719744

---

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>

### DIFF
--- a/bitwarden_license/bit-browser/src/popup/dirt/health/health-delete-at-risk-item-dialog.component.html
+++ b/bitwarden_license/bit-browser/src/popup/dirt/health/health-delete-at-risk-item-dialog.component.html
@@ -29,7 +29,7 @@
     }
   </div>
   <div bitDialogFooter class="tw-flex tw-flex-col tw-gap-2">
-    <button bitButton type="button" buttonType="danger" (click)="onDeleteItem()">
+    <button bitButton type="button" buttonType="danger" [bitAction]="onDeleteItem">
       {{ "delete" | i18n }}
     </button>
     <button bitButton type="button" bitDialogClose>{{ "cancel" | i18n }}</button>
--- a/bitwarden_license/bit-browser/src/popup/dirt/health/health-delete-at-risk-item-dialog.component.html
+++ b/bitwarden_license/bit-browser/src/popup/dirt/health/health-delete-at-risk-item-dialog.component.html
@@ -5,8 +5,26 @@
       Deleting this login removes it from your vault but doesn't fix the risk. To protect your
       account, change your password first.
     </p>
-    <strong>{{ item().name }}</strong>
-    <!-- TODO: ADDITIONAL RISKS TABLE HERE -->
+    <!-- TODO: update below to read categories properly for item -->
+    @if (true) {
+      <bit-section>
+        <bit-section-header>
+          <p>This password has additional risks:</p>
+        </bit-section-header>
+        <bit-card class="tw-flex tw-flex-col tw-gap-2 tw-justify-start tw-items-start">
+          <div class="tw-w-full tw-flex tw-flex-row tw-items-center tw-justify-start tw-gap-2">
+            <bit-icon-tile icon="bwi-warning" variant="warning" slot="start" />
+            Weak
+            <span class="tw-ml-auto">Yes</span>
+          </div>
+          <div class="tw-w-full tw-flex tw-flex-row tw-items-center tw-justify-start tw-gap-2">
+            <bit-icon-tile icon="bwi-refresh" variant="primary" slot="start" />
+            Reused
+            <span class="tw-ml-auto">6 times</span>
+          </div>
+        </bit-card>
+      </bit-section>
+    }
   </div>
   <div bitDialogFooter class="tw-flex tw-flex-col tw-gap-2">
     <button bitButton type="button" buttonType="danger" (click)="onDeleteItem()">Delete</button>
--- a/bitwarden_license/bit-browser/src/popup/dirt/health/health-delete-at-risk-item-dialog.component.html
+++ b/bitwarden_license/bit-browser/src/popup/dirt/health/health-delete-at-risk-item-dialog.component.html
@@ -5,10 +5,11 @@
       Deleting this login removes it from your vault but doesn't fix the risk. To protect your
       account, change your password first.
     </p>
+    <strong>{{ item().name }}</strong>
     <!-- TODO: ADDITIONAL RISKS TABLE HERE -->
   </div>
   <div bitDialogFooter class="tw-flex tw-flex-col tw-gap-2">
-    <button bitButton type="button" buttonType="danger">Delete</button>
+    <button bitButton type="button" buttonType="danger" (click)="onDeleteItem()">Delete</button>
     <button bitButton type="button" bitDialogClose>Cancel</button>
   </div>
 </bit-simple-dialog>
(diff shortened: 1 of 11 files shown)


Banrion (commented):
⚠️ Important Claude comment: https://github.com/bitwarden/clients/pull/22268/changes#r3732616198

⛏️ Storybook for the health-risk-category-detail.component should cover the delete and use the `action` to log when a delete or cancel was selected

bitwarden_license/bit-browser/src/popup/dirt/health/health-risk-category-detail.component.ts
```diff
@@ -107,6 +120,25 @@ export class HealthRiskCategoryDetailComponent {
     });
   };
 
+  readonly onDeleteItem = async (item: CipherView) => {
```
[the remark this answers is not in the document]
lastbestdev: added in latest

libs/common/src/enums/feature-flag.enum.ts
```diff
@@ -182,7 +182,7 @@ export const DefaultFeatureFlagValue = {
   [FeatureFlag.PasskeyLoginReport]: FALSE,
   [FeatureFlag.AccessIntelligenceReportFileStorage]: FALSE,
   [FeatureFlag.AccessIntelligenceAdoptionUxImprovements]: FALSE,
-  [FeatureFlag.BrowserExtensionHealthReport]: FALSE,
+  [FeatureFlag.BrowserExtensionHealthReport]: true,
```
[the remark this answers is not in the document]
lastbestdev: removed

Banrion (dismissed):

bitwarden_license/bit-browser/src/popup/dirt/health/health-risk-category-detail.component.html
```diff
@@ -53,8 +53,20 @@ <h2 bitTypography="h5">{{ "items" | i18n }}</h2>
                   type="button"
                   bitIconButton="bwi-more-horizontal"
                   label="{{ 'options' | i18n }}"
+                  [bitMenuTriggerFor]="itemMenu"
                 ></button>
               </bit-item-action>
+              <bit-menu #itemMenu>
```
Banrion: ♻️ `bit-menu` should be placed inside the `bit-item-action`

bitwarden_license/bit-browser/src/popup/dirt/health/health-delete-at-risk-item-dialog.component.ts
```diff
@@ -0,0 +1,83 @@
...
+    switch (category) {
+      case RiskCategory.Exposed:
+        return { showWeak: item.hasWeakPassword, showReused: item.hasReusedPassword };
+      case RiskCategory.Weak:
+        return { showWeak: false, showReused: item.hasReusedPassword };
+      case RiskCategory.Reused:
+      default:
+        return { showWeak: false, showReused: false };
+    }
+  });
```
Banrion, line 67: ♻️ Local Claude finding:

♻️ **DEBT**: `additionalRisks` keys off the current category instead of the item's own risk flags, so its correctness depends on an invariant owned by another layer.

<details>

The switch returns only the risks that sit *below* `currentCategory` in the exposed > weak > reused order, and never reads `item.hasExposedPassword` or `item.exposedCount` at all.

That produces the right rows today only because `DefaultVaultHealthReportService.highestRiskCategory` dedups highest-risk-wins, so an item in the `weak` list is guaranteed not to be exposed and an item in the `reused` list is guaranteed to be neither. Nothing in this file states that dependency, and no test pins it. If the report ever lists a login in every category it qualifies for — or the hierarchy order changes — this dialog silently under-reports risk on the one screen whose job is to show risk before a destructive action, and the missing Exposed row would have to be built from scratch.

The acceptance criteria describe the condition as "qualifies for at least one other risk category beyond the current drill-down," which is expressible directly from the item and is robust either way:

```ts
readonly additionalRisks = computed(() => {
  const item = this.item();
  const category = this.currentCategory();

  return {
    showExposed: item.hasExposedPassword && category !== RiskCategory.Exposed,
    showWeak: item.hasWeakPassword && category !== RiskCategory.Weak,
    showReused: item.hasReusedPassword && category !== RiskCategory.Reused,
  };
});
```

If the hierarchy shortcut is deliberate, a comment naming `DefaultVaultHealthReportService.highestRiskCategory` as the reason — plus a test that fails if the dedup is removed — would keep the two layers from drifting apart unnoticed.

</details>

bitwarden_license/bit-browser/src/popup/dirt/health/health-delete-at-risk-item-dialog.component.html
```diff
@@ -0,0 +1,37 @@
...
+              {{ "reused" | i18n }}
+              <span class="tw-ml-auto">{{ "xTimes" | i18n: item().reuseCount }}</span>
+            </div>
+          }
+        </bit-card>
+      </bit-section>
+    }
+  </div>
+  <div bitDialogFooter class="tw-flex tw-flex-col tw-gap-2">
+    <button bitButton type="button" buttonType="danger" (click)="onDeleteItem()">
```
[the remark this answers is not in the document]
lastbestdev: done in latest

bitwarden_license/bit-browser/src/popup/dirt/health/health-delete-at-risk-item-dialog.component.html
```diff
@@ -0,0 +1,37 @@
+<bit-simple-dialog hideIcon>
+  <div bitDialogTitle class="tw-text-left">{{ "deleteItem" | i18n }}</div>
+  <div bitDialogContent class="tw-flex tw-flex-col tw-gap-2 tw-w-full">
+    <p class="tw-text-left">
+      {{ "deleteAtRiskItemDescription" | i18n }}
+    </p>
+    @if (additionalRisks().showWeak || additionalRisks().showReused) {
+      <bit-section>
+        <bit-section-header>
+          <p>{{ "passwordHasAdditionalRisks" | i18n }}</p>
```
Banrion, line 10: ⚠️ Styling of this copy in the Figma differs from the screenshot. The screenshot also has "Reused - Yes" in the table, but it looks like it's out of date.

Please check styling of all text

Banrion (dismissed):

bitwarden_license/bit-browser/src/popup/dirt/health/health-delete-at-risk-item-dialog.component.ts
lastbestdev: Think Claude is hallucinating over intended functionality here. The implementation currently covers showing only the categories of lower risk than the risk category currently being viewed for the item to be deleted. We never show a row for an exposed password in the "additional risks" table, since it is already the highest risk category. 

Unless I'm missing something here, I'm going to disregard this one.

bitwarden_license/bit-browser/src/popup/dirt/health/health-delete-at-risk-item-dialog.component.html
lastbestdev: updated in latest

AlexRubik (approved):
no blockers

PR #22321, opened 2026-08-10
Route remaining SDK-capable Send operations through the SDK

adudek-bw:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-41233

## 📔 Objective

Routes the remaining SDK-capable Send operations through @bitwarden/sdk-internal, behind the existing pm-30110-sdk-sends-api flag

- File Send create/upload via the SDK
- Encryption dedup
- CLI cross-instance receive support


libs/common/src/tools/send/services/send-sdk-api.service.ts
```diff
@@ -238,6 +293,120 @@ export class SendSdkApiService implements SendApiServiceAbstraction {
...
+          if (sendId == null) {
+            throw new Error("Created file send is missing its id.");
+          }
+
+          // `encryptedFileBuffer` crosses the wasm boundary as a JS `number[]` rather than a
+          // `Uint8Array` (the SDK's `Tsify` derive doesn't special-case `Vec<u8>`), costing several
+          // times the ciphertext's byte length. Pull out just the fields the upload needs so
+          // `created` — and the oversized array it holds — isn't kept alive by this closure for the
+          // duration of the (potentially slow) network upload.
+          const encryptedFileBuffer = new Uint8Array(created.encryptedFileBuffer);
```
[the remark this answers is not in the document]
harr1424: @adudek-bw  this seems like it might be worth testing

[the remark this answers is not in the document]
adudek-bw: Addressed in 3e1d043 — capped SDK file-Send create at 100 MB (well under the 500 MB premium limit) so an oversized file fails fast with a clear error before any bytes are read or the SDK is touched, rather than risking OOM.

Verified locally against a real server with the flag on:
- A 150 MB file is rejected immediately by the guard, before hitting the network.
- An 80 MB file completes and round-trips byte-identical through the real `create_file_send`/`upload_send_file` path.

This is a stopgap, not the fix — the durable fix is upstream (`Vec<u8>` → a typed array on `CreateFileSendResponse.encrypted_file_buffer`), tracked in [PM-41234](https://bitwarden.atlassian.net/browse/PM-41234) item 4, along with the mitigation already in place (dropping the oversized intermediate array reference right after copying it, so it isn't held alive for the duration of the upload). 100 MB was chosen conservatively given the multi-GB worst case at the 500 MB ceiling and the tightest-memory context this runs in (a browser extension's MV3 service worker) — happy to adjust if there's a better number, but wanted to land something rather than leave it unguarded.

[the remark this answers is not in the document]
harr1424: ❓ If the premium membership affords a user 500MB are we really correct in capping SDK file-Sends at 100MB?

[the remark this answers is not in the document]
adudek-bw: Good call! Changing to 500MB

libs/common/src/tools/send/services/send-sdk-api.service.ts
```diff
@@ -238,6 +309,126 @@ export class SendSdkApiService implements SendApiServiceAbstraction {
...
+  ): Promise<SdkSendView> {
+    if (file == null) {
+      throw new Error("File send creation requires file data.");
+    }
+    const fileSize = file instanceof ArrayBuffer ? file.byteLength : file.size;
+    if (fileSize > MAX_SDK_FILE_SEND_SIZE_BYTES) {
+      throw new Error(
+        `File is too large to send (max ${MAX_SDK_FILE_SEND_SIZE_BYTES / (1024 * 1024)} MB).`,
+      );
+    }
```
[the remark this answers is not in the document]
adudek-bw: Addressed in 267ed46. Two things converged here:

1. I'd bumped `MAX_SDK_FILE_SEND_SIZE_BYTES` to 500 MB in 5743b63 — matching the pre-existing, already-advertised Send limit (that `maxFileSize` hint predates this PR), rather than inventing a stricter SDK-only ceiling. So the advertised/enforced mismatch you flagged is gone; the memory-multiplication risk at 500 MB is still real, but that's the same tradeoff already tracked in PM-41234, not something a lower number fixes without also lowering the advertised limit.

2. Added a size check in `send-file-details.component.ts`'s `valueChanges` subscription: rejects an oversized file with the existing translated `maxFileSize` toast (same key the hint already uses, same pattern `cipher-attachments.component.ts:218-223` uses) before it ever reaches `setFile`/submit, importing `MAX_SDK_FILE_SEND_SIZE_BYTES` from the service so the two can't drift apart. The service-level guard stays as the backstop for the CLI, which has no form layer to validate at.

Also added tests for both the oversized-file-rejected and at-limit-allowed cases.

harr1424 (dismissed):

djsmith85 (commented):
Started an offline line discussion to tighten scope and design an architecture for receiving Sends without widening the SDK API surface just for Sends.

apps/cli/src/tools/send/commands/create.command.spec.ts
```diff
@@ -284,7 +260,12 @@ describe("SendCreateCommand", () => {
         expect(response.message).toBe("--password and --emails are mutually exclusive.");
       });
 
-      it("should use CLI value when JSON has different value of same type", async () => {
+      // `createSend` resolves `req.emails ?? options.emails`, so the JSON value wins on create.
+      // Note this is the opposite of `SendEditCommand`, where the CLI flag overrides the JSON —
+      // an inconsistency between the two commands, not something this test asserts is correct.
+      // The previous version of this test claimed the CLI value won, but only ever inspected a
+      // hard-coded `sendService.encrypt` mock return, so it never exercised the precedence.
+      it("uses the JSON value when JSON and CLI disagree on the same field", async () => {
```
harr1424, line 268: ❓  Thanks for removing these mocked values throughout this file - do you think it's worth creating a ticket in the backlog to align how this works and eliminate the inconsistency?

apps/cli/src/tools/send/commands/receive.command.spec.ts
```diff
@@ -450,41 +463,117 @@ describe("SendReceiveCommand", () => {
...
+        apiUrl: "https://custom.example.com/api",
+        identityUrl: "https://custom.example.com/identity",
+      });
+    });
+  });
+
+  describe("Pm30110SdkSendsApi gating", () => {
+    // Regression guard: cross-instance token requests build a fresh, uncached SDK client
+    // instead of reusing the shared one. That's a real behavior change for every `bw receive`
+    // call, not just cross-instance ones, so it must stay behind the flag: with the flag off,
```
harr1424: ❓ Am I reading this correctly? 

>cross-instance token requests build a fresh, uncached SDK client instead of reusing the shared one 

>That's a real behavior change for every `bw receive` call, not just cross-instance ones 

Why does this affect all `bw receive` calls? Why is ot not isolated to cross-instance access?

apps/cli/src/tools/send/commands/receive.command.ts
harr1424, line 1: ❓ Do the changes in this file re-implement the changes currently waiting for QA in #22216? 
❓ Would it make more sense to wait for those changes to land first, and then integrate this file and the spec file to support any SDK changes?

libs/common/src/auth/send-access/services/default-send-token.service.spec.ts
```diff
@@ -666,6 +666,48 @@ describe("SendTokenService", () => {
...
+    it("builds a client against the hosting instance for a credentialed request", async () => {
+      await firstValueFrom(
+        service.getSendAccessToken$(
+          sendId,
+          { kind: "password", passwordHashB64: "h4sh" as SendHashedPasswordB64 },
+          endpoints,
+        ),
+      );
+
+      expect(sdkService.ephemeralClientEndpoints).toEqual([endpoints]);
```
harr1424: 🎨 In theory it should be the same behavior, but if we add a regression guard enforcing how we expect email OTP protected Sends to behave it may pay off in the future.

libs/common/src/platform/abstractions/sdk/sdk.service.ts
```diff
@@ -94,6 +94,18 @@ export function uuidAsString<T extends Uuid>(uuid: T): string {
   return uuid as unknown as string;
 }
 
+/**
+ * Per-call endpoint overrides for {@link SdkService.createEphemeralClient}.
+ *
+ * Each omitted field falls back to the app's active environment. Use this to target a server the
+ * app is not signed in to — for example the CLI opening a Send link hosted on a different
+ * Bitwarden instance than the one it is configured against.
+ */
```
harr1424: ❓ Is this even necessary outside of the CLI?

harr1424 (commented):
I am ending my review early because a lot of these changes appear related to [PM-40120](https://bitwarden.atlassian.net/browse/PM-40120) which is still waiting for QA. 

I think it's a good idea to ensure that work passes QA before emulating it here. Additionally, that work was carefully scoped only to the CLI. This PR seems to introduce extensive changes well outside of the CLI, exclusively for the sake of supporting a code path that is only reachable in the CLI (divergent domain access).

apps/cli/src/tools/send/commands/receive.command.ts
adudek-bw: #22216 is a meaningfully more complete re-implementation. Rather than have two incompatible implementations of the same thing land on the same files, I dropped the cross-instance receive work here.

apps/cli/src/tools/send/commands/receive.command.spec.ts
adudek-bw: I reverted the whole cross-instance receive path since #22216 is a more complete implementation of the same thing :)

libs/common/src/auth/send-access/services/default-send-token.service.spec.ts
adudek-bw: Removed in favor of your implementation :)

libs/common/src/platform/abstractions/sdk/sdk.service.ts
adudek-bw: Reverted :)

apps/cli/src/tools/send/commands/create.command.spec.ts
adudek-bw: Agreed, that inconsistency (JSON wins on create, CLI flag wins on edit) predates this PR and is worth its own ticket rather than fixing silently here, it's a behavior change either way and deserves its own review.

harr1424: Tracked in https://bitwarden.atlassian.net/browse/PM-42265

harr1424 (dismissed):
Approving on the understanding that `bw receive` logic will be adjusted in [PM-42266](https://bitwarden.atlassian.net/browse/PM-42266) as discussed offline.

libs/common/src/tools/send/services/send-api-service.selector.ts
```diff
@@ -92,15 +112,12 @@ export class SendApiServiceSelector implements SendApiServiceAbstraction {
...
+   * link while signed in to a different server) is forwarded to whichever service the flag
+   * selects; both honour it.
    */
   async postSendAccess(accessToken: SendAccessToken, apiUrl?: string): Promise<SendAccessResponse> {
-    if (apiUrl != null) {
-      return this.sendApiService.postSendAccess(accessToken, apiUrl);
-    }
-    return (await this.getService()).postSendAccess(accessToken);
+    return (await this.getService()).postSendAccess(accessToken, apiUrl);
   }
```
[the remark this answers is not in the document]
adudek-bw: Confirmed, this was a real bug — verified `getApiUrl()` never returns null and traced the same over-triggering into `getSendFileDownloadData` too. Fixed in daaff0a: reverted `postSendAccess`/`getSendFileDownloadData` to their original pre-this-branch behavior — always route to legacy when `apiUrl` is supplied, regardless of the flag, since that's not a reliable cross-instance signal at this call site. Now an ordinary `bw receive` never touches `createEphemeralClient` at all.

That also resolves the secondary `ErrorResponse`/404 concern as a side effect — since these calls always go to legacy for the CLI's receive flow again, the exact same error type receive.command.ts already checks for is what actually gets thrown.

One follow-up worth flagging: with this fix, nothing in the reachable call graph passes a non-null `apiUrl` through the selector to `SendSdkApiService` anymore, so `withAccessClient`'s cross-instance branch and `createEphemeralClient` are now dead code in practice (only reachable if something bypassed the selector and called `SendSdkApiService` directly, which nothing does). Given the open question on the other thread about whether `createEphemeralClient` belongs on `SdkService` at all, I'd lean toward removing it entirely as a fast follow rather than carrying unreachable code — but wanted to flag it rather than pull that thread further in this already-large revert.

libs/common/src/platform/abstractions/sdk/sdk.service.ts
```diff
@@ -125,6 +137,21 @@ export abstract class SdkService {
...
+   * this is the escape hatch for the rare operation that must target a different server — today
+   * only the CLI's `bw receive` against a Send hosted on another Bitwarden instance. Anything
+   * running against the app's own environment should use {@link client$} instead.
+   *
+   * The returned client is not cached and not reference-counted. **The caller owns it and must
+   * dispose of it** (`using client = await createEphemeralClient(...)`, or `client.free()`).
+   *
+   * @param endpoints Endpoints to override; anything omitted comes from the active environment.
+   */
+  abstract createEphemeralClient(endpoints: SdkEndpointOverrides): Promise<PasswordManagerClient>;
```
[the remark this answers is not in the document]
djsmith85: I agree with this, and since all the logic has been reverted, this should also get removed until potentially reimplemented with the Sends receive command. Splitting receiving Sends into its own (sdk)client structure makes sense and enables switching environments via Send-context. Creating/editing Sends does not have that necessity.

djsmith85 (changes requested):

libs/common/src/platform/abstractions/sdk/sdk.service.ts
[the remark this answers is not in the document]
adudek-bw: Removed in ff49638. Simplified `SendSdkApiService.postSendAccess`/`getSendFileDownloadData` to always use the shared client, and dropped `createEphemeralClient`/`SdkEndpointOverrides`.

djsmith85 (dismissed):
Thanks for the clean up and splitting up the work.

libs/common/src/tools/send/services/send-api-service.selector.spec.ts
```diff
@@ -190,7 +234,10 @@ describe("SendApiServiceSelector", () => {
   describe("postSendAccess", () => {
     const accessToken: SendAccessToken = { token: "tok" } as SendAccessToken;
 
-    it("routes to legacy when apiUrl is supplied, even with the flag on", async () => {
+    // `apiUrl` is not a reliable cross-instance signal at the call site (see PR #22321 review
+    // discussion), so these always route to legacy regardless of the flag — matching the SDK
+    // client's inability to target anything but its own configured environment.
+    it("routes a cross-instance apiUrl to legacy even when the flag is on", async () => {
```
harr1424: ❓ This means that a client attempting to access a Send on a self-hosted environment will never be able to access that Send? As I understand it, they will see a message that the Send does not exist, since it doesn't exist on the configured environment, but this isn't exactly honest.

libs/common/src/tools/send/services/send-api.service.abstraction.ts
```diff
@@ -38,4 +39,28 @@ export abstract class SendApiService {
...
+   * Persists a send from its plaintext view, letting the implementation own encryption.
+   *
+   * Prefer this over {@link save} for new code. `save` requires the caller to encrypt first,
+   * which the SDK path cannot use: the SDK generates the send key itself, so a client-encrypted
+   * payload has to be decrypted straight back to plaintext (and a pre-encrypted file buffer is
+   * unusable outright, since its key would never match the one the SDK generates). Handing over
+   * the plaintext view lets each implementation encrypt exactly once, where it can:
+   * - legacy encrypts client-side via `SendService.encrypt`, then posts the wire form.
+   * - the SDK path forwards the view to the SDK, which encrypts under its own generated key.
+   *
```
harr1424, line 61: ❓ Should we deprecate `save()`?

libs/common/src/tools/send/services/send-sdk-api.service.ts
```diff
@@ -101,6 +115,37 @@ export class SendSdkApiService implements SendApiServiceAbstraction {
     }
   }
 
+  /**
+   * Saves a send from its plaintext view — the preferred entry point for this service.
+   *
+   * Nothing is encrypted client-side: the view goes straight to the SDK, which generates the
+   * send key and encrypts under it. This is what {@link save} cannot do, since it receives a
```
harr1424: ⛏️ This is a bit misleading since the SDK here is just WASM and so the encryption occurs just as "client-side" as it did before. Since this describes something fundamental about our security model it may be worth adjusting for clarity/accuracy.

libs/tools/send/send-ui/src/send-form/services/default-send-form.service.ts
```diff
@@ -124,15 +124,15 @@ export class DefaultSendFormService implements SendFormService {
 
     try {
       const plaintextPassword = this._updatedSendView().password;
-      const sendData = await this.sendService.encrypt(
+      // Hand over the plaintext view and let the API service encrypt: the legacy path encrypts
+      // client-side, while the SDK path encrypts under the send key it generates itself.
```
harr1424: ⛏️ Both of these situations (legacy and SDK) encrypt client-side, it may be worth updating this to describe the actual distinction. If the SDK encrypts using a key it generates itself, what does the legacy TS implementation do?

harr1424 (dismissed):
No blocking findings from me. For continuity, could you please reference the Jira ticket that will be used to implement the work described [here](https://github.com/bitwarden/clients/pull/22321#discussion_r3841566239)?

harr1424 (dismissed):

harr1424:
Looks good to me ✅   Only remaining ask is that a follow-up ticket be created in Jira and linked here for continuity.

djsmith85 (dismissed):

harr1424 (commented):
❓ Do we really want to gate the file size check behind this feature flag? How is the current file size limit enforced? 

https://github.com/bitwarden/clients/pull/22321#discussion_r3776611185

adudek-bw:
> ❓ Do we really want to gate the file size check behind this feature flag? How is the current file size limit enforced?
> 
> [#22321 (comment)](https://github.com/bitwarden/clients/pull/22321#discussion_r3776611185)

So this work is all supposed to be a part of the new SDK path, I guess I'm just hesitant to make any functional changes that cause any issues on main before QA handles all the testing.

harr1424 (approved):

djsmith85 (approved):

PR #22322, opened 2026-08-10
[PM-41801] Switch to using policyAppliesToUser$ for the Autotype Default Policy

coltonhurst:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-41801

## 📔 Objective

Switch to using `policyAppliesToUser$` for the Autotype Default Policy, per code comment recommendations. There should be no functionality changes, except exempt users should no longer have the policy applied to them.

## 🧪 Testing

1. Tested and verified the policy does not apply to Owners (an exempt user)
2. Tested and verified the policy does apply to a member (a non-exempt user)


neuronull (commented):
Where does the "code comment recommendations" live? 
Are we re-using this same policy for GA ? 
Do we have a explicit QA test cases for the policy that regression check behavior here?

coltonhurst:
> Where does the "code comment recommendations" live? Are we re-using this same policy for GA ? Do we have a explicit QA test cases for the policy that regression check behavior here?

Q1: https://github.com/bitwarden/clients/blob/main/libs/common/src/admin-console/abstractions/policy/policy.service.abstraction.ts#L14-L19

Q2: Yes, this policy will be used for both Autotype MVP & Autotype GA.

Q3: Can you clarify your question? I'm not sure what you're looking for.

neuronull:
> Q3: Can you clarify your question? I'm not sure what you're looking for.

I was just asking if we have QA test cases that cover this policy logic. I checked, and we do.

neuronull (approved):

PR #22337, opened 2026-08-11
[PM-40337] Fix desktop SSO launch behind pre-authenticating reverse proxy

coroiu:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-40337

## 📔 Objective

Desktop SSO does not work when the self-hosted server sits behind a pre-authenticating reverse proxy (e.g. Azure Entra Application Proxy).

The desktop launches the system browser at a **fragment** URL (`<vault>/#/sso?...`, built by `SsoUrlService.buildSsoUrl`). Because the web vault uses hash-based routing, the SSO parameters live after the `#` and are never sent on the wire — the browser only issues `GET /`. When the proxy has to re-challenge the user, it can only serialize the path + query it actually received, so the fragment is lost across the sign-in roundtrip. The browser lands back on the bare vault URL with an empty hash, the SPA shows the normal login page, and SSO silently never starts.

This adds an **`sso-launch-connector.html`** page: a real path + query URL that the proxy *can* observe and restore. After the proxy challenge is satisfied the browser returns to the connector with its query string intact, and the connector hands off to the `/#/sso` hash route via a same-origin, browser-local navigation (the same technique the existing `proxy-cookie-redirect` connector uses).

Both desktop launch paths use it, gated on whether the server requires bootstrap (i.e. is behind such a proxy):
- **Packaged builds** (custom-scheme `bitwarden://` callback) — `DesktopLoginComponentService`.
- **AppImage / dev builds** (localhost callback) — the `useSsoLaunchConnector` flag is threaded through the `openSsoPrompt` IPC to `SSOLocalhostCallbackService`, since that path builds its URL in the main process.

Non-proxied servers keep the direct `/#/sso` fragment URL, so there is no behavior change and no new connector dependency for them. The gate reads `ServerCommunicationConfigService.needsBootstrap$` and fails safe to the direct URL.

### Notes for reviewers
- `buildSsoUrl` output is unchanged; the connector variant (`buildSsoLaunchConnectorUrl`) shares a private query-string builder, so existing SSO entry URLs for web/browser/CLI are untouched.
- **Version skew:** the connector is a web asset, so a pre-auth server must be on a build that includes it. The client gate only selects the connector for servers advertising `ssoCookieVendor` bootstrap, so ordinary (including older self-hosted) servers are unaffected.
- **Not covered by automated tests:** the connector surviving a real Application Proxy roundtrip requires the App Proxy test environment (INT-338) to verify end to end.

## 📸 Screenshots

- [ ] TODO: capture desktop SSO completing against the INT-338 Azure App Proxy environment (before/after).

### DIFF
--- a/apps/desktop/src/auth/login/desktop-login-component.service.spec.ts
+++ b/apps/desktop/src/auth/login/desktop-login-component.service.spec.ts
@@ -1,6 +1,6 @@
 import { TestBed } from "@angular/core/testing";
 import { MockProxy, mock } from "jest-mock-extended";
-import { of } from "rxjs";
+import { of, throwError } from "rxjs";
 
 import { DefaultLoginComponentService } from "@bitwarden/auth/angular";
 import { SsoUrlService } from "@bitwarden/auth/common";
@@ -187,6 +187,18 @@ describe("DesktopLoginComponentService", () => {
         );
         expect(ssoUrlService.buildSsoLaunchConnectorUrl).toHaveBeenCalled();
         expect(ssoUrlService.buildSsoUrl).not.toHaveBeenCalled();
+        expect(platformUtilsService.launchUri).toHaveBeenCalled();
+      });
+
+      it("fails safe to the direct SSO URL when bootstrap detection errors", async () => {
+        serverCommunicationConfigService.needsBootstrap$.mockReturnValue(
+          throwError(() => new Error("detection failed")),
+        );
+
+        await service.redirectToSsoLogin("test@bitwarden.com");
+
+        expect(ssoUrlService.buildSsoUrl).toHaveBeenCalled();
+        expect(ssoUrlService.buildSsoLaunchConnectorUrl).not.toHaveBeenCalled();
         expect(platformUtilsService.launchUri).toHaveBeenCalled();
       });
 
--- a/apps/desktop/src/auth/login/desktop-login-component.service.spec.ts
+++ b/apps/desktop/src/auth/login/desktop-login-component.service.spec.ts
@@ -18,6 +18,7 @@ import { ToastService } from "@bitwarden/components";
 import { PasswordGenerationServiceAbstraction } from "@bitwarden/generator-legacy";
 
 import { ElectronPlatformUtilsService } from "../../platform/services/electron-platform-utils.service";
+import { ServerCommunicationConfigService } from "../../platform/services/server-communication-config/server-communication-config.service";
 
 import { DesktopLoginComponentService } from "./desktop-login-component.service";
 
@@ -44,6 +45,7 @@ describe("DesktopLoginComponentService", () => {
   let i18nService: MockProxy<I18nService>;
   let toastService: MockProxy<ToastService>;
   let ssoUrlService: MockProxy<SsoUrlService>;
+  let serverCommunicationConfigService: MockProxy<ServerCommunicationConfigService>;
 
   beforeEach(() => {
     cryptoFunctionService = mock<CryptoFunctionService>();
@@ -63,6 +65,9 @@ describe("DesktopLoginComponentService", () => {
     toastService = mock<ToastService>();
     platformUtilsService.getClientType.mockReturnValue(ClientType.Desktop);
     ssoUrlService = mock<SsoUrlService>();
+    serverCommunicationConfigService = mock<ServerCommunicationConfigService>();
+    // Default to a non-proxied server so SSO uses the direct fragment URL.
+    serverCommunicationConfigService.needsBootstrap$.mockReturnValue(of(false));
 
     TestBed.configureTestingModule({
       providers: [
@@ -78,6 +83,7 @@ describe("DesktopLoginComponentService", () => {
               i18nService,
               toastService,
               ssoUrlService,
+              serverCommunicationConfigService,
             ),
         },
         { provide: DefaultLoginComponentService, useExisting: DesktopLoginComponentService },
@@ -89,6 +95,10 @@ describe("DesktopLoginComponentService", () => {
         { provide: I18nService, useValue: i18nService },
         { provide: ToastService, useValue: toastService },
         { provide: SsoUrlService, useValue: ssoUrlService },
+        {
+          provide: ServerCommunicationConfigService,
+          useValue: serverCommunicationConfigService,
+        },
       ],
     });
 
@@ -136,6 +146,7 @@ describe("DesktopLoginComponentService", () => {
             codeChallenge,
             state,
             email,
+            false,
             undefined,
           );
         } else {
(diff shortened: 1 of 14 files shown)


apps/desktop/src/auth/login/desktop-login-component.service.ts
```diff
@@ -50,25 +54,51 @@ export class DesktopLoginComponentService
...
+
+    // A server that requires bootstrap sits behind a pre-authenticating reverse proxy. In that case
+    // SSO must be started via the launch connector so the launch URL is a real path + query that
+    // survives the proxy's sign-in roundtrip; a direct `/#/sso?...` fragment URL would be dropped,
+    // because the proxy only ever sees `GET /` on the wire. Every other server uses the direct URL,
+    // so this adds no connector dependency for non-proxied deployments. Both callback mechanisms
+    // below (custom-scheme and localhost) share this decision, since it concerns the launch leg.
+    const useSsoLaunchConnector = await firstValueFrom(
+      this.serverCommunicationConfigService.needsBootstrap$(Utils.getHostname(webVaultUrl)),
+    );
```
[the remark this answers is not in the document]
coroiu: Good catch — addressed in ed00600b89. `needsBootstrap$` is now wrapped with `.pipe(catchError(() => of(false)))` so a detection failure falls back to the direct SSO URL instead of silently aborting the launch, matching the "fails safe to the direct URL" behavior described in the PR. Added a regression test ("fails safe to the direct SSO URL when bootstrap detection errors") to guard it.

coroiu:
Addressed the automated review feedback in ed00600b89:

- ⚠️ Bootstrap-detection failure now fails safe to the direct SSO URL (`catchError` on `needsBootstrap$`), with a regression test. (inline thread resolved)
- ♻️ `apps/web/src/connectors/README.md` now documents the new SSO **launch** connector (`sso-launch-connector.html` / `sso-launch.ts`), distinguishes it from the callback connector, and adds a "Desktop launch behind a pre-authenticating reverse proxy" section covering the proxy-gated launch path.

apps/desktop/src/auth/login/desktop-login-component.service.spec.ts
```diff
@@ -145,6 +156,68 @@ describe("DesktopLoginComponentService", () => {
...
+          "test@bitwarden.com",
+          true,
+          undefined,
+        );
+        expect(platformUtilsService.launchUri).not.toHaveBeenCalled();
+      });
+    });
   });
 
   describe("redirectToSsoLoginWithOrganizationSsoIdentifier", () => {
```
JaredSnider-Bitwarden, line 223: 🎨 : We should add a test for ensuring the `orgSsoIdentifier` isn't dropped when building the launch connector URL on a proxied server:


```typescript
describe("pre-auth proxy launch URL selection (packaged app)", () => {
  beforeEach(() => {
    (global as any).ipc.platform.isAppImage = false;
    (global as any).ipc.platform.isDev = false;

    passwordGenerationService.generatePassword.mockResolvedValueOnce("testState");
    passwordGenerationService.generatePassword.mockResolvedValueOnce("testCodeVerifier");
    jest.spyOn(Utils, "fromBufferToUrlB64").mockReturnValue("testCodeChallenge");
  });

  it("uses the launch connector URL with orgSsoIdentifier when the server needs bootstrap", async () => {
    serverCommunicationConfigService.needsBootstrap$.mockReturnValue(of(true));

    await service.redirectToSsoLoginWithOrganizationSsoIdentifier(
      "test@bitwarden.com",
      "orgSsoId",
    );

    expect(ssoUrlService.buildSsoLaunchConnectorUrl).toHaveBeenCalledWith(
      expect.any(String),
      expect.any(String),
      expect.any(String),
      expect.any(String),
      expect.any(String),
      "test@bitwarden.com",
      "orgSsoId",
    );
    expect(ssoUrlService.buildSsoUrl).not.toHaveBeenCalled();
    expect(platformUtilsService.launchUri).toHaveBeenCalled();
  });
});
```

apps/desktop/src/auth/login/desktop-login-component.service.ts
```diff
@@ -50,25 +54,56 @@ export class DesktopLoginComponentService
...
+
+    // A server that requires bootstrap sits behind a pre-authenticating reverse proxy. In that case
+    // SSO must be started via the launch connector so the launch URL is a real path + query that
+    // survives the proxy's sign-in roundtrip; a direct `/#/sso?...` fragment URL would be dropped,
+    // because the proxy only ever sees `GET /` on the wire. Every other server uses the direct URL,
+    // so this adds no connector dependency for non-proxied deployments. Both callback mechanisms
+    // below (custom-scheme and localhost) share this decision, since it concerns the launch leg.
+    // Fail safe to the direct URL: if bootstrap detection errors (e.g. the SDK client call rejects,
+    // or `Utils.getHostname` returns null for a host `tldts` cannot parse), fall back to `false` so
+    // the SSO launch proceeds via the direct URL instead of aborting the whole flow.
```
JaredSnider-Bitwarden, line 68: Ty for the great docs across the board documenting the why.

apps/web/src/connectors/README.md
JaredSnider-Bitwarden, line 1: Thank you for updating these docs!

apps/web/src/connectors/sso-launch.ts
```diff
@@ -0,0 +1,27 @@
...
+ * client cannot launch the browser straight at that fragment URL when a pre-auth proxy is in front:
+ * the proxy only ever sees the path + query on the wire (`GET /`), so when it has to re-challenge
+ * the user it can neither capture nor restore the fragment, and the SSO parameters are silently
+ * dropped across the sign-in roundtrip.
+ *
+ * This connector is a real path + query page (`/sso-launch-connector.html?...`) that the proxy CAN
+ * observe and restore. Once the proxy challenge is satisfied the browser lands back here with its
+ * query string intact, and we hand off to the SSO hash route with a same-origin, browser-local
+ * navigation.
+ */
```
JaredSnider-Bitwarden, line 18: 🎨 : Could we try to centralize the documentation here and try to cut down the duplicative explanations elsewhere (e.g., those other docs can reference this one)?  Otherwise, it can create a bit of a maintenance burden. 


Claude details:
The same explanation appears in at least three places:

The `sso-launch.ts` JSDoc 
The inline comment block in `desktop-login-component.service.ts:60–68`
The `SsoUrlService.buildSsoLaunchConnectorUrl` JSDoc
The natural consolidation point is `sso-launch.ts` — it is the connector, so the full explanation belongs there. The other two locations only need to say why they chose this path, not re-explain the whole proxy mechanism.

JaredSnider-Bitwarden (approved):
Approving b/c all of my comments are really non-blocking. This is excellent work! Thank you for making it easy to understand why these changes were required.

dani-garcia (approved):

PR #22342, opened 2026-08-11
[PM-34808] Step 1 remove flagged logic for policy drawers

JaredScar:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-34808

## 📔 Objective

Removes flagged logic for policy drawers feature-flag


apps/web/src/app/admin-console/organizations/policies/base-policy-edit.component.ts
```diff
@@ -101,8 +101,6 @@ export abstract class BasePolicyEditDefinition {
    * Optional drawer-specific configuration for this policy.
    * When set, {@link PolicyEditDrawerComponent} is used in place of the standard
    * modal dialog, loading {@link v2.component} and rendering the drawer-specific layout.
-   * Drawer routing is gated globally by {@link FeatureFlag.PolicyDrawers} in
-   * {@link PoliciesComponent} — there is no per-policy flag.
    */
   v2?: {
```
eliykat: The nested v2 configuration was only to support the policy drawer feature flag. Now that the flag is being removed, it should be folded back into the main configuration.

This will have a larger blast radius and I was debating whether to do it here or in a follow-up PR. But without removing dead code and old implementations, the extent of the feature flag removal here is unclear. I think it's worth doing both together.

apps/web/src/app/admin-console/organizations/policies/policies.component.ts
```diff
@@ -188,7 +182,7 @@ export class PoliciesComponent {
     const dialogComponent: PolicyDialogComponent =
       policy.editDialogComponent ?? PolicyEditDialogComponent;
 
-    const drawerOpener = this.useDrawer() ? dialogComponent.openDrawer : undefined;
+    const drawerOpener = dialogComponent.openDrawer;
 
     if (drawerOpener) {
```
eliykat: This logic should also be able to be streamlined - the conditional looks like it's handling dialog vs. drawer behavior. The old dialog should be deleted, and then there should only be 1 way to open a drawer.

apps/web/src/app/admin-console/organizations/policies/policy-edit-dialog.component.ts
```diff
@@ -159,7 +154,7 @@ export class PolicyEditDialogComponent implements AfterViewInit {
       });
 
     // For modals, only the beforeunload guard is needed — closePredicate is drawer-only.
-    if (!this.useDrawer() || !this.dialogRef.isDrawer) {
+    if (!this.dialogRef.isDrawer) {
```
eliykat: Again, still supporting modal-only logic.

bitwarden_license/bit-web/src/app/key-management/policies/session-timeout.component.ts
```diff
@@ -39,8 +38,8 @@ export class SessionTimeoutPolicy extends BasePolicyEditDefinition {
   priority = 70;
   component = SessionTimeoutPolicyComponent;
 
-  display$(_organization: Organization, configService: ConfigService) {
-    return configService.getFeatureFlag$(FeatureFlag.PolicyDrawers).pipe(map((v: boolean) => !v));
+  display$(_organization: Organization, _configService: ConfigService) {
+    return of(false);
```
eliykat: This component should be deleted as it is now never used. There will be other dead code components as well. (e.g. ones that now use the simple toggle component)

eliykat (changes requested):
This is narrowly removing the feature flag, but we should also remove all the old dialog infrastructure and components as well. We should only be left with the new drawer components and behavior.

amorask-bitwarden (dismissed):
Billing changes look good.

nick-livefront (dismissed):
Vault ✅

amorask-bitwarden (dismissed):

nick-livefront (dismissed):

apps/web/src/app/admin-console/organizations/policies/policy-edit-definitions/single-org.component.ts
```diff
@@ -1,30 +1,25 @@
...
+  description = "singleOrgPolicyDescV2";
   type = PolicyType.SingleOrg;
   category = PolicyCategory.DataControl;
   priority = 10;
   component = SingleOrgPolicyComponent;
-  v2 = {
-    component: SimpleTogglePolicyComponent,
-    description: "singleOrgPolicyDescV2",
-  };
 }
```
eliykat: This was using the v2 SimpleTogglePolicyComponent, now it's using the bespoke `SingleOrgPolicyComponent` again. Regression?

apps/web/src/app/admin-console/organizations/policies/policy-edit-definitions/single-org.component.html
eliykat, line 1: This looks like it's been updated to match the latest style rather than using SimpleToggle.

bitwarden_license/bit-web/src/app/admin-console/policies/policy-edit-register.spec.ts
eliykat, line 1: This spec file was only for v1/v2 regressions - now that we're dropping v2 we don't need it any more. It can be deleted.

apps/web/src/app/admin-console/organizations/policies/policy-edit-register.spec.ts
eliykat, line 1: Same here - can be deleted.

apps/web/src/app/admin-console/organizations/policies/policy-edit-definitions/auto-confirm-policy.component.html
```diff
@@ -1,16 +1,10 @@
...
-      {{ (showBadge ? "autoConfirm" : "editPolicy") | i18n }}
-      @if (!showBadge) {
-        <span class="tw-text-muted tw-font-normal tw-text-sm">
-          {{ policy()!.name | i18n }}
-        </span>
-      }
+      {{ policy()!.name | i18n }}
     </span>
   </div>
 </ng-template>
```
eliykat, line 10: This was not present in the v2 component. Was that a mistake and it should be added back here, or was that correct and it should be deleted here?

Also - I think this renders in a dialog from the main vault page if the org has had the feature enabled by customer success. (It's opt-in on request.) Have we checked that the drawer component still renders in that dialog format?

apps/web/src/app/admin-console/organizations/policies/policy-edit-definitions/auto-confirm-policy.component.spec.ts
```diff
@@ -340,14 +290,11 @@ describe("AutoConfirmPolicyEditV2Component", () => {
...
+      expect(component.policySteps[0].titleContent).toBe((component as any).step0Title);
       expect(component.policySteps[1].titleContent).toBe((component as any).step1Title);
       expect(component.policySteps[1].bodyContent).toBe((component as any).step1Content);
       expect(component.policySteps[1].footerContent).toBe((component as any).step1Footer);
     });
-
-    it("does not set a custom titleContent for step 0, relying on the dialog's generic name+badge header", () => {
-      expect(component.policySteps[0].titleContent).toBeUndefined();
-    });
   });
```
eliykat, line 307: Changed test case asserts new behavior, but unclear if that's intended.

apps/web/src/app/vault/services/web-vault-prompt.service.ts
```diff
@@ -69,7 +69,7 @@ export class WebVaultPromptService {
   }
 
   private openAutoConfirmFeatureDialog(organization: Organization) {
-    MultiStepPolicyEditDialogComponent.open(this.dialogService, {
+    void MultiStepPolicyEditDialogComponent.openDrawer(this.dialogService, {
```
eliykat: This is where the auto-confirm policy is opened from the vault page. Is `openDrawer` correct here? Please check with Jillian.

apps/web/src/app/admin-console/organizations/policies/policy-drawer-story.helper.ts
```diff
@@ -94,7 +88,7 @@ function buildPolicyDialogMeta(
           },
           {
             provide: DialogRef,
-            useValue: { isDrawer, close: () => Promise.resolve(), closePredicate: undefined },
+            useValue: { isDrawer: true, close: () => Promise.resolve(), closePredicate: undefined },
```
eliykat, line 90: `isDrawer` value should no longer be required.

apps/web/src/app/admin-console/organizations/policies/policy-drawer-story.helper.ts
```diff
@@ -135,10 +129,10 @@ function buildPolicyDialogMeta(
             useValue: { encryptString: () => Promise.resolve({ encryptedString: "encrypted" }) },
           },
           {
-            // Only PolicyEditDialogComponent/MultiStepPolicyEditDialogComponent inject this
-            // directly, but providing it unconditionally is harmless for the drawer component.
+            // Only MultiStepPolicyEditDialogComponent-based policies inject this, but providing it
+            // unconditionally is harmless for the plain drawer component.
             provide: ConfigService,
-            useValue: { getFeatureFlag$: () => of(isDrawer) },
+            useValue: { getFeatureFlag$: () => of(true) },
```
eliykat: Should no longer be required.

eliykat (changes requested):
Nice work, it'll be great to have all this cleaned up. A few comments below, also make sure to review Storybook test changes before merge.

mzieniukbw (dismissed):

apps/web/src/app/admin-console/organizations/policies/policy-drawer-story.helper.ts
JaredScar: We still use `isDrawer` because it's part of the `bit-dialog`'s `DialogComponent` which reads `dialogRef.isDrawer` in two places in the stories. It tells the actual DialogRef that this is a Drawer, so use the Drawer styling classes for it. The stories still rely on this

apps/web/src/app/vault/services/web-vault-prompt.service.ts
JaredScar: Choosing to revert this because we should keep OG behavior rather than change this. If they want this changed in the future, it's a separate ticket

apps/web/src/app/admin-console/organizations/policies/policy-edit-definitions/single-org.component.ts
```diff
@@ -1,36 +1,16 @@
...
-
-    if (!this.policyResponse()) {
-      throw new Error("Policies not found");
-    }
-    if (!this.policyResponse()!.canToggleState) {
-      this.enabled.disable();
-    }
-  }
+  component = SimpleTogglePolicyComponent;
+  warningKey = "singleOrgPolicyMemberWarning";
```
eliykat, line 15: The `warningKey` was not previously used in the v2 component; it was in v1 only. We're not changing any v2 behaviour here so I assume it should be removed.

apps/web/src/app/admin-console/organizations/policies/policy-edit-dialogs/multi-step-policy-edit-modal.component.html
eliykat, line 1: Unfortunate that we still have to maintain this. As discussed, please check with Jillian whether they want to migrate away from this as well.

apps/web/src/app/admin-console/organizations/policies/policy-edit-definitions/auto-confirm-policy.component.ts
```diff
@@ -39,10 +39,6 @@ export class AutoConfirmPolicy extends BasePolicyEditDefinition {
   showDescription = false;
   editDialogComponent = MultiStepPolicyEditDialogComponent;
 
-  constructor(readonly firstTimeDialog: boolean = false) {
-    super();
-  }
```
eliykat, line 51: This `firstTimeDialog` property is used to show some extra content when it's being opened from the vault page. We need to maintain that unless design have said otherwise. This was in the v1 template but not the v2, however I assume the web vault was continuing to open the v1 template all this time, so the change wasn't visible. Now that we're getting rid of v1 we'd better make sure the behavior persists.

apps/web/src/app/vault/services/web-vault-prompt.service.ts
```diff
@@ -69,9 +69,12 @@ export class WebVaultPromptService {
   }
 
   private openAutoConfirmFeatureDialog(organization: Organization) {
-    void MultiStepPolicyEditDialogComponent.openDrawer(this.dialogService, {
+    // Opened as a modal (not a drawer) - this is a proactive onboarding prompt shown on the
+    // vault page, not a policy being edited from the Admin Console policies page, so it should
+    // remain an attention-grabbing modal rather than a side drawer.
```
eliykat, line 74: Please verify that the v2 component renders in the dialog OK, because I don't think we've had this combination before (it was v1 + dialog or v2 + drawer, not v1 + drawer).

apps/web/src/app/admin-console/organizations/policies/policy-edit-drawer.component.ts
```diff
@@ -181,7 +195,7 @@ export class PolicyEditDrawerComponent implements AfterViewInit {
     this.setupDiscardGuard();
   }
 
-  private policyDataHasChanged(oldPolicyData: any, newPolicyData: any) {
+  protected policyDataHasChanged(oldPolicyData: any, newPolicyData: any) {
```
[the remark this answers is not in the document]
eliykat: This Claude [comment](https://github.com/bitwarden/clients/pull/22342#discussion_r3785898906) looks correct to me. I remember we had some regressions around this, so it's possible a fix was applied to the dialog while we were still developing the drawer.

eliykat (dismissed):
I'm OOO next week, so please follow up the team for the final review. As long as we're sure there are no regressions around the items below, we should be good to go.

apps/web/src/app/admin-console/organizations/policies/policy-edit-definitions/single-org.component.ts
JaredScar: `warningKey` is intentioned here. Before the code changes here, the SingleOrgPolicy was directly rendering the warning on the template file so warningKey was never needed. warningKey preserves the warning for v2. Removing it here would actually drop the warning key from the drawer

apps/web/src/app/vault/services/web-vault-prompt.service.ts
JaredScar: This is a good callout. I added test coverage for this now, but it does render correctly :D

amorask-bitwarden (approved):

mzieniukbw (approved):

nick-livefront (approved):

sven-bitwarden (approved):

PR #22246, opened 2026-08-06
[PM-35946] Browser: Health report at risk category detail view

lastbestdev:
## 🎟️ Tracking
https://bitwarden.atlassian.net/browse/PM-35946

## 📔 Objective
The Browser Health feature includes a detail view for each of the at-risk categories (exposed, weak, reused), where the user is able to view the ciphers that have landed in each category. This PR implements the new view complete with support for all 3 categories, using just one component. The correct content is decided by choosing the corresponding category when routing into the view.

## 📸 Screenshots

https://github.com/user-attachments/assets/42ca8726-982f-4b87-9038-23cb99e87475

PR #22235, opened 2026-08-05
[PM-35945] Browser: Show the Health Overview after a scan (gauge and risk categories)

AlexRubik:
## 🎟️ Tracking

[PM-35945](https://bitwarden.atlassian.net/browse/PM-35945)

## 📔 Objective

Builds the Health Overview a premium user sees on the extension's Health tab after a scan: the At-Risk Gauge with its heading and count, plus the Exposed, Weak, and Reused categories with deduplicated counts.

Mostly composition, since everything it composes is already merged. The overview and a reusable category row are new components in `bitwarden_license/bit-browser/src/popup/dirt/health/`; the licensed shell imports the overview directly, replacing the `RESULTS PLACEHOLDER` from [#22209](https://github.com/bitwarden/clients/pull/22209). Only open-source file touched is `messages.json`.

Layout and copy match the Figma frame. Copy is still worth a content pass before merge, since locale keys are immutable once shipped.

### Notes for reviewers

| Topic | Detail |
|---|---|
| Interim scan trigger | The report service is compute-only, so the overview scans on init and takes only the first cipher emission, keeping the breach lookup off every vault edit. One `TODO` block for [PM-39223](https://bitwarden.atlassian.net/browse/PM-39223) to lift out; a test pins the scan-once behaviour. |
| Blank until the report resolves | A green gauge mid-scan would mislead. Errors are logged and nothing renders; the failure screen is [PM-39223](https://bitwarden.atlassian.net/browse/PM-39223)'s. |
| Rows link to `/health/{category}` | Matches the extension's top-level drill-down convention. Those routes land with [PM-35946](https://bitwarden.atlassian.net/browse/PM-35946), so taps log `NG04002` and do not navigate yet. |
| No premium gate | `healthEnabled$` checks the flag and org tier, not premium. That branch is [PM-35944](https://bitwarden.atlassian.net/browse/PM-35944)'s; flagged for QA. |
| `I18nPipe` over `JslibModule` | Matches the shell and gauge beside it. |

**Verification:** 61 unit tests, ESLint and `test:locales` clean, and 25 end-to-end checks in a real browser across both states, compact mode, and dark mode.

## 📸 Screenshots

1-at-risk
2-healthy
3-healthy-dark
4-compact-mode

[...]

PR #22346, opened 2026-08-11
[PM-39223] Browser: Run the vault scan and show its progress and failure states

AlexRubik:
## 🎟️ Tracking

[PM-39223](https://bitwarden.atlassian.net/browse/PM-39223)

## 📔 Objective

Runs the browser extension's vault-health scan from the Health tab root and gives it a progress view and a failure view.

The scan pipeline previously lived inside the Health Overview as a stopgap. Because a failed scan resolved to `null`, a scan in flight and a failed scan both rendered an empty page and were indistinguishable. Moving the scan to the root lets the tab render one of four states: intro, scan progress, scan failure, or overview. The Health Overview becomes presentational: it takes the finished report as a required input and fetches nothing, dropping four injected dependencies, its RxJS pipeline, and its null guard.

The scan starts from the intro's "Scan my vault" on the first visit and automatically on later popup opens. There is no manual rescan control. Within a popup session the report is reused rather than regenerated, so re-entering the tab does not re-scan; closing the popup discards it.

Two pipeline properties are load-bearing and were kept with their regression tests: the null-cipher guard before `take(1)` (`cipherViews$` is shareReplay-cached and can replay `null` first, which would scan an empty vault and strand the user on a permanent "healthy" reading), and `take(1)` itself, so a vault edit does not re-run the breach lookup.

Builds on [#22235](https://github.com/bitwarden/clients/pull/22235).

### Addressed in review

The report service now exposes one status constant and one flat `{ status, report }` state through a single accessor, with a per-user stream so an account switch cannot cross results. The tab reads that state as a plain `toSignal` and keeps the scan trigger (with its reuse guard) as a small RxJS chain. Progress-bar width and heading weight were aligned to the design spec; bar thickness needs a `bit-progress-bar` size input, filed as a follow-up.

### Notes for reviewers

- **The progress bar is animated, not driven by real scan progress, and this is agreed.** The scan is a single `await` with no progress events, so it eases toward a ceiling and is swapped out when the scan resolves. Assistive tech gets the status, not the number: the bar is named by the heading and announces "In progress". Confirmed with Max on 2026-08-18; he asked for failure logging instead, which the failure path now does.
- **The failure view has no "Try again" button, though the frame shows one.** The acceptance criteria say the tab has no manual rescan control, and they govern; reopening the popup re-runs the scan. Flagging it as a deliberate divergence.
- This touches `health.component.html`, which [#22246](https://github.com/bitwarden/clients/pull/22246) also edits. Whichever merges second needs a small rebase, dropping the temporary nav buttons there.

## 📸 Screenshots
s2-scanning

s2-overview

s5-failure-view

[PM-39223]: https://bitwarden.atlassian.net/browse/PM-39223?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


bitwarden_license/bit-browser/src/popup/dirt/health/health-scan-error.component.html
```diff
@@ -0,0 +1,16 @@
...
+    <bit-svg [content]="illustration" aria-hidden="true"></bit-svg>
+  </div>
+
+  <h2 bitTypography="body1" class="tw-mb-1 tw-font-semibold tw-text-fg-heading">
+    {{ "healthScanFailed" | i18n }}
+  </h2>
+
+  <p bitTypography="body2" class="tw-mb-0 tw-text-fg-body-subtle">
+    {{ "healthScanFailedDescription" | i18n }}
+  </p>
```
lastbestdev, line 15: Ticket seems to imply we no longer want the "Try again" button on the failure state, but noting here just in case it should be included

bitwarden_license/bit-browser/src/popup/dirt/health/health-scan-error.component.ts
```diff
@@ -0,0 +1,25 @@
...
+import { I18nPipe } from "@bitwarden/ui-common";
+
+/**
+ * The Health tab's failure view, shown when the vault-health scan does not
+ * complete. Presentational.
+ *
+ * There is no retry control. The design frame shows a "Try again" button, but
+ * the acceptance criteria state the tab has no manual rescan control, and the
+ * criteria govern. The scan runs automatically on every Health tab open, so
+ * reopening the tab is the recovery path.
```
lastbestdev: Nit: we can remove this section of the comment

bitwarden_license/bit-browser/src/popup/dirt/health/health-scan-error.component.spec.ts
```diff
@@ -0,0 +1,46 @@
...
+    expect(fixture.nativeElement.querySelector('[role="alert"]')).not.toBeNull();
+  });
+
+  it("does not offer a rescan control", () => {
+    // Deliberate, and deliberately against the design frame: that frame shows a
+    // "Try again" button, but the acceptance criteria state the tab has no
+    // manual rescan control, and the criteria govern. Reopening the tab is the
+    // recovery path, since the scan runs on every open.
+    expect(fixture.nativeElement.querySelector("button")).toBeNull();
+  });
```
lastbestdev: Generally we want to avoid writing tests that assert what things _don't do_. Having a test go red later because someone adds a button to the component isn't adding value IMO.

Valuable tests cover functionality of the test subject that _does_ exist, so that should the functionality change later, we have a test go red. The other tests in this suite do a good job of that. 👍🏻

bitwarden_license/bit-browser/src/popup/dirt/health/health-scanning.component.ts
```diff
@@ -0,0 +1,45 @@
...
+})
+export class HealthScanningComponent {
+  protected readonly progress = signal(START);
+
+  constructor() {
+    const tick = setInterval(
+      () => this.progress.update((v) => v + (CEILING - v) * EASING),
+      TICK_MS,
+    );
+    inject(DestroyRef).onDestroy(() => clearInterval(tick));
```
lastbestdev, line 43: I see that the [ticket](https://bitwarden.atlassian.net/browse/PM-39223) does not specify that the progress bar should reflect the actual progress of generating the report. I don't recall it being highlighted from product/design at any point either.

I think we should get clarity and sign-off around whether the progress bar should be "real" or not, and proceed from there.

bitwarden_license/bit-browser/src/popup/dirt/health/health-scan-error.component.ts
AlexRubik: https://github.com/bitwarden/clients/pull/22346/commits/2eb6e0615e78978a1aad2e6077bba49fbdd5a7fc

bitwarden_license/bit-browser/src/popup/dirt/health/health-scan-error.component.spec.ts
AlexRubik: https://github.com/bitwarden/clients/pull/22346/commits/2eb6e0615e78978a1aad2e6077bba49fbdd5a7fc

bitwarden_license/bit-browser/src/popup/dirt/health/health.component.ts
```diff
@@ -51,6 +70,45 @@ export class HealthComponent {
...
+        ),
+      ),
+    ),
+    { initialValue: null },
+  );
+
+  /** True while the scan is in flight, including before it reports progress. */
+  protected readonly isScanning = computed(
+    () => (this.scan()?.status ?? "scanning") === "scanning",
+  );
```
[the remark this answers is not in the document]
AlexRubik: https://github.com/bitwarden/clients/pull/22346/commits/2b2ca7cd8bd60be9a58e0445a254f349adfc2c7e

bitwarden_license/bit-browser/src/popup/dirt/health/health-scanning.component.ts
AlexRubik: Clarified with Max. Current animated implementation is fine for now but we'd like to have logging if it doesn't already exist to make debugging easier if report processes fail.

bitwarden_license/bit-browser/src/popup/dirt/health/health.component.ts
```diff
@@ -74,4 +131,46 @@ export class HealthComponent {
...
+        ),
+      ),
+      map((report): HealthScanState => ({ status: "success", report })),
+      catchError((error: unknown): Observable<HealthScanState> => {
+        // Message included so a failed scan is identifiable in a log dump;
+        // asked for in review to make report failures easier to debug.
+        this.logService.error("Vault health scan failed", error);
+        return of({ status: "error" });
+      }),
+      startWith<HealthScanState>({ status: "scanning" }),
```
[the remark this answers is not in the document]
AlexRubik: Measured it: instrumented counters show the back arrow does not re-create `HealthComponent`. Constructor runs once, `buildVaultHealthReport` once, no breach lookups after the back press. Overview returns in 4ms, no progress view. Likely `popupRouterCacheService`. No change needed.

bitwarden_license/bit-browser/src/popup/dirt/health/health.component.ts
```diff
@@ -74,4 +131,46 @@ export class HealthComponent {
...
+   * a failure. Never errors — a failed scan is a state, not an exception, so the
+   * tab can render the failure view instead of tearing down the pipeline.
+   */
+  private runScan$(userId: UserId): Observable<HealthScanState> {
+    return this.cipherService.cipherViews$(userId).pipe(
+      // cipherViews$ is shareReplay-cached with refCount: false and emits null
+      // when the decrypted ciphers are cleared, so a fresh subscriber can
+      // receive null FIRST. Scanning it reports an empty vault and, because
+      // take(1) then completes, strands the user on a permanent "healthy"
+      // reading.
```
lastbestdev: Nit: implementation details of the source observable can be read in the implementation. if we want to keep I think it suffices to just note that `cipherViews$ may emit null when decrypted ciphers are cleared`

bitwarden_license/bit-browser/src/popup/dirt/health/health.component.ts
```diff
@@ -51,6 +70,44 @@ export class HealthComponent {
     { initialValue: false },
   );
 
+  /**
+   * The vault scan's lifecycle, or null before it has been started.
+   *
+   * Read through a single `toSignal`, so there is exactly one subscription no
+   * matter how many times the template reads it. The report service caches, so
+   * a second read would no longer repeat the breach lookup, but a second
+   * subscription would re-enter the pipeline and trigger a second scan.
```
lastbestdev: Nit: it appears this is calling out an issue from development again. This comment can be removed to reduce bloat.

bitwarden_license/bit-browser/src/popup/dirt/health/health.component.ts
```diff
@@ -51,6 +70,44 @@ export class HealthComponent {
...
+   */
+  private readonly scan = toSignal<HealthScanState | null>(
```
lastbestdev: Nit: could we update to `scanState` for the name here?

bitwarden_license/bit-browser/src/popup/dirt/health/health.component.ts
```diff
@@ -74,4 +131,46 @@ export class HealthComponent {
...
+            this.vaultHealthReportService
+              .getVaultHealthReport$(userId)
+              .pipe(filterOutNullish(), take(1)),
+          ),
+        ),
+      ),
+      map((report): HealthScanState => ({ status: "success", report })),
+      catchError((error: unknown): Observable<HealthScanState> => {
+        // Message included so a failed scan is identifiable in a log dump;
+        // asked for in review to make report failures easier to debug.
```
lastbestdev: Nit: comment bloat, can be removed

bitwarden_license/bit-browser/src/popup/dirt/health/health.component.ts
```diff
@@ -74,4 +131,46 @@ export class HealthComponent {
...
+      // reading.
+      filterOutNullish(),
+      // The scan does an external breach lookup; a vault edit must not re-run it.
+      take(1),
+      switchMap((ciphers) =>
+        // buildVaultHealthReport publishes the report into the service rather
+        // than returning it, so the category detail views can still read it
+        // once this tab is destroyed (/health/:category is a sibling route, not
+        // a child, so it cannot inherit the result from here). Rejections still
+        // propagate, which is what drives the failure state below.
```
lastbestdev: Nit: this can probably be removed. don't see much value added in keeping

bitwarden_license/bit-browser/src/popup/dirt/health/health.component.ts
```diff
@@ -1,18 +1,32 @@
...
 import { I18nPipe } from "@bitwarden/ui-common";
 
 import { HealthIntroComponent } from "./health-intro.component";
 import { HealthOverviewComponent } from "./health-overview.component";
+import { HealthScanErrorComponent } from "./health-scan-error.component";
+import { HealthScanningComponent } from "./health-scanning.component";
 import { HealthAccessService } from "./services/health-access.service";
 
+/** Where the Health tab's vault scan is in its lifecycle. */
+type HealthScanState =
```
lastbestdev: Important: we are going to want the `VaultHealthReportService` to own the statuses (i.e. loading, error, success) associated with generating the Health report, rather than this component.

Reason being, if any other components (ex. the at risk category detail view) or services want to reuse the Health report feature, they will want to be able to read the statuses while the report is generated as well. 

Then this component only becomes responsible for:
1. Trigger report generation on component load or when user changes
2. Read report state/results from `VaultHealthReportService`

lastbestdev (changes requested):
The VaultHealthReportService should be responsible for managing/publishing the states surrounding report generation (i.e. `loading, error, success`)

bitwarden_license/bit-browser/src/popup/dirt/health/health.component.ts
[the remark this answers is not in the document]
AlexRubik: Correcting my earlier reply: it does happen. Re-measured in real Chrome, the back arrow added 4 breach lookups. Fixed in https://github.com/bitwarden/clients/pull/22346/commits/58cce7fdc5f128ca30f9ecfe3bf8a5eb8b564099, re-measured at 0, plus a follow-up so a failed scan still retries.

bitwarden_license/bit-browser/src/popup/dirt/health/health.component.ts
```diff
@@ -74,4 +146,64 @@ export class HealthComponent {
...
+   * a sibling route and the popup never reuses routes, so the back arrow
+   * destroys and rebuilds this component; without the guard the user pays a
+   * second breach lookup and watches the progress view again for results they
+   * were reading a moment earlier. The report service is popup-scoped, so
+   * closing the popup still discards everything and the next open scans afresh.
+   *
+   * Never errors, so a failure renders the failure view instead of tearing down
+   * the pipeline.
+   */
+  private generateReport$(userId: UserId): Observable<VaultHealthReportState> {
```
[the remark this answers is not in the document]
AlexRubik: Description fixed, both sentences (you caught one, there was a second). The reuse is deliberate in https://github.com/bitwarden/clients/pull/22346/commits/58cce7fdc5f128ca30f9ecfe3bf8a5eb8b564099 but its scope is wider than back-navigation, so the AC question goes to Product.

bitwarden_license/bit-browser/src/popup/dirt/health/health.component.ts
AlexRubik: https://github.com/bitwarden/clients/pull/22346/commits/67463062ee838019ec8a896038c6edf9d53ce1e8

bitwarden_license/bit-browser/src/popup/dirt/health/health.component.ts
AlexRubik: https://github.com/bitwarden/clients/pull/22346/commits/67463062ee838019ec8a896038c6edf9d53ce1e8

bitwarden_license/bit-browser/src/popup/dirt/health/health.component.ts
AlexRubik: https://github.com/bitwarden/clients/pull/22346/commits/67463062ee838019ec8a896038c6edf9d53ce1e8

bitwarden_license/bit-browser/src/popup/dirt/health/health.component.ts
AlexRubik: https://github.com/bitwarden/clients/pull/22346/commits/67463062ee838019ec8a896038c6edf9d53ce1e8

bitwarden_license/bit-browser/src/popup/dirt/health/health.component.ts
AlexRubik: https://github.com/bitwarden/clients/pull/22346/commits/67463062ee838019ec8a896038c6edf9d53ce1e8

bitwarden_license/bit-browser/src/popup/dirt/health/health.component.ts
AlexRubik: Done in [ca2ca1d726](https://github.com/bitwarden/clients/pull/22346/commits/ca2ca1d7269800874327fecb0ffb4e6cc0816592). The service now owns loading/error/success and publishes them per user, so other consumers can read them. This component only triggers the scan and reads state.

Two side effects: the report is kept during a rescan so the detail view isn't kicked out, and switching accounts mid-scan needed a guard ([ce8aa3ee99](https://github.com/bitwarden/clients/pull/22346/commits/ce8aa3ee99b40c6755794fe185cc4ba6dbab2353)).

bitwarden_license/bit-browser/src/popup/dirt/health/health.component.ts
```diff
@@ -51,6 +86,43 @@ export class HealthComponent {
     { initialValue: false },
   );
 
+  /**
+   * Where report generation is for the active user, per the report service.
+   *
+   * Annotated rather than passed as a type argument: `toSignal`'s only overload
+   * that accepts a non-null `initialValue` declares two type parameters, so one
+   * explicit argument disqualifies it on arity.
```
lastbestdev: Nit: remove

bitwarden_license/bit-common/src/dirt/vault-health/services/implementations/default-vault-health-report.service.ts
```diff
@@ -8,38 +9,112 @@ import { CipherRiskResult } from "@bitwarden/sdk-internal";
...
+ * HealthRiskCategoryDetailComponent routes away from itself on a null report, so
+ * dropping it mid-rescan would eject the user from a category they are reading.
+ * Both reads derive from this single value, so every subscriber agrees no matter
+ * when it subscribed.
+ */
+type ScopedState = {
+  userId: UserId;
+  state: VaultHealthReportState;
+  report: VaultHealthReportView | null;
+};
```
lastbestdev: Instead of keeping two instances of the `report` in the published state from the service, can we remove `state.report` and keep just the various status states? It may be good to rename `state => status` as well. This avoids confusion for consumers of the service, as they won't have to choose which `report` to read from/use.

This would look more like:

```suggestion
type ScopedState = {
  userId: UserId;
  status: VaultHealthReportStatus;
  report: VaultHealthReportView | null;
};
```

Where values of `VaultHealthReportStatus` would look like:
```
// in vault-health-report-status.ts

export const VaultHealthReportStatus = Object.freeze({
  Success: "success",
  Error: "error", 
  Loading: "loading",
  Idle: "idle"
} as const);

export type VaultHealthReportStatus = (typeof VaultHealthReportStatus)[keyof typeof VaultHealthReportStatus];

File: bitwarden_license/bit-common/src/dirt/vault-health/models/vault-health-report-state.ts [hunk 16]
```diff
@@ -0,0 +1,22 @@
+import { VaultHealthReportView } from "./view/vault-health-report.view";
+
+/**
+ * Where report generation is for a given user.
+ *
+ * Published by VaultHealthReportService so every Health page can tell a scan
+ * that is still running from one that failed, rather than reading both as a
+ * missing report.
+ */
+export type VaultHealthReportState =
```
lastbestdev: See other comment for suggested changes to this type/constant.

bitwarden_license/bit-common/src/dirt/vault-health/services/implementations/default-vault-health-report.service.ts
lastbestdev: This also allows us to import and use values from the `VaultHealthReportStatus` constant instead of using "magic strings" for the status values. This reduces the likelihood of bugs and is good practice in general

bitwarden_license/bit-browser/src/popup/dirt/health/health.component.ts
```diff
@@ -51,6 +86,43 @@ export class HealthComponent {
...
+    { initialValue: VAULT_HEALTH_REPORT_IDLE },
+  );
+
+  /** True when report generation did not complete. */
+  protected readonly scanFailed = computed(() => this.scanState().status === "error");
+
+  /** The completed report, or null while generating or after a failure. */
+  protected readonly report = computed(() => {
+    const state = this.scanState();
+    return state.status === "success" ? state.report : null;
```
lastbestdev: A few examples here of where we would use the `VaultHealthReportStatus` constant values (see "error" and "success")

bitwarden_license/bit-browser/src/popup/dirt/health/health-scan-error.component.html
```diff
@@ -0,0 +1,16 @@
+<div
+  class="tw-flex tw-h-full tw-flex-col tw-items-center tw-justify-center tw-px-6 tw-text-center"
+  role="alert"
+>
+  <div class="tw-mb-4 tw-size-24 tw-content-center">
+    <bit-svg [content]="illustration" aria-hidden="true"></bit-svg>
+  </div>
+
+  <h2 bitTypography="body1" class="tw-mb-1 tw-font-semibold tw-text-fg-heading">
```
lastbestdev: This should be styled as an h5 per design spec

```suggestion
  <h2 bitTypography="h5" class="tw-mb-1 tw-font-semibold tw-text-fg-heading">
```

bitwarden_license/bit-browser/src/popup/dirt/health/health-scanning.component.html
```diff
@@ -0,0 +1,29 @@
...
+
+         ariaValueText must differ from that heading. A screen reader announces
+         the bar's name and then its value, so reusing the heading here would
+         speak the same phrase twice from two ARIA slots. -->
+    <bit-progress-bar
+      [value]="progress()"
+      [hideStartHint]="true"
+      [ariaValueText]="'scanInProgress' | i18n"
+    >
+      <h2 slot="label" bitTypography="body1" class="tw-mb-2 tw-font-semibold tw-text-fg-heading">
```
lastbestdev: Same here 
```suggestion
      <h2 slot="label" bitTypography="h5" class="tw-mb-2 tw-font-semibold tw-text-fg-heading">
```

bitwarden_license/bit-browser/src/popup/dirt/health/health-scanning.component.html
```diff
@@ -0,0 +1,29 @@
...
+  <div class="tw-w-full">
+    <!-- The heading is projected as the bar's label rather than sitting outside
+         it. Left unprojected, the bar renders its own screen-reader-only copy of
+         accessibleName, which this live region would announce alongside the
+         visible heading.
+
+         ariaValueText must differ from that heading. A screen reader announces
+         the bar's name and then its value, so reusing the heading here would
+         speak the same phrase twice from two ARIA slots. -->
+    <bit-progress-bar
```
lastbestdev, line 15: This progress bar is both narrower and wider than the design spec suggests it should be. Here is the expected

Image

bitwarden_license/bit-common/src/dirt/vault-health/services/implementations/default-vault-health-report.service.ts
```diff
@@ -8,38 +9,112 @@ import { CipherRiskResult } from "@bitwarden/sdk-internal";
...
+  /**
+   * Counts builds so a superseded one cannot publish.
+   *
+   * A build cannot be cancelled once its promise is in flight, so switching
+   * account mid-generation leaves the abandoned build running. Without this it
+   * would overwrite the new account's state on resolve, and because only one
+   * build runs per Health tab open, nothing would publish again: the tab would
+   * sit on the progress view indefinitely.
+   */
+  private generation = 0;
```
lastbestdev: Instead of using this counter based check to prevent states from publishing over each other async, what do you think about updating the internal `state` to become a map of UserId => Subject, so that each user's Health report results can be published into their own stream in this service?

This will help prevent complexity that arises from handling account switching, for example we will not have to worry about the results of a scan for User A overwriting the scan for User B due to async timing issues. The services creates a subject for each user to publish into.

bitwarden_license/bit-common/src/dirt/vault-health/services/abstractions/vault-health-report.service.ts
```diff
@@ -13,22 +13,35 @@ import { VaultHealthReportView } from "../../models/view/vault-health-report.vie
...
+
   /** Get the latest vault health scan report for a user, run buildVaultHealthReport first to generate the report.
-   * @returns an observable that emits the latest report built for `userId`, or
-   * null until a scan completes for that user
+   * @returns an observable that emits the last report successfully built for
+   * `userId`, or null when there is none. Retained while a rescan is in flight
+   * and after a failed one, so it never blinks to null mid-generation; read
+   * `getVaultHealthReportState$` to tell those apart.
    */
   abstract getVaultHealthReport$(userId: UserId): Observable<VaultHealthReportView | null>;
```
lastbestdev: I think we only need one of these methods. Either is fine, no strong preference

bitwarden_license/bit-browser/src/popup/dirt/health/health.component.ts
```diff
@@ -51,6 +86,43 @@ export class HealthComponent {
...
+          switchMap(() => this.generateReport$(userId)),
+          // Clears the previous account's result the moment the active account
+          // changes, so switching users can never briefly show A's at-risk
+          // counts to B while B's own state is still resolving.
+          startWith(VAULT_HEALTH_REPORT_IDLE),
+        ),
+      ),
+    ),
+    { initialValue: VAULT_HEALTH_REPORT_IDLE },
+  );
```
lastbestdev: If we simplified this, reading the results of the scan could be as simple as taking `toSignal` from the VaultHealthReportService `getVaultHealthReport$(...)`

Then we would just need to make sure that elsewhere this component generates the report on init. It could be done in an `effect` something like this:

```
// within constructor (assumes userId, ciphers, and report are signals in component)
effect(() => {
  const userId = this.userId();
  const ciphers = this.ciphers();
  const report = this.report();

  if (report.status === VaultHealthReportStatus.Error || report.status === VaultHealthReportStatus.Idle) {
    void this.vaultHealthReportService.buildVaultHealthReport(userId, ciphers);
  }

})

File: bitwarden_license/bit-common/src/dirt/vault-health/models/vault-health-report-state.ts (hunk 16 above)
Comment (AlexRubik) [reply bitwarden_license/bit-common/src/dirt/vault-health/models/vault-health-report-state.ts]: https://github.com/bitwarden/clients/pull/22346/commits/5fedc5911e01ef947ef16810ac6a0a8a420e263e

File: bitwarden_license/bit-browser/src/popup/dirt/health/health.component.ts (hunk 17 above)
Comment (AlexRubik) [reply bitwarden_license/bit-browser/src/popup/dirt/health/health.component.ts]: https://github.com/bitwarden/clients/pull/22346/commits/5fedc5911e01ef947ef16810ac6a0a8a420e263e

File: bitwarden_license/bit-browser/src/popup/dirt/health/health.component.ts (hunk 14 above)
Comment (AlexRubik) [reply bitwarden_license/bit-browser/src/popup/dirt/health/health.component.ts]: https://github.com/bitwarden/clients/pull/22346/commits/3c3a0c2bf76c03798f6f3fb2f9aff23e082482d9

File: bitwarden_license/bit-common/src/dirt/vault-health/services/implementations/default-vault-health-report.service.ts (hunk 15 above)
Comment (AlexRubik) [reply bitwarden_license/bit-common/src/dirt/vault-health/services/implementations/default-vault-health-report.service.ts]: Done. The published value is now one flat { status, report } with a VaultHealthReportStatus constant, so there's a single report and no duplicate to choose between.

https://github.com/bitwarden/clients/pull/22346/commits/5fedc5911e01ef947ef16810ac6a0a8a420e263e

File: bitwarden_license/bit-common/src/dirt/vault-health/services/abstractions/vault-health-report.service.ts (hunk 22 above)
Comment (AlexRubik) [reply bitwarden_license/bit-common/src/dirt/vault-health/services/abstractions/vault-health-report.service.ts]: Done. Collapsed to one getVaultHealthReport$ that returns the full state; report-only consumers just map to state.report.

https://github.com/bitwarden/clients/pull/22346/commits/5fedc5911e01ef947ef16810ac6a0a8a420e263e

File: bitwarden_license/bit-browser/src/popup/dirt/health/health.component.ts (hunk 23 above)
Comment (AlexRubik) [reply bitwarden_license/bit-browser/src/popup/dirt/health/health.component.ts]: Took the reading half exactly as suggested: scanState is now toSignal over getVaultHealthReport$. For the trigger I went with a small RxJS chain rather than an effect, because an effect reading the state would retry forever on failure, skip the first-visit CTA gate, and rescan on every vault edit (each has a test). The chain is actually the simpler of the two: it expresses those three guards declaratively in one place (a filter plus a take(1)), whereas the effect would need them bolted back on by hand (a one-shot latch so an error doesn't loop, and reading ciphers via firstValueFrom instead of as a signal), which is more code.

https://github.com/bitwarden/clients/pull/22346/commits/3c3a0c2bf76c03798f6f3fb2f9aff23e082482d9

File: bitwarden_license/bit-browser/src/popup/dirt/health/health-scan-error.component.html (hunk 18 above)
Comment (AlexRubik) [reply bitwarden_license/bit-browser/src/popup/dirt/health/health-scan-error.component.html]: The heading already renders at the h5 size because the bitTypography directive gives both h5 and body1 the same font-size class (!tw-text-base), so body1 already produces the h5 size. Moving to h5 would also pull in tw-text-main and fight the explicit heading colour. The real gap was the weight: the spec is 500 (Medium), so I changed tw-font-semibold to tw-font-medium and kept body1.

https://github.com/bitwarden/clients/pull/22346/commits/7b34a7b08edb56aea0f36abed79c690319ce9a27

File: bitwarden_license/bit-browser/src/popup/dirt/health/health-scanning.component.html (hunk 19 above)
Comment (AlexRubik) [reply bitwarden_license/bit-browser/src/popup/dirt/health/health-scanning.component.html]: Same as the failure heading: kept body1 and switched the weight to tw-font-medium to match the spec's 500. This one is also projected into the progress bar's label slot, which the component already styles, so an h5 level there would be redundant.

https://github.com/bitwarden/clients/pull/22346/commits/7b34a7b08edb56aea0f36abed79c690319ce9a27

File: bitwarden_license/bit-browser/src/popup/dirt/health/health-scanning.component.html (hunk 20 above)
Comment (AlexRubik) [reply bitwarden_license/bit-browser/src/popup/dirt/health/health-scanning.component.html:15]: Width: I capped the bar at tw-max-w-64 (256px), which is within 1px of the Figma width (255px). Tailwind bans arbitrary values, so 256px is the right token to land on rather than an exact 255px. Radius already matches the spec (tw-rounded = 4px). The one thing I couldn't match is the bar's height (its thickness): the Figma spec is 16px, but bit-progress-bar renders a fixed tw-h-2 (8px) and exposes no size input, so we can write a follow up for a size input rather than overriding its internals. We can call that tech debt/out of scope.

https://github.com/bitwarden/clients/pull/22346/commits/c454de3ccd0ec87dfc97958cbdb0b3e730fc913e

File: bitwarden_license/bit-common/src/dirt/vault-health/services/implementations/default-vault-health-report.service.ts (hunk 21 above)
Comment (AlexRubik) [reply bitwarden_license/bit-common/src/dirt/vault-health/services/implementations/default-vault-health-report.service.ts]: Done. Each user gets their own subject, so a late build for one account can't overwrite another, and the counter and its guard are gone.

https://github.com/bitwarden/clients/pull/22346/commits/522c084704ad5ceec441afac2f8900e7c217f006

File: bitwarden_license/bit-browser/src/popup/dirt/health/health.component.ts [hunk 24]
```diff
@@ -51,7 +74,63 @@ export class HealthComponent {
...
+        userId
+          ? this.vaultHealthReportService.getVaultHealthReport$(userId)
+          : of(VAULT_HEALTH_REPORT_IDLE),
+      ),
+    ),
+    { initialValue: VAULT_HEALTH_REPORT_IDLE },
+  );
+
+  /** Set when fetching the ciphers to scan fails, which never reaches the service. */
+  private readonly pipelineFailed = signal(false);
```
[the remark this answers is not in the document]
AlexRubik: https://github.com/bitwarden/clients/pull/22346/commits/4e457f9c436e3837c42d4f3d9c7c03fde12fe877

bitwarden_license/bit-browser/src/popup/dirt/health/health.component.ts
```diff
@@ -74,4 +160,24 @@ export class HealthComponent {
...
+      filterOutNullish(),
+      // Generation does an external breach lookup; a vault edit must not re-run it.
+      take(1),
+      switchMap((ciphers) =>
+        defer(() => this.vaultHealthReportService.buildVaultHealthReport(ciphers, userId)),
+      ),
+      catchError((error: unknown) => {
+        // A cipherViews$ failure never reaches the service, so surface it here for
+        // this user and log it.
+        this.pipelineFailedFor.set(userId);
```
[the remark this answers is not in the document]
AlexRubik: https://github.com/bitwarden/clients/pull/22346/commits/d4b5750124da59e1ac62e1ecf29572b6bcb55ccb

bitwarden_license/bit-browser/src/popup/dirt/health/health.component.ts
```diff
@@ -51,7 +74,73 @@ export class HealthComponent {
...
+            // component, so only build when the service holds nothing or a
+            // failure for this user. Without it every back press rescans.
+            switchMap(() =>
+              this.vaultHealthReportService.getVaultHealthReport$(userId).pipe(
+                take(1),
+                filter(
+                  (state) =>
+                    state.status === VaultHealthReportStatus.Idle ||
+                    state.status === VaultHealthReportStatus.Error,
+                ),
```
lastbestdev: This check prevents the scan from rerunning when navigating away and back to the Health tab. My understanding is we want the scan to run automatically when the tab is opened.

bitwarden_license/bit-browser/src/popup/dirt/health/health.component.ts
```diff
@@ -51,7 +74,73 @@ export class HealthComponent {
...
+  );
+
+  /**
+   * The users whose ciphers fetch failed. A set rather than a single slot so
+   * recording or clearing one account's failure cannot disturb another's, which
+   * is the same per-user isolation the report service enforces with its per-user
+   * streams. This failure never reaches the service, so it is tracked here
+   * rather than in that state.
+   */
+  private readonly pipelineFailedFor = signal<ReadonlySet<UserId>>(new Set());
```
lastbestdev: Is this set truly needed? Wondering if we can simplify to an error state signal that is a plain boolean. I'm not seeing why we need to track UserIds

lastbestdev (commented):
A few smaller things remaining, very close

bitwarden_license/bit-browser/src/popup/dirt/health/health.component.ts
lastbestdev: Looks much better, thank you!

bitwarden_license/bit-browser/src/popup/dirt/health/health.component.ts
AlexRubik: It covers an edge case. Tracking the userId matters because of the popped-out Health window option. That is the expand icon in the extension header, which detaches the current page into its own separate browser window.. That window keeps running with its own router while the active account can change underneath it, switched from the toolbar popup or by the background. With a plain boolean, a scan failure recorded for account A would then show on account B in that same window; keying it by user prevents that. Through the normal account switcher this cannot happen, since switching navigates away and rebuilds the tab. I used a Set so clearing one account's failure cannot disturb another's, but I am happy to drop to a single userId signal if you prefer. A plain boolean would bring the cross-account leak back.

https://github.com/bitwarden/clients/pull/22346/commits/89b7b25141d21ebd8f2055ebc7796e2a72107ff4

lastbestdev: Couldn't the scan failure be cleared out when User B visits the Health tab, and the the ciphers are fetched successfully? (in `startGeneration$()`)

lastbestdev: Understanding now - to resolve I believe we can use a combination of a boolean signal (set to false when cipherViews fetch fails, set to true at beginning of `startGeneration$()`) as well as reading the `status` from the state published from the service (`getVaultHealthReport$()`). 

Reading the `status` from the service should prevent the per-user issues, since the published reports are scoped per user.

bitwarden_license/bit-browser/src/popup/dirt/health/health.component.ts
AlexRubik: https://github.com/bitwarden/clients/pull/22346/commits/379ae32c1dbafbebb34162a4ba63118633c43ef3

bitwarden_license/bit-browser/src/popup/dirt/health/health.component.ts
AlexRubik: Done, that is exactly what I implemented: a boolean for the ciphers-fetch failure plus reading status from the service for the loading screen. https://github.com/bitwarden/clients/pull/22346/commits/f0427f55eb5c8c42d93319d5d0ed039e5d11d084

bitwarden_license/bit-browser/src/popup/dirt/health/health.component.spec.ts
```diff
@@ -143,17 +250,342 @@ describe("HealthComponent", () => {
...
+      await settle();
+
+      expect(scanError()).toBeNull();
+      expect(overview()).not.toBeNull();
+    });
+
+    it("shows the progress view on load while the rescan runs, not the report the service still holds", async () => {
+      // PM-39223 rescans on every load. The component asks the service to show
+      // progress immediately (markScanning), so while the rescan runs the tab
+      // shows the progress view rather than the report the service still holds
```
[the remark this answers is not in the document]
lastbestdev: Nit: should remove this

lastbestdev (approved):
LGTM, great job on this

PR #22352, opened 2026-08-12
[PM-41873] Filter inline autofill menu items as the user types

SakshamG7:
## Tracking

Community feature request: https://community.bitwarden.com/t/inline-autofill-menu-ability-to-filter-items-by-typing-searching/63417

## Objective

I have several logins saved for the same site and got tired of scrolling the inline menu to find the right one. I built this for myself, and since I use Bitwarden for free, I figured I might as well offer back the thing I actually use.

When the inline menu is open and you type in a login username field, the list filters instead of closing. The field you are already typing in acts as the search box, so there is no new UI.

Matching is case insensitive against the cipher name, the login username, and the passkey username. The passkey username is in there because that is the label the list actually renders for passkey entries, so matching on anything else makes those rows vanish while you type.

It is behind a feature flag, `inline-menu-typed-search`. With the flag off the old close on type behavior is preserved exactly.

### Why only login fields

Password, TOTP, card, and identity fields keep close on type.

I scoped this to the login case on purpose. It is the common one, it is the one I hit every day, and it is the one I was confident I could do without having to reason hard about security. Cards are the opposite of that. Filtering there means sending card number and CVV keystrokes across the message bus and holding them in background state until the menu closes, and I did not want to be the one making that call.

Identity fields carry none of that concern and would be reasonable to add. They are just low frequency for me.

If someone with a stronger security background sees a safe way to cover cards, I would rather see that land than have me guess at it.

### Repositioning

`updateInlineMenuPositionAfterRepositionEvent` skips repositioning the list when the focused field has a value. That made sense when typing closed the menu, since there was nothing open left to move. Now that the menu stays open while the user types, that early return can leave an open list unpositioned, so it is skipped while a filter is active.

### Testing

340 tests pass across the two autofill specs. `npm run lint` is clean.

Verified by hand in Chrome against a real vault: filtering as you type, matching on both item name and username, passkey username matching, whitespace only input, the empty state, and scrolling within a filtered list.

### Related PRs

#20543, #20836, and #21988 implement the same feature and are all open. I did not find them until after I had built this, so I am happy to close this in favor of whichever the team prefers. The repositioning issue above is not addressed in any of the three and applies regardless of which one lands.

## Screenshots

On focus, every saved login for the site, which is the problem this is trying to solve.

image

Typing `al` narrows it. Alice matches on the item name, Demo and Demo1 on the username. Bob drops out.

image

The filtered list still scrolls when there are more matches than fit.

image

Matching works on the username alone. `car` finds an item named Example because its username is carol@example.com.

image

Another is searching parts of a username.

image

No match falls back to the existing empty state rather than closing the menu, so backspacing brings the list straight back.

image


abinashvaradhan:
any idea when this feature would be get rolled out.. we desperately need this and its been in the request for so long.

SakshamG7:
> any idea when this feature would be get rolled out.. we desperately need this and its been in the request for so long.

I have no idea when this will be reviewed or even if this feature will be added

jprusik:
I reassigned this one to @audreyality since we have a few PRs with different approaches to this and she's presently coordinating the community PRs and our internal Product team developments around this ask.

audreyality:
Hi @SakshamG7,

PR #22354, opened 2026-08-12
refactor(auth): unlock via UnlockService in login strategies

quexten:
Part of the `setUserKey` deprecation. 

Moves to the unlockService equivalent of `setUserKey`. Ideally all login strategies would use the appropriate unlock method, rather than the decryptedKeyMethod. However, the goal here is to remove `setUserKey`.


JaredSnider-Bitwarden (approved):
LGTM! Love how much better and more clear the new service + method names are.

PR #22356, opened 2026-08-12
refactor(auth): unlock via UnlockService when setting an initial password

quexten:
Part of the `setUserKey` deprecation. https://bitwarden.atlassian.net/browse/PM-41489

Unlocking initializes the SDK from persisted state, so each unlock has to run after the account cryptographic state and KDF config writes it depends on, and before `handleResetPasswordAutoEnrollOld`, which reads the user key back out of state. Ordering is documented at each site.


libs/angular/src/auth/password-management/set-initial-password/default-set-initial-password.service.implementation.ts
```diff
@@ -474,7 +484,6 @@ export class DefaultSetInitialPasswordService implements SetInitialPasswordServi
       masterKeyEncryptedUserKey[1],
       userId,
     );
-    await this.keyService.setUserKey(masterKeyEncryptedUserKey[0], userId);
```
quexten, line 477: Already set by the caller as part of line 353

ike-kottlowski (commented):
Auth changes are good. 

There are conflicts, I will approve when those are fixed.

quexten:
@ike-kottlowski Conflicts are resolved!

coroiu (approved):
Approving core module changes

ike-kottlowski (approved):

PR #22357, opened 2026-08-12
refactor(auth): unlock via UnlockService during TDE JIT registration

quexten:
Part of the `setUserKey` deprecation. Independent of the rest of the chain.

Covers both the SDK registration branch and the legacy `initAccount` branch. In the legacy branch the account cryptographic state write moves ahead of the unlock, since unlocking initializes the SDK from state and the freshly generated key pair has to be persisted first.


libs/auth/src/angular/login-decryption-options/login-decryption-options.component.ts
rr-bw, line 1: From a local Claude review:

⚠️ **IMPORTANT**: Unlocking through `UnlockService` fires shared-unlock side effects before `persistUnlockSharingChoice()` records the user's "don't trust this device" choice.

<details>

`keyService.setUserKey` only wrote state. `unlockService.unlockWithDecryptedUserKey` additionally runs `DefaultUnlockService.runOnUnlockSideEffects`, which invokes every callback registered via `registerOnUnlockAction`. On desktop (`DefaultSharedUnlockLeaderService`, provided in `apps/desktop/src/app/services/services.module.ts:407` from the same injector as this component) and web (`DefaultSharedUnlockFollowerService`, `apps/web/src/app/core/core.module.ts:459`), that callback is:

```ts
private async onUnlock(userId, userKey) {
  if (!(await this.enabled(userId))) {
    return;
  }
  await this.leader!.handle_device_event({
    ManualUnlock: { user_id: asUuid(userId), user_key: userKey.toSdk() },
  });
}
```

`enabled()` reads `unlockSharingDisabled$`, whose default for a brand-new JIT-registered user is `false` (`default-shared-unlock-settings.service.ts:42`), and returns `true` for any non-browser client. So for a user who unchecked "remember this device":

1. `unlockWithDecryptedUserKey` → `ManualUnlock` broadcast, user key handed to peer clients (`JsSharedUnlockDriver.unlock_user`).
2. `persistUnlockSharingChoice()` (line 424) then sets `unlockSharingDisabled = true` — after the key already went out.

This contradicts the invariant stated in `persistUnlockSharingChoice`'s own doc comment ("a TDE device the user declined to trust must not participate in shared unlock"). Previously this could only happen via the 100 ms `pollForUnlockEvents` fallback (a narrow, probabilistic race); routing through `UnlockService` makes it deterministic. Both the SDK path and the legacy `initAccount` path are affected, and it only bites when `SharedUnlockPart1`/`SharedUnlockPart2` is enabled.

Suggested fix: record the trust choice before unlocking, e.g. move `await this.persistUnlockSharingChoice();` above the branch that unlocks (it only writes settings, so persisting it on a later failure is harmless and fails safe), or set `unlockSharingDisabled` before the unlock call and keep the rest where it is.

libs/auth/src/angular/login-decryption-options/login-decryption-options.component.spec.ts
rr-bw, line 1: Suggestion from a local Claude review:

🎨 **SUGGESTED**: Tests assert both calls happened but not that the cryptographic state write precedes the unlock — the constraint this change is built around.

<details>

Both code comments added here state that `setAccountCryptographicState` must run before `unlockWithDecryptedUserKey`, because `initialize_user_crypto` reads `accountCryptographicState$` and `assertNonNullish`-throws when it is missing (`libs/unlock/src/default-unlock.service.ts:185-193`). The tests would still pass if the two statements were swapped back, so a future refactor can silently reintroduce a total failure of TDE JIT registration.

One assertion per path pins it down:

```ts
expect(
  accountCryptographicStateService.setAccountCryptographicState.mock.invocationCallOrder[0],
).toBeLessThan(unlockService.unlockWithDecryptedUserKey.mock.invocationCallOrder[0]);
```

libs/auth/src/angular/login-decryption-options/login-decryption-options.component.ts
quexten: Thanks Ryan! I think this is a valid concern, but I would appreciate insight from the auth perspective on what you expect me to do to resolve, otherwise this is not helpful.

Are you saying "Please move up the persist unlock sharing choice, so that the unlock is not shared via shared unlock in the event that the user does not trust the device"?

(I think that makes sense but I would like to confirm that this is your suggestion)

quexten: Fix applied.

libs/auth/src/angular/login-decryption-options/login-decryption-options.component.spec.ts
quexten: I'm really not a fan of these tests since they are very brittle. I'll track the follow-up work to move state writes to the SDK. Then, we can make much nicer tests that just:

1. Perform the action
2. Spawn a new sdk instance
3. Try to unlock

all without direct state assertions.

quexten:
Rust lint errors are unrelated to this PR.

libs/auth/src/angular/login-decryption-options/login-decryption-options.component.ts
rr-bw: That's great, thanks.

libs/auth/src/angular/login-decryption-options/login-decryption-options.component.spec.ts
rr-bw: What do you think about a test that asserts that `persistUnlockSharingChoice()` runs before any shared-unlock side-effects? I think we should make sure we guard against re-introducing a silent bug where we allow shared-unlock when the user did not trust the device.

rr-bw (approved):
Approved with one non-blocking comment: https://github.com/bitwarden/clients/pull/22357#discussion_r3835020828

libs/auth/src/angular/login-decryption-options/login-decryption-options.component.spec.ts
quexten: Makes sense. Follow-up with tests here: https://github.com/bitwarden/clients/pull/22600

PR #22358, opened 2026-08-12
refactor(key-management): unlock via UnlockService in key connector conversion

quexten:
Part of the `setUserKey` deprecation.  https://bitwarden.atlassian.net/browse/PM-41489

Two behaviour changes fall out of this, both of them fixes:

- The V1 conversion generated a key pair, sent it to the server and never persisted it locally, so the SDK had nothing to initialize from. It now writes the account cryptographic state before unlocking.
- Both conversions unlock after the server call rather than before it, so a failed conversion leaves the user locked instead of unlocked with no local key pair.


mzieniukbw (dismissed):
Nice! You got some merge conflicts though

quexten:
Looks like cargo deny is unrelated.

mzieniukbw (approved):

PR #22364, opened 2026-08-12
[PM-41874] refactor(crypto): move tools callers to @bitwarden/legacy-crypto

quexten:
move callers for tools from the barrel-re-export to the legacy-crypto module directly. The re-exports will be removed in a follow-up PR.

### DIFF
--- a/libs/importer/src/importers/lastpass/access/services/crypto-utils.ts
+++ b/libs/importer/src/importers/lastpass/access/services/crypto-utils.ts
@@ -1,9 +1,10 @@
-import {
-  DANGEROUS_aesEcbDecryptLastpassImport,
-  DANGEROUS_aesCbcDecryptLastpassImport,
-} from "@bitwarden/common/key-management/crypto";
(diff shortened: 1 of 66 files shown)


libs/importer/src/importers/lastpass/access/services/crypto-utils.ts
```diff
@@ -1,9 +1,10 @@
+import { Utils } from "@bitwarden/common/platform/misc/utils";
+// eslint-disable-next-line no-restricted-imports
 import {
-  DANGEROUS_aesEcbDecryptLastpassImport,
+  CryptoFunctionService,
   DANGEROUS_aesCbcDecryptLastpassImport,
-} from "@bitwarden/common/key-management/crypto";
-import { CryptoFunctionService } from "@bitwarden/common/key-management/crypto/abstractions/crypto-function.service";
-import { Utils } from "@bitwarden/common/platform/misc/utils";
+  DANGEROUS_aesEcbDecryptLastpassImport,
```
harr1424, line 6: ⚠️ I need to look at this more but we probably shouldn't be using this if there is a better alternative

harr1424 (approved):

PR #22369, opened 2026-08-12
[PM-41874] refactor(crypto): move platform callers to @bitwarden/legacy-crypto

quexten:
move callers for platform from the barrel-re-export to the legacy-crypto module directly. The re-exports will be removed in a follow-up PR.


libs/vault/src/cipher-form/components/sshkey-section/sshkey-section.component.spec.ts
```diff
@@ -18,6 +18,7 @@ import { SshKeySectionComponent } from "./sshkey-section.component";
 
 jest.mock("@bitwarden/sdk-internal", () => {
   return {
+    ...jest.requireActual("@bitwarden/sdk-internal"),
```
quexten, line 21: Had to add this, otherwise tests break

dani-garcia (approved):

gbubemismith (approved):

PR #22371, opened 2026-08-12
[PM-41874] refactor(crypto): move billing callers to @bitwarden/legacy-crypto

quexten:
move callers for billing from the barrel-re-export to the legacy-crypto module directly. The re-exports will be removed in a follow-up PR.

### DIFF
--- a/apps/web/src/app/billing/individual/upgrade/premium-org-upgrade-payment/services/premium-org-upgrade.service.ts
+++ b/apps/web/src/app/billing/individual/upgrade/premium-org-upgrade-payment/services/premium-org-upgrade.service.ts
@@ -12,15 +12,17 @@ import {
   SubscriptionCadenceIds,
 } from "@bitwarden/common/billing/types/subscription-pricing-tier";
 import { FeatureFlag } from "@bitwarden/common/enums/feature-flag.enum";
-import { EncryptService } from "@bitwarden/common/key-management/crypto/abstractions/encrypt.service";
-import { EncString } from "@bitwarden/common/key-management/crypto/models/enc-string";
 import { ConfigService } from "@bitwarden/common/platform/abstractions/config/config.service";
 import { I18nService } from "@bitwarden/common/platform/abstractions/i18n.service";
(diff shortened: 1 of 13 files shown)


apps/web/src/app/billing/individual/upgrade/premium-org-upgrade-payment/services/premium-org-upgrade.service.ts
```diff
@@ -12,15 +12,17 @@ import {
...
-import { EncString } from "@bitwarden/common/key-management/crypto/models/enc-string";
 import { ConfigService } from "@bitwarden/common/platform/abstractions/config/config.service";
 import { I18nService } from "@bitwarden/common/platform/abstractions/i18n.service";
-import { SymmetricCryptoKey } from "@bitwarden/common/platform/models/domain/symmetric-crypto-key";
 import { OrgKey } from "@bitwarden/common/types/key";
 import { SyncService } from "@bitwarden/common/vault/abstractions/sync/sync.service.abstraction";
 // eslint-disable-next-line no-restricted-imports
-import { LegacyCompatKeyService } from "@bitwarden/legacy-crypto";
+import {
+  EncryptService,
```
cyprain-okeke, line 21: ❓ Not a billing caller so no action needed here, but the OrgKey we're importing from @bitwarden/common/types/key still resolves SymmetricCryptoKey through the platform/models/domain shim. Is that file covered in the follow-up that removes the re-exports? Just want to make sure we have our bases covered before those paths disappear.

cyprain-okeke (commented):
Looks good - just one question.

apps/web/src/app/billing/individual/upgrade/premium-org-upgrade-payment/services/premium-org-upgrade.service.ts
quexten: We've not moved the named keys (https://github.com/bitwarden/clients/blob/8711032269f61e5bd8b3ac363ef5666057d0f6c6/libs/common/src/types/key.ts#L12), but you make a good point. We'll not handle that in this PR but will make a follow-up PR for those.

cyprain-okeke (approved):

PR #22376, opened 2026-08-12
[PM-41854] fix: skip collections coachmark step for users without collections

gbubemismith:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-41854

## 📔 Objective

The new-user coachmark tour silently ended after step 2 for an organization member with no collections.

Step 3 (`shareWithCollections`) anchors to the collections section header, which only renders when the user has at least one collection. `startTour()` filtered steps only on `requiresOrganization`, so the step stayed in the list while its anchor never rendered — the tour sat on an invisible step 3 and step 4 was unreachable.

This adds a `requiresCollections` flag to `CoachmarkStep`, mirroring `requiresOrganization`. Filtering up front keeps the "Step X of Y" indicator accurate.

## 📸 Screenshots

https://github.com/user-attachments/assets/e6486504-4b7e-464d-8f45-32a004fffa0f

### DIFF
--- a/apps/web/src/app/vault/components/coachmark/coachmark.service.spec.ts
+++ b/apps/web/src/app/vault/components/coachmark/coachmark.service.spec.ts
@@ -2,7 +2,9 @@ import { TestBed, fakeAsync, tick } from "@angular/core/testing";
 import { Router } from "@angular/router";
 import { BehaviorSubject, of } from "rxjs";
 
+import { CollectionService } from "@bitwarden/admin-console/common";
 import { OrganizationService } from "@bitwarden/common/admin-console/abstractions/organization/organization.service.abstraction";
+import { CollectionView } from "@bitwarden/common/admin-console/models/collections";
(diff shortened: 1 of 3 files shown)


apps/web/src/app/vault/components/coachmark/coachmark.service.spec.ts
```diff
@@ -192,13 +197,25 @@ describe("CoachmarkService", () => {
...
       hasOrganizations.mockReturnValue(of(true));
+      decryptedCollections$.mockReturnValue(of([{} as CollectionView]));
 
       void service.startTour();
       tick(200);
 
       expect(service.totalSteps()).toBe(4);
     }));
 
+    it("should exclude collection-only steps for org users without collections", fakeAsync(() => {
```
nick-livefront, line 208: 👏

nick-livefront (approved):

PR #22383, opened 2026-08-12
[PM-40330] Add vault list table to desktop behind VFO1Foundation flag

nick-livefront:
## 🎟️ Tracking

[PM-40330](https://bitwarden.atlassian.net/browse/PM-40330)

## 📔 Objective

- `VaultListTableComponent` connects `VaultItemsTableComponent` to the desktop vault, gated behind the `VFO1Foundation` feature flag.
- `CipherRowMenuService`, a service that centralizes the overflow menu visibility logic into s single place. 
- Adds `CipherActionService` to handle entirely within shared libs, with a `cipherActionSuccess$` stream that triggers a vault refresh for desktop.
  - Removes event bubbling for the above actions from VaultComponent; the service handles them directly instead of the component catching and dispatching each case
- I added virtual scrolling to the underlying vault table items by defining the cipher row height. 
- Bug fix: secure note copy i18n key was `"note"` instead of `"typeSecureNote"` in `CopyCipherFieldService`
- Bug fix missing `await` on `promptForPremium()` before opening attachments dialog

## 📸 Screenshots

Screenshot 2026-08-12 at 2 24 17 PM

[PM-40330]: https://bitwarden.atlassian.net/browse/PM-40330?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/desktop/src/vault/app/vault-v3/vault.component.html
```diff
@@ -2,47 +2,70 @@
...
+    />
+    <app-vault-list
+      class="tw-h-full"
+      [ciphers]="ciphers"
+      [collections]="collectionsToDisplay"
+      [allCollections]="allCollections"
+      [allOrganizations]="allOrganizations"
+      [disabled]="refreshing"
+      [showOwner]="true"
+      [showPremiumFeatures]="userHasPremium()"
```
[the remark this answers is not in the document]
nick-livefront: [3ccac08](https://github.com/bitwarden/clients/pull/22383/commits/3ccac0860c821843ae2c55b50ffb6c406ba2e364)

apps/desktop/src/locales/en/messages.json
```diff
@@ -170,6 +170,9 @@
   "sharedFoldersDeleted": {
     "message": "Shared folders deleted"
   },
+  "addToSharedFolder": {
+    "message": "Add to shared folder"
+  },
```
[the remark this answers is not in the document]
nick-livefront: [f5ce60e](https://github.com/bitwarden/clients/pull/22383/commits/f5ce60e4f6917e463b9d0a45f09f179b7f30b517)

libs/vault/src/services/cipher-row-menu.service.ts
```diff
@@ -0,0 +1,198 @@
...
+        variant: "danger",
+      },
+      {
+        id: "permanentlyDelete",
+        label: this.i18nService.t("permanentlyDelete"),
+        icon: "bwi-trash",
+        event: (item) => ({ type: "delete", items: [{ cipher: item }] }),
+        show: (item) => item.edit && CipherViewLikeUtils.isDeleted(item),
+        variant: "danger",
+      },
```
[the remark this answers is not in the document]
nick-livefront: [ae6a3c1](https://github.com/bitwarden/clients/pull/22383/commits/ae6a3c1981590494a5b96f903c13599247be1d2f)

apps/desktop/src/vault/app/vault-v3/vault.component.ts
```diff
@@ -633,56 +652,17 @@ export class VaultComponent<C extends CipherViewLike> implements OnInit, OnDestr
...
-            this.refresh();
-          }
-        }
-        break;
-      case "toggleFavorite":
-        await this.handleFavoriteEvent(event.item);
-        break;
       case "editCipher": {
         const fullCipher = await this.cipherService.getFullCipherView(event.item);
         await this.editCipher(fullCipher);
```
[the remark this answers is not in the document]
shane-melton: @nick-livefront I think we'll need to add these back as suggested

libs/vault/src/components/vault-items-table/vault-items-table-actions-column.component.ts
```diff
@@ -120,10 +119,19 @@ export class VaultItemsTableActionsColumnComponent<C extends CipherViewLike, E>
...
 
   /** The actions visible for `item`, honouring each action's optional `show` predicate. */
-  protected visibleActions(item: C): VaultItemsTableRowAction<C, E>[] {
+  protected visibleActions(item: C): VaultItemsTableRowAction<C>[] {
     return this.rowActions().filter((action) => action.show?.(item) ?? true);
   }
 
+  /** Runs the action, or prompts for a premium upgrade if the action is gated. */
+  protected handleAction(action: VaultItemsTableRowAction<C>, item: C): void {
+    if (action.premiumGated?.(item)) {
```
shane-melton: :question: Is this necessary? The `handleAction` template binding is already gated behind `premiumGated` in the template (the premium button is rendered instead).

shane-melton: This would also let us drop injecting the `PremiumUpgradePromptService` in this component.

libs/vault/src/services/cipher-row-menu.service.ts
```diff
@@ -0,0 +1,211 @@
...
+    { initialValue: false },
+  );
+
+  private readonly restrictedTypes = toSignal(this.restrictedItemTypesService.restricted$, {
+    initialValue: [] as RestrictedCipherType[],
+  });
+
+  /**
+   * Returns the full row action definitions for the cipher overflow menu.
+   * Source of truth is the web vault's `VaultCipherRowComponent`.
```
shane-melton: :pick: Comment will become inaccurate when `VaultCipherRowComponent` is removed and isn't entirely relevant imo.

libs/vault/src/services/cipher-row-menu.service.ts
```diff
@@ -0,0 +1,211 @@
...
+        id: "addFavorite",
+        label: this.i18nService.t("favorite"),
+        icon: "bwi-star",
+        run: (item) => void this.cipherActionService.toggleFavorite(item),
+        show: (item) => !item.favorite && this.showFavorite(item),
+      },
+      {
+        id: "removeFavorite",
+        label: this.i18nService.t("unfavorite"),
+        icon: "bwi-star",
```
shane-melton: :question: Should this icon be different for the unfavorite action? It might be an outline version of star.

apps/desktop/src/vault/app/vault-v3/vault.component.ts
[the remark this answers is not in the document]
nick-livefront: Dangit, nice catch: [b64b010](https://github.com/bitwarden/clients/pull/22383/commits/b64b0102be32689ce13a61f9b8dcd94db0092d4c)

libs/vault/src/services/cipher-action.service.ts
```diff
@@ -0,0 +1,203 @@
...
+    this._cipherActionSuccess.next();
+  }
+
+  private async promptPassword(cipher: CipherViewLike): Promise<boolean> {
+    if (cipher.reprompt === CipherRepromptType.None || this.passwordReprompted) {
+      return true;
+    }
+
+    return (this.passwordReprompted = await this.passwordRepromptService.showPasswordPrompt());
+  }
```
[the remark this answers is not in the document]
nick-livefront: [064344e](https://github.com/bitwarden/clients/pull/22383/commits/064344ecdcf189f9ed398f2277dc9a1db63fb604)

libs/vault/src/services/cipher-row-menu.service.ts
```diff
@@ -0,0 +1,211 @@
...
+    initialValue: [] as RestrictedCipherType[],
+  });
+
+  /**
+   * Returns the full row action definitions for the cipher overflow menu.
+   * Source of truth is the web vault's `VaultCipherRowComponent`.
+   */
+  getRowActions<C extends CipherViewLike>(
+    collections: CollectionView[] = [],
+    handler: (event: VaultItemEvent<C>) => void,
```
shane-melton: :art: This seems fairly brittle. A caller needs to know which events it needs to override and isn't actually forced to provide handlers for those that need it. 

I'd also like to move away from the `VaultItemEvent` interface entirely if we're opting for the shared `CipherActionService`.  Two options I can see:

- Make `handler` strongly typed. Something like:
```
type CipherRowMenuHandlers<C extends CipherViewLike> = {
   view: (item: C) => void | Promise<void>;
   edit: (item: C) => void | Promise<void>;
   clone: (item: C) => void | Promise<void>;
   assignToCollections: (item: C) => void | Promise<void>;
   };
```
- Update desktop to have the same behavior as Web for these actions. (This is ideal long term, but may be more difficult short term).

shane-melton (changes requested):
I like the direction this is going! Just a few more comments/suggestions.

libs/vault/src/components/vault-items-table/vault-items-table-actions-column.component.ts
nick-livefront: Done! [0e51e21](https://github.com/bitwarden/clients/pull/22383/commits/0e51e219e53ad4155ba0a2eed9080b2e80b28a55)

(to answer your question, no not needed)

libs/vault/src/services/cipher-row-menu.service.ts
nick-livefront: [25eb576](https://github.com/bitwarden/clients/pull/22383/commits/25eb576c32f9b08ce821295bb78f3e2315495a5b)

libs/vault/src/services/cipher-row-menu.service.ts
nick-livefront: In prod it actually is the same `bwi-star` regardless of if the cipher is favorited or not. The description for the filled star in storybook is `unfavorite action` so that is pretty clear cut to me. I updated it : [e20ff80](https://github.com/bitwarden/clients/pull/22383/commits/e20ff80295969d2a03da9ff0b1772fb0ad6aeb5e)

libs/vault/src/services/cipher-row-menu.service.ts
nick-livefront: True true true, I opted to make it strongly typed in the interest of time and causing any unintended changes in either the web or desktop: [de71a5c](https://github.com/bitwarden/clients/pull/22383/commits/de71a5cecb2cc6a08bd4c16487c906ea473793cd)

libs/vault/src/components/vault-items-table/vault-items-table.component.html
```diff
@@ -6,6 +6,8 @@
   [filters]="initialFilterValues()"
   [loading]="loading()"
   [selection]="selection"
+  [virtualRowHeight]="56"
+  height="fill"
```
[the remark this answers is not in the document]
nick-livefront: [d33968e](https://github.com/bitwarden/clients/pull/22383/commits/d33968e947012d199723dd07e6de6b16b9c14cbc)

libs/vault/src/components/vault-items-table/vault-items-table-actions-column.component.ts
```diff
@@ -120,10 +119,19 @@ export class VaultItemsTableActionsColumnComponent<C extends CipherViewLike, E>
...
+      void this.premiumUpgradePromptService.promptForPremium();
+      return;
+    }
+    void action.run(item);
+  }
```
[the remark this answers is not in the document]
nick-livefront: Handled above by shane's comment

libs/vault/src/services/cipher-action.service.ts
```diff
@@ -0,0 +1,199 @@
...
+  private readonly i18nService = inject(I18nService);
+  private readonly logService = inject(LogService);
+  private readonly passwordRepromptService = inject(PasswordRepromptService);
+  private readonly premiumUpgradePromptService = inject(PremiumUpgradePromptService);
+  private readonly toastService = inject(ToastService);
+
+  private readonly _cipherActionSuccess = new Subject<void>();
+
+  /** Emits after any action completes (success or no-op). Subscribe to trigger a vault refresh. */
+  readonly cipherActionSuccess$ = this._cipherActionSuccess.asObservable();
```
shane-melton: :art: (non-blocking):`cipherActionSuccess$` doesn't seem entirely accurate. Something like `cipherModified$` would better communicate when this emits (and could also possibly emit the cipherId). 

For example, the `viewAttachments` action is technically successful even though it never emits if the user doesn't modify the attachments.

apps/desktop/src/vault/app/vault-v3/vault.component.ts
```diff
@@ -617,6 +632,10 @@ export class VaultComponent<C extends CipherViewLike> implements OnInit, OnDestr
       .pipe(takeUntil(this.destroy$))
       .subscribe(() => this.refresh());
 
+    this.cipherActionService.cipherActionSuccess$
```
shane-melton: :question: (non-blocking) Should this bubble up via the table verus having to explicitly subscribe via the service? The vault container here otherwise (once we drop the events after vfo1) shouldn't really need to "go around" the table to get to the table action emitters.

shane-melton (dismissed):
Only two non-blocking comments, otherwise I think we're good!

libs/vault/src/services/cipher-action.service.ts
nick-livefront: [00c7508](https://github.com/bitwarden/clients/pull/22383/commits/00c7508b0eb6f747cc8e0a7b8304482c15176d02)

shane-melton (approved):

PR #22384, opened 2026-08-12
[PM-41898] SSH Agent v2: add per-key destination filtering

V1rg1lee:
## 🎟️ Tracking

Related to #13401

GitHub Discussion: https://github.com/orgs/bitwarden/discussions/22382

Community feature request:
https://community.bitwarden.com/t/ssh-agent-allow-specifying-limiting-private-key-offers/79954

## 📔 Objective

The SSH agent can expose enough identities for an SSH server to hit `MaxAuthTries` before OpenSSH reaches the correct key.

This PR adds optional per-key destination filtering to SSH Agent v2 using the verified host-key fingerprint already provided by OpenSSH through `session-bind@openssh.com`.

SSH key items on Desktop can now be associated with one or more SHA256 host-key fingerprints. These settings are stored locally on Desktop and are not added to synchronized cipher data.

When a verified destination is available, the agent returns identities in this order:

1. keys explicitly configured for the destination;
2. unrestricted keys with no destinations configured.

Keys restricted to other destinations are omitted. Ordering within each group remains stable.

If no verified session binding is available, all identities are returned as before.

This allows a user to configure only the key they expect to use for a server and have it prioritized ahead of unrestricted Bitwarden identities, without requiring every other key to be configured first.

The destination setting is an identity-offering optimization, not an authorization boundary. This PR does not add hostname/IP matching, host-key discovery, `ssh-keyscan`, or changes to signing authorization.

## 🧪 Testing

Automated testing includes:

- Rust unit and integration coverage for destination matching, filtering, prioritization, stable ordering, serialization, and the no-session-bind fallback.
- Desktop/UI tests for local persistence, Save/Cancel behavior, validation, new-cipher persistence, and removal of destination restrictions.
- `cargo fmt --check`
- `cargo check -p desktop_napi`
- `node scripts/lint-rust.mjs --only clippy`
- Relevant Jest and TypeScript checks
- ESLint and Prettier checks

Manual end-to-end testing was performed against a real OpenSSH server on Windows.

Before this change:

- the Bitwarden agent returned 6 identities;
- the correct key was reached too late;
- the server disconnected with `Too many authentication failures`.

After configuring only the correct key for the destination and leaving the other keys unrestricted:

- OpenSSH successfully bound the agent connection to the server host key;
- the destination-matched Bitwarden identity was prioritized ahead of unrestricted Bitwarden identities;
- the server accepted the correct RSA key and public-key authentication succeeded.

I also tested the stricter case where unrelated keys were restricted to other destinations. In that case, the agent returned only the matching identity.

### Configured destination

image

### Validation

image

### DIFF
--- a/apps/desktop/desktop_native/napi/index.d.ts
+++ b/apps/desktop/desktop_native/napi/index.d.ts
@@ -482,6 +482,11 @@ export declare namespace sshagent_v2 {
     privateKey: string
     name: string
     cipherId: string
+    /**
+     * SHA-256 host-key fingerprints this key is restricted to offering. Empty means
+     * unrestricted.
+     */
+    destinationFingerprints: Array<string>
   }
 }
 
--- a/apps/desktop/desktop_native/napi/src/sshagent_v2.rs
+++ b/apps/desktop/desktop_native/napi/src/sshagent_v2.rs
@@ -30,6 +30,9 @@ pub mod sshagent_v2 {
         pub private_key: String,
         pub name: String,
         pub cipher_id: String,
+        /// SHA-256 host-key fingerprints this key is restricted to offering. Empty means
+        /// unrestricted.
+        pub destination_fingerprints: Vec<String>,
     }
 
     /// SSH public key data
@@ -217,6 +220,7 @@ pub mod sshagent_v2 {
                     private_key_pem: k.private_key,
                     name: k.name,
                     cipher_id: k.cipher_id,
+                    destination_fingerprints: k.destination_fingerprints,
                 })
                 .collect();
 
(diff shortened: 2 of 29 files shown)


djsmith85 (commented):
Thank you for your contribution @V1rg1lee! The Desktop-Native product team will evaluate your contribution and discuss with the engineering team

Platform-owned changes are limited to the updated napi-file. @bitwarden/team-desktop-native-dev can re-request a review once they are happy with the proposed solution.

V1rg1lee:
Great! If you have any questions at all, please don't hesitate to ask. 😊

V1rg1lee:
Hi @neuronull , I noticed the recent serial_test dependency update in Desktop Native, and the PR is now quite a few commits behind main.

Before you review it, would you prefer me to sync the branch with the latest main and resolve any conflicts, or should I leave it as-is for now to avoid changing the branch during review?

Happy to do whichever you prefer.

dwbit:
Thanks for the proposed contribution, we appreciate your time on this one! In this case, the team is already working on a solution, so stay tuned for updates!

PR #22387, opened 2026-08-12
[PM-40867] feat: Show email verification status and verify action on My account settings page

enmande:
## 🎟️ Tracking

[PM-40867](https://bitwarden.atlassian.net/browse/PM-40867)

## 📔 Objective

- Adds a verification status indicator and CTA to the Email field on Settings > My account, so users can see whether their email is verified and trigger verification from there. 
  - Reuses the existing re-send and confirm flows.
- Removes the "verify your email" vault banner and its now-unused `verifyEmailDesc` locale key.

### Full Exercise
Full exercise showing verification flow as well as "Verified ✅" badge reactivity; if a stale "Verify Email" CTA is displayed (for example, on an old tab from before verification was performed), subsequent clicks will surface the "already verified" error from the server as an _info_ toast and the state is updated.

https://github.com/user-attachments/assets/2fee5ac3-2370-4b98-bd63-20ea65de9d79

### Verified Badge Appears, Banner is Removed

https://github.com/user-attachments/assets/cc6dfe1b-c0f0-4fa1-8702-fef2e1b2f4a8

### Verified Badge

Screenshot 2026-08-12 at 2 35 47 PM

### Verify Email Button

Screenshot 2026-08-12 at 4 29 43 PM

### Rate Limiting

Verify email is rate-limited to 5 calls/60 minutes (server opinion).

https://github.com/user-attachments/assets/0e803518-1a72-4caf-94e4-479d6633563b

[PM-40867]: https://bitwarden.atlassian.net/browse/PM-40867?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/web/src/app/auth/settings/account/profile.component.ts
```diff
@@ -114,4 +114,12 @@ export class ProfileComponent implements OnInit {
       message: this.i18nService.t("accountUpdated"),
     });
   };
+
+  protected readonly verifyEmail = async () => {
+    await this.apiService.postAccountVerifyEmail();
```
JaredSnider-Bitwarden: 🎨 : We should catch the server error and map to a translated your email is already verified to maintain parity with what we had before.

enmande: Updated in 80deeaf.
- Exposed a signal off of `AccountService` for better reactivity (following `email`'s pattern)
- Updated the `ngOnInit` to patch the `AccountService` with the always-fresh `profile` call's email-verified value to improve reactivity
- Added an "Email already verified" locale key and will expose it to the user as an _info_ toast if the server returns "already verified" from a subsequent call.
  - Opted to not use a _success_ toast as no action is taken here (nothing succeeded/changed, only an updated state), but happy to discuss other approaches.

JaredSnider-Bitwarden (approved):
Excellent work! Thank you for the re-addition of the translated error + the component tests that are useful in this case! :shipit:

PR #22398, opened 2026-08-13
[PM-41905] Add Autotype MVP Window Validity Check

coltonhurst:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-41905

## 📔 Objective

If the main window in the desktop app is destroyed, and a user presses the autotype hotkey combination before the quit processes have completed, the app would most likely crash due to no conditional checking that the window is still valid (not null or destroyed) before we send the autotype request from the main process → render process.

Adding a null check prevents this crash, in case there are specific tasks prior to app shut down that need to happen.

## 🧪 Testing

Verified this works by editing `window.main.ts` with the following (uncommitted) change, keeping the app alive, and pressing the keyboard shortcut:

```ts
app.on("window-all-closed", () => {
  if (!isMac() || this.isQuitting || isMacAppStore()) {
    // app.quit();   // DO NOT COMMIT
  }
});
```

Screenshot of the logs:

image

---

I also tested Autotype still works normally, and that it still works while closed to tray (with the "close to tray" setting on in local desktop settings).

---

I also verified that the crash does happen on main (tested on 1316fd8ab7b7b27a4510a6118a31950214860548):

image


apps/desktop/src/autofill/main/main-desktop-autotype-mvp.service.ts
```diff
@@ -102,11 +102,17 @@ export class MainDesktopAutotypeMvpService {
     const result = globalShortcut.register(
       this.autotypeKeyboardShortcut.getElectronFormat(),
       () => {
-        const windowTitle = autotype_mvp.getForegroundWindowTitle();
-
-        this.windowMain.win.webContents.send(AUTOTYPE_MVP_IPC_CHANNELS.LISTEN, {
-          windowTitle,
-        });
+        if (this.windowMain.win != null) {
```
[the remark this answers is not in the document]
coltonhurst: Addressed in https://github.com/bitwarden/clients/pull/22398/commits/cb586a239b52ad40031eddbdfd8db623f93db046

apps/desktop/src/autofill/main/main-desktop-autotype-mvp.service.spec.ts
```diff
@@ -414,5 +414,21 @@ describe("MainDesktopAutotypeMvpService", () => {
...
+
+      // Get the registered callback
+      const registeredCallback = (globalShortcut.register as jest.Mock).mock.calls[0][1];
+
+      expect(() => registeredCallback()).not.toThrow();
+      expect(autotype_mvp.getForegroundWindowTitle).not.toHaveBeenCalled();
+      expect(mockLogService.debug).toHaveBeenCalledWith(
+        "Autotype keyboard shortcut activated, but the main window does not exist.",
+      );
+    });
```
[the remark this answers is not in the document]
coltonhurst: https://github.com/bitwarden/clients/pull/22398/commits/4d779bb955b4444c50de08301638402d50b3d760 and https://github.com/bitwarden/clients/pull/22398/commits/7dcec4fda0a7243ac8a20df8c2c3afd1487e06e8 address this

apps/desktop/src/autofill/main/main-desktop-autotype-mvp.service.ts
```diff
@@ -102,11 +102,17 @@ export class MainDesktopAutotypeMvpService {
...
       () => {
-        const windowTitle = autotype_mvp.getForegroundWindowTitle();
-
-        this.windowMain.win.webContents.send(AUTOTYPE_MVP_IPC_CHANNELS.LISTEN, {
-          windowTitle,
-        });
+        if (this.windowMain.win != null && !this.windowMain.win.isDestroyed()) {
+          const windowTitle = autotype_mvp.getForegroundWindowTitle();
+
+          this.windowMain.win.webContents.send(AUTOTYPE_MVP_IPC_CHANNELS.LISTEN, {
```
neuronull, line 108: Is there value in making changes to the MVP at this point that could potentially cause problems? Phrased differently- what are the chances that a user could encounter this, and what is the outcome if they encounter it but we don't fix it?

coltonhurst: I think the chances of a crash happening are low. However, if a crash were to occur, any logic that does need to happen during the shutdown process (after the window has been destroyed or made null) may not complete. Here are three examples of logic that execute at some point in the app closing process:
- `window.main.ts:194` -> biometric reset prompts are cancelled which allow the biometric service to auto-prompt on reload
- `clipboard.main.ts:11` -> if the clipboard is holding a value from a Bitwarden copy action, it clears the clipboard so copied passwords don't linger after exit
- `electron-storage.service.ts` -> the storage backend is flushed to ensure everything storage related is written to disk

If a crash does occur, in case unintended side effects happen, it would be hard to point to this Autotype issue as the cause.

This was a quick find & easy fix. We can utilize the same logic for GA, as the same situation could occur. With the fix being pretty simple and there being potential complications for a user in the case of a crash, I personally think it's worth it to implement this fix.

neuronull: Are we sure that the app would crash if it happened? ( since you were able to inject a case where the fix logged the debug, did you also check what happens without your changes?)

> main window in the desktop app is destroyed

I don't know enough about what all triggers and the subsequent chain of events to understand the stated problem conditions for this scenario to occur.

For my own learning, can you expand on why both

```this.windowMain.win != null && !this.windowMain.win.isDestroyed())```

these things need to be checked? For example, is the first condition something that happens if it's minimized? And maybe it would be useful to code comment about that since others might wonder too.

It seems like the data to support this scenario happening is loose, based on what I understand. If you feel confident in this though, we can proceed with it. If you can confirm from the manual injection question that the app does crash, that boosts confidence.

Comment (coltonhurst) [reply apps/desktop/src/autofill/main/main-desktop-autotype-mvp.service.ts:108]: I just did a test verifying the app does crash, tested on main. Fair and legitimate questions; I updated this PR description with a screenshot and details 👍

---

For the `null` and `isDestroyed()` question, here is our code:

```ts
if (this.windowMain.win != null && !this.windowMain.win.isDestroyed()) {
          const windowTitle = autotype_mvp.getForegroundWindowTitle();

          this.windowMain.win.webContents.send(AUTOTYPE_MVP_IPC_CHANNELS.LISTEN, {
```

Regarding `null`:
- (Assuming `this.windowMain.win` is null...) On the last line in the above snippet, without the null check, we are attempting to access the `webContents` property on a `null` value, which will fail
- Our own window reference is set to null when the `closed` event is emitted; this is done in `window.main.ts:481`
    - This follows the correct docs pattern mentioned [here](https://www.electronjs.org/docs/latest/api/browser-window#event-closed)

Regarding `isDestroyed()`:
- This gets into Electron internals, which I am no expert of. But, to my knowledge, a window will eventually be "destroyed" by Electron under the covers (there are other ways too, like calling `destroy()` explicitly). When the window is closed, we call a closure that sets our own `win` reference to null (as mentioned above, `window.main.ts:481`). In case the destroy by electron happens *before* the closure runs, `win` might not be null but be destroyed, so we need to also check `isDestroyed()`.
- This matches the pattern found in the `updateWindowState()` function in `window.main.ts:591`

neuronull: thanks for checking that and the extra details. I'm good to move this forward but I do think we should consolidate efforts to the GA going forward, baring MVP bugs that are found.  (2c 🪙)

neuronull (approved):

PR #22405, opened 2026-08-13
Remove unused loginApprovalModelRef

djsmith85:
## 📔 Objective

With previous work moving to the `LoginApprovalDialogComponent`, it was missed to remove the existing loginApproval-modal.

This removes the modalRef and the template within the markup.


ike-kottlowski (approved):
Thanks!

PR #22407, opened 2026-08-13
[PM-41880] Quick fix for edit members dialog dependencies

ttalty:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-41880

## 📔 Objective

Clicking a member on Reporting → Reports → Member access threw `NG0201` instead of opening the
member edit dialog. Added a regression test that opens the dialog through the real `DialogService` and asserts the
component was constructed.

## 📸 Screenshots

image

### DIFF
--- a/bitwarden_license/bit-web/src/app/admin-console/organizations/organizations-routing.module.ts
+++ b/bitwarden_license/bit-web/src/app/admin-console/organizations/organizations-routing.module.ts
@@ -10,10 +10,16 @@ import {
 } from "@bitwarden/common/admin-console/abstractions/organization/organization.service.abstraction";
 import { FeatureFlag } from "@bitwarden/common/enums/feature-flag.enum";
 import { ConfigService } from "@bitwarden/common/platform/abstractions/config/config.service";
+import { safeProvider } from "@bitwarden/ui-common";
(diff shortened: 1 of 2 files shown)


bitwarden_license/bit-web/src/app/admin-console/organizations/organizations-routing.module.ts
```diff
@@ -88,6 +94,22 @@ const routes: Routes = [
...
+            // MemberActionsService and BillingConstraintService. Both are plain `@Injectable()`
+            // (no `providedIn`), and MembersModule is the only NgModule that provides them — this
+            // route does not load MembersModule, so without these the dialog fails with NG0201.
+            // MemberDialogManagerService is listed only because MemberActionsService injects it.
+            //
+            // These have to live on the route rather than in the component's `providers`:
+            // DialogService parents the dialog's injector to the environment injector DialogService
+            // itself was created in, so a component-decorator (node injector) provider is never in
+            // the dialog's resolution chain. Route `providers` land in an environment injector,
+            // which is. The report component itself uses none of these three services.
```
Banrion, line 107: ⛏️ I suggest avoiding adding comments explaining common patterns. It's standard to need to inject services needed for components in a module

Banrion (approved):
One nit-pick, but otherwise good. Thanks for the test case addition!

sven-bitwarden (approved):

Issue #20148, opened 2026-04-13
Browser integration with Firefox is not enabled until manually creating the NativeMessagingHosts directory


eduardgrebe:
### Steps To Reproduce

1. In a new install on MacOS, enable browser integration
2. FireFox extension does not allow enabling biometric unlock ("Biometric unlock is unavailable because the Bitwarden desktop app is closed" while app is open and unlocked)
3. Disabling and re-enabling does not create ~/Library/Application Support/Mozilla/NativeMessagingHosts/com.8bit.bitwarden.json if the ~/Library/Application Support/Mozilla/NativeMessagingHosts/ directory does not exist, preventing biometric unlock in browser extension
4. Manually create ~/Library/Application Support/Mozilla/NativeMessagingHosts/ and toggle "allow browser integration" off and back on
5. Now com.8bit.bitwarden.json is created and the extension can communicate with the app

### Expected Result

Creation of ~/Library/Application Support/Mozilla/NativeMessagingHosts/ if it does not exist, so that the com.8bit.bitwarden.json file can be created.

### Actual Result

~/Library/Application Support/Mozilla/NativeMessagingHosts/com.8bit.bitwarden.json is not created and biometric unlock cannot be enabled until the directory is manually created

### Additional Context

Firefox 149.0.2 (aarch64)
Browser extension 26.3.0

### Operating System

macOS

### Operating System Version

26.4.1

### Installation method

Homebrew

### Build Version

2026.3.1 (58867)

### Issue Tracking Info

- [x] I understand that work is tracked outside of GitHub. A PR will be linked to this issue should one be opened to address it, but Bitwarden doesn't use fields like "assigned", "milestone", or "project" to track progress.

pamperer562580892423:
Yesss... https://github.com/bitwarden/clients/issues/18996

PR #22409, opened 2026-08-13
[PM-41954] fix(desktop): isolate native messaging per browser

### LINKED ISSUE #12697: Can not enable system authentication(unlock with biometric) on bitwarden extension
### Steps To Reproduce

1. Go to extension
2. Click on settings then account security 
3. Click on unlock with biometric and i get the error below even tho my desktop app is set up.

image

image

### Expected Result

enable unlock with biometric

### Actual Result

image

### Screenshots or Videos

_No response_

### Additional Context

This only happens with chromium browsers i have tried chrome, thorium and chromium but on firefox it works as expected

### Operating System

Linux

### Operating System Version

Linux Mint 22 Cinnamon

### Web Browser

Chrome, Firefox

### Browser Version

latest on all

### Build Version

Version: 2024.12.4  SDK: 'main (edee180)'  Server version (Third-party): 2024.2.0

### Issue Tracking Info

Donnerbart:
## 🎟️ Tracking

Fixes #12697

Follows up on #20148, which added directory creation for Firefox only, and #21783, which made the DuckDuckGo manifest step non-fatal.

## 📔 Objective

`generateManifests()` aborts on the first browser that throws, and `main.ts` calls it immediately before `nativeMessagingMain.listen()`. One failing browser therefore stops the IPC server from ever starting, disabling native messaging for every browser rather than just the one that failed. The only trace is a single line in `app.log`; in the app the symptom is that browser integration and biometric unlock silently do not work.

Two causes observed on Linux:

- A browser config directory that exists without a `NativeMessagingHosts` subdirectory. `mkdir` ran for Firefox only, so `linkOrCopy` threw `ENOENT` for chromium-family browsers. This is the same report as #20148, one browser over, and matches the manual workaround people describe in #12697.
- A browser config directory that exists but cannot be written, for example under snap confinement, giving `EACCES`.

In both cases the browser at fault was one the reporter had never used with Bitwarden.

Whether the feature breaks partially or completely depends on the entry point, because the two call sites order the operations oppositely. Toggling browser integration in settings runs `listen()` first, so the socket survives the throw and browsers earlier in the iteration order keep working, which is why Firefox is often the only one that still works. On app startup the throw happens before `listen()`, so no socket is created and no browser works.

### Changes

- Create the native messaging hosts directory for every browser, not just Firefox, on both macOS and Linux.
- Give each browser its own `try`/`catch`, so a failure is logged and skipped instead of aborting the remaining browsers and `listen()`.
- Remove the caller-side "Hard-linked" log lines. `linkOrCopy` already logs whether it linked or fell back to copying, so the extra line reported a hard link even when the file had been copied.

### Note for reviewers

The directory is now created for chromium-family browsers where it previously was not, so the proxy binary gets copied into browser config directories that exist but have no Bitwarden integration set up yet. That is already how Firefox behaves, and the guard is unchanged in that the browser's own config directory must already exist, but it is a behavior change worth calling out.

Logging stays at `error` for a skipped browser. This condition already surfaced as `logService.error("Error while setting up native messaging:", err)` from `main.ts`, so keeping the level avoids making it harder to find in logs than it is today.

Added `native-messaging.main.spec.ts` covering both behaviors. Both tests fail against the unmodified source.


pamperer562580892423:
Only another user here.

> Follows up on #20148, which added directory creation for Firefox only

FYI, I think https://github.com/bitwarden/clients/issues/20148 was superseded by https://github.com/bitwarden/clients/pull/21784 [EDIT: I was wrong here - and corrected that further down] (and I guess there were other changes too, also in regards to "Shared unlock", mostly done by @quexten)

Donnerbart:
Hmm, the verb superseded throws me off a bit. #20148 is an issue, #21784 is a code change. Normally a code change would supersede another code change and close/fix an issue, not supersede an issue. In the end this doesn't matter. I was referencing issues and code changes with similar symptoms and partial fixes for context and history of the current code.

The issues I experienced match #12697 the best, which I hope to fix with this PR. And yes, I ran into this while setting up "shared unlock".

pamperer562580892423:
> Hmm, the verb superseded throws me off a bit. #20148 is an issue, #21784 is a code change.

Ah, Sorry, my mistake. I did not realize you were quoting an issue with #20148. In fact, there were a few issues around that topic - and I think I confused this issue with https://github.com/bitwarden/clients/pull/21141 (and I think that was superseded by https://github.com/bitwarden/clients/pull/21784).

But anyway, I think BW has this issue on their radar... I'm looking forward to a (any) solution at all.

> And yes, I ran into this while setting up "shared unlock".

Probably not, since "shared unlock" is not released yet. (unless you tried to compile future app versions on your own)

pamperer562580892423:
> The issues I experienced match #12697 the best, which I hope to fix with this PR. And yes, I ran into this while setting up "shared unlock".

Probably totally irrelevant for that PR now (and I probably while hide the comment later), but
- what is your BW desktop app version?
- what is the browser extension version you are using?
- are you self-hosting your Bitwarden server? If yes, what exact server version?

Donnerbart:
> Probably not, since "shared unlock" is not released yet. (unless you tried to compile future app versions on your own)

Oh, but I am, and there is no compilation needed. The code is already released, you just need to set two feature flags in the app and the extension to activate it. With those configured the IPC server is started in the desktop app, so then I ran into this issue. You'd probably also run into this if you try to use the biometric unlock from the desktop app (should be the same connection being used).

> what is your BW desktop app version?
```
Version 2026.7.0
SDK 'main (3e060f8)'
Shell 41.7.2
Renderer 146.0.7680.216
Node 24.15.0
Architecture x64
```

> what is the browser extension version you are using?
```
Firefox
Version: 2026.7.0
SDK: 'main (3e060f8)'
Server version (Third-party): 2026.6.0
```

> are you self-hosting your Bitwarden server? If yes, what exact server version?

This is not relevant, since this is client-only feature. Self-hosted vaultwarden, newest version.

quexten:
> A browser config directory that exists without a NativeMessagingHosts subdirectory. mkdir ran for Firefox only, so linkOrCopy threw ENOENT for chromium-family browsers. This is the same report as https://github.com/bitwarden/clients/issues/20148, one browser over, and matches the manual workaround people describe in https://github.com/bitwarden/clients/issues/12697.

@Donnerbart On which chromium based browser are you seeing this? On fedora the NMHS dir seems to be created by chromium automatically.

The changes seem reasonable to me.

Donnerbart:
> On which chromium based browser are you seeing this? On fedora the NMHS dir seems to be created by chromium automatically.

The first folder that caused problems for me was `microsoft-edge`, which I have of course not installed. I think it stumbled on Vivaldi next. Basically a ton of browsers that I don't have on my system.

I assume some of these folders might be created by 1Password (which I use for work), e.g.:
```
$ ls -la ~/.config/vivaldi/NativeMessagingHosts/
total 12
drwx------ 2 donnerbart donnerbart 4096 Jun 17  2024 .
drwx------ 3 donnerbart donnerbart 4096 Jun 17  2024 ..
-rw------- 1 donnerbart donnerbart  483 Aug 14 10:53 com.1password.1password.json
```

I also saw a GS Connect file somewhere, for another browser I don't use. So I think all these different states of folders that might already exist or not, are caused my multiple software products, setting up their files for native messaging. Combine that with an uninstall and incomplete cleanup, and you quickly end up with a wild mix of `~/.config/<browser>` folders existing, with and w/o nested `NativeMessagingHosts` folders.

On top I had the write permission error on some of these config folders due to the snap sandbox of the bitwarden app (I switched to the flatpack installer to work around that).

That probably made it hard to find _the_ one root cause for all of these issues.

PR #22411, opened 2026-08-13
Auth / PM-32421 - CLI - Login Command - Validate SSO account has a CLI-supported decryption path

JaredSnider-Bitwarden:
## 🎟️ Tracking
https://bitwarden.atlassian.net/browse/PM-32421
https://github.com/bitwarden/clients/issues/18992

## 📔 Objective

Consolidates the post-SSO-login decryption path check into a single guard that short-circuits on master password or Key Connector and otherwise throws a remediation-specific error for TDE, PRF passkey, or an MP-encryption org member who never set an MP. Closes the gap where TDE and PRF-only accounts reached "You are logged in!" without a viable decryption path in CLI.

## 📸 Screenshots

Only recording TDE as the tests prove the other cases + it's the core problem we are fixing with displaying a clearer error. 

https://github.com/user-attachments/assets/fc6ba9c8-29f9-4d78-ae79-3d6c319bbf48


apps/cli/src/auth/commands/login.command.ts
```diff
@@ -647,33 +647,49 @@ export class LoginCommand {
...
+          " Log in through the web vault or browser extension.",
       );
     }
+
+    // Invariant: no MP, no KC, no TDE, no PRF. MP is the intended decryption method
+    // for the user's org but they haven't set an MP yet — JIT'd into an
+    // MP-encryption org without completing setup.
+    throw Response.error(
+      "In order to log in with SSO from the CLI, you must first log in" +
+        " through the web vault, the desktop, or the extension to set your master password.",
```
rr-bw, line 692: Nit (non-blocking): `web vault` should be `web app`
- 3 usages here
- 2 usages in test file

rr-bw (approved):

PR #22416, opened 2026-08-13
[PM-40396] tools export routing change from tools to settings

itsadrago:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-40396

## 📔 Objective

Relocate the export route from tools/export to settings/export and add an unconditional redirect from the old path to preserve bookmarks, deep links, and docs. The export component itself is unchanged (i.e. no UI changes); only its route location moves.


apps/web/src/app/layouts/user-layout.component.html
```diff
@@ -28,7 +28,7 @@
         [spotlight]="true"
       ></bit-nav-item>
       <app-coachmark #importCoachmark stepId="importData" />
-      <bit-nav-item [text]="'exportNoun' | i18n" route="tools/export"></bit-nav-item>
```
vleague2: Is this for VFO1? Should it be feature-flagged?

itsadrago: Yes, I believe so. Good point. Do we need to create a FF or is there an existing one that should be used?

vleague2: Vault already made one! `VFO1Foundation = "vfo1-foundation"`

vleague2 (dismissed):

vleague2 (dismissed):

vleague2 (approved):

PR #22419, opened 2026-08-13
[PM-41830] Shrinking Window Causes Buttons to Overlap Owner Element and Pushes More Options Menu Offscreen

jengstrom-bw:
## 🎟️ Tracking

[Jira](https://bitwarden.atlassian.net/browse/PM-41830)

## 📔 Objective

When shrinking the desktop window, the buttons will eventually overlap the Owner column elements and will push the More Options three dot menu offscreen.  The overlap happens on both Login and Card ciphers, but the More Options issue only happens on Login ciphers.  This PR fixes the overlap and pushing issues.

## 📸 Screenshots

https://github.com/user-attachments/assets/99c9f2dc-c334-4820-b51a-bed32b94500c

### DIFF
--- a/apps/desktop/src/vault/app/vault-v3/vault-list.component.ts
+++ b/apps/desktop/src/vault/app/vault-v3/vault-list.component.ts
@@ -36,7 +36,12 @@ import {
   CheckboxModule,
 } from "@bitwarden/components";
 import { I18nPipe } from "@bitwarden/ui-common";
-import { NewCipherMenuComponent, VaultBatchBarService, VaultItem } from "@bitwarden/vault";
+import {
+  NewCipherMenuComponent,
+  VaultBatchBarService,
+  VaultCopyButtonsService,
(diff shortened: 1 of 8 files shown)


apps/desktop/src/vault/app/vault-v3/vault-list.component.ts
```diff
@@ -45,6 +50,35 @@ import { VaultItemEvent } from "./vault-items/vault-item-event";
 // Fixed manual row height required due to how cdk-virtual-scroll works
 export const RowHeight = 76.5;
 export const RowHeightClass = `tw-h-[76.5px]`;
+
+/**
+ * Width of the Options column, sized to the widest action strip a row can draw.
```
nick-livefront, line 55: 👏 Documentation is appreciated!

nick-livefront (dismissed):

nick-livefront (approved):

PR #22426, opened 2026-08-14
[PM-41974] fix(vault): serialize legacy cipher data field before passing to SDK

themactep:
## 🎟️ Tracking

No ticket — community-reported issue affecting self-hosted (Vaultwarden) users.

## 📔 Objective

Fix the browser extension vault getting stuck on skeleton loaders for vaults served by Vaultwarden.

Vaultwarden persists the cipher `data` field as a deserialized JSON object (the legacy flat cipher view: `name`, `notes`, `password`, `username`, `uri`, `uris`, etc.). `Cipher.toSdkCipher()` forwarded that object verbatim to the SDK, whose `Cipher.data` field is a JSON string. The SDK's serde layer rejects the object:

```
Uncaught Error: invalid type: JsValue(Object({...})), expected a string
```

That error rejects the vault decrypt call, so `loading$` never resolves and the vault shows skeleton loaders indefinitely.

This change normalizes `data` at the SDK boundary: `undefined` when null, the string unchanged when already a string, and `JSON.stringify(...)` otherwise.

## 📸 Screenshots

N/A (no UI change).

### DIFF
--- a/libs/common/src/vault/models/domain/cipher.spec.ts
+++ b/libs/common/src/vault/models/domain/cipher.spec.ts
@@ -1036,6 +1036,45 @@ describe("Cipher DTO", () => {
       });
     });
 
+    it("should serialize a legacy object `data` field to a JSON string", () => {
+      const legacyData = { name: "EncryptedString", notes: null };
+      const cipherData: CipherData = {
+        id: "2afb03fd-0d8e-4c08-a316-18b2f0efa618",
+        edit: true,
+        viewPassword: true,
+        organizationUseTotp: false,
+        favorite: false,
+        revisionDate: "2022-01-31T12:00:00.000Z",
+        type: CipherType.Login,
+        name: "EncryptedString",
+        creationDate: "2022-01-01T12:00:00.000Z",
+        data: legacyData as any,
+      };
+
+      const sdkCipher = new Cipher(cipherData).toSdkCipher();
+
+      expect(sdkCipher.data).toBe(JSON.stringify(legacyData));
+    });
+
+    it("should pass through a string `data` field unchanged", () => {
+      const cipherData: CipherData = {
+        id: "2afb03fd-0d8e-4c08-a316-18b2f0efa618",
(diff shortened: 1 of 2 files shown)


pamperer562580892423:
> No ticket — community-reported issue affecting self-hosted (Vaultwarden) users.

Sorry, I'm only another user.

But...

1) ... I'm not entirely sure what you're trying to fix here... but are you aware of this comment (and the corresponding fix and new Vaultwarden release): https://github.com/dani-garcia/vaultwarden/discussions/7473#discussioncomment-17771043 ?!

2) And second, if there was something to fix for Vaultwarden users - wouldn't you have to fix it in Vaultwarden?

themactep:
> 1. ... I'm not entirely sure what you're trying to fix here... but are you aware of this comment (and the corresponding fix and new Vaultwarden release): [v1.37.0 (Upgrade to this version when using clients v2026.7.0+) dani-garcia/vaultwarden#7473 (reply in thread)](https://github.com/dani-garcia/vaultwarden/discussions/7473#discussioncomment-17771043) ?!

```
$ ./vaultwarden -v
Vaultwarden 1.37.1-0cefa4cc
Web-Vault 2026.1.1
```

> 2. And second, if there was something to fix for Vaultwarden users - wouldn't you have to fix it in Vaultwarden?

The client stopped working with the server, not the other way around. That's why I added the patch to this repository. If the name Vaultwarden is inappropriate here, close the pull request, I don't care. I can fix my own configuration if necessary. I'm simply sharing the fix for the benefit of all.

pamperer562580892423:
> The client stopped working with the server, not the other way around.

@themactep Yeah, but with the **Vaultwarden** server only... (and not the official Bitwarden servers)

And According to the Vaultwarden developers, the clients do work again with Vaultwarden 1.37.0 and higher. See e.g. several comments by @BlackDex, like this one: https://github.com/dani-garcia/vaultwarden/discussions/7501#discussioncomment-17893949

PR #22430, opened 2026-08-14
PAM UAT

patriksvensson:


libs/common/src/vault/abstractions/cipher.service.ts
```diff
@@ -95,6 +95,20 @@ export abstract class CipherService implements UserKeyRotationDataProvider<Ciphe
...
+   * Like {@link getAllDecryptedForIds}, but RETAINS PAM-gated ("partial") rows — ciphers whose
+   * sensitive fields the server suppressed, decrypted into a `partial` view carrying only the
+   * title and, for logins, the URIs.
+   *
+   * Opt-in, for the same reason {@link cipherListViewsWithPartials$} is: a surface that names a
+   * gated cipher (the PAM access-request lists, which are *about* gated ciphers) would otherwise
+   * resolve nothing and fall back to raw ids. Every other caller must use
+   * {@link getAllDecryptedForIds}, which excludes partials.
+   */
+  abstract getAllDecryptedForIdsIncludingPartials(
```
Hinton: Lift up into the vault PR.

libs/common/src/vault/services/cipher.service.ts
```diff
@@ -559,8 +560,24 @@ export class CipherService implements CipherServiceAbstraction {
...
+  ): Promise<CipherView[]> {
+    return this.decryptedForIds(this.cipherViewsWithPartials$(userId), ids);
+  }
+
+  /** Pick the requested ids out of a decrypted-view stream, once it has decrypted. */
+  private decryptedForIds(
+    views$: Observable<CipherView[] | null>,
+    ids: string[],
+  ): Promise<CipherView[]> {
     return firstValueFrom(
```
Hinton: Lift up into vault PR.

libs/vault/src/vault-item-dialog/vault-item-dialog.component.ts
Hinton, line 1: @abergs these are going to be pretty risk. I don't see us adding this in the current vault PR though because we aren't certain what UX we want.

libs/common/src/vault/services/cipher.service.ts
```diff
@@ -729,11 +746,23 @@ export class CipherService implements CipherServiceAbstraction {
...
+      ciphers
+        .filter((c) => c.partialData == null)
+        .map(async (cipher) => {
+          return await cipher.decrypt(key);
+        }),
     );
+    if (gated.length > 0) {
+      const [partialViews, failedViews] = (await this.decryptCiphers(gated, userId)) ?? [[], []];
+      decCiphers.push(...partialViews, ...failedViews);
+    }
```
Hinton, line 767: ~@abergs Keep or revert? If keep we should probably lift it up into the other PR.~ 

This seems dependent on vault rollout of feature flags.

bitwarden_license/bit-web/src/app/pam/access-rules/access-rule-edit/access-rule-edit.component.html
```diff
@@ -1,14 +1,51 @@
...
+      <bit-breadcrumb [route]="['..']">{{ "pamAccessRules" | i18n }}</bit-breadcrumb>
+    </bit-breadcrumbs>
+    <!-- A route-less bit-breadcrumb falls back to a focusable button that does nothing and never
+         gets aria-current; a static span carries the current-page semantics without that trap. -->
+    <span
+      bitTypography="h3"
+      aria-current="page"
+      class="tw-inline-block !tw-m-0 !tw-text-fg-heading"
+      >{{ pageTypeKey | i18n }}</span
+    >
```
Hinton: This seems like using breadcrumbs for something they where never designed for  @maxkpower

I rate this as high risk of breaking when bit-header or breadcrumbs changes, and not suggest this gets removed.

PR #22435, opened 2026-08-14
desktop-autofill: Skip registration confirmation

iinuwa:
## 🎟️ Tracking

[PM-41980](https://bitwarden.atlassian.net/browse/PM-41980)

## 📔 Objective

- Skips the passkey registration confirmation screen if UV is shown
- Shows ciphers that don't have matching URIs but do have FIDO credentials with matching RP ID in the passkey attach/overwrite list.

## 📸 Screenshots

https://github.com/user-attachments/assets/64bb8dae-0de1-496c-8103-922375f82b50

[PM-41980]: https://bitwarden.atlassian.net/browse/PM-41980?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/desktop/src/app/services/services.module.ts
```diff
@@ -472,6 +473,7 @@ const safeProviders: SafeProvider[] = [
       DesktopSettingsService,
       DesktopFido2UserVerificationService,
       PasswordRepromptService,
+      DomainSettingsService,
```
iinuwa, line 476: DomainSettingsService dependency is added in the same order as the constructor, so this should be safe.

iinuwa:
Went down a couple rabbit holes with CI. Lints are from cargo deny, which we're ignoring for now, and the other test is flaky, seemingly based on CI runner workload. Marking as ready to review.

abergs (approved):

PR #22425, opened 2026-08-14
[PM-40318] Introduce Web's VaultNextComponet for VFO1

shane-melton:
## 🎟️ Tracking

[PM-40318](https://bitwarden.atlassian.net/browse/PM-40318)

## 📔 Objective

Introduce the `VaultNextComponent` that uses the new, shared `VaultItemsTableComponent` for VFO1. 

Router query parameter mapping and legacy query param handling coming in a followup PR.

## 📸 Screenshots

Screenshot 2026-08-13 at 5 13 46 PM

[PM-40318]: https://bitwarden.atlassian.net/browse/PM-40318?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #22437, opened 2026-08-14
PM-40380: Add New Item button to Vault Toolbar

nikwithak:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-40380

## 📔 Objective

Moves the Add Item button into the new vault table toolbar. This also replaces the old vault list component with the new vault table, gated behind the `vfo1-foundation` feature flag.

## 📸 Screenshots

image

### DIFF
--- a/apps/web/src/app/vault/individual-vault/vault-header/vault-header.component.html
+++ b/apps/web/src/app/vault/individual-vault/vault-header/vault-header.component.html
@@ -75,7 +75,7 @@
     }
   </ng-container>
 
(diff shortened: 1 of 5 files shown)


nick-livefront (commented):
Does this overlap with @shane-melton's https://github.com/bitwarden/clients/pull/22425?

nikwithak:
@nick-livefront Thanks for catching that - I synced up with @shane-melton  and it seems we ended up with some dupe tickets - closing this one out in favor of his #22425

PR #22444, opened 2026-08-15
[PM-40814] Add key id support

quexten:
Adds support for persisting key ids, and for piping them through to request models, and for backfilling them via an encrypted migration.


libs/common/src/key-management/encrypted-migrator/default-encrypted-migrator.ts
```diff
@@ -43,6 +44,12 @@ export class DefaultEncryptedMigrator implements EncryptedMigrator {
     sdkService: SdkService,
   ) {
     // Register migrations here
+
+    this.migrations.push({
+      name: "User Key Id Backfill Migration",
+      migration: new UserKeyIdBackfillMigration(sdkService, syncService, logService),
+    });
```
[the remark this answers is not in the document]
quexten: Accepted.

shane-melton (dismissed):
Vault owned changes look good!

libs/common/src/key-management/encrypted-migrator/migrations/user-key-id-backfill-migration.ts
```diff
@@ -0,0 +1,78 @@
...
+import { UserId } from "../../../types/guid";
+import { withPasswordManagerSdk } from "../../utils";
+
+import { EncryptedMigration, MigrationRequirement } from "./encrypted-migration";
+
+/**
+ * @internal
+ * Records the id of the user's current UserKey with the server.
+ *
+ * Key ids were introduced after V2 user keys, so accounts created or upgraded before the server
```
mzieniukbw: ❓ We are very close into adding user key ids into V1 user key, why not make it future proof now ?
That way, once we add user key ids into V1, we will only have to change SDK.

libs/common/src/key-management/encrypted-migrator/migrations/user-key-id-backfill-migration.ts
```diff
@@ -0,0 +1,78 @@
...
+    assertNonNullish(userId, "userId");
+
+    try {
+      if (!(await this.needsBackfill(userId))) {
+        return "noMigrationNeeded";
+      }
+
+      // The id the server knows is written to state by the crypto sync handler, so a stale sync
+      // reads as "missing". Another device may also have backfilled since the last sync, and the
+      // server rejects a second write.
```
mzieniukbw: This is somewhat confusing. If another device have backfilled key id, then sync handler would set it in state, therefore the client would reject the migration as not needed. It never reaches the server. The only way for server to reject the second write is if sync failed or when another device backfilled key id literally at the same time, as this migration is executing. (milliseconds - unlikely)

libs/common/src/key-management/encrypted-migrator/default-encrypted-migrator.spec.ts
```diff
@@ -158,6 +166,27 @@ describe("EncryptedMigrator", () => {
...
+      });
+      mockV2KeyRotationMigration.needsMigration.mockResolvedValue("needsMigration");
+      mockV2KeyRotationMigration.runMigrations.mockImplementation(async () => {
+        order.push("v2Rotation");
+      });
+
+      await sut.runMigrations(mockUserId, mockMasterPassword);
+
+      expect(order[0]).toBe("backfill");
+      expect(order).toEqual(["backfill", "kdf", "v2Rotation"]);
```
mzieniukbw, line 188: Maybe this test name should be "should run key id backfill, kdf and v2 rotation in that order" ?

libs/common/src/vault/services/default-cipher-encryption.service.spec.ts
```diff
@@ -651,4 +651,83 @@ describe("DefaultCipherEncryptionService", () => {
...
+        encryptedFor: userId,
+        encryptedByKeyId: keyId,
+      });
+
+      const result = await cipherEncryptionService.encrypt(cipherViewObj, userId);
+
+      expect(result!.encryptedByKeyId).toBe(keyId);
+    });
+
+    it("leaves the key id undefined when the SDK reports none, as for a V1 user key", async () => {
```
mzieniukbw: This is true, but only for short period of time. It would be better to describe this tests as: "leaves the key id undefined when the SDK reports none" and that's it.

mzieniukbw (dismissed):

libs/common/src/key-management/encrypted-migrator/migrations/user-key-id-backfill-migration.ts
quexten: I've cleaned up the comment. If another device backfills, but the local device *did not sync recently* and then unlocks, this would be relevant.

libs/common/src/vault/services/default-cipher-encryption.service.spec.ts
quexten: Removing this test, since the PR for key id for type 2 keys merged, so this should never happen anymore.

mzieniukbw (approved):

shane-melton (approved):

PR #18273, opened 2026-01-08
[PM-30612] Add biometric unlock to CLI via desktop app integration

jeanregisser:
## 🎟️ Tracking

Discussion: https://github.com/orgs/bitwarden/discussions/17934

## 📔 Objective

Bring biometric unlock to the CLI by leveraging the Desktop app's existing biometric infrastructure via IPC.

**Why?** Power users and developers use the CLI daily. Typing a master password repeatedly is friction. With this change, `bw get password github` just works with Touch ID (or Windows Hello, Polkit), same experience as the browser extension.

**How it works:**

- CLI connects to the running Desktop app via Unix socket (macOS/Linux) or named pipe (Windows)
- Uses the same secure IPC protocol as the browser extension
- Falls back gracefully to master password if biometrics unavailable or cancelled

### Happy path: Get password with Touch ID

https://github.com/user-attachments/assets/74aa090c-ac3f-4a19-83c4-49b98a888270

### Cancel Touch ID → falls back to master password

https://github.com/user-attachments/assets/93758df2-90bc-4f9c-ad2b-60536cd0e992

### Desktop app closed (or different logged in user) → falls back to master password

https://github.com/user-attachments/assets/cdb78552-c6c3-4243-a3de-eb3c239275e4

### Fingerprint verification (when enabled in Desktop settings)

https://github.com/user-attachments/assets/bfea78ca-0900-491e-940b-8b18eb3162a4

## ✨ Key Changes

- **New:** `NativeMessagingClient` - handles IPC communication with Desktop app
- **New:** `IpcSocketService` - platform-specific socket connection (macOS sandbox-aware)
- **Updated:** `CliBiometricsService` - implements biometric unlock via Desktop app
- **Updated:** `UnlockCommand` - auto-tries biometrics, falls back to password
- **Updated:** `StatusCommand` - shows `biometricUnlock` status

## ⏰ Reminders before review

- [x] Contributor guidelines followed
- [x] All formatters and local linters executed and passed
- [x] Written new unit tests
- [ ] Protected functional changes with optionality (feature flags) - _N/A: follows browser extension pattern_
- [ ] Used internationalization (i18n) for all UI strings - _CLI doesn't use i18n much yet, wanted to avoid too many changes_
- [x] CI builds passed

## 💬 Notes for reviewers

- This mirrors the browser extension's approach, reuses the same Desktop IPC protocol
- Tested on macOS with the sandboxed (App Store) Bitwarden Desktop app
- Windows/Linux support included but not personally tested
[...]

PR #22448, opened 2026-08-15
[PM-42038] Add desktop biometric unlock support to the CLI

nurfed1:
## 🎟️ Tracking

Discussion: https://github.com/orgs/bitwarden/discussions/17934

Follow-up to https://github.com/bitwarden/clients/pull/18273. The previous implementation was closed pending SDK-based IPC support.

## 📔 Objective

Add biometric unlock to the CLI by communicating with the running Bitwarden Desktop application through the SDK IPC framework.

When `bw unlock` is run interactively without a supplied password, the CLI:

- discovers a compatible Desktop instance;
- checks biometric availability for the active account;
- requests biometric unlock through the existing Desktop RPC handler;
- validates the returned user key before passing it to the existing unlock service; and
- falls back to the master-password prompt when Desktop biometrics are unavailable or cancelled.

Password arguments, `--passwordenv`, `--passwordfile`, and `BW_NOINTERACTION=true` retain their existing behavior and skip the biometric attempt.

The implementation uses `IpcClient.newWithSdkInMemorySessions` and adds a Node socket/named-pipe transport for Desktop's native IPC framing and platform-specific endpoints. It does not reimplement the legacy biometric encryption protocol.

The Desktop IPC bridge currently identifies direct socket connections as `BrowserBackground` clients. The CLI receives its own connection ID and SDK session, but Desktop cannot identify it as a separate CLI client type for routing or policy. This implementation therefore uses the existing browser-client routing and requires no Desktop changes. A dedicated CLI endpoint can be introduced separately if desired.

Biometric unlock requires Bitwarden Desktop 2026.7.0 or newer.

### Validation

- End-to-end biometric unlock verified on Arch Linux with Desktop 2026.7.0.
- CLI Jest suite passes: 26 suites and 287 tests.
- Production and test TypeScript checks pass.
- ESLint and Prettier checks pass.
- OSS and commercial CLI webpack builds succeed.
- Windows and macOS endpoint derivation is covered by unit tests; end-to-end runtime testing has currently only been performed on Linux.

## 📸 Screenshots

Not applicable; this is CLI-only behavior.

### DIFF
--- a/apps/cli/src/key-management/cli-biometrics-service.ts
+++ b/apps/cli/src/key-management/cli-biometrics-service.ts
@@ -1,24 +1,102 @@
+import { firstValueFrom } from "rxjs";
+
+import { AccountService } from "@bitwarden/common/auth/abstractions/account.service";
+import { toTsBiometricsStatus } from "@bitwarden/common/key-management/biometrics-status-mapper";
+import { fromTsUserId } from "@bitwarden/common/key-management/utils";
+import { LogService } from "@bitwarden/common/platform/abstractions/log.service";
 import { UserId } from "@bitwarden/common/types/guid";
 import { UserKey } from "@bitwarden/common/types/key";
-import { BiometricsService, BiometricsStatus } from "@bitwarden/key-management";
+import { BiometricsService, BiometricsStatus, KeyService } from "@bitwarden/key-management";
 // eslint-disable-next-line no-restricted-imports
 import { SymmetricCryptoKey } from "@bitwarden/legacy-crypto";
+import {
+  ipcRequestAuthenticateBiometrics,
+  ipcRequestGetBiometricsStatus,
+  ipcRequestUnlockBiometrics,
+} from "@bitwarden/sdk-internal";
+
+import { CliIpcService } from "../platform/services/cli-ipc.service";
 
 export class CliBiometricsService extends BiometricsService {
+  private readonly noInteractionTimeout = 5_000;
+  private readonly interactionTimeout = 60_000;
+
+  constructor(
+    private accountService: AccountService,
+    private keyService: () => KeyService,
+    private logService: LogService,
+    private ipcService: CliIpcService,
+  ) {
+    super();
+  }
+
   async authenticateWithBiometrics(): Promise<boolean> {
-    return false;
+    try {
+      await this.ipcService.verifyDesktopConnection();
+      return await ipcRequestAuthenticateBiometrics(
+        this.ipcService.client,
+        AbortSignal.timeout(this.interactionTimeout),
+      );
+    } catch (error) {
+      this.logService.info("CLI biometric authentication failed", error);
+      return false;
+    }
   }
 
   async getBiometricsStatus(): Promise<BiometricsStatus> {
-    return BiometricsStatus.PlatformUnsupported;
+    const account = await firstValueFrom(this.accountService.activeAccount$);
+    return account == null
+      ? BiometricsStatus.NotEnabledLocally
+      : await this.getBiometricsStatusForUser(account.id);
   }
 
   async unlockWithBiometricsForUser(userId: UserId): Promise<UserKey | null> {
-    return null;
+    try {
+      await this.ipcService.verifyDesktopConnection();
+      const response = await ipcRequestUnlockBiometrics(
+        this.ipcService.client,
+        fromTsUserId(userId),
+        AbortSignal.timeout(this.interactionTimeout),
+      );
+
+      if (response.user_key == null) {
+        return null;
+      }
+
+      const userKey = SymmetricCryptoKey.fromSdk(response.user_key) as UserKey;
+      if (!(await this.keyService().validateUserKey(userKey, userId))) {
+        this.logService.info("CLI biometric unlock failed: desktop returned an invalid user key");
+        return null;
+      }
+
+      return userKey;
+    } catch (error) {
+      this.logService.info("CLI biometric unlock failed", error);
+      return null;
+    }
   }
 
   async getBiometricsStatusForUser(userId: UserId): Promise<BiometricsStatus> {
-    return BiometricsStatus.PlatformUnsupported;
+    try {
+      const desktopVersion = await this.ipcService.verifyDesktopConnection();
+      this.logService.info(`[IPC] Connected to Bitwarden Desktop ${desktopVersion}`);
+      const status = await ipcRequestGetBiometricsStatus(
+        this.ipcService.client,
+        fromTsUserId(userId),
+        AbortSignal.timeout(this.noInteractionTimeout),
+      );
+      const mappedStatus = toTsBiometricsStatus(status);
+      return mappedStatus === BiometricsStatus.NotEnabledLocally
+        ? BiometricsStatus.NotEnabledInConnectedDesktopApp
+        : mappedStatus;
+    } catch (error) {
+      this.logService.info("Could not query Bitwarden Desktop biometric status", error);
+      return BiometricsStatus.DesktopDisconnected;
+    }
+  }
+
+  disconnect(): void {
+    this.ipcService.disconnect();
   }
 
   async getShouldAutopromptNow(): Promise<boolean> {
@@ -27,7 +105,7 @@ export class CliBiometricsService extends BiometricsService {
 
   async setShouldAutopromptNow(value: boolean): Promise<void> {}
   async canEnableBiometricUnlock(): Promise<boolean> {
-    return false;
+    return (await this.getBiometricsStatus()) === BiometricsStatus.Available;
   }
   async setBiometricProtectedUnlockKeyForUser(
     userId: UserId,
--- a/apps/cli/src/key-management/cli-biometrics-service.ts
+++ b/apps/cli/src/key-management/cli-biometrics-service.ts
@@ -32,7 +32,6 @@ export class CliBiometricsService extends BiometricsService {
 
   async authenticateWithBiometrics(): Promise<boolean> {
     try {
-      await this.ipcService.verifyDesktopConnection();
       return await ipcRequestAuthenticateBiometrics(
         this.ipcService.client,
         AbortSignal.timeout(this.interactionTimeout),
@@ -52,7 +51,6 @@ export class CliBiometricsService extends BiometricsService {
 
   async unlockWithBiometricsForUser(userId: UserId): Promise<UserKey | null> {
     try {
-      await this.ipcService.verifyDesktopConnection();
       const response = await ipcRequestUnlockBiometrics(
         this.ipcService.client,
         fromTsUserId(userId),
@@ -78,8 +76,6 @@ export class CliBiometricsService extends BiometricsService {
 
   async getBiometricsStatusForUser(userId: UserId): Promise<BiometricsStatus> {
     try {
-      const desktopVersion = await this.ipcService.verifyDesktopConnection();
-      this.logService.info(`[IPC] Connected to Bitwarden Desktop ${desktopVersion}`);
       const status = await ipcRequestGetBiometricsStatus(
         this.ipcService.client,
         fromTsUserId(userId),
@@ -93,10 +89,6 @@ export class CliBiometricsService extends BiometricsService {
       this.logService.info("Could not query Bitwarden Desktop biometric status", error);
       return BiometricsStatus.DesktopDisconnected;
     }
-  }
-
-  disconnect(): void {
-    this.ipcService.disconnect();
   }
 
   async getShouldAutopromptNow(): Promise<boolean> {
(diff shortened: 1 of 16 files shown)


quexten:
@nurfed1 Thanks already for the contribution! This looks pretty nice. We'll have an internal discussion on a few aspects on this. The next steps from our side are:
- Code review
- QA

Nice work!

apps/cli/src/platform/services/cli-desktop-ipc.transport.ts
```diff
@@ -0,0 +1,226 @@
...
+  disconnect(): void {
+    this.socket?.destroy();
+    this.socket = undefined;
+    this.connection = undefined;
+    this.messageBuffer = Buffer.alloc(0);
+    this.onDisconnect?.();
+  }
+
+  private async connect(): Promise<net.Socket> {
+    if (this.socket != null && !this.socket.destroyed) {
```
quexten: Thought: It would be nice for this to live in `desktop-native` but we don't have that included in the CLI yet. Not blocking from my side but i'll defer to platform.

apps/cli/src/platform/services/cli-ipc.service.ts
```diff
@@ -0,0 +1,91 @@
...
+  IpcClient,
+  IpcCommunicationBackend,
+  OutgoingMessage,
+  ipcRequestDiscover,
+} from "@bitwarden/sdk-internal";
+
+import { CliDesktopIpcTransport } from "./cli-desktop-ipc.transport";
+
+const DESKTOP_DISCOVER_TIMEOUT_MS = 5_000;
+export const MINIMUM_BIOMETRIC_DESKTOP_VERSION = "2026.7.0";
```
quexten: Please update to `2026.9.0`

apps/cli/src/key-management/commands/unlock.command.ts
```diff
@@ -76,6 +98,35 @@ export class UnlockCommand {
     return this.successResponse();
   }
 
+  private async tryBiometricUnlock(userId: UserId): Promise<boolean> {
```
quexten:
```suggestion
  private async tryUnlockWithDesktop(userId: UserId): Promise<boolean> {
```

To spoil the plans a bit, shared unlock is also implemented now. The desired product experience here is the following.

1. If the desktop app is already unlocked, the CLI will be unlocked as part of that
2. If the desktop app is locked you are able to unlock via the desktop app, in this case via biometric unlock. If you are not able to unlock via biometrics, you may see the desktop app window, and unlock there with pin / master password. However, the lock state will be shared.
3. If you lock via the CLI you will lock the desktop app

This PR doesn't need to implement all that, but let's name the function in a way that supports the above plan. The rest can be follow up.

apps/cli/src/key-management/cli-biometrics-service.ts
```diff
@@ -1,24 +1,102 @@
...
+        ? BiometricsStatus.NotEnabledInConnectedDesktopApp
+        : mappedStatus;
+    } catch (error) {
+      this.logService.info("Could not query Bitwarden Desktop biometric status", error);
+      return BiometricsStatus.DesktopDisconnected;
+    }
+  }
+
+  disconnect(): void {
+    this.ipcService.disconnect();
```
quexten: Biometrics service shouldn't own ipc service connecting and disconnecting. The IPC service should connect either way when the CLI is started, and disconnect when closing. Namely, there are other services that may talk between CLI and desktop (such as shared unlock).

apps/cli/src/key-management/cli-biometrics-service.ts
```diff
@@ -1,24 +1,102 @@
...
+    const account = await firstValueFrom(this.accountService.activeAccount$);
+    return account == null
+      ? BiometricsStatus.NotEnabledLocally
+      : await this.getBiometricsStatusForUser(account.id);
   }
 
   async unlockWithBiometricsForUser(userId: UserId): Promise<UserKey | null> {
-    return null;
+    try {
+      await this.ipcService.verifyDesktopConnection();
```
quexten: Can we move this to run on CLI init?

apps/cli/src/platform/services/cli-desktop-ipc.transport.ts
```diff
@@ -0,0 +1,226 @@
...
+
+/**
+ * Returns the desktop IPC endpoints used by desktop_native/core/src/ipc/mod.rs.
+ *
+ * macOS has separate endpoints for the sandboxed App Store build and the
+ * unsandboxed build, so both are attempted.
+ */
+export function getDesktopSocketPaths(
+  platform = os.platform(),
+  homeDir = os.homedir(),
```
quexten: This currently implements a custom reimplementation of the unix socket protocol for IPC. However, it only implements unix sockets, not the named pipe transport the windows desktop app uses.

The extension currently calls the desktop_proxy binary and talks to it via stdio. Could we please use the same here? It will make this file a lot smaller, and at the same time handle all of the protocol for you.

quexten (changes requested):
Nice work!

I think the main concern here is the re-implementation of the socket protocol. We would prefer to go via desktop-proxy in the same way that the extension works.

apps/cli/src/platform/services/cli-desktop-ipc.transport.ts
nurfed1: Thanks — using `desktop_proxy` for the transport makes sense.

Before changing this, what mechanism should the standalone CLI use to locate the installed proxy? The extension calls `connectNative("com.8bit.bitwarden")`, so the browser resolves the native-host manifest or Windows registry registration and launches it. The CLI has no equivalent browser runtime.

I checked current `main`; the test runner’s `selectProxyPath()` only supports `PROXY_PATH`, a local debug build, and the standard macOS application path. Production locations vary across Windows, macOS, Linux packages, AppImage, Flatpak, and Snap.

Should the CLI resolve existing native-host registrations, search canonical Desktop installation paths, or is there another supported mechanism? I assume failure to locate or connect should remain non-fatal for unrelated CLI commands.

One clarification: the current implementation does support the Windows named-pipe endpoint through `net.createConnection()`. I agree, however, that delegating endpoint discovery and protocol handling to `desktop_proxy` is preferable.

quexten: As you point out there is currently no such mechanism, the browser does it via the native-messaging-manifest which specifies the desktop proxy path (configured by the desktop when writing the native messaging manifests).

I believe adding well known locations, along with an (optional) env var override should be sufficient.

apps/cli/src/platform/services/cli-desktop-proxy-path.ts
```diff
@@ -0,0 +1,61 @@
...
+  environment: NodeJS.ProcessEnv = process.env,
+): string[] {
+  if (platform === "win32") {
+    return [
+      environment.LOCALAPPDATA &&
+        path.win32.join(environment.LOCALAPPDATA, "Programs", "Bitwarden", "desktop_proxy.exe"),
+      environment.ProgramFiles &&
+        path.win32.join(environment.ProgramFiles, "Bitwarden", "desktop_proxy.exe"),
+      environment["ProgramFiles(x86)"] &&
+        path.win32.join(environment["ProgramFiles(x86)"], "Bitwarden", "desktop_proxy.exe"),
```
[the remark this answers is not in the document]
nurfed1: @AikidoSec ignore: This does not read file contents. ProgramFiles(x86) is a standard Windows installation root used to construct the fixed Bitwarden Desktop proxy path. Process-environment overrides are an explicit trust boundary; nonstandard paths can already be supplied through BITWARDEN_DESKTOP_PROXY_PATH.

apps/cli/src/platform/services/cli-desktop-proxy-path.ts
```diff
@@ -0,0 +1,61 @@
...
+    ].filter((candidate): candidate is string => candidate != null);
+  }
+
+  if (platform === "darwin") {
+    return [
+      "/Applications/Bitwarden.app/Contents/MacOS/desktop_proxy",
+      path.posix.join(homeDir, "Applications/Bitwarden.app/Contents/MacOS/desktop_proxy"),
```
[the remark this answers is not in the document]
nurfed1: @AikidoSec ignore: This does not read file contents. This is the standard per-user Bitwarden application installation path. The proxy is an installed companion executable within Bitwarden.app and is intentionally trusted in the same way as the installed Desktop application.

apps/cli/src/platform/services/cli-desktop-proxy-path.ts
```diff
@@ -0,0 +1,61 @@
+import * as fs from "fs";
+import * as os from "os";
+import * as path from "path";
+
+export const DESKTOP_PROXY_PATH_ENV = "BITWARDEN_DESKTOP_PROXY_PATH";
+
+/** Returns well-known locations for the desktop native-messaging proxy. */
```
quexten, line 7: Nice thank you for the adjustments here!

apps/cli/src/platform/services/cli-desktop-ipc.transport.ts
```diff
@@ -0,0 +1,214 @@
...
+
+  disconnect(): void {
+    const proxy = this.proxy;
+    const wasConnected = proxy != null || this.connection != null;
+
+    this.proxy = undefined;
+    this.connection = undefined;
+    this.connected = false;
+    this.messageBuffer = Buffer.alloc(0);
+    proxy?.kill();
```
quexten:
```suggestion
    if (proxy != null) {
      // Close the native messaging stream so the proxy tears the Desktop
      // connection down itself, then terminate it as a fallback.
      proxy.stdin.end();
      proxy.kill();
    }
```

Currently the proxy keeps the process alive

apps/cli/src/platform/services/cli-desktop-proxy-path.ts
```diff
@@ -0,0 +1,61 @@
...
+
+  if (platform === "darwin") {
+    return [
+      "/Applications/Bitwarden.app/Contents/MacOS/desktop_proxy",
+      path.posix.join(homeDir, "Applications/Bitwarden.app/Contents/MacOS/desktop_proxy"),
+    ];
+  }
+
+  return [
+    "/opt/Bitwarden/desktop_proxy",
```
quexten, line 32: Please add `/app/Bitwarden/desktop_proxy` for flatpak

quexten, line 32:
```suggestion
    "/opt/Bitwarden/desktop_proxy",
    "/app/Bitwarden/desktop_proxy"
```

apps/cli/src/platform/services/cli-ipc.service.ts
```diff
@@ -0,0 +1,95 @@
...
+    this.communicationBackend = new IpcCommunicationBackend({
+      send: async (message: OutgoingMessage): Promise<void> => {
+        if (message.destination !== "DesktopMain" && message.destination !== "DesktopRenderer") {
+          throw new Error("CLI IPC only supports Bitwarden Desktop destinations");
+        }
+        await this.transport!.send(message);
+      },
+    });
+
+    // The desktop native IPC server currently assigns direct socket clients a
```
quexten, line 90: Essentially, this currently will pretend to be an extension, and from the desktop perspective look like an extension. I'll track a follow-up ticket to fix this (you don't need to solve that here).

quexten: https://bitwarden.atlassian.net/browse/PM-43322

apps/cli/src/platform/services/cli-desktop-ipc.transport.ts
```diff
@@ -0,0 +1,214 @@
...
+  });
+
+export function encodeNativeMessagingFrame(message: IpcMessage | object): Buffer {
+  const payload = Buffer.from(JSON.stringify(message), "utf8");
+  if (payload.length > MAX_MESSAGE_SIZE) {
+    throw new Error(`Desktop IPC message exceeds ${MAX_MESSAGE_SIZE} bytes`);
+  }
+
+  const frame = Buffer.allocUnsafe(4 + payload.length);
+  if (os.endianness() === "LE") {
```
quexten, line 33: For any other reviewers; this is what chrome / firefox do too.

quexten (changes requested):
Mostly looks good, there seems to be a bug where the CLI just stays open after the unlock. I've suggested a change there.

apps/cli/src/platform/services/cli-desktop-proxy-path.ts
```diff
@@ -0,0 +1,62 @@
...
+  platform = os.platform(),
+  homeDir = os.homedir(),
+  environment: NodeJS.ProcessEnv = process.env,
+): string[] {
+  if (platform === "win32") {
+    return [
+      environment.LOCALAPPDATA &&
+        path.win32.join(environment.LOCALAPPDATA, "Programs", "Bitwarden", "desktop_proxy.exe"),
+      environment.ProgramFiles &&
+        path.win32.join(environment.ProgramFiles, "Bitwarden", "desktop_proxy.exe"),
```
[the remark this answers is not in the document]
quexten: @AikidoSec ignore: See other ignored comments.

quexten (approved):
Looks good! Thank you

quexten:
Will need a review from @bitwarden/team-platform-dev then we can merge and QA it.

dani-garcia (approved):
Platform changes LGTM!

My only concern is around connection timeouts making CLI commands slower than they already are, for example if the desktop app is not installed, or is installed but not open. We should make sure that QA checks for regressions on execution time.

quexten:
I will merge this now but move it behind a feature flag in a follow-up PR.

PR #22449, opened 2026-08-16
[PM-42039] fix(desktop): access keyctl keys from their owning thread

### LINKED ISSUE #22093: Unlock with system authentication toggle silently reverts on Debian 12 — panic "should get process keyring: KeyDoesNotExist" in secure_memory/keyctl.rs
### Steps To Reproduce

1. Install Bitwarden desktop via the official `.deb` package on Debian 12 (bookworm).
2. Log in and unlock the vault (PIN or master password — works fine).
3. Go to Settings → Security → and enable **"Unlock with system authentication"** (biometrics).
4. The toggle switches on momentarily, then immediately reverts to off. No error is shown in the UI.

### Expected Result

Biometric unlock should be enabled and persist.

### Actual Result

The setting silently reverts. Running the app from a terminal with logging enabled (`bitwarden-app --enable-logging`) shows the real failure is an unhandled Rust panic inside the biometrics key-storage code, not a UI bug:

```
11:33:39.609 › [NAPI] [INFO] desktop_core::password::password: OS keyring is_locked=false

thread 'tokio-runtime-worker' (20760) panicked at secure_memory/src/secure_key/keyctl.rs:38:14:
should get process keyring: KeyDoesNotExist
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
11:33:39.613 › [Main Biometrics IPC Listener] setKeyForUser failed Error: Panic in async function
```

This repeats identically on every retry (confirmed 3x in the same session).

### Environment

- Bitwarden Desktop version: 2026.7.0
- Install method: official `.deb` from bitwarden.com
- OS: Debian GNU/Linux 12 (bookworm)
- Kernel: 6.1.0-17-amd64 (Debian 6.1.69-1)
- Desktop environment: KDE Plasma 5.27.5, X11 session
- polkit agent: `polkit-kde-authentication-agent-1` running and functional (verified working for other apps, e.g. KWallet)
- Fingerprint reader: working via `fprintd`/PAM for sudo, screen lock, and other PAM-integrated apps (Tailscale)
- Kernel session keyring: present and healthy (`keyctl show` lists `_ses` and `_uid.1000` keyrings normally)

### Notes / Diagnosis

The panic message ("should get process keyring: KeyDoesNotExist") indicates the code path in `secure_memory/src/secure_key/keyctl.rs` requests the calling process's **process keyring** via a kernel `keyctl` lookup without the "create" flag set. Per `keyrings(7)`, a process keyring is *not* created automatically — the kernel only creates one on demand if explicitly requested with `create=1`. Looking it up without that flag returns `ENOKEY` when no process keyring yet exists for the process, which appears to be the case for every fresh Bitwarden process here (a process keyring is not inherited across `exec`, so this should reproduce on every launch, independent of how the app is started — confirmed via both direct app-menu launch and manual terminal launch).

This looks closely related to (but is a distinct underlying error from) #15790 ("`Unlock with system authentication` setting can't be enabled on debian 13"), which reports the same "toggle silently reverts" symptom on Debian 13 with a different internal error (`clientKeyPartB64` conversion failure). Both point to the Linux biometrics/keyring integration being broken across current Debian releases.

nurfed1:
## 🎟️ Tracking

Fixes #22093.

## 📔 Objective

Linux keyrings are attached to task credentials. The existing implementation lazily created a process keyring when a secure key container was initialized. Electron and Tokio may already have created worker threads by then, so an existing thread could have credentials without that keyring. Reading or removing the key from such a thread failed with `KeyDoesNotExist`.

Creating the process keyring before any additional threads are started would allow those threads to inherit it. However, the secure-memory subsystem initializes after the host runtime may have created threads, an external launcher cannot install a keyring into its parent process, and `execve()` clears process keyrings. Relying on early initialization would therefore make correctness depend on Electron startup ordering.

This change gives each keyctl container a dedicated worker thread and thread keyring. Key creation, reads, and invalidation are all routed through that worker for the lifetime of the container.

It also:

- moves the owned, mlocked `MemoryEncryptionKey` to and from the worker without copying it through ordinary heap memory;
- explicitly zeroes the temporary stack buffer used to read the kernel key payload; and
- joins the worker when the container is dropped.

## Testing

- `cargo +nightly-2026-06-29 fmt --all -- --check`
- `cargo test -p secure_memory --lib -- --test-threads=1`
- `cargo clippy -p secure_memory --all-targets -- -D warnings`
- Repeated the pre-existing-thread regression test 25 consecutive times
- `npm run build-native --workspace @bitwarden/desktop`
- Manually enabled and used Desktop biometric unlock on Arch Linux with `SECURE_KEY_CONTAINER_BACKEND=keyctl`
- Confirmed there were no `secretmem` mappings while using the keyctl backend


quexten (approved):
Nice work!

quexten:
I've handed this over to QA. Please note that QA capacity is fairly low at the moment, so it may take a while until we can merge it.

PR #22452, opened 2026-08-16
[PM-42047] fix(desktop): install the Firefox native messaging manifest on macOS

### LINKED ISSUE #22027: Firefox native messaging manifest is never installed on macOS
### Steps To Reproduce

1. On a Mac that has never had the Bitwarden native messaging manifest installed, install Firefox and the Bitwarden Firefox extension.
2. Install Bitwarden Desktop and enable biometric unlock in the desktop app.
3. Open the Firefox extension and try to unlock with biometrics.

### Expected Result

Firefox unlocks with Touch ID, the same way Chrome does on the same machine.

### Actual Result

Firefox never unlocks. `~/Library/Application Support/Mozilla/NativeMessagingHosts/com.8bit.bitwarden.json` is never created, while the Chrome manifest is correctly rewritten at every launch.

### Screenshots or Videos

_No response_

### Additional Context

**Workaround**

```bash
mkdir -p ~/Library/Application\ Support/Mozilla/NativeMessagingHosts
```

Then fully quit and reopen Bitwarden Desktop. The manifest is written into the now-existing directory, which is inside the sandbox exception, and Firefox biometric unlock starts working.

**Related:** #18996, #21046, #13447

### Operating System

macOS

### Operating System Version

26.5.2 (25F84)

### Installation method

Mac App Store

### Build Version

2026.7.0 (65236)

### Issue Tracking Info

fabriziobagala:
## 🎟️ Tracking

Fixes #22027

## 📔 Objective

On macOS the Firefox native messaging manifest is never installed, so biometric unlock never works in Firefox while it works in Chrome on the same machine.

`getDarwinNMHS()` returns one path per browser, used both to detect that the browser is installed and as the directory to install into. That holds for every browser except Firefox: its profiles live in `Application Support/Firefox/`, but its manifests belong in `Application Support/Mozilla/`, which Firefox itself does not create - Bitwarden or another native messaging host does. Gating installation on `existsSync()` of `Mozilla/` therefore makes the manifest its own precondition, and Firefox is skipped forever with `Firefox not found, skipping.` It is also why the Firefox/Zen `mkdir` from #21784 never runs on an affected machine: it sits inside the branch that is never taken.

The fix is limited to detection. A new `getDarwinDetectPaths()` returns the paths that prove a browser is installed - for Firefox, `Application Support/Firefox/` plus the existing `Mozilla/` as a fallback; for every other browser, the current path unchanged. `getDarwinNMHS()` and `writeManifest()` are untouched, so #21568 is not reintroduced. Installing each browser's manifest is now wrapped in a try/catch: Firefox is the first entry in the map, and at startup `main.ts` calls `generateManifests()` and `listen()` inside one `try`, so a failure there would stop the native messaging server from starting at all.

This does mean `Mozilla/` can be created when it is missing, which the note on #21784 rules out ("We do not want to create the browser directory"). Nothing is created for a browser that is not installed - `Mozilla/` only appears once `Application Support/Firefox/` has been observed, and only on macOS - but the letter of that rule is not preserved, and it deserves an explicit decision. Linux, Snap and Flatpak are unchanged.

Both MAS entitlements gain one entry - `/Library/Application Support/Mozilla`, without a trailing slash - alongside the existing `NativeMessagingHosts/` subpath, which is left untouched: a subpath exception rooted at `Mozilla/NativeMessagingHosts` cannot authorize creating its own parent. Widening the leaf entry to `Mozilla/` instead would have made it the only non-leaf grant of the thirteen, and would have reached `Mozilla/ManagedStorage/` and `Mozilla/Certificates/`. This is the one part I cannot exercise - an unsandboxed local build permits the write regardless - so it is worth confirming on a signed MAS build.

Also adds `native-messaging.main.spec.ts`, which the module did not have.


pamperer562580892423:
@fabriziobagala  Just FYI, I think someone else proposed a fix for this issue a few days ago: https://github.com/bitwarden/clients/pull/22409 (also @Donnerbart)

fabriziobagala:
> @fabriziobagala Just FYI, I think someone else proposed a fix for this issue a few days ago: #22409 (also @Donnerbart)

@pamperer562580892423 Thanks for the info, but I hadn't realized anyone was already working on it. I created this PR because the bug had become too annoying, so I tried to fix it. On my MacBook, I applied the workaround I described in the issue. I hope this PR is helpful; if not, feel free to close it.

apps/desktop/src/main/native-messaging.main.ts
```diff
@@ -191,18 +191,25 @@ export class NativeMessagingMain {
...
             const nmhsPath = path.join(browserDirectory, "NativeMessagingHosts");
             const manifestPath = path.join(nmhsPath, "com.8bit.bitwarden.json");
 
-            let manifest: any = await this.generateChromeJson(binaryPath);
-            if (key === "Firefox" || key === "Zen") {
-              // Only generate the NMHS dir if the browser directory exists
-              await fs.mkdir(nmhsPath, { recursive: true });
-              manifest = await this.generateFirefoxJson(binaryPath);
-            }
+            try {
```
quexten, line 202: Nit (non-blocking): We should probably just make writemanifest, mkdir non fallible. I.e make it a try catch and call them tryMkdir, tryWriteManifest instead.

quexten (approved):
Approving from KM's side. Looked at this recently and this seems like a good way to fix the firefox quirk of having a different NMHS dir than compared to the browser dir.

addisonbeck (approved):

addisonbeck:
sending this over to our QA team for testing, then we'll get it merged. Thanks for the contribution.

apps/desktop/src/main/native-messaging.main.ts
fabriziobagala: Agreed, `tryMkdir` / `tryWriteManifest` would read better than wrapping the loop body.

I kept it out of this PR because `writeManifest` is shared with the win32 and linux branches, and the `nativeMessaging.manifests` IPC handler relies on `generateManifests()` throwing to report failures to the renderer. Making the primitives non-fallible would change that on every platform, which felt out of scope for a macOS detection fix.

PR #22480, opened 2026-08-17
widen engines pin to include Node 24

### LINKED ISSUE #22164: `@bitwarden/cli` 2026.7.0 cannot be installed on Node 24 (current Active LTS) — new `engines` pin to `~22`
### Steps To Reproduce

1. Use Node 24 (the current Active LTS line, e.g. 24.18.0).
2. Install the CLI from npm, e.g. `yarn global add @bitwarden/cli` (any installer that enforces `engines` fails; npm shows an `EBADENGINE` warning).

### Expected Result

The npm package supports the latest Node.js LTS release. Per the [official Node.js release schedule](https://github.com/nodejs/Release/blob/main/schedule.json), Node 24 "Krypton" has been LTS since 2025-10-28, and Node 22 "Jod" has been in Maintenance since 2025-10-21.

### Actual Result

Installation fails because 2026.7.0 declares `"engines": { "node": "~22" }`:

https://github.com/bitwarden/clients/blob/355ac6791b88b00e0a92a19907b016c104d20a66/apps/cli/package.json#L94-L97

```
error @bitwarden/cli@2026.7.0: The engine "node" is incompatible with this module. Expected version "~22". Got "24.18.0"
error Found incompatible module.
```

### Screenshots or Videos

N/A (full error output above)

### Additional Context

- This is an install regression: versions up to 2026.6.0 published no `engines` field and installed cleanly on Node 24. The pin was introduced in 2026.7.0 by #20400.
- #20400 kept the CLI on Node 22 pending the @yao-pkg/pkg upgrade (#21949), but that blocker only applies to the packaged standalone binary — the npm-distributed package runs on the user's own Node. Could `engines` be widened (e.g. `"~22 || ~24"`) independently of the pkg upgrade?
- The monorepo itself already develops and builds on Node 24, so the shipped CLI refusing Node 24 at install time seems inconsistent — the root `package.json` requires `>=24.17.0` and `.nvmrc` is `v24`:

https://github.com/bitwarden/clients/blob/355ac6791b88b00e0a92a19907b016c104d20a66/package.json#L243-L246

https://github.com/bitwarden/clients/blob/355ac6791b88b00e0a92a19907b016c104d20a66/.nvmrc#L1

### Operating System

Linux

### Operating System Version

Debian-based (kernel 6.18)

### Shell

Zsh

### Build Version

2026.7.0 (cannot run `bw --version` — installation fails)

### Issue Tracking Info

addisonbeck:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-41258
Relates to https://github.com/bitwarden/clients/pull/20400
Fixes https://github.com/bitwarden/clients/issues/22164

## 📔 Objective

Set node 22 as the supported floor for the CLI, allowing installs on Node 24 without warning without breaking our need to build on 22 for our problematic dependencies.


djsmith85 (approved):
Approving for platform

This works as our packaged CLI explicitly uses Node 22. It uses https://github.com/bitwarden/clients/blob/main/apps/cli/.nvmrc during build-cli.yml instead of the root .npmrc

PR #22484, opened 2026-08-17
[CL-1236] Extension header changes

BryanCunningham:
## 🎟️ Tracking

[CL-1236](https://bitwarden.atlassian.net/browse/CL-1236)

## 📔 Objective

Splits `popup-header` into two bars behind the `vfo1-foundation` flag:

- an **app bar** — dark, branded, carrying the Bitwarden logo plus the chrome that belongs to the
  extension rather than the page (pop out button, account switcher). Renders on every screen that
  uses `popup-header`, dialogs and onboarding pages included, and stays pinned while the page
  scrolls.
- a **page title bar** — the familiar header, carrying the back button, a leading icon tile, the
  title, and an item count. The popup viewport is short, so it collapses while the user scrolls down
  and returns as soon as they scroll back up, or when it takes focus via Shift+Tab.

Both bars live inside a single `<header>`, so the banner landmark covers the whole region in either
flag state.

**With the flag off, the header renders exactly as it does on `main`.** All new styling is gated;
`<header>` keeps painting the one-bar header, which also keeps it reachable by the `[&_header]:`
overrides that pages such as the default password manager prompt rely on.

Also in this PR:

- New slots, all documented in `popup-layout.mdx`: `start` and `title-end` (title bar, flag-gated)
  and `title-suffix` (a control belonging to the title itself, not flag-gated). `end` moves to the
  app bar when that bar exists, with no change on the calling side.
- The legacy default slot is dropped while the flag is on. Its only two consumers — the anon layout
  wrapper and the default password manager prompt — project a Bitwarden logo that the app bar would
  otherwise duplicate.
- `background` is retained as a signal input: `"alt"` is transparent and borderless for pages that
  paint their own background, and applies to the page title bar in the two-bar header.
- New `scrollDirection` util in `libs/components`, which handles jitter, clamping, and programmatic
  jumps, with specs.
- `popup-header` migrated to signal inputs and `OnPush`, clearing its CL-764 and CL-903 suppressions.

### ⚠️ Needs a design look

With the flag on, the default password manager prompt and the anon layout pages now render the dark
app bar over their own illustration / branded background. That follows from the app bar being
extension chrome present on every screen, but there is no design for those two screens specifically
— see the last two screenshot rows.

## 📸 Screenshots

| Screen                     | Flag off | Flag on |
| -------------------------- | -------- | ------- |
| Vault                      | image | image |
| Settings                   |  image | image |
| Send | image | image |
| Generator           | image | image |
| Login (anon layout)        | image | image |

[CL-1236]: https://bitwarden.atlassian.net/browse/CL-1236?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/browser/src/platform/popup/layout/popup-layout.stories.ts
```diff
@@ -656,6 +652,112 @@ class MockVaultTablePageComponent {
...
+          <mock-current-account></mock-current-account>
+        </ng-container>
+        <bit-icon-tile slot="start" icon="bwi-vault" variant="brand" size="sm"></bit-icon-tile>
+        <h1 bitTypography="h3" class="!tw-mb-0.5">My vault</h1>
+        <button
+          type="button"
+          bitIconButton="bwi-angle-down"
+          size="small"
+          label="Switch vault"
+        ></button>
```
vleague2: I was thinking that the title would still be provided in the normal `pageTitle` input, with the button being something like `title-suffix` slot. I feel like this custom override approach would be useful if we needed to fully override the title, but it seems like something we could accomplish with less variability/room for erroneous setup since the custom title itself is pretty much identical to the built-in one

apps/browser/src/platform/popup/layout/popup-layout.stories.ts
```diff
@@ -656,6 +652,112 @@ class MockVaultTablePageComponent {
...
+@Component({
+  selector: "mock-vault-page-v2",
+  changeDetection: ChangeDetectionStrategy.OnPush,
+  template: `
+    <popup-page>
+      <popup-header slot="header" [pageTitle]="''">
+        <ng-container slot="app-actions">
+          <mock-popout-button></mock-popout-button>
+          <mock-current-account></mock-current-account>
+        </ng-container>
```
vleague2: If we're keeping the API where consumers have to pass in these actions, should we leave the slot name the same as it is today so that there is less migration/breaking change? (It looks like it's `end` today)

Alternatively, we handle the buttons internally and don't ask consumers to pass them in. They are identical across pages (I asked Emily in some Figma thread somewhere), even on the new login page, with the exception of the popout button not showing up when the browser is popped out.

vleague2 (commented):
Very initial thoughts (leaving what I've got so far while I get distracted by something else)

apps/browser/src/platform/popup/layout/popup-layout.stories.ts
BryanCunningham: Ok, I can move to that setup. Makes sense to me

BryanCunningham:
Addressed the styling findings in 461f745849.

**Findings 1 and 2 — the `[&_header]:` overrides and the un-gated `background` change.** Both came from the same root cause: background, border, and horizontal padding moved off `<header>` onto the bars inside it, without flag gating. `<header>` now carries the one-bar styling again via a `headerClasses()` computed that returns `""` when the flag is on, and `background` is back as a signal input. **With the flag off the header renders exactly as it does on `main`** — same tokens (`bg-background` / `bg-background-alt`, `secondary-300`), same asymmetric padding, same `alt` border-on-scroll behavior (which needed `PopupPageComponent.isScrolled` restored). Two new specs guard against the regression recurring: one asserting `<header>` paints itself and the title bar carries no classes, one covering the `alt` transparent-until-scrolled case. With the flag on, `background="alt"` applies to the page title bar and leaves the app bar alone, so the illustration stays visible in both states.

**Finding 3 — the app bar on every `popup-header`.** Intended. The app bar is extension chrome rather than page content, so it belongs on dialogs and onboarding pages too. The real problem underneath it was duplicate branding, which is now fixed properly: the legacy default slot is dropped while the flag is on, since its only two consumers (the anon layout wrapper and the default password manager prompt) used it to project a Bitwarden logo the app bar would otherwise render twice. Documented in `popup-layout.mdx`.

**PR metadata.** Description filled in with the tracking link and objective. Screenshots are still to come — the table is there with a row per screen and flag state. Worth flagging while you look at them: with the flag on, the prompt page and the anon layout pages render the dark app bar over their own illustration, and no design exists for those two screens specifically. Called out in the description for a design look rather than blocking here.

apps/browser/src/autofill/popup/default-password-manager/default-password-manager-prompt.component.html
```diff
@@ -9,7 +9,7 @@
       slot="header"
       background="alt"
       pageTitle=""
-      class="[&_header]:!tw-bg-transparent [&_header]:!tw-px-6 bit-compact:[&_header]:!tw-px-3"
```
vleague2, line 12: Shouldn't we leave the existing classes untouched for non-VFO1?

apps/browser/src/platform/popup/layout/popup-header.component.html
```diff
@@ -1,34 +1,84 @@
-<!-- 
-  end padding is less than start padding to prioritize visual alignment when icon buttons are used at the end of the end slot.
-  other elements used at the end of the end slot may need to add their own margin/padding to achieve visual alignment.
+<!--
+  Each slot is declared exactly once. A `select` repeated across branches of a control flow block
+  would bind all matching content to whichever copy comes first, leaving the other branch empty.
+
+  `end` is the trailing edge of the topmost bar, so it belongs to the app bar when that bar exists and
+  to the title bar otherwise. Declaring it here and rendering it into whichever branch is live is the
+  only way to move it between the two.
```
vleague2: Imo this comment is unnecessary, we do this template wrapper approach everywhere and it doesn't need explanation

apps/browser/src/platform/popup/layout/popup-header.component.html
```diff
@@ -1,34 +1,84 @@
...
+  <ng-content select="[slot=end]"></ng-content>
+</ng-template>
+
+<!--
+  One `header` around both bars, so the banner landmark covers the whole region regardless of the flag.
+  The bars themselves are plain containers: the app bar is chrome belonging to the extension, the title
+  bar belongs to the page, and neither is a landmark on its own.
+
+  `header` carries the styling of the one-bar header and nothing else. Two bars each paint themselves,
+  so with the flag on this is a bare landmark.
```
vleague2: I can't figure out what Claude means by "bare landmark" lol

apps/browser/src/platform/popup/layout/popup-header.component.html
```diff
@@ -1,34 +1,84 @@
...
+        @if (showBackButton()) {
+          <button
+            bitIconButton="bwi-angle-left"
+            type="button"
+            [label]="'back' | i18n"
+            [bitAction]="backAction()"
+          ></button>
+        }
+        @if (vfo1Enabled()) {
+          <ng-content select="[slot=start]"></ng-content>
```
vleague2: Okay nitpicking but the `end` slot being in a different row than the `start` slot is a bit confusing. Maybe we do `title-start` to indicate that it is not an equivalent placement to the `end` slot, it is an equivalent to the `title-end` slot

apps/browser/src/platform/popup/layout/popup-header.component.spec.ts
```diff
@@ -0,0 +1,342 @@
...
+    it("renders the title-suffix slot, which the flag does not gate", () => {
+      expect(titleBar().contains(slot("title-suffix"))).toBe(true);
+    });
+
+    /**
+     * `header` has to keep painting the one-bar header. Pages such as the default password manager
+     * prompt reach it with `[&_header]:` overrides, which silently stop working if the background or
+     * the horizontal padding moves onto a descendant.
+     */
+    describe("styling", () => {
```
vleague2: Not sold on this block of tests, you know me and the css tests

(Also a couple other styling tests in this file)

apps/browser/src/platform/popup/layout/popup-header.component.ts
```diff
@@ -1,67 +1,199 @@
...
-  background: "default" | "alt" = "default";
+  protected readonly logo = BitwardenLogo;
+
+  /**
+   * Background treatment of the page title bar.
+   *
+   * - `"default"` sits on an opaque background with a bottom border
+   * - `"alt"` is transparent and borderless, for pages that paint their own background
+   *
+   * Not gated by the flag: a page that owns its background owns it in either header.
```
vleague2: This comment feels like something we will miss when we remove the VFO1 flag and then it will be confusing

apps/browser/src/platform/popup/layout/popup-header.component.ts
```diff
@@ -1,67 +1,199 @@
...
-  };
+  });
+
+  /**
+   * The popup viewport is short, so the title bar gets out of the way while the user reads down the
+   * page. The app bar stays pinned.
+   */
+  private readonly scrollDirection = scrollDirection(this.scrollLayout.scrollableRef);
+
+  /** Only the v1 header uses this, to grow a border under an `alt` bar once the page moves. */
```
vleague2: I think same thought applies for all these comments referencing the flag/v1 -- what is actually necessary for understanding the input? Where do we need flag references, and in those spots, can we leave some more explicit references like "remove when the VFO1 flag is deleted" so that it is more obvious and easier to find/delete later?

apps/browser/src/platform/popup/layout/popup-layout.stories.ts
```diff
@@ -1057,6 +1178,58 @@ export const TransparentHeader: Story = {
...
 };
 
+/**
+ * Under the `vfo1-foundation` flag, `popup-header` renders two bars: a branded app bar carrying the
+ * pop out button and account switcher, and a page title bar carrying the leading icon tile, title,
+ * and item count. Toggle the flag in the Feature Flags addon panel to compare against v1.
+ */
+export const HeaderV2: Story = {
+  globals: enabledFlags(FeatureFlag.VFO1Foundation),
+  parameters: { chromatic: { modes: featureFlagModes(FeatureFlag.VFO1Foundation) } },
```
vleague2: I think this mode thing is making it take 2 snapshots, one with flag on and "off" (I put that in quotes because Chromatic says it's off but clearly it is on lol)

apps/browser/src/platform/popup/components/pop-out.component.ts
```diff
@@ -19,6 +23,28 @@ export class PopOutComponent implements OnInit {
...
+    this.configService?.getFeatureFlag$(FeatureFlag.VFO1Foundation) ?? of(false),
+    { initialValue: false },
+  );
+
+  /**
+   * Every usage projects into the header's `end` slot, which the flag moves onto the dark app bar.
+   * Only `side-nav` is legible against that background.
+   */
+  protected readonly buttonType = computed<ButtonType>(() =>
+    this.vfo1Enabled() ? "side-nav" : "primaryGhost",
```
vleague2, line 45: 💭 Guess later we'll need to think about renaming this variant since it's not just `side-nav` anymore

apps/browser/src/platform/popup/layout/popup-page.component.html
```diff
@@ -6,14 +6,14 @@
...
-    [ngClass]="{
-      'tw-invisible !tw-p-0 !tw-border-none': loading() || nonScrollable.childElementCount === 0,
-      'tw-border-secondary-300': scrolled(),
-      'tw-border-transparent': !scrolled(),
-    }"
+    class="tw-transition-colors tw-duration-200 tw-border-0 tw-border-b tw-border-solid tw-border-transparent tw-py-3 bit-compact:tw-py-2 tw-px-[max(0.75rem,calc((100%-(var(--tw-sm-breakpoint)))/2))] bit-compact:tw-px-[max(0.5rem,calc((100%-(var(--tw-sm-breakpoint)))/2))]"
+    [class.tw-invisible]="hideNonScrollable"
+    [class.!tw-p-0]="hideNonScrollable"
+    [class.!tw-border-none]="hideNonScrollable"
+    [class.!tw-border-border-base]="scrolled()"
```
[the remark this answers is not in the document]
vleague2: @BryanCunningham Is this still relevant?

vleague2 (dismissed):

apps/browser/src/platform/popup/layout/popup-page.component.html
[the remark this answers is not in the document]
BryanCunningham: Yes, but they should all just be border-base probably. I'll update them

vleague2 (dismissed):

vleague2 (approved):

jprusik (approved):
Approved for Autofill team ownership concerns

JaredSnider-Bitwarden (approved):
Two auth file changes LGTM!

shane-melton (approved):

addisonbeck (approved):
Approved the platform changes, which are simple. I think long term we probably shouldn't own pop-out.component. Not sure who else would though 🤔

PR #22492, opened 2026-08-17
fix(pam): simplify access-rules table badges and status labels

maxkpower:
## 📔 Objective

Simplifies the Access rules table as specified in the Figma Designs.


JaredScar:
Tests are failing because you cannot change existing keys in `messages.json` file. If you want to change copies, you must introduce new keys and use the new keys in the codebase. Existing ones cannot be modified, only deleted.

JaredScar (changes requested):
Need to fix failing tests (see my earlier comment)

PR #22494, opened 2026-08-17
Auth/PM-41523 - Organization Invite Service - Introduce NoOp and make DI match client support for org invites

JaredSnider-Bitwarden:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-41523

## 📔 Objective

Claude Summary: 

`DefaultOrganizationInviteService` is bound globally, forcing every Angular client to carry a full, **web-only** invite service. `WebRegistrationFinishService` overrides two methods purely to bridge the null-returning stubs the shared default provides.

This PR binds a `NoopOrganizationInviteService` globally, registers `Default` in web's core module only, and inlines the invite reads into `RegistrationFinishComponent` — dropping sixteen unused DI deps from non-web clients and removing the bridging overrides.

## 📸 Screenshots

Direct Org Invite --> New User Registration works

https://github.com/user-attachments/assets/b582998c-e910-4a2d-b24d-533909f0e529

Open Org Invite --> New User Registration works

https://github.com/user-attachments/assets/d8a7a704-094d-496b-b917-36d1ac45f20b


apps/web/src/app/auth/core/services/registration/web-registration-finish.service.ts
```diff
@@ -51,46 +46,6 @@ export class WebRegistrationFinishService
...
-    }
-
-    const masterPasswordPolicyOpts: MasterPasswordPolicyOptions = await firstValueFrom(
-      this.policyService.masterPasswordPolicyOptions$(null, policies),
-    );
-
-    return masterPasswordPolicyOpts;
-  }
-
   override async buildSdkRegisterRequest(
```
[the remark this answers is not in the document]
JaredSnider-Bitwarden: Fixed with https://github.com/bitwarden/clients/pull/22494/commits/07f58058ab3932bd24489019caac0eb1b44b7c2f

coroiu (approved):
Approving core module changes

libs/common/src/auth/organization-invite/services/implementations/noop-organization-invite.service.ts
```diff
@@ -0,0 +1,135 @@
...
+
+  async unsealOpenOrgInvite(
+    _email: string,
+    _sealedData: string,
+  ): Promise<OpenOrgInviteUnsealResult> {
+    throw new Error(
+      "OrganizationInviteService.unsealOpenOrgInvite called on NoopOrganizationInviteService — " +
+        "DefaultOrganizationInviteService must be registered on this client to unseal invite blobs.",
+    );
+  }
```
rr-bw, line 126: There is a _non-blocking_ regression in an edge case here:

`RegistrationFinishComponent` reads from URL params before deciding to execute invite-specific logic via `tryHydrateOpenInviteFromSealedData()`. So a user _could_ open the Extension via a web browser tab and construct the URL + invite query params, and ultimately trigger `unsealOpenOrgInvite()`.

Claude investigated and determined that this would introduce a regression in graceful error handling. See the details and Claude's recommended fix in the collapsible "Details and fix" below:

> ⚠️ **IMPORTANT**: `unsealOpenOrgInvite` throwing is reachable on browser/desktop and regresses a currently-handled error into an unhandled rejection.
> 
> <details>
> <summary>Details and fix</summary>
> 
> Unlike the other throw-group members, `unsealOpenOrgInvite` is not behind a `getOpenOrgInvite() != null` guard. In `RegistrationFinishComponent.tryHydrateOpenInviteFromSealedData()` the only gate is:
> 
> ```ts
> // registration-finish.component.ts:239
> if (this.sealedOpenOrgInviteData == null || !this.email) {
>   return true;
> }
> // ...feature-flag guard...
> const unsealResult = await this.organizationInviteService.unsealOpenOrgInvite(
>   this.email,
>   this.sealedOpenOrgInviteData,
> );
> ```
> 
> Both values come straight from query params (`handleQueryParams`, lines 355 / 381). `RegistrationFinishComponent` is routed on browser (`apps/browser/src/popup/app-routing.module.ts:438`) and desktop (`apps/desktop/src/app/app-routing.module.ts:167`), and an extension page can be opened in a full tab (`chrome-extension://<id>/popup/index.html#/finish-signup?email=…&sealedOpenOrgInviteData=…`). Neither `tryHydrateOpenInviteFromSealedData()` nor its `ngOnInit()` caller has a `try/catch`, so the throw surfaces as an unhandled promise rejection: `viewState` stays on `Loading` (line 143), no error UI renders, and `stripSealedOpenOrgInviteDataFromUrl()` never runs so a refresh retries.
> 
> On `main` today the same URL hits `DefaultOrganizationInviteService.unsealOpenOrgInvite`, which reads `EMAIL_SEALED_OPEN_ORG_INVITE_SECRET_RECORD_DISK_LOCAL` — physically per-client `disk-local` storage the extension can never have written to — misses, and returns `{ kind: "secret-miss" }` (default-organization-invite.service.ts:534-538). The component already handles that gracefully via `logOpenOrgInviteUnsealError` + `showSealedOpenOrgInviteDecryptionFailed()`.
> 
> So this method isn't really a DI-misconfiguration signal: "no paired secret on this client" is the honest, always-correct answer on a client that can never have sealed one — same reasoning that puts `getOrganizationInvite`/`getOpenOrgInvite` in the read group.
> 
> Suggested fix:
> 
> ```ts
> async unsealOpenOrgInvite(
>   _email: string,
>   _sealedData: string,
> ): Promise<OpenOrgInviteUnsealResult> {
>   // Read-shaped: this client can never have written a paired HighEntropySecret
>   // (disk-local, per-client), so "secret-miss" is the truthful answer and lets
>   // RegistrationFinishComponent render its classified error UI instead of crashing.
>   return { kind: "secret-miss" };
> }
> ```
> 
> `getOpenOrgInviteStatus` and `setOrganizationInvite` can stay in the throw group — both are only reached after a successful unseal (lines 266, 295), so they become unreachable once this returns `secret-miss`.
> 
> Scope note: today the throw is additionally gated by `FeatureFlag.GenerateInviteLink` (default `FALSE`), so this is latent until that flag rolls out — but it ships with the flag.
> 
> </details>

I asked Claude to do further investigation to verify that all `OrganizationInviteService` call sites for "throw" methods are either truly gated behind some sort of "org invite is null" guard, _or_, if the call site is gated behind URL params (like the `unsealOpenOrgInvite()` case above), that we make sure to handle the error gracefully.

Here is the output of that verification (TLDR: all are fine except `unsealOpenOrgInvite`):

> <details>
> <summary>Details</summary>
> 
>   Verified all 7 throw-group methods against their actual call sites. Result: 6 of 7 are safe;
>   unsealOpenOrgInvite is the sole exception, confirmed again.
> 
>   Method: validateAndAcceptDirectOrgInvite
>   Call site(s): accept-org-direct-invite.component.ts
>   Gate: web-only file, doesn't ship on other clients
>   Verdict: ✅ Safe
>   ────────────────────────────────────────
>   Method: acceptOpenOrgInvite
>   Call site(s): accept-org-open-invite.component.ts
>   Gate: web-only file
>   Verdict: ✅ Safe
>   ────────────────────────────────────────
>   Method: setOrganizationInvite
>   Call site(s): 2× web-only accept components
>   Gate: web-only files
>   Verdict: ✅ Safe
>   ────────────────────────────────────────
>   Method: setOrganizationInvite
>   Call site(s): registration-finish.component.ts:295 (shared)
>   Gate: reached only after unsealOpenOrgInvite returns kind === "ok" (line 259-263 early-returns otherwise)
>   Verdict: ✅ Safe transitively — unseal fails first
>   ────────────────────────────────────────
>   Method: getOpenOrgInviteStatus
>   Call site(s): accept-org-open-invite.component.ts
>   Gate: web-only file
>   Verdict: ✅ Safe
>   ────────────────────────────────────────
>   Method: getOpenOrgInviteStatus
>   Call site(s): registration-finish.component.ts:266 (shared)
>   Gate: same unseal-success gate as above
>   Verdict: ✅ Safe transitively
>   ────────────────────────────────────────
>   Method: validateOpenOrgInviteEmailDomain
>   Call site(s): login.component.ts:650, registration-start.component.ts:324 (shared)
>   Gate: preceded by getOpenOrgInvite() == null check — Noop hardcodes return null, not derived from URL/state
>   Verdict: ✅ Safe, unconditional
>   ────────────────────────────────────────
>   Method: sealOpenOrgInvite
>   Call site(s): registration-start.component.ts:280 (shared)
>   Gate: preceded by getOpenOrgInvite() == null check
>   Verdict: ✅ Safe, unconditional
>   ────────────────────────────────────────
>   Method: unsealOpenOrgInvite
>   Call site(s): registration-finish.component.ts:251 (shared)
>   Gate: preceded only by sealedOpenOrgInviteData == null || !email — a raw URL query-param check, not a service-state read
>   Verdict: ❌ Not safe
> 
>   The distinction that matters: every other gate is a call to a Noop read method (getOpenOrgInvite()) that's hardcoded to return null unconditionally — it never inspects URL params, so no crafted URL can flip it. unsealOpenOrgInvite's gate is different in kind: it checks whether the component's own query params are populated, and query params are exactly what an attacker/user controls when hand-constructing a chrome-extension://.../#/finish-signup?email=...&sealedOpenOrgInviteData=... URL. That's why this one is genuinely reachable while the other six are not, even though all seven "look" gated at a glance.
> 
>   The two downstream calls chained after unsealOpenOrgInvite (getOpenOrgInviteStatus, setOrganizationInvite at lines 266/295 in the same component) are fine only because the unseal call throws (or, post-fix, returns secret-miss) before the code ever reaches them — they have no independent protection of their own, they just never get a chance to run. This confirms the earlier finding is the only gap, and the recommended fix (unsealOpenOrgInvite → return { kind: "secret-miss" } instead of throwing) is both necessary and sufficient — no other method needs the same treatment.
> </details>

rr-bw (approved):
Approved with one non-blocking comment.

libs/common/src/auth/organization-invite/services/implementations/noop-organization-invite.service.ts
JaredSnider-Bitwarden: Claude + me:
Thank you for calling this out for discussion. 

Follow-up after further investigation — the "regression" as described requires unsupported flows on both clients which merits the existing throw:

**Desktop (Electron)** — Not reachable: users can't enter URLs.

**Browser Extension** — Technically possible in Chrome, but only via a self-inflicted, unsupported path.
- Chrome will still let a user *manually type* `chrome-extension://<their-id>/popup/index.html#/finish-signup?email=…&sealedOpenOrgInviteData=…` into the address bar. That's the only physical route to the regression.
- To reach it, a user would have to: (i) look up their own extension ID in `chrome://extensions`, (ii) know the internal Angular route, (iii) obtain a valid `sealedOpenOrgInviteData` blob — which only exists in the *web* client's per-client `disk-local` storage — and (iv) hand-craft the URL themselves. That's not a supported flow.

Given that, I'm comfortable leaving `unsealOpenOrgInvite` in the throw group. It's genuinely unsupported.

PR #22498, opened 2026-08-17
[PM-38906] Decrypt collections in a single SDK call

JaredScar:
## 🎟️ Tracking

## 📔 Objective

## Performane Improvements
image

### DIFF
--- a/libs/admin-console/src/common/collections/abstractions/collection-encryption.service.ts
+++ b/libs/admin-console/src/common/collections/abstractions/collection-encryption.service.ts
@@ -21,8 +21,9 @@ export abstract class CollectionEncryptionService {
   /**
    * Decrypts many collections using the SDK for the given userId.
    *
-   * Collections that fail to decrypt are logged and dropped from the result rather than
-   * aborting the rest of the batch.
+   * Collections that fail to decrypt do not abort the rest of the batch. Instead they are
+   * included in the result with `decryptionFailure` set and decryption-dependent fields (e.g.
+   * `name`) left empty, so the item is still shown to the user rather than silently dropped.
    *
    * @param collections The encrypted collection objects
    * @param userId The user ID whose keys will be used for decryption
--- a/libs/admin-console/src/common/collections/abstractions/collection-encryption.service.ts
+++ b/libs/admin-console/src/common/collections/abstractions/collection-encryption.service.ts
@@ -1,3 +1,5 @@
+import { Observable } from "rxjs";
+
 import { Collection } from "@bitwarden/common/admin-console/models/collections/collection";
 import { CollectionView } from "@bitwarden/common/admin-console/models/collections/collection.view";
 import { UserId } from "@bitwarden/common/types/guid";
@@ -12,17 +14,20 @@ export abstract class CollectionEncryptionService {
    * @param collection The encrypted collection object
    * @param userId The user ID whose keys will be used for decryption
    *
-   * @returns A promise that resolves to the decrypted collection view
+   * @returns An observable that emits the decrypted collection view
    */
-  abstract decrypt(collection: Collection, userId: UserId): Promise<CollectionView>;
+  abstract decrypt(collection: Collection, userId: UserId): Observable<CollectionView>;
 
   /**
    * Decrypts many collections using the SDK for the given userId.
    *
+   * Collections that fail to decrypt are logged and dropped from the result rather than
+   * aborting the rest of the batch.
+   *
    * @param collections The encrypted collection objects
    * @param userId The user ID whose keys will be used for decryption
    *
-   * @returns A promise that resolves to an array of decrypted collection views
+   * @returns An observable that emits an array of decrypted collection views
    */
-  abstract decryptMany(collections: Collection[], userId: UserId): Promise<CollectionView[]>;
+  abstract decryptMany(collections: Collection[], userId: UserId): Observable<CollectionView[]>;
 }
--- a/libs/admin-console/src/common/collections/services/default-collection-encryption.service.ts
+++ b/libs/admin-console/src/common/collections/services/default-collection-encryption.service.ts
@@ -1,10 +1,23 @@
-import { catchError, concatMap, firstValueFrom } from "rxjs";
+import {
+  Observable,
+  catchError,
+  concatMap,
+  distinctUntilChanged,
+  map,
+  of,
+  switchMap,
+  tap,
+  throwError,
+} from "rxjs";
 
 import { Collection } from "@bitwarden/common/admin-console/models/collections/collection";
 import { CollectionView } from "@bitwarden/common/admin-console/models/collections/collection.view";
+import { FeatureFlag } from "@bitwarden/common/enums/feature-flag.enum";
+import { ConfigService } from "@bitwarden/common/platform/abstractions/config/config.service";
 import { LogService } from "@bitwarden/common/platform/abstractions/log.service";
 import { SdkService } from "@bitwarden/common/platform/abstractions/sdk/sdk.service";
 import { UserId } from "@bitwarden/common/types/guid";
+import { DecryptCollectionListResult } from "@bitwarden/sdk-internal";
 
 import { CollectionEncryptionService } from "../abstractions/collection-encryption.service";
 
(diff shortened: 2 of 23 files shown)


libs/common/src/enums/feature-flag.enum.ts
```diff
@@ -12,6 +12,8 @@ import { ServerConfig } from "../platform/abstractions/config/server-config";
 export enum FeatureFlag {
   /* Admin Console Team */
   GenerateInviteLink = "pm-32497-generate-invite-link",
+  CollectionBulkDecryptWithFailures = "collection-bulk-decrypt-with-failures",
+  PolicyDrawers = "pm-34804-policy-drawers",
```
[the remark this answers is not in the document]
JaredScar: You're right... This was removed in a recent PR after this PR was made and still had it in the works of implementation

libs/admin-console/src/common/collections/services/default-collection-encryption.service.ts
```diff
@@ -1,59 +1,211 @@
...
+   */
+  private decryptManyWithFailuresV2(
+    collections: Collection[],
+    userId: UserId,
+  ): Observable<CollectionDecryptionResult> {
+    const startTime = performance.now();
+
+    return this.sdkService.userClient$(userId).pipe(
+      concatMap(async (sdk) => {
+        if (!sdk) {
```
quexten: Just so you know this check is no longer needed. SDK can never be null.

libs/common/src/enums/feature-flag.enum.ts
```diff
@@ -12,6 +12,7 @@ import { ServerConfig } from "../platform/abstractions/config/server-config";
 export enum FeatureFlag {
   /* Admin Console Team */
   GenerateInviteLink = "pm-32497-generate-invite-link",
+  CollectionBulkDecryptWithFailures = "collection-bulk-decrypt-with-failures",
```
eliykat, line 15: Will need updating once the feature flag key is updated on the server. (See my review on your server PR.)

libs/admin-console/src/common/collections/abstractions/collection-encryption.service.ts
```diff
@@ -12,17 +24,36 @@ export abstract class CollectionEncryptionService {
...
+   * @param collections The encrypted collection objects
+   * @param userId The user ID whose keys will be used for decryption
+   *
+   * @returns An observable that emits a {@link CollectionDecryptionResult}
    */
-  abstract decryptMany(collections: Collection[], userId: UserId): Promise<CollectionView[]>;
+  abstract decryptManyWithFailures(
+    collections: Collection[],
+    userId: UserId,
+  ): Observable<CollectionDecryptionResult>;
```
eliykat: This says that implementations can use the feature flag to decide which bulk decrypt method to call, but the feature flag switch is actually interior/private to this service. I think that's right, because the caller doesn't need to know what's being used to decrypt the collections, only that they are decrypted.

Suggestion: delete `decryptManyWithFailures` as a separate public interface. Rely on the feature flag switch inside `decryptMany` as you do today, returning only `Observable<CollectionView[]>`. We are not actually doing anything with the `failures` value, so there's no need to add the separate interface. The goal is to use the bulk method, not to expose failures differently. 

That said, please also add logging for the new path so that we log which collection IDs failed to decrypt.

EDIT: related to the above - how _do_ we handle decryption failures today? Do collections show up with `[error: failed to decrypt]` like they used to? Or did moving to the SDK mean we log + drop them?

libs/admin-console/src/common/collections/services/default-collection-encryption.service.ts
```diff
@@ -1,59 +1,180 @@
...
+        const sdkCollections = collections.map((c) => c.toSdkCollection());
+        const success: CollectionView[] = [];
+        const failure: Collection[] = [];
+
+        try {
+          // Fast path: SDK decrypts the entire list in one call. Results are returned in
+          // the same order as the input, so index-based re-association is safe.
+          const sdkViews = ref.value.vault().collections().decrypt_list(sdkCollections);
+          for (let i = 0; i < sdkViews.length; i++) {
+            success.push(CollectionView.fromSdkCollectionView(sdkViews[i], collections[i]));
```
eliykat: This is fragile. A better way to do it is to create a hashmap of collections by ID, then just look up the corresponding collection ID. (I'd like to change how `fromSdkCollectionView` works so we don't need to do this at all, but out of scope here.)

libs/admin-console/src/common/collections/services/default-collection-encryption.service.ts
```diff
@@ -1,59 +1,180 @@
...
+          }
+        } catch {
+          // Batch call failed (e.g. one key is missing); fall back to per-item so that
+          // only the affected collections are reported as failures.
           for (const collection of collections) {
             try {
               const sdkView = ref.value.vault().collections().decrypt(collection.toSdkCollection());
-              views.push(CollectionView.fromSdkCollectionView(sdkView, collection));
```
eliykat: This is exactly what we're trying to move away from and I don't think falling back to it solves anything.

Suggestion: remove the catch block; let `catchError` handle any batch level failures below.

libs/admin-console/src/common/collections/services/default-collection-encryption.service.ts
```diff
@@ -1,59 +1,180 @@
...
-          if (!sdk) {
-            throw new Error("SDK not available");
+    return this.configService.getFeatureFlag$(FeatureFlag.CollectionBulkDecryptWithFailures).pipe(
+      distinctUntilChanged(),
+      switchMap((bulkDecryptEnabled) =>
+        bulkDecryptEnabled
+          ? this.decryptManyWithFailuresV2(collections, userId)
+          : this.decryptManyWithFailuresV1(collections, userId),
+      ),
+    );
```
eliykat, line 56: Per my previous comment: this would all fold back into `decryptMany`.

eliykat (changes requested):

eliykat:
Please also update the PR title to reflect the description (this isn't about using observables more).

libs/admin-console/src/common/collections/abstractions/collection-encryption.service.ts
JaredScar: Removed `decryptManyWithFailures` from the public abstraction. Good callout there. The feature flag just lives inside of `decryptMany`

Today: Failures would just be logged and silently dropped. Failed collections don't appear in the result set rather than showing up as error entries, so I've now preserved that behavior.

libs/admin-console/src/common/collections/services/default-collection-encryption.service.ts
```diff
@@ -1,59 +1,158 @@
...
+   */
+  private decryptManyV2(collections: Collection[], userId: UserId): Observable<CollectionView[]> {
+    const startTime = performance.now();
+
+    return this.sdkService.userClient$(userId).pipe(
+      concatMap(async (sdk) => {
+        using ref = sdk.take();
+
+        const collectionMap = new Map<string, Collection>(
+          collections.map((c) => [c.id as unknown as string, c]),
```
BTreston: Is `as unknown as string` really needed?

libs/admin-console/src/common/collections/services/default-collection-encryption.service.ts
```diff
@@ -1,59 +1,158 @@
...
-          for (const collection of collections) {
-            try {
-              const sdkView = ref.value.vault().collections().decrypt(collection.toSdkCollection());
-              views.push(CollectionView.fromSdkCollectionView(sdkView, collection));
-            } catch (error: unknown) {
-              this.logService.error(`Failed to decrypt collection ${collection.id}: ${error}`);
-            }
+        const success: CollectionView[] = [];
+        for (const sdkView of result.successes) {
+          const id = sdkView.id as unknown as string;
```
BTreston: same question as above?

BTreston (commented):
Do we have any comparison performance data for this? We should add some samples to the PR description to give an idea of the expected uplift

quexten (dismissed):

BTreston (dismissed):

eliykat (commented):
As discussed, my previous feedback was incorrect: we _should_ surface these failures to the user, using the current pattern that Vault uses. Today we drop failures which is functionally invisible to the user (unless they're checking console logs). Holding off on further reviews until we align with Product next week. Assuming they agree, this PR will need to add back the separate public interface, and update calling locations to handle failed decryptions using the dialog. (May need some discussion with Vault.)

JaredScar:
Closing and getting rid of stacked PRs

PR #22499, opened 2026-08-17
[PM-38906] Decrypt Admin Console collections in a single SDK call, preserving decryption failures

JaredScar:
## 🎟️ Tracking

## 📔 Objective

## Performane Improvements
image

### DIFF
--- a/libs/admin-console/src/common/collections/services/default-collection-admin.service.spec.ts
+++ b/libs/admin-console/src/common/collections/services/default-collection-admin.service.spec.ts
@@ -200,10 +200,7 @@ describe("DefaultCollectionAdminService", () => {
       });
 
       it("still decrypts responses that carry no groups/users", async () => {
-        // `ApiService` builds every element as a `CollectionAccessDetailsResponse`, whose
-        // `groups`/`users` default to `[]`, so this shape does not occur on the current call
-        // path. The cast below forces the plain `CollectionResponse` class to cover the
-        // defensive `isCollectionAccessDetailsResponse` branch in `decryptManyV1`.
+        // Forces the plain CollectionResponse shape, which doesn't occur in practice today.
         apiService.getManyCollectionsWithAccessDetails.mockResolvedValue(
           new ListResponse(
             {
@@ -226,8 +223,6 @@ describe("DefaultCollectionAdminService", () => {
 
         const result = await firstValueFrom(service.collectionAdminViews$(orgId, userId));
 
-        // Both constructors yield the same name and an empty `groups`, so assert the branch
-        // taken rather than the resulting shape.
         expect(fromResponse).toHaveBeenCalledTimes(1);
         expect(fromAccessDetails).not.toHaveBeenCalled();
         expect(collectionEncryptionService.decryptManyWithFailures).not.toHaveBeenCalled();
--- a/libs/admin-console/src/common/collections/services/default-collection-admin.service.spec.ts
+++ b/libs/admin-console/src/common/collections/services/default-collection-admin.service.spec.ts
@@ -13,12 +13,13 @@ import {
 } from "@bitwarden/common/admin-console/models/collections/collection.response";
 import { CollectionView } from "@bitwarden/common/admin-console/models/collections/collection.view";
 import { FeatureFlag } from "@bitwarden/common/enums/feature-flag.enum";
-import { EncryptService } from "@bitwarden/common/key-management/crypto/abstractions/encrypt.service";
 import { ListResponse } from "@bitwarden/common/models/response/list.response";
 import { ConfigService } from "@bitwarden/common/platform/abstractions/config/config.service";
 import { LogService } from "@bitwarden/common/platform/abstractions/log.service";
 import { OrganizationId, UserId } from "@bitwarden/common/types/guid";
 import { KeyService } from "@bitwarden/key-management";
+// eslint-disable-next-line no-restricted-imports
+import { EncryptService } from "@bitwarden/legacy-crypto";
 
 import { CollectionService } from "../abstractions";
 import {
--- a/libs/admin-console/src/common/collections/services/default-collection-admin.service.spec.ts
+++ b/libs/admin-console/src/common/collections/services/default-collection-admin.service.spec.ts
@@ -1,0 +1,233 @@
+import { mock, MockProxy } from "jest-mock-extended";
+import { firstValueFrom, of } from "rxjs";
+
+import { ApiService } from "@bitwarden/common/abstractions/api.service";
+import { OrganizationService } from "@bitwarden/common/admin-console/abstractions/organization/organization.service.abstraction";
+import {
+  Collection,
+  CollectionTypes,
+} from "@bitwarden/common/admin-console/models/collections/collection";
+import {
+  CollectionAccessDetailsResponse,
+  CollectionResponse,
+} from "@bitwarden/common/admin-console/models/collections/collection.response";
+import { CollectionView } from "@bitwarden/common/admin-console/models/collections/collection.view";
+import { FeatureFlag } from "@bitwarden/common/enums/feature-flag.enum";
+import { EncryptService } from "@bitwarden/common/key-management/crypto/abstractions/encrypt.service";
+import { ListResponse } from "@bitwarden/common/models/response/list.response";
+import { ConfigService } from "@bitwarden/common/platform/abstractions/config/config.service";
+import { LogService } from "@bitwarden/common/platform/abstractions/log.service";
+import { OrganizationId, UserId } from "@bitwarden/common/types/guid";
+import { KeyService } from "@bitwarden/key-management";
+
+import { CollectionService } from "../abstractions";
+import {
+  CollectionDecryptionResult,
+  CollectionEncryptionService,
+} from "../abstractions/collection-encryption.service";
+
+import { DefaultCollectionAdminService } from "./default-collection-admin.service";
+
+const userId = "59fbbb44-8cc8-4279-ab40-afc5f68704f4" as UserId;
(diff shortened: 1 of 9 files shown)


r-tome (dismissed):

libs/admin-console/src/common/collections/services/default-collection-admin.service.ts
```diff
@@ -126,29 +147,153 @@ export class DefaultCollectionAdminService implements CollectionAdminService {
...
+    }
+
+    return this.decryptManyV1(organizationId, collections, orgKeys);
+  }
+
+  /**
+   * V1 implementation: decrypts all collection names concurrently via `EncryptService`.
+   * Collections that fail to decrypt are shown with a placeholder name rather than being dropped.
+   */
+  private decryptManyV1(
```
eliykat: Revert the current implementation. Rewriting it creates unnecessary work and means that you've got changes outside of the feature flag.

libs/admin-console/src/common/collections/services/default-collection-admin.service.ts
```diff
@@ -126,29 +147,153 @@ export class DefaultCollectionAdminService implements CollectionAdminService {
...
+            ],
+          );
+        }),
+        map(({ successes, failures }) => [...successes, ...failures]),
+        catchError((e: unknown) => {
+          this.logService.error("[DefaultCollectionAdminService] Batch decryption failed", e);
+          return of(
+            collections.map((c) =>
+              CollectionAdminView.fromCollectionAccessDetailsDecryptionFailure(c),
+            ),
```
eliykat, line 207: I think this is handled by `collectionEncryptionService` - are there any extra errors we expect to catch here?

libs/admin-console/src/common/collections/services/default-collection-admin.service.ts
```diff
@@ -126,29 +147,153 @@ export class DefaultCollectionAdminService implements CollectionAdminService {
...
+            "Admin Console",
+            "DefaultCollectionAdminService",
+            "decryptMany (v2, via CollectionEncryptionService)",
+            [
+              ["Items", collections.length],
+              ["Successes", successes.length],
+              ["Failures", failures.length],
+            ],
+          );
+        }),
```
eliykat: The encryption service already handles this - can be removed.

libs/admin-console/src/common/collections/services/default-collection-admin.service.ts
```diff
@@ -126,29 +147,153 @@ export class DefaultCollectionAdminService implements CollectionAdminService {
...
     orgKeys: Record<OrganizationId, OrgKey>,
-  ): Promise<CollectionAdminView[]> {
-    const promises = collections.map(async (c) => {
-      if (isCollectionAccessDetailsResponse(c)) {
-        return CollectionAdminView.fromCollectionAccessDetails(
-          c,
-          this.encryptService,
-          orgKeys[organizationId as OrganizationId],
+  ): Observable<CollectionAdminView[]> {
+    if (collections.length > 0 && collections.every(isCollectionAccessDetailsResponse)) {
```
eliykat: Is this ever called with anything other than `CollectionAccessDetailsResponse[]`? Just check current uses and see if this is necessary or whether we can align this. Otherwise we always need v1.

eliykat (commented):
There are repeated changes from the first PR in the stack (e.g. `default-collection-encryption.service.ts`) - please check and make sure it's stacked correctly. That said, I can't comment on the file because Github gives me an error, so it maybe some stacked PRs weirdness. Please see what you can do.

EDIT: please also update the PR title to be slightly more detailed

libs/admin-console/src/common/collections/services/default-collection-admin.service.ts
JaredScar: The EncryptionService's `catchError` actually rethrows `throwError()` and does not recover... Batch-level failures will therefore still propagate unless we catch it here. Unless we want to propagate the errors to the subscriber instead?

jprusik:
Dropped Autofill-dev team as we no longer have code ownership in the changes; feel free to re-add me if you specifically want our review.

libs/admin-console/src/common/collections/services/default-collection-admin.service.spec.ts
```diff
@@ -0,0 +1,233 @@
...
+  Collection,
+  CollectionTypes,
+} from "@bitwarden/common/admin-console/models/collections/collection";
+import {
+  CollectionAccessDetailsResponse,
+  CollectionResponse,
+} from "@bitwarden/common/admin-console/models/collections/collection.response";
+import { CollectionView } from "@bitwarden/common/admin-console/models/collections/collection.view";
+import { FeatureFlag } from "@bitwarden/common/enums/feature-flag.enum";
+import { EncryptService } from "@bitwarden/common/key-management/crypto/abstractions/encrypt.service";
```
quexten: If you rebase the git stack this should be updated to "@bitwarden/legacy-crypto"

libs/admin-console/src/common/collections/services/default-collection-admin.service.ts
```diff
@@ -126,29 +145,106 @@ export class DefaultCollectionAdminService implements CollectionAdminService {
...
-    const promises = collections.map(async (c) => {
-      if (isCollectionAccessDetailsResponse(c)) {
-        return CollectionAdminView.fromCollectionAccessDetails(
-          c,
-          this.encryptService,
-          orgKeys[organizationId as OrganizationId],
-        );
-      }
+    return Promise.all(
+      collections.map((c) =>
```
quexten: Why `Promise.all`? Promise.all generally is slower than a loop on both v8 and spidermonkey for compute heavy operations such as this one. A simple await loop will be faster.

libs/admin-console/src/common/collections/services/default-collection-admin.service.ts
```diff
@@ -126,29 +145,106 @@ export class DefaultCollectionAdminService implements CollectionAdminService {
...
   }
 
-  private async decryptMany(
+  private decryptMany(
+    organizationId: string,
+    userId: UserId,
+    collections: CollectionResponse[] | CollectionAccessDetailsResponse[],
+    orgKeys: Record<OrganizationId, OrgKey>,
+  ): Observable<CollectionAdminView[]> {
+    if (collections.length > 0 && collections.every(isCollectionAccessDetailsResponse)) {
```
quexten: Is there a plan to drop the V1? This code doesn't cleanly map between feature flagged / non feature flagged but basically does:

- Case A: Length > 0 and every isCollectionAccessResonse
  Case A.1: Feature Flag On
    -> SDK
  Case A.2: Feature Flag Off
    -> Non SDK
- Case B: Length = 0 *OR* at least one not isCollectionAccessResponse
    -> Non SDK

^ So Case B is not handled here. Is there a plan to move this to the SDK directly?

quexten (commented):
Since I reviewed the root pr of the stack, reviewing this too. Looks good.

libs/common/src/admin-console/models/collections/collection-admin.view.ts
```diff
@@ -151,6 +151,70 @@ export class CollectionAdminView extends CollectionView {
...
+   * Creates a placeholder CollectionAdminView for a collection that failed to decrypt via the
+   * SDK bulk path (`decrypt_list_with_failures`). Unlike the personal-vault decryption path, the
+   * Admin Console must never silently drop a collection that fails to decrypt, since admins still
+   * need to see, manage, and delete it. Mirrors the fallback behavior of
+   * {@link fromCollectionAccessDetails}'s catch branch.
+   */
+  static fromCollectionAccessDetailsDecryptionFailure(
+    source: CollectionAccessDetailsResponse,
+  ): CollectionAdminView {
+    const view = new CollectionAdminView({ ...source, name: DECRYPT_ERROR });
```
BTreston, line 188: is `DECRYPT_ERROR`s message being i18n translated when shown in the UI? If not, do we want to be able to do that?

libs/admin-console/src/common/collections/services/default-collection-admin.service.ts
JaredScar: Yes, the idea would be to drop V1! It only exists for when the feature flag is off. I've now handled case B, good catch! 🎉

quexten (dismissed):

BTreston (changes requested):
🎨  Most of the comments in code are very verbose Claude speak. We really shouldn't be adding paragraph long comments for some of these, the meaning is lost in the jargon. Aside from that, one minor question (would probably be a followup so non-blocking).

libs/common/src/admin-console/models/collections/collection-admin.view.ts
JaredScar: It's used in multiple places across the codebase and by multiple teams. It's a good idea for sure, but I'm not sure we want to introduce all the teams to come review this for such a change 😅

BTreston: We wouldn't touch the actual message. We'd just i18n pipe in our owned component where this shows up, no external reviewers or impact. Only cross-team would be asking product what to show. Neither here nor there though for this PR. Just a passing thought, really.

quexten: The better way to do this is to not map to [DECRYPT_ERROR] but instead to put a "failed to decrypt" flag (boolean) onto your domain object. Then, you can actually define the way to render decryption failures, with an i18n'd flag (and maybe a warning icon, and disabled state).

eliykat (commented):
We'll want to hold on this until we've figured out error handling in the base PR. Whatever we do there we'll replicate here, so we should wait on that.

JaredScar:
Closing and getting rid of stacked PRs

PR #22501, opened 2026-08-17
[PM-40318] VFO1 Web vault query param handling

shane-melton:
## 🎟️ Tracking

[PM-40318](https://bitwarden.atlassian.net/browse/PM-40318)

## 📔 Objective

Wire up the query parameters to the Web vault table and add support for legacy query params via the legacy vault redirect guard added in https://github.com/bitwarden/clients/pull/22438.

[PM-40318]: https://bitwarden.atlassian.net/browse/PM-40318?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/components/src/filter-menu/filter-menu.component.ts
shane-melton, line 1: ℹ️ This change was necessary to stop the FilterMenu from throwing an NG0950 error due to the `option.value()` signal being undefined for filter menu options that are not available on first render. e.g. the shared folders (collections) are loaded separately from ciphers and appear after first table render.

libs/components/src/table/v2/table-v2.component.ts
```diff
@@ -448,7 +449,10 @@ export class BitTableV2Component<T = unknown, S extends string = never, F = Reco
       if (!this.urlRestored()) {
         return;
       }
-      const state: ParamState = {};
+      // Seed from the store's current value rather than starting empty, so a param
+      // whose chip hasn't registered yet (e.g. one gated behind an `@if` that's still
+      // waiting on data) survives this write instead of being dropped.
+      const state: ParamState = { ...untracked(this.urlStore) };
```
shane-melton, line 455: ℹ️ Some Vault filters are not always rendered immediately (e.g. shared folders) and if a URL has a query param for the shared folders key, the value would be stripped on first write since the control wasn't registered yet on first render.  This preserves the query param instead of clearing it with an empty starting state.

BryanCunningham (dismissed):

libs/vault/src/components/vault-items-table/vault-items-table.component.html
```diff
@@ -7,6 +7,7 @@
   [loading]="loading()"
   [selection]="selection"
   [virtualRowHeight]="56"
+  [queryParam]="filterNamespace"
```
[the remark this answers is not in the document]
shane-melton: Fixed [08ff7ca](https://github.com/bitwarden/clients/pull/22501/commits/08ff7ca32c14cdd223008bacfbe3da5a14e91dbe)

libs/vault/src/routing/vault-filter-legacy-redirect.guard.ts
```diff
@@ -0,0 +1,143 @@
...
+  if (legacy.type != null && !typeTranslated) {
+    keysToStrip.delete("type");
+  }
+
+  const queryParams: Record<string, string> = {};
+  for (const key of route.queryParamMap.keys) {
+    if (!keysToStrip.has(key)) {
+      queryParams[key] = route.queryParamMap.get(key)!;
+    }
+  }
```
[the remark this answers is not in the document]
shane-melton: Fixed [7151267](https://github.com/bitwarden/clients/pull/22501/commits/71512671f10b28b53ceed6ce542b1184dba16687)

libs/components/src/utils/query-param-store.ts
```diff
@@ -161,6 +161,7 @@ export function queryParamStore<T extends ParamState>(
         queryParams: patch,
         queryParamsHandling: "merge",
         replaceUrl: true,
+        state: { focusAfterNav: false },
```
shane-melton, line 164: ℹ️ Without this, every keystroke in the bit-search input would steal focus to the main content of the page due to the `RouterFocusManagerService`.

libs/components/src/table/v2/table-v2.component.ts
```diff
@@ -388,7 +389,7 @@ export class BitTableV2Component<T = unknown, S extends string = never, F = Reco
         key: signal(SEARCH_FILTER_KEY),
         value: search.value,
         active: computed(() => (search.value() ?? "") !== ""),
-        setValue: (value) => search.writeValue((value as string) ?? ""),
+        setValue: (value) => search.writeValue(value == null ? "" : String(value)),
```
shane-melton, line 392: ℹ️ A number only search text term ("2024") would be silently cast to a number, which would break downstream consumers of the search value which always expect a string type (e.g. `SearchService.isSearchable()`)

nick-livefront (approved):

BryanCunningham (approved):

PR #22521, opened 2026-08-18
[PM-35370] Add temporary master password single input for recover account modals

JaredScar:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-35370

## 📔 Objective

Changes the recover account modal to use 1 singular master password input field rather than 2 confirm master password input fields.

## 📸 Screenshots

image
image
image


libs/auth/src/angular/input-password/input-password.stories.ts
```diff
@@ -229,6 +230,17 @@ export const ChangePasswordDelegation: Story = {
...
+  render: (args) => ({
+    props: args,
+    template: `
+      <auth-input-password [flow]="InputPasswordFlow.AccountRecovery"></auth-input-password>
+      <br />
+      <div>Note: single "Temporary master password" field — no confirm field. Used by the Account Recovery dialog. No buttons; the dialog owns Save/Cancel.</div>
+    `,
+  }),
+};
+
```
JaredSnider-Bitwarden, line 243: :art: Probably worth adding a MP policy options support example as well:
```ts
export const AccountRecovery: Story = {
  render: (args) => ({
    props: args,
    template: `
      <auth-input-password
        [flow]="InputPasswordFlow.AccountRecovery"
        [masterPasswordPolicyOptions]="masterPasswordPolicyOptions"
      ></auth-input-password>
      <br />
      <div>Note: single "Temporary master password" field — no confirm field. Used by the Account Recovery dialog. No buttons; the dialog owns Save/Cancel.</div>
    `,
  }),
};

export const AccountRecoveryNoPolicy: Story = {
  render: (args) => ({
    props: args,
    template: `
      <auth-input-password [flow]="InputPasswordFlow.AccountRecovery"></auth-input-password>
    `,
  }),
};
```

libs/auth/src/angular/input-password/input-password.mdx
```diff
@@ -143,8 +143,15 @@ whether or not the user key should be rotated.<br /><br />
...
-- Account Recovery<br /><br />
+- Emergency Access Takeover<br /><br />
+
+**`AccountRecovery`**
+
+Used when an org admin resets a member's master password via the Account Recovery dialog. Like
+`ChangePasswordDelegation`, no `userId` or `email` is passed down and the dialog owns the
+Save/Cancel buttons. Unlike `ChangePasswordDelegation`, there is only **one** password field (no
+confirm field) and the label reads "Temporary master password". When a policy is in effect, the
+requirements are listed inline beneath the field rather than in the banner callout.<br /><br />
```
JaredSnider-Bitwarden, line 154: 🎨:  This has several specific details which claude loves to generate which are so specific they can rot easily and become a maintenance issue. 

Here's a suggested rewrite:
```md
**`AccountRecovery`**

Used when an org admin resets a member's master password via the Account Recovery dialog.
Like `ChangePasswordDelegation`, the dialog owns its own action buttons, so this flow renders
no buttons of its own. Unlike `ChangePasswordDelegation`, it renders a **single** password field
(no confirm field). When a master password policy is in effect, the requirements render inline
beneath the field.<br /><br />
```

More wordy Claude explanation:

<details>
<summary>🎨 Suggestion — the current paragraph pins several things that can rot silently</summary>

1. `"Temporary master password"` — quotes the i18n copy verbatim; drifts if `temporaryMasterPass` is ever rewritten.
2. `Save/Cancel` — quotes specific button labels; same rot shape as (1).
3. `no userId or email is passed down` — freezes the current `@Input` shape into prose; wrong the moment a future variant needs one.
4. `banner callout` — anchors to a sibling UI element by visual name; rots if that element is renamed or restyled.

The rewrite above keeps the load-bearing behavioral distinction (single field, dialog owns buttons, inline requirements when policy exists) and drops the pieces that describe UX chrome Storybook already shows.

</details>

jrmccannon (approved):
Looks good from AC's end.

libs/auth/src/angular/input-password/input-password.component.html
```diff
@@ -1,78 +1,121 @@
...
+          (click)="generatePassword()"
+        ></button>
+        <button
+          type="button"
+          bitSuffix
+          bitIconButton="bwi-clone"
+          label="{{ 'copyPassword' | i18n }}"
+          (click)="copy()"
+        ></button>
+      }
```
JaredSnider-Bitwarden, line 69: ⚠️ :
The [Claude comment here](https://github.com/bitwarden/clients/pull/22521/changes#r3806466774) actually broke the rendering of the generate and copy buttons. Please re-check UI changes before opening for review if you change the HTML as this would have been caught.

Claude explanation:
It merged the two separate @if blocks (one per button) into a single @if containing both <button bitSuffix> elements. Angular cannot project multiple root nodes from a single @if block, so the buttons fell out of the form field.

Image

libs/auth/src/angular/input-password/input-password.component.html
```diff
@@ -1,78 +1,121 @@
...
         [(toggled)]="showPassword"
       ></button>
-      <bit-hint *ngIf="flow !== InputPasswordFlow.ChangePasswordDelegation">
-        <span class="tw-font-medium">{{ "important" | i18n }} </span>
-        {{ "masterPassImportant" | i18n }}
-        {{ minPasswordLengthMsg }}.
-      </bit-hint>
+      @if (
+        flow !== InputPasswordFlow.ChangePasswordDelegation &&
+        flow !== InputPasswordFlow.AccountRecovery
```
JaredSnider-Bitwarden, line 79: ⚠️ : There are ~5 guards that use exclusive checks like this. If we add another flow, we have to remember to `&& flow !== InputPasswordFlow.<NewFlow>` in all those places vs just using an inclusive pattern. We have been bitten by this before too many times to accept this - please swap to inclusive checks.

libs/auth/src/angular/input-password/input-password.component.html
JaredSnider-Bitwarden, line 1: ⚠️ : The point of this component is to centralize the UI for all password flows. Introducing what is essentially a duplicate of the `auth-password-callout` to put into the `bit-hint` is suboptimal. We should discuss this with design to get agreement to make this change in how we display MP policy requirements across the board vs just in this one flow.  

This image is modified to show both the existing `auth-password-callout` and the new `bit-hint` which displays the same content. 

Image

JaredSnider-Bitwarden, line 1: ⚠️ :

There is incongruity in the Figma vs the screenshots on the PR vs what I'm seeing running locally. 

Figma (note the password strength bar in the `bit-hint` directly below the password form control)
Image


Screenshot on the PR (shows password strength bar matching figma):

Image

When I ran it locally (password strength bar below `bit-hint`):

Image

I have a design concern here about if the `bit-hint` is even meant to support adding custom components like the `tools-password-strength` to it or if we will have display issues - probably a UIF question.

JaredSnider-Bitwarden (changes requested):
Thank you for attempting to fix this. However, we are going to request that Auth takes over this work and works with design to figure out how to move forward on this. Leaving these review comments for posterity and to kickstart discussions.

Patrick-Pimentel-Bitwarden:
Closing this pr for now in favor of having Auth own this body of work. We will use this branch to help highlight the concerns to design for how to navigate through making changes to this component. @JaredScar thank you for your work on this and how we might need to make the designers aware of how this component works. No effort is wasted.

PR #22283, opened 2026-08-07
[PM-40313] Restructure the Password Manager side navigation (web)

gbubemismith:
## 🎟️ Tracking
https://bitwarden.atlassian.net/browse/PM-40313

## 📔 Objective
Restructure the web Password Manager side navigation 

## 📸 Screenshots
| View | Screenshot |
| --- | --- |
| **Vfo1 Sidenav**<br>Vault has personal vault, free org and enterprise organizations | Vfo1 Sidenav |
| **Vaults expanded** | Vaults expanded |
| **Product switcher** | Product switcher |
| **Free user, my vault only** | Free user, my vault only |
| **Centralized org ownership** | Centralized org ownership |

https://github.com/user-attachments/assets/7acbaf2a-d285-4003-8dd2-b6d1aba34809

PR #22533, opened 2026-08-19
[PM-37378] VFO1 persist active filters

shane-melton:
## 🎟️ Tracking
[PM-37378](https://bitwarden.atlassian.net/browse/PM-37378)

## 📔 Objective

Add support for persisting filter query parameters scoped to each vault in the future side nav (All Items, My Vault, and each organization vault).

## 📸 Screenshots

https://github.com/user-attachments/assets/8c5c7a72-2b7a-453f-bde8-effcde8263df

[PM-37378]: https://bitwarden.atlassian.net/browse/PM-37378?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/web/src/app/vault/individual-vault/vault-routing.module.ts
+++ b/apps/web/src/app/vault/individual-vault/vault-routing.module.ts
@@ -64,8 +64,11 @@ const routes: Routes = [
       vaultScopeGuard,
       vaultFilterRestoreGuard,
     ],
-    data: { ...MY_ITEMS_ROUTE_DATA, titleId: "vaults", vaultFilterScope: true } satisfies
-      RouteDataProperties & VaultScopeRouteData,
+    data: {
+      ...MY_ITEMS_ROUTE_DATA,
+      titleId: "vaults",
+      vaultFilterScope: true,
+    } satisfies RouteDataProperties & VaultScopeRouteData,
   },
   // An organization vault's shared folders.
   {
--- a/apps/web/src/app/vault/individual-vault/vault-routing.module.ts
+++ b/apps/web/src/app/vault/individual-vault/vault-routing.module.ts
@@ -10,8 +10,12 @@ import {
   organizationVaultGuard,
   SHARED_FOLDERS_ROUTE,
   vaultFilterLegacyRedirectGuard,
+  vaultFilterRestoreGuard,
   vaultScopeGuard,
+  type VaultScopeRouteData,
 } from "@bitwarden/vault";
+
+import { RouteDataProperties } from "../../core";
 
 import { SharedFoldersComponent } from "./shared-folders/shared-folders.component";
 import { VaultNextComponent } from "./vault-next.component";
@@ -25,7 +29,18 @@ const routes: Routes = [
     routeOptions: {
       path: "",
       canActivate: [vaultFilterLegacyRedirectGuard],
-      data: { titleId: "vaults" },
+      data: { titleId: "vaults" } satisfies RouteDataProperties,
+    },
+    // Filter memory only means anything to the VFO1 vault, so it hangs off the flagged route. That
+    // keeps the pre-VFO1 vault from recording filters it can't read back, and saves
+    // `vaultFilterRestoreGuard` from re-checking the flag itself.
+    flaggedRouteOptions: {
+      path: "",
+      data: { titleId: "vaults", vaultFilterScope: true } satisfies RouteDataProperties &
+        VaultScopeRouteData,
+      // Order matters: the legacy rewrite runs first, so a pre-namespace URL's own filters win
+      // over the remembered ones.
+      canActivate: [vaultFilterLegacyRedirectGuard, vaultFilterRestoreGuard],
     },
   }),
   {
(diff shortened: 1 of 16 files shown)


apps/web/src/app/layouts/user-layout.component.ts
```diff
@@ -49,6 +50,19 @@ export class UserLayoutComponent implements OnInit {
...
+  private readonly router = inject(Router);
+  private readonly vaultFilterMemory = inject(VaultFilterMemoryService);
+
+  /**
+   * Carries the filters the vault was last viewed with, so returning to it from elsewhere in the
+   * app restores them. A plain `/vault` tree until the user has filtered something.
+   */
+  protected readonly vaultRoute = computed(() =>
+    this.router.createUrlTree(["/vault"], {
+      queryParams: this.vaultFilterMemory.paramsFor(ALL_ITEMS_SCOPE),
```
shane-melton: ℹ️ This is a no-op without the VFO1 feature flag enabled. Which is required to render the tablev2 component and populate the filter memory service.

libs/vault/src/routing/vault-filter-memory.service.ts
```diff
@@ -0,0 +1,99 @@
...
+  }));
+
+  constructor() {
+    this.writes
+      .pipe(debounceTime(PERSIST_DEBOUNCE_INTERVAL), takeUntilDestroyed())
+      .subscribe(([scope, params]) => {
+        // Updating rather than replacing so an in-memory view that's missing a scope — because
+        // it was recorded before the read resolved — can't drop it from state.
+        void this.state.update((prev) => ({ ...prev, [scope]: params }));
+      });
```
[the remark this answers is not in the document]
shane-melton: Fixed [26dcb6b](https://github.com/bitwarden/clients/pull/22533/commits/26dcb6b10981f378cbd13f609eaabf80a2e82a8c)

apps/web/src/app/layouts/user-layout.component.ts
shane-melton: Actually, on second thought we can probably move this within the sidenav work already being done in #22283. cc: @gbubemismith

gbubemismith: I think it might be more intuitive to have that change here was the side nav is merged to main

libs/vault/src/routing/README.md
gbubemismith, line 1: 🙌🏼 I love this

libs/vault/src/routing/vault-filter-restore.guard.ts
```diff
@@ -0,0 +1,51 @@
...
+export const vaultFilterRestoreGuard: CanActivateFn = async (route, state) => {
+  const filterMemory = inject(VaultFilterMemoryService);
+
+  const scope = vaultScopeOf(state);
+  if (scope == null) {
+    return true;
+  }
+
+  // The URL states its own filters — including none, when they were just cleared.
+  if (hasFilterParams(state.root.queryParams)) {
```
gbubemismith, line 35: ❓ Should remembered filters apply when the user is looking at Trash/Archive?

apps/web/src/app/layouts/user-layout.component.ts
shane-melton: This was dropped in favor of using a `vaultFilterRestoreGuard` instead of building the sidenav URIs dynamically

libs/vault/src/routing/vault-filter-restore.guard.ts
shane-melton: That's a good point. Trash/Archive will likely have the same "scope" as All Items as it currently stands. We'd have to add additional scopes for Trash and Archive if we want them separate. Or add some special handling around those.

shane-melton:
This got a bit stale so I rebased onto `main`, which dropped the parent branch commits that was merged already. 

The filter memory now keys on the shared `VaultScope` model from `libs/vault/src/models` instead of a parallel string scope. `routing/vault-scope.ts` was renamed to `routing/vault-filter-scope.ts` to remove ambiguity. Lastly, the `vaultFilterRestoreGuard` is registered on the new `:vaultId` scoped routes added by #22567.

apps/web/src/app/vault/individual-vault/vault-routing.module.ts
```diff
@@ -25,7 +29,17 @@ const routes: Routes = [
...
+      data: { titleId: "vaults" } satisfies RouteDataProperties,
+    },
+    // Filter memory only means anything to the VFO1 vault, so it hangs off the flagged route. That
+    // keeps the pre-VFO1 vault from recording filters it can't read back.
+    flaggedRouteOptions: {
+      path: "",
+      data: { titleId: "vaults", vaultFilterScope: true } satisfies RouteDataProperties &
+        VaultScopeRouteData,
+      // Order matters: the legacy rewrite runs first, so a pre-namespace URL's own filters win
+      // over the remembered ones.
```
nick-livefront, line 41: 👏 very helpful comments here

libs/vault/src/routing/README.md
nick-livefront: Make that two of us!

libs/vault/src/routing/vault-filter-scope.spec.ts
nick-livefront, line 1: This is a cool set of tests 🚀

nick-livefront (dismissed):

shane-melton:
@gbubemismith @nick-livefront I had to tweak the Vault component to re-mount the shared table component on scope change. Otherwise, the filters leak between scopes because the table v2 component only reads from the URL once on mount. So when the scope restore guard re-writes the query params to the URL, the table would never pick them up and would instead write the existing filters to the new scope's memory. Details in [d2daf30](https://github.com/bitwarden/clients/pull/22533/commits/d2daf30142a1b5bb1eb5cd6287f6e6e302da61b5)

nick-livefront (dismissed):
I appreciate the explanation!

dani-garcia (dismissed):

nick-livefront (approved):

dani-garcia (approved):

PR #22535, opened 2026-08-19
[PM-40851] Render the shared folder grid on the desktop vault view

jengstrom-bw:
## 🎟️ Tracking

[Jira](https://bitwarden.atlassian.net/browse/PM-40851)

## 📔 Objective

Render the shared folder grid above the collection/item list in the desktop vault-v3 host.
This PR also updates the card grid component to only show 3 rows max based on the container width, per design.

## 📸 Screenshots

https://github.com/user-attachments/assets/780bcbc9-552b-4a34-b0b7-4d058fdd78f6

https://github.com/user-attachments/assets/91c93a05-eac7-41bd-bbc9-aaabb5ac06ac


libs/components/src/accordion/accordion.component.ts
jengstrom-bw, line 1: This fixes a thick border on the collapsable portion of the accordion on the desktop app. If there is a different way UIF wants to accomplish this, I can pull this portion out and it can be done in a follow up ticket

libs/vault/src/components/shared-folder-card-grid/shared-folder-card-grid.component.ts
```diff
@@ -123,6 +137,63 @@ export class SharedFolderCardGridComponent {
...
+   * narrower than {@link MIN_CARD_WIDTH}, so a row holds as many of those — plus the gap between
+   * each pair — as fit, and the `minmax` floor growing with the container caps that at three.
+   *
+   * Falls back to the widest layout until the grid has been measured.
+   */
+  private readonly columns = computed(() => {
+    const width = this.gridWidth();
+    if (width === 0) {
+      return MAX_COLUMNS;
+    }
```
nikwithak, line 188: 🎨 non-blocking: I think you can remove this and it should still work with the `Math.min(Math.max(fit, 1), MAX_COLUMNS);` return statement

nikwithak (approved):

BryanCunningham (approved):

PR #22537, opened 2026-08-19
[PM-42239] Access Intelligence: Match v2 request password change button behavior to v1

lastbestdev:
## 🎟️ Tracking
https://bitwarden.atlassian.net/browse/PM-42239

## 📔 Objective
In version 1 of Access Intelligence, the "Request password change" button on the Critical applications tab allowed admins to click it repeatedly, if there were any at risk members in the Organization. This change reverts the behavior of the button back to match in version 2.

The button itself was also positioned incorrectly after an update from the development of the Access Intelligence coachmark tour a while ago. This is also fixed.

Unit tests were missing for the critical applications tab component, so these changes add them to improve coverage of Access Intelligence V2.

## 📸 Screenshots
**Repeated click behavior (also see correct button positioning)**

https://github.com/user-attachments/assets/006cd75d-ef1b-4990-a954-1ddac77e4ff4


bitwarden_license/bit-web/src/app/dirt/access-intelligence/v2/critical-applications-tab/critical-applications-tab.component.ts
```diff
@@ -117,6 +113,11 @@ export class CriticalApplicationsTabComponent {
     };
   });
 
+  protected readonly enableRequestPasswordChange = computed(() => {
+    const summary = this.applicationSummary();
+    return !!summary && summary.totalAtRiskMemberCount > 0;
```
lastbestdev, line 118: I believe v1 actually looked at the `totalAtRiskMemberCount` field in the summary, but I believe this would cause the button to always be enabled, even when the admin has not marked any critical applications. 

Since having critical applications marked is a prerequisite to assigning password reset tasks, I am checking on the `totalCriticalAtRiskMemberCount` instead here. This will only enable the button when there is at least 1 member with access to a critical application & password.

lastbestdev: Confusing from only looking at this diff - the `summary` value above actually maps the report summary `totalCriticalAtRiskMemberCount` to `totalAtRiskMemberCount` 😅

bitwarden_license/bit-web/src/app/dirt/access-intelligence/v2/critical-applications-tab/critical-applications-tab.component.ts
```diff
@@ -117,6 +113,11 @@ export class CriticalApplicationsTabComponent {
...
+  });
```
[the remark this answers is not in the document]
lastbestdev: This is a non-issue. No extra reset tasks are created in the backend for the same ciphers, and this is what we want.

bitwarden_license/bit-web/src/app/dirt/access-intelligence/v2/critical-applications-tab/critical-applications-tab.component.spec.ts
```diff
@@ -0,0 +1,552 @@
...
+      TestBed.runInInjectionContext(() => {
+        const reportSignal = testAccess(component).report;
+        mockDataService.report$.next(testReport);
+        seen.push(reportSignal());
+        mockDataService.report$.next(testReport);
+        seen.push(reportSignal());
+      });
+
+      expect(seen).toEqual([testReport, testReport]);
+    });
```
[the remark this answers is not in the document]
lastbestdev: Faulty test removed in latest

bitwarden_license/bit-web/src/app/dirt/access-intelligence/v2/critical-applications-tab/critical-applications-tab.component.ts
Banrion, line 118: ❓ I'm curious on why we check `totalAtRiskMemberCount` instead of the count of at risk ciphers? Due to the business logic, I don't think we would ever have a situation where we have an at risk cipher and not an at risk member, but using the cipher as the source of the check is more accurate.

Banrion (approved):

bitwarden_license/bit-web/src/app/dirt/access-intelligence/v2/critical-applications-tab/critical-applications-tab.component.ts
lastbestdev: What you're saying makes sense, but also since the password change tasks/notifications go out to Org members, I also this it makes sense to have it as is - check if any members have access to a critical app with an at risk password

Banrion: My concern is the possible confusion when looking at when can we request security tasks be sent out when the security tasks api only accepts ciphers as the input. The business logic to associate a task to a member is private to that api. Again, product wise, technically correct, but code wise, it's more clear when looking at ciphers. It also guards against a situation of data corruption or bugs if the members to cipher mapping ends up incorrect as far as what ciphers are at risk. It's not a blocking concern though. Just 🎨

PR #22544, opened 2026-08-20
Read PAM failures by code, not by matching the server's English

Hinton:
## 🎟️ Tracking

Last link of a three-repo chain, all targeting `pam/uat`:

- **server** [bitwarden/server#8235](https://github.com/bitwarden/server/pull/8235) — emits stable error codes as RFC 7807 problems
- **sdk-internal** [bitwarden/sdk-internal#1400](https://github.com/bitwarden/sdk-internal/pull/1400) — maps them onto typed error variants
- **clients** — this PR

Fulfils [`docs/pam-server-asks.md`](../blob/pam/uat/docs/pam-server-asks.md), which this PR annotates as resolved rather than deleting: it is the record of *why* the contract exists.

## 📔 Objective

This module kept **two** catalogs of the server's English sentences — eighteen in total, matched with `String.includes()` — because the server had no machine-readable discriminant. They were written weeks apart by different people, and both carried the same note in their own words: _"When the server grows a code, this catalog is the single place to retire."_ It has, so they are.

### Why it mattered

Three of the eighteen are not failures at all. `AlreadyActive`, `AlreadyApproved` and `AlreadyPending` mean the requester already holds what they asked for, and the banner reconciles — collapses the fold-out, re-reads the access state, shows an informational toast. Reword one of those sentences on the server and that reconciliation silently became a red error toast. **No test failed on either side**; the contract lived only in a string literal in this repo.

### What changed

Both classifiers switch on the SDK variant the server's code maps to, and both answer in i18n keys:

```ts
// before
if (message.includes("A reason is required for items that need human approval.")) …
// after
ReasonRequired: { messageKey: "pamRequestAccessErrorReasonRequired", field: "reason" }
```

The request side previously echoed the server's English straight onto the form. It now has its own copy — twelve new keys — which is the other half of what a code buys.

`accessRuleErrorMessage` is deleted along with the catalogs, and with it the `Api`-variant unwrapping underneath: that message is the whole serialized response — envelope, `exceptionMessage`, and the server's filesystem paths — so it was already unfit to show *or log*, which meant a sentence the catalog missed degraded to "something went wrong" rather than to the server's own words. Nothing reads it now; `accessRuleErrorVariant` replaces it.

An unrecognised variant falls back to generic copy, which is what lets the server add codes without waiting on a client release.

Net **−113 lines**, and the module's `CLAUDE.md` now says *switch on `variant`, never on `message`* with the reason, so the pattern does not grow back.

### SDK dependency

The new variants are bridged onto the SDK types exactly the way `NotFound` already was: the Rust side has them, no published `sdk-internal` declares them yet, and the alias collapses on the next bump with nothing else changing. So this **type-checks and passes CI against the current pin** — but the variants only arrive at runtime once sdk-internal#1400 is merged and published, so hold the merge until then.

Verified locally against a WASM build of that branch (`build.sh` + `build.sh -b`, both packages linked): 628 PAM tests pass, `test:types` is clean, lint and prettier are clean.

## 📸 Screenshots

No visual change. The copy shown for a failed request is now ours rather than the server's, and the three reconcile cases behave exactly as before — that behaviour is what this PR protects.


apps/web/src/app/admin-console/organizations/shared/components/collection-dialog/collection-dialog.component.ts
JaredScar, line 123: `Type<unknown>` seems like code-smell to me. Do we need to do it this way?

JaredScar (changes requested):

PR #22546, opened 2026-08-20
fix(pam): render the active lease countdown through the shared access-state badge

maxkpower:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-39642
https://bitwarden.atlassian.net/browse/PM-40498

## 📔 Objective

The active-lease countdown on the Requests page rendered as a raw `<span bitBadge variant="success">`,
so it stayed green to expiry and never warned a member their session was ending.

Routes that cell through the shared `AccessStateBadgeComponent` / `cipherAccessBadgeState` model, so
it escalates to the danger "Ending soon" treatment under five minutes, matching the vault row and the
cipher-view modal.

## 📸 Screenshots

01-before-my-requests
02-after-my-requests

## ⚠️ Known limitations

The shared model has no reachable `unavailable` state: the per-cipher access-state response is
caller-scoped, so nothing on it reveals that another user holds the cipher. The server already knows
(`AccessRule.SingleActiveLease`, `SingleActiveLeaseEvaluator`, and a cross-user `EXISTS` in
`AccessLease_CreateFromApprovedRequest`) but only evaluates it at activation, surfacing a 409.
Exposing it at read time is additive server- and SDK-side and needs no client change.

### DIFF
--- a/bitwarden_license/bit-web/src/app/pam/access-state-badge/access-state-badge.component.ts
+++ b/bitwarden_license/bit-web/src/app/pam/access-state-badge/access-state-badge.component.ts
@@ -5,6 +5,7 @@ import {
   effect,
   inject,
   input,
+  NgZone,
   signal,
 } from "@angular/core";
 
@@ -46,6 +47,7 @@ export class AccessStateBadgeComponent {
   readonly state = input.required<AccessBadgeState | null>();
 
   private readonly i18nService = inject(I18nService);
+  private readonly ngZone = inject(NgZone);
 
   /** Ticks once a second while an active-lease countdown is showing so the label stays live. */
   private readonly now = signal(Date.now());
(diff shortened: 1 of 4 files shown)


bitwarden_license/bit-web/src/app/pam/access-state-badge/access-state-badge.component.ts
```diff
@@ -87,8 +89,14 @@ export class AccessStateBadgeComponent {
...
-      const id = setInterval(() => this.now.set(Date.now()), 1000);
-      onCleanup(() => clearInterval(id));
+      // Outside the Angular zone: a table can host dozens of these at once, and a periodic in-zone
+      // timer both triggers change detection per badge per second and never lets NgZone settle,
+      // which would hang `fixture.whenStable()` for any host embedding them. The signal write
+      // still drives change detection on its own.
+      this.ngZone.runOutsideAngular(() => {
+        const id = setInterval(() => this.now.set(Date.now()), 1000);
+        onCleanup(() => clearInterval(id));
+      });
```
abergs, line 99: improvement, non-blocking: Fine for now, but many badges would create many timers

abergs (approved):

r-tome (approved):

PR #22553, opened 2026-08-20
feat(pam): explain request-gated access when viewing a governed collection

maxkpower:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-40088

## 📔 Objective

A member filtering to a gated collection saw request-gated rows with no collection-level explanation.
Adds a banner above the item list stating the restriction, and lifts the shared gating check into
`services/gated-collection.ts` so the sidebar indicator, this banner and the collection-dialog callout
make one claim from one source.

Icon is `bwi-lock-encrypted`, matching the sidebar indicator.

## 📸 Screenshots

01-before-gated-collection
02-after-gated-collection

## ⚠️ Known limitations

- The banner names the restriction but offers no action; requesting still happens per item, from the
  cipher-view banner.
- It appears only when a single collection is the active filter. Browsing "All items" or an
  organization root shows governed and ungated rows mixed, with the per-row column as the only signal.
- `GovernedCollectionsService` is root-provided so four surfaces share one cached per-org rules read.
  That cache's invalidation is not addressed here: a rule enabled in another tab is not reflected until
  the read refreshes.
- Now that the collection read carries `hasEnabledAccessRule`, this surface and the sidebar indicator
  could drop `GovernedCollectionsService` entirely. Not done here; it is a design change to two
  milestones rather than part of this one.


abergs (approved):

apps/web/src/app/vault/individual-vault/vault-gated-collection-banner.token.ts
```diff
@@ -0,0 +1,16 @@
...
+/**
+ * Optional notice rendered above the vault's item list while a single collection is the active
+ * filter. A host that surfaces a privileged-access feature provides the component class;
+ * `app-vault` mounts it via `NgComponentOutlet`, passing the selected collection's
+ * `organizationId` and `collectionId`. Whether that collection warrants a notice — and whether
+ * its organization has the feature at all — is entirely the provided component's decision, so the
+ * vault never depends on the feature library that implements it. Unprovided, the vault is
+ * unchanged.
+ */
+export const VAULT_GATED_COLLECTION_BANNER = new SafeInjectionToken<Type<unknown>>(
```
gbubemismith: 🎨  Using `Type<Unknown>` here defeats the purpose of `SafeInjectionToken`. The tokens value is a compile time check on the provider. From comments I think we can do this 


```suggestion
export interface VaultGatedCollectionBanner {
  readonly organizationId: InputSignal<OrganizationId | undefined>;
  readonly collectionId: InputSignal<CollectionId | undefined>;
}

export const VAULT_GATED_COLLECTION_BANNER = new SafeInjectionToken<VaultGatedCollectionBanner>(
  "VaultGatedCollectionBanner",
);
```

That way in `vault.component.ts` we can have something like this 

``` ts
protected readonly gatedCollectionBanner: Type<VaultGatedCollectionBanner> | null = inject(
    VAULT_GATED_COLLECTION_BANNER,
    { optional: true },
  );

```

gbubemismith (changes requested):
@maxkpower I just left one comment, every other thing looks good. Nice job!

gbubemismith:
@maxkpower Changes look good. Once the conflicts are fixed, I'll approve this

PR #22556, opened 2026-08-20
fix(pam): match the delete rule dialog to the design's error glyph

maxkpower:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-42278

## 📔 Objective

The delete-access-rule confirmation used `openSimpleDialog` with `type: "warning"`, which renders an
amber glyph. The design calls for a red error mark above a blue primary Delete button, a combination
`openSimpleDialog` cannot express, since it derives glyph, glyph colour and accept-button variant from
the single `type` field and only the glyph is overridable.

Adds `PamConfirmDialogComponent`, a small dialog authored as `<bit-simple-dialog>` in
`bitwarden_license/bit-web/`, which separates those three concerns. No change to `libs/components`:
`DialogModule` already exports `SimpleDialogComponent` and the `bitDialogIcon` directive, and consumers
authoring the dialog directly supply the icon element and footer buttons themselves.

The glyph is **`bwi-clear`**, a circle containing an X. `bwi-error`, despite the name, draws an octagon.

`accessRuleDeleteConfirmOptions` is renamed `accessRuleDeleteConfirmParams` since it no longer returns
`SimpleDialogOptions`; `open()` still resolves a plain boolean, so both call sites keep their guards.

## 📸 Screenshots
01-before-delete-dialog
02-after-delete-dialog

## ⚠️ Known limitations

- **The before glyph reveals a library-level defect, not this milestone's bug.** `type: "warning"`
  renders the class `bwi-exclamation-triangle`, which maps to a font glyph whose internal
  `glyph-name` is `warning` and which draws a **circled exclamation**, not a triangle. Confirmed from
  the font source and by measuring the rendered glyph in two dialogs across both themes. Every
  `type: "warning"` dialog in the product is affected. Out of scope here; worth raising with UI
  Foundation.
- **Enter deletes the rule.** `SimpleDialogComponent` autofocuses the first footer button, which here
  is the destructive Delete. This matches `SimpleConfigurableDialogComponent`'s footer order, so the
  existing behaviour is preserved rather than diverged from, but it deserves a design call.
- `PamConfirmDialogComponent` is generic and reusable; the cancel-request confirm could adopt it with
  only a params factory. Giving that dialog its designed callout would additionally need a projected
  slot, deliberately not built speculatively.
- The accept button in the after capture shows a keyboard focus ring; a re-shot with focus cleared is
  pending so the glyph is the only variable.

### DIFF
--- a/bitwarden_license/bit-web/src/app/pam/confirm-dialog/pam-confirm-dialog.component.html
+++ b/bitwarden_license/bit-web/src/app/pam/confirm-dialog/pam-confirm-dialog.component.html
@@ -1,0 +1,28 @@
+<bit-simple-dialog>
+  <i bitDialogIcon class="tw-text-3xl" [class]="iconClasses" aria-hidden="true"></i>
+
+  <span bitDialogTitle>{{ title }}</span>
+
+  <div bitDialogContent>{{ content }}</div>
+
+  <ng-container bitDialogFooter>
+    <button
+      type="button"
+      bitButton
+      [buttonType]="params.acceptButtonType"
+      id="pam-confirm-dialog_button_accept"
+      (click)="dialogRef.close(true)"
+    >
+      {{ acceptButtonText }}
+    </button>
+    <button
+      type="button"
+      bitButton
+      buttonType="secondary"
+      id="pam-confirm-dialog_button_cancel"
+      (click)="dialogRef.close(false)"
+    >
+      {{ cancelButtonText }}
+    </button>
+  </ng-container>
+</bit-simple-dialog>
--- a/bitwarden_license/bit-web/src/app/pam/CLAUDE.md
+++ b/bitwarden_license/bit-web/src/app/pam/CLAUDE.md
@@ -40,10 +40,10 @@ requester's leasing flow, and the approver's inbox. Gated behind `FeatureFlag.Pa
   collection edit dialog.
 - `confirm-dialog/` — `PamConfirmDialogComponent`, a boolean confirm authored as
   `<bit-simple-dialog>`. Reach for it only when `dialogService.openSimpleDialog` cannot
-  express the design: that helper derives the glyph, the glyph's COLOUR and the accept
-  button's variant from one `type` field, so a red glyph over a blue accept button —
-  what the delete-rule design asks for — is unreachable through it. Everything else
-  stays on `openSimpleDialog`.
+  express the design: that helper lets you override the glyph but derives the glyph's
+  COLOUR and the accept button's variant from one `type` field, so a red glyph over a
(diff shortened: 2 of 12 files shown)


BryanCunningham:
> The before glyph reveals a library-level defect, not this milestone's bug. type: "warning"
renders the class bwi-exclamation-triangle, which maps to a font glyph whose internal
glyph-name is warning and which draws a circled exclamation, not a triangle. Confirmed from
the font source and by measuring the rendered glyph in two dialogs across both themes. Every
type: "warning" dialog in the product is affected. Out of scope here; worth raising with UI
Foundation.

FWIW: This is not an error, per se. And, is a known compromise made on the UIF side. The legacy icon name was kept in tact/mapped to the new names as to not churn the consumer codebase with icon name changes.

BryanCunningham:
@maxkpower The danger type of simple dialog enforces a 'danger' accept button by design. We should probably not override this behavior, for consistency purposes.

BryanCunningham (changes requested):
I understand that the design and the component behavior are different here. Apologies for it seeming like we needed a workaround here. I've talked with design and the current component behavior of enforcing a 'danger' button when the dialog type is danger is what we aligned on moving forward. 

FWIW, we'll also be enforcing a 'warning' button when the dialog type is warning in a future update.

JimmyVo16:
I can review this once the merge conflicts are resolved.

abergs (approved):

bitwarden_license/bit-web/src/app/pam/confirm-dialog/pam-confirm-dialog.component.html
Hinton, line 1: This should be a simple dialog, why are we creating a full dialog here?

Hinton (changes requested):

bitwarden_license/bit-web/src/app/pam/confirm-dialog/pam-confirm-dialog.component.html
abergs: @Hinton `bit-simple-dialog` is that not a simple dialog?

abergs: IF you mean why are we creating a new file here; i think it's because we wanted to minimize modifications to existing files in favor of additive changes

Hinton: That's all supported in the programmatic api https://components.bitwarden.com/?path=/docs/component-library-dialogs-service-simpleconfigurable--docs. Very very few cases needs to actually create a new simple dialog component.

PR #22560, opened 2026-08-20
perf(pam): share one clock across the access-state badges

maxkpower:
## 🎟️ Tracking

Follow-up to a non-blocking review note on #22546: "Fine for now, but many badges would create many timers."

## 📔 Objective

`AccessStateBadgeComponent` started its own `setInterval` whenever it showed an active-lease
countdown, so a vault table with N running leases held N independent per-second timers. The badge
renders on five surfaces, and the vault-row host renders one per table row, so the count scales with
page size rather than with anything meaningful.

Adds `AccessBadgeTickerService`, a root-provided clock built on a single interval shared through
`share({ resetOnRefCountZero: true })`. The interval starts when the first active badge observes it
and is torn down when the last one stops, so a screen with no active lease holds no timer at all.

The badge gates its subscription on its own state, so resting rows never reach the ticker. This is
expressed as `toObservable(...)` piped into `switchMap` and read back through `toSignal`, which keeps
the subscription lifecycle owned by the framework rather than by a hand-managed `subscribe`.

The interval is still created inside `runOutsideAngular`. That rationale is unchanged and now lives
in one place: a periodic in-zone timer would trigger change detection every second for as long as any
badge is active and would never let `NgZone` settle, which hangs `fixture.whenStable()` for any host
embedding a badge.

No behaviour change: same one-second cadence, same countdown, same five-minute danger escalation,
same expired fallback.

## ✅ Tests

Three new component specs pin the property this PR exists for, all of which fail against the previous
implementation:

- resting badges start no timer at all
- three concurrent active badges share exactly one `setInterval`
- the interval survives the first badge's destruction and is cleared only with the last

Plus five service specs covering cadence, sharing, out-of-zone creation, ref-counted teardown, and
restart after a full teardown.

`bitwarden_license/bit-web/src/app/pam`: 63 suites, 653 tests passing. ESLint and Prettier clean on
the touched files; `tsc -p bitwarden_license/bit-web/tsconfig.json` reports nothing for them.

## ⚠️ Known limitations

- All badges now tick from one clock, so they update in lockstep rather than each on its own phase
  offset from when it mounted. For a whole-second countdown this is not observable, and arguably it
  is the more correct behaviour.
- The ticker is PAM-local. If another surface later wants a shared one-second clock, this is worth
  promoting rather than copying.

### DIFF
--- a/apps/web/src/app/pam/user-nav-slot/pam-user-nav-slot.component.spec.ts
+++ b/apps/web/src/app/pam/user-nav-slot/pam-user-nav-slot.component.spec.ts
@@ -85,6 +85,14 @@ describe("PamUserNavSlotComponent", () => {
     expect(navItem()).toBeNull();
   });
 
+  it("generates no box of its own, so rendering nothing takes up no space", () => {
+    organizations$.next([nonPamOrg]);
+    fixture.detectChanges();
+
+    expect(navItem()).toBeNull();
+    expect(fixture.debugElement.nativeElement.classList).toContain("tw-contents");
+  });
+
   describe("the pending-count badge", () => {
     const badge = () => fixture.debugElement.query(By.css("[bitBadge]"));
 
--- a/apps/web/src/app/pam/user-nav-slot/pam-user-nav-slot.component.ts
+++ b/apps/web/src/app/pam/user-nav-slot/pam-user-nav-slot.component.ts
@@ -27,6 +27,7 @@ import { PamNavBadgeService } from "../pam-nav-badge.service";
 @Component({
   selector: "app-pam-user-nav-slot",
   templateUrl: "./pam-user-nav-slot.component.html",
+  host: { class: "tw-contents" },
   imports: [BadgeModule, I18nPipe, NavigationModule],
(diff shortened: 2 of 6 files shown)


apps/web/src/app/layouts/user-layout.component.html
```diff
@@ -2,131 +2,63 @@
...
-      <bit-nav-group
-        icon="bwi-wrench"
-        [text]="'tools' | i18n"
-        route="tools"
-        [open]="toolsNavGroupOpen()"
+    <bit-nav-item icon="bwi-vault" [text]="'vaults' | i18n" route="vault"></bit-nav-item>
+    @if (sendEnabled$ | async) {
+      <bit-nav-item icon="bwi-send" [text]="'send' | i18n" route="sends"></bit-nav-item>
+    }
+    <app-pam-user-nav-slot />
```
BryanCunningham: Any way we conditionally render this in the side nav? The new(vfo1) side nav will apply a flex gap here even if the this container is empty/doesn't render anything.

abergs (approved):

PR #22567, opened 2026-08-21
[PM-42183] VFO1 Side nav vault scoping

shane-melton:
## 🎟️ Tracking

[PM-42183](https://bitwarden.atlassian.net/browse/PM-42183)

## 📔 Objective

Wire the Password Manager side nav's vault destinations to real routes. Every nav destination — All items, My vault, an organization, Archive, and Trash — now renders `VaultNextComponent` scoped by a `:vaultId` route segment, so each is a bookmarkable URL with correct active-route highlighting.

Important changes:

- **New `VaultScope` model** (`libs/vault/src/models/vault-scope.ts`) — a single place that parses the `:vaultId` segment, builds scope URLs, and decides whether a cipher, collection, or organization belongs in a scope.
- **New `vaultScopeGuard`** — redirects to the unscoped vault when the segment names no vault the user has (typo, or a bookmark to an org they've left), instead of rendering an empty vault.
- **`:vaultId` route added** to the individual vault routing module, behind `VFO1Foundation`; falls back to `/vault` when the flag is off. "All items" remains the unscoped route, so existing deep links are unchanged.
- **`VaultNextComponent` scoping** — the ciphers stream no longer pre-filters deleted/archived items; the scope decides. Collections, organizations, and the search index are narrowed to the scope so chips can't offer options no row could match. Import/New item are hidden on Trash and Archive. Banners and onboarding read the full active vault rather than the scoped rows.
- **Legacy redirect handling** — `?type=trash` and `?type=archive` now redirect to their own routes rather than being dropped as unmappable filter params.
- **New `VaultManageNavComponent`** — moves the Manage section's My folders / Archive / Trash items out of `user-layout.component` and into a vault-owned component, keeping the archive premium-upgrade path and adding archive active-route highlighting.
- **`VaultNavSectionComponent`** switched from imperative `router.navigate` to `routerLink`, with exact matching on All items so a scoped child doesn't light it up.

## 📸 Screenshots

https://github.com/user-attachments/assets/343d8296-dc8c-40df-9247-2e6075ad7c24

[PM-42183]: https://bitwarden.atlassian.net/browse/PM-42183?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


shane-melton:
@gbubemismith / @nick-livefront Here's a first pass at implementing the side nav scoping we talked about that could use some eyes on the approach. It should be re-usable on Desktop (and potentially Browser) as is, but please let me know if you see any problems. 

It still needs some work to support the My Items collections for orgs, but I'm leaning towards a followup PR to keep this manageable.

libs/vault/src/components/vault-nav-section/vault-nav-section.component.ts
```diff
@@ -27,17 +28,43 @@ import { MY_VAULT } from "../vault-items-table/vault-items-table.component";
...
-  protected async selectAllItems() {
-    await this.router.navigate(["/vault"]);
-  }
+  protected readonly allItemsRoute = vaultScopeCommands(ALL_ITEMS_SCOPE);
+
+  /**
+   * Every scoped vault route nests under All items' own, so a subset match would leave All items
+   * lit alongside the vault the user picked.
+   */
+  protected readonly allItemsActiveOptions: IsActiveMatchOptions = {
```
gbubemismith, line 45: 👍🏼 nice

libs/vault/src/components/vault-manage-nav/vault-manage-nav.component.html
gbubemismith, line 1: Thanks for doing this

libs/vault/src/components/vault-nav-section/vault-nav-section.component.html
```diff
@@ -12,15 +16,15 @@
...
+          [route]="vaultRoute(vault)"
         ></bit-nav-item>
       </bit-nav-group>
     }
   </ng-template>
 
   @if (nav.vaults.length === 1 && !nav.organizationDataOwnership) {
     @let loneVault = nav.vaults[0];
-    <bit-nav-item [text]="loneVault.label" (mainContentClicked)="selectVault(loneVault)">
+    <bit-nav-item [text]="loneVault.label" [route]="vaultRoute(loneVault)">
```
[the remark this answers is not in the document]
shane-melton: Fixed [1a8e6d6](https://github.com/bitwarden/clients/pull/22567/commits/1a8e6d6d2ba4e9dba710bf8e88ae7e4d329bf554)

libs/vault/src/routing/vault-filter-legacy-redirect.guard.ts
```diff
@@ -139,5 +154,11 @@ export const vaultFilterLegacyRedirectGuard: CanActivateFn = async (route) => {
   }
   Object.assign(queryParams, patch);
 
+  // A scope is a different route, not a param on this one, so it can't ride along on the
+  // snapshot-relative tree the param-only redirect uses.
+  if (scope != null) {
+    return router.createUrlTree(vaultScopeCommands(scope), { queryParams });
```
gbubemismith, line 161: ❓ How will this work on desktop?

shane-melton: I was imagining we make a similar routing module update to Desktop so that they both use the same routes and can share the guard/scope services.

gbubemismith: Okay that makes sense. I'll add that on the desktop nav restructure. I am going to base off this branch

shane-melton: Sounds good!

libs/vault/src/routing/vault-scope.guard.ts
```diff
@@ -0,0 +1,39 @@
...
+  const allItems = () => router.createUrlTree(["/vault"]);
+
+  const scope = parseVaultScope(route.paramMap.get("vaultId"));
+  if (scope == null) {
+    return allItems();
+  }
+
+  if (scope.type !== VaultScopeType.Organization) {
+    return true;
+  }
```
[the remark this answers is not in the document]
shane-melton: Fixed [60073d3](https://github.com/bitwarden/clients/pull/22567/commits/60073d3af726b1a72b78cf16ba03635a4aa2f36a)

BryanCunningham (approved):

gbubemismith:
@shane-melton, I noticed something. I aded the matching routes on desktop; `/vault` with a `:vaultId` child guarded the same way as web. Desktop's `VaultComponent` still filters through `RoutedVaultFilterService` which reads the org from `params.get("organizationId")`  or the `organizationId` / `vaultId` **query** params. Ours is `vaultId` as a *path* param, so all three miss and `createFilterFunction` becomes a no-op. What seems to work is translating the scope into the legacy filter before it's used. Should I handle this on my end or is this a good place to do it?

shane-melton:
@gbubemismith I think we'll just want to drop using the `RoutedVaultFilterService` for desktop when the VFO flag is enabled. It doesn't really apply now that the filtering has been moved into the table itself for VFO1. 

We should be able to use this [`cipherInScope()` helper](https://github.com/bitwarden/clients/blob/1a8e6d6d2ba4e9dba710bf8e88ae7e4d329bf554/libs/vault/src/models/vault-scope.ts#L138-L162) in desktop's vault component when the flag is enabled. 

Either way, this PR probably shouldn't include any desktop changes as its already so large.

gbubemismith (approved):

PR #22585, opened 2026-08-21
[PM-40316] Restructure the desktop side navigation

gbubemismith:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-40316

## 📔 Objective

The desktop counterpart to the web side-nav restructure in PM-40313. All changes sit behind the `vfo1-foundation` flag; the legacy navigation and vault filtering are untouched when it is off.

**Navigation.** The desktop-only vault filter is replaced by the shared `VaultNavSectionComponent` and `VaultManageNavComponent`, so Vaults, Tools and Manage match web. Both navigate through `routerLink`, which let the layout drop its vault, archive and trash handlers along with four service injections.

**Routing.** `/vault/:vaultId` is registered alongside the unscoped `/vault`, guarded by `canAccessFeature(VFO1Foundation, ...)` and `vaultScopeGuard` — the same shape as web's `VaultRoutingModule`. Without it the shared nav's links, and the `vaultFilterLegacyRedirectGuard` redirects to `/vault/trash` and `/vault/archive`, resolved to nothing on desktop.

## 📸 Screenshots

https://github.com/user-attachments/assets/3afffd41-426e-4c20-b500-76457d96df12

### Side Nav with vaults
image

### Light mode 

image


apps/desktop/src/app/layout/desktop-layout.component.html
```diff
@@ -2,16 +2,43 @@
...
+        />
+        <bit-nav-item
+          icon="bwi-download"
+          [text]="'exportNoun' | i18n"
+          (mainContentClicked)="openExport()"
+        />
+      </bit-nav-section>
+
+      <bit-nav-section [label]="'manage' | i18n">
+        <vault-manage-nav />
```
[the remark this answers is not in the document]
gbubemismith: Don't worry about this, there's an upcoming PR for this

apps/desktop/src/vault/app/vault-v3/vault.component.ts
```diff
@@ -251,48 +261,73 @@ export class VaultComponent<C extends CipherViewLike> implements OnInit, OnDestr
...
+    switchMap((vfo1Foundation) =>
+      vfo1Foundation
+        ? combineLatest([
+            this.route.paramMap.pipe(map((params) => params.get("vaultId"))),
+            this.vaultNavService.viewModel$,
+          ]).pipe(map(([vaultId, nav]) => resolveVaultScope(vaultId, nav) ?? ALL_ITEMS_SCOPE))
+        : of(ALL_ITEMS_SCOPE),
+    ),
+    shareReplay({ refCount: true, bufferSize: 1 }),
+  );
```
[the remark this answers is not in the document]
gbubemismith: #22260 Is in the best position to fix this.

apps/desktop/src/app/layout/desktop-layout.component.html
```diff
@@ -2,16 +2,43 @@
   <app-side-nav slot="side-nav">
     <bit-nav-logo [openIcon]="logo" route="." [label]="'passwordManager' | i18n" />
 
-    @if (!vfo1Foundation()) {
+    @if (vfo1Foundation()) {
+      <vault-nav-section />
```
shane-melton, line 6: :+1: Glad we get to reuse these! :tada:

apps/web/src/app/vault/individual-vault/vault-routing.module.ts
```diff
@@ -21,9 +21,13 @@ const routes: Routes = [
     },
   }),
   {
+    // The side nav's vault scopes: `my-vault` and an organization id. "All items" is the unscoped
```
shane-melton: :pick: (non-blocking) I dropped these comments in my original PR as they were explained elsewhere better and sound very claude-like ("whole story here"...). Open to keeping if you find them beneficial.

apps/desktop/src/vault/app/vault-v3/vault.component.ts
```diff
@@ -604,23 +645,37 @@ export class VaultComponent<C extends CipherViewLike> implements OnInit, OnDestr
...
+      .subscribe(
+        ([allCollections, allOrganizations, ciphers, collections, vfo1Foundation, scope]) => {
+          // The table builds its Vault and Shared folders chips from these.
+          this.allCollections = vfo1Foundation
+            ? allCollections.filter((collection) => collectionInScope(collection, scope))
+            : allCollections;
+          this.allOrganizations = vfo1Foundation
+            ? allOrganizations.filter((organization) => organizationInScope(organization, scope))
+            : allOrganizations;
+          this.ciphers = ciphers;
```
shane-melton, line 671: :question: Would it make sense to do the same `cipherInScope` filtering here instead of the `rowFilter` change above? Just trying to think what would be easiest for eventual removal of the VFO1 feature flag.

apps/desktop/src/vault/app/vault-v3/vault.component.ts
shane-melton, line 1: 🎨 (non-blocking) The page header is still always "Vault" and doesn't update based on the scope. We can set the `[title]` on `app-header` in the template based on the scope. Though, I'm okay with leaving this alone since it'll be reworked in the breadcrumb epic.

shane-melton (dismissed):
Changes look good! Just a few non-blocking comments/suggestions.

apps/desktop/src/vault/app/vault-v3/vault.component.ts
gbubemismith: Yeah I left this out cause I saw a ticket that handles this

apps/desktop/src/vault/app/vault-v3/vault.component.ts
gbubemismith: I think having it in rowFilter$ is actually an easier removal. The same filtering is also used by the bulk batch bar.

apps/web/src/app/vault/individual-vault/vault-routing.module.ts
gbubemismith: I wonder why it wasn't removed when I rebased

gbubemismith: Fixed 9e0246aa2c5f154a1475fe55f9222bf5e83391ae

apps/desktop/src/vault/app/vault-v3/vault.component.ts
```diff
@@ -604,23 +645,37 @@ export class VaultComponent<C extends CipherViewLike> implements OnInit, OnDestr
...
-        // Explicitly mark for check to ensure the view is updated
-        // Some sources are not always emitted within the Angular zone (e.g. ciphers updated via WS server notifications)
-        this.changeDetectorRef.markForCheck();
-      });
+      .subscribe(
+        ([allCollections, allOrganizations, ciphers, collections, vfo1Foundation, scope]) => {
+          // The table builds its Vault and Shared folders chips from these.
+          this.allCollections = vfo1Foundation
+            ? allCollections.filter((collection) => collectionInScope(collection, scope))
+            : allCollections;
```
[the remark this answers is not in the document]
shane-melton: @gbubemismith This seems legit and something we will want to address

shane-melton (changes requested):
I think we just need to address the Claude finding above

apps/desktop/src/vault/app/vault-v3/vault.component.ts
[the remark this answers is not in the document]
gbubemismith: Fixed 9e0246aa2c5f154a1475fe55f9222bf5e83391ae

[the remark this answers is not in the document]
shane-melton: @gbubemismith That doesn't seem like right commit hash and I'm not seeing any new commits pushed?

shane-melton (approved):

PR #22588, opened 2026-08-21
[PM-41475] Add the My folders page to desktop

gbubemismith:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-41475

## 📔 Objective

Adds the My folders page to desktop, behind the `VFO1Foundation` feature flag.

The Manage section of the restructured desktop side nav already renders a My folders entry pointing at `route="folders"` (via `VaultManageNavComponent`, from #22585), but no route was registered to serve it. This adds that route and the page it lands on.

The page itself is a thin desktop wrapper around the shared `MyFoldersComponent` in `libs/vault`, which web already uses — it projects the desktop `<app-header>` into the shared component and supplies the page padding desktop pages own for themselves. No folder logic is duplicated.

Also copies the six folder-management i18n keys the shared component needs into the desktop locale, per the repo's i18n rule that shared library strings need the key in each consuming app's `messages.json`.

### My folders page
image

### Folders page with delete dialog
image

image


nick-livefront (dismissed):
An integration PR like this is so simple, it really shows that the planning and thought behind the components is worth it 💯

shane-melton (dismissed):
🚀

shane-melton (approved):

PR #22593, opened 2026-08-22
[PM-40157] Change compile-time flags to default off

trmartin4:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-40157

## 📔 Objective

Reverses the logic of the compile-time `flags` so they default to `false` when not populated, rather than defaulting to `true`. This aligns with how our LaunchDarkly flags work and is safer against accidental enabling of a flag due to misunderstanding of the paradigm.

The only other flag currently in use is `sdk`, which is already set to `true` in all configurations, so there are no side-effects of this change for existing flags.

Split out of #22191 per review feedback so the change is documented independently of the beta icon work and is easy to revert if needed.

### DIFF
--- a/libs/common/src/platform/misc/flags.spec.ts
+++ b/libs/common/src/platform/misc/flags.spec.ts
@@ -5,8 +5,8 @@ describe("flagEnabled", () => {
     process.env.FLAGS = JSON.stringify({});
   });
 
-  it("returns true by default", () => {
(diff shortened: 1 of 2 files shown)


trmartin4:
Tagging @djsmith85 for context from #22191.

djsmith85 (approved):
Approving as standard client and beta client now produce the correct flag settings and add the Is-Prerelease is only added to beta clients.

PR #22445, opened 2026-08-15
Add desktop beta icons

trmartin4:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-40157

## 📔 Objective

Add beta icons to desktop builds.

See https://github.com/bitwarden/clients/pull/22191 for corresponding extension changes.

PR #22595, opened 2026-08-22
[PM-42586] Add translated beta name to desktop tray and about menu

trmartin4:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-42586

## 📔 Objective

Builds on https://github.com/bitwarden/clients/pull/22445.  

1. Adds the "Bitwarden" name on the desktop tray and the Help / About menu to reference "Bitwarden Beta" instead of "Bitwarden".
2. Introduces the Bitwarden logo on the landing pages for desktop (outside of the Vault).
3. Removes unused logo from `AnonLayout` that was orphaned in #17969.

### Layout hierarchy (what's used where)

Background on why the change is in the `landing-header` and not `anon-layout`:

- **`bit-landing-header`**: the header primitive. Renders the Bitwarden logo on the left and a projected slot for actions on the right. **This is the file changed in this PR.**
- **`auth-anon-layout` (`AnonLayoutComponent`)**: an auth-flavored composition of the primitives. Wires in the title, subtitle, hero icon, environment selector, hostname readout, and footer year plus version. Uses `<bit-landing-header>` internally, which is what pulls in the logo.

#### Why the browser extension is different

The extension does not go through `<bit-landing-header>`. `ExtensionAnonLayoutWrapperComponent` sets `[hideLogo]="true"` on `<auth-anon-layout>` and renders its own `<bit-svg>` inside the extension's popup header bar. This is required because the logo needs to sit at the top of the popup shell, not inside the anon-layout's card region. The extension wrapper carries its own `flagEnabled("prereleaseBuild") ? BitwardenLogoBeta : BitwardenLogo` binding to drive that render, and it already worked before this change.

## 📸 Screenshots

image

image

image

### DIFF
--- a/libs/components/src/landing-layout/landing-header.component.ts
+++ b/libs/components/src/landing-layout/landing-header.component.ts
@@ -1,7 +1,8 @@
 import { ChangeDetectionStrategy, Component, input } from "@angular/core";
 import { RouterModule } from "@angular/router";
 
-import { BitwardenLogo } from "@bitwarden/assets/svg";
+import { BitwardenLogo, BitwardenLogoBeta } from "@bitwarden/assets/svg";
+import { flagEnabled } from "@bitwarden/common/platform/misc/flags";
 import { I18nPipe } from "@bitwarden/ui-common";
 
 import { SvgModule } from "../svg";
@@ -38,5 +39,5 @@ import { SvgModule } from "../svg";
 })
 export class LandingHeaderComponent {
   readonly hideLogo = input<boolean>(false);
-  protected readonly logo = BitwardenLogo;
+  protected readonly logo = flagEnabled("prereleaseBuild") ? BitwardenLogoBeta : BitwardenLogo;
 }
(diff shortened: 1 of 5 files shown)


djsmith85 (approved):

djsmith85 (dismissed):

djsmith85 (approved):

libs/components/src/landing-layout/landing-header.component.ts
```diff
@@ -38,5 +39,5 @@ import { SvgModule } from "../svg";
 })
 export class LandingHeaderComponent {
   readonly hideLogo = input<boolean>(false);
-  protected readonly logo = BitwardenLogo;
+  protected readonly logo = flagEnabled("prereleaseBuild") ? BitwardenLogoBeta : BitwardenLogo;
```
BryanCunningham, line 42: @trmartin4 Should this be the vfo1 flag? That is what all the other changes are bifurcated on currently

djsmith85: @BryanCunningham This is expected. Similar comment from Vicky on another layout component: https://github.com/bitwarden/clients/pull/22191#discussion_r3874805994

BryanCunningham (approved):

PR #22599, opened 2026-08-23
[PM-42345] Fix Chromium import when elevated as a different Windows user

detunized:
## 🎟️ Tracking

Chromium password import failure when UAC elevation uses a different Windows account.

## 📔 Objective

Decrypt Chromium App-Bound Encryption data as the original user rather than the elevated administrator. The helper
now captures and impersonates the validated client process token for the user-bound DPAPI layer.


harr1424 (dismissed):
Approving now so that we can get a CI build to CS, before this is finalized please be certain that the current error logging won't leak anything sensitive.

harr1424 (approved):

PR #22600, opened 2026-08-24
tests: add test for device trust / shared unlock / unlock ordering

quexten:
## 🎟️ Tracking

https://github.com/bitwarden/clients/pull/22357#discussion_r3841154221

## 📔 Objective

Adds tests to prevent re-introduction of a bug where you unlock before persisting the trust device option (which also sets the shared unlock option). Essentially, shared unlock is toggled depending on device trust, so unlocking would immediately share device unlock state before persisting the device trust option, which is not intended.


rr-bw (approved):
Thanks for adding these

PR #22611, opened 2026-08-24
[CL-1244] icon tiles in menus

BryanCunningham:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/CL-1244

## 📔 Objective

Allow icon tiles in menu items

#### Multi-select
image

#### Menu 
image

#### Select
image

#### Chip filter
image

#### Filter menu
image


libs/components/src/select/select.component.html
```diff
@@ -13,13 +13,23 @@
...
 >
+  <ng-template ng-label-tmp let-item="item">
+    <div class="tw-flex tw-items-center tw-gap-2 tw-min-w-0" [title]="item.label">
+      <ng-container
+        [ngTemplateOutlet]="leadingVisual"
+        [ngTemplateOutletContext]="{ item: item }"
+      ></ng-container>
+      <span class="tw-truncate">{{ item.label }}</span>
+    </div>
+  </ng-template>
```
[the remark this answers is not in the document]
BryanCunningham: This is the intended result

vleague2 (dismissed):

libs/components/src/select/select.component.ts
```diff
@@ -90,6 +107,7 @@ export class SelectComponent<T> implements ControlValueAccessor {
         this.items.set(
           opts.map((option) => ({
             icon: option.icon(),
+            iconTile: option.iconTile(),
```
jaasen-livefront: Heads-up from a consumer — this line can hang change detection, and I think it's worth a look while the PR is still in draft.

This `afterRenderEffect` both **reads** `option.iconTile()` and **writes** `this.items`. Signal equality is reference-based, so a consumer that builds the tile object inline gets a new reference every render, which never compares equal to the previous one: read → write → re-render → new reference → read → …

Worth noting `iconTile` is the only object-typed field in this map — `icon`, `value`, `label`, `description`, and `disabled` are all primitives, where value equality holds. So this is a new hazard rather than an existing one.

**Symptom:** hard app freeze, no error, no console warning, no stack trace pointing anywhere near the select. I hit it consuming `iconTile` in the vault item form (PM-40323) — the app locked up when opening the cipher form.

**Reproducing binding** — the natural way to write it:

```html
<bit-option [iconTile]="getTile(org)" [value]="org.id" [label]="org.name" />
```

```ts
getTile(org: Organization): IconTileOptions {
  return { icon: "bwi-business", variant: "purple", emphasis: "bold" };
}
```

**Our workaround:** precompute tiles once into a `Map<orgId, IconTileOptions>` and bind `orgIconTiles.get(org.id)` so the reference is stable. So we're unblocked either way — this is about sparing the next consumer.

**Scope:** I checked `filter-menu` — its `effect` reads `label()`, not `iconTile()`, so it doesn't have the same read/write cycle. As far as I can tell this is specific to `bit-select`.

**Possible fixes,** if you agree it's worth addressing:

1. Document it on `IconTileOptions` / the `iconTile` input — "hold this in a computed/field."
2. Give the input a value-based `equal` function. The object is small and flat, so a structural compare is cheap and makes the natural binding style just work with no consumer changes.
3. Skip `items.set()` when the mapped output is equivalent — fixes the cycle for any future object-typed option field, not just this one.

I'd lean toward (2), but you have far more context on the component's design than I do — entirely your call, and quite possibly stable references are the intended contract.

BryanCunningham: @jaasen-livefront I've fixed this [here](https://github.com/bitwarden/clients/pull/22611/changes/983c240d47199f4841c1d1538a7cb6e71adc6060)

Thanks for the heads up!

vleague2 (approved):

jaasen-livefront (approved):
Looks good thanks!

PR #22614, opened 2026-08-24
[PM-40490] feat: Block create-organization route in FedRAMP environment

cyprain-okeke:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-40490

## 📔 Objective

Blocks all self-serve organization-creation flows in the FedRAMP (Gov cloud) environment, where organizations are sales-provisioned.

- Adds `govModeBlockedGuard` to the `/create-organization` and `/settings/add-plan` routes (the latter per the PM-37197 contingency in the ticket's AC), and — per review feedback — to the unauthenticated `/trial-initiation` and `/secrets-manager-trial-initiation` routes, which also create organizations; the shared Gov predicate (`clientIsGovMode$`) falls back to the global environment when signed out. Blocked navigation shows an "Access denied" toast and redirects to `/vault`. The guard deliberately **fails open**: an error while determining the region is logged and navigation is allowed, so US/EU users are never locked out of flows they're entitled to.
- Hides every UI entry point that leads to those routes in Gov mode: the New organization button (org switcher), the product switcher's Organizations item, the "Add plan" nav item, the vault filter's add-organization action, and the Secrets Manager landing CTAs (`sm-landing` hides "Try it now"; `request-sm-access` redirects no-org users to `/sm-landing` instead of `/create-organization`).
- Gov-mode state is derived everywhere through a shared fail-open `activeUserIsGovMode$` helper; a routing spec pins the guard wiring on both routes so dropping `canActivate` fails a test.

**Open product questions (flagged to product, non-blocking):**
1. Blocked deep-links currently show the generic `accessDenied` toast copy — environment-specific copy may be preferable.
2. In Gov mode, a user with no organizations sees the SM landing page without a CTA (only the external "Learn more" link) — confirm this is the intended experience vs. a "contact your representative" message.

## 📸 Screenshots

UI is unchanged outside the FedRAMP environment. In Gov mode the self-serve entry points listed above are hidden; recordings from a Gov-configured environment to follow if needed for QA.


JimmyVo16 (dismissed):

jengstrom-bw (dismissed):

amorask-bitwarden (dismissed):
Looks good from the billing perspective. The only ⛏️ I have is that there's a lot of code-level commenting that I would consider unnecessary. Basically stating "this can't be done in GovMode" when that's clearly what the code is already doing. My preference would be to remove that, but I won't block the PR on it.

vleague2 (changes requested):
The most recent UI Review that I can see on the commits is showing broken stories (in some, the UI is missing entirely). Can you check those stories and check if they need updated mocks?

cyprain-okeke:
@vleague2 Thanks for catching that — checked all stories that touch `ProductSwitcherService` and fixed two things:

1. **The broken snapshots you saw** were the product-switcher / navigation-switcher stories on the mid-PR commits: they provide `ProductSwitcherService` explicitly, so the two new constructor deps (`GovModeService`, `LogService`) threw `NullInjectorError` and the stories rendered nothing. Mocks were added in 7087b42 (class-based `Partial<T>` mocks — these stories can't use jest mocks).
2. **Sponsored Families / Free Bitwarden Families stories** (they render the real `HeaderModule` → `<product-switcher>`) had a subtler, pre-existing issue: `ProductSwitcherService` is `providedIn: "root"`, so it was instantiated in the root injector where the stories' `moduleMetadata` mocks aren't visible — the switcher has been silently failing there since before this PR (`NG0201: ProductSwitcherService -> OrganizationService`). Fixed in ed5c03a by providing `ProductSwitcherService` in the story injector with the full mock set (including the new Gov/Log deps), so the header's switcher now actually renders in those stories. Note this means those two snapshots will show the switcher for the first time.

Verified all four stories render with zero console errors in local Storybook. (Local `storybook dev` was also broken by the out-of-date `EncryptService` import on main — already fixed there by #22666, merged in.)

vleague2 (dismissed):
Approving for UIF

jengstrom-bw (dismissed):

JimmyVo16 (dismissed):

amorask-bitwarden (dismissed):

jengstrom-bw (dismissed):

JimmyVo16 (dismissed):

vleague2 (dismissed):

amorask-bitwarden (dismissed):

JimmyVo16 (dismissed):

apps/web/src/app/platform/gov-mode.ts
```diff
@@ -0,0 +1,38 @@
...
+ * Checks the active account's environment, or the global environment when signed out (e.g. the
+ * unauthenticated trial-initiation routes).
+ *
+ * Fails open: any error while determining the region is logged and emitted as `false`, so
+ * consumers keep their self-serve entry points rather than breaking. The route-level
+ * govModeBlockedGuard remains the backstop for the blocked flows themselves.
+ *
+ * @param context - Optional caller context appended to the fail-open log entry.
+ */
+export const clientIsGovMode$ = (
```
addisonbeck, line 11: Why is this needed versus just using govModeService directly? Why is it just in the web app?

apps/web/src/app/platform/gov-mode.ts
```diff
@@ -0,0 +1,38 @@
...
+  context?: string,
+): Observable<boolean> =>
+  accountService.activeAccount$.pipe(
+    getOptionalUserId,
+    switchMap((userId) =>
+      userId != null ? govModeService.isGovMode$(userId) : govModeService.globalIsGovMode$,
+    ),
+    catchError((error: unknown) => {
+      logService.error(
+        `clientIsGovMode$: failed to determine Gov mode, failing open${context ? ` (${context})` : ""}`,
```
addisonbeck: When would this happen?

apps/web/src/app/platform/gov-mode.ts
```diff
@@ -0,0 +1,38 @@
...
+ * Emits whether this client is connected to the Gov cloud, where organizations are
+ * sales-provisioned and self-serve flows such as organization creation are unavailable (PM-40490).
+ * Checks the active account's environment, or the global environment when signed out (e.g. the
+ * unauthenticated trial-initiation routes).
+ *
+ * Fails open: any error while determining the region is logged and emitted as `false`, so
+ * consumers keep their self-serve entry points rather than breaking. The route-level
+ * govModeBlockedGuard remains the backstop for the blocked flows themselves.
+ *
+ * @param context - Optional caller context appended to the fail-open log entry.
```
addisonbeck: Gov cloud is not the only gov environment. Otherwise, this comment is overly verbose. 

> where organizations are sales-provisioned and self-serve flows such as organization creation are unavailable (PM-40490)

This is irrelevant, for example.

apps/web/src/app/platform/gov-mode.ts
cyprain-okeke: `GovModeService` exposes two streams (`isGovMode$(userId)` and the pre-login `globalIsGovMode$`) and this helper picks between them based on the active account, the guard runs on both authenticated routes and the unauthenticated trial-initiation routes.

apps/web/src/app/platform/gov-mode.ts
cyprain-okeke: removed in ab0e78dbb04.

apps/web/src/app/platform/gov-mode.ts
cyprain-okeke: Trimmed in ab0e78dbb04.

JimmyVo16 (dismissed):

apps/web/src/app/platform/gov-mode.ts
addisonbeck: Could this just be baked in to `isGovMode` in `govModeService`? Would it work to make `userId` nullable, and early return `globalGovMode` if no id is supplied? Having this logic live separate from the `govModeService` seems off, and this sounds like it could be behavior that is expected across clients.

JimmyVo16 (dismissed):

amorask-bitwarden (approved):

vleague2 (approved):

jengstrom-bw (approved):

JimmyVo16 (approved):

PR #22646, opened 2026-08-25
refactor(km): simplify process reload service

quexten:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-42444

## 📔 Objective

Simplifies the process reload service by removing the now unused callback / cancel logic, and by removing the passed in callback.


apps/browser/src/background/main.background.ts
```diff
@@ -1486,31 +1485,20 @@ export default class MainBackground {
       this.logService,
     );
 
-    const systemUtilsServiceReloadCallback = async () => {
-      await this.taskSchedulerService.clearAllScheduledTasks();
-
-      // Close browser action popup before reloading to prevent zombie popup with invalidated context.
-      // The 'reloadProcess' message is sent by ProcessReloadService before this callback runs,
-      // and popups will close themselves upon receiving it. Poll to verify popup is actually closed.
```
quexten, line 1489: Now handled in `browser-process-reload-service`

apps/web/src/app/app.component.ts
```diff
@@ -120,7 +120,6 @@ export class AppComponent implements OnDestroy, OnInit {
           }
           case "locked":
             await this.router.navigate(["/"]);
```
quexten, line 122: Not required since lock service handles this

djsmith85 (approved):

libs/common/src/key-management/process-reload/default-process-reload.service.ts
JaredSnider-Bitwarden, line 1: 🎨 non blocking:  there are no unit tests for any of the process reload service implementations... Might be worth adding.

JaredSnider-Bitwarden (approved):
🥇 Love the new abstraction and refactor. So much cleaner. Reviewed single auth file change + `app.component` logic changes to audit how they interact w/ the auth logout logic there. It all looks good to me. 

Left a non-blocking comment about no service test coverage.

libs/common/src/key-management/process-reload/default-process-reload.service.ts
quexten: Can do that as a follow-up, however i'll say that e2e tests would be more useful here. The services have nearly no complexity in them after the change, and all complexity comes from when process reload is triggered.

.github/codecov.yml
```diff
@@ -41,7 +41,7 @@ component_management:
       name: Key Management - Process Reload
       paths:
         - apps/web/src/app/key-management/services/web-process-reload.service.ts
-        - libs/common/src/key-management/services/default-process-reload.service.ts
+        - libs/common/src/key-management/process-reload/**
```
mzieniukbw, line 44: nit: add browser process reload ?

mzieniukbw (approved):

.github/codecov.yml
quexten: https://github.com/bitwarden/clients/pull/22869

PR #22649, opened 2026-08-25
fix(login-decryption-options): [PM-42457] -…

Patrick-Pimentel-Bitwarden:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-42457

## 📔 Objective

Observed failing here: https://github.com/bitwarden/clients/actions/runs/32513094663/attempts/2?pr=22567 and here: https://github.com/bitwarden/clients/actions/runs/32512713148/attempts/1

Fix the flakey test failing in CI. Through an hour working through this with claude I was **_not_** able to reproduce the same behavior we are seeing in CI locally via lots of test runs performed and tracking of output. There have been legitimate enhancements discovered, namely that an error could leak between the tests with how subscriptions were being handled and that has been resolved.

## Observed failure

```    
==== Pulled from logs ====

2026-08-21T18:23:02.5545318Z FAIL libs/auth tests libs/auth/src/angular/login-decryption-options/login-decryption-options.component.spec.ts (6.053 s)
2026-08-21T18:23:02.5547959Z   ● LoginDecryptionOptionsComponent › shared unlock bootstrap on existing untrusted device › does not trust the device while the account remains locked
2026-08-21T18:23:02.5549178Z 
2026-08-21T18:23:02.5549260Z 
2026-08-21T18:23:02.5549269Z 
2026-08-21T18:23:02.5549278Z 
```

The failure body was **empty** — verified byte-level (`od -c`) on the CI log: the `●` header is followed by four blank lines. No matcher diff, no stack, no console block. Both copies (inline and `Summary of all failing tests`) were empty. Reporters are stock `["default", "jest-junit"]` (`jest.config.js:7`), so nothing stripped it.

## Verified

| # | Claim | How |
|---|---|---|
| 1 | The `not.toHaveBeenCalled()` matcher did not fire | A matcher failure always prints `Expected/Received`. Confirmed twice in this file (mutation check + fault injection both printed full output). |
| 2 | The old Locked test could never fail | `filter(status === Unlocked)` vs stubbed `of(Locked)`; enum `Locked = 1`, `Unlocked = 2`. Mocks rebuilt per test. |
| 3 | The new Locked test can fail | Mutation: inverting the filter to `!==` → `Expected 0 / Received 1`. |
| 4 | New tests are deterministic (no macrotask flush) | 25 randomized-order runs green; `libs/auth` 289/289; `test:types` and lint clean. |
| 5 | Cross-test leakage was real | Injected an `Error` rejection into the bare `.subscribe()` in a test placed first in the file. It failed **"trusts the device and navigates to the vault"** — six tests downstream. |
| 6 | The new `afterEach` closes that leak | Firing the captured `destroyRef.onDestroy` callbacks stops a later `Unlocked` emission from reaching `trustDevice`; commenting the teardown out makes it fail. Both directions checked. |

## Not established

- **The CI failure was never reproduced locally.** 216 runs on pre-fix code — 36 plain/randomized, 180 with `--coverage` under 6-way CPU contention. All green. There is no before/after measurement on the actual symptom.
- **The empty failure body is unexplained.** The leading hypothesis (out-of-band RxJS error attributed to the running test) was falsified: an injected `Error` fails a later test but prints message + full stack (5), and a non-`Error` rejection (`undefined`) fails nothing at all — jsdom logs `Error: Uncaught undefined` and the suite passes. Neither yields an empty body.
- **Whether the cause is in this file.** A worker- or environment-level condition on the runner is not excluded; none of these changes would touch that.

## Why confidence is low-to-moderate that the fixes in this pr solves what we are seeing in our CI

Moderate, not low, because:
- 5 proves the leakage class is real and live, and 6 proves the teardown closes it.
- The component-side `catchError` additions mean errors from those two subscriptions can no longer escape at all — fixed at the source, not relocated.
- 2 / 3 mean the test was untrustworthy on its own terms regardless of the flake.

Not higher, because:
- Nothing was measured against the real symptom (never reproduced).
- The falsified empty-body hypothesis means the actual mechanism is unidentified, so the fix cannot be shown to address it.

## Caveat for reviewers

**A green CI after this change is weak evidence of a fix.** The Locked test no longer yields a macrotask, so it is no longer a viable landing spot for a deferred throw. If something still raises one, the symptom relocates or disappears silently. "It stopped failing" will not distinguish *fixed* from *hidden*.

To convert silence into evidence, ship a detector that fails loudly with the error's `constructor.name`, own property names, `message`, and `stack` for anything reaching Jest's error path in this file.

## Recommendation

Merge on the merits of 2–6 — the test had no teeth and the leakage is real. Treat the CI flake as open until either it recurs (with instrumentation attached) or a detector-backed run goes green.


Patrick-Pimentel-Bitwarden:
Already seeing some changes needed from local claude review. Pushing up some changes.

ike-kottlowski (approved):
By catching the error state and surfacing it we are able to keep the subscribe alive, which should keep the test from failing intermittently.

PR #22654, opened 2026-08-25
Confirm before withdrawing an access request

maxkpower:
## 🎟️ Tracking

No Jira ticket. From a UAT design review.

## 📔 Objective

**The drawer withdrew an access request on a single click, with no confirmation.** Every other surface that withdraws a request, the cipher-view banner and the vault-row menu, already routed through the shared confirmation in `AccessRequestCancelService`, so an already-approved request could be given up by a misclick from the drawer and nowhere else. The My requests rows had the same one-click behaviour, in both tables.

All of them now read "Cancel request", are destructive-styled, and raise the same confirmation with its existing pending-vs-approved wording. `AccessRequestCancelService` gains `cancelRequestById()` and `confirmWithdrawal()`; every entry point shares one options builder, so the dialog copy exists in one place.

Also fixes a second bug: cancelling from the drawer left the list stale until the tab was reloaded. `MyAccessService` now subscribes to `AccessRefreshService.accessChanged$()` in place of its direct `AccessEventService` subscription, so one subscription covers both server push and local mutations. Its own mutations never announce, so nothing reloads twice.

No i18n key is added, removed or changed.

## 📸 Screenshots

To follow, being re-captured against the current tip.

## Notes

- The drawer's failure toast now reads "Couldn't cancel the request." while the list rows keep "Could not cancel request." Both keys pre-exist; worth aligning.
- `DefaultPamNavBadgeService` still subscribes to `AccessEventService` directly, so the nav badge count has the same staleness. Pre-existing and out of scope here.
- The list rows show no spinner during a withdrawal. They use `[disabled]` without `[loading]`, where the drawer uses both.

### DIFF
--- a/bitwarden_license/bit-web/src/app/pam/CLAUDE.md
+++ b/bitwarden_license/bit-web/src/app/pam/CLAUDE.md
@@ -116,12 +116,22 @@ Two services drive every refresh:
 - `AccessEventService` — the server's `RefreshAccessRequest` push, filtered to a bare tick.
 - `AccessRefreshService` — merges that push with this client's own mutations and fans it
   out per cipher, so the cipher-view banner and the gated-cipher reloader react to a local
-  change and a remote one through exactly the same path.
+  change and a remote one through exactly the same path. Subscribing without a cipher id
+  means "whichever item changed, tell me": the reading a surface whose state spans the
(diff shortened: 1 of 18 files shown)


maxkpower:
Closing — we're taking a different approach to the request detail view rather than the drawer pattern.

